PR #10, opened 2026-03-16
fix: use provider-specific default voice when testing TTS in settings

JokerQianwei:
When testing a TTS provider different from the currently active one, the voice from the active provider was sent (e.g. OpenAI's 'alloy' sent to GLM API), causing error 1214. Now falls back to the selected provider's default voice.


cosarah:
Thanks for your bug report and fix

PR #94, opened 2026-03-18
feat: Ollama local LLM support

f1rep0wr:
## Summary
- Add Ollama as a built-in provider (OpenAI-compatible, no API key needed)
- 8 default models: llama3.3, llama3.2, qwen2.5, qwen2.5:32b, mistral, gemma3, deepseek-r1, phi4
- Keyless provider activation: providers with only a baseUrl (no API key) now activate correctly
- `isProviderKeyRequired()` helper for consistent key validation across routes
- i18n strings for Ollama (zh-CN + en-US) and missing Doubao strings
- SSRF guard documentation clarifying server-configured URLs bypass safely

## Security: requiresApiKey no longer accepted from client

Previously, clients could send `x-requires-api-key: false` (header) or `requiresApiKey: false` (body) to skip API key validation for ANY provider, including ones like OpenAI that require keys. This meant a client could consume server-configured credentials without authorization.

Fix: the server now derives `requiresApiKey` exclusively from its own provider registry. The client header and body field are removed from all server-side parsing. For Ollama, the registry says `requiresApiKey: false` so it works keyless. For OpenAI, the registry says `true` so the check is enforced regardless of what a client sends.

## Test plan
- [ ] Configure `OLLAMA_BASE_URL=http://localhost:11434/v1` in `.env.local`, select Ollama model, verify chat works
- [ ] Verify existing providers (OpenAI, Anthropic, etc.) still require API keys
- [ ] Verify sending `x-requires-api-key: false` header does NOT bypass key check for OpenAI
- [ ] Verify classroom generation works with Ollama as `DEFAULT_MODEL`


f1rep0wr:
@wyuc Rebased onto main (95b5c2b), conflicts in spots are all resolved.
- resolveModel now returns providerId on ResolvedModel, so chat/route.ts and classroom-generation.ts use it directly instead of re-parsing the model string
- Removed dead requiresApiKey destructure from verify-model/route.ts to match security fix

cosarah (changes requested):
### Issues

**Important**
- `lib/server/provider-config.ts:170` — Ollama won't activate via env vars alone. The env-var activation path (`loadEnvSection`) still requires `envApiKey` to be set — `if (!envApiKey) continue` skips Ollama when only `OLLAMA_BASE_URL` is configured. The YAML path (line ~138) was correctly updated to `entry?.apiKey || entry?.baseUrl`, but the env-var path was not. Fix: change the activation condition to also accept `envBaseUrl` for providers where `requiresApiKey` is false.

**Minor**
- `lib/server/ssrf-guard.ts` — Comment says server-configured URLs "bypass this check entirely," but the bypass actually happens in `resolve-model.ts` (which only runs SSRF validation on `clientBaseUrl`). Consider moving the note there.

### Summary

Clean security improvement removing the client-controlled `requiresApiKey` bypass, and well-structured Ollama integration. The env-var activation path needs a fix so that `OLLAMA_BASE_URL` alone (without an API key) actually activates the provider — this directly contradicts the `.env.example` setup instructions.

---

Also heads up: main recently landed i18n updates so there's a small merge conflict in the i18n locale fields — should be a straightforward rebase.

PR #111, opened 2026-03-19
fix(generate): persist generated agents in API classrooms

cosarah:
## Summary
- **Bug**: `agentMode=generate` in `/api/generate-classroom` generated agent profiles via LLM but never persisted them to the classroom JSON, so the classroom page couldn't display custom agents
- **Fix**: Added `PersistedAgent` type, enriched server-side agent generation with full display data (avatar, color, priority), persisted agents alongside classroom data, and added `registerPersistedAgents()` for the frontend to load them without IndexedDB

## Test plan
- [x] Call `/api/generate-classroom` with `agentMode: "generate"` and all features enabled
- [x] Verify generated classroom JSON contains `agents` array with full profile data
- [x] Verify `/api/classroom?id=...` returns agents in response
- [x] Verify classroom page loads and displays generated agents in browser


wyuc (changes requested):
Core agent persistence fix looks good. Two things to address:

**1. `engine.ts` TTS fallback bypasses `ttsEnabled` check**

The new condition:
```ts
if ((browserTTSExplicit || !speechAction.audioUrl) && browserTTSAvailable)
```
When `!speechAction.audioUrl` is true, this fires browser TTS even if the user has set `ttsEnabled: false`. Should respect the toggle:
```ts
if ((browserTTSExplicit || (!speechAction.audioUrl && settings.ttsEnabled)) && browserTTSAvailable)
```

**2. `getDefaultAgentsForPersistence()` return type + `registerPersistedAgents()` param type**

Both use inline `Array<{id, name, role, ...}>` instead of the `PersistedAgent` type you already defined. Using `import type { PersistedAgent }` keeps them in sync and avoids drift.

Minor: the engine.ts and roundtable i18n changes are unrelated to agent persistence — would be cleaner as separate commits for bisectability, but not blocking.

wyuc:
Superseded by #336 (merged 2026-04-02) + #373, which fix the same agent-persistence bug with a different approach. The unrelated `engine.ts` TTS change no longer applies — current `main` already gates browser TTS on `ttsEnabled`. Closing. Thanks @cosarah!

PR #123, opened 2026-03-19
fix: prevent video placeholder src from causing 404 requests (#117)

### LINKED ISSUE #117: [Bug]: Video placeholder src causes unnecessary 404 requests
## Description

`<video>` elements render with placeholder IDs (e.g. `gen_vid_1`) as `src` before video generation completes. With `preload="metadata"`, browsers immediately fetch these as relative URLs, resulting in 404s.

## Root Cause

`VideoElement` components set `src={elementInfo.src}` without checking if it's a placeholder (`gen_vid_*` pattern).

## Affected Files

- `components/slide-renderer/components/element/VideoElement/index.tsx`
- `components/slide-renderer/components/element/VideoElement/BaseVideoElement.tsx`

## Suggested Fix

Don't set `src` on `<video>` when the value matches the media placeholder pattern `^gen_(img|vid)_[\w-]+$`.

YizukiAme:
## Summary

Fixes #117 — `<video>` elements render with placeholder IDs (e.g. `gen_vid_1`) as `src`, causing unnecessary 404 requests when `preload="metadata"` triggers a fetch.

### Root Cause

- **`VideoElement/index.tsx`** (edit mode): Sets `src={elementInfo.src}` on `<video>` without checking if it's a media placeholder.
- **`BaseVideoElement.tsx`** (presentation mode): Placeholder detection is gated on `!!stageId`, so contexts without a stageId (e.g. homepage thumbnails) don't recognize placeholders and render them as `<video src>`.

### Fix

- **`VideoElement/index.tsx`**: Add `!isMediaPlaceholder(elementInfo.src)` guard before rendering `<video>`.
- **`BaseVideoElement.tsx`**: Decouple `isPlaceholder` from `stageId` — placeholder detection should always run regardless of context. `stageId` is still used to look up the generation task, but a missing `stageId` no longer causes placeholders to leak into `<video src>`.

Closes #117

### DIFF
--- a/lib/store/settings.ts
+++ b/lib/store/settings.ts
@@ -20,7 +20,7 @@ import { createLogger } from '@/lib/logger';
 const log = createLogger('Settings');
 
 /** Available playback speed tiers */
-export const PLAYBACK_SPEEDS = [1, 1.5, 2] as const;
+export const PLAYBACK_SPEEDS = [1, 1.25, 1.5, 2] as const;
 export type PlaybackSpeed = (typeof PLAYBACK_SPEEDS)[number];
 
 export interface SettingsState {


wyuc (approved):
Clean fix, thanks!

PR #128, opened 2026-03-19
fix: clean up Roundtable UI state after QA/Discussion request failure

### LINKED ISSUE #127: [Bug]: Roundtable UI state not cleaned up after QA/Discussion request failure
### Description

When a `/api/chat` request fails during an active QA or Discussion session (e.g. network error, invalid API key, server 500), the chat panel correctly displays an error message, but **the Roundtable live UI state is never cleaned up**, leaving the interface stuck.

### Symptoms

- Roundtable bubble remains in a blank "live discussion" state and never exits live mode
- Thinking animation (loading dots) may continue spinning indefinitely
- Stop Discussion button remains visible but non-functional
- PlaybackEngine stays in `live` mode, causing toolbar controls to behave incorrectly
- If the user had locally paused playback before the error occurred, the pause UI becomes permanently stuck

The only way for the user to recover is to refresh the page!

### Root Cause

The `catch` blocks in `resumeSession()`, `sendMessage()`, and `startDiscussion()` in `use-chat-sessions.ts` only append an error message to the chat session. They do not notify `stage.tsx` to clean up live UI state, nor do they transition the `PlaybackEngine` out of `live` mode.

Affected state that is not cleaned up:
- **stage.tsx**: `chatIsStreaming`, `chatSessionType`, `thinkingState`, `liveSpeech`, `speakingAgentId`, `isTopicPending`
- **engine.ts**: engine remains in `live` / `paused` (discussion-pending) instead of transitioning back to `idle`

### Steps to Reproduce

1. Start a QA conversation or Discussion
2. While the request is in flight, cause `/api/chat` to fail (e.g. change the API key to an invalid one, or disconnect the network)
3. Observe: chat panel correctly shows an "Error" message
4. Observe: Roundtable remains stuck in the live discussion state

### Expected Behavior

After a request failure:
- Roundtable should exit the live UI (clear bubble, thinking animation, stop button, etc.)
- PlaybackEngine should transition back to `idle` and restore the saved lecture cursor position
- The chat session itself should **not** be hard-closed (it should remain retryable)

### Area

Playback / Chat / Roundtable

YizukiAme:
## Summary

Fixes #127 — When `/api/chat` fails mid-stream during a QA or Discussion session, the Roundtable live UI (bubble, thinking animation, engine mode) is never cleaned up, leaving the interface stuck until page refresh.

## Problem

The `catch` blocks in `resumeSession()`, `sendMessage()`, and `startDiscussion()` only append an error message to the chat panel. They don't notify `Stage` to clean up live UI state, nor do they transition the `PlaybackEngine` out of `live` mode.

This affects users with unstable networks or when the LLM provider returns server errors mid-stream.

## Changes

| File | Change |
|------|--------|
| `use-chat-sessions.ts` | Add `clearLiveSessionAfterError()` helper that cleans current buffer, appends error message, and notifies Stage via `onLiveSessionError` callback |
| `chat-area.tsx` | Pass through `onLiveSessionError` callback |
| `stage.tsx` | Add `handleLiveSessionError()` to reset all live UI state + call `engine.handleDiscussionError()`. Also add `engine.resume()` in `doResumeTopic()` to sync engine mode after soft-pause |
| `engine.ts` | Add `handleDiscussionError()` to exit live mode gracefully without triggering the normal discussion-end flash sequence |

## How it works

```
Request fails in catch block
  → clearLiveSessionAfterError()
    → append error message to chat
    → onLiveSessionError callback
      → Stage: resetLiveState() (clear bubble, thinking, streaming flags)
      → Engine: handleDiscussionError() (live → idle, restore lecture cursor)
```

The session is **not** hard-closed — it remains retryable.

## Verification

- ✅ ESLint: 0 errors on all 4 files
- ✅ TypeScript: only pre-existing `latex-to-omml.ts` error (unrelated)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,3 +34,6 @@ jobs:
 
       - name: TypeScript
         run: npx tsc --noEmit
+
+      - name: Unit Tests
+        run: pnpm test
--- a/app/api/chat/route.ts
+++ b/app/api/chat/route.ts
@@ -59,7 +59,7 @@ export async function POST(req: NextRequest) {
       return apiError('MISSING_REQUIRED_FIELD', 400, 'Missing required field: config.agentIds');
     }
 
-    const { model: languageModel } = resolveModel({
+    const { model: languageModel, apiKey: resolvedApiKey } = resolveModel({
       modelString: body.model,
       apiKey: body.apiKey,
       baseUrl: body.baseUrl,
@@ -67,7 +67,7 @@ export async function POST(req: NextRequest) {
       requiresApiKey: body.requiresApiKey,
     });
 
-    if (!body.baseUrl && !body.apiKey && body.requiresApiKey !== false) {
+    if (!resolvedApiKey && body.requiresApiKey !== false) {
       return apiError('MISSING_API_KEY', 401, 'API Key is required');
     }
 
@@ -112,7 +112,7 @@ export async function POST(req: NextRequest) {
         const generator = statelessGenerate(
           {
             ...body,
-            apiKey: body.apiKey,
+            apiKey: resolvedApiKey,
           },
           signal,
           languageModel,
--- a/app/api/generate/agent-profiles/route.ts
+++ b/app/api/generate/agent-profiles/route.ts
@@ -36,6 +36,8 @@ interface RequestBody {
   sceneOutlines?: { title: string; description?: string }[];
   language: string;
   availableAvatars: string[];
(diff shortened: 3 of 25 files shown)


YizukiAme:
> Approved. There are merge conflicts with main — could you rebase when you get a chance?

Rebased onto latest main — resolved the conflict with #129's livePausedRef (both additions kept, no logical overlap). 
Ready for re-review when you get a chance~

wyuc (approved):

PR #176, opened 2026-03-20
Fix classroom generation language normalization for non-English locales

GRIZZZZZLY:
### Summary
This PR fixes a language handling bug in classroom generation where any non-English language could be coerced to `zh-CN`.

### Problem
The generation pipeline treated `en-US` as a special case and defaulted other languages to Chinese. That caused incorrect output language behavior for languages like Russian.

### Changes
- adds proper normalization for supported generation languages
- adds explicit support for `ru-RU`
- removes hardcoded assumptions that only `zh-CN` and `en-US` are valid generation languages
- improves prompt-level language enforcement in classroom generation

### Why this matters
This fixes a real generation bug and improves the multilingual architecture of the generation pipeline.

### Notes
This PR focuses on generation correctness, not full UI localization.

### DIFF
--- a/components/settings/index.tsx
+++ b/components/settings/index.tsx
@@ -520,6 +520,7 @@ export function SettingsDialog({ open, onOpenChange, initialSection }: SettingsD
         return null;
       case 'pdf': {
         const pdfProvider = PDF_PROVIDERS[selectedPdfProviderId];
+        if (!pdfProvider) return null;
         return (
           <>
             {pdfProvider.icon ? (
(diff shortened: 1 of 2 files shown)


wyuc:
Hey, thanks for this. We have a broader redesign planned for the generation language system (see discussion #333), which will replace the hardcoded language enum entirely. That means changes to `normalizeLanguage` and the generation pipeline here would be short-lived.

Closing this one for now. Your Russian i18n PR #177 is still welcome though, just needs a rebase per the comment there. Thanks!

PR #194, opened 2026-03-21
fix: clean up stale generationSession on failure

### LINKED ISSUE #185: [Bug]: Stale generationSession in sessionStorage causes false Tavily error on retry
### Bug Description

When a classroom generation fails on the `/generation-preview` page (e.g., due to a missing Tavily API key), the stale `generationSession` in `sessionStorage` is **not cleaned up**. 

If the user then refreshes the page or navigates back to `/generation-preview`, the old session (with `webSearch: true`) is reloaded and the web search step is re-triggered — even if the user didn't enable web search this time. This produces a confusing error:

> `Tavily API key is not configured. Set it in Settings → Web Search or set TAVILY_API_KEY env var.`

### Root Cause

In `app/generation-preview/page.tsx`, the `catch` block (line ~727) does **not** call `sessionStorage.removeItem('generationSession')`. The session is only cleaned up on success (line 724) or when the user clicks "Back" (line 747).

### Steps to Reproduce

1. Enable "Web Search" in generation settings (without configuring a Tavily API key)
2. Start classroom generation → it will fail with the Tavily error
3. Go back to the homepage, disable "Web Search", and start a new generation
4. **Refresh** the `/generation-preview` page → the old session with `webSearch: true` is loaded → same Tavily error appears

### Expected Behavior

Failed generation sessions should be cleaned up so that refreshing `/generation-preview` doesn't replay stale settings.

### Suggested Fix

Add `sessionStorage.removeItem('generationSession')` in the `catch` block of `startGeneration()`:

```diff
     } catch (err) {
       if (err instanceof DOMException && err.name === 'AbortError') {
         log.info('[GenerationPreview] Generation aborted');
         return;
       }
+      sessionStorage.removeItem('generationSession');
       setError(err instanceof Error ? err.message : String(err));
     }
```

### Deployment Method

Local development (pnpm dev)

### Browser

Chrome

---

**Note**: This is a one-line fix — just add sessionStorage.removeItem('generationSession') in the catch block.

YizukiAme:
## Summary

Fixes stale `generationSession` in `sessionStorage` causing false Tavily error on retry.

## Problem

When generation fails (e.g. missing Tavily API key), the `generationSession` in `sessionStorage` is **not cleaned up**. On page refresh, the old session with `webSearch: true` is reloaded and the web search step is re-triggered, producing a confusing Tavily error even if web search was disabled for the new attempt.

## Fix

Added `sessionStorage.removeItem('generationSession')` in the `catch` block of `startGeneration()`, so failed sessions are cleaned up and don't replay stale settings.

## Testing

- One-line change, no side effects
- `tsc --noEmit` passes

Closes #185


YizukiAme:
> Thanks for the fix. One question before merging:
> 
> In the normal user flow, if someone goes back to the homepage and reconfigures (e.g. disables web search), `page.tsx:295` overwrites `sessionStorage` synchronously before `router.push`, so the old session is always replaced. And if they just refresh `/generation-preview`, that's a retry with the same config — getting the same Tavily error is expected, not "false."
> 
> Could you clarify the exact reproduction path where a user disables web search but still hits the stale session? I might be missing something.
> 
> The cleanup itself is good hygiene regardless — just want to make sure the issue description matches the actual bug.

Thanks for the thorough review!! I originally ran into this during development — I was getting an unexpected Tavily error even though I hadn't configured or enabled web search at all, and after digging through DevTools I found that manually clearing generationSession from sessionStorage resolved it. It happened a few more times after that, so I suspected it was caused by a previously Tavily-enabled session lingering around. 🤔

The reproduction steps in the issue may not be fully accurate — the root cause might be tied to development-specific scenarios that I can't pinpoint exactly right now...? But after applying this one-line fix, I haven't run into the issue again, and the change is minimal.

If you'd like, I can try to track down and reproduce the exact trigger more rigorously. 🙃

PR #213, opened 2026-03-22
fix(storage): local disk persistence for classroom data generated via browser UI flow (#53)

### LINKED ISSUE #53: [Bug] 课程数据未持久化到服务器磁盘
### 版本
OpenMAIC v0.1.0

### 问题描述
本地部署模式下，生成的课程数据未能持久化保存到服务器磁盘。导致：
1. 换浏览器后无法访问课程
2. 服务器重启后课程丢失

### 重现步骤
1. 运行 `pnpm dev` 启动 OpenMAIC
2. 创建 data/classrooms 目录
3. 生成一个课程
4. 检查 data/classrooms/ 目录为空
5. 换浏览器访问课程链接，显示 "Classroom not found"

### 预期行为
课程数据应该保存到 data/classrooms/{id}.json

### 实际行为
data/classrooms/ 目录始终为空，API 返回 404

### 环境
- Node.js: v22.22.0
- Next.js: 16.1.2
- 操作系统: Linux

AsadAliQuant:
## Summary

Fixes #53 — classrooms generated via the browser UI are now persisted to `data/classrooms/{id}.json` on the server disk, so the classroom URL resolves in any browser/device, not just the one that generated it.

## Related Issues

Closes #53

## Root Cause

OpenMAIC has two generation flows:

- **API/OpenClaw flow** — runs server-side via `classroom-generation.ts`, already called `persistClassroom()` correctly
- **Browser UI flow** — runs client-side, saved to IndexedDB only and never wrote to disk

This meant every classroom created through the normal browser UI was only accessible in the browser that generated it — switching to another browser shows "Classroom not found" because IndexedDB is local to each browser.

## Changes

- `lib/hooks/classroom-persistence.ts` — new `persistClassroomToServer()` helper wrapping `POST /api/classroom` (which triggers the existing atomic `persistClassroom()` disk write). A write is only reported successful after checking `response.ok` **and** the API body — non-2xx, `success: false`, and network errors all return a failure result.
- `app/generation-preview/page.tsx` — persists right after the first scene lands, before navigating to the classroom, so the URL resolves cross-browser *during* generation. Reads a fresh store snapshot after `addScene` so the payload always contains the new scene.
- `lib/hooks/use-scene-generator.ts` — re-persists the full deck on main generation completion and on retry-driven completion (integrated with the current generation state machine: `generationEpoch`, `markGenerationCompleteIfDone`).
- `tests/lib/hooks/classroom-persistence.test.ts` — vitest coverage: 2xx success, non-2xx failure, `success: false` on a 2xx status, HTTP-status fallback when the body isn't JSON, and fetch rejection.

## Persistence policy

Persistence is best-effort and non-fatal: generation never fails because of a failed write; failures are logged as warnings. Success is never reported without validating the response.

## Type of Change

## Verification

**Steps to reproduce / test**

1. Generate a course via the browser UI
2. As soon as the app navigates to `/classroom/{id}`, check that `data/classrooms/{id}.json` exists on disk (before generation finishes)
3. Open the classroom URL in a different browser — should load mid-generation
4. After completion, refresh the second browser — full deck is available

**What I personally verified**

- Generated course in Chrome — `data/classrooms/{id}.json` created with full course data
- Opened same URL in Edge — loaded correctly from disk
- Failed writes log a warning and generation completes normally

**Evidence**

- [x] `pnpm lint`, `npx tsc --noEmit`, and vitest pass
- [x] Manually tested locally

## Notes

Zero new dependencies. No schema changes.

Opening a persisted course in another browser requires pasting the direct classroom URL (e.g. `http://localhost:3000/classroom/{id}`). The home page course list still reads from IndexedDB only — showing disk-persisted courses on the home page is a separate improvement.

### DIFF
--- a/app/classroom/[id]/page.tsx
+++ b/app/classroom/[id]/page.tsx
@@ -52,6 +52,14 @@ export default function ClassroomDetailPage() {
         isCurrent,
         loadFromStorage,
         getCurrentStage: () => useStageStore.getState().stage,
+        getLocalDeck: () => {
+          const state = useStageStore.getState();
+          return {
+            sceneCount: state.scenes.length,
+            outlineCount: state.outlines.length,
+            generationComplete: state.generationComplete,
+          };
+        },
         fetchClassroom: defaultClassroomLoadDeps.fetchClassroom,
         applyFallbackScenes: (args) =>
           defaultClassroomLoadDeps.applyFallbackScenes({
--- a/app/generation-preview/page.tsx
+++ b/app/generation-preview/page.tsx
@@ -47,6 +47,7 @@ import type {
   SessionDocumentSource,
 } from '@/lib/types/generation';
 import { AgentRevealModal } from '@/components/agent/agent-reveal-modal';
+import { persistClassroomToServer } from '@/lib/hooks/classroom-persistence';
 import { createLogger } from '@/lib/logger';
 import {
   type GenerationSessionState,
@@ -1076,6 +1077,23 @@ function GenerationPreviewContent() {
 
       sessionStorage.removeItem('generationSession');
       await store.saveToStorage();
+
+      // Persist to server disk so the URL resolves cross-browser during generation (#53).
+      // Fresh snapshot: `store` predates addScene, its scenes are stale.
+      const fresh = useStageStore.getState();
+      if (fresh.stage) {
+        void persistClassroomToServer(fresh.stage, fresh.scenes).then((result) => {
+          if (result.success) {
+            log.info('[GenerationPreview] First scene persisted to server disk:', fresh.stage?.id);
+          } else {
+            log.warn(
+              '[GenerationPreview] Failed to persist first scene to server disk:',
+              result.error,
+            );
+          }
+        });
+      }
+
       router.push(`/classroom/${stage.id}`);
     } catch (err) {
       setIsOutlineStreaming(false);
--- a/lib/classroom/load-classroom.ts
+++ b/lib/classroom/load-classroom.ts
@@ -39,6 +39,11 @@ export interface RunClassroomLoadArgs<TMediaTasks = unknown, TGeneratedAgentReco
   isCurrent: () => boolean;
   loadFromStorage: (classroomId: string, loadToken: StageSceneLoadToken) => Promise<void>;
   getCurrentStage: () => Stage | null;
+  getLocalDeck: () => {
+    sceneCount: number;
+    outlineCount: number;
+    generationComplete: boolean;
+  };
   fetchClassroom: (classroomId: string) => Promise<ClassroomPayload | null>;
   applyFallbackScenes: (args: {
     loadToken: StageSceneLoadToken;
@@ -68,6 +73,7 @@ export async function runClassroomLoad<TMediaTasks = unknown, TGeneratedAgentRec
   isCurrent,
   loadFromStorage,
   getCurrentStage,
+  getLocalDeck,
   fetchClassroom,
   applyFallbackScenes,
   saveGeneratedAgents,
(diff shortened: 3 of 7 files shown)


wyuc (changes requested):
The persistence result is reported incorrectly on failures. `fetch()` resolves normally for HTTP 4xx/5xx responses, but this code never checks `response.ok` or the API result and unconditionally logs `Classroom persisted to server disk`. A disk/write/validation failure therefore looks successful while cross-browser loading still fails.

Please rebase onto current `main` (the generation completion/failure state machine has changed), check the response before declaring success, and add a test for both success and non-2xx behavior. Please also coordinate the scope with #218 so we land one deliberate browser-to-server persistence flow rather than two overlapping implementations. We won't resolve the branch conflict on the maintainer side.

AsadAliQuant:
Rebased onto current main and addressed the review:

- persistClassroomToServer now checks response.ok and the API body (success: false counts as failure even on 2xx), and treats network errors as failures. A failed write is logged as a warning and is never reported as success.
- Added vitest coverage: 2xx success, non-2xx, success: false on 2xx, HTTP-status fallback when the body isn't JSON, and fetch rejection.
- Persistence policy: best-effort and non-fatal, generation still completes if the write fails, and failures surface as console warnings.

On scope coordination with #218: we converged on the same design (POST /api/classroom writing data/classrooms/{id}.json), so I've proposed consolidating there. 

This PR now also persists right after the first scene lands in generation-preview (the early-access case #218 targeted), plus on main completion and retry-driven completion, reading store.getState() after the scene is added so the payload always includes the latest scene. 

Happy to adjust if you'd prefer a different split.

wyuc (changes requested):
Re-reviewed head `bc51a04`. The prior response-handling and stale-snapshot findings are fixed; the focused local tests pass (23/23 total).

One correctness blocker remains: the early first-scene persistence is fire-and-forget, while completion/retry writes can start independently for the same stage ID. `/api/classroom` atomically replaces the same JSON file but does not version or serialize snapshots. If the earlier one-scene request completes after the final request, it overwrites the full deck with the stale snapshot.

Please serialize/coalesce persistence per stage (or otherwise ensure older snapshots cannot commit after newer ones) and add a deterministic deferred-fetch regression proving the final on-disk payload cannot be replaced by the first-scene payload.

Because #53 is specifically a cross-browser behavior, please also report one real flow on the final SHA: generate in browser A, load after the first scene in a fresh browser/context, wait for completion, then refresh/load again and verify the full scene set. Unit tests of the helper alone do not prove the integration path fires and remains readable.

AsadAliQuant:
Thanks for the detailed review — both points are addressed.

**Write race:** persistence is now serialized per stage, so an older snapshot can never overwrite a newer one on disk. Added a regression test that deliberately delays the first-scene write until after the final write and confirms the full deck is what ends up on disk.

**Real cross-browser flow:** I ran the actual scenario from #53 — generated a 4-scene course in Chrome, opened the link in Firefox right after the first scene appeared, waited for generation to finish, then refreshed Firefox. The full 4-scene deck loaded correctly. Also confirmed the same in Edge and a private window.

Running the real flow caught one more bug that unit tests alone would have missed: the viewer browser cached the partial one-scene copy locally and kept re-reading its own stale copy on every refresh instead of asking the server again, exactly the "generated in one browser, not loadable in another" behavior this issue is about. Fixed: a viewer copy now re-checks the server on load and picks up the fuller deck, while the generating browser and locally edited decks are never touched. A previously stuck viewer heals itself on its next load, which I verified too. Regression tests cover this as well.

One expected limitation, for completeness: the end-of-course completion screen doesn't show in viewer browsers, because the server snapshot doesn't carry the completion flag. All real scenes load fine. Happy to persist that flag server-side in a follow-up if you'd like.

The branch is rebased onto current main and squashed to a single commit. Tests and typecheck are green locally, and CI is green on the current head.

wyuc:
Thank you for the persistence work and, especially, for following through on multiple rounds of review with careful fixes, deterministic race coverage, and a real cross-browser verification. I’m sorry that we kept asking you to refine the implementation before stepping back and reconciling the PR with the repository’s newer storage direction. We should have made that architectural check earlier; the extra iteration was not your fault.

The underlying problem in #53 is real, and this branch demonstrates the expected cross-browser behavior well. However, the storage architecture that has since landed is converging on separate pluggable contracts for course documents (`DocumentStore`), learner state (`RuntimeStore`), and binary assets (`StorageProvider`). Adding browser-triggered writes directly to `/api/classroom` and treating `data/classrooms/{id}.json` as a second document source would create a parallel persistence protocol outside those contracts.

That parallel path would also leave us with two authorities—browser IndexedDB and server JSON—without a shared revision/conflict model, while persisting only stage/scenes rather than the complete document, runtime, and asset lifecycle. We would therefore need to migrate this implementation again when the server/HTTP document backend is wired into the app.

For that reason, we’re going to close this PR without merging and keep #53 open to be addressed through the unified document-storage path. This is an architecture decision, not a reflection on the quality or effort of the implementation. Thank you again for the thoughtful work and for helping expose the important stale-viewer and write-order edge cases; those findings will be valuable in the mainline solution.

PR #218, opened 2026-03-23
fix: persist classroom to server for cross-device access

reddingit:
## Summary

  Fixes 404 errors when accessing classrooms from different devices on the same network. Previously, classrooms were only stored in browser IndexedDB, making them inaccessible from other devices. This change persists classroom data to server-side storage, enabling cross-device access via the API.

  ## Related Issues

  N/A (new fix)

  ## Changes

  - **`app/generation-preview/page.tsx`**: Added server-side persistence call after first scene generation, before navigation to classroom page
  - **`app/classroom/[id]/page.tsx`**: Added server-side update call in `onComplete` callback when all scenes finish generation
  - Data is now saved to `data/classrooms/{id}.json` on the server filesystem
  - Falls back gracefully if server persistence fails (continues with local-only storage)

  ## Type of Change

  ## Verification

  ### Steps to reproduce / test

  1. Start dev server: `pnpm dev`
  2. On Device A (localhost), create a new course and wait for first scene to complete
  3. On Device B (LAN), access `http://<Device-A-IP>:3000/classroom/<id>`
  4. Verify classroom loads successfully instead of 404
  5. Wait for all scenes to complete generation
  6. Refresh on Device B to verify full content is available

  - Code follows existing patterns in the codebase (uses existing `/api/classroom` endpoint)
  - Error handling is robust (try-catch with fallback behavior)
  - No breaking changes - existing single-device workflow unaffected
  - TypeScript types are correct

  ### Evidence

  - [ ] Manually tested locally (pending user verification)
  - [ ] Screenshots / recordings attached (N/A - backend behavior change)

  ## Checklist

### DIFF
--- a/app/classroom/[id]/page.tsx
+++ b/app/classroom/[id]/page.tsx
@@ -27,8 +27,27 @@ export default function ClassroomDetailPage() {
   const generationStartedRef = useRef(false);
 
   const { generateRemaining, retrySingleOutline, stop } = useSceneGenerator({
-    onComplete: () => {
+    onComplete: async () => {
       log.info('[Classroom] All scenes generated');
+      // Update server-side storage with all generated scenes
+      const state = useStageStore.getState();
+      if (state.stage && state.scenes.length > 0) {
+        try {
+          const resp = await fetch('/api/classroom', {
+            method: 'POST',
+            headers: { 'Content-Type': 'application/json' },
+            body: JSON.stringify({
+              stage: state.stage,
+              scenes: state.scenes,
+            }),
+          });
+          if (resp.ok) {
+            log.info('[Classroom] Updated server-side storage with all scenes');
+          }
+        } catch (err) {
+          log.warn('[Classroom] Failed to update server-side storage:', err);
+        }
+      }
     },
   });
 
--- a/app/generation-preview/page.tsx
+++ b/app/generation-preview/page.tsx
@@ -721,6 +721,29 @@ function GenerationPreviewContent() {
         }),
       );
 
+      // Persist classroom to server-side storage for cross-device access
+      const state = store;
+      try {
+        const persistResp = await fetch('/api/classroom', {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({
+            stage: state.stage,
+            scenes: state.scenes,
+          }),
+        });
+        if (persistResp.ok) {
+          const persistData = await persistResp.json();
+          log.info('[Generation] Classroom persisted to server:', persistData.url);
+        } else {
+          log.warn(
+            '[Generation] Failed to persist classroom to server, continuing with local-only storage',
+          );
+        }
+      } catch (persistErr) {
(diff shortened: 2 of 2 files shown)


XerxesSimon:
@reddingit
Thank you so much for your optimized code. I have completed it according to the code you wrote, but now I encounter a problem: how can the front-end data trigger the generated PPT content to be saved to the back-end and then saved as {id}.json? I accidentally triggered the save once before, but I can no longer find the way to do it afterward. I hope you can give me some guidance when you have time.

wyuc (changes requested):
The first persistence write uses a stale Zustand snapshot and does not contain the scene it claims to publish.

`const store = useStageStore.getState()` captures the current state object; `store.addScene(data.scene)` replaces Zustand state, but the later `const state = store` still points to the old object. Therefore `state.scenes` does not include the just-generated first scene. A second device may stop getting a 404, but it receives an empty/incomplete classroom until the later completion write.

Please update onto current `main`, read `useStageStore.getState()` after the scene is added, handle non-2xx responses consistently, and add a focused integration test showing that the POST payload includes the newly generated scene. Since this changes server-side data retention for every browser-generated classroom, please also make the intended persistence/config policy explicit rather than relying only on a silent best-effort write.

AsadAliQuant:
Hey @reddingit 👋, wyuc asked us to coordinate scope since both PRs fix the same bug (classrooms only living in the browser that generated them). 

We converged on the same design: POST /api/classroom, with the server writing data/classrooms/{id}.json.

I'd suggest consolidating into #213, which now covers both scopes:

- The persistence helper checks response.ok and the API body, so failed writes are never reported as success.
- The early write right after the first scene (your generation-preview approach) is included, and it reads store.getState() after the scene is added, avoiding the stale-snapshot issue wyuc flagged here.
- Persistence also runs on main completion and on retry-driven completion, with vitest coverage for the success/failure paths.
- Nothing from your cross-device scenario is lost, the write goes to server disk, so other devices hitting the same server can load the classroom. 

Would you be OK closing this in favor of #213, or would you rather split it differently? 

Either way, thanks for digging into the same bug 🙏

wyuc:
Thank you for working on the cross-device classroom problem and for coordinating the overlapping scope with #213. I’m sorry this remained open while the repository’s storage direction evolved; we should have closed the architectural loop sooner.

The problem is valid, but both branches converge on the same browser-to-`/api/classroom` flow backed by `data/classrooms/{id}.json`. The storage architecture that has since landed is moving course documents, learner runtime, and binary assets behind the separate pluggable `DocumentStore`, `RuntimeStore`, and `StorageProvider` contracts. Landing this disk-JSON path now would introduce a second document authority alongside browser storage without the shared revision, migration, conflict, and asset semantics that the unified path is intended to provide.

We have therefore closed #213 and will close this PR for the same reason, rather than ask either contributor to keep refining an implementation we would later migrate away from. This is an architecture decision, not a judgment on the value of the use case or the effort here. Thank you again for pushing the cross-device scenario forward; it will remain part of the requirements for the unified server-backed document solution.

PR #221, opened 2026-03-23
fix: use resolved API key in chat route instead of client-sent key

### LINKED ISSUE #220: fix: server-configured API key ignored in chat route after #114 refactor
## Bug Description

PR #114 (commit d0c324c) refactored `app/api/chat/route.ts` to use `resolveModel()` but broke the server-side API key fallback.

### Before (correct behavior):

```ts
const effectiveApiKey = resolveApiKey(providerId, body.apiKey);
if (!effectiveApiKey) { return 401; }
// ...
statelessGenerate({ ...body, apiKey: effectiveApiKey }, ...)
```

### After (broken behavior):

```ts
const { model: languageModel } = resolveModel({ apiKey: body.apiKey, ... });
if (!body.baseUrl && !body.apiKey && body.requiresApiKey !== false) { return 401; }
// ...
statelessGenerate({ ...body, apiKey: body.apiKey }, ...)
```

### Root Cause

Two issues:

1. **401 check uses `body.apiKey` instead of the resolved key.** For server-configured providers (where the API key is set via environment variable like `OPENAI_API_KEY`), `body.apiKey` is empty/undefined because the client doesn't send it. The old code checked `effectiveApiKey` (which included the server fallback), so it worked. The new heuristic `!body.baseUrl && !body.apiKey && body.requiresApiKey !== false` incorrectly returns 401 for server-configured providers.

2. **`statelessGenerate` receives `body.apiKey` instead of the resolved key.** Even if the 401 check were fixed, the downstream code would get the empty client-sent key instead of the server-configured one.

### Impact

Any deployment relying on server-side API keys (via environment variables) will get 401 errors on the chat endpoint, even though the keys are properly configured.

### Fix

`resolveModel()` in `lib/server/resolve-model.ts` already computes the effective API key internally (line 41-43) but discards it. The fix should:

1. Add `apiKey: string` to the `ResolvedModel` interface
2. Return the resolved `apiKey` from `resolveModel()`
3. Update `app/api/chat/route.ts` to use the returned `apiKey` for the 401 check and pass it to `statelessGenerate`
4. Audit other routes using `resolveModel` for the same issue

wyuc:
## Summary

- **Fixes a regression from PR #114** where the chat route (`app/api/chat/route.ts`) used the raw client-sent `body.apiKey` instead of the server-resolved API key for both the 401 authentication check and the `statelessGenerate` call
- **Adds `apiKey` field to `ResolvedModel` interface** in `lib/server/resolve-model.ts` so that callers can access the effective API key after server-side env-var fallback resolution
- Server-configured providers (e.g. `OPENAI_API_KEY` env var) now correctly authenticate again on the chat endpoint

Fixes #220

## Test plan

- [ ] Deploy with a server-configured API key (env var only, no client-side key) and verify the chat endpoint returns a streaming response instead of 401
- [ ] Deploy with a client-provided API key and verify it still works as before
- [ ] Verify other API routes (`/api/generate/*`, `/api/pbl/chat`, `/api/quiz-grade`, `/api/verify-model`) are unaffected (they were already correct)

### DIFF
--- a/app/api/chat/route.ts
+++ b/app/api/chat/route.ts
@@ -67,7 +67,7 @@ export async function POST(req: NextRequest) {
       requiresApiKey: body.requiresApiKey,
     });
 
-    if (!resolvedApiKey) {
+    if (!resolvedApiKey && body.requiresApiKey !== false) {
       return apiError('MISSING_API_KEY', 401, 'API Key is required');
     }
 


wyuc:
@cosarah Please review

PR #232, opened 2026-03-23
fix: clear stale modelId/providerId after server provider sync

### LINKED ISSUE #226: bug: stale modelId/providerId after server provider sync
## Description

When the server removes a model or provider from its configuration, `fetchServerProviders()` correctly updates `providersConfig` (model lists, `isServerConfigured` flags) but **never validates the current `modelId`/`providerId` selection** against the new state. This leaves stale values in the store that point to unavailable models/providers.

## Reproduction

1. User selects `gpt-4o-mini` as their model
2. Server admin removes `gpt-4o-mini` from `OPENAI_MODELS`
3. After the next `fetchServerProviders()` call, `modelId` still reads `gpt-4o-mini` even though it's no longer in the available models list

## Expected Behavior

After server sync, if the selected model/provider is no longer available:
- `modelId` should be cleared (or reset to a valid default)
- `providerId` should be cleared if the entire provider becomes unusable (no server config AND no client key)

## Affected Scenarios

1. Server removes the selected model from allowed list → `modelId` stale
2. Server removes entire provider, no client key → `providerId` + `modelId` stale  
3. Server narrows model list, selected model excluded → `modelId` stale

## Test Coverage

These scenarios are documented as `it.fails()` tests in `tests/store/settings-server-sync.test.ts` (PR #144). Once fixed, the `.fails()` modifier should be removed and the tests will pass green.

## Suggested Fix

In `fetchServerProviders()` (in the `set()` callback), after merging provider configs, check if:
- `state.modelId` is still in the updated provider's model list
- `state.providerId` still points to a usable provider

If not, clear the stale selection.

Bortlesboat:
## Summary
- After `fetchServerProviders()` merges updated provider config, validate that `state.modelId` and `state.providerId` still reference usable entries
- If the selected provider lost its server config and has no client API key, clear both `providerId` and `modelId`
- If the selected model was removed from the provider's filtered model list, clear `modelId`

Fixes #226

## Test plan
- [x] All 3 pre-written `.fails()` tests now pass — removed `.fails()` markers
- [x] All 12 existing tests continue to pass (15/15 total)
- [x] Linting and formatting pass


wyuc (approved):
Verified the fix. Follow-up improvements tracked in #263.

PR #235, opened 2026-03-24
fix: improve whiteboard UX with bounded canvas and history fixes

### LINKED ISSUE #228: Optimize whiteboard: bounded canvas, toolbar cleanup, always-on pan, and snapshot/history UX
## Overview

Several UX and logic improvements for the whiteboard component to make it feel more polished and intentional.

---

### 1. Preserve whiteboard boundaries

The current implementation turns the whiteboard into a borderless infinite canvas. We want to keep the whiteboard edges visible — zoom and pan should work **within a bounded area** rather than making it feel unbounded.

### 2. Move the Reset View button

Instead of a standalone button, place it in the **top-right corner** alongside the existing "clear whiteboard" and "minimize whiteboard" buttons for a cleaner toolbar layout.

### 3. Always-on drag/pan

Keep the canvas **always draggable** (no toggle needed), and use cursor states (`grab` / `grabbing`) to signal this to the user.

### 4. Snapshot trigger logic

Right now snapshots are pushed both on clear and on a 2-second debounce after any edit. The debounce path tends to accumulate a lot of intermediate states that users don't really need to restore.

**Proposed change:** Only snapshot before destructive operations (clear / AI clear), since the main use case is *"the board got wiped, I want it back."*

### 5. History panel interaction

The current slide-out panel feels disconnected from the history button. A **popover/dropdown anchored to the button** would feel more natural and match the "click outside to dismiss" behavior users already expect.

cosarah:
## Summary

Refactor the whiteboard canvas to feel bounded (like a real whiteboard) instead of infinite, simplify the toolbar, and fix history snapshot logic to prevent data loss.

## Related Issues

Closes #228

## Changes

- Refactor canvas to a 3-layer architecture: viewport (pointer events) → bounded canvas (white rect, pan/zoom) → content layer (autoFit)
- Add reset view button in toolbar with fade animation (replaces old in-canvas overlay button)
- Enable always-on pan — no toggle needed, cursor is always grab/grabbing
- Add pan boundary clamping (half canvas size in each direction)
- Move history panel from absolute-positioned overlay to popover anchored to history button
- Save current whiteboard content before restore to prevent data loss
- Fix snapshot deduplication to check all history entries, not just the last one (prevents history bloat when switching between two snapshots)
- Remove dead code: `restoredKey` mechanism, `zoomHint` i18n keys, zoom hint overlay
- Use `forwardRef` + `useImperativeHandle` to expose `resetView` from canvas to toolbar

## Type of Change

### Steps to reproduce / test

1. Open whiteboard, let AI add some elements
2. Pan (drag) and zoom (scroll wheel) the canvas — verify the white rectangle moves as a bounded area
3. Verify reset view button appears in toolbar when view is modified, click to reset
4. Open history, restore a snapshot — verify current content is saved as "Before restore"
5. Switch between two history entries repeatedly — verify no duplicate snapshots are created

- Bounded canvas panning and zooming with boundary clamping
- Reset view button fade animation
- History panel popover positioning
- Restore saves current content before overwriting
- Snapshot deduplication across all entries
- Double-click to reset view still works

### Evidence

- [x] CI passes (`pnpm format && pnpm lint && npx tsc --noEmit`)
- [x] Manually tested locally

https://github.com/user-attachments/assets/2e7c8479-42ef-4fd4-b3ee-324d62f17535

## Checklist


wyuc (changes requested):
Reviewed the canvas refactoring and history changes. The history simplification (removing `restoredKey`, save-before-restore) is clean. One architectural issue needs addressing before merge.

### Viewport clips the bounded canvas on pan

The viewport is sized to `canvasWidth × canvasHeight` with `overflow-hidden`. The bounded canvas inside is the same size. Any pan immediately clips the white rectangle at the viewport boundary rather than at the outer container edge.

```
[Content Area]  overflow-hidden         ← should be the only clip boundary
  [container]   overflow-hidden         ← clips
    [wrapper]   exact scaled size       ← constrains
      [viewport] 1000×562.5, overflow-hidden  ← clips again
        [bounded canvas] 1000×562.5     ← moves with pan/zoom
```

The viewport should fill the available workspace and only handle pointer events. The bounded canvas moves freely inside it, clipped only by the outermost container:

```
[Content Area]  overflow-hidden         ← sole clip boundary
  [viewport]    w-full h-full           ← fills workspace, no overflow-hidden
    [bounded canvas] fixed size         ← centered, moves with pan/zoom
      [content]                         ← autoFit
```

Reproduced on a narrow window. The white rectangle's top/right edge is cut off by the viewport, not by the outer container.

Once the viewport fills the workspace, two related things to revisit:

- **Pan boundaries** should account for zoom level. Currently `maxPanX = canvasWidth / 2` is fixed. At high zoom the user can't reach the edges, at low zoom the canvas drifts too far.
- **Zoom origin**. Currently zooms around canvas center regardless of cursor position. Zoom-toward-cursor is more natural for a canvas tool.

wyuc (approved):
Viewport clipping fixed, zoom-toward-cursor and zoom-aware pan boundaries both landed. Verified on narrow viewport. CI green.

PR #266, opened 2026-03-25
fix: validate all provider selections after server sync

### LINKED ISSUE #263: Validate all provider selections after server sync, not just LLM
## Context

#232 fixed stale `modelId`/`providerId` for LLM providers after `fetchServerProviders()`. The same class of bug exists for other provider types.

## What's left

**1. TTS, ASR, PDF, Image, Video providers have the same stale selection problem.**

After server sync, if a TTS/ASR/Image/Video provider loses its `isServerConfigured` flag, the corresponding `ttsProviderId`, `asrProviderId`, etc. still point to the now-unusable provider. Same root cause as #226, just for non-LLM provider types.

**2. Auto-select a valid fallback instead of clearing to empty.**

Currently #232 clears stale selections to `""`. The user then has to manually pick a new model. For a classroom product, it would be better to clear the stale value and then run the existing auto-select logic to pick the next available option silently.

**3. Clean up the `'' as ProviderId` type cast.**

The fix uses `'' as ProviderId` to represent "no selection". If `ProviderId` doesn't include empty string, this should be addressed at the type level rather than with a cast.

## Scope

- Extend the validation logic from #232 to cover `ttsProviderId`, `asrProviderId`, `pdfProviderId`, `imageProviderId`/`imageModelId`, `videoProviderId`/`videoModelId`
- After clearing a stale selection, attempt auto-select instead of leaving it empty
- Fix the ProviderId type to support empty/unset state cleanly

Ref: #226, #232

wyuc:
## Summary
- Extract `isProviderUsable`, `validateProvider`, `validateModel` into `lib/store/settings-validation.ts`
- Replace PR #232's LLM-only inline validation with unified validation for all 6 provider types (LLM, TTS, ASR, PDF, Image, Video)
- Usability checks `isServerConfigured || apiKey` — does NOT use `requiresApiKey` (e.g. mineru needs server baseUrl, not API key)
- Auto-select fallback instead of clearing to empty: TTS→browser-native-tts, ASR→browser-native, PDF→unpdf
- Image/Video: auto-recover provider+model from empty state when server adds providers; auto-disable generation when all providers removed; prevent re-enabling until a provider becomes available
- No auto-enable on recovery — user controls when to turn on generation
- PDF providers now recognized with baseUrl-only config (`requiresBaseUrl` option in `loadEnvSection`)
- Remove "setup needed" prompt from header and home page (auto-select makes it unnecessary)
- Clean up dead setup-indicator CSS animations

Closes #263

## Test plan
- [x] 16 unit tests for validation functions (isProviderUsable, validateProvider, validateModel)
- [x] 34 integration tests for stale selection across all provider types
- [x] 23 server provider-config tests including baseUrl-only PDF scenarios
- [x] 74 total tests pass, typecheck clean, lint clean
- [x] Manual verification: provider add/remove/restore cycles for LLM, Image, TTS

### DIFF
--- a/lib/store/settings.ts
+++ b/lib/store/settings.ts
@@ -910,29 +910,24 @@ export const useSettingsStore = create<SettingsState>()(
               }
 
               // === Validate current selections against updated configs ===
-              const llmFallback = [
-                ...Object.entries(newProvidersConfig)
+              // Build fallback: server-configured first, then client-key-only
+              const buildFallback = <T extends string>(
+                config: Record<string, { isServerConfigured?: boolean; apiKey?: string }>,
+              ): T[] => [
+                ...Object.entries(config)
                   .filter(([, c]) => c.isServerConfigured)
-                  .map(([id]) => id as ProviderId),
-                ...Object.entries(newProvidersConfig)
+                  .map(([id]) => id as T),
+                ...Object.entries(config)
                   .filter(([, c]) => !c.isServerConfigured && !!c.apiKey)
-                  .map(([id]) => id as ProviderId),
+                  .map(([id]) => id as T),
               ];
-              const ttsFallback = Object.keys(newTTSConfig).filter(
-                (id) => newTTSConfig[id as TTSProviderId]?.isServerConfigured,
-              ) as TTSProviderId[];
-              const asrFallback = Object.keys(newASRConfig).filter(
-                (id) => newASRConfig[id as ASRProviderId]?.isServerConfigured,
-              ) as ASRProviderId[];
-              const pdfFallback = Object.keys(newPDFConfig).filter(
-                (id) => newPDFConfig[id as PDFProviderId]?.isServerConfigured,
-              ) as PDFProviderId[];
-              const imageFallback = Object.keys(newImageConfig).filter(
-                (id) => newImageConfig[id as ImageProviderId]?.isServerConfigured,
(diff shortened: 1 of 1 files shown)


cosarah (changes requested):
### Issues

**Important**

`settings.ts:925-939` — Fallback arrays for TTS/ASR/PDF/Image/Video only include server-configured providers, while LLM's fallback also includes providers with client-side API keys. This means if a user manually sets a client API key for a non-LLM provider (e.g. OpenAI TTS) and the server config for that provider is later removed, the fallback will skip it and force-switch to browser-native — even though the client key is still valid. Consider unifying the fallback strategy across all provider types.

**Minor**
- `settings.ts:1150-1171` — Auto-select spreads after validation on first load, silently overriding validated results. A comment documenting the intended precedence would help maintainability.
- `settings-validation.ts:30` / `settings.ts:977` — `validateProvider('')` short-circuits to `''`, forcing auto-recover to duplicate the "pick first fallback" logic.
- `settings.ts:1000-1003` — `|| imageModels[0]?.id` is redundant with `validateModel`'s own fallback.

### Summary

Solid PR — clean extraction of validation helpers, unified approach for all 6 provider types, thorough test coverage. Main concern is the inconsistent fallback strategy for non-LLM providers that could drop valid client-key configurations.

PR #319, opened 2026-03-28
fix: chat auto-scrolls to top after thinking completes

xoknight:
## Summary
- Fix streaming auto-scroll effect that was a no-op: `scrollContainerRef` has no fixed height and never overflows, so `el.scrollTop = el.scrollHeight` did nothing. The real scrollable ancestor is the parent div in `chat-area.tsx`.
- Replace with `bottomRef.scrollIntoView({ behavior: 'instant', block: 'end' })` which correctly targets the nearest scrollable ancestor.
- Change `activeBubbleId` scroll from `block: 'nearest'` to `block: 'end'` — the active bubble is always the latest message, and `nearest` could leave the viewport mid-list after thinking transitions.

## Root Cause
In `chat-session.tsx`, three `useEffect` hooks manage auto-scroll:
1. **Effect 1** (`msgCount` dep): `bottomRef.scrollIntoView` — works correctly
2. **Effect 2** (`session.messages` dep): `scrollContainerRef.scrollTop = scrollHeight` — **broken** because the element has no height constraint and never overflows
3. **Effect 3** (`activeBubbleId` dep): `activeBubbleRef.scrollIntoView({ block: 'nearest' })` — works but `nearest` can scroll to mid-list

When thinking completes, `onAgentEnd` removes empty messages and `onAgentStart` adds new ones. If React batches these, `msgCount` stays the same → Effect 1 doesn't fire. Effect 2 should compensate but is a no-op → scroll position is lost during re-layout → chat jumps to top.

## Test plan
- [ ] Start a conversation, let AI think, verify scroll stays at bottom when response arrives
- [ ] During text streaming, verify chat stays pinned to bottom
- [ ] Scroll up manually during streaming, verify auto-scroll doesn't force back down (isAtBottomRef guard)


wyuc (changes requested):
The diagnosis is partly right—the actual scrolling element is the outer `overflow-y-auto` container in `chat-area.tsx`, so assigning `scrollTop` on `scrollContainerRef` inside `ChatSessionComponent` is ineffective. However, this patch changes the write side to `bottomRef.scrollIntoView()` without fixing the read side.

`handleScroll` is still attached to the same inner element that has no constrained height and does not scroll. Scrolling the real outer container does not update `isAtBottomRef`, so it effectively remains `true`. With this patch, every `session.messages` update can therefore call `scrollIntoView()` and force a user who deliberately scrolled up to read history back to the bottom—the exact behavior the guard is meant to prevent.

Please make the actual scroll container own/report the bottom state (for example by lifting the logic into `chat-area.tsx`, passing the real scrolling element/ref down, or observing the bottom sentinel), then use that state to gate streaming auto-scroll. Please verify these three cases: (1) streaming stays pinned when already at the bottom, (2) manually scrolling up suppresses auto-scroll while content continues streaming, and (3) the thinking-to-answer transition remains at the bottom. A short reproduction/recording of the original jump and the fixed behavior would also resolve the earlier unanswered verification request. Rebase onto current `main` before re-review.

wyuc:

PR #386, opened 2026-04-09
fix: resolve DNS before SSRF validation to prevent rebinding bypass

### LINKED ISSUE #378: [Bug]: SSRF guard bypassed via DNS rebinding (hostname-only validation)
### Bug Description

The SSRF guard in `lib/server/ssrf-guard.ts` validates hostnames as strings (e.g. `localhost`, `192.168.*`) but does not resolve DNS before checking. An attacker can register a domain that resolves to a private IP (e.g. `127.0.0.1` or `169.254.169.254`), bypassing all private-network checks.

This is a security concern for public/cloud deployments where SSRF can be used to:
- Access cloud instance metadata (`169.254.169.254`) and leak IAM credentials
- Probe internal network services
- Reach unauthenticated internal APIs

### Steps to Reproduce

1. Register a domain (e.g. `attacker.com`) with a DNS A record pointing to `127.0.0.1`
2. In OpenMAIC, supply `http://attacker.com:11434/v1` as a model endpoint URL
3. The SSRF guard passes it — the hostname is not in the blocklist
4. The server-side fetch connects to `127.0.0.1`

### Expected Behavior

The SSRF guard should resolve the hostname to an IP address first, then validate the resolved IP against the private-range blocklist.

### Actual Behavior

The guard only checks the hostname string, so any domain resolving to a private IP bypasses the check.

### Suggested Fix

Resolve DNS before validation:

```typescript
import dns from 'node:dns';

const { address } = await dns.promises.lookup(parsed.hostname);
// Then check `address` against private IP ranges instead of `parsed.hostname`
```

This should be applied regardless of the `ALLOW_LOCAL_NETWORKS` setting — when that flag is off, DNS-resolved private IPs should still be blocked.

### Additional Context

- This is a pre-existing issue, not introduced by #366
- Only affects deployments where untrusted users can supply custom URLs (e.g. model endpoints)
- Self-hosted deployments with `ALLOW_LOCAL_NETWORKS=true` are not affected (private IPs are already allowed)

YizukiAme:
## Summary

Fix SSRF guard DNS rebinding bypass by resolving hostnames to IP addresses before validation.

Closes #378

## Problem

The SSRF guard in `lib/server/ssrf-guard.ts` only checked hostname strings (e.g. `localhost`, `192.168.*`), allowing attackers to register a domain resolving to a private IP and bypass all private-network checks. This could expose cloud instance metadata (`169.254.169.254`), internal APIs, and other local services.

### Core Fix (`lib/server/ssrf-guard.ts`)

- **`validateUrlForSSRF` is now `async`** — resolves hostnames via `dns.promises.lookup` with `{ all: true, verbatim: true }` and rejects if **any** returned address is private/local
- **`isPrivateIP()` helper** covers:
  - IPv4: `127.0.0.0/8`, `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `169.254.0.0/16`, `0.0.0.0`
  - IPv6: `::1`, `::`, `fc00::/7` (ULA), `fe80::/10` (link-local), `fec0::/10` (deprecated site-local)
  - IPv4-mapped IPv6: `::ffff:127.0.0.1`, `::ffff:7f00:0001`
  - **6to4 tunnel** (`2002::/16`): extracts embedded IPv4 from bits 16-47
  - **Teredo tunnel** (`2001:0000::/32`): extracts XOR-inverted client IPv4 from bits 96-127
- **IP literals detected via `net.isIP()`** — validated directly without DNS lookup, avoiding false positives from "DNS failure = block"
- **Fail-closed** — DNS resolution errors or empty results reject the request

### Async Propagation (20+ files)

- `resolveModel()` and `resolveModelFromHeaders()` in `lib/server/resolve-model.ts` are now `async`
- All direct `validateUrlForSSRF()` callers and indirect callers via `resolveModel`/`resolveModelFromHeaders` updated with `await`
- **No behavioral changes** — existing production-only gating (`NODE_ENV === 'production'`) and per-route SSRF policies are preserved

### Tests (`tests/server/ssrf-guard.test.ts`)

14 test cases covering:
| Category | Cases |
|----------|-------|
| Public pass-through | Public hostname, IPv4 literal, IPv6 literal |
| Protocol blocking | Invalid URL, `ftp://`, `file://`, `javascript:` |
| Hostname fast-path | `localhost`, `.local` |
| Private IPv4 literals | `127.0.0.1`, `10.x`, `172.16-31.x`, `192.168.x`, `169.254.x`, `0.0.0.0` |
| Private IPv6 literals | `::1`, `fd00::`, `fe80::`, `fec0::`, `::ffff:127.0.0.1` |
| 6to4 tunnel | Private embedded (blocked) + public embedded (allowed) |
| Teredo tunnel | Private embedded (blocked) + public embedded (allowed) |
| DNS rebinding | Mocked lookup → `127.0.0.1` → blocked |
| Mixed DNS answers | Public + private → blocked |
| DNS failure | Lookup throws → blocked (fail-closed) |

## Verification

-  `pnpm vitest run tests/server/ssrf-guard.test.ts` — 14/14 passed
-  `pnpm build` — no type errors

## Out of Scope

- **TOCTOU / IP pinning**: The resolved IP at validation time could differ from the IP at fetch time. Mitigating this requires connecting directly by resolved IP, which is a larger change. Tracked separately.
- **ISATAP tunnel**: Extremely rare in practice, deferred.

*Assisted by Claude Opus 4.6 and GPT-5.4.*

### DIFF
--- a/app/api/azure-voices/route.ts
+++ b/app/api/azure-voices/route.ts
@@ -26,7 +26,7 @@ export async function POST(req: NextRequest) {
     }
 
     // Validate baseUrl against SSRF
-    const ssrfError = validateUrlForSSRF(baseUrl);
+    const ssrfError = await validateUrlForSSRF(baseUrl);
     if (ssrfError) {
       return apiError('INVALID_URL', 403, ssrfError);
     }
--- a/app/api/chat/route.ts
+++ b/app/api/chat/route.ts
@@ -68,7 +68,7 @@ export async function POST(req: NextRequest) {
       model: languageModel,
       apiKey: resolvedApiKey,
       providerId,
-    } = resolveModel({
+    } = await resolveModel({
       modelString: body.model,
       apiKey: body.apiKey,
       baseUrl: body.baseUrl,
--- a/app/api/generate/agent-profiles/route.ts
+++ b/app/api/generate/agent-profiles/route.ts
@@ -66,7 +66,7 @@ export async function POST(req: NextRequest) {
     }
 
     // ── Model resolution from request headers ──
-    const { model: languageModel, modelString: _modelString } = resolveModelFromHeaders(req);
+    const { model: languageModel, modelString: _modelString } = await resolveModelFromHeaders(req);
     modelString = _modelString;
 
     // ── Build prompt ──
--- a/app/api/generate/image/route.ts
+++ b/app/api/generate/image/route.ts
@@ -41,7 +41,7 @@ export async function POST(request: NextRequest) {
     const clientModel = request.headers.get('x-image-model') || undefined;
 
     if (clientBaseUrl && process.env.NODE_ENV === 'production') {
-      const ssrfError = validateUrlForSSRF(clientBaseUrl);
+      const ssrfError = await validateUrlForSSRF(clientBaseUrl);
       if (ssrfError) {
         return apiError('INVALID_URL', 403, ssrfError);
       }
--- a/app/api/generate/scene-actions/route.ts
+++ b/app/api/generate/scene-actions/route.ts
@@ -77,7 +77,7 @@ export async function POST(req: NextRequest) {
     }
 
     // ── Model resolution from request headers ──
-    const { model: languageModel, modelInfo, modelString } = resolveModelFromHeaders(req);
+    const { model: languageModel, modelInfo, modelString } = await resolveModelFromHeaders(req);
     outlineTitle = outline?.title;
     resolvedModelString = modelString;
 
--- a/app/api/generate/scene-content/route.ts
+++ b/app/api/generate/scene-content/route.ts
@@ -73,7 +73,7 @@ export async function POST(req: NextRequest) {
     };
 
     // ── Model resolution from request headers ──
-    const { model: languageModel, modelInfo, modelString } = resolveModelFromHeaders(req);
+    const { model: languageModel, modelInfo, modelString } = await resolveModelFromHeaders(req);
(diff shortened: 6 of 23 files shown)


YizukiAme:
## Follow-up: TOCTOU / IP Pinning

This PR fixes the primary DNS rebinding bypass (hostname-only validation), but a **time-of-check-time-of-use (TOCTOU)** gap remains:

1. **Validation time**: `dns.lookup("attacker.com")` → resolves to `8.8.8.8` (public) →  passes
2. **Fetch time** (milliseconds later): system DNS resolves again → attacker's DNS server now returns `127.0.0.1` → connects to localhost

To fully mitigate this, the server would need to **pin the resolved IP** and connect directly to it (setting `Host` header manually), bypassing the second DNS resolution. This requires a custom HTTP agent and is a significantly larger change.

**Planned follow-up issue** (if this PR is merged):
- Implement IP pinning or connect-by-resolved-IP for all SSRF-guarded fetch paths
- Also consider covering ISATAP tunnel addresses (`::0200:5efe:x.x.x.x`)

The current fix already blocks the vast majority of real-world DNS rebinding attacks — exploiting the TOCTOU window requires precise millisecond-level DNS TTL manipulation.

YizukiAme:
so many conflicts?! will fix and address tomorrow

YizukiAme:
all conflicts resolved~

wyuc (approved):
LGTM

PR #389, opened 2026-04-09
fix(i18n): add missing Ollama and Doubao provider names for ru-RU

### LINKED ISSUE #388: i18n(ru-RU): missing Ollama and Doubao provider names
After merging #94 (Ollama support), `ru-RU.json` is missing two `providerNames` keys that were added to en-US, zh-CN, and ja-JP:

- `ollama`
- `doubao`

Quick fix — just add the two keys to `lib/i18n/locales/ru-RU.json`.

cosarah:
## Summary
- Add missing `providerNames.ollama` and `providerNames.doubao` keys to `ru-RU.json`
- These were added to en-US, zh-CN, ja-JP in #94 but ru-RU was missed

Closes #388

### DIFF
--- a/components/settings/index.tsx
+++ b/components/settings/index.tsx
@@ -520,7 +520,12 @@ export function SettingsDialog({ open, onOpenChange, initialSection }: SettingsD
                 <Box className="h-8 w-8 text-muted-foreground" />
               )}
               <div>
-                <h2 className="text-lg font-semibold">{selectedProvider.name}</h2>
+                <h2 className="text-lg font-semibold">
+                  {t(`settings.providerNames.${selectedProvider.id}`) !==
+                  `settings.providerNames.${selectedProvider.id}`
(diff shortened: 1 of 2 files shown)


cosarah:
Two small i18n fixes:

1. **ru-RU missing keys** — #94 added `providerNames.ollama` and `providerNames.doubao` to en-US/zh-CN/ja-JP but missed ru-RU. This adds them.

2. **Settings detail panel hardcoded Chinese** — The provider list (left side) already uses `t('settings.providerNames.${id}')` for i18n lookup, but the detail header (right side) was rendering `provider.name` directly from the registry, which is hardcoded Chinese for Doubao ("豆包") and SiliconFlow ("硅基流动"). Now both sides use the same i18n lookup with fallback to `provider.name` for unregistered/custom providers.

wyuc (approved):

PR #393, opened 2026-04-10
feat: claude search

### LINKED ISSUE #392: [Feature]: Adding Claude for Web Search
### Problem or Motivation

Claude has a web search tool which I would like to use.

### Proposed Solution

Add Claude as a provider for Web Search

### Alternatives Considered

_No response_

### Area

Classroom generation

### Additional Context

_No response_

joseph-mpo-yeti:
## Summary

Adds a new Claude-based web search provider (via Anthropic Messages + web_search_* tools) and upgrades the app to support multiple web search providers end-to-end (store, UI, API, and server generation), including SSRF protections for page/content fetching.

## Related Issues

Closes #392 

## Changes

- Introduces searchWithClaude (including page-content enrichment + SSRF guard) and adds Claude/Tavily to WEB_SEARCH_PROVIDERS.
- Refactors web search settings/state to support provider selection, enable/disable toggle, Claude model/tool configuration, and server/env API key resolution.
- Updates the web-search API + generation flows (toolbar, preview, classroom generation) to pass provider + providerConfig; adds tests and icons.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. pnpm build 
2. pnpm start

### Evidence

- [x] Manually tested locally

#### Before
before-tavily-settings
before-web-search

#### After
after-tavily-settings
after-claude-settings
after-web-search-tavily
after-web-search-claude

## Checklist


wyuc (changes requested):
Rebase onto current main to resolve conflicts.

Two things to fix:

1. `handleSaveModel` doesn't check for duplicate model IDs. Add a uniqueness check before saving to avoid React key collisions.

2. i18n: `ar-SA.json` was just merged and several new keys were added recently (#418, #430, #435). Make sure all 4 locale files have the new web search keys after rebasing.

wyuc:
Bug: `searchWithClaude` manually constructs the API URL (`${baseUrl}/v1/messages`) via `proxyFetch`, but test connection goes through the Anthropic SDK. These two paths handle base URLs differently, so test connection can pass while actual search 404s (e.g. when the base URL ends with `/v1`).

Fix: use the Anthropic SDK (`messages.create`) for the search call too, same as test connection. This aligns URL handling and avoids the manual endpoint construction entirely.

wyuc:
The `web_search` tool needs `allowed_callers: ["direct"]` in the tool definition. Without it, models like Haiku that don't support programmatic tool calling will 400. Since the search call is always user-initiated (not model-to-model), `"direct"` is the correct caller type here.

wyuc:
Hi @joseph-mpo-yeti — apologies this took so many rounds and a while to get back to you. I went through everything again and it all holds up.

**Review items, all addressed:**
- Duplicate model-ID check is in (`handleSaveModel` rejects collisions, correctly excluding the row being edited).
- Search and Test Connection now share one path (`createAnthropic({ baseURL })`), so the URL-handling divergence is gone.
- `allowed_callers: ["direct"]` is injected on every tool. Moving from the raw SDK to `@ai-sdk/anthropic` plus a fetch-layer patch is a reasonable call given the SDK does not emit that field.

**End-to-end verified:** I ran a real generation on a topic whose answer is past the model's training cutoff (the current Node.js LTS). Web search returned live sources and the generated slides correctly produced "Node.js 24 'Krypton', Active LTS, latest v24.16.0", so fresh results genuinely flow into the output.

**Before merge:** just needs a rebase onto `main` — it's currently conflicting.

Optional nits, take or leave: dedup `result.sources` by URL (the model often returns the same page several times), and drop the `final payload` debug log since it dumps the full request body.

Once rebased I'm happy to merge. Thanks for your patience on this one.

lib/web-search/claude.ts
```diff
@@ -0,0 +1,166 @@
...
+/** Fetch a URL and return plain text extracted from its HTML. Returns empty string on any failure. */
+async function fetchPageContent(url: string): Promise<string> {
+  const ssrfError = await validateUrlForSSRF(url);
+  if (ssrfError) {
+    log.warn(`Blocked page fetch due to SSRF check [url="${url}" reason="${ssrfError}"]`);
+    return '';
+  }
+  log.info(`Fetching page content: ${url}`);
+  try {
+    const res = await proxyFetch(url, {
```
wyuc: This validates only the initial source URL, then calls `proxyFetch` with the default redirect behavior. A public attacker-controlled URL can redirect to localhost, cloud metadata, or another private-network target without the redirect destination being revalidated. The fetched body is then included in `sources`, so this is also an exfiltration path. Prefer removing the second page fetch and using Claude's answer/citations directly. If page enrichment is still required, use `redirect: 'manual'`, resolve and validate every `Location` hop, cap redirect count and response bytes, and add a redirect-to-private-address regression test.

wyuc (changes requested):
I re-reviewed this against current `main`. My May 29 guidance that this only needed a rebase is no longer current: the branch is now 194 commits behind and a test merge produces 17 conflicts. More importantly, `main` has since gained a mature multi-provider web-search framework with Tavily, Bocha, Brave, Baidu, MiniMax, Doubao, and SearXNG. Most of this PR's store, settings, toolbar, route, and generation refactor has therefore been superseded.

Please rebuild this as a focused forward-port on current `main`, preserving the valuable Claude adapter, provider registration/config, translations, and tests while plugging into the existing `searchWeb` dispatcher and managed-provider rules. Carrying the old framework changes through the conflicts would risk regressing the providers and security boundaries added since April. There is also one P1 SSRF issue in the adapter that must be fixed; details are attached inline.

The previous functional fixes and real end-to-end verification were valid on the old base, and I appreciate the work that went into them. This request is about the repository moving substantially underneath the PR, not a repeat of those resolved rounds.

lib/server/web-search-config.ts
```diff
@@ -22,6 +23,7 @@ const OFFICIAL_CLIENT_BASE_URLS: Record<WebSearchProviderId, string[]> = {
     'https://api.search.brave.com',
   ],
   baidu: ['https://qianfan.baidubce.com'],
+  claude: ['https://api.anthropic.com', 'https://api.anthropic.com/v1'],
```
wyuc, line 28: [P2] Normalize the bare Anthropic API root to `/v1`. This allowlist accepts `https://api.anthropic.com`, but `@ai-sdk/anthropic` appends `/messages` directly, producing `https://api.anthropic.com/messages`, which returns 404. Reject the bare form or normalize it to `https://api.anthropic.com/v1` before passing it to the SDK, and add a regression test for the final request URL.

lib/web-search/claude.ts
```diff
@@ -0,0 +1,132 @@
...
+const CLAUDE_MAX_OUTPUT_TOKENS = 4096;
+
+const log = createLogger('ClaudeSearch');
+
+/**
+ * Haiku-tier and pre-4.6 models only support the basic web_search_20250305
+ * tool; 4.6+ Opus/Sonnet models use web_search_20260209 (dynamic filtering).
+ */
+function usesBasicWebSearchTool(modelId: string): boolean {
+  return /haiku|-4-5|-4-1|-4-0|-3-/.test(modelId);
```
wyuc: [P2] Recognize dated Claude 4.0 model IDs as basic-search models. Canonical IDs such as `claude-sonnet-4-20250514` and `claude-opus-4-20250514` do not match this regex, so they are routed to `web_search_20260209` even though the implementation comment and compatibility boundary require `web_search_20250305` before 4.6. This is reachable through both `WEB_SEARCH_CLAUDE_MODELS` and the request model field.

lib/web-search/claude.ts
```diff
@@ -0,0 +1,132 @@
...
+ * otherwise). Patch it into outgoing request bodies at the fetch layer.
+ */
+async function fetchWithAllowedCallers(url: string, init?: RequestInit): Promise<Response> {
+  if (init?.method === 'POST' && typeof init.body === 'string') {
+    try {
+      const body = JSON.parse(init.body);
+      if (Array.isArray(body?.tools)) {
+        body.tools = body.tools.map((tool: Record<string, unknown>) =>
+          tool.allowed_callers ? tool : { ...tool, allowed_callers: ['direct'] },
+        );
```
wyuc: [P2] Preserve dynamic filtering for Claude 4.6+ models. Injecting `allowed_callers: ["direct"]` into every `web_search_20260209` request explicitly bypasses the code-execution caller that provides dynamic filtering, contradicting the selected tool version and the adapter comment. Only force direct execution for models/tool versions that actually require it; cover both the Sonnet 5 and Haiku 4.5 request bodies.

lib/web-search/claude.ts
```diff
@@ -0,0 +1,132 @@
...
+        content: 'Referenced by the Claude web-search answer.',
+        score: 1,
+      });
+    }
+
+    return {
+      answer: result.text,
+      sources: [...sources.values()],
+      query,
+      responseTime: Date.now() - startTime,
```
wyuc: [P2] Return `responseTime` in seconds to preserve the existing `WebSearchResult` contract. Every sibling adapter reports seconds, while this expression returns milliseconds, so `/api/web-search` exposes values roughly 1000× larger only for Claude. Use `(Date.now() - startTime) / 1000` and assert the unit in the adapter test.

wyuc (changes requested):
Thanks for rebuilding this as a focused forward-port on the current multi-provider framework. The previous SSRF issue is resolved: the adapter no longer fetches attacker-influenced result URLs, and I did not find regressions to the existing providers.

Four correctness/integration issues remain before this is ready to merge; details are attached inline. CI is green, but these runtime paths are not covered by the current mocked tests.

wyuc (approved):
Re-reviewed commit 9dd70f117bb0d746323872f0a5959c8fedd14945. The previously requested adapter and integration fixes are addressed, CI is green, and the production-built Settings flow passed browser verification. The bare Anthropic-root Request URL preview remains a minor UI inconsistency, but the server-side request normalization is correct and this is non-blocking.

PR #397, opened 2026-04-10
fix: stabilize pptxgenjs workspace install build

abrahamU:
## Summary
- Ensure the local pptxgenjs workspace build handles TypeScript sources reliably during postinstall.
- Switch Rollup TypeScript binding to an ESM import and explicit TS/TSX include pattern.
- Update the lockfile after reinstall and remove stale env example file.

## Test plan
- pnpm --filter pptxgenjs build
- pnpm install


wyuc:
Thanks for catching and fixing this, @abrahamU — the rollup change is spot on.

I've opened #616 with just that fix so we can merge it cleanly. This PR also deleted `.env.example` (which is still the template users copy to `.env.local`) and rewrote the whole `pnpm-lock.yaml`, and it's since gone out of sync with `main`, so rather than untangle it I pulled the rollup change into a fresh branch and credited you as co-author. Closing this in favor of #616 — thanks again!

Issue #287, opened 2026-03-26
[Bug]: Authentication (CODE/ACCESS_CODE) fails to trigger a password prompt and erroneously overrides the Server-side API Key


albert8899:
### Bug Description

I have successfully deployed the project on Vercel and configured HIDE_USER_API_KEY=true to protect my credentials. However, I need to restrict access exclusively to my students (Authentication).

I attempted to use CODE or ACCESS_CODE in Environment Variables, but encountered two critical issues:

No Authentication UI: The web interface does not trigger a password/access code prompt upon entry.

Credential Override: If a user manually enters the access code (e.g., Agent20251220) in the "API Key" field of the settings, the system incorrectly treats this code as the real DOUBAO_API_KEY for API requests, leading to a The API key format is incorrect error from the provider.

[中文]
我已通过 Vercel 成功部署，并配置了 HIDE_USER_API_KEY=true 来隐藏私钥。为了 700 名学生的调研实验，我需要实现访问限制（Authentication）。

我尝试了在环境变量中设置 CODE 或 ACCESS_CODE，但遇到了以下问题：

无鉴权弹窗：网页前端无法自动弹出密码/访问码的验证框。

凭证覆盖错误：如果学生在设置里的“API Key”处输入访问码，系统会错误地将其作为 API_KEY 发送给后端模型供应商，导致报出格式错误（Key format incorrect）。

### Steps to Reproduce

1. Set DOUBAO_API_KEY, DOUBAO_BASE_URL, and CODE=YourPassword in Vercel.

2. Open the site in Incognito mode.

3. No password prompt appears; sending a message fails or requires manual Key input.

### Expected Behavior

[English]

Mandatory Auth Prompt: If CODE or ACCESS_CODE is set, the site should prompt for a password before allowing any interaction.

Separation of Concerns: Once the Access Code is verified, the system should strictly use the server-side DOUBAO_API_KEY for requests, and the Access Code should never be passed into the API Key field of the provider's request header.

[中文]

强制鉴权弹窗：如果设置了访问码，站点应在用户操作前强制弹出密码框。

逻辑分离：访问码仅用于前端解锁。解锁后，系统应强制使用服务端配置的真实 API Key进行请求，而不应让访问码覆盖或污染 API 凭证字段。

### Actual Behavior

Actual Behavior / 实际行为
[English]

Missing Authentication Barrier: Even with CODE or ACCESS_CODE set in environment variables, the application loads directly without any password prompt or overlay.

[...]

PR #411, opened 2026-04-12
feat: ACCESS_CODE site-level authentication (#407)

### LINKED ISSUE #407: feat: ACCESS_CODE site-level authentication
Implement a simple ACCESS_CODE environment variable for self-hosted and shared deployments, as described in the v0.1.1 roadmap (#406).

**Approach:**
- Next.js middleware checks all requests for a valid session cookie
- First-time visitors see a password modal
- Server-side verify endpoint checks code, sets session cookie
- All subsequent requests (including API) authenticated via cookie
- No ACCESS_CODE set → zero behavioral change

**Non-goals:** user accounts, role-based access, token-based auth.

Related: #287, Discussion #293, #406

wyuc:
## Summary

- Add `ACCESS_CODE` env var that gates access to the entire app
- When set, unauthenticated users see a full-screen password modal and cannot call APIs
- When not set, zero behavioral change

## How it works

1. **Next.js middleware** checks all requests for a valid `openmaic_access` cookie
2. First-time visitors see a full-screen modal prompting for the access code
3. Code is verified server-side via `/api/access-code/verify`, which sets an HMAC-signed HttpOnly cookie (7-day TTL)
4. All subsequent requests (including API calls) are validated by verifying the HMAC signature in the cookie
5. If `ACCESS_CODE` is not set, middleware exits immediately — zero overhead

## New files (10)

| File | Purpose |
|---|---|
| `middleware.ts` | Gate all requests, validate HMAC-signed cookie |
| `app/api/access-code/verify/route.ts` | POST: validate code, set signed cookie |
| `app/api/access-code/status/route.ts` | GET: return {enabled, authenticated} |
| `components/access-code-guard.tsx` | Layout wrapper: fetch status, show/hide modal |
| `components/access-code-modal.tsx` | Full-screen password gate with motion animations |

## Modified files

- `app/layout.tsx` — wrap children with `<AccessCodeGuard>`
- `.env.example` — add `ACCESS_CODE`
- `lib/i18n/locales/*.json` (4 files) — add `accessCode.*` translation keys (en-US, zh-CN, ja-JP, ru-RU)

## Security

- Cookie value is `timestamp.HMAC-SHA256(ACCESS_CODE, timestamp)` — never the plaintext code
- Middleware validates HMAC signature on every request (not just cookie existence)
- Access code comparison uses constant-time comparison (`timingSafeEqual`)
- Middleware uses Web Crypto API for Edge Runtime compatibility
- HttpOnly + SameSite=Lax cookie
- **Threat model**: prevent casual unauthorized access to shared deployments. Not enterprise security.

## Code Review Summary

Reviewed by automated code reviewer. All Critical and Important findings addressed:

| Severity | Issue | Resolution |
|---|---|---|
| Critical | Cookie only checked for existence, not validated | HMAC-signed tokens + middleware validation |
| Critical | Timing attack on string comparison | `timingSafeEqual` in verify endpoint |
| Important | Middleware 401 missing `errorCode` | Added to match `apiError` shape |
| Important | Status endpoint not using `apiSuccess` | Fixed |
| Important | Dead `accessCode.submit` i18n key | Removed from all 4 locales |
| Suggestion | Guard renders null during loading (white flash) | Now renders children, modal overlays on top |
| Suggestion | Network error silently disables access control | Defaults to `enabled: true` on error |
| Fix | Node.js `crypto` not available in Edge Runtime | Replaced with Web Crypto API |

## Test plan

- [ ] Set `ACCESS_CODE=test123` in `.env.local`, start dev server
- [ ] Visit app → full-screen modal appears
- [ ] Enter wrong code → inline error message
- [ ] Enter correct code → modal dismisses, app works
- [ ] Refresh page → no modal (cookie persists)
- [ ] `curl /api/verify-model` without cookie → 401
- [ ] Remove `ACCESS_CODE` from env, restart → no modal, all APIs work
- [ ] Verify dark mode renders correctly

Closes #407
Related: #287, Discussion #293, Roadmap #406


cosarah (approved):
LGTM! Clean architecture and solid security fundamentals (HMAC-signed tokens, `timingSafeEqual`, `httpOnly`/`secure` cookies, fail-closed error handling).

Two minor notes for awareness (non-blocking):

1. **Token has no server-side expiry check** — the HMAC verification only validates the signature, not the timestamp age. The cookie `maxAge` handles this in practice, but an extracted token would remain valid indefinitely as long as `ACCESS_CODE` stays the same.

2. **Duplicate HMAC verification logic** — `verify/route.ts` (Node `crypto`) and `middleware.ts` (Web Crypto API) implement the same token verification independently. Necessary for Edge Runtime compatibility, but worth keeping in mind if the token format ever changes.

PR #417, opened 2026-04-13
fix(settings): persist custom provider baseUrl on creation (#414)

### LINKED ISSUE #414: fix: custom provider baseUrl not persisted — stored as defaultBaseUrl instead
## Bug

When adding a custom OpenAI-compatible provider (e.g. OpenRouter), the Base URL entered in the Add Provider dialog is stored as `defaultBaseUrl` instead of `baseUrl`. The `baseUrl` field is hardcoded to `""` on creation.

### Root Cause

`components/settings/index.tsx` L458-471 — `handleAddProvider`:

```typescript
[newProviderId]: {
  apiKey: '',
  baseUrl: '',                              // ← always empty
  defaultBaseUrl: providerData.baseUrl,       // ← user input goes here
  ...
}
```

Downstream, `model-selector.tsx` reads `providerConfig.baseUrl` (empty string) for verify-model requests. The server receives `baseUrl: ""`, treats it as `undefined`, falls back to server config lookup, finds nothing for the custom provider ID, and defaults to `api.openai.com` — resulting in **"Cannot connect to API"** after 3 retries.

### Impact

- **Verify model** always fails for newly added custom providers
- **Generation** with custom providers also fails (same `baseUrl` resolution path)
- The provider-config-panel shows the correct placeholder from `defaultBaseUrl`, masking the bug

### Repro

1. Settings → Add Provider → type: OpenAI Compatible, name: OpenRouter, Base URL: `https://openrouter.ai/api/v1`
2. Add a model (e.g. `google/gemma-4-31b-it:free`), enter API key
3. Click verify → **"Cannot connect to API"**

### Expected

Verify should send the baseUrl to the server and succeed (or fail with an auth error if the key is wrong).

### Fix Direction

Either:
- Set `baseUrl: providerData.baseUrl` on provider creation (instead of `''`), or
- Make consumers fall back to `defaultBaseUrl` when `baseUrl` is empty

YizukiAme:
## Summary

Fix custom OpenAI-compatible provider `baseUrl` not being persisted on creation, causing verify-model and generation requests to fail.

Closes #414

## Root Cause

`handleAddProvider` in `components/settings/index.tsx` hardcoded `baseUrl: ''` while storing the user-entered URL in `defaultBaseUrl`. All downstream consumers (model-selector, provider-config-panel, model-edit-dialog, use-chat-sessions) only read `baseUrl`, so the server received an empty string and fell back to SDK defaults (e.g. `api.openai.com`).

### Core fix
- **`components/settings/utils.ts`**: New `createCustomProviderSettings()` that stores the user-entered URL in both `baseUrl` (for requests) and `defaultBaseUrl` (for placeholder/preview)
- **`components/settings/index.tsx`**: `handleAddProvider` now uses `createCustomProviderSettings()`

### Migration
- **`lib/store/settings.ts`**: `promoteLegacyCustomProviderBaseUrls()` runs on rehydrate — promotes `defaultBaseUrl` → `baseUrl` for existing custom providers (`!isBuiltIn && !baseUrl && defaultBaseUrl`). Does not affect built-in providers.

### Refactor
- **`components/settings/utils.ts`**: New `createVerifyModelRequest()` to unify verify-model request body construction across 3 call sites (model-selector, provider-config-panel, model-edit-dialog)

### Tests
- **`tests/settings/custom-provider-baseurl.test.ts`**: Tests that `createCustomProviderSettings` stores `baseUrl` correctly and `createVerifyModelRequest` builds the correct request body
- **`tests/store/settings-server-sync.test.ts`**: Tests that migration promotes `defaultBaseUrl` for custom providers and leaves built-in providers untouched

## Verification

- `npm test`: 98 passed (0 failed)
- `npx eslint` on changed files: all clean

*Assisted by Claude Opus 4.6 and GPT 5.4*

### DIFF
--- a/components/settings/index.tsx
+++ b/components/settings/index.tsx
@@ -34,7 +34,7 @@ import { toast } from 'sonner';
 import { type ProviderId } from '@/lib/ai/providers';
 import { PROVIDERS, MONO_LOGO_PROVIDERS } from '@/lib/ai/providers';
 import { cn } from '@/lib/utils';
-import { getProviderTypeLabel } from './utils';
+import { createCustomProviderSettings, getProviderTypeLabel } from './utils';
 import { ProviderList } from './provider-list';
 import { ProviderConfigPanel } from './provider-config-panel';
 import { PDFSettings } from './pdf-settings';
@@ -456,17 +456,7 @@ export function SettingsDialog({ open, onOpenChange, initialSection }: SettingsD
     const newProviderId = `custom-${Date.now()}` as ProviderId;
     const updatedConfig = {
       ...providersConfig,
-      [newProviderId]: {
-        apiKey: '',
-        baseUrl: '',
-        models: [],
-        name: providerData.name,
-        type: providerData.type,
-        defaultBaseUrl: providerData.baseUrl || undefined,
-        icon: providerData.icon || undefined,
-        requiresApiKey: providerData.requiresApiKey,
-        isBuiltIn: false,
-      },
+      [newProviderId]: createCustomProviderSettings(providerData),
     };
     setProvidersConfig(updatedConfig);
     setShowAddProviderDialog(false);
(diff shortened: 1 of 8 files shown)


wyuc (approved):
LGTM

PR #420, opened 2026-04-13
fix(#419): hide Ollama from model selector when not configured

### LINKED ISSUE #419: [Bug]: Ollama provider always visible in model selector even without any configuration
### Bug Description

The Ollama provider always appears in the homepage model selector dropdown regardless of whether it's actually configured — neither a server-side environment variable (`OLLAMA_BASE_URL`) nor a client-side API key/base URL setting is required for it to show up.

All other providers correctly require either a server-side API key or a client-side API key to be visible.

### Steps to Reproduce

1. Deploy or run OpenMAIC **without** setting any Ollama-related environment variables (no `OLLAMA_BASE_URL`, no `OLLAMA_API_KEY`)
2. Do **not** configure Ollama in the client-side settings either
3. Open the homepage and click the model selector in the bottom-left of the input box
4. Observe that Ollama and its models (Llama 3.3, Llama 3.2, Qwen 2.5, etc.) are listed as available

### Expected Behavior

Ollama should **not** appear in the model selector when neither the server nor the client has configured it. It should only be visible when:
- The server has set `OLLAMA_BASE_URL` (or equivalent YAML config), **or**
- The user has explicitly configured Ollama's base URL in client-side settings

### Actual Behavior

Ollama always appears in the model selector. Selecting an Ollama model and attempting to generate will fail at runtime since no Ollama instance is reachable.

### Root Cause Analysis

The model selector filtering logic in `components/settings/model-selector.tsx` (line 72-78) checks:

```typescript
(!config.requiresApiKey || config.apiKey || config.isServerConfigured) &&
config.models.length >= 1 &&
(config.baseUrl || config.defaultBaseUrl || config.serverBaseUrl)
```

Ollama's provider definition in `lib/ai/providers.ts` (line 958-964) has:
- `requiresApiKey: false` → first condition short-circuits to `true` (no key check at all)
- `defaultBaseUrl: 'http://localhost:11434/v1'` → third condition is always `true`
- Hardcoded models → second condition is always `true`

So Ollama unconditionally passes all three filters.

Other providers like OpenAI have `requiresApiKey: true`, so they correctly require either `config.apiKey` (client-side) or `config.isServerConfigured` (server-side) to appear.

### Suggested Fix

For providers with `requiresApiKey: false`, the filter should additionally check that the provider is either server-configured or has a user-provided base URL — not just rely on `defaultBaseUrl`. Something like:

```typescript
const isConfigured = config.requiresApiKey
  ? (config.apiKey || config.isServerConfigured)
  : (config.isServerConfigured || config.baseUrl); // exclude defaultBaseUrl for unconfigured keyless providers

isConfigured && config.models.length >= 1 && (config.baseUrl || config.defaultBaseUrl || config.serverBaseUrl)
```

This aligns with the server-side behavior in `lib/server/provider-config.ts` (line 176-182), which already requires keyless providers like Ollama to have an explicit `baseUrl` env var set before activating them.

### Deployment Method

### Additional Context

The server-side provider config (`lib/server/provider-config.ts`) already handles this correctly — Ollama is in the `keylessProviders` set and requires at least a `baseUrl` to be activated. The issue is only in the client-side filtering logic.

cosarah:
## Summary

- Fix Ollama provider unconditionally appearing in the model selector even when neither server-side nor client-side has configured it
- For keyless providers (e.g. Ollama), now require either `isServerConfigured` or an explicit user-provided `baseUrl` to appear — `defaultBaseUrl` alone no longer qualifies
- Update shared `isProviderUsable()` utility with the same logic and add test coverage

Closes #419

## Test plan

- [x] Unit tests: 58 tests pass (including 5 new keyless-provider cases)
- [x] Manual: Ollama not shown in homepage model selector when unconfigured
- [x] Manual: Other configured providers still appear correctly
- [x] Manual: Ollama appears when user sets a baseUrl in settings


wyuc (approved):
Tested locally, LGTM

PR #435, opened 2026-04-16
fix(export): use latest classroom name from IndexedDB for ZIP filename (#434)

### LINKED ISSUE #434: bug: exported ZIP filename uses original prompt instead of renamed classroom name
## Description

When a classroom is renamed on the home page and then exported as a ZIP, the filename still uses the original prompt text (up to 500 characters) instead of the user-given name.

## Steps to Reproduce

1. Create a classroom with a long prompt (e.g. "Please create a course about the history of ancient Roman architecture and its influence on modern design")
2. Go back to the home page
3. Rename the classroom to something short (e.g. "Roman Architecture")
4. Re-enter the classroom and export as ZIP
5. Observe the downloaded filename

## Expected Behavior

The exported ZIP should be named `Roman Architecture.maic.zip` (using the renamed name).

## Actual Behavior

The exported ZIP is named `Please create a course about the history of ancient Roman architecture and its influence on modern design.maic.zip` (the original prompt text).

## Root Cause Analysis

Two compounding issues:

1. **Stage name initialized from raw prompt** (`app/generation-preview/page.tsx:356`): `stage.name` is set to `extractTopicFromRequirement(requirement)`, which returns the prompt text verbatim (up to 500 chars) instead of a concise topic name.

2. **Stale in-memory cache** (`lib/store/stage.ts:275`): `loadFromStorage` skips reloading from IndexedDB when the store already has the stage in memory. So when the user renames on the home page (which updates IndexedDB via `renameStage()`), the Zustand store still holds the old prompt-based name. The export hook (`lib/export/use-export-classroom.ts:29`) reads from the Zustand store, not IndexedDB.

## Suggested Fix

- In `use-export-classroom.ts`, read the latest `stage.name` from IndexedDB before generating the filename (or ensure the store is refreshed after rename).
- Optionally, improve `extractTopicFromRequirement` to produce a shorter default name.

wyuc:
## Summary
- Read the latest `stage.name` from IndexedDB before generating ZIP, instead of relying on the potentially stale Zustand store cache
- Fixes the issue where renaming a classroom on the home page had no effect on the exported ZIP filename

## Test plan
- [x] `pnpm check` (Prettier) passed
- [x] `pnpm lint` (ESLint) 0 errors
- [x] `npx tsc --noEmit` passed
- [x] `vitest run tests/export/` 5/5 passed
- [ ] Manual: rename a classroom on home page → re-enter → export ZIP → verify filename uses the new name

Closes #434

### DIFF
--- a/lib/export/use-export-classroom.ts
+++ b/lib/export/use-export-classroom.ts
@@ -5,7 +5,7 @@ import { saveAs } from 'file-saver';
 import { toast } from 'sonner';
 import { useStageStore } from '@/lib/store/stage';
 import { useI18n } from '@/lib/hooks/use-i18n';
-import { getGeneratedAgentsByStageId } from '@/lib/utils/database';
+import { db, getGeneratedAgentsByStageId } from '@/lib/utils/database';
 import {
   CLASSROOM_ZIP_FORMAT_VERSION,
   CLASSROOM_ZIP_EXTENSION,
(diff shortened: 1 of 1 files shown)


cosarah (approved):
LGTM

Issue #456, opened 2026-04-19
Spotlight overlay leaves focused element dimmed; image/video spotlight measurement broken


wyuc:
## Problem

The slide-renderer spotlight effect (`SpotlightOverlay`) is broken in two independent ways.

### Bug 1 — `backdrop-filter` breaks SVG `<mask>` cutout

The dim layer is an SVG `<rect>` painted over the slide and clipped by an SVG `<mask>` so the focused element shows through. The same rect carries `className="backdrop-blur-[1.5px]"`, which compiles to `backdrop-filter: blur(1.5px)`.

In affected browsers, `backdrop-filter` and SVG `<mask>` composite in the wrong order: the blurred backdrop is painted as if the cutout did not exist, so the focused area looks dimmed even though the mask is correctly defined. Visible symptom: a white border appears around the focused element (it is rendered outside the mask) but the inside of the border stays dimmed.

Reproduction: trigger a spotlight on any text element — the focused area is dim instead of clear.

### Bug 2 — Image and video spotlights collapse to a 0-height cutout

`SpotlightOverlay.measure()` looks for `.element-content` inside the element wrapper and falls back to the wrapper itself. Seven of the nine `Base*Element` components carry `.element-content` on their fixed-size container; `BaseImageElement` and `BaseVideoElement` do not. The fallback hits the outer `.screen-element` div, which is a static block whose absolute-positioned children do not contribute to its bounding box, so `getBoundingClientRect()` returns `{ width: <slide width>, height: 0 }`.

Result: the mask cutout becomes a zero-height strip pinned to the top of the slide, the white border collapses with it, and the entire slide reads as a black overlay.

Reproduction: trigger a spotlight on any image or video element — the entire slide darkens with no visible cutout or border.

## Affected files

- `components/slide-renderer/Editor/SpotlightOverlay.tsx`
- `components/slide-renderer/components/element/ImageElement/BaseImageElement.tsx`
- `components/slide-renderer/components/element/VideoElement/BaseVideoElement.tsx`

## Expected behavior

- Spotlight on any element type clears the focused area and dims the rest, with a thin white border around the focus.
- Image and video elements behave the same as text/shape/etc.

PR #457, opened 2026-04-19
fix(spotlight): restore cutout for text and add element-content for image/video

wyuc:
## Summary

Fixes #456.

- **Drop `backdrop-blur-[1.5px]` from the dim SVG `<rect>`** in `SpotlightOverlay`. The class compiles to `backdrop-filter: blur(1.5px)`, which composites in the wrong order with SVG `<mask>` and bypasses the cutout, leaving the focused area dimmed. The 1.5px blur is visually negligible on a translucent overlay; removing it fully restores the cutout.
- **Add `element-content` class to `BaseImageElement` and `BaseVideoElement`.** `SpotlightOverlay.measure()` walks `#screen-element-{id} → .element-content`, falling back to the outer wrapper when the class is missing. The other 7 `Base*Element` components carry it; image and video did not. The fallback hit a static `<div>` whose absolutely-positioned children do not contribute to its bounding box, so the measured rect was `{w: <slide>, h: 0}` — the cutout collapsed to a zero-height strip pinned to the slide top, reading as a full black overlay.

Diff is 3 files / +8 / -4. The original `motion.rect` cutout animation is preserved.

## Test plan

- [ ] `pnpm check`, `pnpm lint`, `npx tsc --noEmit` all pass
- [ ] Spotlight on a text title (transparent fill) shows clear focus area + thin white border
- [ ] Spotlight on an image element shows the image clearly, dim everywhere else, with the white border around the image
- [ ] Spotlight on a video element behaves the same as image
- [ ] Cutout opening animation (mask shrinks from larger to tight) still plays

### DIFF
--- a/components/slide-renderer/Editor/SpotlightOverlay.tsx
+++ b/components/slide-renderer/Editor/SpotlightOverlay.tsx
@@ -122,13 +122,17 @@ export function SpotlightOverlay() {
                 </mask>
               </defs>
 
-              {/* Dimmed Background */}
+              {/* Dimmed Background. No backdrop-filter: combined with SVG <mask>
+                 it breaks compositing (backdrop bypasses the mask cutout) in some
+                 browsers, leaving the focused area dimmed despite the cutout.
+                 Tailwind 3 silently dropped `backdrop-blur-[1.5px]` on SVG via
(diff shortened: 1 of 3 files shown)


cosarah (approved):
clean change, LGTM
tested locallly

PR #473, opened 2026-04-23
feat(registry): optimize default nickname logic for user participants

bytedaring:
Updated the fallback key for user participant default names from `common.you` to `profile.defaultNickname`. This ensures consistency with the default values used on the profile page.

## Summary

This PR refactors the default nickname acquisition logic for user participants. By switching the fallback key from `common.you` to `profile.defaultNickname`, we ensure that the user's display name remains consistent across the Registry and the Profile modules.

## Related Issues

N/A

## Changes

- Modified the naming fallback logic in the participant registry.
- Updated the i18n/config key reference from `common.you` to `profile.defaultNickname`.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Log in to the application and navigate to the participant list (Registry).
2. Find a user participant who has not set a custom nickname.
3. Observe the fallback name displayed and verify it matches the value in `profile.defaultNickname` (instead of `common.you`).

- Verified that the fallback logic correctly triggers when the primary nickname is missing.
- Manually confirmed that the change aligns with the Profile page's display logic.

### Evidence

- [x] Manually tested locally

## Checklist


wyuc (changes requested):
The fallback-key change itself is reasonable and matches the profile UI, but this head is not green: the lint/typecheck job fails in `pnpm check` because `lib/orchestration/registry/store.ts` is not Prettier-clean. The PR also includes an unrelated formatting change in the `voiceConfig` block.

Please update onto current `main`, keep the diff to the fallback-key change, run Prettier/check, and add or update the focused participant-name test so the translation fallback behavior is explicit. This should be a quick re-review once CI is green.

wyuc:

PR #481, opened 2026-04-24
[codex] Support recent AI model providers

wyuc:
## Summary

- Add recently released LLM providers and model entries, including OpenRouter, Tencent Hunyuan, Xiaomi MiMo, updated DeepSeek V4 IDs, Kimi, Qwen, MiniMax, Grok, OpenAI, Anthropic, GLM, SiliconFlow, and Ollama ordering updates.
- Add OpenRouter, Tencent Hunyuan, and Xiaomi logos and server env wiring without treating `hy3` as an env prefix.
- Add OpenAI image generation support and expose it in media settings.
- Keep built-in provider models ordered by the registry on settings rehydrate so persisted local settings do not leave providers like MiniMax in stale reverse order.

## Why

The provider registry had fallen behind recently released models, and existing persisted settings could preserve older model ordering even after the registry was updated. That made newly curated provider lists appear inconsistent in the UI unless users cleared local settings.

## Validation

- `pnpm vitest run tests/store/settings-server-sync.test.ts tests/ai/minimax-provider.test.ts tests/server/provider-config.test.ts tests/media/openai-image-adapter.test.ts`
- `pnpm exec tsc --noEmit`
- `pnpm check:i18n-keys`
- `git diff --check`

## Notes

- DeepSeek official IDs are `deepseek-v4-pro` and `deepseek-v4-flash`; OpenRouter keeps provider-scoped slugs such as `deepseek/deepseek-v4-pro`.
- MiniMax now hydrates in newest-to-oldest order even for users with older persisted settings.


cosarah (approved):
Tested locally, works as expected. LGTM 🚀

PR #490, opened 2026-04-25
feat(prompts): conditional media snippets in generation prompts

### LINKED ISSUE #483: Make image/video element instructions conditional in generation prompts
## Problem

When image or video generation is disabled in the generation config, the AI still tries to emit image/video elements during outline and slide generation. The fallback logic intercepts these and prevents redundant generation calls, but the unwanted placeholders are still left behind in the produced content.

The root cause is **not** missing post-processing — it's that the prompts always describe the full media-element capability and only tell the AI "do not use it" via a negative override (`mediaGenerationPolicy` in `lib/orchestration/outline-generator.ts:86-99`). LLMs follow "do not do X" instructions far less reliably than simply not exposing X in the first place.

There are also two concrete gaps in the current implementation:

1. **Slide content generation receives no media policy at all.** `scene-generator.ts:671-681` passes nothing equivalent to `mediaGenerationPolicy`, so even when the outline correctly omits `mediaGenerations`, the slide content prompt still describes `ImageElement` / `VideoElement` / `gen_img_*` / `gen_vid_*` placeholders in full and the model can hallucinate them.
2. **The negative-override pattern is wasted tokens** — every generation call carries the full image + video instruction blocks (~300–600 tokens across outline + N slides) regardless of whether they are usable.

## Proposed Solution

Make the image and video instruction sections **truly pluggable** based on the `imageGenerationEnabled` / `videoGenerationEnabled` flags, rather than always-included-with-override.

### Implementation

1. **Extract image/video instruction blocks into snippets**
   - In `lib/prompts/templates/requirements-to-outlines/system.md`: lines covering when to request media generation, decision rules between image vs video, language handling for image labels, etc.
   - In `lib/prompts/templates/slide-content/system.md`: `ImageElement` rules, `VideoElement` rules, `gen_img_*` / `gen_vid_*` placeholder usage.
   - Move each into its own snippet file (e.g., `snippets/image-instructions.md`, `snippets/video-instructions.md`).

2. **Extend the prompt loader with conditional include syntax**
   - `lib/prompts/loader.ts` currently supports `{{snippet:name}}` unconditionally.
   - Add a conditional form, e.g. `{{snippet:image-instructions?if=imageEnabled}}` that resolves to empty string when the boolean variable is falsy.
   - This new capability is reusable for future flag-gated features (e.g., the new code element).

3. **Wire the flags through to both prompt builders**
   - Outline generator (`lib/orchestration/outline-generator.ts`): replace the `mediaGenerationPolicy` negative-override approach with the conditional snippets.
   - Slide content generator (`lib/orchestration/scene-generator.ts:671-681`): pass `imageEnabled` / `videoEnabled` into the slide content prompt variables.

4. **Rewrite transitional sentences** where the existing prose mixes image and video together, so each snippet reads cleanly when included alone or omitted.

### Why Option B over a smaller patch

A minimal alternative would be to simply propagate the existing `mediaGenerationPolicy` string to the slide content prompt. This is roughly half an hour of work and would mostly close the gap, but it preserves the negative-override pattern — which is unreliable for LLMs and wastes tokens on every call. Going straight to the snippet-based conditional approach:

- **Reliably eliminates unwanted placeholders** — the model cannot reference a capability it was never told about.
- **Real token savings** — per-call savings compound across outline + N×slide generations, especially for long courses.
- **Establishes a reusable pattern** — future element types (code, etc.) can gate themselves the same way.

### Trade-offs / Risks

- One-time refactor cost (~half a day): template splitting, loader extension, regression testing.
- Need to verify the four orthogonal combinations (image on/off × video on/off) all produce coherent prompts and good output.
- Extra care when splitting snippets so context isn't broken (sentences that currently mention image and video together need to be rewritten).

## Acceptance Criteria

- [ ] When image generation is disabled, generated outlines and slides contain no image-related placeholders or `ImageElement` references.
- [ ] When video generation is disabled, generated outlines and slides contain no video-related placeholders or `VideoElement` references.
- [ ] When both are disabled, the slide content prompt is meaningfully shorter (measurable token reduction).
- [ ] The prompt loader supports conditional snippet inclusion as a general-purpose feature.
- [ ] All four flag combinations produce coherent, well-formed prompts.

## Related Code

- `lib/prompts/loader.ts` — prompt template loader
- `lib/prompts/templates/requirements-to-outlines/system.md` — outline generation prompt
- `lib/prompts/templates/slide-content/system.md` — slide content generation prompt
- `lib/orchestration/outline-generator.ts:86-99` — current `mediaGenerationPolicy` negative-override
- `lib/orchestration/scene-generator.ts:671-681` — slide content prompt construction (currently missing media policy)
- `lib/store/settings.ts:101-133` — `imageGenerationEnabled` / `videoGenerationEnabled` flags

YizukiAme:
Closes #483

## Summary

Replace the always-included media instruction blocks + negative-override (`mediaGenerationPolicy`) pattern with **conditional snippet includes** gated on `imageEnabled` / `videoEnabled` flags. When a media capability is disabled, the model simply never sees the related instructions — no more "do NOT use X" overrides that LLMs unreliably follow.

### Prompt loader — conditional syntax (`lib/prompts/loader.ts`)

- Extended `processSnippets` to support `{{snippet:name?if=conditionName}}` — resolves to empty string when the condition is falsy.
- `buildPrompt` uses the caller's variables as snippet conditions by default; callers can pass an explicit `conditions` record via `PromptLoadOptions` when needed.
- `processSnippets` is now exported for direct testing.

### 15 new media snippet files (`lib/prompts/snippets/`)

Extracted from the monolithic `system.md` templates into independently gatable snippets:

| Snippet | Used in | Gated on |
|---------|---------|----------|
| `image-instructions` | outline | `imageEnabled` |
| `video-instructions` | outline | `videoEnabled` |
| `media-safety-guidelines` | outline | `mediaEnabled` |
| `outline-image-fields` | outline | `hasSourceImages` |
| `outline-media-field` | outline | `mediaEnabled` |
| `outline-image-resources` | outline (user) | `hasSourceImages` |
| `slide-image-instructions` | slide | `imageElementEnabled` |
| `slide-generated-image-instructions` | slide | `generatedImageEnabled` |
| `slide-video-instructions` | slide | `generatedVideoEnabled` |
| `slide-image-resources` | slide (user) | `mediaElementEnabled` |
| `slide-image-user-rule` | slide (user) | `imageElementEnabled` |
| `slide-video-user-rule` | slide (user) | `generatedVideoEnabled` |
| `slide-source-image-checklist` | slide | `imageElementEnabled` |
| `slide-generated-image-checklist` | slide | `generatedImageEnabled` |
| `slide-video-checklist` | slide | `generatedVideoEnabled` |

### Orchestration wiring

- **`outline-generator.ts`**: Removed `mediaGenerationPolicy` negative-override; passes `imageEnabled`, `videoEnabled`, `mediaEnabled`, `hasSourceImages` as template variables.
- **`scene-generator.ts`** (**closes the second gap from the issue**): Now derives `imageElementEnabled`, `generatedImageEnabled`, `generatedVideoEnabled` from the scene outline's `mediaGenerations` and passes them into the slide content prompt.
- **`scene-outlines-stream/route.ts`**: Same flag wiring as `outline-generator.ts` for the server-side streaming path.
- **`interactive-outlines/user.md`**: Removed the orphaned `{{mediaGenerationPolicy}}` reference (interactive scenes don't use `mediaGenerations`).

## Test coverage

- **`tests/prompts/media-conditional.test.ts`** (8 tests): Unit tests covering all 4 flag combinations for both outline and slide prompts, plus a token-length reduction assertion.
- **`tests/generation/media-prompt-wiring.test.ts`** (2 tests): Integration tests using mock `AICallFn` to verify end-to-end flag propagation through `generateSceneOutlinesFromRequirements` and `generateSceneContent`.

All 10 new tests pass.

*Assisted by GPT-5.5 and Claude Opus 4.6 & Manually investigated and audited.*

### DIFF
--- a/app/api/generate/scene-outlines-stream/route.ts
+++ b/app/api/generate/scene-outlines-stream/route.ts
@@ -189,20 +189,11 @@ export async function POST(req: NextRequest) {
       }
     }
 
-    // Build media generation policy based on enabled flags
+    // Build media snippet conditions based on enabled flags.
     const imageGenerationEnabled = req.headers.get('x-image-generation-enabled') === 'true';
     const videoGenerationEnabled = req.headers.get('x-video-generation-enabled') === 'true';
-    let mediaGenerationPolicy = '';
-    if (!imageGenerationEnabled && !videoGenerationEnabled) {
-      mediaGenerationPolicy =
-        '**IMPORTANT: Do NOT include any mediaGenerations in the outlines. Both image and video generation are disabled.**';
-    } else if (!imageGenerationEnabled) {
-      mediaGenerationPolicy =
-        '**IMPORTANT: Do NOT include any image mediaGenerations (type: "image") in the outlines. Image generation is disabled. Video generation is allowed.**';
-    } else if (!videoGenerationEnabled) {
-      mediaGenerationPolicy =
-        '**IMPORTANT: Do NOT include any video mediaGenerations (type: "video") in the outlines. Video generation is disabled. Image generation is allowed.**';
-    }
+    const mediaGenerationEnabled = imageGenerationEnabled || videoGenerationEnabled;
+    const hasSourceImages = (pdfImages?.length ?? 0) > 0;
 
     // Build teacher context from agents (if available)
     const teacherContext = formatTeacherPersonaForPrompt(agents);
@@ -218,7 +209,10 @@ export async function POST(req: NextRequest) {
       pdfContent: pdfText ? pdfText.substring(0, MAX_PDF_CONTENT_CHARS) : 'None',
       availableImages: availableImagesText,
       researchContext: researchContext || 'None',
-      mediaGenerationPolicy,
+      hasSourceImages,
+      imageEnabled: imageGenerationEnabled,
+      videoEnabled: videoGenerationEnabled,
+      mediaEnabled: mediaGenerationEnabled,
       teacherContext,
       userProfile: userProfileText,
     });
--- a/lib/generation/outline-generator.ts
+++ b/lib/generation/outline-generator.ts
@@ -83,20 +83,11 @@ export async function generateSceneOutlinesFromRequirements(
       ? `## Student Profile\n\nStudent: ${requirements.userNickname || 'Unknown'}${requirements.userBio ? ` — ${requirements.userBio}` : ''}\n\nConsider this student's background when designing the course. Adapt difficulty, examples, and teaching approach accordingly.\n\n---`
       : '';
 
-  // Build media generation policy based on enabled flags
+  // Build media snippet conditions based on enabled flags.
   const imageEnabled = options?.imageGenerationEnabled ?? false;
   const videoEnabled = options?.videoGenerationEnabled ?? false;
-  let mediaGenerationPolicy = '';
-  if (!imageEnabled && !videoEnabled) {
-    mediaGenerationPolicy =
-      '**IMPORTANT: Do NOT include any mediaGenerations in the outlines. Both image and video generation are disabled.**';
-  } else if (!imageEnabled) {
-    mediaGenerationPolicy =
-      '**IMPORTANT: Do NOT include any image mediaGenerations (type: "image") in the outlines. Image generation is disabled. Video generation is allowed.**';
-  } else if (!videoEnabled) {
-    mediaGenerationPolicy =
-      '**IMPORTANT: Do NOT include any video mediaGenerations (type: "video") in the outlines. Video generation is disabled. Image generation is allowed.**';
-  }
+  const mediaEnabled = imageEnabled || videoEnabled;
+  const hasSourceImages = (pdfImages?.length ?? 0) > 0;
 
   // Use simplified prompt variables
   const prompts = buildPrompt(PROMPT_IDS.REQUIREMENTS_TO_OUTLINES, {
@@ -105,7 +96,10 @@ export async function generateSceneOutlinesFromRequirements(
     pdfContent: pdfText ? pdfText.substring(0, MAX_PDF_CONTENT_CHARS) : 'None',
     availableImages: availableImagesText,
     userProfile: userProfileText,
-    mediaGenerationPolicy,
+    hasSourceImages,
+    imageEnabled,
+    videoEnabled,
+    mediaEnabled,
     researchContext: options?.researchContext || 'None',
     // Server-side generation populates this via options; client-side populates via formatTeacherPersonaForPrompt
     teacherContext: options?.teacherContext || '',
--- a/lib/generation/scene-generator.ts
+++ b/lib/generation/scene-generator.ts
@@ -633,14 +633,20 @@ async function generateSlideContent(
     }
   }
 
+  const generatedImageEntries = outline.mediaGenerations?.filter((mg) => mg.type === 'image') ?? [];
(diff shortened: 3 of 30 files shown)


YizukiAme:
forgot the on-on condition test! fixed

wyuc (changes requested):
Thanks for the deep dive on this one. The diagnosis in #483 is spot on, the negative-override to source-omission switch is the right direction, and the scene-generator gap fix plus the 4-flag test coverage are great.

One architectural concern before merging: the snippet split feels too fine-grained.

### Concern

6 of the 15 new snippets are ≤2 lines (e.g. `outline-image-fields.md`, `slide-image-user-rule.md`). At that size the file overhead outweighs the modularity benefit, and reading a prompt now requires jumping across many files and mentally simulating the include plus condition engine, instead of reading one markdown top to bottom. Prompt quality is a core competency for us, so keeping templates reviewable matters a lot.

I think we're also coupling two orthogonal concerns here: conditional rendering (a templating concern) and file-level extraction (a modularity concern). They don't have to ship together. Plus `?if=name` is a bare string, so typos silently render nothing and tests still pass.

### Suggested direction

Replace the file-per-fragment pattern with inline conditional blocks:

```
{{#if imageEnabled}}
### AI-Generated Image Requests
...
{{/if}}
```

Concretely:

1. Add `{{#if name}}...{{/if}}` block syntax to the loader (about 10 lines, one regex pass between snippets and interpolation).
2. Inline the 6 micro-snippets back into their host templates as `{{#if}}` blocks.
3. Keep `image-instructions.md`, `video-instructions.md`, and `media-safety-guidelines.md` as plain `{{snippet:name}}` snippets, wrapped in `{{#if}}` at the call site so the condition lives in the host rather than in the filename.
4. Drop `PromptLoadOptions.conditions` and the `?if=` syntax. `{{#if foo}}` can read template variables directly.

Wiring in outline-generator, scene-generator, and the route can stay as is, and your 10 tests carry over with trivial `buildPrompt` simplifications.

### Why request-changes

`lib/prompts/` is touched by everyone tuning generation quality, so it's cheaper to set the right pattern now than to migrate later. Happy to pair or sketch the loader change if that helps.

YizukiAme:
Thanks for the thorough review!

My original thinking was to make every conditional block a standalone file so each snippet could be independently tested and reused across templates. But yeah, at 1-line granularity the file overhead kills readability, and anyone tuning a prompt now has to mentally `cat` 8+ files to see what the model actually receives. 🙃

Reworked in `5185500` following your suggested direction:

1. **Added `{{#if name}}...{{/if}}` block syntax** to the loader (~10 lines, single regex pass between snippets and interpolation)
2. **Inlined 9 micro-snippets** (all ≤4 lines) back into their host templates as `{{#if}}` blocks
3. **Kept the large instruction snippets** (`image-instructions`, `video-instructions`, `slide-image-instructions` etc.) as `{{snippet:name}}` includes, wrapped in `{{#if}}` at the call site
4. **Dropped `?if=` syntax and `PromptLoadOptions.conditions`** — `{{#if}}` reads directly from the `variables` record passed to `buildPrompt`

Processing order: snippets → conditional blocks → variable interpolation. All 37 tests pass, orchestration wiring unchanged.

wyuc (approved):
LGTM :)

PR #496, opened 2026-04-26
[codex] add VoxCPM2 TTS provider

### LINKED ISSUE #495: Add VoxCPM2 TTS provider with browser voice profiles, cloning, and Agent Bar integration
## Background

OpenMAIC should support locally deployed VoxCPM2 as a TTS provider without requiring OpenMAIC to own or wrap the model runtime. The app should act as an adaptor: users configure a backend endpoint, and OpenMAIC forwards TTS requests to that service.

This enables users to run VoxCPM2 through different deployment modes, such as vLLM-Omni, Nano-VLLM, or the official Python API, while keeping the frontend experience consistent.

VoxCPM project: https://github.com/OpenBMB/VoxCPM

## Scope

Add a new VoxCPM2 TTS provider with:

- Configurable backend type and Base URL.
- Configurable model name for OpenAI-compatible backends.
- Browser-local voice profile management.
- Prompt-based voices.
- Reference-audio voice cloning.
- Audio upload and recording for clone references.
- Agent Bar voice selection support.
- TTS preview support for normal VoxCPM voices.
- Dynamic `Auto Voice` support that generates a VoxCPM voice prompt from agent metadata.

### Provider configuration

Users can configure VoxCPM2 from the TTS settings panel:

- Select backend type:
  - vLLM-Omni / OpenAI-compatible speech endpoint.
  - Nano-VLLM endpoint.
  - Official Python API endpoint.
- Enter Base URL.
- Enter model name where applicable.
- Test TTS directly from the settings panel.

OpenMAIC should only adapt request formats and forward to the selected backend.

### Voice profiles

Users can create and manage VoxCPM voice profiles in the browser:

- Prompt voice: name + voice prompt.
- Cloned voice: name + reference audio + optional reference text / description.
- Reference audio can be uploaded or recorded.
- Voice profiles are stored locally in the browser, aligned with the serverless product model.

### Agent Bar integration

Once VoxCPM2 has a valid Base URL, its voices should enter the global Agent Bar voice pool.

Agent Bar should support:

- Selecting VoxCPM voices per agent.
- Searching the voice dropdown.
- Previewing concrete VoxCPM voices.
- Hiding preview for dynamic Auto Voice.
- Continuing to support existing providers and browser voices.

### Auto Voice

VoxCPM should include an `Auto Voice` option.

When selected, OpenMAIC should generate a VoxCPM voice prompt from available agent information, such as:

- Agent name.
- Role.
- Persona / character description.

The generated prompt should be used at synthesis time, not stored as a fixed voice profile.

## Acceptance criteria

- VoxCPM2 appears as a configurable TTS provider.
- Users can configure backend type, Base URL, and model name.
- TTS test works from the settings panel for supported backends.
- Prompt voices can be added, deleted, selected, and previewed.
- Cloned voices can be created from uploaded or recorded reference audio.
- Concrete VoxCPM voices appear in the Agent Bar voice dropdown.
- VoxCPM voice dropdown supports search and has usable density for larger voice lists.
- Auto Voice appears as a selectable VoxCPM voice when the provider is configured.
- Auto Voice generates prompts from agent metadata during playback.
- Auto Voice is not previewable.
- Existing TTS providers continue to work unchanged.
- Saved user settings and existing character voice configs remain compatible.

## Non-goals

- OpenMAIC should not own the VoxCPM model runtime.
- OpenMAIC should not ship built-in cloned voice assets.
- OpenMAIC should not require server-side persistence for browser-created VoxCPM voice profiles.

wyuc:
## Summary

Adds VoxCPM2 as a configurable TTS provider for OpenMAIC, with OpenMAIC acting as an adaptor to user-managed VoxCPM backends rather than owning the runtime.

Related issue: Closes #495

## What changed

- Added VoxCPM2 provider configuration with backend selection, Base URL, and configurable model name for OpenAI-compatible endpoints.
- Added backend request adapters for vLLM-Omni, Nano-VLLM, and the official Python API style endpoint.
- Added browser-local VoxCPM voice profiles for prompt voices and reference-audio cloned voices.
- Added uploaded and recorded reference-audio support for cloned voices.
- Integrated VoxCPM voices into the Agent Bar global voice pool when the provider is configured.
- Added searchable Agent Bar voice selection and preview behavior for concrete VoxCPM voices.
- Added dynamic Auto Voice support that generates a VoxCPM prompt from agent metadata at synthesis time.
- Added VoxCPM settings UI and official VoxCPM logo asset.

## Validation

- `pnpm exec tsc --noEmit`
- Manually tested local VoxCPM endpoints through SSH tunnels during development.

## Notes

- Auto Voice is intentionally not previewable because its prompt depends on agent context.
- Voice profiles are stored in the browser to preserve the serverless setup model.


wyuc:
## Code Review

Reviewed locally against `e613b757..1ad5ae8c`. Architecture is sound (clean adapter, settings migration is correct, existing providers untouched), but there are blockers before merge.

### Blockers

**1. Hardcoded Chinese strings throughout the new UI (i18n regression).**
Only one new key (`providerVoxCPMTTS`) was added per locale. Every other VoxCPM label, placeholder, toast, and aria-label in `components/settings/tts-settings.tsx` (~40 occurrences) and `components/agent/agent-bar.tsx:249-258, 526-535` is a literal Chinese string. Non-zh users see Chinese UI; screen readers get Chinese aria-labels. Also: `ko-KR.json` was not touched at all — please add the key there too.

**2. `components/generation/media-popover.tsx` lost the inline TTS provider/voice/speed picker (-157 lines).**
The TTS tab is now just a toggle + hint text (lines 291-302). This affects *all* providers, not only VoxCPM, and is not mentioned in the PR description. Either restore the picker (passing `voxcpmProfiles` through `getAvailableProvidersWithVoices`) or call out the removal explicitly.

**3. Reference-audio storage is unbounded.**
`lib/audio/voxcpm-voices.ts:171-200` stores user-uploaded WAV blobs in IndexedDB with no size cap, and `tts-providers.ts` base64-encodes them per request. A multi-MB reference balloons by ~33% in the JSON body and may hit Next.js `bodyParser.sizeLimit`. Add a size/duration validator (e.g. ≤10 MB / ≤60s) at upload time with a clear error.

### Important

**4. SSRF guard is dev-bypassed and the VoxCPM default Base URL is `http://127.0.0.1:8000`** (`lib/audio/constants.ts:746`). `app/api/generate/tts/route.ts:61` only validates URLs when `NODE_ENV === 'production'`. With a default-on internal URL, self-hosted deployments without `NODE_ENV=production` expose the OpenMAIC server as an open proxy to the local network. Apply SSRF validation regardless of env, or gate the localhost default behind an explicit opt-in.

**5. Server-side Auto Voice fallback is context-less.**
`tts-providers.ts:240-244` calls `buildAutoVoxCPMVoicePrompt()` with no args when `voicePrompt` is missing, producing the same generic prompt for every agent. Client callers thread context, but `lib/server/classroom-media-generation.ts:260` does not — every classroom-export Auto Voice line gets the fallback. Either thread agent context through, or block Auto Voice in the export pipeline.

**6. `buildAutoVoxCPMVoicePrompt` gender detection is broken.**
`lib/audio/voxcpm.ts:111` requires `/female|woman/i.test(...)` AND `!/male|man/i.test(...)`. The string `"woman"` matches both because of the `man` substring, so the female branch is unreachable for English personas. Use `\bword\b` boundaries.

**7. PII in Auto Voice prompt.**
`voxcpm.ts:118` interpolates `agentName` into the prompt sent to the user-configured backend. Worth a one-line note in settings UI and stripping control chars from `agentName`.

### Minor

- `useVoxCPMVoiceProfiles` is mounted in 6+ components; every voice change triggers 6 IndexedDB scans. Consider a Zustand slice.
- Hardcoded preview text `'你好，这是 VoxCPM 音色试听。'` in `tts-settings.tsx:782` instead of `t('settings.ttsTestTextDefault')`.
- vLLM-Omni payload hardcodes `voice: 'default'` (`tts-providers.ts:285`) — add a comment.
- `getVoxCPMAudioFormat` returns `webm`/`ogg` but `getAudioResponseFormat` doesn't recognize them and falls back to `mp3`.

### Coupling note

The refactor of `getAvailableProvidersWithVoices(..., voxcpmProfiles)` propagated to every TTS consumer. Default arg `= []` softens it, but it's wider call-site sprawl than necessary. The `media-popover.tsx` deletion (#2) is the single biggest coupling concern — TTS picker gone for all providers, not a VoxCPM-scoped change.

### Verdict

**Request changes — do not merge as-is.** Blockers 1 and 2 must be addressed; 3 should be filed as a follow-up if not fixed in this PR. CI Lint/Typecheck and E2E are still pending — wait for green before flipping to ready.

wyuc:
Addressed the review feedback in `0f72e15`.

Summary:
- Moved VoxCPM UI strings behind i18n keys and verified locale key alignment for the 5 supported locale files.
- Restored the media popover TTS provider / voice / speed picker, including VoxCPM browser profiles.
- Added reference-audio validation before saving: max 10 MB and max 60 seconds.
- Applied SSRF validation outside production too; local-network TTS now requires explicit `ALLOW_LOCAL_NETWORKS` opt-in.
- Simplified Auto Voice prompt generation: it now uses agent `persona` directly as the VoxCPM voice prompt, with only a small role/name fallback.
- Avoided server-side classroom export using VoxCPM Auto Voice without agent context.
- Removed the unused `selectedVoxCPMBackend`, switched preview copy to existing i18n text, documented vLLM `voice: default`, and recognized `webm`/`ogg` audio response formats.

Validation:
- `pnpm exec tsc --noEmit`
- `pnpm run lint` (passes with existing warnings)
- `pnpm run check:i18n-keys`
- Prettier check on touched files

wyuc:
Follow-up UI cleanup pushed in `aec7085`:

- Restored the media popover TTS tab to a switch-only control. Provider, voice, speed, and preview remain in the unified Agent Bar/settings flow instead of being duplicated here.
- Simplified the toolbar model pill to show only the provider icon plus the thinking setting, keeping the full provider/model details in the tooltip and model picker.

Validation: `pnpm exec tsc --noEmit`, focused Prettier check, and `pnpm run lint` pass. Lint still reports only pre-existing warnings unrelated to this PR.

cosarah (commented):
Independent review of HEAD aec7085. Prior round's blockers (i18n, ref-audio cap, SSRF env-agnostic, Auto Voice fallback, classroom export, media popover) all confirmed landed.

### Important

- `lib/audio/voxcpm-voices.ts:301` — `normalizeVoxCPMReferenceAudio` runs on every TTS request, but the stored blob is already normalized at save time (line 237). This re-decodes via Web Audio API and re-encodes WAV per request, spawning a fresh AudioContext each time. Use the stored blob directly.
- `lib/audio/voxcpm-voices.ts:20-21, 230-260` — Per-profile size/duration caps land, but there's no profile-count cap, no eviction, and no `QuotaExceededError` handling around `db.voiceProfiles.put`. Once IndexedDB fills, saves throw and surface only as a generic toast. Add an upper bound (e.g. 50 profiles) or wrap with a quota-aware error path.
- `lib/audio/voxcpm-voices.ts:182-204` — `useVoxCPMVoiceProfiles` is mounted in settings, agent-bar, providers-with-voices, etc. Every `VOXCPM_VOICE_PROFILES_CHANGED` event triggers N independent IndexedDB scans. Consider a Zustand slice or shared subscription.
- `lib/audio/tts-providers.ts:243-245` — When `voice === VOXCPM_AUTO_VOICE_ID` and no `voicePrompt` is passed, the server throws "VoxCPM Auto Voice requires agent context", which `/api/generate/tts` returns as a 500 `GENERATION_FAILED`. Validate at the route layer and return 400 with a specific code.

### Minor

- `lib/audio/voxcpm-voices.ts:174` — Hardcoded `language: 'zh'` for all browser-saved profiles regardless of source audio. Voice-resolver language matching will mis-rank profiles cloned from non-Chinese audio. Prompt for language at save, or default to `'auto'`.
- `lib/audio/voxcpm-voices.ts:34-42` — `blobToBase64` builds the data string char-by-char; for a 10 MB WAV that's ~13 MB of string concat in a tight loop. Prefer `FileReader.readAsDataURL`.
- `lib/audio/voxcpm.ts:96-102` — Persona is sanitized for control chars but unbounded in length. A multi-KB persona becomes the entire voice prompt verbatim. Truncate to ~200 chars.
- `lib/audio/tts-providers.ts:285-288` — `buildVoxCPMTargetText` strips `()（）` to protect the inline `(prompt)text` syntax, but not newlines or other control chars; multi-line persona will produce malformed targetText.
- `lib/audio/voxcpm.ts:5` vs `VOXCPM_MODEL_ID` — Dual model IDs (`voxcpm2` vs `VoxCPM2`) wired through a settings migration. Fine for now; kill the duplicate once the migration window closes.
- `lib/utils/database.ts:196` — `_DATABASE_VERSION = 10` is unused; either consume it in `this.version()` or delete.
- `components/generation/media-popover.tsx:291-298` — TTS tab is intentionally toggle-only now, but provider/voice/speed are reachable only via Settings or Agent Bar. A one-line hint here would aid discoverability.

### Verdict

Comment-level review. No remaining blockers; Important items above are real but addressable.

wyuc:
Follow-up fixes pushed in `27ade4a`:

- Avoid re-normalizing cloned reference audio on every TTS request; stored normalized blobs are now used directly.
- Added route-level validation for VoxCPM Auto Voice without `voicePrompt`, returning 400 with `VOXCPM_AUTO_VOICE_REQUIRES_CONTEXT` instead of a generic 500.
- Sanitized Auto Voice prompt inputs and capped generated prompts to 200 chars.
- Sanitized inline VoxCPM target prompts by removing control chars/newlines and parentheses before building `(prompt)text`.
- Switched reference-audio base64 conversion to `FileReader.readAsDataURL`.
- Marked browser-saved VoxCPM profiles as `language: auto` in both profile options and voice resolver output.

Validation: `pnpm exec tsc --noEmit`, focused Prettier check, and `pnpm run lint` pass. Lint still reports only existing unrelated warnings.

wyuc:
Tested in 3 backends on an A800 80GB machine
@cosarah

cosarah (approved):
LGTM

PR #501, opened 2026-04-27
[codex] Fix Claude Haiku 4.5 thinking controls

### LINKED ISSUE #497: Claude Haiku 4.5 should not expose thinking effort option
## Problem

In `lib/ai/model-metadata.ts:235`, `claude-haiku-4-5` is registered with `anthropicManualEffort`, which exposes effort values `['none', 'low', 'medium', 'high', 'max']` and an adjustable thinking budget.

However, Claude Haiku 4.5 does not support extended thinking / effort selection. Showing the effort selector in the per-model thinking config is misleading and the request adapter may emit an unsupported `thinking` parameter.

## Expected

Configure Claude Haiku 4.5 as a non-thinking model — remove the entry from `THINKING_CAPABILITIES` (or use a no-thinking capability) so:

- Settings UI does not render an effort/budget selector for it.
- API requests do not include the `thinking` field.

## Reference

- `lib/ai/model-metadata.ts:235` — current registration
- `lib/ai/model-metadata.ts:112-125` — `anthropicManualEffort` definition

wyuc:
## Summary

- Change `claude-haiku-4-5` from Anthropic manual effort levels to a budget-based Anthropic thinking capability.
- Map Anthropic budget controls to `thinking: { type: 'enabled', budgetTokens }` without sending `effort` / `output_config.effort`.
- Add metadata, providerOptions, and Anthropic request serialization tests for Claude Haiku 4.5.

## Context

Fixes #497.

Claude Haiku 4.5 supports extended thinking, but the supported control is Anthropic thinking budget (`thinking.budget_tokens`), not the manual effort selector exposed by the previous `anthropicManualEffort` metadata.

## Validation

- `pnpm exec vitest run tests/ai/thinking-config.test.ts tests/ai/llm-thinking-options.test.ts tests/ai/anthropic-serialization.test.ts`

The serialization test uses a mocked `fetch` and does not send any real Anthropic API request.

### DIFF
--- a/lib/ai/llm.ts
+++ b/lib/ai/llm.ts
@@ -148,6 +148,14 @@ function buildThinkingProviderOptions(
 
     case 'anthropic': {
       if (mode === 'disabled') return { anthropic: { thinking: { type: 'disabled' } } };
+
+      if (thinking.control === 'toggle-budget' || thinking.control === 'budget-only') {
+        const budget = pickThinkingBudget(thinking, config);
+        return budget === undefined
+          ? undefined
+          : { anthropic: { thinking: { type: 'enabled', budgetTokens: budget } } };
+      }
+
       const effort = getAnthropicEffort(thinking, config);
       if (!effort) return undefined;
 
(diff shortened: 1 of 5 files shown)


cosarah (approved):
tested locally, LGTM

PR #505, opened 2026-04-27
fix(export): skip malformed SVG path shapes instead of aborting PPTX export

### LINKED ISSUE #504: Export PPTX fails outright when a slide contains a malformed SVG path
## Summary

Importing a `.maic.zip` course whose `manifest.json` contains a single malformed SVG `path` on any slide makes the **Export PPTX** action fail entirely with a "导出失败 / Export failed" toast — no PPTX file is produced, even though the rest of the course is fine and the slide renders normally in the canvas.

## Reproduction

1. Import a course archive that contains at least one shape with an unparseable `path` value on a slide.
   - Real-world example we hit: a shape whose `path` was `M 1 0.5 alert 0.5 0.5 0 1 1 0 0.5 A 0.5 0.5 0 1 1 1 0.5 Z` — likely an upstream LLM hallucination where an `A` arc command got expanded into the literal token `alert`.
2. Open the imported classroom — it loads and renders fine. Browser console shows a non-fatal SVG warning: `<path> attribute d: Expected number, "M 1 0.5 alert 0.5 0.5 0 1…"`.
3. Click the download icon → **Export PPTX**.
4. The toast immediately shows "导出失败 / Export failed". No file downloads.

## Root cause

In `lib/export/svg-path-parser.ts`, `toPoints(d)` calls `new SVGPathData(d)` directly. `svg-pathdata` throws synchronously on unrecognised commands, e.g.:

```
SyntaxError: Unexpected character "l" at index 9. Command cannot follow comma
    at SVGPathDataParser.parse
    at new SVGPathData
    at toPoints (lib/export/svg-path-parser.ts)
    at buildPptxBlob (lib/export/use-export-pptx.ts, shape branch)
```

That throw propagates up through `buildPptxBlob` and is caught only by the top-level `withExportGuard`, which surfaces a generic "Export failed" toast. **One bad path on one slide blocks the entire export.**

For comparison, the sibling helper `getSvgPathRange` in the same file already wraps `new SVGPathData(...)` in `try/catch` and returns a zero range on parse failure — but `toPoints` does not, so the export pipeline is the strictest layer in the stack.

## Expected behaviour

A single malformed shape path should be skipped (with a warning logged), and the rest of the slides should be exported normally — matching how the in-app slide renderer already tolerates the same input.

## Environment

- Branch: `main` @ 93c4cbb
- Repro reliably with the `.maic.zip` described above. Affects both `Export PPTX` and the resource-pack export (which calls `buildPptxBlob` internally).

wyuc:
## Summary
- Make `toPoints` (in `lib/export/svg-path-parser.ts`) mirror the tolerant pattern of its sibling `getSvgPathRange`: warn and return `[]` on parse failure, instead of letting `svg-pathdata`'s `SyntaxError` propagate.
- In the shape branch of `buildPptxBlob` (`lib/export/use-export-pptx.ts`), skip the element when `toPoints` returns no points so a single bad path no longer blocks the entire export.
- Add unit tests covering the happy path and the exact malformed input that triggered #504 (`M 1 0.5 alert 0.5 0.5 0 1 1 0 0.5 A 0.5 0.5 0 1 1 1 0.5 Z`).

Fixes #504.

## Test plan
- [x] `pnpm exec vitest run tests/export/svg-path-parser.test.ts` — 3/3 pass.
- [x] `pnpm exec tsc --noEmit` and `pnpm exec eslint` clean on touched files.
- [x] Playwright repro spec exercising the full UI flow (import affected `.maic.zip` → open classroom → click Export PPTX) captured the original `SyntaxError: Unexpected character "l" ...` in `Export failed:` console output before the fix, and "Export successful" toast after the fix.
- [ ] **Not yet verified manually in a browser** — reviewer should sanity-check by importing a problematic `.maic.zip` and clicking Export PPTX.

### DIFF
--- a/lib/export/svg-path-parser.ts
+++ b/lib/export/svg-path-parser.ts
@@ -1,5 +1,8 @@
 import { SVGPathData } from 'svg-pathdata';
 import arcToBezier from 'svg-arc-to-cubic-bezier';
+import { createLogger } from '@/lib/logger';
+
+const log = createLogger('SvgPathParser');
 
 const typeMap = {
   1: 'Z',
@@ -32,9 +35,20 @@ export type SvgPath = ReturnType<typeof parseSvgPath>;
 /**
  * 解析SVG路径，并将圆弧（A）类型的路径转为三次贝塞尔（C）类型的路径
  * @param d SVG path d属性
+ *
+ * Returns an empty array if the path is malformed (e.g. unrecognised commands).
+ * Mirrors the defensive behaviour of {@link getSvgPathRange}: a single bad path
+ * (often produced by upstream LLM hallucinations) shouldn't take down the whole
+ * PPTX export.
  */
 export const toPoints = (d: string) => {
-  const pathData = new SVGPathData(d);
+  let pathData: SVGPathData;
+  try {
+    pathData = new SVGPathData(d);
+  } catch (err) {
+    log.warn(`Failed to parse SVG path "${d}":`, err);
+    return [];
+  }
 
   const points = [];
   for (const item of pathData.commands) {
--- a/lib/export/use-export-pptx.ts
+++ b/lib/export/use-export-pptx.ts
@@ -581,7 +581,9 @@ async function buildPptxBlob(
             x: el.width / el.viewBox[0],
             y: el.height / el.viewBox[1],
           };
-          const points = formatPoints(toPoints(el.path), ratioPx2Inch, scale);
+          const rawPoints = toPoints(el.path);
+          if (!rawPoints.length) continue; // Malformed path — toPoints already logged.
+          const points = formatPoints(rawPoints, ratioPx2Inch, scale);
 
           let fillColor = formatColor(el.fill);
           if (el.gradient) {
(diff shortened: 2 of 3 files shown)


wyuc:
## Code review summary

Ran `superpowers:code-reviewer` against `93c4cbb..7f993f4`. Verdict: **Ready to merge: Yes** — no Critical or Important issues.

**Strengths called out**
- Pattern matches the sibling `getSvgPathRange` try/catch idiom in the same file.
- 46 lines / 3 files, no scope creep.
- Regression test pins the exact malformed input from #504; warn goes through `createLogger` (not `console`).
- Coupling audited: `toPoints` has only 2 callers (both in `use-export-pptx.ts`); `formatPoints([])` safely returns `[]`; no other consumers of `SvgPoints` rely on non-emptiness.

**Minor suggestions (not blocking)**
1. *Line element branch (`use-export-pptx.ts:669`)* could get the same empty-check guard for symmetry. Reviewer agrees the current behavior is safe — `getLineElementPath` is built from numeric coords, can't produce unrecognised-command tokens — so I'm leaving it as-is. Worth an inline comment here in a future cleanup pass.
2. Optionally add a `toPoints('')` test. Low value; skipped.
3. `parseSvgPath` and the `SvgPath` type are dead exports with the same throw-on-bad-input weakness. Out of scope for this PR; flagged for a follow-up cleanup ticket.

No code changes from review. Marking ready to merge once a maintainer eyeballs.

cosarah (approved):
tested locally, LGTM

PR #513, opened 2026-05-01
feat: add Xiaomi MiMo TTS provider with voice cloning support

ql-wade:
## Summary

Add **Xiaomi MiMo** as a built-in TTS provider, enabling text-to-speech synthesis via Xiaomi's MiMo API (OpenAI-compatible chat/completions endpoint with audio modality).

### Features

- **`generateXiaomiTTS()`** — Full TTS implementation using `/chat/completions` + `modalities=["text","audio"]`
- **4 models supported:**
  - `mimo-v2.5-tts` (default)
  - `mimo-v2.5-tts-voiceclone` (voice cloning from reference audio)
  - `mimo-v2.5-tts-voicedesign` (custom voice design)
  - `mimo-v2-tts` (legacy V2)
- **9 built-in voices:** 冰糖, 茉莉, 苏打, 白桦 (Chinese) + Mia, Chloe, Milo, Dean (English) + mimo_default
- **MP3 output**, configurable speed range 0.5x–2.0x
- **Domestic China endpoint:** `https://token-plan-cn.xiaomimimo.com/v1` (no proxy needed)

### Files Changed

| File | Change |
|------|--------|
| `lib/audio/tts-providers.ts` | +58 lines: `generateXiaomiTTS()` implementation + case branch |
| `lib/audio/constants.ts` | +27 lines: Provider registration (models, voices, config) |
| `lib/audio/types.ts` | +1 line: Type definition for `xiaomi-tts` |
| `lib/store/settings.ts` | +1 line: Default settings entry |
| `components/settings/audio-settings.tsx` | +1 line: UI display name |
| `lib/server/provider-config.ts` | +1 line: `TTS_XIAOMI` env var mapping |

### Configuration

```bash
# .env.local
TTS_XIAOMI_API_KEY=your_xiaomi_api_key
TTS_XIAOMI_BASE_URL=https://token-plan-cn.xiaomimimo.com/v1  # optional, default provided
```

Then go to **Settings → Text-to-Speech → Xiaomi MiMo TTS** to configure.

### Testing

Tested with MiMo V2.5 TTS model. Voice cloning and voice design modes follow the same API pattern.

📌 Related: This complements the existing Xiaomi MiMo LLM provider (`xiaomi:mimo-v2.5-pro`) added in v0.2.1.

### DIFF
--- a/components/settings/audio-settings.tsx
+++ b/components/settings/audio-settings.tsx
@@ -44,6 +44,7 @@ function getTTSProviderName(providerId: TTSProviderId, t: (key: string) => strin
     'doubao-tts': t('settings.providerDoubaoTTS'),
     'elevenlabs-tts': t('settings.providerElevenLabsTTS'),
     'minimax-tts': t('settings.providerMiniMaxTTS'),
+    'xiaomi-tts': 'Xiaomi MiMo TTS',
     'browser-native-tts': t('settings.providerBrowserNativeTTS'),
   };
   return names[providerId];
--- a/lib/audio/constants.ts
+++ b/lib/audio/constants.ts
@@ -928,6 +928,33 @@ export const TTS_PROVIDERS: Record<BuiltInTTSProviderId, TTSProviderConfig> = {
     speedRange: { min: 0.7, max: 1.2, default: 1.0 },
   },
 
+  'xiaomi-tts': {
+    id: 'xiaomi-tts',
+    name: 'Xiaomi MiMo TTS',
+    requiresApiKey: true,
+    defaultBaseUrl: 'https://token-plan-cn.xiaomimimo.com/v1',
+    models: [
+      { id: 'mimo-v2.5-tts', name: 'MiMo V2.5 TTS' },
+      { id: 'mimo-v2.5-tts-voiceclone', name: 'MiMo V2.5 TTS VoiceClone' },
+      { id: 'mimo-v2.5-tts-voicedesign', name: 'MiMo V2.5 TTS VoiceDesign' },
+      { id: 'mimo-v2-tts', name: 'MiMo V2 TTS' },
+    ],
+    defaultModelId: 'mimo-v2.5-tts',
+    voices: [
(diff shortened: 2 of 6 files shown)


wyuc (changes requested):
Xiaomi MiMo TTS would be a useful domestic provider, but several advertised paths are not implemented yet:

1. `generateXiaomiTTS()` says it uses chat completions with `modalities=["text", "audio"]`, and the PR summary makes the same claim, but the actual request body never sends `modalities`. As written, there is no explicit request for an audio response. Please align the payload with the verified MiMo API contract and cover the exact request body in a provider test.
2. The provider exposes `mimo-v2.5-tts-voiceclone` and `mimo-v2.5-tts-voicedesign`, and the PR title claims voice-cloning support, but all four models use the same ordinary `{ messages, audio: { voice, format } }` payload. There is no reference-audio upload/configuration, clone identifier, or voice-design prompt anywhere in the UI/config path. These options are therefore misleading/nonfunctional. Please either implement and test their required inputs end to end, or remove those models and the voice-cloning claim from this focused PR.
3. The provider declares a configurable `speedRange`, but `generateXiaomiTTS()` never reads or sends `config.speed`, so changing speed in Settings has no effect. Implement the API's supported speed field or do not advertise the control.
4. There are no Xiaomi provider tests. Please add mocked success/error/malformed-response coverage, assert the model/voice/audio payload, and include one real API smoke result proving the returned base64 decodes to playable MP3.

The branch also conflicts with current `main`, where the TTS settings/provider architecture has evolved, and the existing CI fails Prettier on `lib/audio/constants.ts`. Please rebase, resolve the current integration points, and get refreshed CI green before re-review.

wyuc:
Thanks @ql-wade for the early MiMo TTS work. Closing in favour of #1432, which adds built-in Xiaomi MiMo TTS and ASR providers on the current audio stack. If voice cloning is still something you'd like to see, a follow-up on top of #1432 would be very welcome.

Issue #539, opened 2026-05-09
Refactor generated video handling into a first-class VideoManifest


wyuc:
## Problem

Generated videos currently rely on placeholder IDs flowing through multiple loosely coupled steps:

1. Outline generation creates `mediaGenerations` entries.
2. Sequential IDs like `gen_vid_1` are rewritten into globally unique IDs.
3. Video generation uses the rewritten ID as the media task key.
4. Slide content generation asks the model to copy that same ID into `VideoElement.src`.
5. Renderer/export/action code later resolves `src` through the media generation store.

This makes `VideoElement.src` carry two different meanings:

- a real playable URL
- a generated video placeholder ID like `gen_vid_xxx`

The flow is fragile because the slide JSON depends on the model copying the exact generated video ID. If the model emits a valid-looking but wrong placeholder, e.g. `gen_vid_1`, the video can be generated successfully but never render, because the slide points to a different key.

## Proposed Direction

Introduce a stage/classroom-level `VideoManifest` and make generated video references explicit.

Example:

```ts
type VideoManifest = Record<string, {
  type: 'video';
  prompt: string;
  aspectRatio?: string;
  status: 'pending' | 'generating' | 'done' | 'failed';
  url?: string;
  poster?: string;
  error?: string;
}>;
```

Slide video elements should reference generated videos via `mediaRef`:

```ts
{
  type: 'video',
  mediaRef: 'video_abc123'
}
```

`src` should eventually mean only a concrete playable URL. Generated videos should be resolved through a shared resolver:

```ts
resolveVideo(stageId, mediaRef)
```

## Scope

This proposal intentionally covers generated videos only.

Images, PDF source images, generated images, audio, and other media types should continue using the current flow for now. If the video path works well, the same pattern can be generalized into a broader `MediaManifest` later.

## New Flow

1. Generate structured outlines.
2. Collect video generation requests from outline `mediaGenerations`.
3. Create `VideoManifest` entries with system-owned video IDs.
4. Start video generation tasks from the manifest.
5. Generate slide JSON using `mediaRef` for video elements.
6. Validate that every video `mediaRef` exists in `VideoManifest`.
7. Renderer/export/action engine resolve `mediaRef` through the manifest.
8. Once video generation finishes, manifest entries receive `url` and optional `poster`.

## Benefits

- Removes ambiguity from `VideoElement.src`
- Stops relying on the model to preserve placeholder IDs exactly
- Gives generated videos a proper data contract
[...]

PR #540, opened 2026-05-09
feat: add generated video manifest refs

wyuc:
## Summary

- Introduce a video-only manifest helper and `mediaRef` support for generated video elements.
- Update slide video prompts and post-processing so generated videos use validated `mediaRef` values instead of `src: gen_vid_*`.
- Resolve generated videos through `mediaRef` in renderer, actions, PPT export, and hosted classroom media replacement while preserving legacy `src: gen_vid_*` fallback.

Related: #539

## Test Plan

- `pnpm exec tsc --noEmit`
- `ALLOW_LOCAL_NETWORKS=false pnpm test`
- `pnpm lint` (0 errors, existing warnings remain)

### DIFF
--- a/app/generation-preview/page.tsx
+++ b/app/generation-preview/page.tsx
@@ -23,6 +23,7 @@ import {
 import { getCurrentModelConfig } from '@/lib/utils/model-config';
 import { db } from '@/lib/utils/database';
 import { MAX_PDF_CONTENT_CHARS, MAX_VISION_IMAGES } from '@/lib/constants/generation';
+import { buildVideoManifestFromOutlines } from '@/lib/media/video-manifest';
 import { nanoid } from 'nanoid';
 import type { Stage } from '@/lib/types/stage';
 import type { SceneOutline, PdfImage, ImageMapping } from '@/lib/types/generation';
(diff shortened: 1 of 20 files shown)


wyuc:
Agent CR summary:

- Ran a focused subagent code review over `main..feat/video-manifest`, covering the video manifest/mediaRef flow, renderer/export/server resolution paths, backward compatibility, and tests.
- Review found no Critical issues.
- Important findings were:
  - hallucinated generated video refs were not removed when a slide had no generated video requests;
  - direct video `src` could be overwritten when `mediaRef` was also present;
  - `videoManifest` lifecycle fields could become stale against the client media store.
- Addressed these in `90fac28`:
  - invalid generated video refs are now removed even when there are zero valid video requests;
  - direct video URLs are preserved and stray `mediaRef` is dropped during normalization;
  - server media replacement only fills missing/placeholder `src`, never overwrites direct URLs;
  - `videoManifest` is now explicitly a generated video request index, with runtime status kept in media tasks/media files;
  - added regression coverage for these cases.

Validation after fixes:

- `pnpm check`
- `pnpm exec tsc --noEmit`
- `pnpm lint` (0 errors, existing warnings remain)
- `ALLOW_LOCAL_NETWORKS=false pnpm test`
- GitHub CI run `25597992934` passed: E2E + Lint/Typecheck/Unit Tests.

cosarah (approved):
Tested generation, PPT import/export, and legacy `src: gen_vid_*` fallback — all working. Manifest resolution clean across renderer, actions, export, and hosted media paths. LGTM.

PR #544, opened 2026-05-10
fix(whiteboard): code element captures internal scroll/drag (#530)

### LINKED ISSUE #530: [Bug]: Whiteboard code element cannot reliably capture internal scroll/drag when content overflows
﻿## Bug Description
When a code element on the whiteboard contains long lines or enough lines to overflow its visible area, interacting inside the code element does not reliably scroll or inspect the code itself. Whiteboard pan/zoom/selection handling captures the interaction first, which makes the code content difficult to read.

## Steps to Reproduce
1. Add a code element to the whiteboard.
2. Make the code content overflow horizontally and/or vertically.
3. Try to scroll inside the code area with a mouse wheel or trackpad.
4. Try dragging inside the code area to inspect overflowing content.
5. Observe that the whiteboard often pans, zooms, or selects instead of letting the code element consume the interaction.

## Expected Behavior
When the pointer is over a scrollable code element, scroll and drag gestures intended for the code viewport should be handled by the code element first. Whiteboard pan/zoom should only take over when the code element has no remaining scrollable room in that direction.

## Actual Behavior
The whiteboard interaction layer captures the gesture before the code element can fully consume it, especially when content overflows. This makes long or tall code blocks hard to inspect.

## Deployment Method
Local development (pnpm dev / npm run dev / yarn dev)

## Browser
Chrome

## Operating System
Windows 11

## Relevant Logs / Screenshots
No console error.

Relevant implementation references:
- `components/slide-renderer/components/element/CodeElement/BaseCodeElement.tsx`: wheel interception only exists for the code body.
- `components/slide-renderer/components/element/CodeElement/BaseCodeElement.tsx`: the current logic only checks vertical scrollability and does not cover horizontal overflow or pointer-down/drag propagation.
- `components/slide-renderer/Editor/Canvas/index.tsx`: the canvas blank-area handler captures `onMouseDown` and routes into selection or viewport dragging.

## Additional Context
Other editable element wrappers in the slide renderer commonly stop propagation on pointer or mouse down, but the code element path appears to lack equivalent protection for internal interactions. Horizontal overflow is also not covered by the current wheel logic.

cosarah:
## Summary

Whiteboard code elements with overflowing content now reliably capture internal scroll, wheel, and drag gestures instead of being preempted by the whiteboard's pan/zoom layer. Code content also stays in sync with its surrounding element box during a whiteboard pan.

## Related Issues

Closes #530

## Changes

- `BaseCodeElement`: drag-to-scroll inside the code body via pointer capture (works for both axes), so users can pan code content with the mouse.
- Wheel events on the code element are contained: native wheel listeners stop propagation so the whiteboard zoom handler never fires while the cursor is over code.
- Wrap rendered lines in a `min-width: max-content` container so horizontal overflow is reachable.
- Wrapper `onPointerDown` / `onClick` stop synthetic propagation so clicks on the header / border do not start a whiteboard pan.
- `userSelect: 'none'` + `touchAction: 'none'` on the code body — drag-scroll is the unified gesture (chosen for consistency over conditional select-vs-drag based on overflow state).
- Stop the per-line code lag during whiteboard pan by memoizing the `AnimatedElement` wrapper and removing `layout="position"` from `CodeLineRow`'s motion.div.
- Handle `lostpointercapture` and unmount-mid-drag so capture / cursor never leak.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Add a code element to the whiteboard with content that overflows both vertically and horizontally.
2. Move the cursor over the code body and use the mouse wheel — only the code scrolls; whiteboard zoom is not triggered.
3. Press and drag inside the code body — the content scrolls in both axes; the whiteboard does not pan.
4. Move the cursor outside the code element and wheel / drag — whiteboard zoom and pan work normally.
5. Pan the whiteboard while a code element is on screen — the code lines stay aligned with the element box (no visible lag).
6. Alt-tab away mid-drag and back — the cursor returns to `grab` and a fresh drag works.

- Wheel-scroll vertical and horizontal inside the code body, including Shift+wheel.
- Drag-scroll vertical and horizontal inside the code body.
- Wheel and drag outside the code element (header / border / blank canvas) still drive whiteboard zoom / pan.
- Whiteboard pan no longer causes code content to lag behind the element box.
- Type-checking passes (`npx tsc --noEmit`).

Not verified:
- Touch / mobile devices.
- Code element inside the slide editor (the editor does not currently mount the code element type).

### Evidence

- [x] Manually tested locally

## Checklist

### DIFF
--- a/components/slide-renderer/components/element/CodeElement/BaseCodeElement.tsx
+++ b/components/slide-renderer/components/element/CodeElement/BaseCodeElement.tsx
@@ -366,32 +366,108 @@ function CodeLineRow({
 
 export function BaseCodeElement({ elementInfo, animate }: BaseCodeElementProps) {
   const { language, lines, fileName, showLineNumbers = true, fontSize = 14 } = elementInfo;
+  const wrapperRef = useRef<HTMLDivElement>(null);
   const codeBodyRef = useRef<HTMLDivElement>(null);
 
-  // Prevent wheel events from bubbling to the whiteboard zoom handler
-  // when the code body has scrollable content in the wheel direction.
+  // Drag-to-scroll inside the code body. Press and drag with the primary button
+  // to pan the code content vertically and horizontally. The pointer is
+  // captured so the gesture survives leaving the body bounds, and pointer
+  // capture also implicitly prevents the whiteboard pan from kicking in.
+  useEffect(() => {
+    const el = codeBodyRef.current;
+    if (!el) return;
+
+    let dragging = false;
+    let startX = 0;
+    let startY = 0;
+    let startScrollLeft = 0;
+    let startScrollTop = 0;
+    let activePointer: number | null = null;
+
+    const onPointerDown = (e: PointerEvent) => {
+      if (e.button !== 0) return;
+      dragging = true;
+      activePointer = e.pointerId;
+      startX = e.clientX;
+      startY = e.clientY;
+      startScrollLeft = el.scrollLeft;
+      startScrollTop = el.scrollTop;
+      el.setPointerCapture(e.pointerId);
+      el.style.cursor = 'grabbing';
+    };
+
+    const onPointerMove = (e: PointerEvent) => {
+      if (!dragging || e.pointerId !== activePointer) return;
+      el.scrollLeft = startScrollLeft - (e.clientX - startX);
+      el.scrollTop = startScrollTop - (e.clientY - startY);
+    };
+
+    const endDrag = (e: PointerEvent) => {
+      if (e.pointerId !== activePointer) return;
+      dragging = false;
+      activePointer = null;
+      if (el.hasPointerCapture(e.pointerId)) el.releasePointerCapture(e.pointerId);
+      el.style.cursor = 'grab';
+    };
+
+    el.style.cursor = 'grab';
+    el.addEventListener('pointerdown', onPointerDown);
+    el.addEventListener('pointermove', onPointerMove);
+    el.addEventListener('pointerup', endDrag);
+    el.addEventListener('pointercancel', endDrag);
+    return () => {
+      el.removeEventListener('pointerdown', onPointerDown);
+      el.removeEventListener('pointermove', onPointerMove);
+      el.removeEventListener('pointerup', endDrag);
+      el.removeEventListener('pointercancel', endDrag);
+    };
+  }, []);
+
+  // Always consume wheel events inside the code body so the whiteboard zoom
+  // handler never fires while the cursor is over code. Users can wheel outside
+  // the code element to zoom.
   useEffect(() => {
     const el = codeBodyRef.current;
     if (!el) return;
 
     const handleWheel = (e: WheelEvent) => {
-      const { scrollTop, scrollHeight, clientHeight } = el;
-      const canScroll = scrollHeight > clientHeight;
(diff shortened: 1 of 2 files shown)


cosarah:
## Self-review notes

Ran an independent code review against this branch and applied the actionable feedback. Recording it here for transparency.

### Findings vs. fixes

| # | Severity | Finding | Resolution |
|---|----------|---------|------------|
| 1 | BUG | `touchAction: 'none'` + `userSelect: 'none'` removes browser-native touch scroll and text selection inside the code body. | **Kept intentionally.** Decided to make drag-to-scroll the single unified gesture (decision A): consistent UX across overflow / non-overflow code blocks, no mode-switching. Touch is covered by the pointer-event drag path; selection is sacrificed to keep the rule simple. |
| 2 | BUG | Drag-to-scroll fires even when content does not overflow, swallowing clicks / caret placement. | Same as #1 — accepted as the cost of a single consistent gesture. |
| 3 | BUG | `lostpointercapture` not handled; an interrupted drag (alt-tab, devtools) could leave `dragging` stuck and the cursor at `grabbing`. | **Fixed.** Added `lostpointercapture` listener that runs the same end-drag cleanup; the same cleanup runs on unmount. |
| 4 | IMPROVEMENT | Wrapper-level native `pointerdown` / `mousedown` / `click` listeners are redundant — whiteboard pan is itself a React `onPointerDown`, so synthetic stopPropagation suffices. | **Fixed.** Replaced the native listeners with React `onPointerDown` / `onClick` on the wrapper, matching the `BaseVideoElement` pattern. Removed ~25 lines. |
| 5 | IMPROVEMENT | Wrapper `wheel` listener registered with `passive: true` is misleading — it does not call `preventDefault`. | **Fixed.** Removed the `passive: true` flag (it had no effect since we only stopPropagation). The wrapper wheel listener stays as a native listener because the whiteboard's wheel handler is itself native. |
| 6 | IMPROVEMENT | `memo(AnimatedElement)` comment about "every frame" is inaccurate — the parent rerenders on every pan-state setState, not every frame. | **Fixed.** Reworded the comment. |
| 7 | IMPROVEMENT | Removing `layout="position"` AND adding `memo` is belt-and-suspenders; line insert in the middle of a block now jumps rather than slides. | **Accepted.** The lag during whiteboard pan was the user-visible regression; insert-jump on a non-final line is a much rarer path during live code generation, where lines typically stream in at the bottom. Keeping both fixes is the safer choice. |
| 8-13 | NIT | Misc style / dead-code points (combine the two body effects, fewer cursor writes, single ref read). | **Mostly fixed.** The two body-level effects were merged into one. The cursor is still set imperatively but cleaned up on every drag end. |
| 14 | NIT | Mobile / touch not covered by tests. | Acknowledged — manual desktop verification only. |

### What is intentionally not fixed

- Text selection inside code is gone for everyone. Trade-off discussed above; the original issue (#530) was specifically about scroll/drag capture, not selection.
- Browser Ctrl+wheel zoom is swallowed inside the code body (already true at the whiteboard viewport level — not a regression).

wyuc (approved):
Chrome E2E verified locally on PR head: classroom smoke passed; code element wheel / horizontal wheel / drag scroll internally without whiteboard pan/zoom; outside drag/wheel still pan/zoom the whiteboard; no page errors observed.

PR #552, opened 2026-05-10
fix: prevent memory leaks and silent export failures

arnow117:
## Summary

- **use-browser-asr**: Added cleanup on unmount to abort active SpeechRecognition and release the microphone. Without this, navigating away while recording leaves the mic locked.
- **use-streaming-text**: Stabilized `onComplete` callback with `useRef` to prevent animation restarts when parent re-renders. Previously, any re-render that created a new `onComplete` function reference would restart the entire streaming animation.
- **use-export-pptx**: Added `resp.ok` checks at 3 `fetch()` calls to avoid silently embedding corrupted images/media when HTTP requests fail (e.g., 403, 404, 500). Failed resources are now skipped with a warning log.

## Test plan

- [ ] Verify mic is released after unmounting a component that uses `useBrowserASR` while recording
- [ ] Verify streaming text animation does not restart when a sibling component re-renders
- [ ] Export a presentation containing external images that return 404 — confirm no crash and a warning is logged instead of a corrupted PPTX


wyuc (approved):
LGTM

PR #554, opened 2026-05-10
fix(orchestration): restore agent attribution in director summary

### LINKED ISSUE #511: Multi-agent dialogue: off-topic replies, role confusion, premature discussion end
## Problem

Multi-agent classroom dialogue has three recurring quality issues:

1. **Off-topic responses (答非所问)** — agents reply without addressing the actual question content.
2. **Role confusion between user and student agents** — teacher/director treats the human user as if they were a student agent, or attributes user statements to student agents (and vice versa).
3. **Director ends discussion prematurely** — when the human user asks a substantive question, the director closes the discussion (`已结束`) instead of routing it to the appropriate agent for a real answer.

## Example

Topic: axis of symmetry. User pushes back on whether the Tiananmen gate (a 3D building) can be called an axisymmetric figure.

```
奇奇:    哇,蝴蝶真的好漂亮!我想到了,咱们的天安门城楼是不是也是轴对称的呀?
林老师:  奇奇观察得真仔细!天安门城楼确实是轴对称的。... 谁来大胆猜猜看?
思思:    肯定可以
你(用户): 天安门城楼是建筑物,为什么也是轴对称图形呢?蝴蝶和剪纸我觉得是对折能够
         完全重合是对称轴对称图形,那那个城楼对折能对折吗?城楼它是三D的怎么办?
思思:    我觉得肯定能重合!只要找到正中间那条对称轴,两边的点就像照镜子一样一对一地对上呢。
林老师:  思思说得真妙,"照镜子"抓住了点对点的本质!虽然建筑是立体的,但在数学里,
         我们观察的是它的正面轮廓。... 严丝合缝地重合啦!
你(用户): 我们当然可以说天安门城楼的照片是轴对称图形,但我们不能说天安门城楼这个物体
         是一个轴对称的吧。
已结束
```

Issues visible in this trace:

- **Off-topic**: User's first question targets the 3D-vs-2D distinction. 思思 ignores it and just reasserts symmetry. 林老师's reply pivots to "照镜子" without addressing the dimensionality concern raised by the user.
- **Role confusion**: 林老师 attributes the "照镜子" insight to 思思, but 思思 only said "肯定能重合" — the framing came from the user's challenge. Teacher conflates whose contribution is whose.
- **Premature end**: User's second message is a direct correction/objection ("不能说天安门城楼这个物体是一个轴对称的"). Director responds with `已结束` instead of routing to teacher or peer agents for engagement.

## Expected behavior

- Agents must address the actual content of the prior turn, not generic topic continuation.
- Director/teacher must distinguish user turns from student-agent turns, and never re-attribute user statements to student agents.
- User questions — especially substantive challenges or objections — should trigger continued discussion, not termination. Termination should be gated on resolution, not on user input arrival.

## Suggested investigation

- Review director prompt for end-of-discussion trigger conditions; check whether user turn is being misclassified as a closing signal.
- Review teacher/agent prompts for speaker-attribution rules; ensure user is always tagged distinctly from student agents in the dialogue history fed back to the LLM.
- Check whether agent turns are conditioned on the *latest* user/agent message specifically, not just the rolling topic.

ashutoshrana:
## Summary

Fixes #511

Three reported symptoms — off-topic replies, role confusion, and premature END — all trace to one root cause: `summarizeConversation()` collapsed every `role: 'user'` message to the same generic label, stripping the `[AgentName]:` prefix that `convertMessagesToOpenAI()` adds when re-encoding peer agent turns. The director LLM had no way to distinguish a substantive human challenge from a brief agent acknowledgment, so it would emit `END` on unresolved student questions.

## Root Cause

\`\`\`
Teacher:   "The Tiananmen gate is axisymmetric..."
Xiao Ming: "[Xiao Ming]: Yes, symmetric from the front!"   ← re-encoded as role:'user'
Student:   "Wait — the gate is 3D. Can 3D objects really be axisymmetric?"
\`\`\`

Before this fix, the director summary showed both as identical `[User]` lines. The director read the second as a follow-up that appeared resolved and emitted `END`, leaving the student's real challenge unanswered.

## Changes

**`lib/orchestration/summarizers/conversation-summary.ts`**
- Added `AGENT_PREFIX_RE` to detect the `[AgentName]:` prefix on `role:'user'` messages
- `summarizeConversation()` now labels peer agent turns `[Agent: Name]` and genuine human messages `[Student (Human)]` — the director can distinguish them
- Added `extractLastHumanMessage()` to retrieve the last unaddressed human turn (skips agent-prefixed messages)

**`lib/orchestration/director-prompt.ts`**
- Added `buildOpenStudentQuestionSection()` — surfaces the most recent unresolved human question explicitly in the director's system prompt with a clear instruction not to emit `END` while it remains unanswered
- Added optional `openAIMessages` parameter (10th, backward-compatible)

**`lib/orchestration/director-graph.ts`**
- Passes the already-computed `openaiMessages` variable to `buildDirectorPrompt()` — no extra computation, one-line change

**`lib/prompts/templates/director/system.md`**
- Added `{{openStudentQuestionSection}}` placeholder so the new section is actually injected into the director prompt. Without this, `interpolateVariables()` silently drops any variable with no matching placeholder — the fix would be a no-op.

## Tests

- `tests/orchestration/conversation-summary.test.ts` — 38 new tests covering role label correctness, the exact #511 scenario replay, content truncation, `maxMessages` slicing, and all `extractLastHumanMessage` cases
- `tests/prompts/templates.test.ts` — 3 regression tests confirming `{{openStudentQuestionSection}}` is injected when a human message is present and absent otherwise

**331 tests passing, zero lint errors, zero new TypeScript errors.**

> Note: `lib/export/latex-to-omml.ts` has a pre-existing `TS2307: Cannot find module 'mathml2omml'` error present on `main` before this branch — not introduced here.

## Scope

This fix is limited to the director summary and prompt layer. No changes to graph topology, LangGraph state, `convertMessagesToOpenAI()`, or any agent-facing logic.


cosarah:
Thanks for working on this fix and for adding tests around the attribution issue.

 I do see a problem with the current approach, though: it relies on regex-matching the  rendered message text ([Name]: ...) to infer whether a message came from a human or  from another agent. In the real runtime path, that is not a reliable source of truth.
 For example, actual user messages can already be prefixed during conversion, so this  logic can misclassify a real human message as an agent message, which means Open  Student Question may never be populated when it actually should be.

 I think the fix should use the real message role/source from structured metadata  (originalRole, agentId, etc.) instead of parsing display text. In other words, the  director should make this decision based on the actual speaker identity, not on a
 regex over the formatted content.

ashutoshrana:
Hi @cosarah — you're right, and the issue is more fundamental than just occasional misclassification.

After tracing the full data flow: `use-chat-sessions.ts` sets `senderName: t('common.you')` on **every** human message. `message-converter.ts` then applies a `[senderName]:` prefix to all `role:'user'` content unconditionally (line 92–93). So in the director path, **all** `role:'user'` messages carry a `[Name]:` prefix — `AGENT_PREFIX_RE` matched them all as agent turns, meaning `extractLastHumanMessage()` always returned `null` and the Open Student Question section never fired.

The fix uses metadata as you suggested:

**`extractLastHumanMessage()`** now takes `StatelessChatRequest['messages']` (pre-conversion) and uses `msg.metadata?.originalRole === 'user'` as the discriminator — set by `use-chat-sessions.ts` on every genuine human message. Text is extracted from `msg.parts` directly, no content parsing.

**`summarizeConversation()`** uses `msg.role` as the discriminator. In the director path (`convertMessagesToOpenAI` called without `currentAgentId`), peer agent messages stay as `role:'assistant'` — they are never re-encoded as `role:'user'`. So `role` is reliable here. The `[senderName]:` prefix is stripped from user content cosmetically for readability, not used for logic.

**`buildDirectorPrompt()`** 10th param changed from `OpenAIMessage[]` to `StatelessChatRequest['messages']`; `director-graph.ts` passes `state.messages` (already in scope at line 164) instead of the converted `openaiMessages`.

Happy to push an updated commit if this direction looks right — or adjust the approach if you'd prefer a different pattern.

cosarah:
This direction looks right to me. Keeping sender prefixes in the conversation context is still useful for attribution, but the human vs. agent distinction should come from the real structured role/source metadata rather than regex-matching formatted text. Using the pre-conversion message metadata for Open Student Question and limiting prefix handling to presentation/readability feels like the correct fix.

ashutoshrana:
Updated commit addresses the feedback — the revised implementation is on the branch.

Quick summary of what changed:
- `extractLastHumanMessage()` now takes pre-conversion `StatelessChatRequest['messages']` and uses `msg.metadata.originalRole === 'user'` as the discriminator. No content parsing.
- `summarizeConversation()` uses `msg.role` for human/agent discrimination (reliable in the director path where no peer-agent re-encoding occurs). The `[senderName]:` prefix is stripped from summary output for readability only.
- `buildDirectorPrompt()` 10th param updated to `StatelessChatRequest['messages']`; `director-graph.ts` passes `state.messages` directly.

Happy to make any further adjustments.

cosarah (dismissed):
LGTM. Root cause analysis is accurate — `role:'user'` collision in director summary cleanly resolved by switching the discriminator from regex prefix matching to `msg.role` + `metadata.originalRole`. `extractLastHumanMessage` correctly uses the source-of-truth field set in `use-chat-sessions.ts`, and the `{{openStudentQuestionSection}}` placeholder injection in `system.md` is required (silent-drop on missing placeholder is a real footgun — good catch).

Test coverage is strong: 47 new tests including the exact #511 replay, edge cases (multi-part text, blank content, agent-only conversation), and prompt-injection regression. Full suite 347 passing, tsc clean.

Verified by running the dev server. Behavior matches expectation: directors no longer prematurely emit `END` on unresolved student challenges.

Minor non-blocking note: `extractLastHumanMessage` always surfaces the most recent human turn even after it has been addressed, so the director relies on its own judgment ("NOT been fully addressed") to know when to allow `END`. Worth watching in extended sessions, but the prompt wording handles it for now.

wyuc:
Thanks @ashutoshrana — the labeling fix and the move from regex to `msg.role` + `metadata.originalRole` is the right shape. Two things I'd like to discuss before merge:

**1. Could we move the open-question logic into the prompt instead?** @cosarah's non-blocking note already hints at this — `extractLastHumanMessage()` always surfaces the latest human turn whether it's been addressed or not, which means director ends up judging "fully addressed?" on its own anyway. If we're trusting director's judgment regardless, the procedural extraction mainly adds a layer to maintain.

What I'd lean toward is folding the same intent into `director/system.md` as declarative rules:

- context encoding (what `[Agent: Name]` vs `[Student (Human)]` mean, that they interleave)
- END conditions (only after a `[Student (Human)]` substantive question gets a *targeted* peer response — not a generic ack)
- attribution rule (teacher must not re-attribute a `[Student (Human)]` phrasing to an `[Agent: Name]`)

If that lands cleanly, `extractLastHumanMessage` / `buildOpenStudentQuestionSection` / the placeholder / the `director-graph.ts` wiring can all go away, and the fix becomes purely prompt-layer. Happy to hear pushback if there's a reason the procedural form is preferable.

**2. End-to-end regression for #511.** The 47 unit tests verify director's *inputs*, not its *behavior*. With the move toward "let director judge from rules," behavior verification matters more. We have eval infra (`tests/eval/`) — one fixture per symptom would be ideal: off-topic (3D Tiananmen), mis-attribution ("照镜子"), premature END (substantive objection). Each should **fail on `main`, pass with the fix** — otherwise it doesn't really prove the fix solved the symptom.

Out of scope for this PR — happy to file a separate issue and let you or @cosarah pick it up, whoever has bandwidth.

ashutoshrana:
Thanks @wyuc — both points are reasonable and I want to engage with them properly.

**On point 1 (prompt-layer vs. procedural):**

I think you're right that the procedural approach adds a layer without changing the fundamental reliance on the director's judgment. If the director is ultimately deciding "fully addressed?", surfacing the last human message via code doesn't remove that dependency — it just adds indirection.

Worth separating what can go away from what should stay:

- `extractLastHumanMessage`, `buildOpenStudentQuestionSection`, the `{{openStudentQuestionSection}}` placeholder, and the `director-graph.ts` wiring — all can go away under the prompt-layer approach.
- The `summarizeConversation()` changes (role-based discrimination, stripping the `[senderName]:` prefix from summary output) should stay regardless — the director prompt consumes this summary and the cleaner labels (`Student (Human)` vs. `Agent`) are still useful context even without the extracted question section.

For `director/system.md`, the declarative rules would cover three things:

1. **Context encoding** — clarify that `[Student (Human)]` and `[Agent: Name]` labels identify the actual speaker type, and that the director should use the summary labels, not infer from raw message structure.
2. **END condition** — only emit `END` after a `[Student (Human)]` substantive question has received a direct, targeted response from a named agent. A generic acknowledgment does not satisfy this.
3. **Attribution rule** — do not re-attribute a `[Student (Human)]` challenge to an `[Agent: Name]` when routing or summarizing resolution status.

Happy to push a revised commit with this shape — or share the draft rules first if you'd prefer to review before I push.

**On point 2 (eval fixtures):**

Should I include these in this PR, or would you prefer to track them as a separate follow-up issue?

wyuc:
Thanks @ashutoshrana —

1. **Push directly**, no need to share the draft first — I'll review on the diff.
2. **Keep the eval fixtures in this PR.** Once the fix is purely prompt-layer the unit-test surface mostly disappears, so the e2e fixtures become the real regression guard and belong with the change. These go under **`/eval/`** (a new domain dir following the `outline-language` pattern — `runner.ts` + `judge.ts` + `scenarios/*.json` + an `eval:` npm script), not `tests/eval/` (that's just vitest for the eval framework itself). If the harness proves fiddly, the deterministic premature-END scenario (assert director output ≠ `END`) is the must-have here; off-topic / mis-attribution can land as a skeleton + tracked follow-up if needed.

wyuc:
Direction is right — the prompt-layer rules + role-based labels are the correct shape. Two things before merge:

**1. CI is red, and not just prettier.** `pnpm check` fails on formatting first, which masks a real break: the refactor dropped `buildOpenStudentQuestionSection` / the 10th `buildDirectorPrompt` param, but `tests/prompts/templates.test.ts` still has 3 `open student question section` tests (~lines 219–293) hitting the removed API — typecheck + assertions will fail once prettier passes. Please `prettier --write` the two files **and** delete those 3 stale tests.

**2. The eval doesn't discriminate yet.** I ran `eval:orchestration` locally with a before/after on `system.md`:

| | author's fixture | a fixture matching the real #511 trace (declarative objection after the discussion appears to wrap up) |
|---|---|---|
| main `system.md` (no rules #10–12) | PASS | PASS |
| PR `system.md` (with #10–12) | PASS | PASS |

All four pass — the fixture passes *before* the fix too, so it can't prove the rules do anything. The premature-END symptom looks model-dependent/probabilistic — it didn't reproduce on `gemini-3-flash-preview` at temp 0, which a single deterministic run won't catch.

Suggest pulling the eval (and the `eval:orchestration` script) out of this PR so #554 lands as a clean prompt-layer fix, and tracking it as a follow-up issue — framed as "verify #511 is actually fixed" (reproduce the bug first), not "add fixtures". Happy to file it.

`conversation-summary.ts` and its unit tests are solid — keep those.

ashutoshrana:
Updated commit addresses both blockers from the 2026-05-22 review.

**What changed (8 files, −388 lines):**

- Removed `eval/orchestration/` (judge, runner, reporter, types, scenario) and the `eval:orchestration` npm script — the fixture passed on `main` before the fix, so it doesn't discriminate. Tracking as a follow-up per your suggestion.
- Deleted 3 stale tests from `tests/prompts/templates.test.ts` (`open student question section` × 3) — these called `buildDirectorPrompt` with a 10th param that no longer exists. They were masked by the prettier failure; once formatting passed they would have broken tsc.
- Fixed prettier formatting in `tests/orchestration/conversation-summary.test.ts` and `tests/prompts/templates.test.ts`.

**Validation (local, before push):**
- `pnpm run check` — clean
- `pnpm run test` — 333 passed, 0 failures
- `pnpm run lint` — 0 errors
- `tsc --noEmit` — clean

`conversation-summary.ts`, its 38 unit tests, and `director/system.md` rules 10–12 are unchanged.

wyuc (approved):
LGTM — both blockers from the 2026-05-22 review addressed, CI green. Thanks for the patient iteration @ashutoshrana. Tracking the deferred items (agent-side prompt clarification, e2e regression eval, a few small refinements) in a follow-up issue: #598.

PR #557, opened 2026-05-11
fix: preserve discussion triggers in imported classroom zips

### LINKED ISSUE #556: Imported .maic.zip classrooms silently skip discussion actions because discussion.agentId is not remapped
## Summary
Classrooms imported from `.maic.zip` can fail to trigger scene-level `discussion` actions. The imported scene still carries the original course's `discussion.agentId`, but imported agents are recreated with new IDs. When playback reaches that action, the engine treats the agent as unselected/invalid and skips the discussion silently.

## Steps to Reproduce
1. Create or export a classroom that contains at least one `discussion` action.
2. Export it as `.maic.zip`.
3. Import that zip as a new classroom.
4. Play the scene containing the `discussion` action.

## Actual
The discussion does not trigger. No proactive discussion card appears, and playback continues as if the discussion had already been consumed.

## Expected
The imported classroom should trigger the same `discussion` action as the original classroom.

## Root Cause
- Export only normalizes `speech.audioId` for portability.
- `discussion.agentId` is serialized as a raw runtime ID.
- Import recreates agents with brand-new IDs.
- Import remaps `multiAgent.agentIndices`, but does not remap `discussion.agentId`.
- Playback skips `discussion` when its `agentId` is not in the currently selected agent list.

## Suggested Fix
1. Make action-level agent references portable in the zip format.
2. Remap imported `discussion` initiators to the newly created agent IDs during import.
3. Add a backward-compatible fallback for existing ZIPs:
   - If a legacy `discussion.agentId` is invalid after import, assign a valid student/non-teacher agent instead of leaving it stale.
4. Add regression tests for discussion round-trip.

## Test Gap
Current classroom zip tests cover audio reference rewriting, but not `discussion` agent reference round-tripping.

cosarah:
## Summary
- store portable discussion initiators in classroom zip manifests using `agentIndex` when possible
- remap imported discussion initiators to newly created agent IDs during `.maic.zip` import
- preserve legacy/default-agent discussion initiators for preset classrooms and add regression tests

## Testing
- frontend manual verification
- `/home/yangshen/maic/OpenMAIC/node_modules/.bin/vitest run tests/export/classroom-zip.test.ts`
- pre-commit checks: `prettier`, `eslint --fix`, `tsc --noEmit`

Closes #556


cosarah:
Self-review:

- I checked both import paths this change affects: generated-agent classrooms and preset/default-agent classrooms.
- During review I found one compatibility edge case and fixed it in a follow-up commit: preset classrooms could lose a legacy `discussion.agentId` on export when no manifest agent index was available.
- I verified the final diff preserves three behaviors:
  - new exports use portable `agentIndex` for discussion initiators when an exported agent mapping exists
  - imports remap discussion initiators onto the newly created generated-agent IDs
  - legacy/default-agent initiators are preserved when the classroom does not carry generated agents
- Regression coverage was expanded around both round-trip and legacy fallback behavior.
- I do not have further findings in the final PR diff.

Residual risk:
- coverage here is at the manifest rewrite/unit-test level; there is still no end-to-end browser test that exports, re-imports, and plays through a discussion trigger.

wyuc (commented):
E2E verified locally for PR #557.

Flow tested:
1. Created a classroom with one slide page containing a generated-agent `discussion` action after the lecture speech.
2. Exported it as `.maic.zip`.
3. Imported that ZIP into a fresh browser instance/context.
4. Played the imported page from the canvas and waited for the page playback to finish, not just a fixed delay after clicking Play.
5. Confirmed the discussion proactive card appeared, clicked Join, and verified the discussion response rendered in both the roundtable and Chat panel.

Checks observed:
- Exported `manifest.json` stores the discussion trigger as `agentIndex`.
- Import rewrites it to the new generated agent id in the target classroom.
- Joining the discussion sends `/api/chat` with `sessionType: "discussion"` and `triggerAgentId` equal to the imported/generated agent id.
- The discussion content played successfully after import.

Screenshots:

Discussion card after page playback finished

Discussion speaking after Join

Also ran:
- `pnpm test tests/export/classroom-zip.test.ts`

Looks good from the import/export + playback path I tested.

wyuc (approved):

PR #198, opened 2026-03-21
feat: add pre-generation outline review

nkmohit:
**Summary:**
Adds an optional "Review Outline" step between outline generation and content generation. Off by default so the existing flow is unchanged for casual users. When enabled, generation pauses after the outline is drafted and shows a review page where users can edit scene titles, descriptions, key points, reorder scenes, change scene types, add new scenes, or remove unwanted ones before confirming and starting content generation.

**Related Issues:**
Related to #173. Context from discussion in #150 where wyuc noted a mandatory confirmation step added friction this addresses that by making the review step opt-in via a toolbar toggle.

**Changes:**
- Added "Outline" tab to the generation toolbar popover with a "Review Outline" toggle (off by default)
- Added outline review page that renders `OutlinesEditor` (previously unused component) after outline streaming completes
- Generation pauses on outline step when toggle is enabled and only proceeds on user confirmation
- Added i18n translations for English and Chinese
- Added persisted setting for the toggle state

**Type of Change:**
- [x] New feature (non-breaking change that adds functionality)

**Steps to reproduce / test:**
1. Enable "Review Outline" in the generation toolbar (Image/Video/TTS/ASR/Outline tabs)
2. Enter a topic and click "Enter Classroom"
3. Watch outline generate — generation pauses and shows the review page
4. Edit, reorder, or remove scenes as needed
5. Click "Confirm and generate course" to proceed
6. Verify content generation uses the edited outlines
7. Disable the toggle and confirm generation proceeds automatically as before

**What you personally verified:**
- Toggle is off by default, existing generation flow unchanged
- Review page renders correctly with all outline items editable
- Scene type dropdown, reorder controls, delete, and add scene all work
- "Back to requirements" returns to the input page
- "Confirm and generate course" proceeds with edited outlines
- i18n works in both English and Chinese
- `pnpm exec tsc --noEmit` passes

**Evidence:**
- [x] CI passes (pnpm check && pnpm lint && npx tsc --noEmit)
- [x] Manually tested locally
- [x] Screenshots attached

screenshots: 
Screenshot 2026-03-21 at 3 20 01 PM
Screenshot 2026-03-21 at 3 21 20 PM
Screenshot 2026-03-21 at 3 22 15 PM
Screenshot 2026-03-21 at 3 22 36 PM
Screenshot 2026-03-21 at 3 22 52 PM

**Checklist:**
- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [x] My changes do not introduce new warnings

Issue #173, opened 2026-03-20
[Feature]: Add outline review and refinement step before content generation


nkmohit:
### Problem or Motivation

After outline generation, the app waits 800ms and then immediately starts scene content generation by calling `/api/generate/scene-content `with no user confirmation step. Users currently can’t remove unwanted scenes, reorder items, or adjust titles before the expensive generation runs.

This is also different from #68 (post-generation editing) because this feature enables controlled pre-generation, saving unnecessary calls for scenes the user doesn’t want.

### Proposed Solution

Two-phase approach:

Short term (UI wiring):

- Wire up the existing OutlinesEditor component (`components/generation/outlines-editor.tsx`) into the outline->content flow in `app/generation-preview/page.tsx`.
- After `/api/generate/scene-outlines-stream` finishes streaming, pause on the outline step and render OutlinesEditor.
- Only proceed to content generation after the user clicks a confirm action, using the (possibly edited) outlines.

Long term (conversation loop):

- Add a conversational refinement loop so users can provide natural-language feedback (e.g., “remove the quiz scenes”, “add a section on positional encoding”, “make it more beginner friendly”).
- The AI regenerates/refines the outline based on the original requirement + prior outline + user feedback.
- Repeat until the user confirms, then start scene content generation.

### Alternatives Considered

- Do post-generation editing only (but is less efficient because content generation already happened).
- Full regeneration from scratch (works but slower and more costly than targeted outline refinement).

### Area

Classroom generation

### Additional Context

- There is an OutlinesEditor component in `components/generation/outlines-editor.tsx` that appears to be planned for this use case but is currently unused.
- This change would give curators control before expensive generation and supports both manual editing now and AI-assisted iteration later.

wyuc:
The short-term UI wiring described here shipped in #558 (`a5209d7`): `OutlinesEditor` is now wired into `app/generation-preview/page.tsx` so generation pauses after outline streaming completes and the user can review/edit before content generation runs.

The long-term conversational refinement loop (natural-language outline edits) is not addressed yet — happy to track that under a follow-up issue if you'd like to spec it out further.

Thanks @nkmohit for the original proposal and for #198, which informed the placement design.

PR #558, opened 2026-05-11
feat(outline-review): clickable streaming card morphs into editor

wyuc:
## Summary

Replaces the placement, interaction, and visual design of the pre-generation outline review surface. The previous toolbar-toggle approach (#198) is dropped: the streaming preview card on `/generation-preview` is itself the entry into the editor, and morphs into it via a shared `layoutId` while SSE keeps streaming uninterrupted.

Implements #173 (short-term UI wiring; conversational refinement loop is still future work).

## Related issues / PRs

- Implements #173 — *Add outline review and refinement step before content generation*
- Supersedes #198 — *feat: add pre-generation outline review*. That PR added a "Review Outline" toggle inside the home toolbar's media popover; we revisited the placement and chose not to add another always-visible toolbar control because the home toolbar is already dense (model / PDF / web search / media / interactive / voice / submit). This PR keeps the default fast path and exposes review on demand from the streaming preview card itself.

If maintainers prefer to land #198's approach instead, this PR can be closed without merging — they cover the same product surface with different placement strategies.

## Demo flow

1. Submit any requirement on `/`.
2. Generation streams the outline into the small preview card (now tilted + with a "点击审阅" hint).
3. Click the card mid-stream to expand into the full editor; SSE keeps writing scenes into the editor in real time (read-only while streaming).
4. After streaming, edit titles / descriptions / key-points / type / quiz config; reorder by drag; insert with hover-revealed `+` between scenes; delete with in-place confirm popover.
5. Confirm to proceed, or collapse to fall back to the small card with a 2.5s auto-continue.

## What changed

**Streaming → editor morph** (`app/generation-preview/components/visualizers.tsx`, `app/generation-preview/page.tsx`)
- Streaming card is now an interactive `motion.div` (tilt + hover-straighten + always-visible expand affordance).
- `previewPhase: 'review'` can be entered mid-stream via `outlineReviewIntentRef`; SSE-completion ORs in this intent to decide whether to wait on confirmation.
- Editor's source of truth is `session.sceneOutlines ?? streamingOutlines`, so collapsed and expanded views stay in sync after edits.
- Header collapse button shrinks back to the preview card. Mid-stream collapse keeps SSE running; post-stream collapse re-arms the no-review path's 2.5s auto-continue timer.
- Auto-start condition broadened so a refresh during early-review reattaches to a fresh SSE rather than stalling.

**Editor redesign** (`components/generation/outlines-editor.tsx`)
- Document-style layout (Notion-feeling): drag handle + number share a flex left rail aligned to the title baseline; title/description are inline-editable borderless textareas with auto-resize.
- ⋮ overflow menu and explicit move-up/down removed; reordering is drag-only, with hover-revealed delete (popover-confirmed) next to the type pill.
- `+` insert dividers between scenes (hover-revealed); faint `+` at list edges replaces the "Add scene" bar so insertion is uniform.
- Quiz config popover redesigned: stepper for question count, segmented control for difficulty, multi-select pills for question types — fixes a pre-existing UI bug where the single-select silently truncated the LLM's `questionTypes` array to one element.
- Streaming placeholder row pulses below the latest scene during SSE.

**Draft cache fix** (`lib/hooks/use-draft-cache.ts`, `app/page.tsx`)
- Surfaced while testing "Back to requirements". `useDraftCache` initialised state via `useState`'s lazy initialiser, but under React 18+ SSR the initialiser only runs server-side (returns `undefined` because `window` is missing) and React reuses that hydrated state on the client without re-running it — so `cachedValue` was permanently `undefined` and nothing ever rehydrated. The home page then masked the symptom with a derived-state pattern that initialised `prev` from the same always-undefined value, making the comparison never fire.
- Switched the hook to an effect-based read so consumers receive a real state update after mount, and replaced the home page's derived-state restore with a `useEffect` + `useRef` guard.
- Side benefit: the PBL chat draft cache and quiz-answer cache start working (same broken pattern, same fix).
- Moved `localStorage.removeItem('requirementDraft')` in `generation-preview/page.tsx` to after the review promise resolves, so navigating back from the editor preserves the draft.

**i18n** (`lib/i18n/locales/{en-US,zh-CN,zh-TW,ja-JP,ru-RU,ar-SA}.json`)
- 14 new keys per locale: expand hint, streaming progress / waiting / waiting-confirm, collapse, delete confirmation copy, insert hint, quiz config summary, etc.
- Locale coverage test (`tests/i18n/outline-review-locales.test.ts`) extended to assert the new keys plus three `{{count}}` interpolation contracts.

## Removed from #198's scope

- The "Outline" tab and "Review Outline" toggle inside the home toolbar's media popover — explicitly out of scope per the placement design (no new always-visible toolbar control).
- The redundant "Review outline" button + auto-continue caption that lived under the preview card — the card itself is now the entry point.

## Test plan

- [x] `pnpm exec tsc --noEmit`
- [x] `pnpm vitest run tests/i18n/outline-review-locales.test.ts tests/store/settings-server-sync.test.ts` — 46/46 pass
- [ ] E2E: existing `e2e/tests/generation-flow.spec.ts` cases (review-from-preview, persist always-review, auto-open-when-always-review, persisted-review-session-resume) — locators are role-based and should be unaffected by the redesign, but worth running in CI.
- [ ] Manual flows:
  - [ ] Click streaming card mid-stream → editor expands, scenes keep arriving live
  - [ ] Collapse mid-stream → small card reappears, SSE keeps going
  - [ ] Collapse post-stream → small card + 2.5s auto-continue
  - [ ] "Always review outlines" toggle persists; auto-opens editor on next generation
  - [ ] Drag reorder, hover insert `+`, delete with popover confirm
  - [ ] Quiz scene with mixed `questionTypes` from LLM round-trips through editor
  - [ ] "Back to requirements" → home page restores the original input
  - [ ] Refresh during early-review → SSE restarts and editor refills

## Status

Draft for review. Particularly looking for opinions on:
1. **Placement-vs-#198 trade-off** — confirm we want to drop the toolbar toggle approach?
2. **Collapse-then-auto-continue UX** — currently 2.5s with no progress bar (we discussed and felt a countdown bar reads as CTA); fine to ship as-is?
3. **Quiz `questionTypes` fix** — pre-existing bug, OK to fold into this PR or split out?


wyuc:
Pushed `7dddcc4` addressing the code review. Summary of changes:

**Critical**
- ✅ E2E selectors (Critical 1) — `e2e/pages/generation-preview.page.ts` now matches the streaming card's `outlineExpandHint` aria-label (Tap to review / 点击审阅 / クリックでレビュー / Нажмите, чтобы проверить / اضغط للمراجعة) instead of the removed "Review outline" button.

**Important**
- ✅ RU/AR plurals (Important 2) — added CLDR plural variants. EN gets `_one` for the three count-bearing strings; RU adds `_one` and `_few` for `quizConfigSummary`; AR adds `_one` and `_two` for `outlineEditorSummary`, `outlineEditorStreamingProgress`, `quizConfigSummary`. Bare key remains as `_other` fallback. zh / zh-TW / ja have a single plural form so no change.
- ✅ Promise leak (Important 3) — `waitForOutlineReviewChoice` now accepts an `AbortSignal` and rejects on abort. `goBackToHome` and unmount cleanup already abort the controller, so the parked promise settles instead of stranding the awaiting `startGeneration` closure. The existing `AbortError` catch swallows it silently.
- ✅ Mid-stream collapse (Important 4) — added inline comment clarifying intent.
- ✅ Keyboard reorder (Important 5) — drag handle responds to Ctrl/⌘ + ↑/↓ when focused; copy switched from `dragScene` to a new `dragSceneHint` that surfaces the shortcut; `aria-keyshortcuts` declared.
- ✅ KeyPointInput blur (Important 6) — dropped `onBlur={onSubmit}`. Only Enter / comma commits. No more surprise chips when clicking away.

**Minor**
- ✅ Dead i18n keys (Minor 9) — removed `reviewOutlineAction`, `moveSceneUp`, `moveSceneDown`, `dragScene`, `sceneDescriptionLabel`, `keyPointsLabel`, `keyPointsPlaceholder`, `quizConfigLabel`, `addScene` from all 6 locales and from the coverage test.
- ✅ Layout thrash (Minor 11) — `useAutoResize` now `requestAnimationFrame`-batches its measure+write, with cancellation on cleanup.
- ✅ `isConfirmingOutlines` reset (Minor 14) — fallback path of `handleConfirmOutlines` now clears the flag before driving the transition itself.

**Deferred (not done in this commit)**
- The reviewer flagged a "removed `replaceThumbnails` / `revokeThumbnailSlideMediaUrls`" in `app/page.tsx`. Verified locally — that's a `origin/main..HEAD` two-dot diff artifact: my branch was rooted on `46b61de` and upstream commit `22c637c` (after my root) introduced those removals. Three-way merge will preserve them. Not a real removal in this PR.
- Reviewer also recommended rebasing the branch on `origin/main` so reviewers see the real diff. Holding on that pending maintainer preference — happy to rebase if you'd like the file count to reflect actual blast radius.

Verification:
- `pnpm exec tsc --noEmit` ✅
- `pnpm vitest run tests/i18n/outline-review-locales.test.ts tests/store/settings-server-sync.test.ts` → 46/46 ✅
- E2E not re-run locally (needs full dev server boot); selectors updated per the page-object change above.

---
*Edited to escape `#N` references that GitHub auto-linked to unrelated PRs/issues.*

wyuc:
Follow-up commits to get CI green after the rebase + bot review:

**`2880980` chore: prettier + drop redundant session check**
- Ran prettier on `outlines-editor.tsx` (Lint job was red).
- Removed `&& session` from the StepVisualizer `onExpandOutline` conditional — `session` is guaranteed non-null by the early-return guard above (flagged by `github-code-quality` bot's two inline comments at line 1299).

**`2fa5565` fix: silence react-hooks/set-state-in-effect on draft restore**
- ESLint's `react-hooks/set-state-in-effect` rule fired on the new `useDraftCache` effect read and the `app/page.tsx` restore effect. Added the same `eslint-disable` comment block already used for the file's other localStorage hydration effects — the `setState`-in-effect is intentional because lazy `useState` initialisers run on the SSR server with `window` undefined and React doesn't re-run them on hydration.

**`18019e4` revert: drop CLDR plural variants until key-alignment check supports them**
- `scripts/check-i18n-keys.mjs` enforces strict key equality across all locales, which is incompatible with i18next's per-locale plural-suffix keys (RU needs `_one`/`_few`/`_many`/`_other`, AR adds `_two`/`_zero`, while zh / ja need none). Any partial pluralization breaks alignment.
- Rolled back the `_one` / `_few` / `_two` variants I added in `7dddcc4`. The original "ungrammatical for n=1" wording from the reviewer's Important #2 stays for now — proper fix would teach `check-i18n-keys.mjs` to ignore CLDR plural suffixes when comparing key sets, then re-add the variants. Filing as follow-up.

**Also:** branch is now rebased on `origin/main` per the reviewer's recommendation, so the diff stat reflects actual blast radius (18 files, ~2k LOC) instead of the inflated two-dot view.

CI status:
- ✅ `Lint, Typecheck & Unit Tests` (1m26s)
- ✅ `Analyze (javascript-typescript)` (1m28s)
- ✅ `E2E Tests` (2m19s)

cosarah:
Tested locally — flow works as advertised: streaming card morph, mid-stream expand/collapse, drag reorder, hover `+` insert, popover delete, quiz config round-trip, "always review" persistence, and the draft-cache restore on "back to requirements".

Two small things worth considering before merging:

1. **Empty outline scenes pass through unchecked.** `insertOutlineAt` creates a scene with empty `title`/`description`/`keyPoints`, and `handleConfirmOutlines` only guards `length === 0`. The downstream LLM still produces something reasonable (it has the original requirement as context), but the resulting scene title is blank and the sidebar shows a perpetual loading skeleton for that page. Suggest filtering empty scenes on confirm, or disabling the confirm button (with a hint) when any scene has no title.

2. Minor follow-up only — even with (1) fixed, a fallback title in `scene-generator.ts` (e.g. `Scene N`) and ensuring empty-content failures move the outline to `failedOutlines` would harden the never-loading-state case.

LGTM otherwise — happy to merge once (1) is addressed.

cosarah (approved):
Approving — the empty-outline edge case can wait, since the outline schema is slated for refactor anyway. Nice work on the morph interaction and the draft-cache fix.

Issue #562, opened 2026-05-12
[MAIC Editor] Slide surface — first concrete static-display surface


wyuc:
Implements `SceneEditorSurface` for slides — Phase 1's first concrete surface, covering the full slide editing scope per #560. Reusable form components live alongside since they're imported into surface popovers.

Surface registers at module load but no chrome to render it yet — the user-visible part lands in the Pro mode chrome PR.

### Surface plumbing
- `components/edit/use-slide-edit-commands.ts` — orchestrator hook over `slide-ops` + `useHistorySnapshot` + `useCanvasStore`
- `components/edit/surfaces/slide/SlideCanvas.tsx` — wraps existing `slide-renderer/Editor/Canvas`
- `components/edit/surfaces/slide/SlideSurface.tsx` — assembles `SurfaceState`
- `components/edit/surfaces/slide/register.ts` — side-effect registers slideSurface

### Surface contributions
- `components/edit/surfaces/slide/insert-items.tsx` — Text / Shape (rectangle) / Image inserts
- `components/edit/surfaces/slide/commands.tsx` — global commands (ruler / grid)
- `components/edit/surfaces/slide/floating-actions.tsx` — element-type-aware floating popovers covering full Phase 1 capabilities:
  - **Text**: font picker, size, bold / italic / underline, color, alignment, bullet
  - **Shape**: fill color + opacity, outline controls, flip H/V
  - **Image**: replace (upload / URL), border radius, mask color, **crop trigger**, flip H/V
  - **Any**: geometry popover (x/y/w/h/rotate), z-order via overflow `⋯`
  - **Multi-select**: align + distribute popover

### Slide-level surface
- Slide background editor (solid color + image) — surfaced in command bar or dedicated popover; ergonomics decided in chrome PR's design pass

### Reusable form components
- `components/edit/surfaces/slide/inspector-controls.tsx` — TextInput, NumberInput, ColorInput, SwitchRow, FontSelect, OpacityField, FlipControls, OutlineControls, GeometryFields, AlignmentGrid (used inside popovers)
- `components/edit/surfaces/slide/ImagePicker.tsx` — file upload + URL paste; cleanly closes popover after pick
- `components/edit/surfaces/slide/ImageSourceField.tsx` — clean inspector field that hides base64 data URLs

### i18n
- 6-locale `edit.*` namespace: insert / actions / fields / sections / sidebar (all UI strings; no hardcoded text in components)

## Design notes

[...]

Issue #560, opened 2026-05-12
[MAIC Editor] Phase 1 — framework + Slide as first surface (RFC #547)


wyuc:
## What MAIC Editor is — and isn't

MAIC classrooms compose scenes of different kinds. This project builds
the editing framework for the **static display components** in those
classrooms — the parts that present non-interactive content to
learners. It is intentionally **not** a "slide editor": slide is one
form of static display, the most common one today, and the natural
starting point.

The framework treats each static display medium as a pluggable
surface implementing a shared contract (`SceneEditorSurface`).
Phase 1 ships the framework + the **slide surface**.

Beyond Phase 1, the same framework extends to:

- **Richer slide DSL elements** — early experiments with code blocks;
  charts / tables / video / formulas as natural follow-ups
- **New static display media** — HTML-based rendering is being
  explored as a parallel surface alongside slide
- **AI-native authoring (Phase 2)** — an AI editing assistant that
  proposes ops against the existing surface state, leveraging the
  same `SceneEditorSurface` contract (no contract change required)

> Interactive scenes (quiz / PBL / live discussion) are a **different
> paradigm** with separate editing concerns. They are **not**
> intended to plug into `SceneEditorSurface`; when they need editors,
> those live as a sibling framework that may share primitives
> (history kernel, i18n, Pro toggle) but not the chrome contract.

## Design principles

Load-bearing — every sub-PR's scope decision defers to these.

### 1. Round-trip is non-negotiable

Every editing operation in any static display surface must round-trip
cleanly through the corresponding export path (PPTX for slide; future
surfaces declare their own export targets). This isn't a fallback —
it's a hard gate on what we ship. **A property the export pipeline
can't represent cannot be edited.** Users who outgrow our editor
should always have an escape hatch into desktop tools.

This implementation realizes
[RFC #547](https://github.com/THU-MAIC/OpenMAIC/discussions/547).

### 2. AI-first; manual editing is the fine-tune layer

MAIC's product loop is: **AI generates → user lightly refines →
playback / export**. The editor is intentionally **not a destination
authoring tool** — its job is the minimum sufficient refinement on
AI output. Phase 2 brings an AI editing assistant for structural
changes (rewrite, reorganize, restyle); Phase 1 manual editing
exists to cover what AI can't precisely do (specific color, exact
alignment, pixel positions, factual fixes, replacing assets).

Practical consequence on Phase 1 scope: anything AI handles better
[...]

Issue #563, opened 2026-05-12
[MAIC Editor] Pro mode chrome — full-screen editor + Pro toggle entry + docs


wyuc:
The user-visible PR. After this lands in `feat/maic-editor-v0`, MAIC Editor's Phase 1 is live and ready for the integration PR to main. Includes documentation since they ship together.

### Pro mode chrome
- `components/edit/ProMode/*` — full-viewport editor (replaces playback chrome entirely when `mode === 'edit'`):
  - `ProModeEditor.tsx` — top-level layout: thin top bar + thumbnail rail + canvas + floating selection bar
  - `ProTopBar.tsx` — back-to-playback button + slide title + undo/redo + insert items + global commands. **No playback elements** (no "current scene" label, no download, no language switcher, no theme toggle).
  - `ProSidebar.tsx` — slide-only thumbnail rail, hover-revealed `⋯` menu inside each card (move up / move down / delete), "+ Add slide" footer; no quiz/interactive entries
  - `ProFloatingBar.tsx` — selection-anchored property popovers (renders contributions from SurfaceState)

### Wiring
- `components/header.tsx` — Pro toggle (real Switch + "Pro" label, violet active state) between header pill and download button
- `components/stage.tsx` — when `mode === 'edit'`, render `<ProModeEditor>` viewport-level instead of nesting playback panels
- `components/stage/scene-renderer.tsx` — route `mode === 'edit'` through registry to resolved surface

### Docs
- `CONTRIBUTING.md` — new section "Adding an editor surface for a new scene type": SceneEditorSurface contract overview, walkthrough using the slide surface as reference, registration pattern, DSL constraints (PPTX-representability for slide; per-type for other surfaces)
- `docs/editor-architecture.md` (new, optional) — diagram + writeup of chrome ↔ surface decoupling
- Update [RFC #547](https://github.com/THU-MAIC/OpenMAIC/discussions/547) Discussion: post implementation summary as a reply, link to merged PRs, note out-of-scope items deferred

## UX decisions (pre-designed)

A wireframe + per-region behavior spec lands in the PR description for review **before** code. Key decisions to be confirmed there:

- Pro mode is a **page-level takeover**, not a panel swap inside playback layout
- Sidebar action buttons live **inside** the thumbnail card via hover `⋯` menu — never as an external column
- No fixed right inspector panel; all properties via floating popovers
- Slide-only sidebar in Pro mode (other static-display surfaces will own their own sidebars when implemented)
[...]

PR #564, opened 2026-05-12
feat(maic-editor): framework primitives + edit StageMode

### LINKED ISSUE #561: [MAIC Editor] Framework — stage edit mode + slide-ops kernel + SceneEditorSurface contract
First sub-PR for #560. The full framework / engine layer in one PR. No UI consumers yet — surface and chrome land in subsequent PRs.

### Stage edit mode boundary
- `lib/types/stage.ts` — extend `StageMode` union: `'autonomous' | 'playback'` → `'autonomous' | 'playback' | 'edit'`
- `lib/store/stage.ts` — `setMode` clears canvas selection (`useCanvasStore.resetCanvasState()`) when leaving `edit`
- `components/stage.tsx` — `useEffect` auto-exits `edit` mode when current scene becomes uneditable (pending generation / no scenes / generating outlines / null current scene)

### slide-ops operation kernel
- `lib/edit/slide-ops.ts` — operation reducer covering: `element.add`, `element.update`, `element.updateMany`, `element.deleteMany`, `element.duplicate`, `element.reorder`, `element.align`, `element.removeProps`, `text.updateContent`, `slide.update`. Plus `applySlideEditOperation` dispatcher.
- `lib/edit/slide-edit-elements.ts` — factories: `createDefaultTextElement`, `createDefaultShapeElement`, `createDefaultImageElement`, `createDefaultSlide`, plus `plainTextToParagraphHtml` / `htmlToPlainText` helpers (incl. `ShapeSpec` for future extension)
- `tests/edit/slide-ops.test.ts` — full operation coverage
- `tests/edit/slide-edit-elements.test.ts` — factory contracts
- `tests/edit/stage-mode.test.ts` — mode transition behavior

### SceneEditorSurface contract + registry
- `lib/edit/scene-editor-surface.ts` — interface definitions:
  - `SceneEditorSurface<TContent, TSelection>` — what each scene type registers
  - `SurfaceState<TContent, TSelection>` — what the surface's hook returns to chrome
  - Contribution primitives: `InsertPaletteItem`, `FloatingAction`, `EditorCommand`, `EditorHint` (reserved); both `InsertPaletteItem` and `FloatingAction` support optional `popoverContent` for property surfaces
  - `UiAffordance` shared base, `SurfaceHistory`
- `lib/edit/scene-editor-registry.ts` — Map-based registry with `register` / `resolve`

### i18n placeholders
- `lib/i18n/locales/{zh-CN,en-US,ja-JP,zh-TW,ar-SA,ru-RU}.json` — placeholder keys: `stage.editCourse`, `stage.doneEditing`. Full `edit.*` namespace lands in the slide-surface PR alongside surface implementation.

## Operation design guardrail

Every operation in this PR must produce data round-trippable through the existing PPTX export path. Reviewers should sanity-check: does this op modify a field that PPTX format can represent?

## Acceptance

- [ ] `tsc --noEmit` clean
- [ ] All new tests pass; existing tests still pass
- [ ] No new lint errors
- [ ] `pnpm build` green
- [ ] Reviewer-confirmed: every operation maps to a PPTX-representable mutation
- [ ] No user-visible changes (mode never set to `'edit'` from any UI yet)

**Target branch**: `feat/maic-editor-v0`
**Depends on**: nothing
**Estimated diff**: ~1100 lines (mostly tests)

Tracking: #560

wyuc:
## Summary

First sub-PR of the MAIC Editor Phase 1 (RFC #547, tracking #560). Adds the framework primitives — types, store transitions, and the `SceneEditorSurface` contract — without shipping any UI consumers. The EditShell chrome and the slide surface registration come in follow-up sub-PRs.

**About positioning.** MAIC Editor is the editing framework for *static display components* in MAIC classrooms — the parts that present non-interactive content to learners. Slide is the first surface; other static-display media (e.g. HTML-based) will plug into the same contract in later phases.

## What ships

- **`StageMode`** gains `'edit'` alongside `'autonomous' | 'playback'`. `setMode` resets canvas selection when leaving `'edit'` so stale selection never bleeds into playback.
- **Auto-exit guard** in `<Stage>`: leaves `'edit'` whenever the current scene becomes uneditable (no scenes, pending generation, no current scene). Pre-wires the follow-up Pro toggle so it can never strand the user in an empty edit shell.
- **`SceneEditorSurface` contract** + a tiny in-memory registry under `lib/edit/`. The shell will resolve surfaces by `SceneType`; surfaces register themselves at module init. Surfaces declare `CanvasComponent`, `useSurfaceState()`, and contribute insert palette items, floating actions, commands, and (reserved for AI) inline coach hints.
- **Slide kernel** (`lib/edit/slide-ops.ts`): immutable, history-aware operations — `slide.update`, element `add` / `update` / `updateMany` / `delete` / `deleteMany` / `reorder` / `duplicate` / `align` / `removeProps`, plus `text.updateContent`. Single entry point so undo/redo and (future) PPTX-export round-tripping stay coherent.
- **Slide element factories** (`lib/edit/slide-edit-elements.ts`) for default text / shape / image elements and HTML ↔ plain-text helpers.
- **i18n**: `stage.editCourse` + `stage.doneEditing` across all 6 locales (en-US, zh-CN, zh-TW, ja-JP, ru-RU, ar-SA). Consumed by the header toggle in the next sub-PR.
- **Tests**: vitest coverage for the slide kernel (operations + history) and the edit-mode store transition (entry + canvas reset on exit).

## Explicitly out of scope

- **No UI entry into edit mode** — no Pro toggle in the header yet. The new `'edit'` mode is reachable only programmatically until the follow-up chrome PR lands.
- **No EditShell / EditModeSidebar** — chrome ships in a follow-up.
- **No slide surface registration** — the kernel is there but the slide surface that calls `register()` ships in a follow-up.

## Type widening

`PBLRenderer`'s `mode` prop was `'autonomous' | 'playback'`; widened to `StageMode` so `<SceneRenderer>` (which already passes `StageMode`) type-checks. The prop is unused inside the renderer, so behavior is unchanged.

## CI

`ci.yml` now also runs on PRs targeting `feat/maic-editor-v0`. The MAIC Editor lands as a series of stacked sub-PRs against that long-lived branch, and without this entry none of them get a style/lint/type/test gate — regressions would only surface when `feat/maic-editor-v0` finally merges back to `main`, which is a much worse place to catch them. Push trigger stays `main`-only since nobody pushes directly to the long-lived branch.

Note: GitHub Actions reads workflow config from the PR's **base** branch, so this change only starts gating once #564 merges into `feat/maic-editor-v0` — the next sub-PR (#562) will be the first to actually run CI.

## Test plan

- [x] `pnpm test` → 335 passed (was 317, +18 from new files)
- [x] `pnpm exec tsc --noEmit` → clean
- [x] `pnpm lint` → 0 errors (7 pre-existing warnings, none introduced)
- [x] `pnpm check:i18n-keys` → 6 locales aligned
- [x] `pnpm build` → compiled successfully

Tracking issue: #560.
Closes #561.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
+    branches: [main, feat/maic-editor-v0]
 
 concurrency:
   group: ci-${{ github.ref }}
--- a/components/scene-renderers/pbl-renderer.tsx
+++ b/components/scene-renderers/pbl-renderer.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import { useCallback } from 'react';
-import type { PBLContent } from '@/lib/types/stage';
+import type { PBLContent, StageMode } from '@/lib/types/stage';
 import type { PBLProjectConfig } from '@/lib/pbl/types';
 import { useStageStore } from '@/lib/store/stage';
 import { PBLRoleSelection } from './pbl/role-selection';
@@ -10,7 +10,7 @@ import { useI18n } from '@/lib/hooks/use-i18n';
 
 interface PBLRendererProps {
   readonly content: PBLContent;
-  readonly mode: 'autonomous' | 'playback';
+  readonly mode: StageMode;
   readonly sceneId: string;
 }
 
--- a/components/stage.tsx
+++ b/components/stage.tsx
@@ -49,6 +49,7 @@ export function Stage({
   const { t } = useI18n();
   const {
     mode,
+    setMode,
     getCurrentScene,
     scenes,
     currentSceneId,
@@ -259,6 +260,19 @@ export function Stage({
     await chatAreaRef.current?.endActiveSession();
     doSessionCleanup();
   }, [doSessionCleanup]);
+
+  // Auto-exit edit mode whenever the current scene becomes uneditable
+  // (pending generation, no scenes, currently generating). Without this guard,
+  // a follow-up sub-PR's Pro toggle could strand the user in an empty edit shell.
+  useEffect(() => {
+    if (mode !== 'edit') return;
+    const isPending = currentSceneId === PENDING_SCENE_ID;
+    const stillEditable =
+      !isPending && scenes.length > 0 && generatingOutlines.length === 0 && !!currentScene;
+    if (!stillEditable) {
+      setMode('playback');
+    }
+  }, [mode, currentSceneId, scenes.length, generatingOutlines.length, currentScene, setMode]);
 
   const clearPresentationIdleTimer = useCallback(() => {
     if (presentationIdleTimerRef.current) {
--- a/components/stage.tsx
+++ b/components/stage.tsx
@@ -3,6 +3,7 @@
 import { useState, useEffect, useRef, useMemo, useCallback } from 'react';
 import { useStageStore } from '@/lib/store';
 import { PENDING_SCENE_ID } from '@/lib/store/stage';
+import { isCurrentSceneEditable } from '@/lib/edit/stage-mode';
 import { useCanvasStore } from '@/lib/store/canvas';
 import { useSettingsStore } from '@/lib/store/settings';
 import { useI18n } from '@/lib/hooks/use-i18n';
@@ -262,14 +263,17 @@ export function Stage({
   }, [doSessionCleanup]);
 
   // Auto-exit edit mode whenever the current scene becomes uneditable
-  // (pending generation, no scenes, currently generating). Without this guard,
-  // a follow-up sub-PR's Pro toggle could strand the user in an empty edit shell.
+  // (pending generation, no scenes, currently generating). Predicate lives in
+  // lib/edit/stage-mode so it can be unit-tested without rendering Stage.
   useEffect(() => {
     if (mode !== 'edit') return;
-    const isPending = currentSceneId === PENDING_SCENE_ID;
-    const stillEditable =
-      !isPending && scenes.length > 0 && generatingOutlines.length === 0 && !!currentScene;
-    if (!stillEditable) {
+    const editable = isCurrentSceneEditable({
+      currentSceneId,
+      sceneCount: scenes.length,
+      generatingOutlineCount: generatingOutlines.length,
+      hasCurrentScene: !!currentScene,
+    });
+    if (!editable) {
       setMode('playback');
     }
   }, [mode, currentSceneId, scenes.length, generatingOutlines.length, currentScene, setMode]);
--- a/lib/edit/scene-editor-registry.ts
+++ b/lib/edit/scene-editor-registry.ts
@@ -5,7 +5,21 @@ const surfaces = new Map<SceneType, SceneEditorSurface>();
 
 export const sceneEditorRegistry: SceneEditorRegistry = {
   register: (surface) => {
+    // Re-registering the same instance is benign (HMR re-executes module init
+    // and the second pass passes the identical surface object). Only warn when
+    // a different surface tries to take the slot, since that silently masks
+    // bugs like accidental double-imports from divergent paths.
+    const existing = surfaces.get(surface.sceneType);
+    if (existing && existing !== surface && process.env.NODE_ENV !== 'production') {
+      console.warn(
+        `[sceneEditorRegistry] overwriting existing surface for "${surface.sceneType}". ` +
+          `If this is HMR, call unregister first.`,
+      );
+    }
     surfaces.set(surface.sceneType, surface as SceneEditorSurface);
+  },
+  unregister: (sceneType) => {
+    surfaces.delete(sceneType);
   },
   resolve: (sceneType) => surfaces.get(sceneType),
 };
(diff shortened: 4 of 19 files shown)


wyuc:
Check #561 and corresponding sub-issues, review thoroughly plz :)

cosarah (commented):
Reviewed against issue #561 scope — framework, kernel, contract, i18n placeholders, tests all land as specified. The five findings below are kernel-level concerns that won't get fixed by the downstream surface (#562) or chrome (#563) PRs, so they belong here.

### Issues

**Important**
- `lib/edit/slide-ops.ts:295-310` — `getElementListRange` is a forked copy of `lib/utils/element.ts#getElementListRange` / `getElementRange`. The local version ignores `PPTLineElement` geometry (`start`/`end`) and rotation, and treats line `height` as 0 — so `element.align` produces wrong bounds for line or rotated elements. The canonical helper is already used elsewhere (e.g. `lib/hooks/use-canvas-operations.ts:286`); reuse it instead of forking.
- `lib/edit/slide-ops.ts:104-128` — history stack is unbounded. Each op `structuredClone`s the previous present into `past`; long edit sessions leak memory and slow undo. Add a cap (e.g. `MAX_HISTORY = 50`) and drop the oldest entry on overflow.
- `lib/edit/slide-ops.ts:141-143` — `slide.update` takes `patch: Partial<Slide>` and does `Object.assign(draft.canvas, patch)`, which accepts `{ elements: [...] }` / `{ animations: [...] }` and replaces the collections wholesale, bypassing the dedicated element ops. Narrow to `Partial<Omit<Slide, 'elements' | 'animations'>>` (or split into `slide.updateMeta`) so the kernel keeps a single source of truth for collection mutations.
- `lib/edit/slide-ops.ts:198-212` — `element.duplicate` silently skips ids missing from `idMap`, and neither `add` nor `duplicate` checks for id collisions with existing elements. Add a contract check (throw or filter) so duplicate ids can't slip into a slide.
- `lib/edit/scene-editor-registry.ts:6-11` — `register` silently overwrites a prior registration for the same `SceneType`. With HMR or accidental double-import the original surface vanishes with no signal. Dev-warn on duplicate, or no-op idempotently; consider exposing `unregister` for HMR cleanup.

**Minor**
- `lib/edit/slide-ops.ts:243-245` — `cloneElement` uses `JSON.parse(JSON.stringify(...))` while `cloneSlideContent` uses `structuredClone`. Pick one; JSON round-trip drops `undefined`, `Date`, typed arrays.
- `lib/edit/slide-ops.ts:101-108` — `applySlideEditOperation` pushes a history snapshot even when the op is a no-op (e.g. `element.update` against a missing id). `produce` returns the same reference when nothing changed; skip the push in that case so undo doesn't replay empty steps.
- `lib/edit/slide-ops.ts:135-233` — after `produce`, pushing `target.present` directly onto `past` is safe (immer's structural sharing); the extra `cloneSlideContent` is a double allocation.
- `tests/edit/slide-ops.test.ts` — coverage gaps worth filling now while the kernel is fresh: every `element.align` direction beyond `horizontal`, line / rotated elements in `align`, no-op paths for `update` / `delete` / `reorder` / `removeProps`, `element.add` index clamping, `element.duplicate` with default offset, and history "future cleared after new op after undo".
- `tests/edit/stage-mode.test.ts` — doesn't exercise the auto-exit guard. Extracting the predicate would let you lock in the policy without rendering `<Stage>`.
- `components/scene-renderers/pbl-renderer.tsx:13-17` — `mode` widened to `StageMode` but unused inside the renderer. Fine for now; flag for cleanup when PBL gets its own surface.

### Summary

Scope-complete for #561. The kernel is well-shaped and the immutable-history approach is sound — but the duplicated geometry helper (silently wrong for lines/rotation), unbounded history, and the `slide.update` escape hatch are real correctness/maintenance issues that shouldn't slip past this PR. Verdict: comment.

wyuc:
## Agent CR sweep summary

After your review I ran two passes of an independent code-review subagent against this branch's tip.

**Round 1** (against `f1bcffe` — first fix commit). All 5 Important + 4 high-value Minor items in your review confirmed addressed; subagent surfaced two new defense-in-depth notes:

1. `element.duplicate`'s shallow spread left nested refs (`start` / `end` tuples, `outline`, `points`) shared with the source — latent footgun for ops that mutate nested arrays in place.
2. `slide.update` was type-narrowed but had no runtime guard — `as any` at the call site could still bypass it.

**Round 2** (against `311befd` — follow-up). Both addressed: `element.duplicate` now does `structuredClone(current(element))` (deep), and `slide.update` throws at runtime when `'elements' in patch || 'animations' in patch`. Tests cover both. No new findings.

**Verification:**
- vitest: 374 / 374 (was 317 pre-review, +57 new coverage)
- Local tsc · lint · prettier · i18n · build — all green
- Upstream CI (Lint+Typecheck+Unit Tests, E2E Tests) — both passing on this PR's tip (the ci.yml widening picked itself up via the merged workflow that GitHub resolves for `pull_request` events)

cosarah (approved):
All five Important findings + five of six Minors fixed; remaining `PBLRenderer.mode` is fine to defer. Bonus catches — `element.duplicate` deep-clone via `structuredClone(current(...))` and the runtime guard on `slide.update` — close two real escape hatches I missed.

Local verify: `pnpm test -- tests/edit` → 374 pass, `tsc --noEmit` clean, lint unchanged.

Issue #570, opened 2026-05-13
[MAIC Editor] Enablement infrastructure (pre-slide-surface)


wyuc:
Parent: #560. Lands before #562 (slide surface).

### 1. Feature flag
- `NEXT_PUBLIC_MAIC_EDITOR_ENABLED`, default OFF.
- Gates only the Pro toggle in `Header`; `StageMode` union unchanged.
- Retained as kill-switch post-rollout.

### 2. Slide DSL schema versioning
- `slide.schemaVersion: number` on slide content; current shape is v1.
- `migrateSlide(slide)` keyed by version; v1 body empty, shape + tests in.
- Slide loader funnels through `migrateSlide`.
- Tests: pure, idempotent, v1 round-trip is identity.

### 3. Round-trip integration test harness
- New `tests/edit/round-trip/` directory.
- Harness: fixture slide + ops list → apply → PPTX export → reimport → diff.
- Initial cases: `noop` + one trivial op (e.g. change a text element's content).
- Each #562 element adds its own case.
- Runs as part of `pnpm test`.

### 4. Edit history persistence
- Op history (capped at 50 in #564) autosaves to localStorage, keyed **per scene**.
- On Stage mount: check the key; if present, prompt restore / discard.
- i18n: 2 strings (prompt title + body) × 6 locales.

### 5. Concurrency guards
- **AI regen lock** — reverse of #564's auto-exit. Edit mode holding a scene → any regenerate-this-scene call (manual retry, auto-on-failure) is **refused** (pessimistic; no queue).
- **Multi-tab detect** — `BroadcastChannel` with `localStorage` event fallback. Second tab attempting edit on the same course is refused with a clear message.
- i18n: 2 strings × 6 locales.

## Out of scope

- Server-side history sync (Phase 2).
- Cross-device conflict resolution.
- Optimistic regen guard (revisit if pessimistic friction shows up).

## Acceptance criteria

- [ ] Flag OFF: zero behavior change vs `feat/maic-editor-v0` tip
- [ ] Flag ON: Pro toggle behaves as in #568
- [ ] `slide.schemaVersion` in type + serialized; `migrateSlide` tested
- [ ] Round-trip harness with noop + trivial op cases, green in CI
- [ ] Refresh mid-edit → per-scene restore prompt on remount; accept restores, dismiss clears
- [ ] Regenerate-this-scene refused while edit mode holds it
- [ ] Second tab edit on same course refused
[...]

PR #568, opened 2026-05-13
refactor(maic-editor): drop EditModeSidebar; clean Pro mode chrome

wyuc:
## Why

`EditModeSidebar` (the 240px file-list panel with hover-revealed
up/down/delete action columns and a `+ Add slide` footer, shipped
in #565) is not the design we want for Pro mode. The tracking
issue (#560) wording is "minimal top bar + slide thumbnail rail",
and the design owner explicitly wants the sidebar gone in v0 —
future slide navigation will be redesigned and plugged into a
reserved slot.

PR #565 also kept the playback chrome wrapped around the editor
in Pro mode (Header, sidebar, Roundtable, ChatArea all stayed
mounted with only the sidebar swapped); the editor's `EditShell`
chrome (CommandBar / FloatingToolbar / HintRail) was never
actually visible since no surface registers yet. This PR replaces
that panel-swap with a cleaner canvas-slot takeover: in Pro mode
the editor owns the canvas region while the global Pro toggle
stays in the Header above.

## What ships

**Drop `EditModeSidebar` and its scaffolding.**
- `components/edit/EditModeSidebar.tsx` deleted.
- `lib/edit/reorder-scenes.ts` + `tests/edit/reorder-scenes.test.ts`
  deleted (only consumer was the sidebar).
- `edit.sidebar.*` i18n block (8 keys × 6 locales) deleted:
  `pages` / `addSlide` / `newSlide` / `expand` / `collapse` /
  `moveUp` / `moveDown` / `delete`.
- `CommandBar`'s sidebar-toggle button (which consumed
  `edit.sidebar.expand` / `collapse`) removed.

**Pro mode canvas takeover.** `Stage` keeps the playback `Header`
mounted in both modes — it owns the global Pro toggle Switch,
which is the entry **and** exit affordance (closing the Switch
exits; no separate Done-editing button). When
`mode === 'edit' && currentScene`:
- `SceneSidebar`, `Roundtable`, and `ChatArea` are not mounted.
  (The previous `AnimatePresence` slide-out on `ChatArea` is gone
  too — simpler since the two states are never mounted
  simultaneously.)
- The canvas slot renders `<EditShell scene={currentScene} />`
  instead of `<CanvasArea>`. EditShell internally resolves the
  surface via `sceneEditorRegistry`; when none is registered it
  falls through to the `edit.unsupportedScene` placeholder. With
  no surfaces registered in this PR, every scene type falls
  through to that placeholder — the verifiable v0 behavior.

[...]

PR #571, opened 2026-05-14
feat(maic-editor): enablement infrastructure (pre-slide-surface)

wyuc:
Implements #570. Lands before #562. Stacked on #568 (the diff against
`feat/maic-editor-v0` includes #568's commits until that merges).

## What ships

1. **Feature flag** — `NEXT_PUBLIC_MAIC_EDITOR_ENABLED`, default OFF.
   Pro toggle in `Header` only renders when ON. `StageMode` unchanged.

2. **Slide DSL schema versioning** — `SlideContent.schemaVersion?: number`
   (current = 1). `migrateSlideContent` / `migrateScene` are pure +
   idempotent; the stage store funnels `setScenes` / `addScene` through
   `migrateScene` so legacy data gets stamped at the boundary.

3. **Round-trip test harness** — `tests/edit/round-trip/`. Caveat: no
   PPTX → Slide reimport exists in the codebase, so the harness does
   `apply ops → buildPptxBlob → JSZip parse → assert content survived`
   instead of a full reimport diff. Per-op cases extend it in #562.
   `buildPptxBlob` is now exported (the hook is still the only runtime
   caller).

4. **Edit history persistence** — per-scene `persist` / `load` / `has` /
   `clear` helpers under `maic-editor:slide-history:${sceneId}`, plus a
   standalone `SlideHistoryRestorePrompt` dialog. 4 new i18n strings × 6
   locales. **Not wired into Stage yet** — #562 owns the entry effect.

5. **Concurrency guards** —
   - `isSceneEditLocked` predicate for the AI-regen lock (defensive;
     current call paths can't structurally trigger it).
   - `tryAcquireEditLock` / `refreshEditLock` / `releaseEditLock` /
     `isEditLockHeldByOther` for multi-tab, localStorage-backed; atomic
     acquire, idempotent heartbeat, stale-lock takeover after 3×
     heartbeat (default 15s).
   - Standalone `MultiTabEditConflictPrompt`. 3 new i18n strings × 6
     locales. **Not wired into Stage yet** — same reason as (4).

## Why dialogs aren't wired here

The slide-surface PR owns the edit-entry effect machinery — that's where
the history-state lifecycle and the per-tab `tabId` ref naturally live.
Shipping half-wired dialogs now means speculatively building Stage state
we'd restructure on contact with the surface.

## Out of scope

Server-side history sync (Phase 2); cross-device conflict resolution;
optimistic regen guard.

## Test plan

- [x] prettier · lint (0 new warnings) · tsc · `check:i18n-keys` · build — green
- [x] `pnpm test` — **410 passing / 9 failing**; the 9 are baseline
      `tests/server/ssrf-guard.test.ts` (DNS-dependent). **+45 new tests**
      across feature-flag (5), slide-schema (9), round-trip (2),
      slide-history-persistence (8), regen-lock (4), edit-mode-lock (17).
- [x] Manual: flag unset → no Pro toggle; flag set → Pro toggle behaves
      as in #568. No new user-visible affordances.


wyuc:
## Self-CR pass

Ran a code-review loop. Verdict: `READY`. Findings:

- **Feature flag correctness** — `isMaicEditorEnabled` parses
  `'true'`/`'1'` only; everything else (unset, `'false'`, `'yes'`)
  falsy; tests cover all branches. Stage call site at
  `components/stage.tsx:1025` passes `undefined` when off; Header
  gates the Switch on `onToggleEditMode &&`. ✓
- **Schema migration purity + idempotency** — same-reference return
  on already-current input; new object preserving canvas otherwise;
  non-slide scenes pass through identity. Store funneling in
  `setScenes` / `addScene` short-circuits via identity equality. ✓
- **Forward-compat fix** (CR follow-up commit \`0c1b02c\`): content
  written with a future `schemaVersion >= CURRENT` is now returned
  untouched instead of silently downgraded.
- **Round-trip harness honesty** — noop test unzips and asserts
  `ppt/slides/slide1.xml` present (not just blob size); needle test
  uses a distinctive string and asserts `.toContain(NEEDLE)` against
  the actual slide XML. ✓
- **Storage failure handling** — all four persistence helpers wrap
  storage access in try/catch; tests cover corrupted JSON and
  quota-exceeded. Multi-tab lock helpers same. ✓
- **Multi-tab lock semantics** — self-renewal allowed; refuses fresh
  other-tab lock; steals after `LOCK_STALE_MS`; `release` no-ops
  when another tab now owns; `refresh` no-ops on other-tab lock.
  Tests cover all branches. ✓
- **No new upstream coupling** — new `lib/edit/` modules touch only
  the public types (`Scene`, `SlideContent`, `StageMode`,
  `SlideEditHistory`); EditShell imports unchanged.
- **Not-wired claim is honest** — both `SlideHistoryRestorePrompt`
  and `MultiTabEditConflictPrompt` are unmounted in this diff. #562
  owns the entry effect.
- **i18n alignment** — 7 new keys × 6 locales; `check:i18n-keys`
  passes.

Tests: **46 new** (feature-flag 5 + slide-schema 10 + round-trip 2
+ slide-history-persistence 8 + regen-lock 4 + edit-mode-lock 17),
all green. 9 pre-existing \`tests/server/ssrf-guard.test.ts\` failures
are baseline (DNS-dependent, sandbox-only).

Ready for review.

cosarah (approved):
LGTM. Locally verified:

- Feature flag gates Pro toggle (with/without `NEXT_PUBLIC_MAIC_EDITOR_ENABLED`)
- 45 new tests pass (lock, schema migration, history persistence, regen lock, round-trip)
- Confirmed lock/history APIs are infra-only with no runtime callers yet — matches PR scope note (#562 owns Stage wiring)

Code quality: lock atomicity and stale-takeover logic are sound; `migrateSlideContent` is idempotent and forward-compatible; localStorage paths are SSR-safe and exception-tolerant. No blocking issues.

PR #579, opened 2026-05-17
feat(maic-editor): slide surface — skeleton + infra wiring + geometry (1/3)

wyuc:
First of three sub-PRs implementing the **slide** SceneEditorSurface for MAIC Editor (tracking: #560; builds on #564/#565/#568/#571). Draft until the stack lands.

Registering the surface lights up Pro mode for slide scenes via EditShell. This infra-first slice ships the parts every later element depends on:

**In this PR**
- `SceneEditorSurface` impl + registry registration; the surface owns a `SlideEditHistory` (the #564 kernel), not the live stage store.
- Reuses the **unmodified** slide renderer canvas behind a surface-owned scene context; drag/resize/rotate commits are diffed into canonical `element.update` ops — one gesture = one undo step.
- Numeric x/y/w/h/rotate popover as the precise fallback.
- #571 wiring: cross-tab edit lock + conflict prompt, per-scene history persistence + restore prompt, regen-lock guard.
- Prime gate: the renderer normalizes content on mount (text auto-height) before any user action; those pre-prime commits are folded into the baseline rather than staged as undoable/persisted edits, so the restore prompt no longer fires spuriously on first entry.
- Per-op round-trip test for the geometry op; bridge + session unit tests; `edit.geometry` keys in all 6 locales.

**Upstream-shared changes (minimal, justified)**
1. One additive, backward-compatible optional `controller` prop on `SceneProvider` — the uncontrolled/playback path is unchanged — so staged edits don't write through to the live lesson.
2. A `FloatingToolbar` trigger-nesting fix: it wrapped `PopoverTrigger` around `<Tooltip>` (a provider, not a DOM node), so no `popoverContent` action could ever open. This PR is the first `popoverContent` consumer to exercise that path.

**Deferred to later sub-PRs (2/3, 3/3):** text rich-text + insert palette; image replace/crop/flip; z-order; slide management; slide background. Out of scope by design (AI-first ceiling): shape elements, multi-select align/distribute, intermediate z-order, keyboard shortcuts.

Local CI green (prettier / eslint / tsc / i18n / vitest / e2e).


wyuc:
Self-reviewed before requesting eyes. Key points for reviewers:

- **Upstream coupling** is held to two minimal, additive changes: the optional `SceneProvider` `controller` prop (uncontrolled/playback path byte-unchanged) so staged edits don't write through to the live lesson; and a `FloatingToolbar` trigger-nesting fix (it wrapped `PopoverTrigger` around `<Tooltip>`, a provider — no `popoverContent` action could open; this PR is the first consumer).
- Review surfaced and fixed two real issues: the renderer's ResizeObserver text auto-height was being staged as a user edit (spurious restore prompt on entry) — now classified by pointer-gesture and folded into the baseline; and the geometry numeric panel is gated off for line elements (`PPTLineElement` omits height/rotate).
- One gesture = one undo step, one-shot history-restore, cross-tab lock lifecycle, and the regen-lock guard are all exercised; per-op geometry round-trip + bridge/session unit tests added. Local + CI green.
- Scope is geometry only; text/insert (PR2) and image/z-order/slide-mgmt/bg (PR3) are separate follow-up sub-PRs.

cosarah (approved):
LGTM. Clean separation via the `controller` prop on `SceneProvider` keeps the playback path untouched while letting the surface stage edits in its own `SlideEditHistory`. The gesture-vs-ResizeObserver discriminator (and the test for it) is a nice touch — undo survives text auto-height reflows. Bridge → diff → canonical ops keeps export/round-trip honest, and the per-op geometry round-trip test guards against tautologies.

Verified locally:
- Drag/resize/rotate = one undo step
- Numeric `GeometryPopover` commits flow into the same history
- `Move` action correctly disabled for `line` elements
- `MultiTabEditConflictPrompt` fires on a second tab
- No spurious `SlideHistoryRestorePrompt` on first entry (prime-gate works)

One small UX nit — bug or intended?

If a user makes edits, undoes everything back to the baseline, exits Pro mode, then re-enters, the `SlideHistoryRestorePrompt` still fires. Reason: `hasPersistedSlideHistory` only checks key existence, and `applyOp`/`undo`/`redo` all write to localStorage; only `Discard` clears it. So `present === baseline` with empty `past`/non-empty `future` is structurally indistinguishable from "has unsaved work" at restore-decision time.

Easy fix if it matters: in `slide-edit-session.ts`'s `replace()`, when `history.past.length === 0` and the present deep-equals the seeded baseline, call `clearPersistedSlideHistory` instead of persisting. Non-blocking — happy to defer to 2/3 or a follow-up if intentional.

wyuc:
Fixed in `4838220`. Simpler than the deep-equal: `replace()` clears the persisted history when the resulting `past` is empty — only reachable when pristine or fully undone (the non-user path never calls `replace()`), so it's equivalent without tracking the baseline. Unit test added.

cosarah:
Re-reviewed the force-push (`4838220`). The restore-prompt-after-full-undo fix is exactly the right shape:

- `replace()` clears the persisted key when `past.length === 0` instead of writing it. The comment correctly pins down the invariant that makes this safe — `replace` is only reached from user paths (`applyOp` / user-edit `commitContent` / `undo` / `redo`); `seed` and the non-user normalization branch go through raw `set` and never persist. So `past.length === 0` here unambiguously means "back at the seeded baseline."
- Test coverage matches: seed never persists → first `applyOp` persists → `undo` back to baseline clears. The in-memory localStorage shim coexists cleanly with the install-and-restore pattern in `slide-history-persistence.test.ts` (ran both files together locally — 20/20 green).

No regressions to the gesture / ResizeObserver semantics. Approval still stands.

PR #577, opened 2026-05-17
[codex] fix server LLM model auto-selection

xuruiray:
## Summary

Fix server-configured LLM startup so a provider selected by server sync also gets a usable default model when the current `modelId` is empty.

## Related Issues

Fixes #498

## Changes

- Auto-fill the first available model for a server-configured LLM provider when server sync falls back from an unusable default provider, matching the existing image/video recovery behavior.
- Add a regression test that mirrors the reported DeepSeek server-only setup: default `openai` provider, empty `modelId`, server response containing `deepseek` with `deepseek-chat`.
- Add a narrow TypeScript compatibility fix for the OpenAI-compatible `fetch` wrapper so `next build` passes with the current Next/TypeScript fetch type.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [x] CI/CD or build changes

### Steps to reproduce / test

1. Start from a store state with `providerId: openai` and an empty `modelId`.
2. Mock `/api/server-providers` to return only `deepseek` with `models: ['deepseek-chat']`.
3. Run `fetchServerProviders()` and verify the store selects `providerId: deepseek` and `modelId: deepseek-chat`.
4. Run the full local validation commands listed below.

### What you personally verified

- RED: `pnpm vitest run tests/store/settings-server-sync.test.ts -t "selects the server LLM model"` failed before the fix with `expected '' to be 'deepseek-chat'`.
- GREEN: the same targeted test passed after the store sync change.
- `pnpm vitest run tests/store/settings-server-sync.test.ts` passed: 47 tests.
- `pnpm test` passed: 45 test files, 334 tests.
- `pnpm lint && pnpm check && pnpm build` passed. Lint reports 7 existing warnings and 0 errors.
- `pnpm exec playwright install chromium` was run because the local Playwright browser cache was missing.
- `CI=1 pnpm test:e2e` passed: 11 Playwright tests. I used CI mode because a pre-existing `next dev` process held `.next/dev/lock` on port 3000; CI mode starts the production server on 3002 without touching that process.

### Evidence

- [x] CI-style local checks passed (`pnpm lint && pnpm check && pnpm build`)
- [x] Unit/integration tests passed (`pnpm test`)
- [x] E2E tests passed (`CI=1 pnpm test:e2e`)
[...]

PR #581, opened 2026-05-17
fix(settings): enforce "usable provider ⇒ concrete model" invariant (#580)

### LINKED ISSUE #580: Follow-up (#577/#498): make "usable provider ⇒ a concrete model is always selected" an invariant, and align it across code + UX
#577 functionally fixed #498 — verified e2e on `main@2ed39da` (server-only DeepSeek → `modelId` resolves to `deepseek-chat`, "Enter Classroom" enters generation, no toast). But #577 is a *reactive* patch and leaves two alignment gaps. This issue tracks converging them under one invariant.

### Invariant

`(usable provider) ⟺ (a concrete model of that provider is always selected)`. A "usable provider" = configured (server-configured **or** client API key) **and** has ≥1 model. The `(valid provider, empty modelId)` state must not be representable. Only two legal states remain:

- **A — no usable LLM provider:** one clear "configure provider" entry point.
- **B — ≥1 usable provider:** resolved `(providerId, modelId)` is always `(usable provider, a concrete model)`; the selector always shows a checked model.

**Default model rule (decided):** first available model of the resolved provider. One single rule, **shared across LLM / image / video**.

### Layer 1 — Code alignment

- Resolve provider **and** model atomically at the source — every event that makes a provider usable (server sync setting `isServerConfigured`, user entering an API key, switching provider) sets `modelId` to a concrete model in the same place. No reactive "detect empty later and patch" (that's the #577 shape).
- Extract one shared `resolveSelectedModel(...)` used by LLM/image/video. Today these diverge: LLM (post-#577) `validateModel() || (isServerConfigured ? models[0]?.id : '') || ''` — gated; image/video `recovered* || validateModel() || models[0]?.id || ''` — unconditional. Asymmetry origin: #266 added `validateModel`'s empty-input short-circuit + the recovery tail for image/video only; LLM stayed bare until #577 patched it asymmetrically.
- `validateModel` may still validate, but model **selection resolution** must never yield empty as a final state when a usable provider exists.

### Layer 2 — UX alignment

Under the invariant these states become unreachable / inconsistent and should be removed or unified (refs verified on `main@2ed39da`):

| Today | Correct form |
|---|---|
| `app/page.tsx:293-303` `if(!currentModelId)` → setup toast **+** force-open SettingsDialog | Gate on **"is there a usable provider"** (state A), and reuse the single Configure-Provider affordance — no toast + forced-dialog pair |
| `generation-toolbar.tsx:139` `configuredProviders.length>0 ? selector : "Configure Provider" CTA` | Keep — this is the correct shape; make it the **single** canonical "not configured" component |
| `generation-toolbar.tsx:767` trigger label `… \|\| t('settings.selectModel')` ("Select Model") | Dead under invariant — always `provider / modelName` |
| `generation-toolbar.tsx:784-786` `providerEntries.length===0` → `noModelsFound` | Unreachable — remove |
| `settings.modelNotConfigured` toast | State eliminated — remove |
| `components/settings/model-selector.tsx` (dead code, zero refs; carries its own no-provider/no-model empty states) | Delete the file |

### Non-goal

Re-fixing #498 (functionally resolved by #577). This is invariant + consolidation only.

Refs: #498, #577, #266

wyuc:
Closes #580.

Establishes the invariant **`(usable provider) ⟺ (a concrete model of that provider is always selected)`** for LLM / image / video, resolving provider **and** model atomically at the source. The previously-representable `(usable provider, empty modelId)` state is eliminated.

### Layer 1 — code
- New pure `resolveSelectedModel()` — like `validateModel` but no empty-input short-circuit, so it returns `''` only when the provider has no models.
- `fetchServerProviders` reconcile: symmetric LLM/image/video provider-recovery + `resolveSelectedModel`; removed the asymmetric #577 `isServerConfigured`-gated LLM tail and the separate first-load block (now also covers the client-API-key path #498/#577 missed).
- One shared `resolveLLMSelection()` used by **both** `setProviderConfig` (single edit: key add/clear, model-list edit) and `setProvidersConfig` (bulk: delete provider, import, reset) — keep current provider if still usable, else first usable, else State A. Image/video config setters get the symmetric treatment.

### Layer 2 — UX (states now unreachable)
- Shared `isLLMProviderConfigured` / `hasUsableLLMProvider`; landing page gates generation on a usable provider and drops the `modelNotConfigured` toast + forced settings dialog. Toolbar uses the shared predicate; dead "Select Model" label/tooltip removed.
- Deleted unused `components/settings/model-selector.tsx`; pruned the orphan `settings.selectModel` i18n key (6 locales).

### Deviations from the issue's literal table (please confirm)
- **Kept** `generation-toolbar` `providerEntries.length === 0` branch — it is reachable via *search* (search-no-results), not the dead no-provider state the issue assumed.
- **Kept** `settings.modelNotConfigured` i18n key — still used by the chat path (`use-chat-sessions.ts`), a separate reachable surface.
- **Keyless providers** (ollama/lemonade) now require an explicit user `baseUrl` to count as "configured" (registry `defaultBaseUrl` alone isn't user intent) — without this, `hasUsableProvider` was true while the reconcile never selected them, re-enabling generation with no model.

### Testing
- New `resolveSelectedModel` / `hasUsableLLMProvider` unit tests; server-sync invariant tests (client-key resolution, atomic API-key entry/clear, provider switch, custom-model delete, bulk delete, serverModels-preference + first-load regression guards). 96/96 store tests, full suite green (the only failures are pre-existing environmental `ssrf-guard` DNS tests, untouched here). `tsc` + `eslint` clean. Added an e2e spec (State A / State B).

Based on `main@2ed39da`; will rebase before merge if `main` advances.


cosarah (approved):
LGTM. Code: shared `resolveLLMSelection` symmetrically applied to single + bulk config edits + image/video setters; `isLLMProviderConfigured` consolidates the gate; `resolveSelectedModel` cleanly separates selection-resolution from validation. Comments explain the invariant well.

Locally tested on pr-581 worktree (port 3001):
- State A (no usable provider): generate button disabled, single "Set up model" CTA, no toast, no forced dialog.
- State B: toggling `requiresApiKey` on the active provider with no apiKey immediately switches `providerId` to the next usable one (or to State A) — capsule updates without page refresh. Earlier mis-report on my end was caused by testing against an older dev server.
- Clearing/adding apiKey, deleting the selected custom model, and bulk provider delete all re-resolve atomically as described.
- Keyless providers correctly require an explicit user `baseUrl` before counting as configured.

Unit + store tests 364/365 (the one failure is the pre-existing `ssrf-guard` DNS test, unrelated). `tsc` clean.

Issue #183, opened 2026-03-21
Per-provider proxy support for server-side fetch calls


wyuc:
Node.js native `fetch()` doesn't respect `HTTP_PROXY`/`HTTPS_PROXY`. We have a `proxyFetch` wrapper using undici, but currently only Tavily web search uses it. All other server-side providers (LLM, TTS, ASR, image, video) use raw `fetch()`, causing `fetch failed` errors for users who need a proxy.

Blanket proxying isn't ideal either — domestic providers (Seedance, Kling, Qwen, etc.) don't need it and may break if routed through an overseas proxy. A per-provider `proxy: true` flag in config would let each provider opt in.

PR #589, opened 2026-05-22
feat: add SearXNG self-hosted web search provider support

### LINKED ISSUE #569: [Feature]: 希望增加对本地部署 Web Search API 的支持。
### Problem or Motivation

无论在网页还是 .env.example 只提供了商用 Web Search API 的配置，希望可以支持本地部署的 Web Search API 的配置。例如： SearXNG 。

### Proposed Solution

感觉提供一组环境变量就行。

例如：

```
SearXNG _API_KEY=
SearXNG _BASE_URL=
```

### Alternatives Considered

_No response_

### Area

API / Backend

### Additional Context

_No response_

tongshu2023:
@
## Summary

Fixes #569

Adds support for SearXNG (https://github.com/searxng/searxng), the most popular self-hosted metasearch engine. Users running privacy-conscious or air-gapped deployments can now use their own SearXNG instance as a web search provider instead of relying on third-party APIs.

## Changes

- **New provider**: `searxng` — registered with `requiresApiKey: false`, endpoint `/search?format=json`
- **New file**: `lib/web-search/searxng.ts` — implementation using SearXNG JSON API with optional Bearer token auth
- **Provider config**: Added `SEARXNG_BASE_URL` / `SEARXNG_API_KEY` env vars, plus `keylessProviders` support for web search in `provider-config.ts`
- **URL validation**: Modified `resolveSafeClientWebSearchBaseUrl` to allow any base URL when the provider has an empty allowed-list (self-hosted providers)
- **Store**: Added SearXNG to `getDefaultWebSearchConfig` with `requiresApiKey: false`
- **i18n**: Added SearXNG display names in all 6 supported locales (`en-US`, `zh-CN`, `zh-TW`, `ar-SA`, `ja-JP`, `ru-RU`)
- **Docs**: Updated `.env.example` with SearXNG configuration comments

## Test Plan

- [x] TypeScript compilation passes (only pre-existing `mathml2omml` error unrelated to this PR)
- [x] All 13 modified + 1 new files reviewed
- [x] `WEB_SEARCH_PROVIDERS` registry iterated in settings store and toolbar UI — SearXNG auto-appears
- [ ] Manual test: configure `SEARXNG_BASE_URL` to a running SearXNG instance and verify search results
- [ ] Manual test: verify SearXNG appears in Settings > Web Search provider selector
@


wyuc (changes requested):
Thanks @tongshu2023 — verified end-to-end against a real SearXNG instance; the feature works (search, client base URL, Bearer auth). A few fixes before merge:

- **CI**: `prettier --check` fails on `searxng.ts` — run `pnpm exec prettier --write lib/web-search/searxng.ts`.
- **Missing icon**: `constants.ts` points to `/logos/searxng.png` but no asset is added — 404s, no icon in Settings.
- **Empty Base URL field**: no `defaultBaseUrl`, so the one provider that *needs* a user-supplied URL shows no placeholder. Surface `SEARXNG_DEFAULT_BASE_URL` as `defaultBaseUrl`.
- **`/search` doubling**: `buildSearXNG…Url` appends `/search` unconditionally — `brave.ts` guards with `endsWith('/search')`, match it.
- **`proxyFetch` ignores `NO_PROXY`**: a deployment with `HTTP_PROXY` set routes the localhost SearXNG request through the external proxy and fails. Fix in `proxy-fetch.ts` or track as follow-up (#183/#529).

Nit: typo `buildSearXNGSearcHUrl`.

tongshu2023:
﻿Thanks for the thorough end-to-end verification @wyuc! All five items + the typo addressed in c20c386:

- ✅ **Prettier** applied to `searxng.ts`
- ✅ **Icon**: added the official SearXNG logo as `public/logos/searxng.svg` (from the upstream brand assets) and pointed `constants.ts` at it
- ✅ **defaultBaseUrl**: `http://localhost:8888` now surfaces in `WEB_SEARCH_PROVIDERS.searxng`, so the Settings base URL field shows a placeholder (same duplicated-literal pattern brave uses)
- ✅ **`/search` doubling**: guarded with `endsWith('/search')`, matching `brave.ts` — covered by a test
- ✅ **`NO_PROXY`**: implemented in `proxy-fetch.ts` rather than deferring — `shouldBypassProxy()` follows the curl convention (exact host, domain suffix with/without leading dot, optional `:port`, `*`), and `proxyFetch` falls back to direct fetch on match. This should also help #183/#529. 6 dedicated tests.
- ✅ **Typo**: `buildSearXNGSearcHUrl` → `buildSearXNGSearchUrl`

Added `tests/web-search/searxng.test.ts` (URL building, Bearer auth, maxResults slicing, error path) — 29 tests green locally across web-search + proxy-fetch, plus a clean `tsc --noEmit`.

tongshu2023:
Thanks for the thorough end-to-end review @wyuc! All items addressed, and the branch is rebased onto main (resolved the registration-point overlap with the new MiniMax provider — both now coexist).

- **CI / prettier**: formatting applied to all touched files; `prettier --check`, `eslint`, and `tsc --noEmit` are clean locally.
- **Missing icon**: added `public/logos/searxng.svg` (official brand asset from the SearXNG repo, `client/simple/src/brand/searxng.svg`); `constants.ts` now points to the `.svg`.
- **Empty Base URL field**: `SEARXNG_DEFAULT_BASE_URL` (`http://localhost:8888`) now lives in `lib/web-search/constants.ts` and is surfaced as `defaultBaseUrl`, so `web-search-settings.tsx` renders it as the placeholder. It stays a placeholder (not pre-filled) since the user must point at their own instance.
- **`/search` doubling**: the URL builder now guards with `endsWith('/search')`, matching `brave.ts`.
- **Typo**: `buildSearXNGSearcHUrl` → `buildSearXNGSearchUrl`.
- **Tests**: added `tests/web-search/searxng.test.ts` — default base URL fallback, the `/search` no-doubling guard, Bearer-header presence/absence, result mapping with `maxResults`, and non-OK error shape. 29/29 in `tests/web-search` pass.

On **`proxyFetch` ignoring `NO_PROXY`**: agreed it belongs in `proxy-fetch.ts` rather than per-provider — I'd rather not bury a cross-cutting proxy fix inside this feature PR. I'm happy to send a follow-up PR adding `NO_PROXY`/localhost bypass to `proxyFetch` (referencing #183/#529) right after this one.

tongshu2023:
Follow-up for the `proxyFetch`/`NO_PROXY` item is up: #718 (loopback always bypasses the proxy + standard `no_proxy`/`NO_PROXY` semantics). With that, every item from the review has a home.

wyuc:
Closing as superseded by #842, which has now landed on `main` with the SearXNG provider, server-managed configuration, UI/i18n wiring, URL normalization, focused tests, and the hardened client/server URL trust boundary. Thank you for the original implementation and the detailed follow-up work here; it clearly helped validate the use case.

PR #593, opened 2026-05-24
fix: remove defaultBaseUrl fallback in TTS availability check

### LINKED ISSUE #574: TTS provider availability check produces false positives, breaking multi-agent TTS when local servers aren't running
## Summary

`getAvailableProvidersWithVoices` in `lib/audio/voice-resolver.ts` marks keyless local TTS providers (e.g. VoxCPM, Lemonade) as available whenever their `defaultBaseUrl` is set in the static registry — even when the user has neither configured nor deployed the server. The inflated voice list is then handed to the agent-profile generator, which can assign an unreachable provider to an agent, causing TTS failures at runtime.

## Reproduction

1. Locally deploy **only** the Lemonade TTS server (e.g. `http://localhost:13305/v1`). Do **not** deploy VoxCPM.
2. Do not configure VoxCPM in settings (no `baseUrl`/`serverBaseUrl`).
3. Generate a classroom via `/generation-preview`.
4. Observe agent profiles: some agents get `voiceConfig.providerId = 'voxcpm-tts'`, others `'lemonade-tts'`.
5. During TTS generation, VoxCPM-assigned agents error out (connection refused on `127.0.0.1:8000`); Lemonade-assigned agents succeed.

## Root cause

`lib/audio/voice-resolver.ts:133-142`:

\`\`\`ts
const isKeylessLocalProvider =
  !config.requiresApiKey &&
  !!(
    providerConfig?.serverBaseUrl?.trim() ||
    providerConfig?.baseUrl?.trim() ||
    config.defaultBaseUrl   // ← always truthy for built-in keyless local providers
  );
\`\`\`

VoxCPM (`lib/audio/constants.ts:746-761`) declares `requiresApiKey: false` and `defaultBaseUrl: 'http://127.0.0.1:8000'`. The third disjunct hits unconditionally, so VoxCPM is reported as available regardless of user config or server liveness. Lemonade hits the same disjunct (`defaultBaseUrl: 'http://localhost:13305/v1'`), but in the repro it happens to be running.

The inflated list is consumed by `app/generation-preview/page.tsx:676-704` (`getAvailableVoicesForGeneration`) and fed to `/api/generate/agent-profiles` as `availableVoices`. The LLM is explicitly prompted to *\"try to use different voices for each agent\"* (`app/api/generate/agent-profiles/route.ts:102-107`), so it distributes voices across providers — happily handing out VoxCPM voices that cannot be served.

At TTS time the failure is silent per agent: `app/generation-preview/page.tsx:885-933` just increments `ttsFailCount`. The course is generated with a partially broken agent voice assignment.

## Impact

- Imported and freshly generated classrooms can ship with agents bound to TTS providers that error at playback.
- Effect is non-deterministic from the user's perspective: \"some agents work, some don't\".
- Affects any keyless local provider with a `defaultBaseUrl` in the registry. Today: VoxCPM, Lemonade. Future keyless local providers will inherit the same bug.

## Proposed fix

Drop the `config.defaultBaseUrl` fallback from the availability check so that a keyless local provider only counts as available when the user has explicitly set `serverBaseUrl` or `baseUrl`:

\`\`\`ts
const isKeylessLocalProvider =
  !config.requiresApiKey &&
  !!(providerConfig?.serverBaseUrl?.trim() || providerConfig?.baseUrl?.trim());
\`\`\`

UX follow-up: in TTS settings, offer a one-click \"Use default URL\" action that writes `defaultBaseUrl` into the user config, so default-port deployments stay one click away.

Optional hardening (separate change):

- Health-probe keyless local providers (with short cache) before listing them.
- Surface per-agent TTS failures during generation instead of silently counting them.

## Affected files

- \`lib/audio/voice-resolver.ts\` — availability logic
- \`app/generation-preview/page.tsx\` — consumer; silent failure counter
- \`app/api/generate/agent-profiles/route.ts\` — voice distribution prompt

tongshu2023:
﻿## Summary
Fixes #574 — TTS provider availability check produces false positives for local keyless providers.

## Problem
The `isKeylessLocalProvider` condition in `lib/audio/voice-resolver.ts` included `config.defaultBaseUrl` as a fallback. This caused local keyless TTS providers (VoxCPM, Lemonade) to be marked as "available" even when the user had not deployed the corresponding server. Agents were then assigned to unreachable providers, resulting in silent TTS failures (connection refused) for some agents while others worked fine.

## Root Cause
`lib/audio/voice-resolver.ts` lines 133-139 — the `isKeylessLocalProvider` check unconditionally accepted `config.defaultBaseUrl` as a valid URL, ignoring whether the user had actually deployed and configured the server.

## Changes
- **1 file changed**: `lib/audio/voice-resolver.ts`
- Removed `config.defaultBaseUrl` from the `isKeylessLocalProvider` availability check
- Local keyless providers are now only available when the user explicitly sets `serverBaseUrl` or `baseUrl` in their configuration

## Test Plan
1. Deploy only Lemonade TTS server (do not deploy VoxCPM)
2. Do not configure VoxCPM base URL in settings
3. Generate a classroom
4. ✅ All agent voices should use providers with actual configured URLs (no VoxCPM)
5. ✅ Previously broken agents should no longer be assigned to unreachable providers

Related to #587 — its logs show `lemonade-tts: fetch failed`, consistent with this fix's symptom. Worth asking the reporter to verify whether Lemonade was explicitly configured after this merges.

### DIFF
--- a/lib/audio/voice-resolver.ts
+++ b/lib/audio/voice-resolver.ts
@@ -134,8 +134,7 @@ export function getAvailableProvidersWithVoices(
       !config.requiresApiKey &&
       !!(
         providerConfig?.serverBaseUrl?.trim() ||
-        providerConfig?.baseUrl?.trim() ||
-        config.defaultBaseUrl
+        providerConfig?.baseUrl?.trim()
       );
     const isLocalVoxCPM =
       providerId === VOXCPM_TTS_PROVIDER_ID &&
--- a/lib/audio/voice-resolver.ts
+++ b/lib/audio/voice-resolver.ts
@@ -132,10 +132,7 @@ export function getAvailableProvidersWithVoices(
     const isServerConfigured = providerConfig?.isServerConfigured === true;
     const isKeylessLocalProvider =
       !config.requiresApiKey &&
-      !!(
-        providerConfig?.serverBaseUrl?.trim() ||
-        providerConfig?.baseUrl?.trim()
-      );
+      !!(providerConfig?.serverBaseUrl?.trim() || providerConfig?.baseUrl?.trim());
     const isLocalVoxCPM =
       providerId === VOXCPM_TTS_PROVIDER_ID &&
       !!(providerConfig?.serverBaseUrl?.trim() || providerConfig?.baseUrl?.trim());
--- a/tests/audio/voice-resolver.test.ts
+++ b/tests/audio/voice-resolver.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, it } from 'vitest';
+import { getAvailableProvidersWithVoices } from '@/lib/audio/voice-resolver';
+
+describe('getAvailableProvidersWithVoices keyless availability (#574)', () => {
+  it('excludes a keyless provider when only defaultBaseUrl exists (no user-configured URL)', () => {
+    // lemonade-tts is keyless and ships defaultBaseUrl; with no user config it must not be listed
+    const providers = getAvailableProvidersWithVoices({});
+    expect(providers.map((p) => p.providerId)).not.toContain('lemonade-tts');
+  });
+
+  it('excludes a keyless provider whose config has neither serverBaseUrl nor baseUrl', () => {
+    const providers = getAvailableProvidersWithVoices({
+      'lemonade-tts': { enabled: true },
+    });
+    expect(providers.map((p) => p.providerId)).not.toContain('lemonade-tts');
+  });
+
+  it('includes a keyless provider when the user sets serverBaseUrl', () => {
+    const providers = getAvailableProvidersWithVoices({
+      'lemonade-tts': { serverBaseUrl: 'http://localhost:13305/v1' },
+    });
+    expect(providers.map((p) => p.providerId)).toContain('lemonade-tts');
+  });
+
+  it('includes a keyless provider when the user sets baseUrl', () => {
+    const providers = getAvailableProvidersWithVoices({
+      'lemonade-tts': { baseUrl: 'http://localhost:13305/v1' },
+    });
+    expect(providers.map((p) => p.providerId)).toContain('lemonade-tts');
+  });
+
+  it('ignores whitespace-only URLs for keyless providers', () => {
+    const providers = getAvailableProvidersWithVoices({
+      'lemonade-tts': { serverBaseUrl: '   ', baseUrl: '' },
+    });
+    expect(providers.map((p) => p.providerId)).not.toContain('lemonade-tts');
+  });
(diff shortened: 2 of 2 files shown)


YizukiAme:
Hello @tongshu2023! The fix itself is clean👍, but Prettier is failing:
```
pnpm exec prettier --check lib/audio/voice-resolver.ts
# Code style issues found in the above file.
```
Could you run `pnpm exec prettier --write lib/audio/voice-resolver.ts` and push the formatting fix? Should be a one-liner change.

wyuc (changes requested):
Thanks @tongshu2023, fix matches the #574 proposal exactly. Before merge:

- CI red on prettier. Run `pnpm exec prettier --write lib/audio/voice-resolver.ts` and re-push.
- Add a vitest: keyless provider with only `defaultBaseUrl` (no user-set `serverBaseUrl`/`baseUrl`) should NOT appear in the returned list.
- Add `Related to #587` in the description. Its logs show `lemonade-tts: fetch failed`, consistent with this fix's symptom, but we can't tell from the report whether the user configured Lemonade explicitly. Worth asking the reporter to verify after merge.

Non-blocking: `isLocalVoxCPM` (lines 140–142 / 151) becomes redundant after this fix. Happy to take it as follow-up.

tongshu2023:
﻿Thanks for the review @wyuc @YizukiAme! All three items addressed in 08d28bd:

- ✅ Prettier formatting applied to `lib/audio/voice-resolver.ts`
- ✅ Added `tests/audio/voice-resolver.test.ts` — covers the regression case (keyless provider with only `defaultBaseUrl` and no user-set `serverBaseUrl`/`baseUrl` is NOT listed), plus positive controls (serverBaseUrl / baseUrl / apiKey paths) and whitespace-only URL handling. 6 tests, all green locally.
- ✅ Description updated with `Related to #587`

Agreed on `isLocalVoxCPM` being redundant after this change — happy to send that cleanup as a follow-up PR once this merges.

tongshu2023:
Closing this: the provider-enablement refactor in #665/#666 supersedes it. `isTTSProviderConfigured()` in `lib/audio/provider-enablement.ts` now deliberately omits the `defaultBaseUrl` fallback (documented in its doc comment as #665 symptom 1), and the regression case this PR's test covered is now exercised by `tests/audio/provider-enablement.test.ts` and `tests/audio/voice-resolver-enablement.test.ts`. Nothing left to merge — thanks @wyuc @YizukiAme for the earlier reviews. The `isLocalVoxCPM` redundancy noted as follow-up was also removed by the same refactor.

PR #599, opened 2026-05-26
test+fix(orchestration): rip maxTurns; restore answer-the-user's-question behavior (#598 follow-up)

### LINKED ISSUE #598: Follow-up to #554: remaining items for #511
## Context

#554 lands a focused director-layer fix for one of #511's three symptoms (premature `END`). Scope was narrowed during review and the regression eval was deferred. This tracks what's outstanding.

## 1. Agent-side attribution clarification (covers #511 symptoms ① and ②)

#554 only modifies the director path. The other two #511 symptoms originate in the **agent** path and are unchanged:

- ① **Off-topic replies** — agent ignores the actual content of the prior turn.
- ② **Role confusion** — teacher misattributes a human turn to a peer agent (the "照镜子" case).

In `director-graph.ts:290`, `convertMessagesToOpenAI(state.messages, agentId)` re-encodes peer agents as `role:'user'` with a `[AgentName]:` prefix. The human's turn is also `role:'user'` with `[You]:` (or its localised equivalent) prefix. The agent system prompt doesn't currently document this encoding, so the model has no structural signal to distinguish them.

**Fix:** add one or two sentences in `lib/orchestration/prompt-builder.ts` clarifying that `role:'user'` messages prefixed `[Name]:` are AI peers, not the human student. This was item 3 of @ashutoshrana's original plan in #511; it was dropped from #554 to keep scope tight.

## 2. End-to-end regression for premature-END

Deferred from #554. The unit tests there verify the director's **inputs** (summary labels), not its **behaviour**. The premature-END symptom is model-dependent — a deterministic e2e fixture that fails on `main` and passes with the fix is the real regression guard.

Frame as **"verify #511 is actually fixed"**, not "add fixtures": reproduce the bug deterministically first (which models / prompts / seeds?), then encode the regression. If it doesn't reproduce reliably on any shipped model config, that's data we need before trusting the fix.

Lands under a new `/eval/orchestration/` domain dir (not `tests/eval/`), following the `outline-language` pattern: `runner.ts` + `judge.ts` + `scenarios/*.json` + an `eval:orchestration` npm script.

## 3. Director Rule #11 wording

Rule #11 in `director/system.md` implicitly treats every `[Student (Human)]` turn as a question, so it over-blocks `END` when the human just acknowledges ("ok continue", "got it"). Suggested rewording: gate on **"the most recent `[Student (Human)]` line asks a question or raises an objection AND appears after the last substantive `[Agent]` answer"**.

## 4. Test coverage for `[…]:`-leading content

`SENDER_PREFIX_RE = /^\[[^\]]+\]:\s*/` in `conversation-summary.ts` greedily strips the first `[…]:` pattern. Content like `"[FYI]: my answer is…"` loses `[FYI]: ` in the summary. Low-probability in practice; a single test pinning the chosen behaviour (strip vs preserve) would document the contract.

## 5. JSDoc contract on `summarizeConversation`

Add a line clarifying that `system` role messages don't appear in director-path input (`convertMessagesToOpenAI` filters non-user/assistant), so the `'System'` label branch is defensive-only.

---

Items 1 and 2 are substantive. 3–5 are polish on #554's own work. Happy to coordinate via comments if anyone wants to bundle multiple.

wyuc:
Closes #598 follow-up items 1 and 2 (and surfaces additional in-product fixes discovered during the prod investigation).

The root symptom this PR addresses: agents drift onto adjacent topics rather than answering the user's literal question ("答非所问"). The director then keeps picking peer agents for "variety" instead of routing back to the teacher to actually answer. Frustration signals from the user are just one sub-case where this is loud enough to notice; the underlying bug fires earlier, often on the first response.

### Commit 1: premature-END regression guard

5 scenarios modelled on #511's described symptoms. A/B across 9 models × 5 scenarios × 5 samples per variant: **0% END rate in both variants on every scenario**. The prompt-layer rules added in #554 don't measurably change director END behavior on these inputs. Kept as a regression guard so any future change that raises the END rate trips this eval.

### Commit 2: director question-answering eval

5 synthesized scenarios — 3 first-turn drift (no frustration yet), 2 with escalated frustration. Decision rule-judged: USER | TEACHER | OTHER_AGENT | END. A/B: current main vs main with rule 13 stripped.

### Commit 3: director rule 13 — Answering the User's Question

Constrained-output rule injected at the end of the # Rules section. Triggers whenever the conversation has an unanswered user question. Constraints: route to LITERAL `role: teacher` by default, USER cue when ambiguous or after multiple unresolved complaints, forbid peer-agent "variety" routing and END while the question is open.

On google:gemini-3-flash-preview, 5 samples per variant:

| Scenario | baseline | with_rule |
|---|---|---|
| math_quadratic_axis_drift_no_frustration | 100% | 100% |
| bio_dark_reaction_drift_no_frustration | 80% | 100% |
| english_team_isare_drift_no_frustration | 20% | 100% |
| physics_inertial_mass_escalated_frustration | 0% | 100% |
| calculus_product_rule_drift_no_frustration | 60% | 100% |
| **mean** | **52%** | **100%** |

3 of 5 scenarios reproduce the bug at baseline (≤60% correct, peer-agent "variety" routing dominates). The rule lifts every scenario to 100%.

### Commit 4: agent rule — Answering the User's Question

Adds a section to the agent system prompt that directs the agent, on every response where the user message contains a question, to lead with the concrete answer to the literal question before any other content. Pairs with the director rule — director routes to the teacher when the question is unanswered; the teacher now has explicit guidance for what to do when dispatched.

### Commit 5: rip maxTurns end-to-end + distinguish errors from completion

**maxTurns rip**: the director graph used `maxTurns: turnCount + 1` to force "one director→agent cycle per request" — a hack expressing the single-round contract as a turn limit. Replaced with explicit graph topology (`agent_generate → END`). The user-facing "Max Discussion Turns" setting is also removed: with rule-driven routing (cue_user / END), a numeric cap is redundant and was previously masking director bugs by force-ending.

**Error vs completion**: before, every non-`cue_user` outcome of the client agent loop mapped to `status: 'completed'`, conflating `end` (legitimate completion) with `empty_turns` (parser failure / model timeout) and `no_done` (SSE infra failure). The new switch routes the error paths through `clearLiveSessionAfterError`, which appends an inline System message and sets the new `status: 'error'`. The session list gets a red `AlertCircle` icon for the error state.

Touches: `director-graph.ts`, `lib/types/chat.ts`, `lib/store/settings.ts`, `components/agent/agent-bar.tsx`, `components/settings/agent-settings.tsx`, `lib/chat/agent-loop.ts`, `components/chat/use-chat-sessions.ts`, `components/chat/session-list.tsx`, `eval/whiteboard-layout/runner.ts`, 6 i18n locales.

## Test plan

- [x] `pnpm check` clean
- [x] `pnpm test` — no new failures (9 pre-existing ssrf-guard DNS-bound failures unrelated)
- [x] `npx tsc --noEmit -p tsconfig.json` clean
- [x] `EVAL_DIRECTOR_MODEL=<model> pnpm eval:orchestration` — premature-END eval passes (regression guard)
- [x] `EVAL_DIRECTOR_MODEL=<model> npx tsx --env-file=.env.local eval/orchestration/answering-runner.ts` — 5/5 scenarios PASS, mean 100% correct rate with rule, baseline 52%
- [ ] dev server sanity (manual) — multi-agent discussion, user asks specific question → check teacher answers directly; agent failure to produce output → check session-list error icon


wyuc:
## CR loop summary (2 rounds, converged)

Ran 2 rounds of code review via the `superpowers:requesting-code-review` subagent on the full PR diff (`origin/main..HEAD`).

**Round 1** flagged 2 Important issues:
1. `components/chat/use-chat-sessions.ts:460` — dead `settingsState` binding (orphan from the maxTurns rip), tripping `@typescript-eslint/no-unused-vars`.
2. `package.json` — missing `eval:orchestration:answering` convenience script (asymmetric with `eval:orchestration`).

Both fixed in `9a90e84` (4 lines).

**Round 2** verified:
- ESLint clean on the touched file.
- `pnpm exec tsc --noEmit` clean.
- `pnpm check` (prettier) clean.
- `pnpm check:i18n-keys` passes (7 locales, including pt-BR).
- `AgentLoopOutcome.reason` union is exhaustively handled by the new outcome switch.
- `maxTurns` / `currentTurn` / `max_turns` rip is complete; only remaining references are two explanatory comments in `director-graph.ts`.

Round 2 verdict: loop converged, no new findings.

Out-of-scope items deliberately left for follow-up (called out but not addressed in this PR):
- Pre-existing `directorNode` error catch still routes to `shouldEnd: true` (i.e. completed rather than error) — independent path from this PR's outcome-handler upgrade.
- `AgentSettings` component is orphaned on `origin/main` too; not introduced by this PR.
- Retry-last-message affordance for the new `'error'` session state — separate UX work.

xuyuanwei678:
1.After asking a question and then following up, the teacher no longer responds.
image

xuyuanwei678:
image
2.When I type “Can you take a look at this?” without any context, there’s a chance that nobody will respond.


image
3. When I ask a genuinely ambiguous question, such as “Can you take a look at this?” without any context, the user is not asked to clarify.

wyuc:
Thanks @xuyuanwei678 — pushed fixes for 2 & 3, and traced 1.

**Issues 2 & 3 (ambiguous "帮我看一下这个" → no response / no clarification)** — fixed in `ded3493`.

Root cause: the director cued `USER` for a vague turn, which makes no agent speak — control just returns to the user, so they face silence ("nobody responds"); and when an agent did speak it guessed a topic. Fix: director rule 13 now routes any unaddressed user turn (question / vague / frustration) to the `role: teacher` agent only — USER cue is forbidden as a response to a user turn — and the agent prompt tells the teacher to ask one short clarifying question (with concrete options) when the message is too vague, instead of guessing or going silent. Verified: "帮我看一下这个" now yields e.g. *"…你想讨论的是代码逻辑，还是 Agent 的决策流程图呢?"*. The director eval passes all 7 scenarios on gemini-3-flash-preview (10 samples each), incl. the vague follow-up (20% → 100%).

**Issue 1 (follow-up → 已结束 with no response)** — I couldn't reproduce on this PR branch (7 configs: single/multi agent, QA/discussion, direct API/real UI) nor on current `origin/main`. The most likely causes both point at older code, and this PR closes each:

1. **Director ending erroneously.** Simulating the deployed pre-#554 prompt (anti-END rules stripped), the director ENDs ~60% of the time on a "topic discussed but the specific question wasn't answered" turn — rule 4 ("END if complete") + rule 5 ("don't drag on") win with no guard against it. Rule 13 now forbids END (and overrides rules 4/5) while a user turn is unaddressed.

2. **Provider API instability.** A flaky/empty/timed-out model response was silently treated as a normal completion → a quiet 已结束 with no agent text — which matches the screenshot (correct first answer, then nothing). The error-handling change in this PR surfaces that as an explicit error state (red icon + a System message to retry) instead of a silent "completed", and the maxTurns removal drops the single-agent force-end path that also produced 已结束.

The deployed test/prod site currently runs that older code. Could you confirm which environment you tested? A repro on a fresh checkout of this branch — with the `[Director] …` server log line for the follow-up request — would let me close this one for certain.

xuyuanwei678:
Thanks for the clarification.

I was actually testing locally on this branch (feat/598-eval-orchestration) with the DeepSeek API when I first hit issue 1. After pulling the latest changes and retesting, I can no longer reproduce issues 1, 2, or 3.

Given that, issue 1 was likely caused by the provider API instability you mentioned (e.g. timeout or empty response). This PR now surfaces that case explicitly instead of silently ending the conversation. LGTM from my side.

xuyuanwei678 (approved):

PR #601, opened 2026-05-26
feat(maic-editor): nav rail, scene management, Pro mode chrome rework

wyuc:
Closes the deferred work from #560 / #562 — Phase 1 slide surface Pro mode now has a
working left rail with full slide management, plus an architectural rework of the
Stage chrome that lifts two long-standing UX bugs.

### Slide nav rail + scene management
- `SlideNavRail` — vertical thumbnail strip via motion.dev `Reorder.Group`, drag-to-
  reorder, drag-to-resize, collapse toggle. All scene types render first-class
  (slide gets `ThumbnailSlide`, non-slide gets playback `SceneSidebar` mockups via
  the extracted `SceneThumbnailContent`).
- `+ Add` between every thumb plus one above the first. Per-tile menu has Rename /
  Duplicate / Delete; Rename also responds to double-click; Enter commits, Esc
  cancels. Delete uses a toast Undo restoring at the original index.
- New store action `useStageStore.insertSceneAfter` — validates stage id, splices,
  rebalances `order`, triggers `debouncedSave`.
- Drag-in-flight writes the rail width directly on the DOM and only commits to the
  settings store on mouseup, so 60 Hz drag never serializes the (large) persisted
  object.

### Pro mode chrome rework
- `stage.tsx` (was 1391 LOC) is now a 113-line dispatcher: mode + cross-tab edit-lock
  + Pro toggle coordination + multi-tab conflict prompt. Everything else moved into
  two independent roots:
  - `PlaybackChromeRoot` — every bit of playback chrome and engine state. Exposes
    `teardown()` via forwardRef so the toggle can await SSE / engine / TTS shutdown
    before unmounting.
  - `EditChromeRoot` — Pro mode chrome (EditShell + SlideNavRail + HeaderControls
    trailing). Owns ``body[data-maic-editor]`` lifecycle.
- `StageGrid` is a CSS-Grid named-slot shell (`top / left / right / bottom / center`)
  with the left column spanning all rows, so the sidebar reaches the absolute top
  edge in both modes.
- The two shells now have an **identical outer shape** — sidebar full-height-left,
  Header 80 px top-right of content, content fills. Mode toggle cross-fades only the
  contents.
- `motion.layoutId` on the Pro Switch + settings pill morphs them between modes so
  the click target never teleports. Drawer-style swap (`AnimatePresence`, 280 ms
  `CHROME_EASE`) layered via `absolute inset-0` keeps `canvasStore` to a single
  writer across the transition.

### Two bugs the rework solves
- **Chrome remount on slide ↔ non-slide swap** — old `EditShell` returned two
  different component types (`WithSurface` vs `ReadOnly`) so React unmounted the
  whole subtree (CommandBar + leftRail + SlideNavRail) on every scene-type change.
  Fixed via a `NOOP_SURFACE` fallback in `lib/edit/noop-surface.tsx` and a
  `SurfaceStateRunner` keyed by `scene.type` that publishes surface state up to a
  stable Frame.
- **Rail drag jitter / thumbnail rescaling** — `motion.aside` was running motion's
  element-tracking pipeline every frame even with `animate={false}`. Replaced with
  plain `<aside>` + CSS transition, direct-DOM width write during the gesture,
  `Reorder.Item layout="position"` so reorder still animates but parent-driven
  resizes don't fight it.

### Other UX
- Floating insert toolbar — Text box / Image moved out of CommandBar into a
  collapsible pill that floats centered above the canvas. State persists in
  `settings.editInsertToolbarCollapsed`.
- `ProsemirrorEditor` auto-focuses when its element becomes the editing target —
  inserting a Text box no longer requires a second click before typing.
- Right cluster of the chrome bar now reads `[language · theme · settings]  [PRO]
  [Download]` in both modes (Download was previously playback-only and lived inside
  the settings pill).

## Verification
- `pnpm test`: 480 / 489 pass. The 9 failures in `tests/server/ssrf-guard.test.ts`
  reproduce on `feat/maic-editor-v0` (base) — pre-existing, unrelated.
- `pnpm check`: prettier clean.
- `pnpm lint`: 0 errors, 7 pre-existing warnings.
- e2e: drag-to-reorder, rename (menu + double-click + Esc), insert-before-first,
  text auto-focus, drag tracking precision (1 px cursor lock), mode swap
  (slide↔slide, slide↔interactive, scene-type-swap chrome stability).

## Out of scope / follow-ups
- `Editor/Canvas` vs `ScreenCanvas` unification — would let the canvas survive a
  mode swap without remount. Separate slide-renderer refactor.
- Right-side properties panel + bottom timeline strip — slots already wired in
  `StageGrid`, plug in once surface contracts grow.

## Test plan
- [ ] Open a multi-slide deck with at least one non-slide scene; toggle Pro — right
      cluster (settings pill, PRO, Download) shouldn't jump.
- [ ] Click between slide thumbs — chrome doesn't flicker. Click an interactive
      thumb — chrome stays mounted; canvas shows playback renderer + read-only
      badge.
- [ ] Drag the rail handle — handle glued to cursor, thumbs scale smoothly.
- [ ] Rename via the three-dot menu and again by double-clicking the title; Esc
      cancels.
- [ ] Insert a Text box via the floating toolbar — caret lands inside the new
      element, type immediately.
- [ ] Hover the gap between two thumbs — \`+\` pops out to the right, not buried
      under the active card's violet ring.


cosarah:
Review notes from manual verification:

1. Deleting the first slide and clicking Undo restores it as the second slide. The restore logic clamps `entry.index - 1` to 0 and then inserts after that anchor, so index 0 cannot be restored correctly.

2. `mode` is transient UI state, but it is not reset when entering a classroom through SPA navigation. Refresh starts in playback mode, while entering edit mode in one classroom, leaving it, and then opening another classroom can keep the app in `edit` mode.

3. Inserting a slide at the top updates the scene order, but the lecture/script data appears independent and is not reindexed or remapped. This can make the visible scene order diverge from the script order.

wyuc:
Thanks for the pass. #1 and #2 fixed in 9f4f0e0; pushing back on #3.

**#1 Undo restore at index 0** — the toast handler now prepends via `setScenes([entry.scene, ...live])` when `entry.index === 0`. The clamp-then-anchor path was silently shifting the restored slide to position 1. e2e added: delete Slide 1 → Undo → restored to position 1.

**#2 `mode` not reset on SPA classroom navigation** — `loadFromStorage` and the server-side load path both now set `mode: 'playback'` on every classroom load, matching the refresh-path semantics.

**#3 insert at top vs lecture/script ordering** — scene scripts (`scene.actions[]`) are stored per-scene by `scene.id` and survive insertion unchanged; the new blank scene starts with no actions, which is the intended "blank-insert hands content authoring to the user" model for Phase 1. The `outline.order ↔ scene.order` join is pre-existing (used by `generatingOutlines` on load) and post-insert still correctly reports "nothing pending". A proper outline↔scene rebind by id would be the right fix if we need it, but it goes beyond PR3a's scope. Happy to file a follow-up issue if you'd like.

xuyuanwei678 (approved):

Issue #596, opened 2026-05-25
[Feature]: 建议优化内网部署版本


ccc898:
### Problem or Motivation

在企业内网（未连接互联网）部署后，如果将网页版openmaic生成的课程导入，无法正常显示图片，无法正常使用小程序。内网上直接生成的课程无法通过链接分享给内网上的其他用户，数据只能保存本电脑的浏览器中，或者只能导出后发给其他用户。

### Proposed Solution

建议优化公网网页版生成的课程内容及导出方式，增加导出内网版，即将文件中的公网资源替换为内网中可以使用的资源。建议本地部署版增加服务器保存方式，方便直接分享链接给其他用户使用，也方便内网用户将所有课程汇总到一个企业培训网页。

### Area

Classroom generation

wyuc:
问题都收到了，后续可以通过打包下载在线资源的形式让离线使用体验更好，欢迎持续关注～

wyuc:
Update: the offline-export part of this request is now handled.

#613 (merged) inlines the external assets referenced by interactive scenes — KaTeX, Three.js (incl. `three/addons`), Tailwind CDN, Google Fonts, and images — into the exported HTML as `data:` URIs at export time, routed through the SSRF-guarded media proxy to bypass browser CORS. Exported `.maic.zip` classrooms and Resource Packs now play fully offline after import into an intranet/air-gapped instance, with no public-CDN calls at playback time. Assets that can't be fetched at export (e.g. CORS-locked image hosts) are reported rather than silently dropped. Classrooms exported before this change still reference CDNs and need re-exporting.

The other half of this issue — server-side persistence and shareable links so courses survive a browser change and can be distributed by URL — is a separate track and is not addressed here. That's covered by #582 / #585 / #237 / #225 / #214 / #53. Leaving this issue open for that part.

A follow-up to extend offline coverage to non-bundled ESM / relative module imports (via a Resource Pack file layout) is noted on #613.

ccc898:
是不是还没有发布到网页版中？我刚试了一下，在网页版生成课堂后，再导入到内网上，还是无法正常显示。内网上需要重新部署吗？

liuzhengcai:
部署后，已经创建好的课程，如何能发布给别人学习，我把课程链接发送到另一台机器，打不开

wyuc:
Second update on this issue. The offline-export half was answered earlier (#613). Here is
where the other half — server-side storage and sharing on an intranet — now stands.

[...]

PR #603, opened 2026-05-27
feat: vendor KaTeX + Three.js for intranet-friendly interactive HTML (#596)

wyuc:
## Summary
- Add `three@0.160.0` as a dep (keeping existing `katex@^0.16.33`); postinstall copies them to `public/vendor/{katex,three}/` (gitignored).
- `interactive-post-processor.ts` now injects `/vendor/katex/...` instead of `https://cdn.jsdelivr.net/.../katex/...`.
- `visualization3d-content/system.md` importmap example now points to `/vendor/three/...` instead of `https://unpkg.com/...`.
- `eslint.config.mjs`: add `public/vendor/**` to globalIgnores (third-party assets should not be linted).
- README + README-zh: short note on intranet/airgapped behavior.

## Why
Refs #596. Every interactive scene that contains math (KaTeX injected unconditionally) or a 3D visualization (Three.js via unpkg) currently requires public-internet egress at playback time. Intranet-deployed OpenMAIC instances cannot reach those CDNs, so the scenes render unstyled formulas / broken 3D widgets. After this PR, all such assets are served from the OpenMAIC instance itself at `/vendor/...`, populated automatically during `pnpm install`.

## Scope boundary
This PR only changes the *injection path* for newly generated HTML. It does NOT rewrite paths at Resource Pack export time, nor bundle vendor files into the exported ZIP. That is a planned follow-up (PR-B). Classrooms generated before this PR still reference jsdelivr URLs inside their stored scene `content.html`; they need regeneration to pick up the new paths.

## Test plan
- [x] New unit tests in `tests/generation/interactive-post-processor.test.ts` cover: KaTeX injection target (`/vendor/katex/...`), absence of any `cdn.jsdelivr.net` / `unpkg.com` URLs, no double-injection guard, plus LaTeX delimiter regression
- [x] `pnpm test` full suite — only the 9 pre-existing `tests/server/ssrf-guard.test.ts` DNS-dependent failures remain; no new failures introduced
- [x] `pnpm lint` / `pnpm check` clean
- [ ] Manual: dev server, generate a math-bearing interactive scene, confirm KaTeX loads from `/vendor/katex/` in DevTools Network
- [ ] Manual: generate a 3D visualization scene, confirm Three.js loads from `/vendor/three/`


wyuc:
Updated approach: instead of editing the visualization3d prompt to instruct the LLM to use \`/vendor/three/\` (which depends on LLM adherence), I added a post-processor rewriter that catches LLM-written CDN URLs for vendored libraries and rewrites them to \`/vendor/...\` at generation time.

This covers both directions:
- LLM might keep writing \`unpkg.com/three@.../...\` per its training prior — rewriter normalizes.
- LLM might write \`cdn.jsdelivr.net/.../katex@.../...\` in HTML directly — rewriter normalizes (the old injector's substring guard previously let these through unchanged).

The prompt-template edit is therefore redundant and was reverted to keep the prompt aligned with the actual mechanism and to avoid LLM confusion.

Mechanism stack now:
1. Post-processor rewrites known-CDN URLs for vendored libs (KaTeX, Three.js) → \`/vendor/...\`
2. Post-processor injects KaTeX assets if no \`katex\` substring present (unchanged from current \`main\`-branch behavior, only the injected paths changed)

Tests added in \`tests/generation/interactive-post-processor.test.ts\`: 11 new cases for the rewriter (unpkg + jsdelivr variants, version-agnostic patterns, addons path-suffix preservation, unrelated-URL preservation, non-vendored artifact left alone) plus 2 integration tests for the full \`postProcessInteractiveHtml\` flow.

wyuc:
Superseded by #613 (export-layer asset inlining).

The vendoring approach here was heavier (committed static libs + postinstall + a per-library CDN rewriter) and only covered libraries we pre-vendored — it missed Tailwind CDN, Google Fonts, and remote/OSS images, which are exactly the assets that break offline in practice. #613 instead inlines *any* external asset referenced by interactive-scene HTML into `data:` URIs at export time (routed through the SSRF-guarded `/api/proxy-media` to bypass browser CORS), so the exported `.maic.zip` / Resource Pack is fully self-contained with no committed static-lib footprint. Closing in favor of that.

PR #611, opened 2026-05-28
feat: add support for TXT and DOCX file uploads

Kevin7012:
## Summary
  - Add `lib/document-parser.ts` with parsers for TXT, DOCX, and PDF
  - Install `mammoth` library for DOCX text extraction
  - Update toolbar UI to accept .pdf, .txt, .docx files
  - Update API route to handle all three document types
  - Update UI text from "PDF" to "Document" in all 6 languages
  - Fix duplicate scene key in outlines visualizer

  ## Test plan
  1. Upload a PDF file and verify parsing still works
  2. Upload a TXT file and verify parsing works
  3. Upload a DOCX file and verify parsing works
  4. Verify UI labels correctly show "Document" instead of "PDF"


wyuc:
Closing as superseded by the document-extraction work that has already landed in #741 and #844. Current `main` now has the document extractor boundary, TXT/Markdown handling, MinerU-backed Office/image formats, the `/api/extract-document` route, and the M3 multi-file bundle UI. This older branch also conflicts with that implementation and includes unrelated generated/report/package changes, so it is no longer a viable merge path. Thank you for helping establish the need for non-PDF uploads.

PR #612, opened 2026-05-28
feat(maic-editor): gate slide scene creation until inserted scenes are playable

wyuc:
## Why

Editor-created slide scenes (blank insert + duplicate) ship without playback `actions`. The playback engine treats an action-less scene as zero-dwell and skips straight past it, so a freshly inserted slide is effectively unplayable. Seeding default actions on new scenes is a separate change; until that lands, this gates the two scene-creation entry points so the editor stays coherent as an in-place "fine-tune the AI-generated deck" tool for the MVP release.

## What

- Add `lib/edit/scene-creation-enabled.ts` exporting `SCENE_CREATION_ENABLED = false`.
- Hide the inter-thumb `+` insertion zones in `SlideNavRail`.
- Hide the per-slide Duplicate menu item in `ThumbItem`.
- Keep reorder / delete / rename. These are playback-safe: scene `actions` are per-scene and id-bound, so reordering or removing scenes never produces an unplayable scene.

The insert/duplicate handlers and `createBlankSlideScene` / `duplicateSlideScene` helpers are untouched and still referenced behind the flag, so re-enabling is a one-line flip once new scenes are seeded with default actions.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `eslint` clean on changed files
- [x] `prettier --check` clean
- [x] `vitest run` edit suite green (480 passed; the only local failures are environment-dependent `ssrf-guard` DNS tests, unrelated to this diff)
- [x] e2e (`e2e/tests/slide-scene-creation-gate.spec.ts`): generates a classroom via the mocked pipeline, enters Pro mode, asserts no `+` insertion zones and that the per-slide overflow menu has only Rename + Delete (no Duplicate). Verified as a real guard by flipping `SCENE_CREATION_ENABLED` on → the e2e fails (4 insert zones, 3 menu items), then flipping back.

Target branch: `feat/maic-editor-v0`

### DIFF
--- a/components/edit/SlideNavRail/InsertionZone.tsx
+++ b/components/edit/SlideNavRail/InsertionZone.tsx
@@ -30,6 +30,7 @@ export function InsertionZone({ label, onInsert }: InsertionZoneProps) {
         onClick={onInsert}
         aria-label={label}
         title={label}
+        data-testid="slide-nav-insert"
         className="absolute inset-0 z-10 outline-none focus-visible:opacity-100"
       >
         <span className="sr-only">{label}</span>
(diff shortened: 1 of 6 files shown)


cosarah (approved):
Approved. Manually verified the editor gate behavior: insert and duplicate are hidden in Pro mode, while rename, delete, reorder, and playback of existing scenes continue to work.

Issue #398, opened 2026-04-11
[RFC] SSRF hardening: redirect re-validation, second-hop protection, IP pinning


YizukiAme:
## Background

PR #386 fixed the primary SSRF DNS rebinding bypass. During follow-up analysis, we identified a few remaining hardening opportunities in the server-side outbound HTTP layer.

### 1. Redirect Re-validation (Low-Medium)
Several SSRF-guarded routes follow HTTP redirects by default. A validated initial URL could redirect to an internal address that is never re-checked. Requires an authenticated user to intentionally enter a malicious URL, so practical impact is limited to multi-tenant deployments.

### 2. Second-Hop URL Validation (Low)
Some server-side code fetches URLs returned in provider API responses (e.g. download links for generated media). These second-hop URLs are not validated against SSRF rules. Would require a compromised AI provider to exploit.

### 3. TOCTOU / IP Pinning (Low)
DNS is resolved during validation but re-resolved independently at fetch time, leaving a theoretical race window. Exploitation requires millisecond-precision DNS manipulation — extremely difficult in practice.

### 4. ISATAP Tunnel Coverage (Informational)
A rare IPv6 tunnel format is not currently detected by `isPrivateIP()`. Deprecated protocol, negligible practical risk.

## Proposed Solution

Create a unified `outboundFetch()` server-side HTTP client that:
- Resolves DNS once and pins the validated IP for the actual connection
- Handles redirects manually with per-hop re-validation
- Supports two trust modes: strict (user-supplied URLs) and permissive (server-configured, e.g. Ollama)
- Falls back to alternate resolved IPs on connection failure

Implementation is ready locally and passing all tests.

A lighter alternative: simply add `redirect: 'manual'` to all affected routes without building the full transport layer.

## Proposed PR Roadmap (if full approach is desired)

| PR | Scope | Independent Value |
|----|-------|-------------------|
| **A** | Transport core + AI SDK injection | LLM API calls protected |
| **B** | Direct-fetch route migration | All user-supplied-URL routes protected |
| **C** | Second-hop + adapter layer | Provider-returned URLs protected |

Each PR is independently mergeable.

## Questions for Maintainers

1. **Is this level of hardening desired**, or would the lightweight fix (`redirect: 'manual'` on affected routes) be sufficient?
2. **Is replacing `fetch()` with a unified transport across server routes acceptable?**

## Related
- Original vulnerability: #378
- DNS rebinding fix (merged): #386

[...]

PR #613, opened 2026-05-28
feat: inline external assets at export for offline classrooms (#596)

wyuc:
## Summary
Interactive scenes reference external CDN assets (KaTeX, Three.js, Tailwind CDN, Google Fonts, images). After exporting a classroom and importing it into an offline/intranet instance, those CDN refs break. This PR inlines every external asset referenced by interactive-scene HTML into `data:` URIs **at export time** (when the network is available), so the exported course is fully self-contained.

## Approach
- New `lib/export/inline-assets.ts` (+ `inline-assets-importmap.ts`, `inline-assets-shared.ts`): scan HTML for external refs (`<link>`, `<script>`, ``, `<source>`, `<style>`/inline `url()`, and `<script type=importmap>` entries), fetch each (cached, deduped), rewrite to `data:` URIs.
- Stylesheets get nested `url(...)` fonts/images recursively inlined (KaTeX fonts, Google Fonts → gstatic).
- ES-module importmap entries become `data:text/javascript` URIs. The `three/addons/` prefix mapping is resolved by walking the module-import graph in the scene's module scripts and emitting explicit full-specifier `data:` entries (a `data:` URI cannot be a directory prefix). Verified that `data:` importmap targets work for transitive ESM imports with network blocked.
- Wired into both `.maic.zip` export (`use-export-classroom.ts`) and Resource Pack export (`use-export-pptx.ts`).
- Assets that can't be fetched (CORS-restricted hosts, 404) are reported and left as original URLs (online still works; offline degrades for just that asset).

## Scope / non-goals
- Import-side fetching of old ZIPs is out of scope: classrooms exported before this change must be re-exported. (Old ZIPs can't be fixed at import time without network, in which case re-export is simpler.)
- `content.url`-only scenes (whole-iframe remote page, no `content.html`) are not inlined.
- No server-side fetch proxy: CORS-blocked assets remain external.
- Generation/runtime is unchanged — this is purely an export-layer transform.

## Test plan
- [x] Unit: `tests/export/inline-assets.test.ts`, `tests/export/inline-assets-importmap.test.ts`, `tests/export/export-classroom-inline.test.ts` — scanner, fetcher/cache, CSS font recursion, orchestrator (links/scripts/img/css), importmap direct + `three/addons` graph closure (incl. nested + cycle), failure/CORS handling, dedup, namespace-safety, `.maic.zip` wiring.
- [x] `pnpm test` — no new failures vs baseline.
- [x] e2e offline simulation (network fully blocked) for KaTeX, Three.js+OrbitControls, Tailwind CDN, Google Fonts, remote image: all render from inlined output; all break from the original CDN-referencing HTML. (Screenshots captured locally.)

Supersedes the vendoring approach (committed static libs + postinstall + per-library rewriter), which only covered pre-vendored libraries; export-scan covers any CDN the model emits.


wyuc:
## Code review pass

Ran a multi-angle correctness review over the full diff. Summary of what it surfaced and how it was resolved.

### Fixed
- **Import-map prefix fallback.** Expanding `three/addons/` into explicit per-module `data:` entries no longer deletes the original prefix key. A sub-path imported dynamically (and therefore not seen during static analysis) would otherwise lose its only mapping and fail to resolve — even online. Explicit specifier entries shadow the prefix per the import-map spec, so keeping both is safe; the statically-imported modules still load offline from their `data:` entries.
- **Proxy error classification.** `/api/proxy-media` now forwards an upstream 4xx as a 4xx and collapses only 5xx to 502. Previously every upstream non-2xx became 502, which the client-side fetcher treats as transient and retries 3× — wasting time on permanently-missing assets.
- **Proxy response size cap.** Added a 25 MiB guard (Content-Length check plus a post-blob check) so the proxy can't buffer an unbounded upstream response into server memory.
- **Minor.** Re-export `FetchAsset` from the module barrel; `resolveSpecifier` uses `spec in imports` so an explicitly-mapped empty-string specifier isn't mistaken for unmapped.

### Known limitations (documented, not blocking)
- Modules fetched from **non-bundled ESM CDNs** (e.g. esm.sh-style split packages) whose internal imports are relative or resolve outside the page import-map are not recursively inlined. The common case is fully handled: three.js ships pre-bundled module builds, and its addons import only the bare `three` specifier.
- Following redirects re-validates **each hop** through the SSRF guard, which closes redirect-to-internal. The validate-then-fetch DNS window (rebinding) is a property of the shared guard rather than this change; full IP-pinning is tracked in #398.

### Verification
- **Unit:** reference scanner, cached fetcher, recursive CSS `url()`/font inlining, import-map graph closure, woff2 preference, proxy-media redirect + per-hop SSRF, and both export wirings.
- **Offline e2e (all network blocked):** KaTeX, Three.js + OrbitControls, Tailwind, Google Fonts, and a remote image all render from the inlined output; all break from the original CDN-referencing HTML.
- **Real-browser CORS e2e:** a direct `fetch('https://cdn.tailwindcss.com')` fails with a CORS error, while the same asset via `/api/proxy-media` succeeds. This guards the class of bug that Node-based tests can't catch, since Node's `fetch` ignores CORS.

cosarah (commented):
I tested the current behavior and it does successfully convert some external assets to data URIs. A few remaining URLs such as `http://www.w3.org/2000/svg` are XML namespace identifiers rather than fetchable assets, so keeping those is fine.

A few offline-export concerns remain:

1. CSS nested asset failures are not reported. The PR inlines `url(...)` inside downloaded CSS, such as fonts and background images, but if one of those nested assets fails to fetch, it is left as a remote URL without being added to `report.failed`. The export can look fully successful while still depending on network access.

2. External module scripts are not recursively bundled. A `<script type="module" src="https://.../app.js">` is converted to a `data:` URL, but imports inside that script, such as `import "./chunk.js"` or `import "pkg"`, are not followed. This can break offline execution.

3. Importmap recursion misses relative imports inside fetched modules. The importmap path can recursively inline mapped specifiers, but if a fetched module imports `./helper.js`, that relative dependency is skipped because it is not resolved against the fetched module’s original URL.

Suggested design improvement: instead of embedding all interactive dependencies as `data:` URLs inside `manifest.json`, consider exporting them as separate ZIP entries, for example under `interactive-assets/` or per-scene asset folders, then rewriting HTML/CSS/importmap/module references to relative paths. This would keep `manifest.json` readable, avoid large base64 payloads in JSON, and provide a real file base for resolving nested relative module imports offline.

wyuc:
Thanks, this was a thorough pass. Going point by point.

**1. CSS-nested asset failures now reported** — fixed in `2cb4e22`. `inlineCssUrls` returns the failed nested URLs and both call sites merge them into `report.failed`, so a font or background that can't be fetched no longer hides behind a successful-looking export.

**2 & 3. Module-graph gaps** — both real, thanks for catching them. An external `<script type="module" src>` isn't recursed into, and a relative `./helper.js` inside a fetched module isn't resolved against that module's own URL.

I checked how often these actually fire against the interactive scenes our generator produces (18 scenes across 3 real courses): external module scripts 0/18, relative imports inside fetched modules 0/18. It's structural, not luck. The prompt steers the model to inline its module code rather than reference an external entry, and the only common importmap library is three.js, which ships a bundled single-file module whose addons import the bare `three` specifier. So neither pattern shows up in practice. I'd rather document these as known limitations than ship a fragile in-browser module bundler that the real workload never exercises.

**4. ZIP entries vs data: URIs** — agreed for the Resource Pack export, and I'd frame the benefit slightly differently: with a real file base, relative `import` / `url()` / importmap references resolve natively, so #2 and #3 disappear without any source rewriting at all. That makes it the right place to solve them, as a follow-up.

The `.maic.zip` path can't get that file base, though. On import its scene HTML is stored as a string in IndexedDB and rendered via an iframe `srcDoc`, whose base URI is the host page, so relative paths resolve against the OpenMAIC server rather than any bundled file. Switching to `blob:` doesn't help (no directory either), and a service-worker virtual FS is far heavier than this feature warrants. Self-contained `data:` URIs are what let that path work with no changes to the import pipeline or renderer. So my proposal is: keep `.maic.zip` on data: URIs, move the Resource Pack to a file layout separately, and fold the module-graph fixes into that. Does that split work for you?

cosarah (approved):
Looks good to me. The CSS-nested asset failure reporting issue is fixed, and I agree the remaining module-graph limitations are better handled as a follow-up with a Resource Pack file-layout approach rather than blocking this PR.

PR #622, opened 2026-05-30
docs: add Fumadocs documentation site

wyuc:
## Summary

Adds a standalone, statically-exportable documentation site for OpenMAIC under `packages/docs`, built with [Fumadocs](https://fumadocs.dev) (Next.js + Tailwind). Intended to be served under `/docs`.

## What's included

- **6 locales** with full i18n: English (default), 简体中文, 繁體中文, 日本語, Русский, العربية — Arabic rendered RTL.
- **URL scheme**: English is unprefixed (`/docs/getting-started`) to preserve existing links; other locales live at `/docs/<lang>/<slug>`. Prefixed English paths (`/docs/en/...`) redirect to the canonical unprefixed URLs.
- **Static export** (`output: 'export'`) so it can be hosted as plain static files behind any web server or CDN — no Node runtime required.
- **Static full-text search** (Orama, build-time index) that works on the exported static site.
- **Brand theming** via CSS variables, plus a "Live Demo" link back to open.maic.chat in the sidebar nav.
- Content: Getting Started, Configuration, Supported Models, Deployment, VoxCPM.

## Kept isolated from the main app

So it doesn't touch the main app's build, lockfile, or CI:

- excluded from the pnpm workspace (own lockfile, build and deploy);
- ignored in `.prettierignore` (own toolchain, like the other packages);
- excluded from the root `tsconfig.json`. ESLint already ignores `packages/**`.

## Build & verify

```bash
cd packages/docs
pnpm install
pnpm build   # next build + static export -> out/
```

Produces static HTML for all locales plus the search index. The production build is clean (no hydration/console errors); RTL, locale switching, and static search were verified.

## Out of scope

Deployment (serving the static `out/` at `open.maic.chat/docs`) is handled downstream and is intentionally not part of this PR.


wyuc:
## Code review (CR loop)

Ran an iterative review → fix → re-review loop until no actionable findings remained.

**Round 1 — fixed**
- **CJK search returned nothing.** zh-CN / zh-TW / ja used the English (whitespace) tokenizer, which indexes space-less CJK text as a single token. Now they use the mandarin / japanese tokenizers (`@orama/tokenizers`) on **both** the build-time index (`app/static.json/route.ts`) and the client query (`components/search-dialog.tsx`), so the index and queries agree.
- **Locale config was duplicated across four files** (drift-prone). Consolidated the language list, default locale, per-locale tokenizer, RTL set, labels, and the `/docs` basePath into a single source `lib/locales.mjs`, consumed by i18n, the search index/dialog, `next.config.mjs`, and the postexport script.
- Removed dead `gitConfig.branch`.

**Round 2 — fixed**
- Refreshed the stale README route table (it listed scaffold routes that don't exist).
- Removed the remaining hand-maintained locale lists: the language-switcher labels and the postexport sanity list now derive from the single source.

**Round 3 — clean.** `tsc --noEmit` and the static build pass; no new findings.

**Known limitation (documented, deferred).** The English (default, unprefixed) route is a single `[slug]` segment to avoid a hydration conflict, so nested English pages would 404; other locales (catch-all) support nesting. See README → Known limitations.

**Verification.** `pnpm build` (static export) is clean; Chinese search returns results (e.g. “模型” → 13 hits); English search unchanged; all six locales render, Arabic RTL.

cosarah (approved):
Verified the static docs export locally. Routes, locales, and search look good.

PR #623, opened 2026-05-30
feat: support custom MiniMax voice_id

emmett001:
## Feature

Add support for custom MiniMax voice_id in MiniMax TTS settings.

## Changes

- Added custom Voice ID input field
- Persist voice_id configuration
- Inject voice_setting.voice_id into MiniMax TTS request body
- Support custom voice during TTS testing and Agent playback

## Example

ttv-voice-2026053021002326-CYyvUlBU

### DIFF
--- a/app/generation-preview/page.tsx
+++ b/app/generation-preview/page.tsx
@@ -878,7 +878,9 @@ function GenerationPreviewContent() {
                   language: languageDirective,
                 })),
               }
-            : undefined;
+            : settings.ttsProviderId === 'minimax-tts'
+              ? (ttsProviderConfig?.providerOptions || undefined)
+              : undefined;
         const speechActions = (data.scene.actions || []).filter(
           (a: { type: string; text?: string }) => a.type === 'speech' && a.text,
         );
--- a/app/generation-preview/page.tsx
+++ b/app/generation-preview/page.tsx
@@ -879,7 +879,7 @@ function GenerationPreviewContent() {
                 })),
               }
             : settings.ttsProviderId === 'minimax-tts'
-              ? (ttsProviderConfig?.providerOptions || undefined)
+              ? ttsProviderConfig?.providerOptions || undefined
               : undefined;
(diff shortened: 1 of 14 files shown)


emmett001:
This PR adds support for custom MiniMax voice_id configuration through provider options, allowing users to specify a custom voice when using MiniMax TTS.

wyuc (changes requested):
Custom MiniMax voice IDs are useful, but the current implementation introduces a second voice source that silently overrides the existing one.

1. `generateMiniMaxTTS()` always prefers `providerOptions.voiceId` over `config.voice`. Once a user enters a custom ID, later choosing a built-in MiniMax voice—or using an agent-specific `voiceConfig.voiceId`—has no effect until the separate custom field is manually cleared. The request/log/UI still present `ttsVoice` as selected while another hidden value wins. Please keep one authoritative voice value (for example an editable/creatable voice selector that writes `ttsVoice`), or define explicit override state and make the UI/agent behavior unambiguous.
2. The branch manually adds MiniMax `providerOptions` plumbing to generation preview, settings preview, scene narration, discussion TTS, and two agent preview paths. Current `main` has since introduced `resolveAgentVoiceOptions()` as the single source of truth specifically to prevent these paths drifting. Please rebase and integrate the base provider options once through the current resolver/preview abstraction rather than reintroducing six provider-specific branches.
3. `app/page.tsx` only adds `suppressHydrationWarning` to the greeting and is unrelated to custom MiniMax voices. Please remove that change from this PR.
4. Add focused regression coverage proving the selected custom ID reaches `voice_setting.voice_id` for Settings preview, narration, discussion, and agent preview; prove clearing it restores the selected built-in/agent voice; and assert provider options survive the API request without changing other providers. Include one real MiniMax smoke using a valid custom voice ID and confirm playable audio.

The branch currently conflicts with `main`; please rebase onto the current TTS/voice-resolver architecture and run refreshed CI before re-review.

wyuc:

PR #624, opened 2026-05-30
Make server-configured providers admin-managed, not client-overridable (#620)

### LINKED ISSUE #620: Simplify server vs. client provider config: lock server-configured providers instead of allowing client override
## Problem

The relationship between server-configured and client-configured providers is currently a tri-state — server API key, client base URL, client API key — where the client can partially echo or override the server's config. This is the root cause of a recurring bug class: when the settings UI sends back a base URL with an empty key, routes that treat "any base URL means fully client-configured" drop the server key and fail with "missing API key". #533 patches this for ~10 routes by adding a `canUseServerApiKeyForBaseUrl` guard, but it's treating a symptom of an over-flexible design.

## Proposal

Make a server-configured provider **admin-managed and not overridable** from the client:

- If a provider is server-configured, the client cannot change its key or base URL. The settings UI shows it as read-only ("managed by server").
- A user who wants their own credentials for the same vendor **adds a separate provider** (custom-provider support already exists), rather than overriding the managed one.

## Benefits

- Deletes the entire bug class: the client never sends config for a managed provider, so the `canUseServerApiKeyForBaseUrl` guard and the edge cases around it (e.g. server key configured without an explicit base URL) become unnecessary.
- No need to expose the server's base URL to the client at all (see below).
- Clear mental model: "managed by admin" vs. "my own provider", with no confusing half-states.

## Stop exposing server base URLs to the client

Today `/api/server-providers` (`getServerProviders()` and the TTS/ASR/PDF/image/video/web-search variants) returns each server-configured provider's **base URL** to the frontend, stored as `serverBaseUrl`. The API key is already withheld, but the base URL can still reveal internal infrastructure (a private gateway/proxy endpoint). If managed providers aren't client-overridable, the client doesn't need the base URL — the endpoint should return only a "managed" flag (and the allowed model list), not the URL.

## Tradeoff

This removes one existing flow: overriding a server-configured provider's key with your own for the *same* provider (e.g. a shared/rate-limited server key that a power user replaces with their own). Under this design they would add a separate provider instead. This seems an acceptable simplification, but flagging it explicitly.

## Scope

- Settings UI: render managed providers read-only; drop the override key/base-URL inputs and region toggles for them.
- `fetchServerProviders` / settings store: stop populating `serverBaseUrl`; keep an `isServerConfigured` flag only.
- `/api/server-providers`: return a managed flag + allowed models, not base URLs.
- Provider routes (~10) + `resolveModel`: managed providers resolve key and base URL purely server-side; ignore client-sent overrides. `canUseServerApiKeyForBaseUrl` can then be removed.

## Relationship to #533

#533 is the correct tactical fix for the current design and can merge independently. This issue proposes the structural change that would make that guard unnecessary going forward.

wyuc:
Closes #620.

## What

A server-configured provider is now **authoritative**: the operator owns its API key and base URL, and the client can no longer override either. This collapses the old tri-state (server key / client base URL / client key) that was the root of a recurring "missing API key" bug class — when the settings UI echoed a base URL with an empty key, routes that treated "any base URL ⇒ fully client-configured" dropped the server key. #533 patched that per-route with `canUseServerApiKeyForBaseUrl`; this removes the guard entirely by fixing the design.

## Server

- `resolve*ApiKey/BaseUrl` ignore client values when the provider is server-configured; new `isServerConfiguredProvider(section, id)`; `canUseServerApiKeyForBaseUrl` + `normalizeProviderBaseUrl` deleted.
- `getServer*Providers()` no longer return base URLs — only the allowed model list (LLM/image) and presence (the "managed" flag) — so an internal gateway/proxy endpoint is never exposed to the client.
- Routes (chat / image / video / tts / asr / pdf + verify-*) resolve managed providers purely server-side; SSRF validation now applies only to unmanaged (client-supplied) base URLs.

## Client

- `serverBaseUrl` dropped from `ProviderSettings`, the settings store, request-building (TTS base URL no longer echoes the server URL) and voice resolution; only `isServerConfigured` / `serverModels` remain.
- Settings UI renders managed providers read-only: the key / base-URL override inputs and region toggles are hidden, leaving a "managed by server" notice (copy updated across 7 locales).
- When the operator pins an allowed model list (`MODELS` ⇒ `serverModels`), the model catalog is locked too (read-only, no add/reset/edit/delete). Without a pinned list the server manages only credentials and the user keeps curating models. This removes the prior inconsistency where the panel let you add models that the generation toolbar then filtered out.

## Tradeoff

Overriding a server provider's key with your own for the *same* provider is removed; add a separate custom provider instead (as flagged in the issue).

## Out of scope / follow-ups

- The server does **not** validate a request's model against `serverModels` — it remains a client-side soft restriction; hard enforcement would be a separate change.
- Image providers never wired `serverModels` to the client (pre-existing), so their catalog isn't locked here.

## Verification

`tsc --noEmit`, `vitest` (379), `prettier --check`, `eslint` (0 errors), i18n key alignment — all green. New `e2e/tests/managed-provider-620.spec.ts` (2 passing) covers managed (read-only, locked catalog) vs unmanaged (editable) providers; provider-config unit tests rewritten for the new contract.


wyuc:
## Self-review (cr loop) summary

Reviewed the full diff across two lenses — correctness/security and consistency/simplification. Core invariant verified: for a server-configured (managed) provider there is no path where a client-sent `apiKey`/`baseUrl` influences resolution — the `resolveSection*` short-circuit on server config, every route gates `clientApiKey`/`clientBaseUrl`, and SSRF validation runs only on genuinely client-supplied (unmanaged) URLs. No correctness or security bugs found.

One item was flagged and deliberately **not** changed: the `/api/web-search` route still validates a client base URL against the provider allowlist and returns 400 for an out-of-allowlist URL, even for a managed provider, rather than silently ignoring it. This is intentional defense-in-depth (fail loud on a malicious client base URL such as an internal address) and is covered by `tests/web-search/route.test.ts` ("rejects client-controlled base URLs outside the provider allowlist"). The managed-authority resolution itself is already enforced inside `resolveWebSearchApiKey/BaseUrl`, so a valid client value is ignored and the server's config wins; only malicious input is rejected loudly.

Remaining notes are non-actionable: the `managed ? undefined : …` idiom varies slightly between routes (both forms correct), `optionalOverride` copy is now only used in the keyless web-search branch, and model-catalog locking is LLM-only because the store only wires `serverModels` for the LLM `providers` map (pre-existing). Verification: `tsc`, `vitest` 379, `prettier`, `eslint` (0 errors), i18n alignment, e2e 2 passing.

cosarah (commented):
Review note: I see two follow-ups before merge.

1. Managed web-search providers still validate client-sent baseUrl before resolving server config. With the new managed-provider semantics, client key/baseUrl should be ignored for server-configured web-search providers, not rejected. This can break users with stale local baseUrl values.

2. Legacy localStorage state may still keep serverBaseUrl. fetchServerProviders() spreads existing provider config and no longer clears that removed field, so old clients can keep a server URL in persisted client state.

Suggested coverage: add regression tests for managed Bocha/Tavily ignoring client baseUrl, and for clearing legacy serverBaseUrl during server-provider sync.

Verified locally:
- pnpm test -- tests/server/provider-config.test.ts tests/store/settings-server-sync.test.ts tests/store/settings-validation.test.ts tests/server/web-search-config.test.ts tests/web-search/route.test.ts
- pnpm test:e2e -- managed-provider-620.spec.ts
- git diff --check origin/main...HEAD

wyuc:
Thanks @cosarah — both addressed in 6782c55.

1. **Managed web-search now ignores client base URL.** You're right: for a server-configured provider the client key/baseUrl should be dropped, not rejected. The route now gates them on `isServerConfiguredProvider('webSearch', id)` like the other routes, so a managed provider with a stale local base URL just uses the server config instead of 400-ing. The allowlist validation still runs for unmanaged providers, where the client base URL is actually used — I split `tests/web-search/route.test.ts` into "unmanaged rejects out-of-allowlist" and "managed ignores client baseUrl" to cover both. (This supersedes my earlier note defending the rejection: I'd only weighed the SSRF angle and missed that the client URL is irrelevant for a managed provider.)

2. **Legacy `serverBaseUrl` cleared.** Bumped the persisted settings schema to v3 and strip `serverBaseUrl` from every provider config map in both `migrate` and `merge`, so old clients drop the stale field on rehydrate. Added a rehydrate regression test asserting it's removed while `isServerConfigured` is preserved.

Verification: full unit suite 381 pass, e2e 2 pass, tsc/prettier/eslint clean.

cosarah (approved):
Approved. No blocking issues found; latest changes address my previous concerns.

PR #629, opened 2026-05-31
fix(interactive): keep-alive iframe pool — interactive scenes no longer reload on remount (#619)

### LINKED ISSUE #619: [Bug]: Interactive scene iframe has no keep-alive — reloads on every remount (scene switch, mode toggle, re-parent)
## Summary

Interactive scene iframes reload (re-execute their HTML/scripts) whenever the playback chrome unmounts and remounts. The most visible trigger is toggling Pro mode (edit ↔ playback), but the root cause is structural and not specific to the editor: any path that tears down and rebuilds `PlaybackChromeRoot` re-mounts the iframe from scratch.

## Steps to reproduce

1. Open a classroom whose current scene is an interactive (iframe) scene.
2. Toggle Pro mode on, then off (or trigger any flow that unmounts/remounts the playback chrome).
3. Observe the interactive scene reload: the iframe re-parses its HTML and re-runs its scripts, losing any in-iframe state.

## Root cause

`Stage` switches between `EditChromeRoot` and `PlaybackChromeRoot` via `AnimatePresence` as two mutually exclusive branches keyed by mode. Entering Pro mode unmounts the whole `PlaybackChromeRoot` subtree; exiting rebuilds it.

```
Stage (mode swap via AnimatePresence)
  └─ PlaybackChromeRoot      ← unmounted on edit, rebuilt on return
       └─ CanvasArea
            └─ SceneRenderer
                 └─ InteractiveRenderer  ← <iframe srcDoc> mounts fresh
```

`InteractiveRenderer` renders `<iframe srcDoc={patchedHtml}>` directly in the React tree (`components/scene-renderers/interactive-renderer.tsx`). There is no iframe reuse/persistence (e.g. portaling a single iframe into a stable host node), so a remount means a full reload.

This is pre-existing behavior of the playback chrome. The editor work merely adds a prominent new way to trigger it.

## Impact

- Interactive widget state is lost on every Pro mode toggle.
- Visible reload flash; wasted work re-parsing/re-running iframe content.
- Affects any future remount trigger, not just Pro mode.

## Candidate fixes (need a direction decision)

1. Persist the iframe: portal one iframe instance into a stable DOM host outside the mode-swap subtree, show/hide instead of unmount.
2. Keep `PlaybackChromeRoot` mounted across mode swaps (hidden, not unmounted) so its iframe survives.
3. Cache/restore iframe state at the widget messaging layer (`widget-iframe` store) — partial, only covers state the widget cooperates on.

Option 1 or 2 fixes it generally; both touch the playback chrome architecture, so this should land as its own change rather than riding along with the slide-editor PR.

## Notes

Out of scope for the MAIC Editor slide-surface PR (#615) — filing separately so that PR stays focused.

wyuc:
Fixes #619.

## Problem
Interactive scene iframes reload (re-parse HTML, re-run scripts, lose in-iframe state) on **every remount**, not just the Pro mode toggle. `InteractiveRenderer` rendered `<iframe srcDoc>` inline in the scene subtree, so any unmount — or even a DOM re-parent — makes the browser drop the document. Triggers: Pro mode toggle (PlaybackChromeRoot unmount), scene switch and back, any PlaybackChromeRoot remount.

## Fix — keep-alive iframe pool
Persist the iframes in a stable host at the `Stage` root, outside the mode-swap / scene subtree, keyed by sceneId.

- **`lib/store/interactive-iframe-pool.ts`** — per-scene entries (`srcDoc`/`src`, content hash, tracked rect, visibility), LRU-bounded (`CAP=3`). Re-mounting with an unchanged hash keeps the existing content reference, so the iframe is never re-set; only a content-hash change rebuilds it (the single intended reload path).
- **`InteractiveIframeHost`** — mounted once in `Stage`, portals one persisted `<iframe>` per entry to `document.body`, positioned over each scene's rect via `position: fixed`. Hidden with `visibility` (never `display:none`, which can drop the document on re-show) in edit mode and whenever the placeholder is gone. Owns the `widget-iframe` postMessage registration.
- **`InteractiveRenderer`** — now a placeholder: registers content, marks the scene active/visible, reports its rect via a rAF loop (same mechanism as `useTrackedRect`), and hides (not evicts) on unmount.

Mode toggle, back-and-forth scene switching, and re-renders now hit the cache with zero reload and zero flash.

## Coupling
Self-contained, interactive-scene path only. `widget-iframe` messaging API unchanged (`PlaybackChromeRoot.getSendMessage` untouched). No change to the playback engine / SSE / edit-lock.

## Known minor (v1)
Brief content pop on the edit→playback fade-in (iframe shows at full opacity while the edit chrome fades out). Acceptable for v1; can refine with an animation-aware visibility gate later.

## Tests
- `tests/store/interactive-iframe-pool.test.ts` — 13 tests: create, hash-change rebuild, hide-keeps-alive, rect tracking, active, LRU eviction (incl. never evicting the active scene, recency refresh). All green.
- `pnpm test`: 620 pass / 13 new. (The 9 `ssrf-guard` failures are pre-existing, DNS-dependent, unrelated.)
- `tsc --noEmit` and `eslint` clean on changed files.

## Verification still pending
Live visual e2e (Playwright screenshots of a real interactive classroom across Pro-mode toggle and scene-switch-back) not yet run — needs a dev server + seeded interactive content. Will attach once verified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,6 @@ eval/orchestration/results-answering/
 
 # e2e screenshot artifacts
 e2e/screenshots/
+
+# local e2e screenshots for #619 review
+e2e/__screenshots-619/
--- a/components/edit/PlaybackChromeRoot.tsx
+++ b/components/edit/PlaybackChromeRoot.tsx
@@ -439,8 +439,15 @@ export const PlaybackChromeRoot = forwardRef<PlaybackChromeRootHandle, PlaybackC
         engineRef.current.stop();
       }
 
-      // Get widget iframe messaging callback for interactive scenes (keyed by sceneId)
-      const widgetSendMessage = useWidgetIframeStore.getState().getSendMessage(currentScene.id);
+      // Widget iframe messaging callback for interactive scenes, resolved lazily
+      // at send time (keyed by sceneId). The interactive iframe now lives in the
+      // keep-alive host (#619), which registers its postMessage callback a commit
+      // after this engine is built — so resolving eagerly here would capture null
+      // on a scene's first visit and silently drop every widget action. Looking it
+      // up per-send always sees the live registration.
+      const sceneIdForWidget = currentScene.id;
+      const widgetSendMessage = (type: string, payload: Record<string, unknown>) =>
+        useWidgetIframeStore.getState().getSendMessage(sceneIdForWidget)?.(type, payload);
 
       // Create ActionEngine for playback (with audioPlayer for TTS and widget messaging)
       const actionEngine = new ActionEngine(
--- a/components/scene-renderers/InteractiveIframeHost.tsx
+++ b/components/scene-renderers/InteractiveIframeHost.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useEffect, useRef, type CSSProperties } from 'react';
+import { useEffect, useRef, useState, type CSSProperties } from 'react';
 import { createPortal } from 'react-dom';
 import { useStageStore } from '@/lib/store';
 import { useWidgetIframeStore } from '@/lib/store/widget-iframe';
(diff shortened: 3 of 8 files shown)


wyuc:
## Code-review loop

Two review rounds (multi-angle finders + verification). Converged; fixes pushed in `deff283`.

**Fixed:**
- **Widget actions dropped on first visit** — `PlaybackChromeRoot` captured the widget postMessage callback eagerly at engine construction, but the iframe (and its registration) now lives in the keep-alive host, which registers one commit later. Eager capture saw `null` on a scene's first visit and silently no-op'd every widget action. Now resolved lazily at send time.
- **Iframe vanished in fullscreen presentation** — `requestFullscreen` runs on the stage subtree, so a body-portaled iframe was outside it. Portal target now follows `document.fullscreenElement`.
- **Stale iframes across classroom switch** — pool now `reset()`s when the host unmounts.
- **0×0 flash before first layout** — only show once the rect has a real measured size.
- **Hash-collision could skip a rebuild** — dropped the djb2 hash; compare `srcDoc`/`src` by value (string `===` is value equality, so keep-alive still holds).

**Accepted tradeoffs / known minors (not fixed):**
- Entering/exiting fullscreen presentation re-parents the iframe and reloads it once. The only alternative (always body-portal) leaves the widget invisible during presentation, which is worse — a deliberate present-toggle reloading is acceptable where an incidental remount is not.
- Scene-rail `Interactive Preview` thumbnail (`ThumbnailInteractive`) still reloads on re-render — a separate, stateless surface, deliberately out of scope here (pending a call on whether to track it separately).

**Verified clean:** `vitest` 620 pass (+13 new; the 9 `ssrf-guard` failures are pre-existing/DNS-dependent), `tsc` + `eslint` clean, Playwright e2e green.

wyuc:
Thumbnail reload (separate, stateless surface) is now tracked in #630 — deliberately out of scope here. CI green; moving to ready.

cosarah (commented):
I can reproduce a regression when toggling Edit mode on an interactive scene. Scene switching works and the iframe state is preserved, but after entering Edit mode and returning to playback, the interactive scene briefly appears with the preserved state and then becomes a persistent blank white canvas.

This looks like a visibility race rather than an iframe reload. The placeholder calls show(sceneId) on mount and hide(sceneId) in cleanup, while visibility is keyed only by sceneId. With Stage's AnimatePresence cross-fade, an outgoing PlaybackChromeRoot cleanup can run after the new PlaybackChromeRoot has already mounted and shown the same scene, so the stale cleanup hides the pooled iframe. The iframe document stays alive, which matches the observed state flash followed by white screen.

I think the pool needs some ownership/versioning for the active placeholder mount so stale cleanup cannot hide a newer renderer instance for the same scene.

cosarah:
Update: I retried and this appears to be intermittent rather than consistently reproducible. The symptom still points to the same timing race: in the failing runs, the preserved iframe state flashes first, then the pooled iframe becomes hidden. It may be worth stress-testing fast Edit mode toggles / cross-fade cleanup ordering.

cosarah (approved):
Manually verified the latest owner-token update. Scene switching and Edit mode toggling preserve the interactive iframe state, and the intermittent blank-after-return issue no longer reproduces in my testing. The fullscreen behavior matches the intended tradeoff. LGTM.

PR #632, opened 2026-06-01
fix: hide unsupported MiniMax Hailuo Fast T2V model

### LINKED ISSUE #631: MiniMax Hailuo 2.3 Fast is exposed in the T2V selector but unsupported by the current adapter
OpenMAIC currently lists `MiniMax-Hailuo-2.3-Fast` in the MiniMax video model selector. However, the current MiniMax video adapter sends Text-to-Video requests.

Actual MiniMax API behavior:

`model MiniMax-Hailuo-2.3-Fast does not support Text-to-Video mode`

Hailuo 2.3 Fast requires Image-to-Video input with `first_frame_image`. Therefore, exposing only the Fast model ID in the current T2V flow is misleading: users may expect OpenMAIC to consume the Hailuo Fast quota, but the request fails before a video task is created.

Expected behavior:

- Remove or disable `MiniMax-Hailuo-2.3-Fast` in the current T2V-only selector, or
- Expose it only after OpenMAIC has a real I2V / first-frame workflow.

Related to #591.

wyuc:
## Summary

- Remove `MiniMax-Hailuo-2.3-Fast` from the current MiniMax Text-to-Video model list.
- Add a regression test to keep the T2V-only selector from exposing the unsupported Fast model.

## Why

The current MiniMax video adapter submits Text-to-Video requests. MiniMax Hailuo 2.3 Fast requires Image-to-Video input with `first_frame_image`, so selecting the Fast model in the current flow causes the API to reject the request before a video task is created.

This keeps the UI aligned with the adapter's actual capabilities until OpenMAIC has a dedicated I2V / first-frame workflow.

Fixes #631.

## Test Plan

- `pnpm test tests/media/minimax-video-provider.test.ts`
- `pnpm test tests/media/minimax-video-provider.test.ts tests/store/settings-server-sync.test.ts`
- `git diff --check`

### DIFF
--- a/lib/media/video-providers.ts
+++ b/lib/media/video-providers.ts
@@ -85,9 +85,10 @@ export const VIDEO_PROVIDERS: Record<VideoProviderId, VideoProviderConfig> = {
     name: 'MiniMax Video',
     requiresApiKey: true,
     defaultBaseUrl: 'https://api.minimaxi.com',
+    // Hailuo 2.3 Fast requires Image-to-Video with first_frame_image; this
+    // provider currently submits Text-to-Video requests only.
     models: [
       { id: 'MiniMax-Hailuo-2.3', name: 'Hailuo 2.3' },
-      { id: 'MiniMax-Hailuo-2.3-Fast', name: 'Hailuo 2.3 Fast' },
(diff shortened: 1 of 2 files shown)


cosarah (approved):
LGTM, verfied locally

PR #636, opened 2026-06-02
fix(pdf): override undici bodyTimeout for MinerU to prevent 5-min cutoff

hxf0223:
**问题:** 使用 MinerU 解析大型 PDF 时，Node.js 内置 fetch（undici）有默认的 bodyTimeout=300s（5 分钟），会在 AbortSignal.timeout() 生效前触发，导致请求被中断并出现 TypeError: fetch failed，MINERU_TIMEOUT_MS 配置无效。

**变更:** 在 pdf-providers.ts 中为 MinerU 上传请求创建一个 undici Agent（dispatcher），将 bodyTimeout 和 headersTimeout 设置为 MINERU_TIMEOUT_MS（默认 900000ms = 15 分钟），并将其传入 fetch。同时保留 AbortSignal.timeout() 作为二次保护。

**原因解释:** undici 的内部超时优先于外部 AbortSignal，因此必须修改 undici 的超时配置以匹配预期的最长处理时间。


wyuc (changes requested):
The MinerU timeout problem is legitimate—Undici's dispatcher defaults `headersTimeout` and `bodyTimeout` to 300 seconds, so a longer `AbortSignal.timeout()` alone does not extend those limits. The current branch needs cleanup before the fix can merge:

1. The final commit is the complete, unrelated #662 web-search/query-rewrite change (`app/api/web-search/route.ts`, `lib/ai/llm.ts`, and `lib/server/classroom-generation.ts`). Please remove it so this PR contains only the MinerU timeout fix. #662 also has separate review findings and must not hitchhike here.
2. The branch conflicts with current `main`, where the upload path was refactored from the old PDF-only `parseWithMinerU()` body into `parseWithMinerUDocument()` and now handles multiple document MIME types. Please rebase and apply the timeout at the current shared self-hosted MinerU request path so DOCX/PPTX/XLSX/image extraction receives the same protection.
3. The patch creates a new `UndiciAgent` for every parse and never closes or reuses it. Each Agent owns per-origin dispatchers/connections and exposes `close()`/`destroy()`; repeated uploads can accumulate idle pools/resources. Please use a stable reusable dispatcher (with a validated timeout) or close it in `finally` after the response body has been fully consumed.
4. Validate `MINERU_TIMEOUT_MS` rather than passing `NaN`, zero, or a negative value from raw `parseInt` into both the Agent and `AbortSignal.timeout()`. Add focused coverage for default/custom/invalid values and for dispatcher cleanup/reuse.

Once the branch is reduced to this concern, rebased onto current `main`, and the dispatcher lifecycle is fixed, happy to re-review.

wyuc:

PR #638, opened 2026-06-02
fix(export): guard toPoints() against arc-first SVG paths

### LINKED ISSUE #637: PPTX export crashes (TypeError) on shape paths that start with an arc (A) command
### Description

`toPoints()` in `lib/export/svg-path-parser.ts` throws a `TypeError` for any SVG `path` whose first command is an arc (`A`) — or whose leading commands push no anchor point. In the arc branch it reads `points[points.length - 1].type` without checking that a previous point exists; when `points` is empty, `lastPoint` is `undefined` and `lastPoint.type` throws.

`toPoints()` documents (and `getSvgPathRange()` implements) the contract that a malformed path should return `[]` so a single bad / LLM-hallucinated shape doesn't take down the export — this contract was added in #505. The arc-first case slips past it because `new SVGPathData(d)` parses such a path successfully; the crash happens later in the command loop, which #505's try/catch doesn't cover.

### Impact

The error propagates out of the per-element loop in `lib/export/use-export-pptx.ts` (which relies on `toPoints` returning `[]` and is not wrapped per-element), aborting the **entire "Export to PPTX"** for the classroom. A single shape element with an arc-first `path` is enough.

### Steps to reproduce

```ts
import { toPoints } from '@/lib/export/svg-path-parser';
toPoints('A 0.5 0.5 0 1 1 1 0.5');
// → TypeError: Cannot read properties of undefined (reading 'type')
```

### Environment

- OpenMAIC `main` (v0.2.1)

### Suggested fix

Guard the arc branch with `!lastPoint` (skip the arc when there is no preceding anchor point), completing the malformed-path contract from #505. PR to follow.

ly-wang19:
## Summary

`toPoints()` (`lib/export/svg-path-parser.ts`) threw a `TypeError` for SVG paths whose first command is an arc (`A`), because the arc branch read `points[points.length - 1].type` with an empty `points` array. The function documents that malformed paths must return `[]` (mirroring `getSvgPathRange`, contract added in #505) so one bad / LLM-hallucinated shape doesn't break export — but the arc-first case slipped through (`SVGPathData` parses it fine; the crash is later in the loop), **aborting the entire PPTX export**.

This adds a `!lastPoint` guard and a regression test.

## Related Issues

Closes #637
Follow-up to #505.

## Changes

- `lib/export/svg-path-parser.ts`: in the arc branch, skip the arc when there is no preceding anchor point instead of dereferencing `undefined`.
- `tests/export/svg-path-parser.test.ts`: regression test for an arc-first path.

## Type of Change

### Steps to reproduce / test

1. Before the fix: `toPoints('A 0.5 0.5 0 1 1 1 0.5')` throws `TypeError: Cannot read properties of undefined (reading 'type')`.
2. After the fix: it returns `[]`.
3. `pnpm vitest run tests/export/svg-path-parser.test.ts` → 4 passed (the new test fails on `main`, passes with the fix).

- TDD: the new test reproduces the crash on `main` and passes with the fix.
- The post-guard `lastPoint.x` / `lastPoint.y` reads remain valid (the guard guarantees a M/L/Q/C point).
- Existing `toPoints` / `getSvgPathRange` tests still pass.

### Evidence

- [x] CI passes (`pnpm check` ✓, `pnpm lint` ✓ 0 errors, `npx tsc --noEmit` ✓, `vitest` 4/4 ✓)
- [x] Manually tested locally

## Checklist

- [x] I have added/updated documentation as needed (N/A — code + test only)

---

🤖 **AI-assisted PR** (Claude). Self-reviewed the diff before submission.


wyuc (approved):
Verified the root cause and impact: `toPoints()` dereferences `points[points.length-1]` in the arc branch, which is `undefined` when an arc is the first command — `SVGPathData` parses it fine so #505's constructor-level try/catch doesn't catch the throw. It's called unguarded in the export loop (use-export-pptx.ts:590 and :675, no per-element catch), so one arc-first shape aborts the entire PPTX export.

The `!lastPoint` guard is the correct minimal fix: it short-circuits before the `.type` read, the post-guard whitelist guarantees `lastPoint.x/.y` are valid, and the Z-close case was already handled. Regression test mirrors the repro and follows TDD. CI green, mergeable.

LGTM 👍 (one tiny non-blocking nit: the comment says it 'returns []', but it actually skips just the unanchored arc and keeps the rest — which is the better behavior; wording could match.)

PR #642, opened 2026-06-02
fix(web-search): allow keyless Brave as a server-side default

### LINKED ISSUE #641: Keyless Brave Search can't be used as a server-side web search provider
Brave Search is `requiresApiKey: false` (`lib/web-search/constants.ts`) and `searchWithBrave` already falls back to **scraping Brave's public results page** when no API key is provided (`lib/web-search/brave.ts`). But the server-side wiring never lets keyless Brave activate or be selected:

1. `loadEnvSection` for web search (`lib/server/provider-config.ts`) is **not** given `keylessProviders`, so Brave is only loaded into the server config when `BRAVE_API_KEY` is set — which defeats keyless use. (The LLM / TTS / ASR / image sections all pass `keylessProviders` for their keyless providers like Ollama / Lemonade.)
2. `resolveServerWebSearchProviderId` only auto-selects providers that have an `apiKey`, so keyless Brave is never chosen.

**Net effect:** in the autonomous classroom-generation flow, `resolveClassroomWebSearchConfig` returns `undefined`, so web search is **silently disabled** for an operator who wants free, keyless Brave search.

### Suggested fix
Pass `keylessProviders: new Set(['brave'])` to the web-search `loadEnvSection` (activatable via `BRAVE_BASE_URL`, mirroring Ollama/Lemonade), add Brave to the `resolveServerWebSearchProviderId` fallback chain, and document `BRAVE_BASE_URL` in `.env.example`. PR to follow.

ly-wang19:
## Summary

Brave Search is `requiresApiKey: false` and `searchWithBrave` already scrapes Brave's public results page when no key is set — but the server-side wiring never let keyless Brave **activate** or be **selected**, so in the autonomous classroom-generation flow `resolveClassroomWebSearchConfig` returned `undefined` and web search was silently disabled for operators who wanted free, keyless Brave.

This mirrors the existing keyless handling for Ollama/Lemonade: set `BRAVE_BASE_URL` (e.g. `https://search.brave.com`) to enable keyless Brave, which then becomes a valid server default (after any keyed providers).

## Related Issues

Closes #641

## Changes

- `lib/server/provider-config.ts`: pass `keylessProviders: new Set(['brave'])` to the web-search `loadEnvSection` (so `BRAVE_BASE_URL` alone activates Brave), and add Brave to the `resolveServerWebSearchProviderId` fallback chain (after keyed providers).
- `.env.example`: document `BRAVE_BASE_URL` for keyless Brave.
- `tests/server/provider-config.test.ts`: tests that keyless Brave is selected via `BRAVE_BASE_URL`, and that a keyed provider still takes priority.

## Type of Change

### Steps to reproduce / test

1. With no web-search keys but `BRAVE_BASE_URL=https://search.brave.com`, the server now resolves `brave` and web search works (previously: `undefined`, disabled).
2. `pnpm vitest run tests/server/provider-config.test.ts` → 31 passed (2 new).

- New tests cover keyless selection + keyed-provider priority.
- Behavior-preserving for existing setups: Brave only activates when `BRAVE_BASE_URL` is explicitly set; keyed-provider resolution is unchanged.

### Evidence

- [x] CI passes (`pnpm check` ✓, `pnpm lint` ✓, `npx tsc --noEmit` ✓, `vitest` 31/31 ✓)

## Checklist

- [x] I have added/updated documentation as needed (`.env.example`)

---

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -190,6 +190,10 @@ TAVILY_API_KEY=
 BOCHA_API_KEY=
 BOCHA_BASE_URL=https://api.bocha.cn
 BRAVE_API_KEY=
+# Brave also works without a key by scraping its public results page. Set
+# BRAVE_BASE_URL (e.g. https://search.brave.com) to enable Brave as a keyless
+# server-side provider — no BRAVE_API_KEY required.
+BRAVE_BASE_URL=
 BAIDU_API_KEY=
 BAIDU_BASE_URL=https://qianfan.baidubce.com
 # Dedicated MiniMax web-search vars avoid conflicting with the LLM MINIMAX_* endpoint.
--- a/lib/server/provider-config.ts
+++ b/lib/server/provider-config.ts
@@ -308,7 +308,9 @@ function buildConfig(yamlData: YamlData): ServerConfig {
     pdf: loadEnvSection(PDF_ENV_MAP, yamlData.pdf, { requiresBaseUrl: true }),
     image,
     video: loadEnvSection(VIDEO_ENV_MAP, yamlData.video),
-    webSearch: loadEnvSection(WEB_SEARCH_ENV_MAP, yamlData['web-search']),
+    webSearch: loadEnvSection(WEB_SEARCH_ENV_MAP, yamlData['web-search'], {
+      keylessProviders: new Set(['brave']),
+    }),
     ttsDisabled: collectDisabledTTS(yamlData.tts),
   };
 }
@@ -566,5 +568,8 @@ export function resolveServerWebSearchProviderId(preferredProviderId?: string): 
   if (webSearch.bocha?.apiKey) return 'bocha';
   if (webSearch.baidu?.apiKey) return 'baidu';
   if (webSearch.minimax?.apiKey) return 'minimax';
+  // Brave needs no API key (it scrapes the public results page), so a configured
+  // keyless Brave is a valid fallback even though it has no apiKey.
+  if (webSearch.brave) return 'brave';
   return Object.keys(webSearch)[0];
 }
--- a/tests/server/provider-config.test.ts
+++ b/tests/server/provider-config.test.ts
@@ -315,6 +315,30 @@ providers:
     });
   });
 
+  describe('resolveServerWebSearchProviderId', () => {
+    it('selects keyless Brave when only BRAVE_BASE_URL is set (no API key)', async () => {
+      vi.stubEnv('BRAVE_BASE_URL', 'https://search.brave.com');
+      const {
+        resolveServerWebSearchProviderId,
+        getServerWebSearchProviders,
+        resolveWebSearchBaseUrl,
+      } = await import('@/lib/server/provider-config');
+      // Brave is activated as a keyless server provider via base URL alone. The
+      // exposed map shows only presence (managed flag), not the base URL (#624).
+      expect(getServerWebSearchProviders().brave).toEqual({});
+      expect(resolveWebSearchBaseUrl('brave')).toBe('https://search.brave.com');
+      // ...and Brave is then auto-selected as the server default.
+      expect(resolveServerWebSearchProviderId()).toBe('brave');
+    });
+
+    it('prefers a keyed provider over keyless Brave', async () => {
+      vi.stubEnv('BRAVE_BASE_URL', 'https://search.brave.com');
+      vi.stubEnv('TAVILY_API_KEY', 'tvly-key');
+      const { resolveServerWebSearchProviderId } = await import('@/lib/server/provider-config');
+      expect(resolveServerWebSearchProviderId()).toBe('tavily');
+    });
+  });
+
   describe('baseUrl-only providers (e.g. mineru)', () => {
     it('includes PDF provider from YAML when only baseUrl is configured (no apiKey)', async () => {
       yamlOverride = `


wyuc (dismissed):
LGTM on the logic — keyless Brave mirrors the Ollama/Lemonade pattern, keyed providers still win in `resolveServerWebSearchProviderId`, and the tests cover both cases. Confirmed it stays opt-in: `webSearch.brave` is only set when `BRAVE_BASE_URL` is configured, and the classroom flow is still gated by `enableWebSearch` (off by default), so no silent scraping.

Before it can land:
1. Branch conflicts with `main` — please rebase.
2. Keyless mode scrapes Brave's public page; please confirm you verified it returns results end-to-end.

ly-wang19:
Thanks @wyuc — both points addressed:

**1. Rebase** ✅ — rebased onto `main`; the `provider-config.ts` conflict (your `ttsDisabled` vs my keyless `webSearch`) is resolved keeping both. Now `MERGEABLE`, 43 tests green.

**2. Verify keyless returns results end-to-end** — I did, and it turned up a real problem: **it currently returns 0 results.** A live scrape of `search.brave.com` (app headers) is HTTP 200 with 20 `data-type="web"` snippets, but `parseBraveSearchHtml` extracts nothing — Brave moved the result title from `<span class="search-snippet-title">` to `<div class="… search-snippet-title …">`, so the title regex misses and every snippet is skipped. (The existing brave test stayed green because its fixture still used the old `<span>` markup.)

Fixed in **#688** (accept `<span>`/`<div>`, fixtures updated to current markup). With that fix the same scrape returns real results end-to-end, e.g. `Photosynthesis - Wikipedia -> en.wikipedia.org/wiki/Photosynthesis`.

So this PR should land **after / together with #688** — otherwise it enables a keyless provider that returns nothing.

One caveat worth a note: keyless scraping is **best-effort** — rapid repeat requests get rate-limited/challenged by Brave (some responses come back empty). The keyed Brave API path is unaffected. Happy to add a short "best-effort, keyed recommended" note to the docs/env example if you want it in this PR.

tongshu2023:
Heads-up on a merge-order interaction with #589 (SearXNG provider), which touches the same two spots in `provider-config.ts`:

1. **`loadEnvSection` opts** — #589 passes `keylessProviders: new Set(['searxng'])` at the same call site. Trivial union on whichever lands second: `new Set(['brave', 'searxng'])`.
2. **`resolveServerWebSearchProviderId` tail** — a real semantic difference, not just a textual conflict. #589 currently ends with `return Object.keys(webSearch).find((id) => webSearch[id]?.apiKey)` (keyless providers never auto-selected). Your `if (webSearch.brave) return 'brave'` is the better call given the `keylessProviders` gating: anything in the map got there via an explicit operator opt-in (`BRAVE_BASE_URL` / `SEARXNG_BASE_URL`), so treating it as a valid fallback default is right, and #589's blanket skip would silently swallow your keyless-Brave default.

No action needed on this PR — flagging so the rebase burden lands on #589, not here. Once this merges I'll rebase #589 to: union the keyless set, and follow this PR's pattern (`if (webSearch.brave) return 'brave'; if (webSearch.searxng) return 'searxng';`) instead of the apiKey-only fallback.

wyuc:
**Second-pass review after #688 landed — end-to-end confirmed working; three findings before merge**

First, verification: on current `main` (with #688), a live fetch of `search.brave.com/search?q=photosynthesis` using the app's headers parses 20 `data-type="web"` snippets into real results (`Photosynthesis - Wikipedia -> en.wikipedia.org/wiki/Photosynthesis`), so the keyless path now delivers end-to-end. A repeat request from a different egress got HTTP 429, which confirms the best-effort caveat — so yes, please do include the short "best-effort, keyed recommended" note you offered for `.env.example`.

Three findings from a second review pass, all verified against the code:

**🟠 P2 — Setting `BRAVE_BASE_URL` silently downgrades users' own Brave API keys to scraping**

Once the operator sets `BRAVE_BASE_URL`, Brave becomes server-managed: the web-search route drops the client-sent key (managed guard in `app/api/web-search/route.ts`), and `resolveSectionApiKey` returns the managed entry's `apiKey || ''` — empty by construction for a keyless entry. So a user who configured their own (paid) Brave Search API key in Settings silently goes from the official JSON API to regex-scraping the public page. The classroom path (`lib/server/web-search-config.ts` → `resolveWebSearchApiKey`) downgrades the same way. This is the first managed-with-empty-key case in `webSearch`, where the managed-is-authoritative rule masks a real client credential with an empty one.

Two ways to resolve — your pick:
- Minimal: document the override in `.env.example` — a line under `BRAVE_BASE_URL` noting it makes Brave server-managed, so client Brave keys are then ignored (scrape mode).
- Better: for a keyless-capable managed entry, fall back to the client key when the managed entry's key is empty. This doesn't reintroduce the #533 tri-state: `searchWithBraveApi` always talks to the fixed `api.search.brave.com`, so the client key never flows to an operator-controlled base URL.

**🟡 P3 — `resolveServerWebSearchProviderId(preferredProviderId)` rejects keyless Brave as an explicit preference while accepting it as a fallback**

The preference check requires a truthy `apiKey` (`webSearch[preferredProviderId]?.apiKey`), so with keyless Brave plus any keyed provider configured, `preferredProviderId: 'brave'` returns the keyed provider — contradicting the new fallback a few lines below. No caller passes a preference today, so it's latent, but it's this PR's own feature interacting with the exported API. A presence-based check for keyless providers fixes it.

**🟡 P3 — New tests can leak host env: `ENV_PREFIXES_TO_CLEAR` clears neither `BRAVE` nor `BAIDU`**

A dev/CI host with `BAIDU_API_KEY` set fails "selects keyless Brave when only BRAVE_BASE_URL is set" (`baidu` is checked before the keyless fallback); a host with `BRAVE_API_KEY` makes it pass for the wrong reason (keyed path instead of the keyless fallback under test). Adding both prefixes keeps host state out of the tests, per the file's own header comment.

All three are small. Once they're addressed (P2 can be consciously resolved as doc-only if you prefer), this is good to merge — and #589 will follow this PR's pattern per the earlier thread.

wyuc (changes requested):
Formalizing the three blockers from the second-pass review:

1. `BRAVE_BASE_URL` makes Brave server-managed with an empty managed key, so the route/classroom config ignores a user's own Brave API key and silently downgrades the official API path to best-effort page scraping. Please preserve the client key when the managed key is empty, or at minimum explicitly document this override and its 429/challenge limitations.
2. `resolveServerWebSearchProviderId(preferredProviderId)` requires a truthy API key, so an explicitly preferred keyless Brave is rejected even though the same config is accepted as the fallback. Make the preference check presence/configuration based for keyless providers.
3. The new provider-config tests do not clear `BRAVE_*` or `BAIDU_*`, so host environment variables can make the keyless test pass for the wrong reason or fail by selecting Baidu first. Add both prefixes to the test cleanup and verify the keyless path explicitly.

The feature is useful and the live scrape path works after #688, but these need to be addressed and the branch refreshed onto current `main` before merge.

wyuc:
Closing as superseded by #842, which has landed on `main` and now covers the requested keyless Brave server configuration/fallback together with the SearXNG provider work. The associated issue #641 has been closed as completed.

Issue #646, opened 2026-06-02
[MAIC Editor] Phase 2 — universal scene-DSL editor + AI editing (MAIC Agent)


wyuc:
## What MAIC Editor is (refined from Phase 1)

Phase 1 (#560) framed MAIC Editor as an editor for **static display components**, and drew a line between "static" and "interactive" scenes — putting quiz / PBL / live "out of contract." That axis was a **category error**: static-vs-interactive describes a scene's *runtime*, not whether its underlying object can be edited.

A MAIC scene is a **DSL object**. It carries more than displayed content:

- **content** — the slide canvas, the quiz's questions, an interactive widget's config, …
- **actions** — playback orchestration: narration (speech), spotlight, laser, whiteboard cues
- **structure / metadata** — order, whiteboards, multi-agent config

**MAIC Editor is the editing framework for the MAIC scene DSL.** Its reach is the whole object — every scene type, content *and* actions *and* structure. Nothing is excluded by runtime paradigm; there are only surfaces not yet built. "Slide editor" was always too narrow; "static-display editor" was the same narrowing one level up.

### What keeps "edit everything" tractable

Two load-bearing bounds carried from Phase 1 keep universality from becoming an unbounded scope:

1. **Bounded by the DSL.** "Everything" means *every field of the finite scene DSL*, not "any conceivable feature." The DSL is the contract and the boundary.
2. **The AI-first ceiling (Phase 1 principle #2 still holds).** The framework is universal *by architecture*; manual surfaces stay **minimal** and cover only what humans do better than AI. AI covers the rest. "Edit everything" is a statement about reach, not a commitment to a hand-built UI for every field.

### Two editing dimensions

- **Content surfaces** — per scene type, via the pluggable `SceneEditorSurface` contract. slide = freeform canvas (shipped); **quiz = structured form (Phase 2)**; interactive / PBL config = later surfaces. Same chrome contract; quiz is the proof it generalizes beyond canvas.
- **Actions / playback** — a cross-cutting layer every scene carries. Edited in Phase 2 via the **AI editing sidebar**, whose first capability is **(re)generating a scene's `actions` to match its current content**.

> Sibling-framework caveat, narrowed: editing that configures **interactive behavior / runtime orchestration** (PBL agents + issue board, live discussion) is genuinely a different editing task and may stay a sibling. But this is about the *editing shape*, not "interactive scenes are out" — and it is not a Phase 2 concern.

## New in Phase 2: MAIC Agent

[...]

PR #648, opened 2026-06-02
feat(maic-editor): slide content surface — z-order, background, image crop/flip (PR3b)

### LINKED ISSUE #647: [MAIC Editor] Slide content surface completion — z-order, background, image crop/flip
Parent: #646 · Branch: `feat/maic-editor-v1`

Completes the **slide content surface** — the remaining surface-level affordances deferred from Phase 1's PR3 split (PR3a shipped nav rail + chrome; this is PR3b). Track 1 of #646; **independent of MAIC Agent, can start first**.

All building blocks already exist on `main`; this issue is wiring + UI, not new kernel work:

- z-order op — `lib/edit/slide-ops.ts` already has `element.reorder`
- slide background — `lib/edit/slide-ops.ts` already has `slide.update` (metadata: theme / background / viewport)
- reusable controls — `components/edit/surfaces/slide/ColorPicker.tsx`, `ImagePicker.tsx`
- image crop — `components/slide-renderer/components/element/ImageElement/ImageClipHandler.tsx`

Surface-level only: **no chrome / store changes**. Renderer touches stay additive + `editable`-guarded + playback byte-unchanged (the Phase 1 rule).

### Z-order (any element)
- To-front / to-back (2-way only; intermediate forward/backward stays AI's domain, per the AI-first ceiling).
- Reuse the existing `element.reorder` op. Surface via the non-text anchored bar — extend `AnchoredDeleteBar.tsx` into an element-type-aware bar (or an overflow `⋯`), since today it is delete-only for non-text elements.

### Slide background (slide-level)
- Solid color + image. Reuse `slide.update` + `ColorPicker` + `ImagePicker`.
- This is **slide-level**, not element-anchored — needs a non-anchored affordance. Decide placement during implementation (command bar entry vs a dedicated slide-background control); keep it consistent with the anchored-bar IA from #590/#601.

### Image element
- Replace (reuse `ImagePicker`), **crop** (reuse `ImageClipHandler` — expose a crop trigger from an image-type anchored bar), flip H/V.
- Likely needs an image-type anchored bar (e.g. `AnchoredImageBar`) or making the non-text bar type-aware. Reuse `AnchoredBar.tsx` shell.

### Round-trip (hard gate)
- Every new op must round-trip through PPTX export: z-order = element order, slide background, image crop + flip. Add per-op cases to the existing round-trip harness (`tests/edit/slide-ops.test.ts` + the export round-trip suite).

## Deferred minors from #590 (optional fold-in)
Land here or separately:
- `use-tracked-rect.ts` rAF → `ResizeObserver`
- `editor-fonts.ts` route-scoping (currently ships on every page)
- `LineElement` 0×0 rect risk for the generic anchored bar
- font-size clamp consistency (8/96 in the bar vs legacy)
- interactive-component unit tests (anchored bars, ColorPicker, the create-selection text branch)

## Out of scope
- Anything requiring chrome or store changes
- Shape element editing, multi-select align/distribute (AI's domain per #560/#646 principle #2)
- New slide DSL element types (chart / table / video / formula)

## Success criteria
- [ ] z-order to-front/to-back, slide background (solid + image), image replace/crop/flip all work on the slide surface
- [ ] every new op round-trips cleanly through PPTX export (CI gate)
- [ ] playback byte-unchanged with feature flag OFF; renderer touches additive + `editable`-guarded
- [ ] `tsc --noEmit`, `pnpm test`, `pnpm lint`, `pnpm build` green
- [ ] targets `feat/maic-editor-v1`

_Estimated ~400–600 LOC, no chrome/store changes._

wyuc:
Completes the **slide content surface** (PR3b) — the surface-level affordances deferred from Phase 1's PR3 split. Closes #647. Track 1 of #646.

Surface-level only: **no chrome / store-shape changes**. All affordances ride existing slots (`SceneEditorSurface` insert strip + selection-anchored bars); op dispatch funnels through `useSlideEditSession.applyOp`. Renderer touches stay additive + `editable`-guarded; playback byte-unchanged with the flag OFF.

All kernel building blocks already existed on `main` — this is wiring + UI:
- z-order → existing `element.reorder`
- slide background → existing `slide.update` + `ColorPicker` / `ImagePicker`
- image crop → reuses the renderer's `clipingImageElementId` + `ImageClipHandler` (no new op); flip/replace → `element.update`
- the PPTX exporter already serializes `<p:bg>`, `<a:srcRect>` (crop) and `flipH`/`flipV`

## What's in

- **Z-order (all elements)** — `ZOrderButtons` (to-front/to-back, 2-way only; intermediate steps stay AI's domain), on both the element bar and the text bar.
- **Type-aware non-text bar** — `AnchoredDeleteBar` → `AnchoredElementBar`: images get replace/crop/flip; every non-text element gets z-order + delete.
- **Image controls** — replace (clears stale crop), crop (arms `ImageClipHandler`), flip H/V.
- **Slide background** — a "Background" insert-strip item with a solid (`ColorPicker`) / image (`ImagePicker`) popover.
- **Round-trip (hard gate)** — new `tests/edit/round-trip/{z-order,background,image-crop-flip}.test.ts`, asserting each op survives PPTX export (element order, `<p:bg>` solid+image, `<a:srcRect>`, `flipH/flipV`).
- **#590 deferred minors** — `use-tracked-rect` min-anchor clamp (degenerate line elements) + idle-stop rAF (re-arms on pointer press / resize / scroll); shared `FONT_SIZE_MIN/MAX` constants. (editor-fonts route-scoping was already handled by `preload-editor`'s dynamic import.)

## Tests

- `pnpm exec tsc --noEmit`, `pnpm lint`, `pnpm check`, `pnpm build` green.
- `pnpm exec vitest run` green (the `ssrf-guard` failures are local no-DNS only; CI has DNS).
- New unit + round-trip suites; e2e (`slide-content-surface-647.spec.ts`) enters Pro mode and verifies the background popover + z-order bar render and anchor.

## Notes for review

- Targets `feat/maic-editor-v1` (Phase 2 integration branch, mirrors how v0 used `feat/maic-editor-v0`).
- Image replace/crop/flip have no DOM-render e2e because the repo has no jsdom/RTL test infra (all existing slide tests are pure-logic); those ops are covered by `image-actions.test.ts` + the `image-crop-flip` round-trip suite instead. Open to adding RTL infra if preferred.


cosarah (changes requested):
Requesting changes for the Crop button only.

The other surface features I checked look OK: z-order, background, image replace, flip, and flip undo. Crop has several blockers.

Issues:

1. Crop commits are not reliably undoable.

`ImageClipHandler` can confirm crop via Enter or outside-click, and the commit goes through the renderer `updateElement()` path rather than a canonical `useSlideEditSession.applyOp()` operation. In the slide edit surface, non-pointer commits are treated as non-user normalization commits, so they update present content without pushing an undo entry. The result is that a confirmed crop can be applied but Undo does not restore the previous image state.

2. The crop operation does not behave like crop.

`ImageElement.handleClip()` writes both `clip` and element geometry (`left`, `top`, `width`, `height`). In practice this makes the result behave more like resizing/compressing the image than applying a stable crop. The UX needs a defined invariant here, e.g. whether crop preserves the outer frame or intentionally changes it, and the implementation/tests need to enforce that invariant.

3. Crop mode does not exit correctly when clicking inside the scene/canvas container.

Clicking blank space in the scene clears selection, which hides the anchored floating bar, but it does not clear `clipingImageElementId` or confirm/cancel crop. That leaves the editor in a confusing crop state with the visible surface controls gone.

4. Flip + crop produces unexpected results.

The normal image branch applies `flipStyle`, but the crop-mode branch renders `ImageClipHandler` without that flip transform. So the crop UI and range calculation operate on the unflipped image, then the normal flipped rendering is restored after confirmation. This makes flip + crop produce a result that does not match what the user saw while cropping.

Please either remove/hide Crop from this PR, or make crop a proper editor-surface operation with undo, clear exit/cancel behavior, defined geometry semantics, flip-aware handling, and e2e coverage.

xuyuanwei678 (approved):

wyuc:
Removed the Crop button in `86e5ed7` rather than ship the unintegrated clip path. All four points trace to the pre-existing renderer crop mechanism (`ImageElement.handleClip` via `clipingImageElementId`), which isn't an editor-surface op: the commit bypasses `useSlideEditSession.applyOp` (so Enter-confirmed crops aren't undoable), it rewrites element geometry instead of cropping within a stable frame, it doesn't clear crop mode on canvas-click, and the clip handler isn't flip-aware. Making crop correct needs renderer changes, which fall outside this PR's surface-only / playback-byte-unchanged scope.

This PR now ships z-order, slide background, image replace, and flip only. Crop will return as a dedicated PR with a canonical undoable op, a defined geometry invariant, exit/cancel on canvas-click, a flip-aware `ImageClipHandler`, and e2e — tracked in a dedicated follow-up issue.

PR #652, opened 2026-06-03
fix(audio): revoke blob URL when audio.play() rejects

### LINKED ISSUE #651: Blob URL leaks when audio.play() rejects during TTS playback
In `AudioPlayer.play()` (`lib/utils/audio-player.ts`), the IndexedDB playback branch creates an object URL via `URL.createObjectURL(audioRecord.blob)` and revokes it **only inside the `'ended'` listener**. If `await this.audio.play()` rejects — an expected, recoverable condition (autoplay policy before the first user gesture, decode error, interrupted load) — the `'ended'` event never fires, so the blob URL **and its backing audio blob leak for the lifetime of the document**.

During lecture playback `play()` is called once per speech action, so repeated failures (e.g. autoplay blocked before the first gesture, or rapid stop/skip) accumulate leaked blobs. The playback engine catches the rejection and falls back to a reading-time timer, so this path is genuinely exercised.

### Suggested fix
Revoke the object URL when `play()` rejects, then re-throw. PR to follow.

ly-wang19:
## Summary

In `AudioPlayer.play()`, the IndexedDB playback branch revokes its `URL.createObjectURL` blob URL **only inside the `'ended'` listener**. When `audio.play()` rejects (autoplay policy before the first user gesture, decode error, interrupted load) that listener never fires, so the blob URL and its backing audio blob leak for the lifetime of the document. `play()` is called once per speech action during lecture playback, so repeated failures accumulate.

This revokes the URL when `play()` rejects, then re-throws (all other paths unchanged).

## Related Issues

Closes #651

## Changes

- `lib/utils/audio-player.ts`: wrap the IndexedDB-branch `audio.play()` in a try/catch that revokes the blob URL on rejection.
- `tests/audio/audio-player-leak.test.ts`: regression tests (revokes on `play()` rejection; does not revoke prematurely on success).

## Type of Change

- Tests cover both the rejection path (URL revoked) and the success path (URL not revoked — revocation stays deferred to `'ended'`).
- Behavior-preserving otherwise; only the previously-leaking rejection path changes.

### Evidence

- [x] CI passes (`pnpm check` ✓, `pnpm lint` ✓, `npx tsc --noEmit` ✓, `vitest` 2/2 ✓)

## Checklist

---


wyuc (approved):
LGTM. Revoking the blob URL on the play() rejection path is the right fix, and the tests cover both the reject (revoked) and success (deferred to 'ended') cases. Pure memory hygiene, no behavior change on the happy path.

PR #659, opened 2026-06-03
fix(generation): preserve raw LaTeX through JSON.parse

### LINKED ISSUE #658: json-repair silently corrupts raw LaTeX (\frac, \times, ...) before any repair runs
### Summary

`parseJsonResponse` / `tryParseJson` (`lib/generation/json-repair.ts`) silently
corrupt math content when an LLM emits **raw** (single-backslash) LaTeX inside a
JSON string value, e.g. `"\frac{1}{2}"`, `"\times"`, `"\theta"`.

### Root cause

`tryParseJson` **Attempt 1** is a plain `JSON.parse(jsonStr)`. For LaTeX commands
whose name starts with a **valid JSON escape letter** — `\b \f \n \r \t \u` —
`JSON.parse` does **not** throw; it decodes the escape and returns successfully:

| raw LaTeX | `JSON.parse` decodes to |
| --- | --- |
| `\frac` | U+000C (form feed) + `rac` |
| `\times` | U+0009 (tab) + `imes` |
| `\nu` | U+000A (newline) + `u` |
| `\beta` | U+0008 (backspace) + `eta` |

Because Attempt 1 *succeeds* (with corrupted output), it returns immediately. The
existing LaTeX repairs ("Fix 1 / Fix 2") live in **Attempt 2**, which only runs
**after** Attempt 1 throws — so they never see this class. They do catch commands
like `\sum` / `\alpha`, whose `\s` / `\a` are invalid escapes that make Attempt 1
throw, which is why the bug went unnoticed for the throwing subset.

### Reproduction

The model emits a single backslash; the JSON string on the wire is:

```
{"latex":"\frac{1}{2}"}
```

```ts
const r = parseJsonResponse<{ latex: string }>('{"latex":"\\frac{1}{2}"}');
// r.latex === "rac{1}{2}"  (form feed + "rac"), NOT "\frac{1}{2}"
```

### Impact

Any slide/scene whose content contains a `\frac`, `\times`, `\theta`, `\tau`,
`\tan`, `\to`, `\top`, `\text`, `\nu`, `\nabla`, `\rho`, `\beta`, … emitted with a
single backslash is silently mangled (a control character is injected and the
command name is truncated). Fractions and multiplication are ubiquitous in a
math-education product, so this is a fairly common corruption.

### Expected

Raw LaTeX is preserved through parsing, while genuine JSON escapes (multi-line
`\n`, tabs in code, `\uXXXX`) remain intact.

### Fix

I have a fix that normalises stray backslashes inside string literals **before**
the first parse — a single linear scan that is a strict no-op on already-valid
JSON — and will open a PR referencing this issue.

ly-wang19:
## What & why

LLM responses frequently contain **raw** (single-backslash) LaTeX such as
`\frac`, `\times`, `\theta` inside JSON string values. `JSON.parse` does **not**
throw on these — it silently decodes `\f` → form feed, `\t` → tab, `\n` →
newline, `\b` → backspace — corrupting the math content.

The corruption was never reachable by the existing repair logic: `tryParseJson`
Attempt 1 is a plain `JSON.parse`, which **succeeds** (with corrupted output) for
these and returns immediately. The "Fix 1 / Fix 2" LaTeX repairs live in Attempt
2, which only runs after Attempt 1 *throws* — so they only ever caught commands
like `\sum` / `\alpha` (whose `\s` / `\a` are invalid escapes).

Fixes #658.

## Approach

Normalise stray backslashes inside string literals **before** the first parse,
with a single linear scan (`normalizeJsonStringBackslashes`, no regex
backtracking). It is a **strict no-op on already-valid JSON**:

- valid escapes `\"` `\\` `\/` `\n` `\r` `\t` `\uXXXX` are preserved — including a
  newline that precedes a lowercase word, and tabs inside code;
- `\n` / `\r` / `\t` stay control escapes **unless** the maximal lowercase run
  spells a known LaTeX command (`\nu`, `\rho`, `\times`, …), so genuine newlines
  and tabs are never reinterpreted;
- the one deliberate judgement call: `\b` / `\f` followed by letters are read as
  LaTeX (`\beta`, `\frac`), because a backspace / form-feed never appears in real
  educational content;
- the unfixable-by-construction case (a real newline/tab that *exactly* spells a
  command, e.g. literal `"\nu"` meaning newline + "u") degrades to the LaTeX
  reading — acceptable for a math product, and astronomically rare.

This supersedes the previous per-letter regex fixes (removed), which left
`\b\f\n\r\t`-initial commands silently corrupted and dropped LaTeX punctuation
like `\%` / `\_`.

## Test plan

`tests/generation/json-repair.test.ts` (new cases, all green):

- raw `\frac`, `\times`, `\theta`, `\nu_i`, `\rho`, `\beta`, `\sum \alpha`,
  `\left(\frac{a}{b}\right)`, `\% \_` are preserved end-to-end through
  `parseJsonResponse`;
- no-op cases: `\n`-before-lowercase-word, tab in code, `\uXXXX`, escaped
  backslash, already-escaped `\\frac`;
- direct unit checks that structure is untouched and a standalone `\n` stays a
  newline while `\frac` is double-escaped.

```
pnpm prettier --check   # clean
pnpm eslint             # clean
npx tsc --noEmit        # clean
npx vitest run tests/generation/   # 22 passed
```

No user-facing strings are added or changed (no i18n impact).

---

*This PR was written with AI assistance (Claude) and has been self-reviewed
end-to-end: I verified the silent-corruption root cause against the real
`parseJsonResponse` flow, and confirmed the normaliser is a no-op on valid JSON
across a multi-line/tab/unicode/escaped-backslash corpus before committing.*

### DIFF
--- a/lib/generation/json-repair.ts
+++ b/lib/generation/json-repair.ts
@@ -13,6 +13,180 @@ function repairQuotedPropertyFragments(jsonStr: string): string {
   );
 }
 
+/**
+ * LaTeX command names that begin with a JSON control-escape letter (n/r/t) and
+ * therefore collide with `\n` `\r` `\t`. Used by
+ * {@link normalizeJsonStringBackslashes} to tell e.g. `\nu` / `\times` (LaTeX)
+ * apart from a genuine newline / tab escape. The list is intentionally limited
+ * to common math commands; an unlisted command simply keeps the previous
+ * behaviour (no regression) — it does not break parsing.
+ */
+const NRT_LATEX_COMMANDS = new Set([
+  // \n…
+  'nu',
+  'nabla',
+  'neq',
+  'ne',
+  'ni',
+  'not',
+  'notin',
+  'nleq',
+  'ngeq',
+  'nless',
+  'ngtr',
+  'nmid',
+  'nparallel',
+  'nsubseteq',
+  'nsupseteq',
+  'nsim',
+  'ncong',
+  'nearrow',
+  'nwarrow',
+  'nexists',
+  'neg',
+  'natural',
+  'nrightarrow',
+  'nleftarrow',
+  'nleftrightarrow',
+  'nprec',
+  'nsucc',
+  // \r…
+  'rho',
+  'right',
+  'rightarrow',
+  'rightleftharpoons',
+  'rightharpoonup',
+  'rightharpoondown',
+  'rangle',
+  'rceil',
+  'rfloor',
+  'rbrace',
+  'rbrack',
+  'rtimes',
+  'rvert',
+  'rmoustache',
+  // \t…
+  'theta',
+  'times',
+  'tau',
+  'tan',
+  'tanh',
+  'text',
+  'textbf',
+  'textit',
+  'textrm',
+  'texttt',
+  'textstyle',
+  'tfrac',
+  'tbinom',
+  'therefore',
+  'tilde',
+  'to',
+  'top',
+  'tag',
+  'triangle',
+  'triangledown',
+  'triangleleft',
+  'triangleright',
+  'triangleq',
+  'tt',
+]);
+
+/**
+ * Re-escape stray backslashes inside JSON string literals so that raw LaTeX
+ * (which LLMs frequently emit) survives `JSON.parse`.
+ *
+ * Why this is needed: `JSON.parse` does not throw on `"\frac"` or `"\times"` —
+ * it silently decodes `\f` -> U+000C (form feed) and `\t` -> U+0009 (tab),
+ * corrupting the math content. Sequences whose command starts with a letter
+ * that is not a valid escape (`\sum`, `\alpha`) make `JSON.parse` throw instead,
+ * so the previous regex fixes only ever saw that second class and left the
+ * silent corruption in place. Normalising here, before the first parse, closes
+ * both paths with a single linear scan (no catastrophic backtracking).
+ *
+ * The function is a strict no-op on already-valid JSON: every valid escape
+ * (`\"` `\\` `\/` `\n` `\r` `\t` `\uXXXX`) is preserved, including newlines that
+ * precede a lowercase word and tabs inside code. The only deliberate
+ * reinterpretation is a bare `\b` / `\f` followed by letters — a
+ * backspace / form-feed reading never occurs in real educational content,
+ * whereas `\beta` / `\frac` are ubiquitous, so those are read as LaTeX.
+ */
+export function normalizeJsonStringBackslashes(jsonStr: string): string {
+  // Fast path: no backslash means no escapes to normalise, so the scan below
+  // would be a pure copy. Most structure-heavy JSON hits this.
+  if (!jsonStr.includes('\\')) return jsonStr;
+
+  let out = '';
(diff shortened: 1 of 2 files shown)


wyuc (changes requested):
Thanks for digging into this — the underlying bug is real and subtle (`JSON.parse` silently decoding `\f`/`\t`/`\n`/`\b` in raw LaTeX), and normalizing before the first parse is a reasonable direction. Before merging I'd like a more robust version, because as written the pass both leaves part of #658 unfixed and introduces a new corruption class for non-math content. Two blocking concerns:

**1. Unlisted `\n`/`\r`/`\t`-initial LaTeX commands are still silently corrupted, with no diagnostic.**
The n/r/t path only rescues commands present in the hardcoded `NRT_LATEX_COMMANDS` set; anything else falls through to `out += '\\' + next` and `JSON.parse` decodes it to a control char — the exact silent corruption from #658, just on a smaller surface, and it fails invisibly. Real examples that regress:
- `{"latex":"\twoheadrightarrow"}` → `run="twoheadrightarrow"` ∉ set → TAB + `woheadrightarrow`
- `{"latex":"\nsubset A"}` (set has `nsubseteq`, not `nsubset`) → newline + `subset A`
- also `\restriction`, `\nvdash`, `\thickapprox`, `\rightrightarrows`, …

The comment "an unlisted command simply keeps the previous behaviour (no regression)" is a bit misleading — parsing doesn't break, but the content is still corrupted, which is the defect being fixed. A hardcoded command list can never be complete.

**2. New regression: valid `\n`/`\t`/`\r` escapes in ordinary multiline content get rewritten when the next run happens to spell a listed command.**
Because the discriminator is "maximal lowercase run equals a command", legitimate control escapes collide:
- `{"latex":"line1\nu = 1"}` → `\\nu` → the real newline is lost
- `{"v":"x\to"}` (TAB + `o`) → literal `\to`; `{"v":"is\not"}` (newline + `ot`) → literal `\not`
- reachable too via `\ne`, `\ni`, `\tag`, `\tan`, `\top`, `\text`, …

So the docstring's "strict no-op on already-valid JSON … every valid escape preserved byte-identical" isn't accurate (this, plus `\b`/`\f` which are unconditionally reinterpreted). This can corrupt any multiline generated text/code, not just math.

**Suggested direction.** The `\n`-vs-`\nu` ambiguity is fundamentally unresolvable at the JSON-string level without semantics, so I'd avoid a heuristic that rewrites *already-valid* JSON. A parse-first fallback sidesteps both problems:
- Try a plain `JSON.parse` first. If it succeeds, only re-process when the decoded value contains suspicious control characters (a lone `\b`/`\f`, or a `\t`/`\n` sitting mid-word) rather than unconditionally; if it throws, run the normalization as today.
- Or, at minimum, keep the current approach but emit a diagnostic when a control escape is preserved next to a lowercase run, and document the residual gap explicitly instead of claiming a strict no-op.

The O(n) single-pass structure and the `\\`/`\uXXXX`/escaped-quote handling all look correct — it's specifically the unconditional rewrite of valid JSON that I'd like reworked. Happy to help iterate.

wyuc:

PR #660, opened 2026-06-03
feat(generation): opt-in parallel scene-content generation

### LINKED ISSUE #572: Opt-in parallel scene generation (hybrid two-phase)
### Problem

Scene generation in `lib/hooks/use-scene-generator.ts` runs strictly serial — content → actions → TTS, one outline at a time (see the `// Serial generation loop` at line ~329). For an N-scene classroom this is N × (content + actions + TTS) of wall-clock latency, which dominates the post-outline wait.

### Proposal

A hybrid two-phase parallel mode:

1. **Phase 1 — concurrent content.** `Promise.all` over pending outlines (capped) to fetch `/api/generate/scene-content`. Failures mark individual outlines and don't pause the batch.
2. **Phase 2 — serial actions + TTS.** Keep the current in-order for-of loop for `/api/generate/scene-actions` + TTS so that `previousSpeeches` threading and the pause-on-failure UX stay intact.

This keeps narrative coherence (previousSpeeches carryover) and ordered UI rendering, while parallelizing the heaviest per-scene LLM call. Expected speedup ≈ 2–3× at concurrency 3.

### Configuration

- Gated by a **server-side config flag**, **default OFF**.
- Concurrency is **configurable** (e.g., `PARALLEL_SCENE_CONCURRENCY`, default 3 when enabled).
- Rationale for default-off: many users bring API keys with low per-key concurrency quotas; a bursty default would surprise them with 429s. Opt-in keeps the out-of-box behavior unchanged.

### Out of scope

- A "full parallel" path (content + actions + TTS all concurrent) — would break previousSpeeches coherence.
- The `classroom-job-runner.ts` server path is also serial today but is not on the web critical path; can adopt the same pattern later.
- The unused `generateFullScenes` in `lib/generation/scene-generator.ts` (no callers) — worth removing as cleanup alongside this work.

ly-wang19:
## What & why

Implements the hybrid two-phase design from #572. Scene generation was strictly
serial (content → actions → TTS, one outline at a time), so an N-scene classroom
paid `N × (content + actions + TTS)` of wall-clock latency. This parallelises the
heaviest, safest step.

Closes #572.

- **Phase 1 — concurrent content.** Fetches `/api/generate/scene-content` for the
  pending outlines with bounded concurrency. Content is the only per-scene step
  independent of cross-scene state (no `previousSpeeches`), so it is safe to run
  in parallel. A content failure marks just that outline (and clears its spinner)
  without pausing the batch.
- **Phase 2 — serial actions + TTS.** Unchanged in-order loop, so
  `previousSpeeches` threading, ordered `addScene`, the `abort`/`epoch` guards,
  and the pause-on-failure UX all stay intact. Outlines whose content failed in
  Phase 1 are skipped.

## Opt-in / config

- Gated by **`PARALLEL_SCENE_CONCURRENCY`** (server-side), **default `0` = off**,
  clamped to `10`. With it unset the content map is `null` and the loop is
  **byte-for-byte the original serial path** — out-of-box behaviour is unchanged.
- Default-off rationale (from the issue): many keys have low per-key concurrency
  quotas; a bursty default would surface as 429s.

### One wiring decision (happy to change)

The generation loop is **client-side** (`useSceneGenerator`) but the flag is
**server-side**. As I flagged in the issue, I went with **option (b): piggyback on
the existing `/api/server-providers` response** (no new endpoint, runtime, not
build-time) — the settings store already fetches it at init, so the hook reads
`useSettingsStore.getState().parallelSceneConcurrency` with zero extra requests.
It is a small, self-contained change to swap to a `NEXT_PUBLIC_` env or a
dedicated endpoint if you'd prefer.

## Notable implementation detail

The bounded worker pool is extracted to **`lib/utils/concurrency.ts`**
(`mapWithConcurrency`, with an early-stop hook for abort/epoch) so the concurrency
logic is unit-tested in isolation. (`lib/export/inline-assets.ts` has an
equivalent private helper that could be consolidated onto this in a follow-up — I
left it untouched to keep this PR scoped.)

## Test plan

```
npx vitest run          # 82 files, 661 tests — all pass
npx tsc --noEmit        # clean
pnpm prettier --check   # clean
pnpm eslint             # clean
```

- `tests/utils/concurrency.test.ts` (new): order preservation, the concurrency cap
  is never exceeded, limit clamps to item count, early-stop via `shouldContinue`,
  empty list.
- `tests/server/provider-config.test.ts`: `getParallelSceneConcurrency` default-off,
  positive read, clamp-to-10, and zero/negative/non-numeric → off.
- The hook loop itself isn't unit-tested — the repo has no DOM/hook test harness
  (no `@testing-library` / jsdom env), and the serial path is unchanged. The novel
  logic (the pool) is covered above.

No user-facing strings are added or changed (no i18n impact).

---

*Written with AI assistance (Claude) and self-reviewed end-to-end: I traced the
existing serial loop's `previousSpeeches`/abort/epoch/pause semantics and kept
Phase 2 identical, and confirmed the full suite (661 tests) is green with the
settings-store change.*


wyuc (commented):
## Review — opt-in parallel scene-content generation

Nice change — the default-off gating is genuinely inert, the serial path is untouched, and extracting `mapWithConcurrency` with isolated tests is the right call. I built this branch and exercised both paths end-to-end through the real generation flow (not just unit tests); findings below. One UX trade-off is worth addressing before merge, the rest are minor.

### What I verified
- **Serial (flag unset/0):** byte-for-byte the original loop — content fetches are strictly sequential (peak in-flight = 1), scenes assemble in order, 0 regressions.
- **Parallel (flag = 10):** content fetches genuinely overlap (measured peak in-flight = 6 on a 9-scene deck, correctly clamped to item-count rather than over-spawning to 10), 0 failures, and Phase 2 still assembles scenes in the correct order despite out-of-order content completion. The `previousSpeeches`/ordered-`addScene` invariants hold.

### Main point: the barrier delays incremental rendering, and it scales with ⌈N/C⌉
Phase 2 (which does the `addScene` that actually paints a scene) starts only after `await mapWithConcurrency(...)` — i.e. after **every** content fetch resolves. The first scene is unaffected (it's generated eagerly in `generation-preview` before the classroom loop), but the **second scene onward** must wait for the *entire* content batch before any of them render.

Measured, same build, on a deck where scene count exceeds concurrency (`PARALLEL_SCENE_CONCURRENCY=2`, ~11 scenes ≈ 5 waves):

| | 1st→2nd scene gap | pattern |
|---|---|---|
| serial | **~72 s** | steady cadence, scenes appear one by one |
| parallel C=2 | **~5m20s** | first scene, then a long freeze on a single page, then scenes 2..N burst in together |

So in the `N > C` regime the user sees the first page promptly, then what looks like a stall on one page for the whole content phase, then a burst. Total wall-clock still improves (content is parallelized), but the perceived experience can be *worse* than serial because it reads as "done, then frozen." The gap grows with `⌈(N−1)/C⌉`.

### Suggested fix: pipeline instead of barrier (decouple spawn from consume)
Keep the content fetches concurrent, but **await them in order inside the existing serial loop** rather than awaiting the whole pool up front. Because each fetch is already in flight, awaiting scene *i* returns as soon as *its* content is ready — typically already done, hidden behind the actions+TTS of the previous scene. First classroom scene then paints after `content(1) + actions(1) + TTS(1)` (same as serial), and you keep the full concurrency benefit with no stall.

Roughly — turn the eager `await`-all into a bounded map of in-flight promises and pull per-outline in the loop:
```ts
// bounded (<=C in flight via a semaphore), but returns promises immediately — no barrier
const contentPromises = useParallelContent
  ? lazyBoundedMap(pending, C, (o) => fetchSceneContent({ outline: o, /* … */ }, signal), { shouldContinue })
  : null;

for (const outline of pending) {                 // unchanged ordered Phase-2 loop
  if (aborted) break;
  const contentResult = contentPromises
    ? await contentPromises.get(outline.id)        // already running → resolves when scene i is ready
    : await fetchSceneContent(/* … */);            // serial fallback
  if (!contentResult.success || !contentResult.content) { /* same continue/pause handling */ continue; }
  // Step 2/3: actions + TTS  (still serial, still uses previousSpeeches)
  store.getState().addScene(scene);                // paints incrementally
}
```
`lazyBoundedMap` is `mapWithConcurrency` turned inside-out: each item's promise acquires a `Semaphore(C)` before fetching, so all promises exist immediately but only `C` are in flight at once. This preserves every invariant the PR already relies on — content has no cross-scene dependency (safe to run ahead), while actions+TTS stay strictly serial and in order (`previousSpeeches` intact). It may even make the separate eager-first-scene path redundant.

If you'd rather keep this PR scoped to the barrier version, that's reasonable — but please add a sentence to the PR/issue noting the time-to-incremental-render trade-off in the `N > C` case, and file a follow-up for the pipelined version.

### Minor
- **Unexpected throw in Phase 1 aborts the whole batch.** The fetches run under `Promise.all`; the happy path returns `{success:false}` objects (handled), but if `fetchSceneContent` ever *rejects* (non-Abort), the whole pool rejects and **no** outline reaches Phase 2 — whereas the serial path would have committed earlier scenes. Low probability, but a `try/catch` inside the Phase-1 callback that routes unexpected throws into the same "mark this outline failed" path would make it symmetric with serial.
- **`setCurrentGeneratingOrder(pending[0].order)` during Phase 1** is cosmetic-only while several fetches are in flight — minor, but the "current scene" indicator will be slightly misleading during the content phase.
- **Nit:** the `Math.max(0, Math.floor(...))` clamp is applied in both `settings.ts` and the hook; a short comment that it's intentional belt-and-suspenders would help.
- **Nit:** `concurrency.test.ts` asserts `expect(peak).toBe(3)`; `toBeLessThanOrEqual(3)` matches the actual invariant and is less timing-sensitive (the clamp test already uses `toBeLessThanOrEqual`).

Overall: correct and safe to merge as-is; the pipeline change is the one I'd genuinely encourage, since it removes the only real downside.

ly-wang19:
Thanks for the unusually thorough review — building both paths and *measuring* the `N > C` time-to-incremental-render gap is exactly the kind of thing unit tests miss, and you’re right that the barrier’s "first page, then a frozen page, then a burst" reads as worse than serial.

Took the **pipeline** route (pushed in `0eaf2a5`):

- New `lazyBoundedMap` (the "`mapWithConcurrency` turned inside-out" you described): a FIFO semaphore caps in-flight fetches at `C`, but it returns one promise per outline **immediately**. The serial loop `await`s them **in order**, so scene *i* resolves as soon as *its* content is ready — usually already done behind scene *i-1*’s actions/TTS. First scene now paints after `content(1)+actions(1)+TTS(1)`, same as serial, no stall. `mapWithConcurrency` stays as a thin `Promise.all(lazyBoundedMap(...))` wrapper so its tests are unchanged.

All four minors addressed too:
- **Unexpected throw:** the fetch callback now `try/catch`es and returns a `{success:false}` result, so a non-Abort throw routes through the same mark-failed path as serial instead of rejecting the pool. With the pipeline, earlier scenes are already committed by the time a later fetch settles — symmetric with serial.
- Dropped the cosmetic Phase-1 `setCurrentGeneratingOrder`; the loop sets it per outline.
- Added a comment that the double clamp is intentional belt-and-suspenders.
- `concurrency.test.ts` now asserts `toBeLessThanOrEqual(limit)`, plus new `lazyBoundedMap` tests including the no-barrier property (first promise resolves without forcing the last item to run).

Full suite green (664 tests), tsc/lint/format clean. The separate eager-first-scene path in `generation-preview` is now arguably redundant as you noted — happy to file a follow-up to remove it rather than widen this PR.

wyuc (approved):
Pipeline revision looks great — `lazyBoundedMap` + in-order consumption is exactly the no-barrier shape, and the throw-symmetry/clamp/test cleanups are all in. I rebuilt the branch and re-ran both paths end-to-end: serial unchanged, and the parallel `N>C` stall is gone (second scene now paints right after the first instead of after the whole content batch). Thanks for the fast, thorough turnaround. LGTM. (A follow-up to drop the now-redundant eager-first-scene path sounds right.)

PR #662, opened 2026-06-03
fix(llm): increase maxOutputTokens for search query rewrite to suppor…

hxf0223:
When using reasoning models (e.g. DeepSeek V4 Pro), the `web-search-query-rewrite`
call consistently produces empty output and causes JSON parse failures:

```text
WARN] [LLM] [web-search-query-rewrite] LLM returned empty text
{"finishReason":"length","outputTokens":256,"textTokens":0,"reasoningTokens":256}
[WARN] [Generation] Attempt 1 parse error: Unexpected end of JSON input
[ERROR] [Generation] Failed to parse JSON from response
[WARN] [SearchQueryBuilder] Query rewrite returned empty output, falling back to raw requirement
```

**Root cause:** Reasoning models share `max_tokens` between internal thinking tokens
and visible output tokens. With `maxOutputTokens=256`, all 256 tokens are consumed
by the model's internal reasoning, leaving zero tokens for visible output.

Verified via direct API call:
- `max_tokens=256` → `finish_reason: length`, `text_tokens: 0`, `reasoning_tokens: 256`
- `max_tokens=2048` → `finish_reason: stop`, output: `{"query": "..."}`

## Changes

- **`app/api/web-search/route.ts`**: `maxOutputTokens: 256 → 2048`
- **`lib/server/classroom-generation.ts`**: `maxOutputTokens: 256 → 2048`
- **`lib/ai/llm.ts`**: add diagnostic warning when LLM returns empty text, logging
  `finishReason`, full usage breakdown (including `reasoningTokens`), and model name
  to make future debugging easier

## Impact

Only affects the web search query rewrite call. No impact on non-reasoning models
(256 tokens is already sufficient for a short JSON query on those models).

### DIFF
--- a/app/api/web-search/route.ts
+++ b/app/api/web-search/route.ts
@@ -87,7 +87,7 @@ export async function POST(req: NextRequest) {
               { role: 'system', content: systemPrompt },
               { role: 'user', content: userPrompt },
             ],
-            maxOutputTokens: 256,
+            maxOutputTokens: 2048,
           },
           'web-search-query-rewrite',
           undefined,
--- a/lib/ai/llm.ts
+++ b/lib/ai/llm.ts
@@ -282,6 +282,21 @@ export async function callLLM<T extends GenerateTextParams>(
         continue;
       }
 
+      // Diagnostic: warn when the model returns empty text so we can identify root cause
+      if (!result.text || result.text.trim().length === 0) {
+        const resultAsUnknown = result as unknown as Record<string, unknown>;
+        log.warn(`[${source}] LLM returned empty text`, {
+          finishReason: result.finishReason,
+          usage: result.usage,
+          hasReasoning: !!resultAsUnknown.reasoning,
+          reasoningLength:
+            typeof resultAsUnknown.reasoning === 'string'
+              ? resultAsUnknown.reasoning.length
+              : undefined,
+          model: getModelId(params),
+        });
+      }
+
       return result;
     } catch (error) {
       lastError = error;
--- a/lib/server/classroom-generation.ts
+++ b/lib/server/classroom-generation.ts
@@ -217,7 +217,7 @@ export async function generateClassroom(
           { role: 'system', content: systemPrompt },
           { role: 'user', content: userPrompt },
         ],
-        maxOutputTokens: 256,
+        maxOutputTokens: 2048,
       },
       'web-search-query-rewrite',
     );


wyuc (commented):
Thanks for tracking this down — the diagnosis is correct and the fix is legitimate. The two edited sites are the only two `web-search-query-rewrite` callers, and they were the only hardcoded-small budgets (every other generation site already uses the dynamic `modelInfo?.outputWindow`). The reasoning/output token-sharing root cause is real and well explained, and blast radius is low.

A few things before merge:

**1. The new diagnostic reads the wrong field — its reasoning signals are dead.**
We're on `ai@^6`. In AI SDK v6, `GenerateTextResult.reasoning` is an **array** of reasoning parts and `reasoningText` is the **string**. So:

```ts
hasReasoning: !!resultAsUnknown.reasoning,          // an array is always truthy → always true
reasoningLength: typeof resultAsUnknown.reasoning === 'string'
  ? resultAsUnknown.reasoning.length : undefined,    // never typeof 'string' → always undefined
```

`hasReasoning` is `true` even with no reasoning, and `reasoningLength` is always `undefined` — ironic for a diagnostic meant to surface exactly the reasoning-heavy case. Please read `result.reasoningText` instead. (`usage.reasoningTokens` is the one field here that already carries the real signal.)

**2. Consider disabling thinking for this call rather than just enlarging the budget.**
`callLLM` already accepts a per-call `thinking?: ThinkingConfig`, which supports `enabled: false` / `mode: 'disabled'`. The query-rewrite wants a short deterministic JSON object — reasoning here adds latency and **billed reasoning tokens** for no benefit. Passing `thinking: { enabled: false }` would fix the empty-output failure *and* avoid spending the tokens, instead of paying for up to 2048 tokens of reasoning the call doesn't need. At minimum, `2048` deserves a comment explaining why.

**3. Duplicated magic number.** This PR is itself the evidence — the same `256 → 2048` edit had to be applied in two copy-pasted call sites. A shared constant/helper would stop them drifting. Out of scope, but worth a follow-up.

Happy to merge once #1 is fixed; #2 is the stronger fix if you're up for it.

wyuc:
@hxf0223 Confirmed on our side — the diagnosis is right and the fix is legitimate (the two edited sites are the only `web-search-query-rewrite` callers). The only thing blocking merge now is a merge conflict against `main`. Could you rebase onto the latest `main` and resolve it? Once it's mergeable we'll get this in. Thanks!

wyuc (changes requested):
Formalizing the blocker from my earlier review on this exact head.

The empty-output diagnosis is valid, but the new diagnostic reads AI SDK v6's reasoning result incorrectly: `result.reasoning` is an array of reasoning parts, while the string field is `result.reasoningText`. As written, `hasReasoning` is true even for an empty array and `reasoningLength` is always undefined, so the diagnostic cannot identify the case it was added for.

Please rebase onto current `main`, switch this diagnostic to the correct v6 field (with a focused test), and consider disabling thinking for this short JSON rewrite rather than paying for a much larger reasoning budget. I can re-review after the author-side rebase.

wyuc:

PR #666, opened 2026-06-04
feat(tts): unify provider-enablement model, default browser-native off (#665)

### LINKED ISSUE #665: TTS provider availability is inconsistent and uncontrollable; make it a unified enablement model and default browser-native off
## Summary

Which TTS providers/voices are available is decided by several ad-hoc gates in `getAvailableProvidersWithVoices` (`lib/audio/voice-resolver.ts`), a hardcoded browser-native branch in the agent UI, and hardcoded browser-native fallbacks in the resolution paths. The result is inconsistent and uncontrollable. The symptoms below share one root cause: **there is no unified provider-enablement model.**

## Symptoms

**1. Unconfigured keyless providers are force-shown (bug).** `voice-resolver.ts:132-146`:
```js
const isKeylessLocalProvider =
  !config.requiresApiKey &&
  !!(isServerConfigured || providerConfig?.baseUrl?.trim() || config.defaultBaseUrl);
```
`config.defaultBaseUrl` is always set for local providers (e.g. `lemonade-tts` -> `http://localhost:13305/v1`), so its 54 voices are always selectable even with nothing running there; synthesis then fails.

**2. The per-provider `enabled` flag is ignored.** `lib/store/settings.ts` defines per-provider `enabled` (+ `setTTSEnabled`/`setProviderConfig({enabled})`), but `getAvailableProvidersWithVoices` never reads it — a user can't hide a configured provider.

**3. Browser-native TTS can't be disabled.** In `components/agent/agent-bar.tsx` it's appended unconditionally whenever `browserVoices.length > 0`; no toggle.

**4. Agent voice auto-assignment relies on fragile, implicit fallback.** `resolveAgentVoice` (`voice-resolver.ts`) and `resolveVoiceForAgent` (`lib/hooks/use-discussion-tts.ts`): with no valid `voiceConfig`, they pick `availableProviders[0]` by index, falling back to a hardcoded `browser-native-tts` only when no provider exists. So browser-native is avoided only by being excluded/last in the list; `availableProviders[0]` is an arbitrary `Object.entries` order; and the hardcoded fallback bypasses any enablement model (contradicts #3).

## Proposed direction

A single provider-enablement model honored everywhere (picker, auto-assignment, fallback):

- **Availability requires real configuration** — drop `|| config.defaultBaseUrl` so a keyless provider shows only when server-configured or given an explicit base URL (fixes Lemonade; VoxCPM still shows when configured).
- **Honor `enabled`** at the **user/client** level, and support **admin/server-level disable** via `server-providers.yml`/env (server precedence) so deployments can force a provider off.
- **Browser-native becomes a first-class provider in this model, default OFF and fully opt-in** — never auto-assigned, no hardcoded fallback. (Native voice quality is poor; it shouldn't be a silent default.)
- **Route auto-assignment/fallbacks through the model** — deterministic selection among *enabled* providers; remove the hardcoded `browser-native-tts` bypass and the implicit ordering dependency.
- **When no provider is enabled/available**, the toolbar TTS control (`components/audio/tts-config-popover.tsx`) stays off and shows a **CTA** prompting the user to enable Browser Native TTS (or configure a provider) — instead of silently falling back.

## Scope / non-goals
- Provider-neutral; no change to synthesis logic itself.
- Behavior change: fresh installs with no configured TTS will have audio **off by default**, with a clear CTA to enable browser-native or configure a provider (intentional — replaces the poor-quality silent fallback).
- Found on `main` @ `2586cb5`.

wyuc:
Closes #665.

## Problem
TTS provider availability was decided by several ad-hoc gates in `getAvailableProvidersWithVoices`, a hardcoded browser-native branch in the agent UI, and hardcoded browser-native fallbacks in the resolution paths — inconsistent and uncontrollable, all rooted in the absence of a unified provider-enablement model.

## Approach
A single enablement model honored everywhere (picker, auto-assignment, fallback, toolbar CTA).

**`lib/audio/provider-enablement.ts`** (new, pure):
- `isTTSProviderConfigured` — availability. **Drops `|| defaultBaseUrl`** so a keyless local provider (Lemonade) shows only when server-configured or given an explicit base URL (fixes symptom 1). browser-native is intrinsically configured.
- `isTTSProviderEnabled` = configured && `enabled !== false` && `!serverDisabled` (server precedence).
- `listEnabledTTSProviderIds` / `hasAnyEnabledTTSProvider`.

**Resolution** (`voice-resolver`, `use-discussion-tts`):
- `getAvailableProvidersWithVoices` → `getEnabledProvidersWithVoices`, filtered through the predicate (the dead `enabled` flag is now live — symptom 2).
- `resolveAgentVoice` picks deterministically among enabled providers (canonical registry order), honors a `voiceConfig` only while its provider is enabled, **never auto-assigns browser-native**, and returns `null` instead of the hardcoded `browser-native-tts` fallback (symptom 4). Teacher path routes through the model too.

**Browser-native** (symptom 3): a first-class provider, **default OFF, fully opt-in**. Surfaced only when the user enabled it and it is not server-disabled. `v3 → v4` migration forces it off for existing users.

**Server/admin disable** (server precedence): `server-providers.yml` (`tts.<id>.enabled: false`) and env (`TTS_<PROVIDER>_ENABLED=false`, incl. `TTS_BROWSER_NATIVE_ENABLED`). Surfaced via a per-provider `disabled` flag on `/api/server-providers` and merged into the store as `serverDisabled`.

**Empty state**: when no provider is enabled/available, `tts-config-popover` shows a CTA — one-click **Enable Browser Native TTS** + a link to provider settings — instead of silently falling back. `tts-settings` gains a per-provider enable toggle (disabled + noted when server-off).

## Behavior change (intentional, per issue)
Fresh installs with no configured TTS have audio **off by default**, with a clear CTA. Existing browser-native users are migrated off and re-enable via the CTA; users with a configured server/key provider are unaffected.

## Tests
- `tests/audio/provider-enablement.test.ts` — predicates incl. Lemonade-not-shown, server-disable precedence, browser-native opt-in.
- `tests/audio/voice-resolver-enablement.test.ts` — enabled filtering + resolver determinism + null-when-none + no browser-native auto-assign.
- `tests/server/provider-config.test.ts` — server force-disable via env & YAML, env-true overrides YAML.
- `tests/store/settings-server-sync.test.ts` — browser-native default OFF + serverDisabled merge/reset.
- `tests/i18n/tts-enablement-locales.test.ts` — new keys across all 7 locales.

Verification: `tsc --noEmit` clean, eslint clean, `next build` passes, full vitest suite green (the only failures are the pre-existing `ssrf-guard` DNS-sandbox tests, unrelated and failing on clean `main`).

Found on `main` @ `2586cb5`.

## Downstream note (openmaic-live)
Touches `lib/store/settings.ts` (persist v4 migration, new `serverDisabled` field) and the `/api/server-providers` TTS payload shape (`{ disabled? }`) — coordinate the -live rebase accordingly.


wyuc:
## Code-review round 1 (self-review, multi-angle)

Ran a 7-angle review (correctness ×3, cleanup ×3, altitude) over the diff. Highest-signal finding: the enablement model was honored in the **discussion path + picker** but **not** in the primary **lecture-generation, playback, and server** paths, which still trusted the raw `ttsProviderId === 'browser-native-tts'` check. Fixed in `95d7f1b`:

**Coherence (model honored everywhere)**
- `use-scene-generator` (both gates) + `generation-preview`: server TTS generation now gates on `isTTSProviderEnabled`, so a disabled/unconfigured global provider no longer triggers failing POSTs or a silent half-broken state.
- `playback/engine`: browser-native client-side speech fires only when browser-native is actually enabled (opt-in).

**Server precedence enforced end-to-end**
- `/api/generate/tts` rejects a force-disabled provider (`isServerTTSProviderDisabled` + new `PROVIDER_DISABLED` code) — the disable is no longer advisory-only on the client.
- `classroom-media-generation` and `/api/health` no longer count force-disabled providers as available.

**Edges**
- Empty `TTS_<P>_ENABLED=` no longer overrides a YAML disable (treated as "no opinion").
- `hasAnyEnabledTTSProvider` short-circuits.
- Added tests for the force-disable helper + empty-env edge.

### Consciously deferred (not bugs)
- **Memoization** of `getEnabledProvidersWithVoices`/`hasAnyEnabledTTSProvider` in `agent-bar`/`tts-config-popover`/`use-discussion-tts`: these mirror the pre-existing unmemoized pattern on `main` (the old `getAvailableProvidersWithVoices` was called the same way) — not a regression introduced here.
- **`validateProvider`/`isProviderUsable` not checking `serverDisabled`**: the *effective* behavior is correct (downstream gating + route enforcement + `buildFallback` already exclude server-disabled); the residual is a cosmetic selection/pill mismatch only.
- **Predicate overlap** (`isTTSProviderConfigured` vs `isProviderUsable`) and the **settings toggle-row** markup duplication: design-level cleanup, out of scope for this fix.

### Verification
`tsc --noEmit` clean · eslint clean · `next build` passes · full vitest suite green except the pre-existing `ssrf-guard` DNS-sandbox failures (fail identically on clean `main`).

wyuc:
## E2E (real headless browser) — `web-e2e` skill

Ran a real-browser pass against a live dev server (`pnpm dev` :3007), driving the actual Settings UI. The session had no Playwright **MCP** connected, so per the skill's sanctioned fallback I drove a headless Chromium via a Playwright script (fresh context = clean localStorage = default config; app rendered in zh-CN). No seeding/mocking.

### Verdict: ✅ PASS (after fixing one regression the run surfaced)

| Check | Result |
|---|---|
| Settings dialog opens | ✅ |
| New per-provider “启用此提供方” toggle + hint render in TTS section | ✅ |
| Server-managed provider (Qwen TTS 服务端) **enabled ON** by default | ✅ (after fix) |
| Browser-native provider **OFF** by default (opt-in) | ✅ `aria-checked=false` |
| Toggling the switch flips state | ✅ false→true |
| Console errors | none |

### Regression caught by the run → fixed (`c331130`)
First pass showed the **default-selected provider was Qwen (服务端, server-managed) with its enable toggle OFF** — i.e. a provider the operator configured server-side was hidden from the picker/auto-assignment. Root cause: `getDefaultAudioConfig` defaulted most built-ins to `enabled:false`, and now that the model honors `enabled`, those configured providers got hidden (pre-change `isServerConfigured` bypassed `enabled`). My design intent ("other built-ins default enabled:true") was never actually implemented.

Fix: all built-in TTS providers default `enabled:true` (browser-native stays opt-in `false`); v3→v4 migration normalizes the never-user-editable flag the same way. Re-ran e2e → all 5 checks green: Qwen ON by default, browser-native OFF, toggle flips.

### Evidence (local screenshots)
`step-04-tts-section` (toggle + hint), `step-06-selected-provider-on` (Qwen 服务端 ON), `step-07-browser-native-off` (浏览器原生 TTS toggle OFF with hint “关闭后…不会出现在语音选择器中，也不会被自动分配”).

### Not covered
- The toolbar empty-state **CTA** (`tts-config-popover`) — that component is not currently mounted anywhere in the app, so it's unreachable via UI; covered by unit tests only.
- Full lecture-generation / playback audio paths — would need a live classroom generation (LLM, ~minutes); the gating logic is covered by the code-review fixes + unit tests.

Verification after the fix: `tsc` clean · eslint clean · `next build` passes · vitest green (105 store+audio tests incl. new default/migration cases).

wyuc:
## E2E round 2 — 音色选择 (voice picker) 深测 ✅ PASS

First round only exercised the Settings enable toggle. This round drives the actual **voice picker** (the AgentBar 音色选择器 — the main UI surface of `getEnabledProvidersWithVoices`) in a real headless browser, plus the strongest dynamic check.

| Check | Result |
|---|---|
| AgentBar shows voice pills (Qwen 服务端 enabled) | ✅ |
| Teacher voice picker opens (搜索音色) | ✅ |
| Picker lists the **enabled** provider's voices — group "Qwen TTS · Qwen3 TTS Flash" (芊悦/Cherry, 苏瑶/Serena, …) | ✅ |
| Browser-native / OS voices **absent** from the list | ✅ |
| Search "Cherry" filters to 芊悦 (Cherry) only | ✅ |
| Selecting Cherry closes picker + writes back → pill = "芊悦 (Cherry)" | ✅ |
| **Dynamic**: disable Qwen's enable toggle in Settings → reopen AgentBar → voice pills = **0** | ✅ |

The last check is the key end-to-end proof: the per-provider `enabled` flag actually governs which voices are selectable — turn the only enabled provider off and every voice disappears from the picker. No console errors.

**Not covered this round:** browser-native-enabled→voices-appear (headless Chromium has no `speechSynthesis` OS voices, so the list stays empty even when enabled); per-agent pills in 预设 mode (default is `auto` → only the Teacher pill shows; same `resolveAgentVoice` + picker component); full lecture generation/playback (LLM, minutes).

wyuc:
## Follow-up fix — empty-state voice pill (`8dcf1f9`)

The round-2 e2e surfaced a UX gap (raised in review): when no TTS provider is enabled (browser-native OFF by default + none configured), the AgentBar hid the voice pill **entirely** via `showVoice = availableProviders.length > 0` — the dropdown just vanished. That `showVoice` gate is pre-existing, but it was practically unreachable before #665 (browser-native was an always-on fallback with OS voices); #665 made the empty state reachable, and silently vanishing contradicts #665's own "never silently off, show a CTA" principle.

**Fix:** AgentBar voice pills (teacher + agent) now always render. With TTS on but no enabled provider, they render a **"无可用音色 / No voice"** pill that opens **Settings → TTS** on click (shared `NoVoicePill`). Dropped the hide-everything gate; added `agentBar.noVoice` across all 7 locales + parity test.

Verified in a real headless browser: disable the only enabled provider → the pill **still renders** (not hidden) → clicking it opens the Settings TTS section. `tsc`/eslint/`next build` clean; 108 i18n+store+audio tests green.

wyuc:
## Simplified the empty-state fix (`941d55e`, force-pushed)

Per review feedback, dropped the earlier `NoVoicePill` + "open Settings" CTA — it was over-engineered. Now the empty case is handled exactly like the existing **TTS-disabled** state: the voice pill always renders but **greyed and non-interactive** (`disabled={!ttsEnabled || availableProviders.length === 0}`), instead of vanishing. No new component, no CTA, no settings-jump.

Verified in a real headless browser: disable the only enabled provider → the AI教师 voice pill still renders (greyed "🔇 Cherry") and is inert on click (no picker, no dialog) — same effect as turning TTS off. `tsc`/eslint/`next build` clean.

cosarah (commented):
Findings:

1. The TTS server-disabled state is not considered when validating the currently selected TTS provider. fetchServerProviders() excludes disabled providers from fallback candidates, but validateProvider() still keeps the current provider if it has a local API key because isProviderUsable() ignores serverDisabled. If an admin disables the selected provider while another TTS provider is available, scene/lecture TTS generation will still see the disabled provider selected and skip generation instead of falling back.

2. Several touched files are not Prettier-formatted: lib/audio/provider-enablement.ts, lib/audio/voice-resolver.ts, lib/hooks/use-discussion-tts.ts, lib/hooks/use-scene-generator.ts, and tests/audio/voice-resolver-enablement.test.ts. This will fail the repo formatting check.

wyuc:
Thanks for the review.

**1. Server-disabled provider stays selected.** Good catch — confirmed. `isProviderUsable()` did not account for `serverDisabled`, so `validateProvider()` kept a force-disabled provider whenever it had a local API key, and generation then skipped instead of falling back. Fixed: `serverDisabled` now forces `isProviderUsable()` to `false`, so the current selection is re-pointed to an available provider on the next sync. Added unit tests for both the predicate and the re-point path.

**2. Prettier.** Formatted all touched files; `prettier --check` is clean now.

Both pushed in the amended commit.

cosarah:
LGTM， verified locally

cosarah (approved):
LGTM, verfied locally

PR #672, opened 2026-06-07
feat(tts): per-agent auto-voice quality + register-once timbre stability (#670)

### LINKED ISSUE #670: Upgrade auto-voice quality: per-agent voice-design prompt + register-once/reference-by-id timbre stability
## Summary

The `voxcpm:auto` "auto voice" produces weak, unstable timbre. This proposes upgrading it into a proper per-agent voice identity: generate a real voice-design prompt per agent at profile-generation time, then stabilize timbre by registering it once and referencing it by id, with a re-register-on-invalid safety net.

## Problem

Today the auto voice prompt comes from `buildAutoVoxCPMVoicePrompt(context)` (`lib/audio/voxcpm.ts`), which essentially reuses the agent's **persona** text (sanitized, ≤200 chars), falling back to `role + name`, then the literal `"natural classroom voice"`. Two issues:

1. **Persona is not a voice description.** It describes personality / teaching style, not vocal timbre (gender/age/pitch/texture/pace). Fed as a voice-design prompt, it gives generic, poorly-matched voices.
2. **Text-only voice design drifts.** VoxCPM2's own docs note "every generation can still have subtle, unique variations", so the same agent sounds different across utterances within a single session. Timbre is not stable.

## Proposed pipeline

1. **Per-agent voice-design prompt at profile generation.** Extend `app/api/generate/agent-profiles/route.ts` so the LLM emits, alongside name/role/persona, a structured voice-design descriptor per agent following a 3-layer recipe — **identity** (gender/age/role) + **texture** (pitch/vocal quality) + **delivery** (emotion/pace) — in the course language and consistent with the persona. Persist it on the agent (`AgentConfig`).

2. **Bootstrap → register → reference-by-id (stable timbre).** When the TTS backend supports voice registration (e.g. vLLM-Omni `POST /v1/audio/voices`):
   - Synthesize the voice-design prompt **once** into a reference clip.
   - Register it as a custom voice → obtain a `voice_id`.
   - Subsequent TTS references `voice=voice_id` (pre-encoded latents), which locks timbre across the session and shrinks the per-request payload (no re-sending the reference, no re-running text voice-design).
   - The **reference audio is the source of truth**; `voice_id` is a **deterministic key** (e.g. hash of the reference) so registration is idempotent.

3. **register-on-invalid robustness.** If a referenced voice is missing (backend restart / cache eviction / GC), re-register from the stored reference audio and retry, transparently.

4. **Graceful degradation for backends without registration.** Keep today's inline voice-design path (the `(prompt)text` prefix) as the fallback for backends that can't register voices, accepting its drift.

## Affected code (pointers)

- `app/api/generate/agent-profiles/route.ts` — emit the voice-design descriptor in the prompt + output schema + parse onto the agent.
- `lib/orchestration/registry/types.ts` — carry the descriptor (and resolved `voice_id`) on `AgentConfig`.
- `lib/audio/voxcpm.ts` — `buildAutoVoxCPMVoicePrompt` prefers the descriptor; add the 3-layer recipe.
- `lib/audio/voxcpm-voices.ts` — `getVoxCPMProviderOptions` auto branch drives bootstrap→register→reference-by-id when supported; register-on-invalid.

## Scope / non-goals

- Provider-neutral; this is about **auto-voice quality + timbre stability**, distinct from TTS provider enablement (#665).
- Timbre stability comes from **reference-audio reuse**, not longer text prompts (per VoxCPM2 docs); prompt quality only sets the *initial* character.
- Keeps the inline voice-design path as a fallback, so no regression for backends without a registration API.

wyuc:
Closes #670.

## Problem
`voxcpm:auto` derived its voice prompt from the agent **persona**, which describes personality, not vocal timbre. The result was generic, poorly-matched voices — and because text voice-design drifts, the same agent could sound different (even switching apparent gender) across utterances in a single session.

## What this does
**1. Per-agent voice design.** `agent-profiles` now emits a 3-layer `voiceDesign` (identity / texture / delivery) per agent, in the course language, persisted on `AgentConfig`.

**2. Register-once / reference-by-id, provider-neutral.** A `VoiceRegistrationAdapter` seam + a generic endpoint `POST /api/generate/voice` synthesize the descriptor into a reference clip **once**, register it on the backend, and return a deterministic `voiceId`. Later TTS references `voice=<voiceId>` (pre-encoded latents), locking timbre and shrinking the payload. vLLM-Omni is the first adapter; other registration-capable providers plug in by adding an adapter. Backends without registration keep the inline `(prompt)text` path (graceful degradation).

**3. Single voice resolver.** All TTS paths (lecture narration, multi-agent discussion, voice preview) resolve through one `resolveAgentVoiceOptions(agent, …)` that reads the agent profile, so they can't drift out of sync. Agents without a `voiceDesign` (e.g. preset agents) fall back to their persona as the seed — stable, if generic.

**4. Robustness.** Deterministic id = hash(descriptor + provider + model); a GC'd registered voice is re-registered transparently via a proactive existence check. Managed TTS providers can pin their model server-side (`resolveTTSModel`).

## Tests
- Unit: voice-design helpers + deterministic id; the registration adapter (multipart contract, list-based existence, bootstrap); provider→adapter dispatch; the resolver's voiceDesign-vs-persona source and narrator selection.
- E2E (browser): generating a course registers the voice once and every narration TTS reuses the single registered `voiceId`; the registered voice is present on the backend.

## Known limitation / follow-up
The persisted stage snapshot doesn't yet carry `voiceDesign` (only persona), so a reopened/generated classroom resolves a stable persona-based voice rather than its original descriptor voice. Persisting the full agent profile on the stage is a clean follow-up.

### DIFF
--- a/app/api/generate/agent-profiles/route.ts
+++ b/app/api/generate/agent-profiles/route.ts
@@ -12,7 +12,7 @@ import { createLogger } from '@/lib/logger';
 import { apiError, apiSuccess } from '@/lib/server/api-response';
 import { resolveModelFromRequest } from '@/lib/server/resolve-model';
 import { AGENT_COLOR_PALETTE } from '@/lib/constants/agent-defaults';
-import { normalizeVoiceDesign } from '@/lib/audio/voxcpm';
+import { normalizeVoiceDesign } from '@/lib/audio/voice-design';
 
 const log = createLogger('Agent Profiles API');
 
--- a/app/api/generate/agent-profiles/route.ts
+++ b/app/api/generate/agent-profiles/route.ts
@@ -12,6 +12,7 @@ import { createLogger } from '@/lib/logger';
 import { apiError, apiSuccess } from '@/lib/server/api-response';
 import { resolveModelFromRequest } from '@/lib/server/resolve-model';
 import { AGENT_COLOR_PALETTE } from '@/lib/constants/agent-defaults';
+import { normalizeVoiceDesign } from '@/lib/audio/voxcpm';
 
 const log = createLogger('Agent Profiles API');
 
@@ -128,6 +129,10 @@ Requirements:
   - Use the "path" value as the avatar field in the output
 - Each agent must be assigned one color from this list: ${JSON.stringify(AGENT_COLOR_PALETTE)}
   - Each agent must have a different color
+- Each agent needs a "voiceDesign" object describing their VOCAL identity (not personality), written following the language directive and consistent with the persona, as three short comma-free phrases:
+  - "identity": gender + age + role (e.g. "middle-aged male teacher")
+  - "texture": pitch + vocal quality (e.g. "warm low-pitched slightly husky")
+  - "delivery": emotion + pace (e.g. "calm measured encouraging")
 ${voicePrompt}
 
 Return a JSON object with this exact structure:
@@ -137,6 +142,7 @@ Return a JSON object with this exact structure:
       "name": "string",
       "role": "teacher" | "assistant" | "student",
       "persona": "string (2-3 sentences)",
+      "voiceDesign": { "identity": "string", "texture": "string", "delivery": "string" },
       "avatar": "string (from available list)",
       "color": "string (hex color from palette)",
       "priority": number (10 for teacher, 7 for assistant, 4-6 for student)${voiceJsonField}
@@ -170,6 +176,7 @@ Return a JSON object with this exact structure:
         color: string;
         priority: number;
         voice?: string;
+        voiceDesign?: unknown;
       }>;
     };
 
@@ -211,6 +218,8 @@ Return a JSON object with this exact structure:
         }
       }
 
+      const voiceDesign = normalizeVoiceDesign(agent.voiceDesign);
+
       return {
         id: `gen-${nanoid(8)}`,
         name: agent.name,
@@ -221,6 +230,7 @@ Return a JSON object with this exact structure:
         priority:
           agent.priority ?? (agent.role === 'teacher' ? 10 : agent.role === 'assistant' ? 7 : 5),
         ...(voiceConfig ? { voiceConfig } : {}),
+        ...(voiceDesign ? { voiceDesign } : {}),
       };
     });
 
--- a/app/api/generate/tts/route.ts
+++ b/app/api/generate/tts/route.ts
@@ -111,7 +111,8 @@ export async function POST(req: NextRequest) {
     };
 
     log.info(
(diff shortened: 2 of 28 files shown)


wyuc:
**Self-review before requesting review (high-effort, multi-angle).** Round 1 surfaced 6 actionable items, all fixed; a verification pass found no regressions.

1. Strip parentheses from the voice-design prompt so a paren can't break the `(prompt)text` bootstrap delimiter.
2. Gate `POST /api/generate/voice` on `isServerTTSProviderDisabled` (#665) — the `tts` route enforced it, this sibling didn't.
3. Check `voiceExists` before re-registering a client-cached clip (no redundant per-session re-upload).
4. Dedup concurrent `ensureRegisteredVoice` via an in-flight promise map.
5. Warm up only the narrator, not every generated agent.
6. Collapse the duplicated vLLM-Omni speech payload.

**Accepted (follow-ups, not blocking this PR):**
- Stage snapshot persists `persona` but not `voiceDesign`; a reopened generated course resolves a stable persona-based voice rather than its original descriptor voice. Persisting the full agent profile on the stage (and sourcing narrator resolution from the stage instead of the global registry, which would remove the `pickNarratorAgent` heuristic) is the clean follow-up.
- Preset agents fall back to persona as the voice seed: stable but generic (persona isn't a vocal spec); per-preset descriptors are a follow-up.
- Settings "Test TTS" / popover previews use the inline path (no registration) — provider-config test surfaces without an agent.
- base64↔Blob helpers and the `/tts`↔`/voice` provider-config prelude are duplicated; extracting shared helpers is a follow-up.

cosarah (commented):
One issue: `registeredThisSession` / `inFlight` in `lib/audio/voice-registration-client.ts` are keyed only by `voiceId`. If the user changes the VoxCPM backend/base URL in the same browser session, `ensureRegisteredVoice()` can skip revalidation and return an id registered on the old backend; `lib/audio/tts-providers.ts` then sends that stale `registeredVoiceId` and ignores the inline prompt, so the new backend can fail with voice-not-found instead of re-registering/falling back. Please include backend/config identity in the memoization key or revalidate on config changes.

cosarah (commented):
Small remaining issue: `memoKeyFor()` in `lib/audio/voice-registration-client.ts` keys the session/in-flight memo by `voiceId` and `ttsBaseUrl`, but not `ttsApiKey`. Since voice registration/existence checks and later speech calls are authenticated, switching API keys/accounts on the same base URL can still reuse a `registeredVoiceId` from the old credentials and skip revalidation. Please include an API-key identity in the memo key; memory-only or hashed is fine.

cosarah (approved):
LGTM, verified locally

PR #675, opened 2026-06-07
fix(export): include line control points in getElementRange

### LINKED ISSUE #674: getElementRange clips line elements with curve/elbow control points
### Summary

`getElementRange` in `lib/utils/element.ts` computes a too-small bounding box
for `line` elements that have control points, and also assumes the nearer
endpoint sits at offset 0.

### Code

```ts
if (element.type === 'line') {
  minX = element.left;
  maxX = element.left + Math.max(element.start[0], element.end[0]);
  minY = element.top;
  maxY = element.top + Math.max(element.start[1], element.end[1]);
}
```

It only looks at `start`/`end` (and only via `Math.max`), and uses
`minX/minY = left` — assuming `min(start, end) === 0`.

### Impact

`PPTLineElement` carries optional `broken`, `broken2`, `curve`, `cubic` control
points (see `lib/types/slides.ts`). A curved or elbow line bows out past its
endpoints, so its computed range is too small. That clips the line on PPTX
export (`lib/export/use-export-pptx.ts` calls `getElementRange`) and under-sizes
its selection box in the editor (`useMouseSelection.ts`).

Example: a line `{ left: 100, top: 100, start: [0,0], end: [40,0], curve: [20,60] }`
reports `maxY = 100`, but the curve reaches `y = 160`.

This is the same class of bug as the SVG bounding-box issue fixed in #656.

### Fix

Collect every point the line can reach (both endpoints plus any present control
points) and take `Math.min`/`Math.max` on each axis. Backward-compatible for
plain `[0,0] → [w,h]` lines (their min offset is already 0).

ly-wang19:
## What & why

`getElementRange` (`lib/utils/element.ts`) computed a `line` element’s bounding box as `Math.max(start, end)` with `minX/minY = left` — ignoring the optional `broken`/`broken2`/`curve`/`cubic` control points a line can bow out to, and assuming the nearer endpoint sits at offset 0.

`PPTLineElement` carries those control points (`lib/types/slides.ts`), so a curved or elbow line extends past its endpoints and its range came out too small — clipping it on PPTX export (`use-export-pptx.ts`) and under-sizing its selection box in the editor (`useMouseSelection.ts`). Same class of bug as the SVG bbox fix in #656.

Closes #674.

## Fix

Collect every reachable point (both endpoints + any present control points) and take `min`/`max` per axis. Backward-compatible for plain `[0,0] → [w,h]` lines (their min offset is already 0), so the common case is unchanged.

## Test plan

New `tests/utils/element.test.ts` (5 cases, all green):
- plain endpoint-only line is unchanged;
- box uses the min of the endpoints (not just `left`);
- box expands for a quadratic `curve` control point (the headline clip);
- includes `broken`/elbow points;
- includes both `cubic` bezier points, incl. negative offsets.

`npx tsc --noEmit`, `prettier --check`, `eslint` all clean; full suite unaffected.

---

*AI-assisted (Claude), self-reviewed: verified the control-point fields on `PPTLineElement` and that the fix is a no-op for normal `[0,0]→[w,h]` lines before committing.*

### DIFF
--- a/lib/utils/element.ts
+++ b/lib/utils/element.ts
@@ -253,8 +253,13 @@ export const getLineElementPath = (element: PPTLineElement) => {
     const mid = element.broken.join(',');
     return `M${start} L${mid} L${end}`;
   } else if (element.broken2) {
-    const { minX, maxX, minY, maxY } = getElementRange(element);
-    if (maxX - minX >= maxY - minY)
+    // Pick the elbow orientation from the endpoint extent, computed locally.
+    // getElementRange now folds in control points (#674); reusing it here would
+    // let the broken2 point itself flip the orientation. This preserves the
+    // exact pre-#674 heuristic (max of the endpoint offsets per axis).
+    const extentX = Math.max(startArr[0], endArr[0]);
+    const extentY = Math.max(startArr[1], endArr[1]);
+    if (extentX >= extentY)
       return `M${start} L${element.broken2[0]},${startArr[1]} L${element.broken2[0]},${endArr[1]} ${end}`;
     return `M${start} L${startArr[0]},${element.broken2[1]} L${endArr[0]},${element.broken2[1]} ${end}`;
   } else if (element.curve) {
--- a/lib/utils/element.ts
+++ b/lib/utils/element.ts
@@ -82,10 +82,34 @@ export const getElementRange = (element: PPTElement) => {
   let minX, maxX, minY, maxY;
 
   if (element.type === 'line') {
-    minX = element.left;
-    maxX = element.left + Math.max(element.start[0], element.end[0]);
-    minY = element.top;
-    maxY = element.top + Math.max(element.start[1], element.end[1]);
+    // Include every point the line can reach — endpoints AND the optional
+    // control points (broken/broken2 elbows, quadratic curve, cubic bezier) —
+    // and take min/max on both axes. The previous code used only
+    // `Math.max(start, end)` with `minX/minY = left`, which clipped curved or
+    // elbow lines (their control points extend past the endpoints) and assumed
+    // the nearer endpoint always sits at offset 0.
+    const xs = [element.start[0], element.end[0]];
+    const ys = [element.start[1], element.end[1]];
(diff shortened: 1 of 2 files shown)


wyuc (changes requested):
Nice catch folding the control points into the range — the endpoint/`broken`/`curve`/`cubic` handling all look correct, and decoupling the `broken2` elbow orientation from `getElementRange` is the right call. One blocking issue before merge:

**`broken2` over-inflates the bbox on the axis the elbow never draws.**

`getLineElementPath` renders a `broken2` elbow using only **one** coordinate of the control point, picked by orientation:
- horizontal (`extentX >= extentY`): `M start L broken2[0],start[1] L broken2[0],end[1] end` → uses `broken2[0]` only; `broken2[1]` is never drawn.
- vertical: uses `broken2[1]` only; `broken2[0]` is never drawn.

But `getElementRange` unconditionally pushes `broken2[0]` into `xs` **and** `broken2[1]` into `ys`, so the off-axis coordinate stretches the box along an axis the line never occupies.

Concrete case (`left=top=0`): `start=[0,0]`, `end=[100,20]`, `broken2=[50,500]`. Rendered path is `M0,0 L50,0 L50,20 100,20` (max y = 20), but `getElementRange` returns `maxY = 500` — a selection/export box ~25× taller than the visible line, which then feeds the wrong dimensions into the PPTX export.

This is the same "`broken2` is special" reasoning you already applied to the orientation decoupling — it just needs to extend to the range too. Suggested fix: mirror the orientation when including `broken2` — add `broken2[0]` to `xs` only on the horizontal route, and `broken2[1]` to `ys` only on the vertical route (or derive the elbow corner set directly).

Also worth a test: the existing `[50,500]` case only asserts the path, not the range, so this slipped through. A `getElementRange` assertion on a `broken2` line would lock it down.

Everything else — coordinate space, plain-line tightening, `cubic` unioning both control points, the convex-hull bound for `curve`/`cubic` — looks correct. Happy to help iterate.

wyuc:

PR #685, opened 2026-06-08
feat(vocational): add task engine and procedural skill widgets

jackefn:
## Summary

Add an independent vocational education task-engine generation path and procedural-skill widget support.

## Related Issues

N/A

## Changes

- Add `taskEngineMode` support for vocational task generation and route vocational requests through a dedicated task-engine outline prompt.
- Add an experimental vocational test toggle on the home page that enables task-engine mode.
- Add `procedural-skill` widget type support, including widget outline fields for procedure type, task, tools, ordered steps, success criteria, and error consequences.
- Add procedural-skill widget content prompts with required learner controls, feedback panel, reset path, stable selectors, and existing platform message contracts.
- Extend widget teacher-action prompts so generated actions can target procedural-skill widgets through existing `SET_WIDGET_STATE`, `HIGHLIGHT_ELEMENT`, `ANNOTATE_ELEMENT`, and `REVEAL_ELEMENT` message types.
- Add safeguards and prompt contracts so normal interactive generation does not silently emit procedural-skill widgets unless vocational task-engine mode is enabled.
- Add focused tests for task-engine routing, procedural-skill prompt contracts, widget thinking metadata, and supported widget type behavior.

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [x] CI/CD or build changes

### Steps to reproduce / test

1. Check i18n key alignment:
   `pnpm check:i18n-keys`
2. Run vocational-module focused tests:
   `pnpm vitest run tests/generation/procedural-skill-routing.test.ts tests/generation/task-engine-outline-route.test.ts tests/generation/widget-thinking-policy.test.ts tests/prompts/diagram-content-contract.test.ts tests/prompts/game-content-contract.test.ts tests/prompts/interactive-outlines-widget-types.test.ts tests/prompts/procedural-skill-content-quality-contract.test.ts tests/prompts/task-engine-outlines.test.ts tests/prompts/widget-teacher-actions-procedural-skill.test.ts`
3. Run project checks:
   `pnpm check`
   `pnpm lint`
   `npx tsc --noEmit`

- `pnpm check:i18n-keys` passed.
- Focused vocational tests passed: 9 test files, 30 tests.
- `pnpm check` passed.
- `pnpm lint` passed with existing warnings only; no lint errors.

### Evidence

- [x] Manually tested locally

Local verification logs:

```text
pnpm check:i18n-keys
i18n key alignment check passed (7 locale files, source: en-US.json).

pnpm vitest run ...vocational focused tests...
Test Files  9 passed (9)
Tests       30 passed (30)

pnpm check
All matched files use Prettier code style!

pnpm lint
0 errors, 4 warnings

npx tsc --noEmit
passed
```

## Checklist


wyuc (changes requested):
Thanks for splitting this out of the larger PR and for the thorough test coverage and local verification. The `procedural-skill` widget contract is genuinely well thought through: the decision/judgment requirement, the consequence-feedback path, success-criteria gating, the null-safe DOM rules, and reusing the existing four `postMessage` types instead of inventing a new protocol are all the right calls. The widget itself is the strong part of this PR.

My main request is about **how this lands**, plus a set of correctness and scope issues to resolve first.

## Overall recommendation: gate the whole path behind a server-side flag, default off

Right now the vocational path is wired directly into the default generation flow as a third outline-planner branch (`taskEngineMode ? TASK_ENGINE_OUTLINES : interactiveMode ? INTERACTIVE_OUTLINES : REQUIREMENTS_TO_OUTLINES`), with a vocational-specific normalizer on one side and a sanitizer on the other. That shape does not scale: a fourth or fifth teaching mode means another boolean, another nested ternary, another bespoke normalizer, and retrofitting every other mode's sanitizer to stay mode-aware.

Suggested approach:

1. Put the entire task-engine path behind a **server-side feature flag, defaulting to off**. While off, generation behavior is byte-for-byte identical to today, so this can merge without changing the default product or risking existing courses.
2. Treat full availability as a follow-up. We would like to introduce a general pedagogy/course-mode routing mechanism in OpenMAIC first (an explicit course mode that carries its own allowed widget set, scene-structure targets, suitability gate, and post-normalization). Once that exists, vocational becomes one pluggable mode (similar to a skill/plugin) populated as data, rather than a hardcoded branch. We are happy to align with you on that interface so this work slots into it cleanly.

This keeps your widget and prompts in the tree now, and avoids baking a routing pattern we will have to unwind later.

## Must fix before merge

**1. Remove the changes that are unrelated to vocational generation.** Several edits in this PR change behavior for *all* users and are not gated by the new mode. They each deserve their own PR with their own contract tests and rollback boundary:

- `app/api/generate/scene-content/route.ts`: the hardcoded `if (providerId === 'glm')` thinking-disable. A provider special-case does not belong inside a generation route, and the fact that the same guard would have to be repeated on other generation paths (e.g. the classroom path builds its own `aiCall` with no such guard) shows it is at the wrong layer. If GLM with thinking genuinely degrades scene content, please handle it through provider/thinking-capability configuration in a separate change. Drop it from this PR.
- `lib/prompts/templates/diagram-content/*` and `game-content/*`: the stable-SVG-id contract, visibility/JSON contracts, and the playable-payload contract. These look like reasonable improvements, but they change every diagram and game generation. As one concrete regression risk, the diagram example icon was changed from an emoji glyph (`🎯`) to a plain word (`target`); since nothing in the prompt maps an icon name to a glyph, the model is now steered to render literal words like "target" into the SVG. Please move these to a dedicated prompt-quality PR.
- `lib/prompts/templates/interactive-outlines/user.md`: adding `visualization3d` to the ordinary Interactive Mode distribution and required-field allowlist changes the widget mix for all interactive courses. Separate, deliberate change.

**2. `errorConsequences` is computed but never reaches content generation.** The outline normalizer defaults and persists `errorConsequences`, the type carries it, and a test asserts it is populated, but `generateWidgetContent`'s `procedural-skill` case does not pass it into the prompt variables, and `procedural-skill-content/user.md` has no `{{errorConsequences}}` placeholder. The consequence/GO-STOP framing is the core value of the widget, so this data currently dead-ends. Either thread it through to the content prompt or drop the field.

**3. Widget-outline defaults diverge from the existing builder.** The new normalizers set `language: 'javascript'` (code) and `gameType: 'strategy'` (game), while the existing `buildWidgetOutline` and `generateWidgetContent` in `lib/generation/scene-generator.ts` use `'python'` and `'quiz'`. The same default now lives in three places with two different values. Please reuse the existing default builder rather than redefining defaults in the route.

**4. The persisted mode is mislabeled.** On submit, vocational mode forces `interactiveMode: true`, which is persisted on the stage and rendered as the "Interactive Mode" badge, while `taskEngineMode` is never persisted at all. A vocational course is therefore stored and displayed as an interactive course, with no record that it was vocational. Persist the actual mode and avoid overloading `interactiveMode`.

## Design (worth addressing, can follow the flag)

**5. The keyword-regex classifier in the route.** `looksLikeVocationalTask` is a frozen bilingual keyword list used as the fallback classifier. It misfires in both directions: common English words like `go`, `stop`, `tool`, and `safety` cause an ordinary scene to be force-converted into a fabricated procedural-skill widget, and a genuinely vocational scene whose wording misses the list is downgraded to a static slide. It also second-guesses the `widgetType` the planner already chose. Prefer trusting the planner's `widgetType` and a config-driven gate over an in-route regex.

**6. Gating by rewrite vs. gating by allowlist.** `sanitizeNonTaskEngineOutline` lets the model spend tokens producing a `procedural-skill` outline and then rewrites it into a diagram. Combined with the prompt-level "do not emit it" instructions, the gate is split across two places that can drift, and it only exists on the streaming route. A per-mode allowed-widget set, so the generator simply cannot select `procedural-skill` when the mode is off, would remove both the rewrite code and the per-path guard problem.

**7. Normalizer duplication.** The seven `normalizeTaskEngine*Outline` functions are near-identical (spread, set `type: 'interactive'`, set one `widgetType`, apply per-field defaults) and overlap the existing `buildWidgetOutline` / outline-fallback logic. A single table-driven normalizer keyed by widget type would remove most of this and is what causes issue #3.

## Nits

- The task-engine `system.md` and `user.md` restate each other's rules (scene counts, ratios, first-slide layout, self-check), and the NEV-A12 example appears three times across the two files. Keep the rubric in one place.
- The `procedural-skill-content` prompt re-pastes the platform/iframe `postMessage` contract that already lives in other widget content prompts, where it has already drifted (`event.data.type` vs `event.data.action`). Consider a shared prompt fragment.

Overall: the widget and its prompt contract are solid and we want them in. The asks are to unbundle the unrelated global changes, fix the four correctness items, and land the vocational path behind a default-off server flag so we can open it up properly once the pedagogy-routing mechanism is in place. Happy to pair on the flag and on that routing interface.

jackefn:
Thanks for the detailed review. I have pushed an update addressing the requested changes. I kept the larger pedagogy/course-mode routing mechanism out of this PR and only added the default-off flag/gating needed for this landing step.

## Main changes:

- Put the vocational/task-engine path behind a server-side feature flag: `OPENMAIC_ENABLE_VOCATIONAL === "true"`, defaulting to off.
- When the flag is off, `taskEngineMode` is ignored and generation silently falls back to the existing standard / interactive behavior.
- Kept the public UI flag separate: `NEXT_PUBLIC_SHOW_VOCATIONAL_TEST_UI` only controls whether the test toggle is shown, and is not used as the server-side gate.

## I also removed the unrelated global changes from this PR:

- Removed the GLM scene-content thinking override.
- Reverted the diagram-content prompt hardening changes.
- Reverted the game-content playable-payload/global prompt changes.
- Reverted the ordinary interactive-outline `visualization3d` distribution / allowlist changes.

## Correctness fixes:

- Threaded `errorConsequences` through procedural-skill content generation and added it to the procedural-skill prompt.
- Removed route-level default drift for ordinary widget types, so ordinary widget defaults are no longer redefined in the task-engine route.
- Removed the keyword-regex classifier that promoted outlines into `procedural-skill`.
- Persisted `taskEngineMode` separately from `interactiveMode`.
- Added a small `Vocational Mode` badge for saved vocational stages, while ordinary interactive courses still show the existing `Interactive Mode` badge.
- Canonicalized duplicate streamed outline ids before emitting them, so repeated ids such as `scene_4` do not leak to the preview UI.

## Prompt cleanup:

- Reduced duplication between the task-engine system/user prompts.
- Consolidated the NEV-A12 example.
- Narrowed the procedural-skill postMessage wording to the existing `event.data.type` contract.

## Validation:

- `pnpm check:i18n-keys`
- `pnpm check`
- `pnpm lint`
- `pnpm exec tsc --noEmit`
- `pnpm test`

All passed. Lint has 0 errors and 4 pre-existing warnings.

wyuc:
Thanks, this is a strong turnaround. I re-reviewed `e6debb09` and verified the main asks: the default-off flag, the reverted scope-creep (GLM thinking, diagram/game/visualization3d prompts), the `errorConsequences` threading, the removed keyword classifier, and the separated `taskEngineMode` persistence all look right, and CI is green. Three items remain; only the first is blocking.

**1. (blocking) The flag gates outline-prompt selection, not content generation.** `isVocationalTaskEngineEnabled()` is read only in `scene-outlines-stream/route.ts`, to pick the planner. The content stage does not re-check it: `generateWidgetContent`'s `case 'procedural-skill'` has no guard, and it is reachable with the flag off via paths that bypass the outline route:

- a replayed/persisted session, where `app/generation-preview/page.tsx` reuses `currentSession.sceneOutlines` and posts straight to `/api/generate/scene-content`;
- a direct `scene-content` POST with a client-supplied outline;
- the `lib/server/classroom-generation.ts` job path, where `applyOutlineFallbacks` does not strip `procedural-skill`.

So "flag off ⇒ vocational unreachable" is not structural. An outline that already carries `widgetType: 'procedural-skill'` (for example generated once with the flag on, then the flag turned off, then replayed) will still render the widget. Suggest computing the effective decision once (e.g. `resolveVocationalActive(requirements)`) and gating the content stage as well, or having `generateWidgetContent` refuse/downgrade `procedural-skill` when the flag is off, so the guarantee holds regardless of where the outline came from.

**2. The new "Vocational Mode" badge can mislabel.** `taskEngineMode` is persisted from the client request (`generation-preview` stores `!!requirements.taskEngineMode`) and the `done` SSE event does not echo the server's effective mode. With `NEXT_PUBLIC_SHOW_VOCATIONAL_TEST_UI` on but `OPENMAIC_ENABLE_VOCATIONAL` off, generation correctly falls back to the standard path, but the stage is saved with `taskEngineMode: true` and the homepage renders the "Vocational Mode" badge over non-vocational content. Echo the effective mode in `done` and persist that instead of the requested flag.

**3. Flag parsing is inconsistent.** `isVocationalTaskEngineEnabled()` uses `process.env.OPENMAIC_ENABLE_VOCATIONAL === 'true'`, bypassing the file's own `readBoolean()` helper (whose header comment documents `'true'` or `'1'` as truthy, and which the sibling readers use). `OPENMAIC_ENABLE_VOCATIONAL=1` would silently leave the feature off. Use `readBoolean` for consistency.

Item 1 is worth fixing before merge since it is the actual flag-off guarantee; 2 and 3 are small. Once 1 is addressed I am happy to approve.

wyuc (approved):
Verified `50c0aefd`. All three follow-up items are addressed cleanly:

1. **Content-stage gate (the blocking one).** `generateWidgetContent` now takes `allowProceduralSkill` (defaults false) and refuses a `procedural-skill` outline with a warning + null return before it ever builds the prompt, and `applyOutlineFallbacks` sanitizes a stray procedural-skill outline into a diagram at the fallback stage. The new `resolveVocationalActive(requirements)` = `taskEngineMode && isVocationalTaskEngineEnabled()` is the single server-authoritative decision, and I traced every live caller — `scene-content/route.ts` and `classroom-generation.ts` both derive `allowProceduralSkill` from it (the classroom path has no `taskEngineMode`, so it is always false). With `OPENMAIC_ENABLE_VOCATIONAL` unset there is no remaining path to a procedural-skill widget, including the replay/persisted-session and direct-POST paths.
2. **Badge / effective mode.** The done event now carries the server-effective `taskEngineMode`, the client persists from `resolveTaskEngineModeFromOutlineDoneEvent(...)` rather than the raw request flag, and the badge is set from that. The UI-on / server-off mislabel is gone.
3. **Flag parsing.** `isVocationalTaskEngineEnabled()` now uses `readBoolean`, so `OPENMAIC_ENABLE_VOCATIONAL=1` works.

New tests cover the content-stage gate with real negative assertions (flag off ⇒ procedural prompt never reached / outline downgraded), and CI is green.

One optional, non-blocking follow-up: the sanitize logic now exists in two places (`sanitizeNonTaskEngineOutline` in the outline-stream route and `sanitizeProceduralSkillOutline` in `outline-generator.ts`) with identical behavior — worth consolidating later, but not a blocker.

Thanks for the fast, on-point turnaround. LGTM — approving.

PR #692, opened 2026-06-08
perf(editor): dedupe alignment snap-lines in O(n) instead of O(n²)

### LINKED ISSUE #691: uniqAlignLines is O(n²) (findIndex in a loop) on the drag/scale hot path
### Summary

`uniqAlignLines` (`lib/utils/element.ts`) dedupes alignment snap-lines with an
`Array.findIndex` inside a `forEach`, which is **O(n²)**.

### Code

```ts
export const uniqAlignLines = (lines: AlignLine[]) => {
  const uniqLines: AlignLine[] = [];
  lines.forEach((line) => {
    const index = uniqLines.findIndex((_line) => _line.value === line.value); // O(n)
    ...
  });
  return uniqLines;
};
```

### Impact

It runs on **every drag/scale mousemove** (`useDragElement` / `useScaleElement`)
to merge the snap lines, and the line count grows with the number of nearby
element edges. On element-dense canvases (tens–hundreds of elements) the
quadratic merge is ~n² comparisons per pointer move at ~60 Hz, which shows up as
drag jank.

### Fix

Dedupe in O(n) via a `Map` keyed on `value`. A `Map` preserves first-occurrence
insertion order (re-`set`-ting an existing key keeps its slot), so the output
order and the range-merge semantics are unchanged — behavior-preserving, just
linear.

ly-wang19:
## What & why

`uniqAlignLines` (`lib/utils/element.ts`) merges alignment snap-lines with `Array.findIndex` inside a `forEach` — **O(n²)**:

```ts
lines.forEach((line) => {
  const index = uniqLines.findIndex((_line) => _line.value === line.value); // O(n)
  ...
});
```

It runs on **every drag/scale mousemove** (`useDragElement` / `useScaleElement`), and the snap-line count grows with the number of nearby element edges. On element-dense canvases that’s ~n² comparisons per pointer move at ~60 Hz → drag jank.

Closes #691.

## Fix

Dedupe in **O(n)** via a `Map` keyed on `value`. A `Map` preserves first-occurrence insertion order (re-`set`-ting an existing key keeps its slot), so the **output order and the range-merge semantics are unchanged** — behavior-preserving, just linear.

## Test plan

New `tests/utils/uniq-align-lines.test.ts` (5 cases): dedupe + outer-bound range merge, first-occurrence order preserved, single line, empty input, and many-duplicates collapse. `tsc`/`prettier`/`eslint` clean.

No user-facing strings (no i18n impact). (Separate test file from `element.test.ts` to avoid overlap with #675, which also touches `element.ts`.)


wyuc (approved):
Focused review complete. The Map-based implementation preserves first-seen ordering and the existing outer-range merge semantics while reducing duplicate lookup from O(n²) to O(n). Verified on the refreshed main base: fresh frozen-lockfile install, targeted uniqAlignLines tests (5/5), and TypeScript all pass locally; refreshed GitHub Lint/Typecheck/Unit and E2E checks are green.

PR #694, opened 2026-06-08
test(tts): cover splitLongSpeechText / splitLongSpeechActions

### LINKED ISSUE #693: Add test coverage for the TTS text-splitting utilities (tts-utils.ts)
### Summary

`lib/audio/tts-utils.ts` — the TTS text/action splitting used on the generation
and playback paths (`splitLongSpeechText`, `splitLongSpeechActions`,
`TTS_MAX_TEXT_LENGTH`) — has **no test coverage**. The splitting logic
(sentence → clause → hard-split, with a per-provider length cap) is exactly the
kind of edge-case-prone code that benefits from a regression lock.

### Proposal

Add unit tests covering: sentence-boundary splitting, packing short sentences up
to the limit, clause-level fallback, hard-splitting punctuation-free runs, the
**invariant that no chunk exceeds `maxLength` (and none is empty)**, and the
`splitLongSpeechActions` contract (no-limit provider → untouched; short/non-speech
→ untouched; over-limit → chunked sub-actions with `<id>_tts_<n>` ids, the parent
`audioId` dropped, and text preserved across the split).

I verified the `maxLength` invariant holds against adversarial inputs before
writing the tests — this is coverage for correct behavior, not a bug fix.

ly-wang19:
## What & why

`lib/audio/tts-utils.ts` (TTS text/action splitting on the generation + playback paths) had **no test coverage**. The sentence → clause → hard-split chunking with a per-provider length cap is edge-case-prone and worth a regression lock.

Closes #693.

## Coverage added (`tests/audio/tts-utils.test.ts`, 9 cases)

- `splitLongSpeechText`: within-limit passthrough (trimmed), sentence-boundary split (punctuation kept), packing consecutive short sentences up to the limit, clause-punctuation fallback for over-long sentences, hard-splitting a punctuation-free run, and the **invariant** that no chunk exceeds `maxLength` (and none is empty) across several limits.
- `splitLongSpeechActions`: provider with no limit → actions returned untouched (same ref); short speech + non-speech → untouched; over-limit speech → chunked sub-actions with `<id>_tts_<n>` ids, the parent `audioId` dropped, each chunk ≤ limit, and the text preserved across the split.

## Note

No source change — coverage only. I verified the `maxLength` invariant holds against adversarial inputs (long no-punctuation runs, clause-heavy text, boundary lengths) before locking it in tests, so this documents correct behavior rather than fixing a bug. `tsc`/`prettier`/`eslint` clean.


wyuc (approved):
Focused review complete. The added coverage locks the important TTS splitting boundaries and invariants without changing production behavior. Verified on the latest main (including #696): targeted TTS utility tests pass 9/9 locally, the final diff contains only the new test file, and refreshed GitHub Lint/Typecheck/Unit and E2E checks are green.

PR #696, opened 2026-06-08
feat(generation): parallelize within-scene TTS generation

### LINKED ISSUE #695: Parallelize TTS generation within a scene (follow-up to #660)
### Summary

Follow-up to #660 (opt-in parallel scene **content**). The remaining serial cost
in classroom generation is **TTS**: `generateTTSForScene`
(`lib/hooks/use-scene-generator.ts`) renders a scene's speech clips one at a time
in a `for … await` loop.

### Why it's safe to parallelize

Unlike cross-scene parallelism (ruled out in #572 because it breaks
`previousSpeeches` threading), the speech actions **within a single scene are
independent**: each renders its own audio under its own `audioId`
(`tts_s<order>_<actionId>`), stored separately in IndexedDB, with no ordering or
carry-over dependency. Playback order is determined later by the action list, not
by generation order.

### Proposal

When the server opts into parallel generation, render a scene's speech clips with
**bounded concurrency** (reusing the existing `PARALLEL_SCENE_CONCURRENCY` knob
and the `mapWithConcurrency` helper added in #660) instead of one at a time.
Default (`0`/unset) keeps the strictly-serial behaviour, and the bound keeps TTS
providers' per-key 429 quotas in check — same safety stance as #660.

(A dedicated `PARALLEL_TTS_CONCURRENCY` knob could be a later refinement if TTS
and LLM quota profiles need to diverge; reusing the one knob keeps this change
focused.)

ly-wang19:
## What & why

Follow-up to #660 (just merged — opt-in parallel scene **content**). The remaining serial cost in classroom generation is **TTS**: `generateTTSForScene` rendered a scene’s speech clips one at a time in a `for … await` loop.

**Why it’s safe**: unlike cross-scene parallelism (ruled out in #572 because it breaks `previousSpeeches` threading), the speech actions **within a scene are independent** — each writes its own audio under its own `audioId` (`tts_s<order>_<actionId>`), stored separately, with no ordering/carry-over. Playback order comes from the action list, not generation order.

Closes #695.

## Change

When the server opts into parallel generation (`PARALLEL_SCENE_CONCURRENCY > 1`, the #660 knob), render the scene’s speech clips with **bounded concurrency** via the `mapWithConcurrency` helper #660 added; otherwise the original strictly-serial `for … await` loop runs.

- **Default (`0`/unset) is byte-for-byte the original serial behaviour.**
- The bound keeps TTS providers’ per-key 429 quotas in check — same safety stance as #660.
- Per-clip failures are still counted (not thrown), so one bad clip never aborts the scene — unchanged from before.

Reuses the existing `parallelSceneConcurrency` setting (no new env var / plumbing). A dedicated `PARALLEL_TTS_CONCURRENCY` could be a later refinement if TTS vs LLM quota profiles need to diverge — noted in the issue.

## Test plan

`npx vitest run` — **94 files / 742 tests pass**; `tsc`/`prettier`/`eslint` clean. The concurrency mechanism itself is already unit-tested (`tests/utils/concurrency.test.ts`, incl. the cap); the serial default path is unchanged. (The hook isn’t unit-tested — no DOM/hook harness in the repo — same as #660.)


wyuc (approved):
Focused review complete. Speech actions within one scene use distinct audio IDs and independent storage writes, so bounded parallel execution preserves playback order and failure accounting. The opt-in/default-serial behavior and concurrency cap are preserved. Verified on the refreshed main base with a fresh frozen-lockfile install, concurrency tests (8/8), file lint, full TypeScript, and green GitHub Lint/Unit/E2E checks.

PR #698, opened 2026-06-08
feat(generation): parallelize scene content on the server path

### LINKED ISSUE #697: Apply parallel scene-content generation to the server path (follow-up to #660)
### Summary

Follow-up to #660. That PR added opt-in parallel scene-**content** generation to
the **client** path (`useSceneGenerator`). The **server** path —
`lib/server/classroom-generation.ts`, used by `/api/generate-classroom` and the
background job-runner — is still strictly serial:

```ts
for (const [index, outline] of outlines.entries()) {
  const content = await generateSceneContent(safeOutline, ...);   // serial
  const actions = await generateSceneActions(safeOutline, content, ...);
  createSceneWithActions(...);
}
```

So server-side / background generations get **none** of #660's speedup. #572
listed this server path as the explicit "adopt the same pattern later" item.

### Proposal

Mirror #660's pipeline on the server: pre-warm scene **content** with bounded
concurrency (reusing the server helper `getParallelSceneConcurrency()` and the
`lazyBoundedMap` helper, both from #660), then keep the existing serial loop for
**actions + scene assembly + progress reporting** so scenes stay in outline order
and progress events fire in order.

- Gated by the same `PARALLEL_SCENE_CONCURRENCY` env (default `0`/unset =
  unchanged strictly-serial behaviour).
- Content has no cross-scene dependency, so running it ahead is safe; a content
  failure resolves to `undefined` and is skipped per-scene exactly like today.

ly-wang19:
## What & why

Follow-up to #660 (merged), which added opt-in parallel scene-**content** generation to the **client** path. The **server** path — `lib/server/classroom-generation.ts`, used by `/api/generate-classroom` and the background job-runner — was still strictly serial (`for … of outlines` → `await generateSceneContent` → `await generateSceneActions`), so server-side / background generations got none of #660’s speedup. #572 listed this as the explicit "adopt the same pattern later" item.

Closes #697.

## Change

Mirror #660’s pipeline on the server:
- **Pre-warm scene content** with bounded concurrency via `lazyBoundedMap`, reusing the server-side `getParallelSceneConcurrency()` helper (both added in #660).
- Keep the existing **serial loop** for actions + scene assembly + `onProgress` reporting, so scenes stay in **outline order** and progress events fire in order.

Properties:
- Gated by the same `PARALLEL_SCENE_CONCURRENCY` env. **Default `0`/unset is byte-for-byte the original serial pipeline.**
- Content has no cross-scene dependency, so running it ahead is safe; a content failure resolves to `undefined` and is **skipped per-scene exactly as today** (the parallel callback also catches throws → `undefined`, so one bad scene never aborts the batch).

## Test plan

`npx vitest run` — **742 tests pass**; `tsc`/`prettier`/`eslint` clean. The concurrency mechanism is already unit-tested (`tests/utils/concurrency.test.ts`); the serial default path is unchanged. (`classroom-generation.ts` orchestrates LLM/IO and has no existing unit test — same as the rest of the server pipeline.)

### DIFF
--- a/lib/server/classroom-generation.ts
+++ b/lib/server/classroom-generation.ts
@@ -15,6 +15,8 @@ import type { AICallFn } from '@/lib/generation/pipeline-types';
 import type { AgentInfo } from '@/lib/generation/pipeline-types';
 import { getDefaultAgents } from '@/lib/orchestration/registry/store';
 import { createLogger } from '@/lib/logger';
+import { lazyBoundedMap } from '@/lib/utils/concurrency';
+import { getParallelSceneConcurrency } from '@/lib/server/provider-config';
 import { isProviderKeyRequired } from '@/lib/ai/providers';
 import { resolveClassroomWebSearchConfig } from '@/lib/server/web-search-config';
 import { resolveModel } from '@/lib/server/resolve-model';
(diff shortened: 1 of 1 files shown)


wyuc (changes requested):
Nice — I verified the parallelization is real (`lazyBoundedMap` kicks off the first `limit` scenes eagerly via the semaphore's synchronous `pump()`, so awaiting `contentPromises[index]` in the serial loop doesn't gate later scenes), the concurrency is bounded, ordering/progress stay in outline order, there's no unhandled-rejection path, and the default (`PARALLEL_SCENE_CONCURRENCY` unset) path is byte-equivalent. No shared-state races either (`aiCall` is stateless, `agents` is read-only, agent RNG runs only in the serial actions phase).

One blocking issue before merge:

**The parallel path silently changes failure semantics, and the comment describing it is inaccurate.**
In the default serial path, `await generateSceneContent(...)` is *not* wrapped in try/catch, so a thrown provider/network/parse error rejects and aborts the whole job (fail-fast). In the parallel path, the task lambda catches the throw and returns `undefined`, so that scene is skipped and the job can still complete "successfully" with an incomplete set of scenes. The comment says content failure is "skipped per-scene (same as the serial path)", but the serial path skips only on a falsy *return* — a *throw* aborts it. So enabling the opt-in doesn't just parallelize; it also converts fail-fast into partial-success, which can persist incomplete classrooms as successful.

Could you either:
- re-throw content errors in the parallel path so it preserves the serial fail-fast behavior (parallelism shouldn't change error handling); or
- if partial-success is intended, make both paths consistent and update the comment + document that the opt-in changes failure semantics.

Minor (non-blocking): peak in-flight `aiCall` count is `concurrency + 1`, since the serial loop's `generateSceneActions` call overlaps the background content calls. Sizing `PARALLEL_SCENE_CONCURRENCY` to exactly a provider's per-key quota can momentarily hit `quota + 1` and see sporadic 429s — worth a note in the env var's doc comment.

The mechanism itself is solid; it's just the error-handling divergence I'd like resolved. Thanks!

wyuc:

PR #699, opened 2026-06-08
fix(agent): respond to the user's turn before lecturing (+ answer-content eval)

wyuc:
## Problem

The agent (teacher) often does not respond to the user's latest message — it greets, opens a lecture, or keeps narrating the current slide instead of answering the question or honoring the request. Routing is usually correct (the director picks the teacher), so the existing director question-answering eval (#511/#599) cannot catch it: it never generates or judges the agent's actual reply.

## Two parts, in one PR

**1. A new `answer-content` eval (the measuring stick).**
- Runs the real agent-generate inputs — `buildStructuredPrompt()` + `convertMessagesToOpenAI()` — against the model, parses output with the app's runtime parser, and LLM-judges two signals: `leads_with_answer` (first sentence addresses the literal ask) and `answered_anywhere`. Their gap quantifies "drift-then-answer".
- A/B mirrors the `answering-runner` rule-13 strip: `baseline` removes the section under test, `with_rule` is as-shipped.
- Scenarios are synthetic/anonymized with a live-shaped classroom context (topical slide deck, stage + `languageDirective`, student profile), covering opening-lecture override, ignored format/capability/navigation requests, ignored correction, frustration re-ask, adjacent pivot, vague-clarify, plus clean controls.

**2. The fix (prompt).**
- Generalizes the agent-system `# Answering the User's Question` section into `# Responding to the User's Turn`: the user's latest message always takes priority over continuing the lecture, with explicit handling for navigation/pacing, format/language, can't-do capability requests, corrections, frustration, and vague asks.

## Result

`gemini-3-flash-preview`, n=10 samples/variant:

| Variant | Mean leads-with-answer |
|---|---|
| baseline (section stripped) | 42% |
| with_rule (as-shipped) | **93%** |

All 12 scenarios pass; answered-anywhere 95%. The eval was used red→green to develop the fix and to surface two eval-fairness bugs (technical-term code-switching; verbal navigation when no page-turn action exists), now corrected in the judge.

## Run

```
EVAL_AGENT_MODEL=<provider:model> EVAL_JUDGE_MODEL=<provider:model> \
  pnpm eval:orchestration:answer-content
```


wyuc:
**Review loop summary (converged)**

This PR pairs a behavior fix with the eval that measures it.

**Fix** — generalized the agent-system `Answering the User's Question` section into `Responding to the User's Turn`: the user's latest message takes priority over continuing the lecture, with explicit handling for navigation/pacing, format/language, can't-do capability requests, corrections, frustration, and vague asks; plus a carve-out so a *standalone* acknowledgement doesn't trigger a manufactured reply.

**Eval** — new `answer-content` harness: runs the real agent-generate inputs (`buildStructuredPrompt` + `convertMessagesToOpenAI`), parses with the runtime parser, and LLM-judges `leads_with_answer` vs `answered_anywhere`; the A/B strips the section under test. Scenarios are synthetic with live-shaped classroom context (slide deck + stage + student profile).

**Result** (`gemini-3-flash-preview`, n=10): mean leads-with-answer **42% → 93%**, 11/12 scenarios pass.

**Reviewed over 3 rounds; fixed:** prompt over-broadness (no acknowledgement carve-out; conflict with agent-initiated discussions), then the carve-out being too loose (ack+request hybrids, pacing-word collision); eval empty-scenario set producing a false PASS; `extractTexts` surfacing action-JSON as fake text; plus minor dedup/dead-code cleanups.

**Accepted limitations (by design):** the LLM judge can't see deck order, so the navigation rubric trusts a verbal transition; an all-errored scenario currently scores 0% (visible in per-scenario detail); the `chem` opener sits near the model's reliability boundary (~60–80% across runs) and was deliberately not over-tuned — use n≥10 for stable aggregates. The harness is offline (not wired into CI).

---

@cosarah — would you mind doing some hands-on conversation testing on top of the automated eval? The harness is synthetic and single-turn by design, so real multi-turn dialogue is the gap it can't cover. Especially valuable: mid-lecture interruptions ("skip to the next page", "explain in <language>", "make a video"), corrections, frustration ("you didn't answer my question"), and bare acknowledgements ("ok" / "嗯") to confirm the agent keeps teaching rather than over-responding. Thanks!

cosarah (commented):
**Findings**

- P2: `lib/prompts/templates/agent-system/system.md:59` tells agents to “do it” for navigation requests, and `eval/orchestration/answer-content-judge.ts:42` counts a verbal transition as success, but the agent action schema has no slide-navigation action. As a result, “skip to next page” can pass the eval while the UI stays on the same `currentSceneId`. Either expose and execute a real navigation action, or make the prompt/judge require saying the agent cannot directly change slides.

- P2: `eval/orchestration/answer-content-judge.ts:79` coerces judge fields with `Boolean(...)`. If the judge returns `"false"` as a string, it becomes `true`, falsely passing samples. Parse only actual booleans, or strictly normalize `"true"`/`"false"` and mark malformed output as an error.

- P3: `eval/orchestration/answer-content-runner.ts:297` excludes errored samples from the denominator. A scenario can pass with one good sample and the rest generation/judge errors. For an eval gate, errors should count as failures or fail the scenario when usable sample count is below requested `EVAL_SAMPLES`.

wyuc:
Thanks @cosarah — addressed all three in d4cb7e1:

- **Navigation (P2):** went with the honest reframe. Agents have no slide-navigation action, so the prompt, judge, and scenario now require the reply to *acknowledge* the request and be honest that it can't change the slide itself (continue with the next point verbally / tell the user how to navigate). Pretending it flipped the slide no longer passes. (Wiring a real navigation action is a larger change I kept out of this PR's scope.)
- **Judge booleans (P2):** parsed strictly now (real boolean or `"true"`/`"false"`); malformed judge output is flagged as an error instead of coerced.
- **Error accounting (P3):** errors count as failures — the denominator is all requested samples, so a scenario can't pass on a single good sample.

Re-ran (gemini-3-flash-preview, n=10): navigation **0% → 100%** under the honest framing; aggregate leads-with-answer ~93%, no regression. One caveat: the opener scenarios (math/chem) sit near the model's reliability boundary (~40–100% across runs), so their per-scenario verdict flickers — documented as a known limitation; use n≥10 for stable aggregates.

wyuc:
@cosarah friendly ping 🙂 — all three findings are addressed in d4cb7e1 (details in the comment above) and CI is green. Could you take another look when you get a chance? Thanks!

cosarah (commented):
Findings:

1. P2 `eval/orchestration/answer-content-runner.ts:272`
`leadText` is `texts[0]`, meaning the first JSON text block, not the first sentence. Since the prompt encourages 2-3 short sentences and models often put them in one `type:"text"` object, a response like "Today we'll discuss parabolas. The formula is x=-b/(2a)." can be judged as leading with the answer even though the first sentence drifted. That weakens the exact "first sentence drifts, second answers" regression this eval is meant to catch. Please split/extract the first sentence from the first text block before judging `leads_with_answer`.

2. P3 `eval/orchestration/answer-content-runner.ts:63`
The eval hardcodes `TEACHER_ACTIONS` and omits canonical teacher actions such as `play_video` and `wb_draw_chart` from `lib/orchestration/registry/types.ts:58`. The runner claims to exercise the real agent-generate prompt path, but the available-action section differs from production. Please import `ROLE_ACTIONS.teacher` or `getActionsForRole('teacher')` so benchmark prompt fidelity does not drift.

cosarah (commented):
Findings:

1. P2 `eval/orchestration/answer-content-runner.ts:247`
`leadFromTexts()` now limits judging to the first two sentences, but still only inside `texts[0]`. Structured agent output explicitly allows multiple `type:"text"` blocks interleaved with actions, so a valid reply like `[{"type":"text","content":"Sure."},{"type":"text","content":"The derivative is 2x."}]` will be judged on only `"Sure."` and fail `leads_with_answer`, even though it follows the prompt's allowed acknowledgement-then-answer pattern. It should collect the first two speech sentences across consecutive/ordered text chunks, not just the first text block.

cosarah (approved):
Approved. The current eval result is within the expected range.

Issue #621, opened 2026-05-30
RFC: MAIC ETL — document extraction and knowledge-source pipeline for OpenMAIC


wyuc:
## Summary

OpenMAIC currently treats uploaded course material as a PDF-centered path: upload a single PDF, parse it with a PDF provider, truncate the extracted text if needed, and pass `pdfText` / `pdfImages` into generation.

That model has served the first version well, but it is now limiting several related efforts:

- multi-format course material upload
- MinerU support beyond PDF
- multi-document course material bundles
- long textbook handling and chapter-aware preprocessing
- question bank / reference material support
- future RAG and knowledge-source workflows

This RFC proposes **MAIC ETL** as a restrained, incremental framework for these capabilities.

The goal is not to build a full knowledge-base product first. The goal is to generalize the existing PDF parsing path into a document extraction pipeline, using MinerU multi-format support as the first reference implementation.

## Motivation

Several current issues and PRs point to the same architectural pressure:

- #332 notes that the upload pipeline is hardcoded around PDF, and that PRs touching new formats or document processing conflict with each other.
- #611 adds TXT/DOCX/PDF parsing, but still needs a stable place to land that does not keep extending the old PDF-specific model.
- #41 and #140 request MD/TXT and DOCX support.
- #148 and #160 request multiple PDFs per generation.
- #52 and #65 request long-document summarization instead of simple truncation.
- #335 describes a real textbook use case where large illustrated documents need chapter slicing and image filtering.
- #43 requests question-bank reference material during generation.
- #258 already made web search query rewriting aware of uploaded PDF text, showing that downstream features need document context.
- #515 shows an external knowledge-graph-to-classroom workflow that could later fit into the same source framework.

These should not become a pile of unrelated feature patches. They should gradually form a small standard layer inside OpenMAIC.

## Proposed Direction

Introduce **MAIC ETL** as a staged roadmap:

```text
Source / File
  -> Extractor Provider
  -> DocumentArtifact
  -> DocumentBundle
  -> Transform
  -> Index / Retrieve
  -> Generation / Classroom QA
```

The first milestone should stay narrow:

```text
file -> document extractor provider -> DocumentArtifact
```

No vector database, no full RAG workspace, and no general knowledge-base UI in the first milestone.

## Milestone 0: Rename the Boundary

Generalize the current PDF-specific concepts into document-level concepts.

Examples:

- `PDFProvider` -> `DocumentExtractorProvider`
[...]

PR #704, opened 2026-06-09
feat(document): add extractor provider foundation

jackefn:
## Summary

Introduce the first MAIC ETL foundation layer from #621 by routing the existing PDF parsing path through a document extraction boundary.

This PR adds `DocumentExtractorProvider` and `DocumentArtifact` abstractions, wraps the existing `unpdf`, self-hosted MinerU, and MinerU Cloud PDF providers as document extractors, and preserves the existing `/api/parse-pdf` response shape through a compatibility adapter.

This PR intentionally does not add RAG, vector storage, persistent knowledge workspaces, multi-document upload, or a knowledge-base UI.

## Related Issues

Related to #621

## Changes

- Add document extraction types:
  - `DocumentExtractorProvider`
  - `DocumentExtractorConfig`
  - `DocumentExtractorInput`
  - `DocumentArtifact`
  - document blocks, assets, citations, and diagnostics
- Add a document extractor registry with provider lookup and MIME-type validation.
- Wrap the existing PDF providers as document extractors:
  - `unpdf`
  - `mineru`
  - `mineru-cloud`
- Declare provider capabilities for current PDF-backed extractors.
- Add PDF compatibility adapters between existing `ParsedPdfContent` and the new `DocumentArtifact`.
- Route `/api/parse-pdf` internally through `extractDocument(...)` while preserving the current API response shape.
- Add tests for registry behavior, MinerU capability metadata, unsupported MIME errors, and PDF compatibility round-tripping.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Run focused document extraction tests:
   `pnpm vitest run tests/document/extractor-registry.test.ts tests/document/pdf-compat.test.ts`
2. Run formatting check:
   `pnpm check`
3. Run i18n key check:
   `pnpm check:i18n-keys`
4. Run lint:
   `pnpm lint`
5. Run TypeScript:
   `npx tsc --noEmit`

- Focused document extraction tests passed: 2 test files, 6 tests.
- `pnpm check` passed.
- `pnpm check:i18n-keys` passed.
- `pnpm lint` passed with existing warnings only and no errors.
- `npx tsc --noEmit` passed after clearing stale local `.next/dev/types` generated from previous branch checkouts.
- Verified this PR does not introduce RAG APIs, vector database code, knowledge-base UI, persistent workspace logic, or vocational task-engine changes.

### Evidence

- [x] Manually tested locally

Local verification logs:

```text
pnpm vitest run tests/document/extractor-registry.test.ts tests/document/pdf-compat.test.ts
Test Files  2 passed (2)
Tests       6 passed (6)

pnpm check
All matched files use Prettier code style!

pnpm check:i18n-keys
i18n key alignment check passed (7 locale files, source: en-US.json).

pnpm lint
0 errors, existing warnings only

npx tsc --noEmit
passed
```

## Checklist


wyuc (commented):
Nice split. This is the right first step toward the #621 MAIC ETL foundation, and the scope discipline is good: it correctly stays at the extraction boundary and does not pull in RAG, vector storage, persistent workspaces, or a knowledge-base UI. I ran a real PDF through `/api/parse-pdf` after the change and confirmed the extracted text and the full response shape (`text`, `metadata.{pdfImages,imageMapping,pageCount,parser,processingTime}`) are preserved, so the compatibility goal holds on the live path.

Nothing here is blocking, since the happy path always takes the `providerRaw` fast-path. The notes below are about making this a load-bearing foundation rather than a pass-through, and about not leaving a trap for the layers that will sit on top.

**1. The artifact round-trip is currently a pass-through, and the fallback is lossy.** `documentArtifactToParsedPdfContent` returns `providerRaw` verbatim whenever it looks like a `ParsedPdfContent`, which is always true on the parse-pdf path. So the `DocumentArtifact` `blocks`/`assets` built in the forward adapter are constructed and discarded every request, and the reconstruct-from-blocks branch is effectively dead. More importantly, that fallback branch silently drops `tables`, `formulas`, `layout`, `taskId`, and rewrites `parser` to the provider id. When the RAG work rebases onto this and feeds an artifact it built itself (no `ParsedPdfContent` in `providerRaw`), MinerU's structured output would vanish on the way back, which is exactly the structured-document case that path cares about. Suggest either making the artifact actually load-bearing (consume `blocks`/`assets`) or trimming the unused reconstruction, and in any case making the fallback preserve tables/formulas/layout.

**2. PDF is baked into the "document" boundary.** `DocumentExtractorProviderId` is a verbatim duplicate of `PDFProviderId`, `toPdfProviderId` is an unchecked `as` cast, and `createPdfBackedDocumentExtractor` indexes `PDF_PROVIDERS[...]` and reads `.name` at module load. The moment a non-PDF provider id is added to the document union without a matching `PDF_PROVIDERS` entry, `pdfProvider.name` throws at import time and takes down the parse-pdf route on load. Also `parsedPdfToDocumentArtifact` spreads PDF-specific keys (`imageMapping`, `pdfImages`, `parser`) into the supposedly provider-neutral `DocumentArtifact.metadata`, which re-introduces the coupling `providerRaw` was meant to isolate. For the boundary to admit a non-PDF extractor later (Milestone 1/2), the provider-id space and the neutral metadata should not assume PDF.

**3. Declared capabilities are never used for selection.** `selectDocumentExtractorProvider` chooses by MIME membership plus an optional `preferredProviderId` only; the `DocumentExtractorCapabilities` matrix is set but read nowhere. #621 calls for capability-aware selection, and with three providers all claiming `application/pdf` the current tiebreak is silent array order. The matrix also restates `PDF_PROVIDERS[id].features` in a second format (and `mineru` is marked `async: false` while the local server also polls), so it is a second source of truth that will drift. Either wire selection through capabilities or drop the matrix until something consumes it.

**4. The registry duplicates an existing pattern.** The new Map-based registry is a third lookup over the same three ids alongside `PDF_PROVIDERS` and the `lib/ai/providers.ts` registry. Worth mirroring the existing shape rather than adding a parallel one.

**Minor:** the unknown-provider error message changed from `Unknown PDF provider` to `Unknown document extractor provider` (any assertion/log keyed on the old text breaks); the data-URL mime parse `src.slice(5, src.indexOf(';'))` produces a truncated mime when a data URL has no `;`; and several modeled types (`DocumentCitation`, `DocumentDiagnostic`, `outline: unknown`, the `file` asset variant) are never produced (`citations`/`diagnostics` are hardcoded empty), so they are speculative for M0.

Overall this is a good direction and I am glad it landed as its own PR. 1 and 2 are the ones I would address before the RAG/extraction layers build on it.

wyuc (approved):
Verified the revision (`4a9f2c2`). All review points are addressed:

1. The artifact round-trip is no longer a pass-through veneer: `documentArtifactToParsedPdfContent` reconstructs `tables`/`formulas`/`layout` from `blocks` on both the `providerRaw` path and the reconstruct-from-blocks fallback, so MinerU structured fields are no longer silently dropped (covered by `tests/document/pdf-compat.test.ts`). `taskId` is only absent on the no-`providerRaw` path, which is inherent (MinerU-only, no block representation) and codified in the test.
2. The boundary no longer bakes in PDF: `DocumentExtractorProviderId` is now `string`, provider construction is null-safe (no module-load throw for an unknown id), and `parsedPdfToDocumentArtifact` keeps `DocumentArtifact.metadata` provider-neutral (no `pdfImages`/`imageMapping`/`parser` leak).
3. Selection is now capability-aware (`requiredCapabilities` filter in `selectDocumentExtractorProvider`, with positive/negative tests).
4. `/api/parse-pdf` is behavior-preserving: same response shape, default `unpdf`, validations intact. I also ran a real PDF end-to-end through the UI upload and the `/api/parse-pdf` endpoint on this head, and text and metadata shape are preserved.

`tests/document` 8/8, CI green.

Two non-blocking follow-ups for when the extraction layers build on this: have the parse-pdf route actually pass `requiredCapabilities` so capability selection is exercised on the real path, and consider carrying `taskId` through the reconstruct path. Neither blocks this foundation PR. LGTM, approving.

PR #706, opened 2026-06-09
perf(renderer): diff code lines in O(n) via a prev-line map

### LINKED ISSUE #705: perf(renderer): O(n²) line diffing in code-block typing animation
In `@maic/renderer`, `BaseCodeElement` computes per-line animation states in a `useEffect` that runs on every `[lines, animate]` change (i.e. on each streaming update while a code block types out). The diff against the previous render uses two passes:

```ts
const prevIds = new Set(prevLinesRef.current.map((l) => l.id));
for (const line of lines) {
  if (!prevIds.has(line.id)) states.set(line.id, { type: "inserted", timestamp: 0 });
}
for (const line of lines) {
  const prev = prevLinesRef.current.find((p) => p.id === line.id);   // O(prevLines) per line
  if (prev && prev.content !== line.content) states.set(line.id, { type: "replaced", timestamp: 0 });
}
```

The first loop already builds a `Set` for O(1) membership — but the second loop does a linear `prevLinesRef.current.find(...)` for **every** line just to compare content, making it **O(lines × prevLines) = O(n²)**. Code blocks routinely run 50-300 lines, and this fires repeatedly during the typing animation, so it is an on-render hot path.

### Fix
Build a single `Map<id, prevLine>` once and fold both checks into one O(n) pass (insert vs. replace are mutually exclusive: a line absent from prev is "inserted", a line present with changed content is "replaced"). This also removes the redundant second iteration and is consistent with the `Set` the code already uses in the first loop. Small PR to follow.

ly-wang19:
## Problem

`BaseCodeElement`'s line-diff `useEffect` runs on every `[lines, animate]` change — i.e. on each streaming update while a code block types out. It built a `Set` for the insert check (O(1)), then did a **second** pass with a linear `prevLinesRef.current.find(...)` per line for the content/replace check:

```ts
const prevIds = new Set(prevLinesRef.current.map((l) => l.id));
for (const line of lines) {
  if (!prevIds.has(line.id)) states.set(line.id, { type: "inserted", timestamp: 0 });
}
for (const line of lines) {
  const prev = prevLinesRef.current.find((p) => p.id === line.id);   // O(prevLines) per line
  if (prev && prev.content !== line.content) states.set(line.id, { type: "replaced", timestamp: 0 });
}
```

That second loop is **O(lines × prevLines) = O(n²)**. Code blocks routinely run 50-300 lines and this fires repeatedly during the typing animation, so it's an on-render hot path.

## Fix

One `Map<id, prevLine>`, one O(n) pass. Insert and replace were already mutually exclusive (a line either isn't in prev → inserted, or is in prev with changed content → replaced), so the merged `if/else if` is behavior-identical and also drops the redundant second iteration. Consistent with the `Set` the first loop already used.

```ts
const prevById = new Map(prevLinesRef.current.map((l) => [l.id, l]));
for (const line of lines) {
  const prev = prevById.get(line.id);
  if (!prev) states.set(line.id, { type: "inserted", timestamp: 0 });
  else if (prev.content !== line.content) states.set(line.id, { type: "replaced", timestamp: 0 });
}
```

## Verification
- `prettier` clean, `eslint` clean, `tsc --noEmit` error count unchanged (the only errors are pre-existing unresolved-workspace-module noise; the edited region type-checks clean).
- Behavior preserved: identical `states` map for any (prev, current) line pair — `typing` on first render, `inserted` for new ids, `replaced` for same id + changed content, unchanged otherwise.

Closes #705


wyuc (approved):
Reviewed after updating the branch onto the latest main. The final diff remains a single +4/-8 renderer hunk replacing the per-line linear lookup with a previous-line Map, preserving behavior while reducing the diff pass from O(n²) to O(n). Local @openmaic/renderer validation passed (19 test files / 220 tests, plus tsc --noEmit), and the refreshed GitHub CI passed both Lint, Typecheck & Unit Tests and E2E Tests.

PR #668, opened 2026-06-05
feat(packages): introduce maic-import and maic-renderer workspace packages

xuyuanwei678:
## Summary

Extracts the PPTX import pipeline and the slide renderer from the OpenMAIC app
into two reusable workspace packages

- **`packages/maic-import/`** (was `pptxtojson-pro`) — parses `.pptx` files and
  emits OpenMAIC-shaped `Slide[]`. Includes adapter, import-pipeline, model
  nodes, serializers (text/shape/table/chart/image/math) and shape presets.
- **`packages/maic-renderer/`** (was `slide-renderer`) — React component for
  rendering PPTist-style `Slide` JSON. Read-only canvas in v1:
  - `<SlideCanvas>` main entry (props or `<SlideRendererProvider>` context),
    auto-fits viewport.
  - Element renderers: text, shape (gradient/pattern), image (clip/outline),
    table, chart, video, line, latex, code.
  - Four effect overlays: highlight, spotlight, laser, zoom (all optional).
  - `slideToPng` standalone snapshot util (html2canvas-pro).
  - Self-hosted Chinese font set (`maic-renderer/fonts.css`, 22 woff2 faces).
  - Tailwind dependency dropped from runtime; styles are data-driven inline.

Other highlights:

- Iterative parser+renderer fidelity passes (image soft edges, per-cell table
  borders, text fill scope, math/text serializer expansions, shape preset
  growth).
- Postinstall builds both packages and runs `sync-maic-import.mjs` to copy the
  parser's prebuilt dist into `public/vendor/maic-import/` so the app can
  URL-import it (bypassing Turbopack's pdfjs-dist dynamic-require trap).
- README / DESIGN docs for both packages, including install, quickstart, API
  surface, Tailwind setup and effect overlays.
- The two packages are renamed in their final commit to align with the MAIC
  product family naming.

## Test plan

- [ ] `pnpm install` succeeds (workspace deps resolve, postinstall builds both
      packages, sync script copies dist).
- [ ] `pnpm dev` boots; `/slide-renderer-demo` renders the bundled demo deck
      and lets you import a real `.pptx` via the upload button.
- [ ] Exported PNG via "导出 PNG" matches the on-screen slide layout.
- [ ] Existing classroom routes (`/classroom/...`) still render slides
      correctly (no regressions from the rename — workspace symlinks point at
      the new paths).
- [ ] `pnpm build` succeeds end to end.

Made with [Cursor](https://cursor.com)

PR #707, opened 2026-06-09
refactor(app): consume @maic/dsl + @maic/renderer in the app (slide types + read-only thumbnails)

wyuc:
## Why

#668 extracted the slide stack into the `@maic/{dsl,renderer,importer}` workspace packages, but the app itself still renders from its own duplicated copies — so `@maic/dsl` had 0 import sites in the app and `@maic/renderer` was only wired for fonts/Tailwind, not actually used to render. This starts dogfooding the packages back into the app so the copies can't drift again.

## What

**1. `lib/types/slides.ts` → thin re-export of `@maic/dsl`.**
The hand-maintained file was the seed for `@maic/dsl`; the app now consumes the package (`@maic/dsl` added as a direct dep). All 51 exported symbols are covered by the package, so every `from '@/lib/types/slides'` import site keeps working unchanged — no call-site churn. `ShapePathFormulasKeys` / `ElementTypes` move from local `const enum` to the package's regular `enum` (a `const enum` can't be re-exported across a package boundary under `isolatedModules`); runtime value usage is identical.

**2. Read-only slide thumbnails render via `@maic/renderer`'s `SlideCanvas`.**
A small `SlideThumbnail` wrapper owns the outer box sizing + lazy placeholder and delegates rendering to `SlideCanvas`. Three read-only surfaces switch to it: playback scene sidebar, scene-thumbnail content (editor nav rail), and home recent-course cards. The wrapper preserves the legacy thumbnail video treatment (muted, play-badged `<video>`, placeholder filtering) via the renderer's `renderVideo` slot.

## Not in scope

The full-size editing canvas is untouched — `@maic/renderer` v1 is read-only, editing is v2.

## Test plan

- `tsc --noEmit` clean; `next build` OK.
- e2e (Playwright): 19/19 green, including the `recent-video-thumbnail` specs (caught a missing play-badge on the first pass → fixed via `renderVideo`).
- Manual real-generation smoke: generated a course end-to-end and confirmed the generated slides (text + shapes + LaTeX) render correctly in the sidebar thumbnails via `@maic/renderer`.

## Follow-ups (not in this PR)

- Migrate the editor nav rail / full-size canvas once `@maic/renderer` gains an editing surface (v2).
- Promote `lib/types/stage.ts` pure types into `@maic/dsl`; reserve `@maic/exporter`.

## Minor heads-up

`@maic/dsl`'s built `dist/index.js` uses extensionless re-exports (`export * from './slides'`), which fails bare-node ESM resolution (Next's bundler is unaffected). Worth emitting `.js` extensions in the dsl build.

## Follow-up fix in this PR: media placeholders in thumbnails

`@maic/renderer` renders `element.src` as-is, but this app's generated media keeps `gen_img_*`/`gen_vid_*` placeholder refs in the slide and tracks the generated `objectUrl` in the media-generation store (the slide's `src` is never mutated). A statically-rendered thumbnail therefore showed broken placeholders and didn't update when a failed image's retry later succeeded. `SlideThumbnail` now resolves placeholders against the store via `useResolvedSlide` (mirroring the full-size canvas's `useResolvedImageSrc`), subscribed through a primitive signature of just this slide's resolutions so unrelated task churn doesn't re-render thumbnails.


cosarah (commented):
Findings

- P2 components/slide-renderer/SlideThumbnail.tsx:58: useResolvedSlide only resolves el.src, but generated videos are currently keyed by mediaRef. A mediaRef-only video never subscribes to its task, so classroom thumbnails stay on the placeholder after generation or retry. Please reuse getVideoMediaRefForElement or resolve both mediaRef and legacy src.

- P3 components/slide-renderer/SlideThumbnail.tsx:65: the signature separator is literal NUL bytes. Git treats the new .tsx source as binary, so GitHub diffs and review tooling degrade. Please replace it with an escaped delimiter such as \\0 or \\u0000, or a visible string delimiter.

wyuc:
Both addressed in 173e40a:

- **P2 (mediaRef-keyed videos):** `useResolvedSlide` now derives each element's task key via `getVideoMediaRefForElement` for videos (mediaRef, with legacy `gen_vid_*` src fallback) — same subscription as `BaseVideoElement` — and keeps the placeholder-src key for images, matching `useResolvedImageSrc`. A mediaRef-keyed video that already carries a real playable src keeps it while its task is unresolved; only placeholder srcs are blanked.

- **P3 (NUL delimiter):** verified — the file was indeed being treated as binary (`Bin` in diffstat). The signature delimiter is now a visible `|`; the source is plain UTF-8 again, so diffs render normally from this commit on.

wyuc:
@cosarah gentle ping — both findings from your review are addressed in 173e40a (P2: videos now keyed by `getVideoMediaRefForElement`; P3: NUL delimiter → `|`). CI is green and the branch is mergeable. Could you take another look when you have a moment? Thanks!

cosarah (approved):
LGTM, verfied locally

PR #709, opened 2026-06-09
fix(export): preserve attributes before an empty style attribute (#682)

### LINKED ISSUE #682: formatAttributes drops earlier attributes when style is empty
### Summary

`formatAttributes` in `lib/export/html-parser/stringify.ts` drops every
attribute accumulated *before* an empty `style` attribute, because the reduce
callback returns `''` (which becomes the new accumulator) instead of the
accumulated string.

### Code

```ts
export const formatAttributes = (attributes: ElementAttribute[]) => {
  return attributes.reduce((attrs, attribute) => {
    const { key, value } = attribute;
    if (value === null) return `${attrs} ${key}`;
    if (key === 'style' && !value) return '';   // <- resets the accumulator
    ...
  }, '');
};
```

### Impact

For `[{class:'foo'}, {style:''}, {id:'bar'}]`, `toHTML` emits `<div id='bar'>` —
`class='foo'` is lost. `toHTML` runs `formatAttributes` for every element, so any
element with an empty `style=""` after another attribute loses the earlier
attributes on stringify (e.g. HTML round-trips in edit/export).

### Fix

Return the accumulator instead of an empty string: `return attrs;` (still omits
the empty style attribute, but preserves the others).

tongshu2023:
### Bug

`formatAttributes` in `lib/export/html-parser/stringify.ts` drops every attribute accumulated **before** an empty `style` attribute, because the empty-style branch returns `''` — which becomes the new `reduce` accumulator — instead of the accumulated string. Closes #682.

```ts
if (key === 'style' && !value) return '';   // resets the accumulator
```

For an element like `{ id: 'box', style: '' }`, serialization produced `<div>` instead of `<div id='box'>` — the `id` (and any other earlier attribute) silently disappears from exported HTML.

### Fix

Return the accumulator unchanged so the empty `style` is skipped without discarding prior attributes:

```ts
if (key === 'style' && !value) return attrs;
```

### Tests

Adds `tests/export/stringify.test.ts` (none existed for this module):

- `formatAttributes`: value attribute quoting, `null` → boolean attribute, single→double quote switch when the value contains `'`, empty `style` omitted, **regression: attributes before an empty `style` are kept**, non-empty `style` preserved
- `toHTML`: element whose first attribute precedes an empty `style` (regression), void-tag serialization, text/comment passthrough

### Verification

```
$ npx vitest run tests/export/stringify.test.ts
 Test Files  1 passed (1)
      Tests  9 passed (9)
```

`prettier --check`, `eslint`, and `tsc --noEmit` all clean.

### DIFF
--- a/lib/export/html-parser/stringify.ts
+++ b/lib/export/html-parser/stringify.ts
@@ -5,7 +5,7 @@ export const formatAttributes = (attributes: ElementAttribute[]) => {
   return attributes.reduce((attrs, attribute) => {
     const { key, value } = attribute;
     if (value === null) return `${attrs} ${key}`;
-    if (key === 'style' && !value) return '';
+    if (key === 'style' && !value) return attrs;
 
     const quoteEscape = value.indexOf("'") !== -1;
     const quote = quoteEscape ? '"' : "'";
--- a/tests/export/stringify.test.ts
+++ b/tests/export/stringify.test.ts
@@ -1,0 +1,80 @@
+import { describe, expect, it } from 'vitest';
+import { formatAttributes, toHTML } from '@/lib/export/html-parser/stringify';
+import type { AST, ElementAttribute } from '@/lib/export/html-parser/types';
+
+describe('formatAttributes', () => {
+  it('formats a value attribute with single quotes', () => {
+    const attrs: ElementAttribute[] = [{ key: 'id', value: 'main' }];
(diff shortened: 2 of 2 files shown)


tongshu2023:
Apologies — I missed that #683 (opened earlier by the team) already covers this. Closing in favor of that PR. The tests added here are free to cherry-pick if useful.

PR #711, opened 2026-06-09
fix(audio): gate SpeechButton on ASR availability

### LINKED ISSUE #710: SpeechButton stays clickable when ASR is turned off
### Describe the bug

The microphone `SpeechButton` (e.g. at the bottom-right of the home page input box) remains enabled and clickable even after ASR is turned off (`asrEnabled = false`). Clicking it still starts a recording for a feature the user has explicitly disabled.

### Root cause

`SpeechButton` is a "dumb" component: it only honors the `disabled` prop passed by each call site and never reads the `asrEnabled` state itself (`components/audio/speech-button.tsx`). The gate is therefore wired ad-hoc per call site, and the home page (`app/page.tsx`) passes no `disabled` at all, so it is never gated. Beyond the toggle, the button is also not gated when the selected provider is unusable (requires an API key that isn't set) or when the browser lacks Web Speech API support for `browser-native`.

### Steps to reproduce

1. Open the home page.
2. In the media/ASR settings, turn ASR off.
3. The mic button next to the input is still highlighted and clickable.

### Expected behavior

When ASR is unavailable (toggled off, selected provider not configured, or browser unsupported), the button should be visibly disabled.

### Proposed fix

Centralize ASR availability into a single `useASRAvailable()` hook and have `SpeechButton` consume it internally, so every call site is gated uniformly. PR to follow.

wyuc:
Closes #710

### Problem

`SpeechButton` only honors the `disabled` prop and never reads ASR state, so call sites must each remember to wire it. The home page passes nothing, leaving the mic button clickable even when ASR is turned off.

### Fix

- Add `lib/hooks/use-asr-available.ts` — a single source of truth combining the three things that gate speech input: the `asrEnabled` toggle, whether the selected provider is usable (keyless / has a key / server-managed, mirroring the `cfgOk` check in the media popover), and Web Speech API support for `browser-native` (via `useSyncExternalStore`, SSR-safe).
- Have `SpeechButton` consume `useASRAvailable()` internally. The external `disabled` prop still composes on top, so existing call-site conditions (e.g. `isLoading`) keep working. Every call site — home, PBL chat, quiz — is now gated uniformly; new call sites are protected by default.

No call-site changes required.

### Test

- `tsc --noEmit` clean (no new errors); `eslint` clean.
- Toggling ASR off greys out and disables the button across all call sites; switching to an unconfigured key-required provider, or a browser without Web Speech support, does the same.

### DIFF
--- a/components/audio/speech-button.tsx
+++ b/components/audio/speech-button.tsx
@@ -3,6 +3,7 @@
 import { useCallback, useEffect, useRef } from 'react';
 import { Mic, Loader2 } from 'lucide-react';
 import { useAudioRecorder } from '@/lib/hooks/use-audio-recorder';
+import { useASRAvailable } from '@/lib/hooks/use-asr-available';
 import { useI18n } from '@/lib/hooks/use-i18n';
 import { cn } from '@/lib/utils';
 import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/ui/tooltip';
@@ -44,6 +45,13 @@ export function SpeechButton({
 
   const active = isRecording || isProcessing;
 
+  // Gate on ASR availability (toggle + provider configured + browser support)
+  // so every call site is disabled uniformly when ASR is off/unusable — but
+  // never while actively recording/processing, so the user can always click to
+  // stop (the recorder has no auto-stop and would otherwise leave the mic open).
+  const asrAvailable = useASRAvailable();
+  const isDisabled = (disabled || !asrAvailable) && !active;
+
   const handleClick = () => {
     if (isRecording) {
       stopRecording();
@@ -62,7 +70,7 @@ export function SpeechButton({
       <TooltipTrigger asChild>
         <button
           type="button"
-          disabled={disabled || isProcessing}
+          disabled={isDisabled || isProcessing}
           onClick={handleClick}
           className={cn(
             'relative flex items-center justify-center rounded-lg transition-all duration-200 shrink-0 cursor-pointer',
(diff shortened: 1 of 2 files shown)


wyuc:
### Self-review summary

- **No blockers.** The availability gate mirrors the existing `cfgOk` check in the media popover for built-in providers and the `asr-settings` logic for custom providers; `browser-native` (keyless) is always considered configured. `useSyncExternalStore` keeps the Web Speech support check SSR-safe (server snapshot assumes supported, so no false-disable before hydration). Hook order and selector reference-stability are sound.
- **Addressed one finding:** the gate is no longer applied while actively recording/processing (`isDisabled = (disabled || !asrAvailable) && !active`). The recorder has no auto-stop, so disabling the button mid-recording could otherwise strand an open mic with no way to click stop. This also closes the same latent case for the external `disabled` prop.
- **Verified in a browser:** toggling ASR off greys out and disables the mic button (`disabled`, `opacity: 0.4`, `pointer-events: none`) across all call sites; toggling it back on restores it. No console errors.
- `tsc --noEmit` and `eslint` clean on both changed files.

cosarah (commented):
Findings:

- P2 `lib/hooks/use-asr-available.ts:38`: `useASRAvailable` treats custom keyless ASR providers as available even when they have no configured model. Existing ASR provider selection hides custom ASR until `customModels.length > 0` (`components/generation/media-popover.tsx:176`), but the new hook only checks key/server config. After adding a custom ASR provider, the mic can be enabled while the provider is still incomplete, leading to a recording flow that fails at transcription time. The availability predicate should include the custom-provider model requirement, ideally shared with the popover logic.

- P3 `lib/hooks/use-asr-available.ts:40`: API keys are checked with `!!providerConfig?.apiKey`, but the recorder only sends keys after `.trim()` (`lib/hooks/use-audio-recorder.ts:65`). A whitespace-only key enables the button and then fails on the server as missing credentials. Use `providerConfig?.apiKey?.trim()` here.

wyuc:
@cosarah Thanks — both confirmed real and fixed in be55116:

- **P2 (custom provider without models):** the hook now requires at least one configured model for custom providers (`modelOk = !!builtIn || customModels.length > 0`), mirroring the media-popover rule that hides model-less customs. I kept the predicate local to the hook with a comment pointing at the popover rule rather than extracting a shared helper — happy to extract one if you'd prefer the popover to consume it too.
- **P3 (whitespace-only key):** now checked with `apiKey?.trim()`, matching what the recorder actually sends.

`tsc --noEmit` and `eslint` clean on the changed file.

cosarah (approved):
LGTM，verified locally

PR #712, opened 2026-06-09
fix(export): convert PPTX shadow offset from px to pt (#678)

### LINKED ISSUE #678: PPTX export: shadow offset not converted from px to pt
### Summary

In the PPTX exporter (`lib/export/use-export-pptx.ts`, `getShadowOption`), the
shadow **blur** is converted from px to pt but the shadow **offset** is not, so
exported shadows are offset too far.

### Code

```ts
return {
  type: 'outer',
  color: c.color.replace('#', ''),
  opacity: c.alpha,
  blur: shadow.blur / ratioPx2Pt,   // ✓ px → pt
  offset,                           // ✗ still px
  angle,
};
```

pptxgenjs shadow `blur` and `offset` are both in **points**, but the slide model
is in **pixels**. `ratioPx2Pt = (96 / 72) × (viewportSize / 960)`.

### Impact

At the default viewport, `ratioPx2Pt ≈ 1.333`, so every exported shadow sits
~33% farther from its element than authored. The blur radius is already correct,
which makes the mismatch visible (sharp offset, soft blur).

### Fix

Convert the offset the same way as blur: `offset: offset / ratioPx2Pt`.

tongshu2023:
### Bug

In the PPTX exporter (`lib/export/use-export-pptx.ts`, `getShadowOption`), the shadow **blur** is converted from px to pt but the shadow **offset** is not. pptxgenjs treats both `blur` and `offset` as **points**, while the slide model is in **pixels** (`ratioPx2Pt = (96 / 72) × (viewportSize / 960)`). Closes #678.

```ts
blur: shadow.blur / ratioPx2Pt,   // ✓ px → pt
offset,                           // ✗ still px
```

At the default viewport `ratioPx2Pt ≈ 1.333`, so every exported shadow sits ~33% farther from its element than authored — and because the blur radius is already correct, the mismatch is visible (sharp offset, soft blur).

### Fix

Convert the offset with the same ratio as the blur:

```ts
offset: offset / ratioPx2Pt,
```

### Tests

`getShadowOption` is a pure function; exported it for testing (mirrors the already-exported `buildPptxBlob` used by the round-trip tests) and added `tests/export/shadow-option.test.ts`:

- offset is converted px → pt (regression)
- blur and offset share the same ratio (stay aligned)
- default offset (4px) is converted
- direction → angle/magnitude mapping across all 8 h/v quadrants
- color/opacity formatting

### Verification

```
$ npx vitest run tests/export/shadow-option.test.ts
 Test Files  1 passed (1)
      Tests  5 passed (5)
```

`prettier --check`, `eslint`, and `tsc --noEmit` all clean. The only source change is the one-line offset conversion plus the `export` keyword.

### DIFF
--- a/lib/export/use-export-pptx.ts
+++ b/lib/export/use-export-pptx.ts
@@ -251,7 +251,7 @@ function formatPoints(points: SvgPoints, ratioPx2Inch: number, scale = { x: 1, y
 
 // ── Shadow config ──
 
-function getShadowOption(shadow: PPTElementShadow, ratioPx2Pt: number): pptxgen.ShadowProps {
+export function getShadowOption(shadow: PPTElementShadow, ratioPx2Pt: number): pptxgen.ShadowProps {
   const c = formatColor(shadow.color);
   const { h, v } = shadow;
 
@@ -296,7 +296,7 @@ function getShadowOption(shadow: PPTElementShadow, ratioPx2Pt: number): pptxgen.
     color: c.color.replace('#', ''),
     opacity: c.alpha,
     blur: shadow.blur / ratioPx2Pt,
-    offset,
+    offset: offset / ratioPx2Pt,
     angle,
   };
 }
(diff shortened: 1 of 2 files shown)


tongshu2023:
Apologies — I missed that #679 (opened earlier by the team) already covers this. Closing in favor of that PR. The tests added here are free to cherry-pick if useful.

PR #713, opened 2026-06-09
fix(audio): track string context when splitting Doubao TTS stream (#676)

### LINKED ISSUE #676: Doubao TTS stream splitter miscounts braces inside string values
### Summary

The Doubao TTS provider (`lib/audio/tts-providers.ts`) splits its streamed
response — a run of concatenated JSON objects with no delimiter — by counting
`{`/`}` braces, but the counter does not track string context. A `{` or `}`
inside a JSON string value mis-aligns the object boundaries.

### Code

```ts
let depth = 0;
let start = -1;
for (let i = 0; i < responseText.length; i++) {
  if (responseText[i] === '{') { if (depth === 0) start = i; depth++; }
  else if (responseText[i] === '}') {
    depth--;
    if (depth === 0 && start >= 0) { /* JSON.parse(slice) */ }
  }
}
```

### Impact

If any object's string value contains a brace — e.g. an error
`{"code":1,"message":"bad {input}"}` — the `}` inside `message` decrements
`depth` early, so the object is split mid-string. `JSON.parse` of the fragment
throws and the chunk is dropped (audio lost, or the real error is swallowed).

### Fix

Track string context (and escaped quotes) while scanning, the same way the
depth-scanner in `lib/generation/json-repair.ts` does. Extracting a small
`splitConcatenatedJsonObjects(text)` helper makes it unit-testable.

tongshu2023:
### Bug

The Doubao TTS provider (`lib/audio/tts-providers.ts`) splits its streamed response — a run of concatenated JSON objects with no delimiter — by counting `{`/`}` braces, but the counter does not track string context. A `{` or `}` inside a JSON string value mis-aligns the object boundaries. Closes #676.

For example, an error chunk `{"code":1,"message":"bad {input}"}` — the `}` inside `message` decrements `depth` early, so the object is split mid-string. `JSON.parse` of the fragment throws and the chunk is dropped (audio lost, or the real error swallowed). Worse, every following object in the stream stays mis-aligned.

### Fix

Extract a pure, unit-testable `splitConcatenatedJsonObjects(text)` helper that tracks string context and escapes while scanning — the same approach the depth-scanner in `lib/generation/json-repair.ts` already uses — and refactor the Doubao loop to iterate over its output. Behavior is unchanged for well-formed streams; only brace-in-string (and escaped-quote) cases are now handled correctly.

### Tests

Adds `tests/audio/tts-providers-split.test.ts`:

- single object; two concatenated objects with no delimiter
- **brace inside a string value is not split (regression)**
- following objects stay aligned after a brace-in-string object
- escaped quotes inside string values
- nested objects treated as one top-level object
- empty / noise / stray-closer inputs

### Verification

```
$ npx vitest run tests/audio/tts-providers-split.test.ts
 Test Files  1 passed (1)
      Tests  8 passed (8)
```

`prettier --check`, `eslint`, and `tsc --noEmit` all clean.

### DIFF
--- a/lib/audio/tts-providers.ts
+++ b/lib/audio/tts-providers.ts
@@ -845,6 +845,51 @@ export async function getCurrentTTSConfig(): Promise<TTSModelConfig> {
 export { getAllTTSProviders, getTTSProvider, getTTSVoices } from './constants';
 
 /**
+ * Split a run of concatenated JSON objects (no delimiter between them) into
+ * individual object strings. Tracks string context and escapes so that a `{`
+ * or `}` inside a string value does not mis-align object boundaries.
+ */
+export function splitConcatenatedJsonObjects(text: string): string[] {
+  const objects: string[] = [];
+  let depth = 0;
+  let start = -1;
+  let inString = false;
+  let escapeNext = false;
+
+  for (let i = 0; i < text.length; i++) {
+    const char = text[i];
+
+    if (escapeNext) {
+      escapeNext = false;
+      continue;
+    }
+    if (char === '\\' && inString) {
+      escapeNext = true;
+      continue;
+    }
+    if (char === '"') {
+      inString = !inString;
+      continue;
+    }
+    if (inString) continue;
+
+    if (char === '{') {
+      if (depth === 0) start = i;
+      depth++;
+    } else if (char === '}') {
+      if (depth === 0) continue; // stray closer outside any object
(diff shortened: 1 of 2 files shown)


tongshu2023:
Apologies — I missed that #677 (opened earlier by the team) already covers this. Closing in favor of that PR. The tests added here are free to cherry-pick if useful.

PR #714, opened 2026-06-09
fix(mathml2omml): call includes() instead of indexing it, preserving trailing text (#680)

### LINKED ISSUE #680: mathml2omml drops trailing text in text containers (includes indexed, not called)
### Summary

In the vendored `mathml2omml` package, `parse.js` indexes the `includes`
**method** instead of calling it, so a condition is always falsy and trailing
text inside a MathML text container is silently dropped.

### Code

`packages/mathml2omml/src/parse-stringify/parse.js`:

```js
const textContainerNames = ['mtext', 'mi', 'mn', 'mo', 'ms']   // line 8 — an array
...
textContainerNames.includes(current.name) &&                   // line 48 — correct
...
textContainerNames.includes[arr[level].name] &&                // line 82 — BUG: indexes the fn
```

`Array.prototype.includes` is a function; `includes[name]` reads a property of
that function object, which is `undefined` for any tag name — so the whole
branch (the "trailing text node" handler) never runs. Line 48 in the same file
shows the intended form.

### Impact

Trailing text after a nested closing tag inside a text container
(`mtext`/`mi`/`mn`/`mo`/`ms`) is never added to the tree, so it is missing from
the OMML — affecting math rendered into PPTX export via `latexToOmml`.

### Fix

Call the method: `textContainerNames.includes(arr[level].name)`. One character.
`dist/` is gitignored and rebuilt by `postinstall` (`npm run build`), so the
runtime bundle picks up the fix on install.

tongshu2023:
### Bug

In the vendored `mathml2omml` package, `parse.js` indexes the `includes` **method** instead of calling it:

```js
textContainerNames.includes[arr[level].name] &&   // always undefined → falsy
```

`textContainerNames` is an array; `arr[level].name` is a string like `"mtext"`. `Array.prototype.includes` has no such index property, so the expression is always `undefined` (falsy). The whole "trailing text node" branch is therefore never taken, and trailing text inside a text container (`mtext`/`mi`/`mn`/`mo`/`ms`) is silently dropped. Closes #680.

### Fix

Call the method:

```js
textContainerNames.includes(arr[level].name) &&
```

(The sibling check on line 48 already uses the correct `includes(...)` call.)

### Verification

Before/after on the rebuilt package:

```
input:  <math><mtext>a<mspace/>b</mtext></math>
buggy:  OMML contains "b" → false   (trailing "b" dropped)
fixed:  OMML contains "b" → true
```

Adds `tests/export/mathml2omml-trailing-text.test.ts` (3 cases: trailing text in `<mtext>` and `<mi>`, plus a plain-text-container sanity check):

```
$ npx vitest run tests/export/mathml2omml-trailing-text.test.ts
 Test Files  1 passed (1)
      Tests  3 passed (3)
```

`prettier --check` and `eslint` clean. Source change is a single character-level edit (`[ ]` → `( )`).

### DIFF
--- a/packages/mathml2omml/src/parse-stringify/parse.js
+++ b/packages/mathml2omml/src/parse-stringify/parse.js
@@ -79,7 +79,7 @@ export function parse(html, options = {}) {
       }
       if (
         level > -1 &&
-        textContainerNames.includes[arr[level].name] &&
+        textContainerNames.includes(arr[level].name) &&
         nextChar !== '<' &&
         nextChar
       ) {
--- a/tests/export/mathml2omml-trailing-text.test.ts
+++ b/tests/export/mathml2omml-trailing-text.test.ts
@@ -1,0 +1,27 @@
+import { describe, expect, it } from 'vitest';
+import { mml2omml } from 'mathml2omml';
+
+/**
+ * Regression for #680: parse.js indexed `textContainerNames.includes[...]`
+ * instead of calling `.includes(...)`, so the "trailing text node" branch was
+ * always skipped and trailing text inside a text container was dropped.
+ */
+describe('mml2omml trailing text in text containers (#680)', () => {
+  it('preserves text that trails a child element inside <mtext>', () => {
+    const omml = mml2omml('<math><mtext>a<mspace/>b</mtext></math>');
+    // The trailing "b" used to be silently dropped.
+    expect(omml).toContain('b');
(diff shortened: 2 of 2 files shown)


tongshu2023:
Apologies — I missed that #681 (opened earlier by the team) already covers this. Closing in favor of that PR. The tests added here are free to cherry-pick if useful.

PR #716, opened 2026-06-09
fix(vocational): send taskEngineMode on the on-demand scene generation path

### LINKED ISSUE #715: bug(vocational): scenes 2..N lose procedural-skill — taskEngineMode not sent on the on-demand generation path
## Summary

In the new vocational task-engine feature (#685), the server-side vocational gate is driven by `requirements.taskEngineMode` sent in the `/api/generate/scene-content` request body. **Only the first scene's request carries `requirements`.** Scenes 2..N (and all retries / resume-after-reload) are generated by `useSceneGenerator.generateRemaining`, whose `fetchSceneContent` body never includes `requirements`. As a result every scene after the first silently **downgrades from `procedural-skill` to `diagram`**, losing the procedural-skill training mechanism for most of a vocational course.

## Root cause / trace

**Scene 1** (`app/generation-preview/page.tsx:836`) sends it:
```ts
body: JSON.stringify(withThinkingConfig({
  outline: firstOutline, allOutlines: outlines, ...
  requirements: currentSession.requirements,   // carries taskEngineMode ✅
}))
```

**Scenes 2..N** go through `useSceneGenerator.generateRemaining` → `fetchSceneContent` (`lib/hooks/use-scene-generator.ts:365` and the retry at `:610`). That request body threads `stageInfo`, `languageDirective`, `agents`, … but **no `requirements` / `taskEngineMode`** (grep for `taskEngineMode` in that file returns nothing).

**Server** (`app/api/generate/scene-content/route.ts:128-130`):
```ts
const vocationalActive = resolveVocationalActive(requirements); // undefined → false
const effectiveOutline = applyOutlineFallbacks(outline, !!languageModel, {
  allowProceduralSkill: vocationalActive,                       // false for scenes 2..N
});
```

**Downgrade** (`lib/generation/outline-generator.ts:203-205`):
```ts
if (outline.widgetType === 'procedural-skill' && !options.allowProceduralSkill) {
  return sanitizeProceduralSkillOutline(outline);   // → widgetType: 'diagram', strips task/tools/steps/successCriteria
}
```

## Why it's clearly a bug (internal contradiction)

The task-engine **outline** route marks the operation scenes as `procedural-skill` (the prompt targets a "10-14 vocational mixed-scene ratio"), and the stage is persisted with `taskEngineMode: true` and shows the vocational badge — **but the content path then strips `procedural-skill` from every scene except the first.** The outline and the generated content disagree.

## Impact
- Silent: no error, only a server-side `log.warn`; the stage still reports `taskEngineMode: true`.
- Most scenes of every vocational course render as generic `diagram` HTML instead of the procedural-skill widget.
- Affects the normal interactive multi-scene flow, retries, and resume-after-reload (`generateRemainingRef` at `:683`).

## Why existing tests don't catch it
`tests/generation/scene-content-route-vocational-gate.test.ts` calls the route handler directly with `requirements` supplied (or deliberately omitted to assert the safe-false downgrade). No test exercises the real multi-scene caller (`generateRemaining`), which is exactly the path that always omits `requirements`.

## Proposed fix (open to your preference)

`taskEngineMode` is already persisted on the stage (`lib/types/stage.ts:67`, `lib/utils/database.ts:54`, written in `stage-storage.ts`), so the minimal fix mirrors scene 1 on the 2..N path:
1. Add `taskEngineMode?: boolean` to `GenerationParams` (`use-scene-generator.ts`).
2. Populate it at the `generateRemaining(...)` call-site (`app/classroom/[id]/page.tsx:167`) from `stage.taskEngineMode`.
3. Include `requirements: { taskEngineMode: params.taskEngineMode === true }` in both `fetchSceneContent` bodies (`:365`, `:610`) so the server re-gates via the existing `resolveVocationalActive` (still ANDed with the `OPENMAIC_ENABLE_VOCATIONAL` env flag — no security change).

Alternatively you may prefer a server-authoritative shape (e.g. persist `taskEngineMode` onto the outline itself so the client doesn't need to re-send it). Since this is your feature, I'd rather align on the approach than presume — **happy to open the PR** whichever way you prefer.

(Possibly related: the server bulk path `lib/server/classroom-generation.ts` builds `requirements = { requirement }` with no `taskEngineMode`, so vocational scenes generated there would downgrade too — worth confirming whether that path is meant to support the task engine at all.)

ly-wang19:
Fixes #715.

## Problem

The vocational task-engine gate is driven by `requirements.taskEngineMode` in the `/api/generate/scene-content` request body (`resolveVocationalActive` → `applyOutlineFallbacks({ allowProceduralSkill })`). Only the **first** scene sends it (`app/generation-preview/page.tsx`). Scenes 2..N and retries are generated by `useSceneGenerator.generateRemaining` / `fetchSceneContent`, which never included `requirements` — so `vocationalActive` was `false` for them and `applyOutlineFallbacks` rewrote every `procedural-skill` outline to `diagram`. The outline says `procedural-skill`, the content path strips it; most scenes of a vocational course silently lose the task-engine widget. (Full trace in #715.)

## Fix (minimal, client-side)

`taskEngineMode` is already persisted on the stage (`lib/types/stage.ts`, `lib/utils/database.ts`, written in `stage-storage.ts`), so this mirrors what scene 1 already does:

- add `taskEngineMode?: boolean` to `GenerationParams`
- populate it at the single `generateRemaining(...)` call site from `stage.taskEngineMode`
- include `requirements: { taskEngineMode: params.taskEngineMode === true }` in both `fetchSceneContent` bodies

The retry path (`retrySingleOutline`) inherits it automatically — it reuses `lastParamsRef.current`. **Server contract is unchanged**; the flag is still ANDed with the `OPENMAIC_ENABLE_VOCATIONAL` env gate, so this is not a security/routing change — it only stops the client from dropping a flag it already sends for scene 1.

## Note on approach

This is the smallest fix that restores parity with the first-scene path. If you'd prefer a **server-authoritative** shape instead — e.g. persisting `taskEngineMode` onto the outline so the content route doesn't depend on the client re-sending it — I'm happy to revise this PR that way; it's your feature and I'd rather match your intended design.

## Verification
- `prettier` clean, `eslint` clean, `tsc --noEmit` error count unchanged (only pre-existing unresolved `@maic/*` workspace-module errors; neither edited file appears).
- No test added: the server gate this feeds is already covered by `tests/generation/scene-content-route-vocational-gate.test.ts`; the gap is purely the client hook's request body, and the repo has no React-hook (renderHook + fetch-mock + Zustand) harness to assert it without adding disproportionate test infra. Happy to add one if you'd like to introduce that harness.

### DIFF
--- a/app/classroom/[id]/page.tsx
+++ b/app/classroom/[id]/page.tsx
@@ -175,6 +175,7 @@ export default function ClassroomDetailPage() {
           agents: params.agents,
           userProfile: params.userProfile,
           languageDirective: params.languageDirective || stage.languageDirective,
+          taskEngineMode: stage.taskEngineMode,
         });
       });
     } else if (outlines.length > 0 && stage) {
--- a/lib/hooks/use-scene-generator.ts
+++ b/lib/hooks/use-scene-generator.ts
@@ -83,6 +83,7 @@ async function fetchSceneContent(
     };
     agents?: AgentInfo[];
     languageDirective?: string;
+    requirements?: { taskEngineMode?: boolean };
   },
   signal?: AbortSignal,
 ): Promise<SceneContentResult> {
@@ -269,6 +270,8 @@ export interface GenerationParams {
   agents?: AgentInfo[];
   userProfile?: string;
   languageDirective?: string;
+  /** Vocational task-engine flag; gates procedural-skill generation server-side (see resolveVocationalActive). */
+  taskEngineMode?: boolean;
 }
 
 export function useSceneGenerator(options: UseSceneGeneratorOptions = {}) {
@@ -372,6 +375,7 @@ export function useSceneGenerator(options: UseSceneGeneratorOptions = {}) {
               stageInfo: params.stageInfo,
               agents: params.agents,
               languageDirective: params.languageDirective,
+              requirements: { taskEngineMode: params.taskEngineMode === true },
             },
             signal,
           );
@@ -617,6 +621,7 @@ export function useSceneGenerator(options: UseSceneGeneratorOptions = {}) {
             stageInfo: params.stageInfo,
             agents: params.agents,
             languageDirective: params.languageDirective,
+            requirements: { taskEngineMode: params.taskEngineMode === true },
           },
           signal,
         );


wyuc:
The bug and minimal fix still look valuable: current `main` still does not thread `stage.taskEngineMode` through `generateRemaining` / retry scene-content requests. However, this branch now conflicts with `main` in the touched classroom/generator code. Please rebase onto the latest `main`, resolve the conflicts on your branch, and push the refreshed result; once it is mergeable we can re-run review and CI. Thanks!

wyuc:
@ly-wang19 thanks for the fix, and sorry this sat so long. The bug still reproduces on current `main`, but the on-demand generation path had moved (`app/classroom/[id]/page.tsx` → `components/classroom/ClassroomSurface.tsx`, and `fetchSceneContent` now takes `requirements`), so I merged `main` into your branch and re-applied the same change there (`a139040`): the stage's `taskEngineMode` is forwarded through `GenerationParams`, and `requirements: { taskEngineMode: true }` is sent on the scene-content requests only when it's set, so ordinary generation payloads are unchanged.

wyuc (approved):
Approving the re-applied fix: scenes generated on demand (2..N) now carry the stage's taskEngineMode to /api/generate/scene-content, matching the first scene. tsc, eslint and prettier are clean on the changed files.

PR #717, opened 2026-06-10
fix(web-search): accept <div> result titles in Brave keyless scraper

### LINKED ISSUE #687: Brave keyless scraper returns 0 results — result-title markup changed (span→div)

tongshu2023:
## Summary
Fixes #687 — the keyless Brave web-search path returns 0 results because Brave moved the result title from a `<span>` to a `<div>`.

## Problem
`parseBraveSearchHtml` matched the title with a `<span class="search-snippet-title">`-only regex. Brave's current markup is:

```html
<div class="title search-snippet-title line-clamp-1 …" title="Title">Title</div>
```

So `titleMatch` was always null, `if (!title) continue;` skipped every snippet, and the scraper returned an empty list even though the page contained 20 `data-type="web"` blocks (see the live evidence in #687).

## Fix
Accept either tag for the title element, using a backreference so the open/close tags stay consistent:

```ts
/<(span|div)[^>]*class="[^"]*search-snippet-title[^"]*"[^>]*>([\s\S]*?)<\/\1>/i
```

## Tests
- Updated the `parseBraveSearchHtml` and `searchWithBrave` fixtures to Brave's current `<div>` markup (the old fixtures had drifted from the live page, which is why the suite stayed green through the breakage).
- Added a dedicated legacy `<span>` case so back-compat stays covered.
- `tests/web-search`: 25 tests green; `prettier --check` clean.

Note on live verification: repeat scrapes from my IP are currently answered with HTTP 429 (Brave rate-limits rapid keyless requests — consistent with the best-effort caveat in #687), so the fixtures mirror the live markup captured in the issue's evidence rather than a fresh scrape.

### DIFF
--- a/lib/web-search/brave.ts
+++ b/lib/web-search/brave.ts
@@ -71,10 +71,11 @@ export function parseBraveSearchHtml(html: string, maxResults: number): WebSearc
     const url = decodeHtml(linkMatch[1].trim());
     if (!url || isBraveOwnedUrl(url)) continue;
 
+    // Brave has used both <span> and <div> for the result title over time
     const titleMatch = block.match(
-      /<span[^>]*class="[^"]*search-snippet-title[^"]*"[^>]*>([\s\S]*?)<\/span>/i,
+      /<(span|div)[^>]*class="[^"]*search-snippet-title[^"]*"[^>]*>([\s\S]*?)<\/\1>/i,
     );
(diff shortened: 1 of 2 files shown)


tongshu2023:
Apologies — I missed that #688 (opened earlier by the team) already covers this. Closing in favor of that PR. The tests added here are free to cherry-pick if useful.

Issue #529, opened 2026-05-05
[Bug]: Local/private network URLs are not allowed


geekyouth:
### Bug Description

<img width="780" height="620" alt="Image" src="https://github.com/user-attachments/assets/752dd92b-571c-4194-ae98-f66c4470d984" />

<img width="998" height="342" alt="Image" src="https://github.com/user-attachments/assets/f1873871-177c-465f-904f-ffdc0eff57a4" />

商业AI 响应很慢，服务繁忙，我有大量自建的api服务，就是不让用
Local/private network URLs are not allowed

### Steps to Reproduce

如上

### Expected Behavior

允许私有网络api 提供商

### Actual Behavior

Local/private network URLs are not allowed

### Deployment Method

Docker

### Relevant Logs / Screenshots

```shell

```

wyuc:
Thanks for the report. This is tracked in #455 : set `ALLOW_LOCAL_NETWORKS=true` on the server to allow local or private endpoints. Closing as a duplicate so the discussion stays in one place; please add any extra details there.

PR #718, opened 2026-06-10
fix(proxy): bypass proxy for loopback hosts and honor NO_PROXY

tongshu2023:
## Summary

`proxyFetch` routes every request through the configured HTTP(S) proxy, including loopback targets. A deployment with `HTTP_PROXY` set sends a `http://localhost:8888` SearXNG (or any self-hosted/intranet service) request to the external proxy, where `localhost` resolves to the **proxy machine**, and the request fails. Spotted by @wyuc while verifying #589.

Related to #183 (per-provider proxy opt-in — this PR is the standards-based subset) and the self-hosted deployment pain in #529.

## Changes

- **Loopback always bypasses the proxy**: `localhost`, `*.localhost`, `127.0.0.0/8`, `::1`. Proxying loopback is never meaningful — the proxy's loopback is not the caller's.
- **Honor `no_proxy` / `NO_PROXY`** with the conventional (curl-style) semantics:
  - comma-separated entries, whitespace and empty segments ignored
  - `*` wildcard bypasses everything
  - optional `:port` per entry, matched against the effective port (protocol default `80`/`443` when the URL has none)
  - domain-suffix matching: `example.com` matches itself and `api.example.com`; a leading dot (`.example.com`) is equivalent
- `shouldBypassProxy(url)` is exported and unit-tested; `proxyFetch` behavior is unchanged when no proxy is configured.

## Tests

`tests/server/proxy-fetch.test.ts` — 9 cases covering loopback, exact/subdomain/dot-prefix matching, per-entry ports with protocol defaults, `*`, IP entries, list parsing, and lowercase `no_proxy`. `vitest`, `prettier --check`, `eslint`, and `tsc --noEmit` all pass locally.


wyuc (approved):
LGTM. Clean, standards-based fix with solid test coverage — loopback always bypasses and NO_PROXY follows curl semantics. CI green. Thanks!

PR #719, opened 2026-06-10
fix(web-search): accept div result titles in Brave keyless scraper

### LINKED ISSUE #687: Brave keyless scraper returns 0 results — result-title markup changed (span→div)

tongshu2023:
## Summary

Fixes #687. The keyless Brave web-search path returned **0 results** because Brave moved the result title element:

- now: `<div class="title search-snippet-title line-clamp-1 ..." title="Title">Title</div>`

The title regex matched `<span>` only, so `if (!title) continue;` skipped every snippet.

## Changes

- `parseBraveSearchHtml` accepts `<span>` **or** `<div>` for the title via a backreferenced `(span|div)` group, so the closing tag must match the opening one.
- Test fixtures updated to Brave's current `<div>` markup; one legacy `<span>` case kept for back-compat (per the issue's suggestion).

## Verification

- Live check just now: `https://search.brave.com/search?q=photosynthesis` returns HTTP 200 with **20** `data-type="web"` snippets — **0** span titles, **20** div titles. The old parser extracts 0 of them; the fixed one extracts them all.
- `vitest run tests/web-search/brave.test.ts` — 3/3 pass; `prettier --check` and `eslint` clean.

### DIFF
--- a/lib/web-search/brave.ts
+++ b/lib/web-search/brave.ts
@@ -71,10 +71,12 @@ export function parseBraveSearchHtml(html: string, maxResults: number): WebSearc
     const url = decodeHtml(linkMatch[1].trim());
     if (!url || isBraveOwnedUrl(url)) continue;
 
+    // Brave has shipped the result title as both <span> and <div> over time;
+    // accept either so a markup change doesn't silently drop every result.
     const titleMatch = block.match(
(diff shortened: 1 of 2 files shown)


tongshu2023:
Closing as a duplicate of #688, which predates this PR and fixes the same root cause (I missed it before opening — apologies for the noise). Leaving my live-verification data over there in support.

Issue #664, opened 2026-06-03
[Bug]: PBL creation fails with some local models.


PhillLittlewood:
### Bug Description

When using Gemma 4 in LMStudio it always fails on PBL Slides.  However Gemma 3 works fine for PBL.

### Steps to Reproduce

1. Load any Gemma 4 26B or 31B.  Tried with several context sizes.
2. Generate a course.
3. About 4 slides in the slide with fail to generate and log shows it's the pbl generation.

### Expected Behavior

Pbl slides should generate.

### Actual Behavior

PBL slides fail to generate with Gemma 4.

### Deployment Method

Local development (npm run dev / pnpm dev / yarn dev)

### Browser

Edgd

### Operating System

Windows 11

### Relevant Logs / Screenshots

```shell

```

### Additional Context

LMStudio immediately reject the pbl request with 500 error.

tongshu2023:
Did some digging — this is almost certainly a **tool-calling capability** mismatch, not a context-size or content problem.

**Why only PBL fails:** PBL is the one generation path in this repo that uses function/tool calling. `lib/pbl/generate-pbl.ts` drives an agentic loop via `generateText` + `tools` + `stopWhen` (tools like `set_mode`, project/agent/issueboard MCPs). Regular outline/slide generation is plain text completion. So a model that can't accept a `tools` array in the request will pass the first ~4 slides and then hard-fail exactly when the PBL scene starts — matching your repro.

**Why the immediate 500:** LM Studio validates the request against the model's chat template before inference. If the template has no tool-call support, it rejects the request right away (typically with `"... does not support tool use"` or a jinja template error in the server log). That also explains the Gemma 3 vs Gemma 4 split: what matters isn't the model generation but **whether the GGUF's chat template (or your selected prompt template in LM Studio) declares tool support** — Gemma 3 community templates with tool support have been around for a while; newer Gemma 4 quants may ship templates without it.

**To confirm:** check the LM Studio server log (Developer tab) for the line accompanying the 500 — it states the rejection reason. Pasting it here would settle this.

**Workarounds meanwhile:**
- For PBL scenes, pick a local model whose template supports tools (LM Studio shows a hammer/tool badge on capable models), or switch the Gemma 4 prompt template to a tool-capable variant if one exists.
- Mixed setups work too: keep Gemma 4 for outline/slides and route a tools-capable model for PBL.

[...]

PR #644, opened 2026-06-02
fix(tts): surface provider rate-limit errors as HTTP 429

ly-wang19:
## Summary

TTS providers throw a plain `Error` on any non-OK response, and `/api/generate/tts` catches everything as `GENERATION_FAILED` (500), so an upstream **429** (rate-limit / quota) was indistinguishable from a real failure and clients had no signal to retry/back off. `TTSRateLimitError` already existed for exactly this (with a TODO noting the route collapses everything), but only Doubao threw it and the route never mapped it.

This wires it up end to end.

## Related Issues

Closes #643

## Changes

- `lib/server/api-response.ts`: add a `RATE_LIMITED` error code.
- `lib/audio/tts-providers.ts`: add a shared `throwIfTtsRateLimited(provider, status)` helper and call it at every provider's `!response.ok` site, so an upstream HTTP 429 becomes a typed `TTSRateLimitError`.
- `app/api/generate/tts/route.ts`: return `429 RATE_LIMITED` when a `TTSRateLimitError` is caught (all other errors unchanged).
- `tests/audio/tts-rate-limit.test.ts`: unit tests for the helper.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

### What you personally verified

- The change is **purely additive** (20 insertions, 0 deletions in `tts-providers.ts`): non-429 responses keep their existing provider-specific error messages; only a 429 short-circuits to the typed error.
- Each guard uses the correct provider name for its block and `response` is in scope at every site.

### Evidence

- [x] CI passes (`pnpm check` ✓, `pnpm lint` ✓, `npx tsc --noEmit` ✓, `vitest` — audio suite 22/22, incl. new tests ✓)

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [x] My changes do not introduce new warnings

---

🤖 **AI-assisted PR** (Claude). Self-reviewed before submission.

PR #722, opened 2026-06-10
fix(pbl): explain the tool-calling requirement when generation is rejected before the first step

ly-wang19:
Refs #664 (first, diagnostic half — see "Scope" below for why it doesn't close it).

## Problem

PBL is the **only** generation path that requires native tool calling (`generateText` + Zod `tools` + `stopWhen`, the agentic MCP loop in `lib/pbl/generate-pbl.ts`). Models whose chat template lacks tools support reject such a request **outright** — #664 reports LM Studio + Gemma 4 returning an instant 500 (while Gemma 3, whose LM Studio template has tools support, works). The raw provider error propagates to the failed-scene UI with no hint that the *model capability*, not the course or the app, is the problem — and retrying can never succeed.

## Fix

Track how many agentic steps completed. When the call fails **before the first step** — the signature of a tools-incapable model rejecting the request — wrap the error with an actionable message:

> PBL generation was rejected before the first step. PBL scenes drive the model through tool/function calls, so the selected model must support tool calling; models without it reject the request immediately. Pick a tool-capable model for PBL scenes, or change this scene's type. Original error: …

Failures **after** steps have completed rethrow unchanged, so mid-loop network/provider errors are not misdiagnosed.

Same shape as the merged #644 (surface TTS provider rate-limits as 429s): convert a cryptic provider failure into an actionable one at the layer that knows the cause.

## Scope (why `Refs` and not `Closes`)

This makes the failure clear and actionable; it does not make tools-incapable models able to generate PBL scenes. That needs a design decision on #664 — e.g. a single-shot JSON text-mode fallback (reusing the existing json-repair machinery) or a `tools` capability flag that gates `pbl` outlines per model. Happy to implement either if maintainers pick a direction; the agentic MCP loop is clearly a deliberate design so I didn't want to parallel it unilaterally.

## Verification
- New regression tests (`tests/pbl/generate-pbl-tool-rejection.test.ts`, 2 cases): zero-step rejection gets the wrapped message incl. the original error; post-step failure keeps its original error.
- `prettier` clean, `eslint` clean, `tsc --noEmit` error count unchanged, adjacent PBL/generation suites green (11/11).

### DIFF
--- a/lib/pbl/generate-pbl.ts
+++ b/lib/pbl/generate-pbl.ts
@@ -285,28 +285,46 @@ export async function generatePBLContent(
   // Run the agentic loop
   const systemPrompt = buildPBLSystemPrompt(config);
 
-  const _result = await callLLM(
-    {
-      model,
-      system: systemPrompt,
-      prompt: `Design a PBL project. Start in project_info mode by setting the project title and description.`,
-      tools: pblTools,
-      stopWhen: stepCountIs(30),
-      onStepFinish: ({ toolCalls, text }) => {
-        if (text) {
-          callbacks?.onProgress?.(`Thinking: ${text.slice(0, 100)}...`);
-        }
-        if (toolCalls) {
-          for (const tc of toolCalls) {
-            callbacks?.onProgress?.(`Tool: ${tc.toolName}`);
+  // PBL is the only generation path that requires native tool calling, and
+  // providers whose chat template lacks tools support reject such requests
+  // outright (e.g. LM Studio templates without a tools section, see #664).
+  // Track loop progress so a rejection before the first step — the signature
+  // of missing tool support — gets an actionable message instead of a raw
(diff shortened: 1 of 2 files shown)


wyuc (changes requested):
The diagnostic goal is useful, but `completedSteps === 0` is not specific to missing tool support. Any failure before `onStepFinish`—for example 401/403 auth errors, 429 rate limits, provider outages, DNS/network failures, or request timeouts—will be rewritten as “the selected model must support tool calling.” That can send users toward changing models when the actual problem is credentials or availability.

Please either classify a tool-capability rejection from reliable provider error metadata/message before presenting it as the cause, or make the zero-step fallback a neutral message that preserves the original failure and only suggests tool support as one possibility. Add negative tests for at least an immediate auth/rate-limit or network error so they are not diagnosed as a tool-capability failure.

wyuc:

PR #723, opened 2026-06-10
fix(agents): persist AgentBar voice picks and mode/selection across reloads

wyuc:
## Problem

User choices made in the AgentBar did not survive page loads:

1. **Per-agent voice picks were lost on refresh.** Picks were stored via `updateAgent(id, { voiceConfig })` on registry records, but the registry persist merge deliberately resets default agents from code (and generated agents are rebuilt from IndexedDB), so the pick silently rolled back to the default voice. Confusingly, the teacher's voice *did* survive, because the teacher pill writes the global `ttsVoice` setting instead.
2. **Visiting a classroom clobbered the AgentBar.** `app/classroom/[id]/page.tsx` unconditionally forced `setAgentMode('auto')` + reset `selectedAgentIds` whenever the stage had generated agents, so merely opening an old course wiped the user's preset/auto choice and agent selection.

Both reproduced end-to-end in a real browser before the fix.

## Approach

User UI preferences get one persistent home, the already-persisted settings store; the agent registry stays a derived runtime view (code + IndexedDB).

- **`agentVoiceOverrides`** (persisted map, keyed by agent id): written by the AgentBar voice picker, consulted at read time by `resolveAgentVoice` with priority *override → agent voiceConfig → deterministic fallback*, validated against enabled providers exactly like `voiceConfig`. `setAgentVoiceOverride(id, undefined)` clears an entry.
- **`agentSelectionIsUserSet`** (persisted flag): set only by real AgentBar interactions (switching to a *different* mode, toggling an agent); cleared whenever a classroom load or generation-preview writes a stage-derived selection. The new pure `restoreAgentSelection` honors the persisted mode/selection on classroom load only when it is user-set **and** still valid for the loaded stage (a preset selection of known non-generated agents, or an auto selection drawn from this stage's generated agents); otherwise it applies the previous stage-derived defaults and marks them not-user-set. This keeps the old behavior for fresh users, imported and server-hydrated classrooms, while a user's explicit choice survives refreshes and classroom visits.

## Known limitations (intentional)

- A user-set preset selection wins over an auto classroom's generated personas — user intent takes precedence. Switching back to Auto-generate in the AgentBar restores the generated roster.
- The picker has no "Auto" reset row yet; the settings API supports clearing an override (`undefined` deletes), UI affordance is a follow-up.

## Testing

- New unit tests: override precedence/validation in `resolveAgentVoice`, persisted override CRUD + real persist-rehydration round-trip (stubbed localStorage, fresh module import), `restoreAgentSelection` including sequential classroom round-trips (auto → preset → auto must not degrade the auto classroom).
- Full suite: 105 files / 796 tests green; `tsc --noEmit`, eslint (one pre-existing warning untouched), prettier clean.
- Browser e2e on a production build: voice pick survives a home refresh; preset mode + selection survive visiting a classroom and returning; a fresh profile visiting an auto classroom still gets its generated agents (no regression).

### DIFF
--- a/app/classroom/[id]/page.tsx
+++ b/app/classroom/[id]/page.tsx
@@ -87,17 +87,19 @@ export default function ClassroomDetailPage() {
       const { useSettingsStore } = await import('@/lib/store/settings');
       const { restoreAgentSelection } =
         await import('@/lib/orchestration/registry/agent-selection');
-      // Keep the user's persisted mode/selection when still valid for this
-      // stage instead of unconditionally forcing auto mode (which clobbered
-      // AgentBar choices on every classroom visit); fall back to the
-      // stage-derived defaults when the persisted choice is stale. Stale
-      // generated IDs (from another stage / pre-bleed-fix) never validate, so
-      // they don't resolve against a leftover registry entry.
+      // Keep the user's explicit AgentBar mode/selection when still valid for
+      // this stage instead of unconditionally forcing auto mode (which
+      // clobbered it on every classroom visit); fall back to the stage-derived
+      // defaults otherwise, marking them as NOT user-set so the next classroom
+      // never mistakes them for a choice. Stale generated IDs (from another
+      // stage / pre-bleed-fix) never validate, so they don't resolve against a
+      // leftover registry entry.
       const settings = useSettingsStore.getState();
       const registry = useAgentRegistry.getState();
       const stage = useStageStore.getState().stage;
-      const next = restoreAgentSelection({
+      const { selection: next, isUserSet } = restoreAgentSelection({
         persisted: { mode: settings.agentMode, selectedAgentIds: settings.selectedAgentIds },
+        persistedIsUserSet: settings.agentSelectionIsUserSet,
         generatedAgentIds,
         stageAgentIds: stage?.agentIds,
         isPresetAgent: (id) => {
@@ -110,6 +112,9 @@ export default function ClassroomDetailPage() {
       if (next.mode !== settings.agentMode) settings.setAgentMode(next.mode);
       if (next.selectedAgentIds !== settings.selectedAgentIds) {
         settings.setSelectedAgentIds(next.selectedAgentIds);
+      }
+      if (isUserSet !== settings.agentSelectionIsUserSet) {
+        settings.setAgentSelectionIsUserSet(isUserSet);
       }
     } catch (error) {
       log.error('Failed to load classroom:', error);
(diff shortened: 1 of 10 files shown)


wyuc:
**Review process summary** (3 internal review rounds before marking ready):

- **Round 1 (6 findings).** The most important one: the original restore logic kept any "persisted selection" that validated, without tracking where it came from — so the stage-derived defaults a classroom load writes could masquerade as a user choice on the next load, and an auto → preset → auto classroom round-trip would permanently downgrade the auto classroom to preset agents. Fixed by tracking provenance with a persisted `agentSelectionIsUserSet` flag: only real AgentBar interactions set it, and `restoreAgentSelection` honors the persisted selection only when it is user-set and still valid for the loaded stage. Also from this round: overrides are clearable (`setAgentVoiceOverride(id, undefined)` deletes), the override type is single-sourced in `voice-resolver`, persist rehydration is pinned by tests, and a comment overstating the old failure mode was tightened.
- **Round 2 (3 findings).** Clicking the already-active mode tab (a visual no-op) no longer promotes stage defaults to a user choice; generation-preview clears the flag when it writes the freshly generated roster; tests now assert setter writes actually reach the storage blob (guards against a future `partialize` omitting the new fields).
- **Round 3.** Zero actionable findings.

**Verification.** Full unit suite green (105 files / 796 tests), `tsc`/`eslint`/`prettier` clean. Manual browser verification on a production build: a non-teacher agent's voice pick survives a home refresh; preset mode + selection survive visiting a classroom and returning; a fresh profile visiting an auto classroom still gets its generated personas (no regression); switching back to Auto-generate restores the generated roster; zero console errors throughout.

cosarah (changes requested):
There is a state consistency issue when switching AgentBar from Preset back to Auto.

`handleModeChange('auto')` sets `agentMode = 'auto'` and marks `agentSelectionIsUserSet = true`, but it does not restore `selectedAgentIds` to the generated agent ids for the current auto-agent classroom. After a Preset -> Auto toggle, the persisted state can become:

```json
{
  "agentMode": "auto",
  "selectedAgentIds": ["default-1"],
  "agentSelectionIsUserSet": true
}
```

I verified this with Playwright by seeding an auto-generated selection (`["gen-teacher", "gen-student"]`) and then toggling Preset -> Auto in the real AgentBar UI. The final state was `agentMode: "auto"` but `selectedAgentIds` still pointed to preset agents.

This means the UI says Auto mode, while playback/discussion code still reads `selectedAgentIds` and continues using preset agents. It also makes the persisted “user-set auto” state invalid for `restoreAgentSelection`, which will discard it on reload.

Please either update `selectedAgentIds` to the valid generated agent ids when switching back to Auto, or avoid marking the selection as user-set unless a valid generated selection is written.

cosarah (approved):
LGTM, verified locally

PR #724, opened 2026-06-10
fix(renderer): make code entrance animation type line-by-line

### LINKED ISSUE #531: [Bug]: Code element entrance animation starts typing on all lines at once instead of line-by-line
﻿## Bug Description
The code element entrance animation is expected to reveal code line by line with a typewriter effect, but currently all lines begin typing at nearly the same time.

## Steps to Reproduce
1. Render a code element with animation enabled.
2. Use a multi-line code sample.
3. Watch the initial entrance animation.

## Expected Behavior
Each line should start after the previous line finishes, or at least after its configured stagger delay, producing a clearly line-by-line typewriter effect.

## Actual Behavior
All lines become active at roughly the same time, so the animation looks like parallel typing rather than a sequential line-by-line reveal.

## Deployment Method
Local development (pnpm dev / npm run dev / yarn dev)

## Browser
Chrome

## Operating System
Windows 11

## Relevant Logs / Screenshots
No console error.

Relevant implementation references:
- `components/slide-renderer/components/element/CodeElement/BaseCodeElement.tsx`: initial first-render animation states are computed for each line.
- `components/slide-renderer/components/element/CodeElement/BaseCodeElement.tsx`: those states are applied asynchronously via `setTimeout(..., 0)`.
- `components/slide-renderer/components/element/CodeElement/BaseCodeElement.tsx`: `CodeLineRow` derives its initial mounted/typing behavior from whether animation state is already present.
- `components/slide-renderer/components/element/CodeElement/BaseCodeElement.tsx`: per-line `typingDelay` is computed, but the delayed mount path appears to be bypassed on the first paint.

## Additional Context
This looks like an ordering issue on first render: rows mount before `animStates` is populated, so they enter the non-animated path first. Once animation state arrives, the rows are already mounted, which collapses the intended stagger into simultaneous typing.

tongshu2023:
## Summary

Fixes #531 — the code element's entrance animation typed all lines simultaneously instead of line-by-line.

## Root cause

Exactly the ordering issue the report hypothesized:

1. First render: `animStates` is an **empty Map** (it was only populated via a `setTimeout(..., 0)` inside an effect).
2. Every `CodeLineRow` therefore mounts with `animState === undefined` → `isNewLine === false` → `useState(!isNewLine || typingDelay === 0)` latches `mounted = true`. The `typingDelays` map is also empty at this point, so every row gets `typingDelay = 0`.
3. One tick later the states arrive, `isNewLine` flips to true for every row — but `mounted` is already latched true, so the delayed-mount path (`setTimeout(() => setMounted(true), typingDelay)`) is a no-op, and the `mounted && isNewLine` effect fires `setTyping(true)` for **all rows in the same tick**.

## Fix

Compute the first-render `'typing'` states synchronously in the `useState` initializer, so rows see their `animState` and cumulative `typingDelay` on the very first render: line 0 types immediately, each later row renders `null` until its stagger delay elapses. The effect now only handles subsequent line edits (`inserted`/`replaced`), with `prevLinesRef` seeded to the initial lines and the first effect run skipped.

**Behavior note:** previously, toggling `animate` false→true marked every line `inserted` (because `prevLinesRef` started empty). With the ref seeded, a toggle no longer replays the entrance — flagging in case that old behavior was intentional.

**Scope note:** `packages/@maic/renderer/src/elements/code/BaseCodeElement.tsx` has the identical first-render pattern, but #706 is currently rewriting that exact effect (prev-line map). To avoid conflicting with it, this PR only fixes the app copy; happy to port the same fix to the package copy once #706 lands.

## Verification

- Code-path trace above; `tsc --noEmit`, `eslint`, `prettier --check` all clean.
- No component-level test added: `vitest.config.ts` only includes `tests/**/*.test.ts` in a node environment (no jsdom/RTL infrastructure for `.tsx` behavior tests). If you'd like one, I can propose a minimal jsdom setup in a separate PR.


wyuc (approved):
LGTM. Verified the code block now reveals line-by-line on first paint instead of all at once. Good call scoping this to the app copy to avoid colliding with #706's package rewrite. Thanks!

PR #727, opened 2026-06-12
fix(importer): correct hanging-indent bullet slot rendering

xuyuanwei678:
## Summary
- Size the bullet slot to the hanging-indent amount (`|indent|`) instead of `marL`, so paragraphs with a wide `marL` no longer push body text past the bullet.
- Reset `text-indent` on the inline-block slot to stop symbol bullets (■ ◆) from drifting left onto adjacent shapes (e.g. slide 3 numbered circles).
- Only pad synthesized-marL symbol bullets; real-marL bullets sit at the natural hanging position to match PowerPoint.
- Add a vitest setup (`test` / `test:watch` scripts + dependency) and serializer tests covering bullets, tab stops, fills, empty paragraphs, tables, and math.

## Test plan
- [x] `pnpm --filter @maic/importer test`
- [x] Visually verify slide 2 (right text box, real marL) and slide 3 (numbered circles, synthesized marL) bullet alignment against the source PPTX.

Made with [Cursor](https://cursor.com)


wyuc (changes requested):
The core fix is correct and well-reasoned — I verified the geometry and the CSS behavior, and ran the new test suite locally (6 files / 17 tests, all pass). But the test wiring is broken: the test command in this PR's own test plan does not actually run, and CI never executes the new tests either. Requesting changes for that one issue.

## What I verified

- **Slot width `marL` → `|indent|`**: correct. The `<p>` already carries `margin-left:marL; text-indent:indent`, so the first line starts at `marL+indent`; a slot of width `|indent|` pushes body text exactly to `marL`, aligned with continuation lines. The old `width:marL` over-pushed the first line whenever a real `marL ≫ |indent|` was present. The synthesized-marL case (`marL == |indent|`) is unaffected.
- **`text-indent:0` on the slot**: correct and a real bug. CSS `text-indent` is inherited and applies to the inline-block's own first line, shifting the bullet glyph left by `N` out of the element. Zeroing it on the slot is the right fix.
- **`padding-left:16px` only for synthesized-marL symbol bullets**: the `marginLeftSynthesized` distinction is sound — real-marL bullets sit at the natural hanging position like PowerPoint.

## Blocker: the new tests never run

`pnpm --filter @maic/importer test` (the command in the PR test plan) fails with:

```
No test files found, exiting with code 1
include: tests/**/*.test.ts
```

There is no package-local vitest config, so vitest walks up and picks the **root** `vitest.config.ts`, whose `include` is `tests/**/*.test.ts` (plural) — while this PR puts tests in `test/` (singular). CI's `pnpm test` also only runs the root `tests/` suite, so the green check does not cover any of the 17 new tests.

Fix: add a `packages/@maic/importer/vitest.config.ts`:

```ts
import { defineConfig } from 'vitest/config';

export default defineConfig({
  test: {
    include: ['test/**/*.test.ts'],
  },
});
```

I verified locally that with this config the suite passes (17/17, ~3.5s). A package-local config is preferable to renaming the dir to `tests/`, because the root config's `setupFiles: tests/setup-env.ts` wouldn't resolve from the package dir anyway. Please also add a CI step (e.g. `pnpm --filter @maic/importer test` in the unit-test job) so these regressions are actually guarded.

## Minor (non-blocking)

- `vitest ^4.1.8` here vs `4.1.0` at the root puts two vitest versions in the lockfile — consider aligning them.
- `tableSerializer.test.ts` calls `tableToElement` at module top level (outside `it`); a parse failure there reports as a collection error instead of a test failure. Consider moving it into `it`/`beforeAll`.
- The assertions match serialized CSS strings (e.g. `width:48px;text-indent:0;`), so they're brittle to style-string formatting changes — acceptable for now, and the comments already note that real geometry needs visual verification. The two test-plan checkboxes (including the visual check of slides 2/3) are still unchecked; please confirm before merge.
- Edge case (pre-existing, fine to ignore): with synthesized marL and `|indent| < 16px`, `padding-left:16px` + border-box collapses the content width to 0 and the glyph overflows the slot. Not introduced by this PR.

wyuc (approved):
LGTM. Test wiring blocker resolved: package-local vitest config + CI step added, vitest aligned, table test moved into beforeAll. Verified locally at this head — `pnpm --filter @maic/importer test` discovers and passes all 6 files / 17 tests. Thanks for the quick turnaround.

PR #735, opened 2026-06-14
feat(maic-agent): Pro-mode "Edit with AI" agent + script timeline editor (v0)

wyuc:
## What

Adds the Pro-mode **"Edit with AI"** agent sidebar and a **script timeline**
editor for a scene's playback actions. v0 ships behind the editor flag.

### Agent sidebar
- assistant-ui runtime over an SSE `AgentEvent` stream; server agent built on a
  pinned `pi` core (see `lib/agent/VENDOR.md`) with an allowlist tool gate + quota seam.
- v0 capability = `regenerate_scene_actions` only; the system prompt bounds the
  agent to regenerating scene narration (no false claims about editing slides).
- UI: solid-violet user bubble, tool card with status badge, quick-prompt chips,
  shimmer "Thinking…" indicator, streaming markdown (Streamdown), stop-response.

### Script timeline (讲解脚本)
- horizontal film-strip of the scene's actions: speech clips + cue cards in
  playback order; hovering a cue replays the real spotlight/laser on its element.
- inline editing: edit speech text, drag to add (speech/spotlight/laser),
  drag to reorder, delete.
- element targeting via an on-canvas pick layer (outlines selectable elements,
  live preview, draggable/collapsible element panel).
- per-speech managed TTS: audio status + preview + regenerate (reuses
  `/api/generate/tts` + the IndexedDB audio cache).
- fix: a laser cue replays as a laser pointer, not a spotlight.

## Test
- `vitest run tests/edit tests/lib/agent` green; `next build` passes; type-checked.

Draft for visual/UX review.


wyuc:
## Code review — 3 rounds, converged (7 → 1 → 1 findings)

Ran a high-effort multi-angle review over this round's changes. All findings fixed and pushed; each fix is type-checked, builds, and `tests/edit` is green.

**Round 1 — correctness + architecture**
- Duplicated slide: cues no longer dangle on stale element ids or share audio-cache keys with the source (reseed action ids, remap `elementId`, drop stale `audioId`).
- Timeline TTS now uses the pipeline's canonical key `tts_s<order>_<id>` and reuses `generateAndStoreTTS`; one `resolveSpeechAudioId` for audio identity.
- Async TTS commits re-read the store and stamp by id, so a concurrent edit isn't clobbered; speech clip keeps a dirty flag so an agent regeneration mid-edit survives blur.
- `pickTarget` keyed by `actionId` (not a positional index); element binding goes through `setElementIdById`; `clearAllEffects` clears it.
- `setElementId` only writes element-bound cue types.
- Consolidated the cue taxonomy, preview side-effects, and audio identity into shared modules (`cue-meta`, `cue-preview`, `regenerate-speech-tts`) — previously duplicated across ActionsBar / regenerate-tool-ui / ElementPickLayer; removed dead palette type; pick-layer hit-test throttled to rAF.

**Round 2 — refactor regressions**
- Deep-clone duplicated-scene actions (nested cue data was aliased by reference).

**Round 3 — agent runtime**
- Guard a superseded run on stop-then-resend, so a stale aborted run can't reset `isRunning` / rewrite the latest message.

**Deferred (not bugs)**
- Converging the edit-mode and playback laser overlays — touches the playback path, needs visual QA before merging.
- Splitting ActionsBar into per-component files — organization only; the cross-file duplication is already removed.

cosarah (commented):
Findings

- P1: CI is currently blocked. `pnpm check` fails Prettier on 19 files, and `pnpm lint` fails on changed code: misplaced `eslint-disable` / `any` in `components/ai-elements/prompt-input.tsx:1023`, the same issue in `lib/hooks/use-audio-recorder.ts:11`, and `setState` inside an effect in `components/edit/ActionsBar/ActionsBar.tsx:363`.

- P1: Stopped or superseded agent runs can still mutate the latest run and stage actions. `handleEvent` applies `tool_execution_end` results directly in `lib/agent/client/use-agent-runtime.ts:111`, while `refresh` always rewrites the last message in `lib/agent/client/use-agent-runtime.ts:146`. The superseded guard only exists in `finally` at `lib/agent/client/use-agent-runtime.ts:231`. A stop-then-resend can let old SSE events overwrite the new assistant message or apply stale regenerated actions.

- P2: Duplicating a slide drops `audioId` but leaves `audioUrl` on speech actions. The clone logic deletes only `audioId` in `lib/edit/slide-defaults.ts:88`, but playback prefers `audioUrl` in `lib/action/engine.ts:221`. Duplicated slides with server-generated TTS will keep playing the source slide’s audio, and regenerating or stamping a new `audioId` will not matter while the stale `audioUrl` remains.

- P2: Timeline edits can clobber concurrent action updates because several handlers commit the render-time `actions` snapshot. Examples: drop/reorder in `components/edit/ActionsBar/ActionsBar.tsx:708`, speech blur commit in `components/edit/ActionsBar/ActionsBar.tsx:893`, and delete/move in `components/edit/ActionsBar/ActionsBar.tsx:903`. If the agent or TTS updates actions while a user is dragging or editing, the later UI commit can revert those updates.

wyuc:
Thanks @cosarah — all four addressed in `2287eca` (+ `0d63d27`). CI should be green now: `pnpm lint` exits 0 (0 errors), `prettier --check` clean on changed files, `tsc`/`tests/edit`/`next build` all pass.

- **P1 CI**: moved the misplaced `eslint-disable` above the `any` line in `prompt-input.tsx` / `use-audio-recorder.ts` (it was annotating the comment line, not the declaration); disabled `set-state-in-effect` on the speech-clip external-text sync; ran Prettier on the changed files; memoized `ElementPickLayer` `elements`.
- **P1 stop/superseded run**: `handleEvent` and `refresh` are now gated on the run still owning `abortRef` (an `isCurrent()` check), not just the `finally`. A superseded run no longer applies `tool_execution_end` results or rewrites the new run's message.
- **P2 duplicate audioUrl**: `duplicateSlideScene` now drops `audioUrl` as well as `audioId`, so a copy can't keep playing the source's server-generated TTS. Actions are also deep-cloned so nested cue data isn't aliased.
- **P2 stale-snapshot commits**: `commit` now applies an updater against the latest store actions; every timeline edit (drag/reorder/blur/delete/move/regenerate) goes through it, so a concurrent agent/TTS update isn't reverted by a render-time snapshot.

Two items intentionally deferred (not bugs): converging the edit/playback laser overlays (touches the playback path — wants visual QA), and splitting ActionsBar into per-component files (org-only; the cross-file duplication is already removed).

cosarah (commented):
Findings

- P1: CI is still blocked by Prettier. `pnpm check` fails with 10 unformatted files, including `app/api/agent/edit/route.ts`, `components/edit/EditShell/EditShell.tsx`, and `lib/agent/runtime/stream-fn.ts`. The GitHub “Lint, Typecheck & Unit Tests” job remains red.

- P2: `ActionsBar` now reads the latest store state before writing actions back, but many operations still identify the target action by the render-time `index`. If the agent or TTS updates `actions` while the user is dragging, editing, or deleting, the later blur/drop/click handler can apply that stale index to the latest array and edit/delete/move the wrong action. Examples are `components/edit/ActionsBar/ActionsBar.tsx:772`, `components/edit/ActionsBar/ActionsBar.tsx:919`, `components/edit/ActionsBar/ActionsBar.tsx:972`, `components/edit/ActionsBar/ActionsBar.tsx:978`, and `components/edit/ActionsBar/ActionsBar.tsx:990`. These handlers should carry `actionId` and resolve the current index from the latest array inside the commit.

wyuc:
Good catches — fixed in `d2c19ad`. My earlier "CI green" was wrong because I'd only formatted this round's diff, not the repo.

- **Prettier**: ran `prettier . --write` repo-wide; `prettier . --check` is now clean (the 10 files were mostly from earlier commits on the branch).
- **E2E**: the `slide-scene-creation-gate` spec asserted the *old* gated behavior; updated it to the new contract — creation is enabled, so the rail exposes insert zones and the per-slide menu is Rename + Duplicate + Delete.
- **Stale index**: timeline edits now identify the target by `actionId` and resolve the index against the latest array inside `commit` (`setSpeechTextById` / `removeById` / `moveById` / `moveByIdDir`; the drag payload carries the id). A concurrent agent/TTS update can no longer make a stale render-time index touch the wrong action.

Locally: `pnpm lint` exit 0, `prettier . --check` clean, `tsc` clean, `tests/edit` + `tests/lib/agent` (249) green, `next build` passes. Watching the CI run now.

cosarah (commented):
Finding: Sending a message from the AI agent panel can crash the editor. 
Trigger: after the user enters edit mode and sends an AI agent message, the first assistant message mounts. The console reports `The result of getSnapshot should be cached to avoid an infinite loop` at `components/edit/AgentPanel/AgentPanel.tsx:65` (`AssistantMessage`) and then React throws `Maximum update depth exceeded` from the edit chrome render path. 
Root cause: the `useMessage` selector in `AssistantMessage` returns a fresh object literal (`{ showThinking, stoppedEmpty }`) on every snapshot. `useMessage` is backed by `useSyncExternalStore`, which compares snapshots by `Object.is`, so equal boolean values still appear changed because the object reference is new. Once a message is sent and the assistant message starts streaming/subscribing, this unstable snapshot causes repeated updates until React hits the depth limit.

wyuc:
Fixed in `ad030c3` — good find, that crash was mine. `AssistantMessage`'s `useMessage` selector now returns a single string phase (`'thinking' | 'stopped' | 'content'`) instead of a fresh `{…}` literal, so the useSyncExternalStore Object.is compare is stable and the render no longer loops. tsc / prettier / build all green.

components/edit/AgentPanel/AgentPanel.tsx
```diff
@@ -0,0 +1,284 @@
...
+import { useAgentRuntime } from '@/lib/agent/client/use-agent-runtime';
+import { MarkdownText } from './markdown-text';
+import { RegenerateSceneActionsUI } from './regenerate-tool-ui';
+
+const MIN_WIDTH = 320;
+const MAX_WIDTH = 640;
+const DEFAULT_WIDTH = 384;
+
+/** Quick-prompt chips — one tap prefills the composer (user reviews, then sends). */
+const QUICK_PROMPTS = ['重新生成讲解旁白', '让讲解更口语一些', '加一个生活化类比'];
```
cosarah: **P1 — i18n missing (project convention).** All new UI strings are hardcoded Chinese here and across `ActionsBar`, `ElementPickLayer`, `regenerate-tool-ui` and the `cue-meta` labels (`有什么想改的？`, these `QUICK_PROMPTS`, `讲解脚本`, `拖入添加`, etc.). Sibling editor components (`CommandBar`, `ThumbItem`, `QuizForm`, …) all go through `useTranslation`/`t()`, and the project rule requires i18n for UI text. This is the main thing to address before leaving draft — inconsistent and unlocalizable as-is.

components/edit/ActionsBar/ActionsBar.tsx
```diff
@@ -0,0 +1,1022 @@
...
+        }
+      }
+      commit((cur) => {
+        let next = cur;
+        for (const a of cur) {
+          if (a.type === 'speech' && a.id)
+            next = setAudioIdById(next, a.id, speechAudioId(order, a.id));
+        }
+        return next;
+      });
```
cosarah: **P2 — stamps `audioId` on every speech line, including failed/skipped ones.** This loop stamps `setAudioIdById` for *every* speech in `cur` with an id, but the regen above skips empty-text lines and swallows per-line failures (the `catch`). So a line whose TTS never generated still gets an `audioId` pointing at a non-existent blob (dangling ref). Not fatal (`audioExists` returns false and the derived key matches), but the semantics are wrong — only stamp lines that actually generated successfully (e.g. collect the ids returned by `regenerateSpeechAudio` and stamp those).

lib/edit/scene-creation-enabled.ts
```diff
@@ -1,11 +1,11 @@
...
  *
- * Flip to `true` once newly-created scenes are made playable.
+ * Enabled now that the editor can author a scene's playback `actions`:
+ * duplicated slides carry the source's actions (playable as-is), and a blank
+ * inserted slide starts with no narration — the user fills it in via the
+ * script timeline / MAIC Agent (the playback engine skips a scene with no
+ * actions until then). Reorder / delete / rename were always playback-safe.
  */
-export const SCENE_CREATION_ENABLED = false;
+export const SCENE_CREATION_ENABLED = true;
```
cosarah, line 11: **P3 — blank inserted slide is silently skipped in playback (UX foot-gun).** With this flipped to `true`, a blank inserted slide ships with zero `actions`, so the playback engine skips the whole slide until the user authors narration. The comment acknowledges it, but from the user's side: insert a slide, hit play, the slide vanishes. Consider seeding the inserted slide with a default speech action, or surfacing a visible "no narration" indicator in the timeline / nav so the skip isn't a surprise.

components/edit/ActionsBar/ActionsBar.tsx
```diff
@@ -0,0 +1,1022 @@
...
+            ? crypto.randomUUID()
+            : `a-${Date.now()}`;
+        const action = makeAction(p.type, id);
+        commit((cur) => insertAt(cur, slot, action));
+        if (p.type === 'speech') setFocusId(id);
+      } else {
+        commit((cur) => moveById(cur, p.id, slot));
+      }
+    },
+    [actions, commit],
```
cosarah: Nit (eslint warning): `actions` is an unnecessary dependency here — `handleDrop` reads the latest state via `commit()` and never touches `actions`. Drop it from the dep array.

lib/agent/tools/regenerate-scene-actions.ts
```diff
@@ -0,0 +1,239 @@
...
+      'Use this after the scene content has been modified (e.g. slide elements changed, quiz questions updated) ' +
+      'so that the actions stay in sync with what is actually on screen. ' +
+      'Only supply the sceneId — the scene data is loaded automatically.',
+    parameters: RegenerateSceneActionsParams,
+
+    execute: async (_toolCallId, params) => {
+      const { sceneId, previousSpeeches, userProfile } = params;
+
+      // ── Resolve trusted scene context from deps (not from model args) ──
+      const ctx_data = deps.getSceneContext(sceneId);
```
cosarah: Nit: `ctx_data` uses snake_case, inconsistent with the camelCase used everywhere else in the repo. `sceneCtx` / `ctxData`.

cosarah (commented):
Reviewed locally — `vitest run tests/edit tests/lib/agent` (249 passed), `tsc --noEmit` and `eslint` all green (1 warning). Architecture is clean: pure `actions-edit` ops, single-source `cue-meta`, by-id (index-stale-safe) edits, latest-state `commit()`, the superseded-run guard, and the non-empty-actions guard against destructive wipes are all nicely done. The trusted-context design (model only supplies `sceneId`) and the allowlist gate + tight system prompt are good guardrails.

Leaving a few inline notes — the i18n gap (P1) is the main one to resolve before this leaves draft; the rest are medium/low.

components/edit/surfaces/slide/ElementPickLayer.tsx
```diff
@@ -0,0 +1,324 @@
...
+    setCollapsed(false);
+    const onResize = () => measureOutlines();
+    window.addEventListener('resize', onResize);
+    window.addEventListener('scroll', onResize, true);
+    return () => {
+      window.removeEventListener('resize', onResize);
+      window.removeEventListener('scroll', onResize, true);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [pickTarget?.sceneId, pickTarget?.actionId, elements.length]);
```
cosarah, line 137: Nit: outlines only re-measure on `elements.length` change (plus resize/scroll). If an element moves/resizes without a count change, the outline boxes go stale until the next resize/scroll event. Acceptable for v0, just flagging.

app/api/agent/edit/route.ts
```diff
@@ -0,0 +1,127 @@
...
+   * Trusted scene/stage context for every scene the agent may act on.
+   * The client includes the active scene (and all sibling scenes) so the
+   * `regenerate_scene_actions` tool can resolve outline + content without
+   * relying on model-fabricated arguments.
+   */
+  sceneContextMap?: SceneContextMap;
+}
+
+export async function POST(req: NextRequest) {
+  if (!isMaicEditorEnabled()) {
```
cosarah, line 43: The endpoint is gated only by the editor feature flag, no additional auth. Consistent with the existing `/api/generate/*` routes, so this is likely intended — just confirming it's deliberate since this one drives an agent loop.

wyuc:
Thanks for the thorough review @cosarah — all addressed in `b19a02e`.

**P1 — i18n (the gate):** Every new UI string now goes through `useI18n`/`t()` like the sibling editor components. `cue-meta` no longer holds Chinese literals — it exposes hook-free i18n **label keys** that callers resolve via `t()`, so the pure module stays hook-free while the text is localizable. Added 68 keys to **all 7 locales** under `edit.{cue,element,agent,timeline,tts,pick,regen}` (the locale-parity test stays green). Covers `AgentPanel`, the `ActionsBar` timeline, `ElementPickLayer`, the regenerate tool card, and the cue/element labels.

**P2 — dangling `audioId`:** "Regenerate all voiceover" now collects the ids that actually synthesized and stamps only those (and only fires the refresh when at least one succeeded), so a skipped/empty/failed line can no longer get an `audioId` pointing at a blob that was never written.

**P3 — blank inserted slide:** The blank slide is now seeded with a single empty speech clip in `createBlankSlideScene`, so it stays playable (no more silent skip on play) and surfaces a "fill me in" starting line in the timeline. While verifying this I caught a follow-on edge case: an empty-text speech line under the default browser-native TTS would speak an empty `SpeechSynthesisUtterance`, which doesn't reliably fire `onend` in Chromium and would hang playback on that slide. Fixed by routing an empty/whitespace speech line to the reading timer instead, with `splitIntoChunks` returning no chunks for blank text as defense-in-depth.

**Nits:** dropped the stale `actions` dep from `handleDrop` (the eslint warning is gone); renamed `ctx_data` → `ctxData`.

**On the remaining notes:**
- *ElementPickLayer outline re-measure* — left as-is for v0, per your flag.
- */api/agent/edit auth* — yes, deliberate: gated by the editor feature flag, consistent with the existing `/api/generate/*` routes.
- *`{{count}} actions` pluralization* — left as-is for now (consistent with the other count strings in the editor); can revisit with proper i18next plural keys if we want it across the board.

CI is green on `b19a02e`.

sophietao20-star:
Tested locally — A few issues currently impact the authoring workflow.

### Bugs

1. **Agent model config is inconsistent with course generation**
   Course generation uses the frontend model config, while `/api/agent/edit` falls back to server-side providers. Generation can work with the frontend model config while the Agent fails due to missing server-side keys.
   **Expected:** Agent editing should use the same active frontend model config.
2. **Agent regenerates even when requested constraints cannot be guaranteed**
   For requests like "rewrite narration but keep existing spotlight/laser counts and bound elements unchanged", the Agent still runs regenerate despite not being able to guarantee the constraint.
   **Expected:** explain that the constraint is unsupported and avoid regeneration.

### UX / Product Polish

1. **Clarify TTS capability differences**
   Users have no expectation that browser-native TTS does not support timeline audio generation features (preview/regenerate/bulk regenerate). This should be surfaced when selecting a TTS provider.
2. **Handle stale audio after speech edits**
   Speech edits can leave existing audio stale, so that the audio and text may be out of sync
   Suggestion: discard the audio after the speech was edited so that regeneration is required, or at least mark it as outdated.
3. **Explain disabled Pro-mode on the course-complete placeholder**
   Add a hover explaining why the Pro-mode button is disabled on this page, and direct users to a real scene.
4. **Prevent unsupported actions on interactive scenes**
   Spotlight/laser actions can still be added even though required element bindings cannot be completed.
   Suggestion: disable unsupported actions on interactive scenes.
5. **Speech editing remains enabled on interactive scenes**
   Interactive pages are intended to be read-only in v0, but speech still supports the full edit flow (add/edit/voice/play), which does not match either the intended scope or the "Interactive · view-only" banner.
   Suggestion: disable speech editing, consistent with spotlight and laser actions.
6. **Preserve slide names in Agent chat history**
   Conversations across pages are merged together, making it difficult to identify which page a message refers to.
   Suggestion: preserve the original `@slide` reference.

### Suggested New Capabilities

1. **Undo for AI results**
   Agent regeneration directly overwrites the current page’s actions. There should be at least one-step rollback so users can recover from a bad AI result.
2. **Local undo/redo for the actions timeline**
   The timeline supports drag, delete, speech edits, and AI overwrites. Local undo/redo would reduce the cost of mistakes.

lib/agent/client/use-agent-runtime.ts
```diff
@@ -0,0 +1,291 @@
...
+                  })),
+            content: scene.content,
+            stageId: scene.stageId,
+            languageDirective: stage?.languageDirective,
+          };
+        }
+
+        const res = await fetch('/api/agent/edit', {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
```
sophietao20-star: Course generation uses the frontend model config, while `/api/agent/edit` falls back to server-side providers. Generation can work with the frontend model config while the Agent fails due to missing server-side keys. The failure currently surfaces as a generic 500 error (see screenshot below).

Image

Expected: Agent editing should use the same active frontend model config.

lib/agent/runtime/build-agent.ts
```diff
@@ -0,0 +1,68 @@
...
+    ? `The current slide is id="${scene.id}" with title "${scene.title}".`
+    : 'There is no active slide.';
+  return [
+    'You are the MAIC Editor assistant, embedded in the slide editor sidebar.',
+    sceneLine,
+    // Capability boundary — keep this tight. The agent has exactly ONE tool
+    // (regenerate_scene_actions). Without firm limits the model cheerfully
+    // claims it can rewrite slide text, images and layout, which it cannot.
+    "Your ONLY capability is regenerating the current scene's spoken narration and its playback actions (讲解旁白/动作), by calling the `regenerate_scene_actions` tool with only the sceneId — outline and content are resolved automatically, so never fabricate them.",
+    'You CANNOT do anything else. You cannot edit slide text, titles, bullets, images, layout, colors or styles; you cannot add, delete, reorder or duplicate slides; you cannot insert quizzes or modify the whiteboard; you cannot edit the slide / PPT / canvas directly. The user edits slide content themselves on the canvas.',
```
sophietao20-star, line 64: The prompt covers unsupported slide/PPT edits, but not constrained regeneration requests. For requests like "rewrite narration but keep existing spotlight/laser counts and bindings unchanged", the current v0 tool cannot reliably guarantee the constraint, yet the Agent still regenerates.
Suggestion: explicitly instruct the Agent to refuse such requests and explain the limitation to the user rather than applying a potentially destructive regenerate.

sophietao20-star (commented):
Left two inline comments for the bug-level issues found during local product testing.

The remaining UX polish and follow-up suggestions are collected in the PR comment:
https://github.com/THU-MAIC/OpenMAIC/pull/735#issuecomment-4728669554

wyuc:
Thanks for the thorough product pass @sophietao20-star — all addressed in `62e77d8`.

### Bugs
1. **Model config** — agent edit now sends the active frontend model config (`x-model`/`x-api-key`/`x-base-url`/`x-provider-type` headers, same as course generation) plus `thinkingConfig` in the body, instead of falling back to a server-side provider. Fixes the 500 when no server key is configured, and now honors the user's active thinking budget/level too. (Server-managed providers are unaffected — the server key stays authoritative for those.)
2. **Constrained regeneration** — the system prompt now refuses requests like "rewrite the narration but keep the existing spotlights/lasers/bindings unchanged" and explains that v0 regenerates the whole scene and can't guarantee that constraint, rather than running a destructive wholesale regenerate.

### UX / consistency
1. **TTS capability** — the settings panel now surfaces a notice when browser-native TTS is selected: the timeline's per-line audio (preview / regenerate / bulk voiceover) is unavailable on it.
2. **Stale audio after edits** — editing a speech line now discards its cached audio (the blob + the stamped fields), so the stale audio can't replay for the new wording; the line reads as un-voiced until regenerated.
3. **Disabled Pro-mode** — the Pro-mode toggle now explains on hover why it's disabled on non-editable pages (e.g. the course-complete placeholder) and points the user to a real scene.
4 & 5. **Read-only scenes** — read-only scene types (no registered editor surface, e.g. interactive / PBL) no longer mount the narration timeline **or** the AI edit panel, so speech / spotlight / laser can't be edited there, consistent with the "· view-only" badge. **Scope note:** I gated the AgentPanel too (not just the timeline) — leaving an AI tool that rewrites a scene's actions on a view-only scene seemed to contradict the read-only intent. Happy to revert to timeline-only if you'd rather keep the panel available there.
6. **Chat history per page** — kept the chat as a single global thread (Cursor-style, per your read), and instead each regenerate tool card now carries an `@<scene>` pill identifying the page it acted on.

### New capabilities → follow-ups
The two undo/redo asks are larger features, so I've split them out: #757 (one-step undo for AI regeneration) and #758 (local undo/redo for the timeline).

CI is green on `62e77d8`.

wyuc:
Heads-up @cosarah @sophietao20-star — to unblock the next phase, this PR has been merged into `feat/maic-editor-v1` (fast-forwarded, full history preserved).

Thank you both for the thorough reviews. Everything raised across the rounds was addressed:
- i18n: all new editor UI strings routed through `useI18n`/`t()`, keys added across every locale (incl. ko-KR);
- TTS/audio: editing a speech line now invalidates its cached audio, and an empty-text line no longer hangs browser-native playback;
- read-only scenes (interactive/PBL) no longer mount the timeline or AI panel — consistent with the "view-only" badge;
- agent model config: the editor agent now uses the active frontend model config + is a routable `maic-agent` stage for `MODEL_ROUTES`;
- the regenerate tool card now carries an `@scene` pill (and no longer overflows on a narrow rail).

The two undo/redo suggestions are tracked as follow-ups: #757 (one-step AI rollback) and #758 (timeline undo/redo).

For anything still open or newly noticed: please park it for the next-phase PR on the editor line — I'll carry it forward there. Thanks again!

Issue #720, opened 2026-06-10
[Epic] @maic SDK family: from in-app implementations to package consumers


wyuc:
## Goal

Evolve the app (and downstream consumers) from "ships its own slide stack" to a
standard consumer of the `@openmaic/{dsl, renderer, importer, exporter, storage}` package
family. End state: the app keeps only business glue (media-store resolution,
lazy placeholders, app chrome); every contract, renderer, importer and exporter
lives in a package with a published boundary.

Dependency rule (established in #668, kept acyclic):
`@openmaic/dsl` → nothing; `renderer` / `importer` / `exporter` → `@openmaic/dsl` only.

> **Scope note:** the family ships under the **`@openmaic`** npm scope (the bare
> `maic` name blocks an `@maic` org); matches the existing `@openmaic/docs`.

## Where we are (done)

- [x] #668 extracted `@openmaic/dsl` (pure contract), `@openmaic/renderer` (read-only
      canvas, v1), `@openmaic/importer` (pptx → DSL); app consumes the importer.
- [x] #707 made the app consume `dsl` + `renderer`: `lib/types/slides.ts` is a
      thin re-export shim of `@openmaic/dsl`; the three read-only thumbnail
      surfaces render through `SlideCanvas` (with reactive `gen_*` media
      resolution in the app-side wrapper).
- [x] #778 / #780 publish foundation: `@openmaic` org/scope claimed, `files` /
      `exports` / `prepublishOnly` build guards, ESM-only builds, a tag-triggered
      publish workflow with npm provenance, and placeholder `0.0.1` of all three
      packages on the `placeholder` dist-tag.

## Phase 1 — Cleanup & consolidation (low risk, immediately actionable)

- [x] Delete dead legacy `ThumbnailSlide/` + `ThumbnailElement` (zero imports
      after #707, ~217 LOC)
- [x] Repoint ~90 `@/lib/types/slides` import sites directly at `@openmaic/dsl`,
      then delete the shim
- [x] Fix `@openmaic/dsl` build: extensionless re-exports broke bare-node ESM
      resolution — resolved (ESM-only builds) in #780
- [x] Publish foundation in place (#778, #780) — see "Where we are"
- [x] Cut the real `0.1.x` `latest` release (push `@openmaic/<pkg>@<version>`
      tags) — gated on the Phase 2 contract changes / a real consumer

## Phase 2 — Grow the DSL (needs contract design)

[...]

PR #736, opened 2026-06-15
chore(phase-1): drop dead ThumbnailSlide + fix @maic/dsl Node ESM resolution

wyuc:
First Phase-1 cleanup from the SDK-consumption epic (#720), follow-up to #707.

## What

**1. Remove dead `ThumbnailSlide` / `ThumbnailElement`** (~215 LOC). These were the in-app read-only thumbnail renderers that `SlideThumbnail` (→ `@maic/renderer`) replaced in #707. They have zero importers now — the editor nav rail renders through `SceneThumbnailContent` → `SlideThumbnail`; the only remaining `ThumbnailSlide` mentions are an unrelated local type alias / util name and historical doc comments.

**2. Make `@maic/dsl` resolve under Node ESM.** The package built with `moduleResolution: bundler` + extensionless re-exports, so `dist/index.js` emitted `export * from './slides'`, which fails bare-node ESM resolution (`ERR_MODULE_NOT_FOUND`) even though Next's bundler tolerated it. Switched the package tsconfig to `NodeNext` and added `.js` extensions to the relative specifiers; `node -e "import('@maic/dsl')"` now loads, and NodeNext enforces extensions going forward.

## Test plan

- `tsc --noEmit` clean for the app, `@maic/dsl`, `@maic/renderer`, and `@maic/importer` (all consume the dsl `.d.ts`).
- `node -e "import('@maic/dsl')"` resolves (16 exports) — previously threw.
- `next build` OK; e2e (Playwright) 19/19 green.

Part of #720 (Phase 1).


cosarah (approved):
LGTM，verified locally

PR #738, opened 2026-06-15
refactor(types): consume @maic/dsl directly and drop the slides re-export shim

wyuc:
Phase-1 cleanup from the SDK-consumption epic (#720), follow-up to #707/#736.

## What

#707 left `lib/types/slides.ts` as a thin `export * from '@maic/dsl'` shim so the ~100 existing `@/lib/types/slides` import sites kept compiling unchanged. This points them all at `@maic/dsl` directly and deletes the shim — the app now consumes the package contract with no indirection, and the legacy module path is gone.

Pure module-specifier migration: every symbol imported from `@/lib/types/slides` was already a re-export of the same `@maic/dsl` symbol, so behavior is identical. Besides the ~100 alias imports it also redirects:
- the two `lib/types` relative importers (`./slides`),
- the e2e fixture's relative import,
- the dynamic `import('../types/slides').Slide` type alias in `stage-storage`,
- and merges the now same-module import pair in `slide-edit-elements`.

## Test plan

- `tsc --noEmit` clean; `eslint` clean (no new warnings).
- `vitest run` 804/804; `next build` OK; e2e (Playwright) 19/19.
- No remaining references to `lib/types/slides` anywhere in source.

Part of #720 (Phase 1).


cosarah (approved):
LGTM, verified locally

Issue #140, opened 2026-03-19
feat: support DOCX file upload


YizukiAme:
## Summary

Add `.docx` as a supported document type alongside existing PDF and text uploads.

## Motivation

Many users have course materials in Word format. Currently they need to manually convert to PDF or text before uploading, which adds friction.

## Proposed Approach

- Extend `SupportedDocumentType` to include `docx`
- Use [mammoth](https://github.com/mwilliamson/mammoth.js) (dynamic import) to extract raw text from DOCX files
- Reuse existing upload validation, IndexedDB storage, and `MAX_PDF_CONTENT_CHARS` truncation flow
- Add DOCX parse error handling with user-facing toast
- Update i18n strings

## Dependencies

Builds on the text/Markdown upload infrastructure from #41.

## Notes

Implementation is ready on a local branch, will open a PR after #41 is merged.

PR #41, opened 2026-03-16
feat: support MD/TXT file upload alongside PDF

YizukiAme:
## What

Enable uploading **Markdown** (`.md`, `.markdown`) and **plain text** (`.txt`) files as course material, alongside existing PDF support. Text files bypass the OCR pipeline entirely — their content is read client-side and passed directly to the generation pipeline.

## Why

Many users have learning materials in Markdown or plain text format. Previously, the upload UI only accepted `.pdf` files, requiring users to convert their notes before uploading.

## Changes

| File | Change |
|------|--------|
| `lib/utils/document-upload.ts` | **[NEW]** Shared helpers — `getSupportedDocumentType()`, `readTextFileContent()`, `SUPPORTED_DOCUMENT_ACCEPT` constant |
| `components/generation/generation-toolbar.tsx` | Use shared `getSupportedDocumentType()` for validation; widen `accept` filter; update UI label |
| `app/page.tsx` | Branch on document type — text files are validated (empty check), then stored to IndexedDB via `storePdfBlob()`, same as PDF |
| `app/generation-preview/page.tsx` | Add text document branch — deferred read from IndexedDB, apply `MAX_PDF_CONTENT_CHARS` truncation, reuse truncation warning |
| `app/generation-preview/types.ts` | Add `documentType` field to `GenerationSessionState` |
| `lib/i18n/common.ts` | Add `docUpload` i18n key |
| `lib/i18n/generation.ts` | Update file limit string to "PDF / Markdown / TXT"; add `emptyFile` error message |

## How it works

```
Text file (.md/.txt)  →  empty check  →  storePdfBlob() to IndexedDB  →  generation-preview reads blob, extracts text, truncates
PDF file (.pdf)       →  storePdfBlob() to IndexedDB                  →  generation-preview runs OCR via /api/parse-pdf as before
```

Both paths store the original file in IndexedDB and only save a reference key in `sessionStorage`, avoiding `sessionStorage` quota issues. Content truncation (`MAX_PDF_CONTENT_CHARS`) is applied uniformly in the generation-preview stage.

## Key design decisions

- **IndexedDB for all file types** — text files go through `storePdfBlob()` just like PDFs, avoiding `sessionStorage` size limits
- **Deferred text extraction** — text content is read and truncated in generation-preview, not at upload time, ensuring consistent `MAX_PDF_CONTENT_CHARS` enforcement
[...]

PR #741, opened 2026-06-15
feat(document): add multi-format course material upload

jackefn:
## Summary

Implement MAIC ETL Milestone 2 by moving the upload flow from PDF-only upload to single-file course material upload.

This PR adds a generic document extraction API, supports PDF/DOCX/PPTX/TXT/Markdown uploads from the homepage, routes files through the document extractor boundary, and keeps the existing generation compatibility shape stable.

## Related Issues

Related to #621, #611, #41, #140

## Changes

- Add `/api/extract-document` for generic course material extraction.
- Add shared MIME helpers and upload accept-list for PDF, DOCX, PPTX, TXT, and Markdown.
- Add a local `plain-text` extractor for TXT and Markdown so these formats do not depend on MinerU.
- Extend the self-hosted MinerU document extractor to support PDF, DOCX, and PPTX.
- Update the homepage upload UI from "Upload PDF" to "Upload course material".
- Preserve single-file upload behavior and existing generation compatibility fields.
- Select providers by explicit PDF parser choice or by extractor capability match.
- Return clear diagnostics for unsupported file types and unsupported provider/format combinations.
- Update generation preview copy to show the uploaded file type dynamically.
- Add document extractor tests for registry behavior, PDF compatibility, and TXT/Markdown extraction.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Start OpenMAIC locally and upload a PDF, TXT, or Markdown file from the homepage course material upload control.
2. Configure MinerU and upload a DOCX or PPTX file to confirm provider-backed extraction.
3. Upload an unsupported file type or force an unsupported provider/format combination and confirm a clear error is returned.

- TXT and Markdown extraction runs locally through `plain-text` and no longer calls MinerU.
- PDF extraction still works through the existing PDF-compatible path.
- DOCX/PPTX are routed to MinerU by capability match when MinerU is configured.
- Unsupported formats and unsupported provider/format combinations return clear diagnostics.
- The generation preview displays the uploaded file type dynamically instead of hardcoding PDF.

### Evidence

- [x] Manually tested locally

Local verification:

```text
pnpm vitest run tests/document/extractor-registry.test.ts tests/document/pdf-compat.test.ts tests/document/text-extractor.test.ts
Test Files  3 passed (3)
Tests       13 passed (13)

pnpm exec tsc --noEmit
passed

pnpm check:i18n-keys
passed

pnpm check
passed

pnpm lint
0 errors, existing warnings only
```

## Checklist


wyuc (commented):
Thanks for this — the extractor-boundary design (registry + capability matching + `ParsedPdfContent` compat shape) is clean, and the PDF/TXT/Markdown paths look solid. One blocking issue on the DOCX/PPTX path, plus a few minor items.

## Blocking

### DOCX/PPTX returns an opaque 500 instead of clear diagnostics

The headline new formats break (or fail unhelpfully) on the two most common setups, which contradicts the "clear diagnostics for unsupported provider/format combinations" goal:

- **No MinerU configured (stock deployment).** For a DOCX/PPTX upload the client sends no `providerId`/`baseUrl` (`app/generation-preview/page.tsx`, the `documentMimeType === 'application/pdf'` guards), so the route capability-matches `mineru` — the only office-capable provider. `mineru` is unmanaged → `resolvePDFBaseUrl('mineru', undefined)` returns `undefined` → `parseWithMinerUDocument` throws "MinerU base URL is required…", which the outer try/catch maps to `apiError('PARSE_FAILED', 500, …)`. Users get a 500 with a raw internal message, not an actionable 4xx.

- **MinerU configured client-side (not server-managed).** The same PDF-only guards in `generation-preview` drop the user's configured `baseUrl`/`apiKey` for non-PDF files, so even a correctly-configured client cannot upload DOCX/PPTX — the request reaches the route without a base URL and fails identically.

Root cause is twofold: (1) `generation-preview` only forwards provider config when the file is a PDF, and (2) `selectDocumentExtractorProvider` matches on static capabilities only, with no check that the matched provider is actually configured/runnable.

Suggested fix:
- Forward `providerId`/`apiKey`/`baseUrl` for the document path regardless of MIME type (let the route decide), so client-configured MinerU works for office formats.
- After provider selection, if the chosen provider is MinerU-backed and has neither a server config (`isServerConfiguredProvider('pdf', id)`) nor a client base URL, return a `400`/`422` with an actionable message (e.g. "DOCX/PPTX requires a configured MinerU extractor") rather than falling through to a 500. Ideally capability matching also factors in configured-availability.

## Minor

- **`getGenerationStepText` produces wrong labels** (`app/generation-preview/types.ts`). The type label is derived from `pdfFileName.split('.').pop().toUpperCase()` with no validation, so `report.v2` renders as "Analyzing V2 file" and an extension-less name renders the whole filename. Prefer the MIME-based branch, or only accept the extension when it's one of the known set.

- **Text charset is effectively ignored** (`lib/document/extractors/text.ts` + `lib/document/mime.ts`). `normalizeDocumentMimeType` strips `;charset=…` before the extractor runs, so `textDecoderForMimeType` always falls back to UTF-8 and a non-UTF-8 TXT/MD upload is silently mojibaked. Either preserve the charset for text types, sniff a BOM, or document UTF-8-only.

- **Dead i18n keys.** `getGenerationStepText` always overrides the `pdf-analysis` step, so `generation.analyzingPdf` / `analyzingPdfDesc` are now unreachable in every locale. Remove or repoint.

## Tests

Registry and text-extractor tests are good. Gaps worth closing:
- No test for the unconfigured-MinerU path — the registry test asserts `mineru` is *selected* for office formats but nothing asserts the route/extract behavior when it can't run (where the 500 hides).
- No test for `extract-document` route error mapping (unsupported MIME → 400, unknown provider → 400, provider/format mismatch → 400).

## Verdict

Mergeable once the DOCX/PPTX path returns an actionable 4xx (and forwards client MinerU config for office formats). A route-level error-mapping test would lock it in. The rest is low-severity polish.

wyuc (commented):
Thanks for the quick turnaround — the 422 path, client-config forwarding, label whitelist, BOM sniffing, dead-key cleanup, and the new route error-mapping test all look good. Two regressions slipped in with the fix, though:

## Blocking (new regressions)

### 1. Non-PDF uploads from the main UI now get rejected with 400

Removing the `documentMimeType === 'application/pdf'` guard means `generation-preview` now *always* forwards `providerId`. The homepage stores `settings.pdfProviderId` (default `unpdf`) for every upload, so a TXT/Markdown/DOCX/PPTX upload arrives with `providerId=unpdf`. In `app/api/extract-document/route.ts` the `!provider.supportedMimeTypes.includes(mimeType)` check then 400s the request ("Document extractor 'unpdf' does not support MIME type 'text/plain'") *before* `selectDocumentExtractorProvider` can pick `plain-text`/`mineru`.

Net effect: the new course-material formats fail on the default path unless the user manually switches the extractor dropdown to a compatible one — which defeats the feature. (TXT/MD should "just work" locally; they no longer do.)

Fix: treat `preferredProviderId` as a *hint*, not a hard constraint — if it doesn't support the file's MIME type, fall back to MIME/capability-based selection instead of returning 400. Alternatively, only forward `providerId` from the client when it actually applies to the file (PDF, or a provider that supports the MIME). Note you still want to forward `apiKey`/`baseUrl` unconditionally so client-configured MinerU keeps working for Office formats (that was the original fix).

### 2. The new 422 guard wrongly blocks `mineru-cloud`

`isMinerUBackedProvider` returns true for both `mineru` and `mineru-cloud`, but `mineru-cloud` resolves its endpoint as `config.baseUrl || MINERU_CLOUD_DEFAULT_BASE` (`lib/pdf/mineru-cloud.ts`) — it needs only an API key, no base URL. So a user who selects `mineru-cloud` for a PDF (key set, no base URL — the normal case) now hits the 422 "requires a configured MinerU … base URL", a path that worked before.

Fix: narrow the guard to `provider.id === 'mineru'` (self-hosted, which genuinely requires a base URL), or check the effective endpoint after resolution (including the cloud default) rather than the raw client base URL.

## Otherwise

Round-1 items are all addressed well. A test covering (a) a non-PDF upload with a default `unpdf` providerId and (b) `mineru-cloud` PDF with key-only would lock both regressions down. Close these two and it's good to go.

yanpgwang:
Re-reviewed the latest push — both round-2 regressions are fixed and now locked by tests:

1. **Incompatible `providerId` treated as a hint** ✅ — the route drops a non-matching preferred provider and falls back to `selectDocumentExtractorProvider`, so a default `unpdf` + TXT/MD upload resolves to `plain-text` instead of 400. Covered by *"treats an incompatible preferred provider as a hint and falls back by MIME type"*.
2. **422 guard no longer blocks `mineru-cloud`** ✅ — narrowed to `provider.id === 'mineru'`, so a key-only `mineru-cloud` PDF passes. Covered by *"allows MinerU Cloud PDF extraction with an API key and no base URL"*.

Local checks: `vitest run tests/document` → **19/19 passed**; `tsc --noEmit` clean; `prettier --check` clean on the changed files; new i18n keys present in all 8 locales.

## One thing I'd suggest fixing in this PR (not as follow-up)

**Office support is added for self-hosted MinerU but not for MinerU Cloud — this PR creates a self-host/cloud capability split.** Today `mineru-cloud` is left PDF-only (`lib/document/extractors/pdf.ts:41` only grants the Office MIME types to `id === 'mineru'`). So a user configured with *only* MinerU Cloud (key, no base URL) who uploads a DOCX gets capability-matched to self-hosted `mineru` → the new 422 → a message telling them to configure a MinerU **server / base URL**, which they don't need and don't have. The headline "multi-format upload" feature silently doesn't apply to the Cloud path.

I'd rather we close this in the PR than ship the inconsistency and fix it later — adding Office to one backend but not the other is the kind of gap that's easy to land now and confusing to debug afterwards.

I verified the Cloud API genuinely supports Office, so this is a real, achievable fix (not an API limitation). Live call to `mineru.net/api/v4/file-urls/batch` with `.docx/.pptx/.doc/.ppt` all returned `code:0` with OSS upload URLs that **preserve the extension** — and that extension is how MinerU infers the document type. Concretely:

- `lib/document/extractors/pdf.ts:41` — grant the Office MIME types to `mineru-cloud` as well.
- `lib/pdf/mineru-cloud.ts` `sanitizeFileName` — relax the `.pdf`-only allowlist to also accept `.docx/.pptx`; and thread the **real** filename/extension through to the upload instead of forcing `document.pdf`. This is the critical part: if the Cloud receives a `.pdf`-named DOCX it will try to parse it as a PDF and fail.
- route the `mineru-cloud` non-PDF path to the Cloud parser (the current `extract()` sends every non-PDF to the self-hosted `parseWithMinerUDocument`).

If you'd prefer to keep this PR scoped to self-hosted MinerU, that's a reasonable call too — but then the UI/422 copy should explicitly say Office requires self-hosted MinerU (Cloud = PDF only), so Cloud users aren't sent to configure a base URL they don't need.

## Minor (non-blocking)

The charset branch in `text.ts` is dead: the route normalizes the MIME type (stripping `;charset=…`) before calling `extract`, so `textDecoderForMimeType`'s `mimeType.match(/charset=/)` (`lib/document/extractors/text.ts:15`) never matches. BOM sniffing is the only effective handling, so a `charset=gbk` file with no BOM still falls back to UTF-8 and mojibakes. Either drop the dead regex (and document UTF-8 + BOM only) or preserve the charset through to the extractor.

## Verdict

Round-2 regressions are resolved and tested. The self-host/cloud Office split is the one thing I'd like addressed before merge (or documented away if Cloud is intentionally out of scope); the charset note is low-severity polish.

yanpgwang:
I re-tested the latest PR head (`d31014c`) and the MinerU Cloud DOCX happy path works with a valid Bearer token, but there are still a few issues to address before merge:

1. Server-managed MinerU Cloud is not detected with API key only. `loadEnvSection(..., { requiresBaseUrl: true })` applies to all PDF providers, so `PDF_MINERU_CLOUD_API_KEY` alone is ignored and `resolvePDFApiKey('mineru-cloud')` returns empty unless `PDF_MINERU_CLOUD_BASE_URL` is also set. Since MinerU Cloud has a default base URL, this should allow API-key-only config for `mineru-cloud`.

2. Office files can be rejected when the browser reports a generic ZIP MIME type. `normalizeDocumentMimeType()` currently trusts any non-empty/non-octet MIME before checking the extension, so `.docx`/`.pptx` files reported as `application/zip` or `application/x-zip-compressed` become unsupported instead of normalizing by extension.

3. Real MinerU Cloud results still report `metadata.parser: "mineru"`, because the cloud ZIP parser reuses `extractMinerUResult()`, which hardcodes `parser: "mineru"`. The route tests mock `"mineru-cloud"`, so they do not catch the real metadata mismatch. Please override the parser metadata in the cloud wrapper or make the shared parser accept a parser id.

yanpgwang (commented):
Re-reviewed the two follow-up commits (`d31014c` + `a214ca1f`). All three round-3 items are addressed and covered by tests:

1. **`mineru-cloud` server config with API key only** ✅ — `baseUrlOptionalProviders: new Set(['mineru-cloud'])` in `lib/server/provider-config.ts`, so `PDF_MINERU_CLOUD_API_KEY` alone activates the provider (default endpoint kicks in). Locked by two new `provider-config.test.ts` cases (env-only and YAML-only).
2. **Generic ZIP MIME types no longer reject Office files** ✅ — `normalizeDocumentMimeType` now falls back to the extension when the browser reports `application/zip` / `application/x-zip` / `application/x-zip-compressed` / `application/octet-stream`. Covered by `mime.test.ts`.
3. **Cloud results now report the correct parser id** ✅ — `parseMinerUZip` overrides `metadata.parser` to `'mineru-cloud'` after `extractMinerUResult`. Asserted in `mineru-cloud.test.ts`.

Also verified end-to-end against the real Cloud API with a Bearer-authenticated `.docx` upload — batch → presigned PUT → poll → ZIP download all pass; `full.md` returns and `sanitizeFileName` preserves the `.docx` extension so `document_type` inference is correct.

Local checks: `vitest run tests/document tests/server/provider-config.test.ts` → **70/70 passed** on `a214ca1f`.

## Non-blocking follow-ups

- **`pdf-compat.ts:63` block type is now inconsistent for Cloud output.** With the round-3 parser-id fix, `parsed.metadata?.parser` is `'mineru-cloud'` for the Cloud path, so the check `parser === 'mineru' ? 'markdown' : 'text'` tags Cloud markdown as `type: 'text'`. Round-trip via `documentArtifactToParsedPdfContent` still merges both types so `ParsedPdfContent.text` is unchanged today, but any future direct consumer of `DocumentArtifact.blocks` will see self-host and cloud diverge. One-line fix: `parser === 'mineru' || parser === 'mineru-cloud'`. Latent, not urgent.
- **`sanitizeFileName` allowlist vs `MINERU_DOCUMENT_MIME_TYPES` mismatch.** The Cloud sanitizer accepts `.doc/.ppt` (legacy Office), but the extractor's `MINERU_DOCUMENT_MIME_TYPES` only lists `pdf/docx/pptx`, and there's no MIME entry for `application/msword` / `application/vnd.ms-powerpoint`. So the legacy formats can never reach the extractor — the allowlist expansion is dead. Either drop `.doc/.ppt` from the sanitizer, or add the legacy MIME types + accept-list entries end-to-end. Cosmetic today.

## Verdict

Round-3 items resolved and tested; real Cloud DOCX pipeline confirmed working. LGTM to merge — the two notes above can be follow-ups.

yanpgwang:
@wyuc round-3 items are resolved and locked with tests, and the two non-blocking follow-ups I flagged (pdf-compat markdown block for `mineru-cloud`, sanitizer allowlist vs `MINERU_DOCUMENT_MIME_TYPES`) got cleaned up in 539f3dd with tests too — LGTM from my side, could you give a final approve when you have a minute?

@jackefn PR is currently `BEHIND` main — please rebase (or merge main in) so it's up to date before the final merge.

yanpgwang (approved):

PR #743, opened 2026-06-15
feat(dsl): promote Stage/Scene lesson skeleton into @maic/dsl (#740)

### LINKED ISSUE #740: [Phase 2] Promote Stage/Scene lesson types into @maic/dsl — contract-scope decision needed
Phase 2 of the `@maic` SDK consumption epic (#720): promote the lesson-structure types (`Stage` / `Scene` / `SceneContent`) into `@maic/dsl`. The README roadmap framed this as "split the pure structural types from the app-coupled ones first" — but on inspection the coupling is shallower than expected, which turns this into a **contract-scope decision** that I'd like a steer on before writing the migration.

## Coupling reality

`lib/types/stage.ts` imports from three sibling modules — `lib/types/action.ts`, `lib/types/widgets.ts`, `lib/pbl/types.ts` — but **all three are themselves fully dependency-free** (zero imports; `Action`'s whiteboard-draw payloads are inlined, not referencing `@maic/dsl` element types). So there is no hard coupling blocker: technically the entire `Stage` / `Scene` / `SceneContent` / `Action` / `WidgetConfig` / `PBLProjectConfig` surface could move into `@maic/dsl` as pure types.

The real question is therefore **what belongs in the contract package**, not whether it compiles.

Breakdown:
| Type | Depends on | Notes |
|---|---|---|
| `SceneType`, `StageMode`, `VideoManifest`, `Whiteboard` | `Slide` only | universal, trivially DSL-able |
| `Stage` | `Whiteboard`, `VideoManifest` | universal lesson container |
| `Scene` | `Action`, `SceneContent`, `Slide` | central; couples to `Action` |
| `SlideContent`, `QuizContent` | `Slide` | universal |
| `InteractiveContent` | `WidgetType/WidgetConfig/TeacherAction` | richer product feature (Ultra-mode widgets) |
| `PBLContent` | `PBLProjectConfig` | richer product feature (project-based learning) |
| `Action` (20+ variants) | — | playback instruction set |

## Options

**A. Minimal — defer.** Leave `Stage`/`Scene` in the app for now; only the slide object-model stays in `@maic/dsl`. Lowest risk, but the roadmap item stays open and the runtime engine contract stays app-side.

**B. Lesson container in DSL, features plug in (recommended).** Move `Stage`, `Scene`, `SceneType`, `StageMode`, `Whiteboard`, `VideoManifest`, `SlideContent`, `QuizContent` into `@maic/dsl`. Make `Scene` generic / open over its `actions` and non-slide `content` so the universal structure lives in the contract while `Action`, widget configs, and PBL stay app-side (or move to sibling packages later). Keeps the DSL lean and focused on the lesson skeleton; needs a small contract-shape change (generic `Scene<TAction, TContent>` or an extension seam).

**C. Move everything.** `Stage`/`Scene`/`Action`/widgets/PBL/quiz all into `@maic/dsl`. Most complete, single source of truth for the whole object model — but pulls product-feature surfaces (Ultra widgets, PBL) into the "pure contract" package, which may be more than the contract should own.

## Recommendation

I lean **B**: the universal lesson skeleton (`Stage`/`Scene`/`Slide`/quiz) is the contract; `Action` and the feature configs are richer/faster-moving and can plug in via generics now and graduate to sibling packages (`@maic/actions`, …) later if desired. But the generic-`Scene` shape is a real API choice, and whether widgets/PBL "belong" in the contract is a product call.

@cosarah — which scope do you want before I implement (and if **B**, do you prefer a generic `Scene<TAction, TContent>` or a registry/extension seam)? Happy to take whichever direction; I'll do the migration + e2e + tests behind it.

Part of #720.

wyuc:
Resolves #740 (Phase 2 of #720). Implements **scope B** from the issue: the universal lesson skeleton moves into `@maic/dsl`, with `Scene` generic so app-side feature surfaces (`Action`, Ultra-mode widgets, PBL) plug in via type parameters.

## Decision

Per the open question in #740, this goes with a **generic `Scene<TAction, TContent>`** rather than a registry / declaration-merge seam. Reasons: explicit at the call site, `isolatedModules`-friendly (no ambient augmentation), straightforward `.d.ts` emission, and type inference stays readable. A registry seam can still be layered on later if desired — this does not foreclose it.

## What moves into `@maic/dsl`

New `packages/@maic/dsl/src/stage.ts`:
- `Stage`, `SceneType`, `StageMode`, `Whiteboard`, `VideoManifest` / `VideoManifestEntry`
- `GeneratedAgentConfig`, `MultiAgentConfig` (previously inline/anonymous — now named)
- `SlideContent`, `QuizContent`, `QuizOption`, `QuizQuestion`
- `SceneContent` — narrowed to the **universal** kinds only: `SlideContent | QuizContent`
- `Scene<TAction = never, TContent extends { type: SceneType } = SlideContent | QuizContent>` — the contract owns the skeleton + universal content; `TContent` is constrained structurally (tagged with `type: SceneType`) so an app can pass a wider union
- Pure guards `isSlideContent` / `isQuizContent`

Defaults (`never` / `SlideContent | QuizContent`) give renderers and importers a feature-free, read-only scene — exactly what they need.

## What stays app-side

- `Action` (and all of `lib/types/action.ts`)
- `WidgetType` / `WidgetConfig` / `TeacherAction` / `InteractiveContent`
- `PBLContent` / `PBLProjectConfig`

These are faster-moving product surfaces; they may graduate to sibling packages (`@maic/actions`, …) later, but not in this PR. The generation re-export block at the bottom of the old `stage.ts` is dropped — verified zero consumers import those five types via `@/lib/types/stage`.

## App wiring

`lib/types/stage.ts` is rewritten as a **shim** (not deleted) so the ~55 existing `import { … } from "@/lib/types/stage"` call sites keep working unchanged:
- re-exports the skeleton types from `@maic/dsl`
- keeps `InteractiveContent` / `PBLContent` locally
- declares the app `SceneContent` as the full four-way union (so `switch (content.type)` call sites keep all four cases)
- aliases `Scene = DslScene<Action, SceneContent>` so existing `Scene` usage keeps its exact semantics (`actions: Action[]`, four-way content)

`Scene` changing from `interface` to a generic `type` alias on the DSL side is safe — confirmed no `interface extends Scene`, no declaration merging, no `implements Scene` anywhere in the tree.

## Tests

- New `packages/@maic/dsl/test/stage.test.ts` (9 contract tests) + `vitest.config.ts` + `test` script on the package — covers the generic defaults, the app-instantiation pattern, guard behavior, the `Whiteboard` `Omit` shape, and the four-way `SceneType`.
- DSL package: `typecheck` clean, 9/9 tests pass.
- Whole app: `tsc --noEmit` clean; `vitest run` → **805 pass / 1 fail**. The single failure (`tests/edit/stage-mode.test.ts > sceneEditorRegistry > registering a different surface … warns in dev`) is **pre-existing on `origin/main`** — verified by stashing this branch's changes, rebuilding, and re-running; it fails identically and is unrelated to types.

## Out of scope (deliberately)

- `Action` / widgets / PBL migration into the DSL or sibling packages — later Phase 2 work
- JSON Schema + pure validator for the stage contract — separate roadmap item
- npm publish — Phase 1 maintainer action

### DIFF
--- a/lib/types/stage.ts
+++ b/lib/types/stage.ts
@@ -10,10 +10,7 @@
 // `Scene` is re-exported as an alias of the app's fully-instantiated
 // `Scene<Action, AppSceneContent>`, so existing `import { Scene }` callers keep
 // the same semantics (actions are `Action[]`, content spans all four kinds).
-import type {
-  Scene as DslScene,
-  SceneContent as DslSceneContent,
-} from '@maic/dsl';
+import type { Scene as DslScene, SceneContent as DslSceneContent } from '@maic/dsl';
 import type { Action } from '@/lib/types/action';
 import type { WidgetType, WidgetConfig, TeacherAction } from '@/lib/types/widgets';
 import type { PBLProjectConfig } from '@/lib/pbl/types';
--- a/lib/types/stage.ts
+++ b/lib/types/stage.ts
@@ -28,9 +28,12 @@ export type {
   QuizOption,
   QuizQuestion,
   QuizContent,
-  isSlideContent,
-  isQuizContent,
 } from '@maic/dsl';
+
+// The two discriminant guards are runtime functions, so they must be value
+// re-exported — a bare `export type {}` erases them and leaves the import as
+// `undefined` at runtime / "cannot be used as a value" at the type level.
+export { isSlideContent, isQuizContent } from '@maic/dsl';
 
 // The contract's `SceneContent` is the universal subset (slide | quiz). Reach it
 // under a distinct name; the app's own `SceneContent` (declared below) is the
--- a/lib/types/stage.ts
+++ b/lib/types/stage.ts
@@ -1,150 +1,55 @@
-// Stage and Scene data types
-import type { Slide } from '@maic/dsl';
+// Stage and Scene data types.
+//
+// The universal lesson skeleton (Stage / Scene / SceneContent / Whiteboard /
+// VideoManifest / SlideContent / QuizContent / …) now lives in `@maic/dsl` and
+// is re-exported below. `Scene` is generic there: the contract owns only the
+// structure + the slide/quiz content kinds, while the playback `Action` set and
+// the richer feature content (interactive widgets, PBL) are app-side and get
+// composed in here.
+//
+// `Scene` is re-exported as an alias of the app's fully-instantiated
+// `Scene<Action, AppSceneContent>`, so existing `import { Scene }` callers keep
+// the same semantics (actions are `Action[]`, content spans all four kinds).
+import type {
+  Scene as DslScene,
+  SceneContent as DslSceneContent,
+} from '@maic/dsl';
 import type { Action } from '@/lib/types/action';
-import type { PBLProjectConfig } from '@/lib/pbl/types';
 import type { WidgetType, WidgetConfig, TeacherAction } from '@/lib/types/widgets';
+import type { PBLProjectConfig } from '@/lib/pbl/types';
 
-export type SceneType = 'slide' | 'quiz' | 'interactive' | 'pbl';
+export type {
+  SceneType,
+  StageMode,
+  Whiteboard,
+  VideoManifestEntry,
+  VideoManifest,
+  GeneratedAgentConfig,
+  MultiAgentConfig,
+  Stage,
+  SlideContent,
+  QuizOption,
+  QuizQuestion,
+  QuizContent,
+  isSlideContent,
+  isQuizContent,
+} from '@maic/dsl';
 
-export type StageMode = 'autonomous' | 'playback' | 'edit';
+// The contract's `SceneContent` is the universal subset (slide | quiz). Reach it
+// under a distinct name; the app's own `SceneContent` (declared below) is the
+// full four-way union so existing `switch (content.type)` call sites keep all
+// four cases.
+export type { SceneContent as SceneContentBase } from '@maic/dsl';
 
-export type Whiteboard = Omit<Slide, 'theme' | 'turningMode' | 'sectionTag' | 'type'>;
-
-export interface VideoManifestEntry {
-  type: 'video';
-  prompt: string;
-  aspectRatio?: string;
-}
-
-export type VideoManifest = Record<string, VideoManifestEntry>;
-
-/**
- * Stage - Represents the entire classroom/course
- */
-export interface Stage {
-  id: string;
-  name: string;
-  description?: string;
-  createdAt: number;
-  updatedAt: number;
-  // Stage metadata
-  languageDirective?: string;
-  style?: string;
-  // Whiteboard data
-  whiteboard?: Whiteboard[];
-  // Generated video requests keyed by the mediaRef used by PPTVideoElement.
-  // Runtime media state lives in the media task store / persisted media files.
-  videoManifest?: VideoManifest;
-  // Agent IDs selected when this classroom was created
-  agentIds?: string[];
-  /**
-   * Server-generated agent configurations.
-   * Embedded in persisted classroom JSON so clients can hydrate
-   * the agent registry without relying on IndexedDB pre-population.
-   * Only present for API-generated classrooms.
-   */
-  generatedAgentConfigs?: Array<{
-    id: string;
-    name: string;
-    role: string;
-    persona: string;
-    avatar: string;
-    color: string;
-    priority: number;
-  }>;
-  /**
-   * True when this classroom was generated with Interactive Mode enabled
-   * (the INTERACTIVE_OUTLINES prompt branch).
-   * Absent on legacy classrooms, imports, and regular-mode generations.
-   */
-  interactiveMode?: boolean;
-  /**
-   * True when this classroom was generated with the vocational Task Engine
-   * path enabled. This is distinct from `interactiveMode`: task-engine
-   * classrooms are interactive, but not every interactive classroom is
-   * vocational.
-   */
-  taskEngineMode?: boolean;
-}
+// The raw, generic contract Scene is reachable under a distinct name for
+// callers (e.g. read-only renderers) that want the feature-free skeleton.
+export type { Scene as SceneShape } from '@maic/dsl';
 
 /**
- * Scene - Represents a single page/scene in the course
- */
-export interface Scene {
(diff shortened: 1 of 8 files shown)


wyuc (commented):
Reviewed the diff — overall solid. The shim approach cleanly preserves the existing `Scene` semantics, and the claims in the PR body check out: no `interface extends Scene` / declaration merging in play, no remaining consumers of the dropped generation re-exports, and the four-case `scene-renderer` still compiles against the widened content union. Three points, none blocking:

**1. (should-fix) Type-guard constraints are narrower than the generic API they're meant to support** — `packages/@maic/dsl/src/stage.ts`

`isSlideContent` / `isQuizContent` are typed `<T extends SceneContent>`, where `SceneContent` is the narrow `SlideContent | QuizContent`. But the docstrings say "or any candidate", and the whole point of generic `Scene<TAction, TContent>` is to let apps widen `TContent` (e.g. `+ InteractiveContent | PBLContent`). A caller doing `isSlideContent(scene.content)` on a widened union is rejected by TS, since the wide union doesn't `extend` the narrow one. Today the guards aren't called on widened content, so it's latent — but it makes them unusable for exactly the generic case they exist to serve. Suggest `<T extends { type: SceneType }>` and narrow from there.

**2. (should-fix) `@maic/dsl` `test` script isn't standalone on a clean checkout** — `packages/@maic/dsl/package.json` / `vitest.config.ts`

The tests self-import from `@maic/dsl`, which resolves through the package `exports` map to `./dist/index.js`. `dist` isn't committed, `vitest.config.ts` has no `resolve.alias` to `src`, and `test` (`vitest run`) doesn't build first — so `pnpm --filter @maic/dsl test` fails on a clean checkout unless the package happens to have been built out-of-band. Suggest one of: a `resolve.alias` mapping `@maic/dsl` → `./src/index.ts` in the vitest config, a `pretest: build`, or importing from the relative source in the tests.

**3. (nit) A type test asserts nothing** — `packages/@maic/dsl/test/stage.test.ts`

The "TAction defaults to never … rejects concrete actions" test uses `expectTypeOf<DefaultScene['actions']>().toMatchTypeOf<unknown[] | undefined>()`, which is trivially true for `never[] | undefined` and doesn't actually test rejection. `.toEqualTypeOf<never[] | undefined>()` would pin the intended behavior. (Minor: `toMatchTypeOf` is deprecated in vitest 4 in favor of `toExtend`.)

wyuc:
Thanks for the review — all three addressed in `ce8a5dc`:

1. **Guard genericity**: relaxed `<T extends SceneContent>` → `<T extends { type: SceneType }>` on both `isSlideContent` / `isQuizContent`, so they accept an app-widened content union (the generic case `Scene<TAction, TContent>` exists to serve). Added a regression test that runs the guards against `SceneContent | InteractiveContent`.

2. **Standalone tests**: added `resolve.alias` (`@maic/dsl` → `./src/index.ts`) to `vitest.config.ts`, so `pnpm --filter @maic/dsl test` works on a clean checkout with no `dist` build. Verified by `rm -rf dist` then re-running — 10/10 pass.

3. **Type-test pin**: switched to `.toEqualTypeOf<never[] | undefined>` so the default actually rejects drift.

`pnpm --filter @maic/dsl typecheck` clean, `pnpm --filter @maic/dsl test` 10/10, whole-app `tsc --noEmit` exit 0.

wyuc (commented):
Re-reviewed `ce8a5dc` — all three are properly addressed:

1. **Guard genericity** ✅ — `<T extends { type: SceneType }>` compiles for an app-widened union, the `content is T & SlideContent` predicate still narrows correctly, and it's not over-loose (`{ foo: 1 }` and `{ type: 'bogus' }` are both rejected). Confirmed with a standalone `tsc --strict` pass.
2. **Standalone tests** ✅ — `resolve.alias` resolves `@maic/dsl` to `./src/index.ts`, so the suite runs with no `dist` build.
3. **Type-test pin** ✅ — `.toEqualTypeOf<never[] | undefined>` is exact (verified bidirectionally; a concrete-action assignment is now rejected).

One small thing the re-review surfaced (pre-existing in the shim, not from this fix commit — flagging since it's part of the PR):

**(nit / latent) Guard functions are re-exported via `export type` in the compat shim** — `lib/types/stage.ts:21-35`

`isSlideContent` / `isQuizContent` are real runtime functions, but they sit inside the shim's `export type { … }` block, which erases them. So `import { isSlideContent } from "@/lib/types/stage"` would error ("cannot be used as a value") / be `undefined` at runtime. No call sites import the guards through the shim today, so it's latent — but since the shim's purpose is back-compat re-export, it'd be worth pulling the two functions into a value export:

```ts
export { isSlideContent, isQuizContent } from "@maic/dsl";
```

while keeping the type-only block for the interfaces.

(Also informational, non-blocking: with the new alias the suite resolves `@maic/dsl` to `src`, so it no longer exercises the published `dist` / `exports` resolution path — fine, just means a future exports-map/build regression wouldn't be caught by these unit tests.)

Otherwise LGTM — the shim preserves the existing `Scene` semantics and the contract split looks clean.

wyuc:
Good catch — fixed in `df0e201`.

**Guard value-export**: pulled `isSlideContent` / `isQuizContent` out of the shim's `export type { … }` block into a plain `export { isSlideContent, isQuizContent } from "@maic/dsl"`. `export type` was erasing the runtime binding, so a future `import { isSlideContent } from "@/lib/types/stage"` would have been `undefined` at runtime / a TS "cannot be used as a value" error. Latent today (no shim-level guard imports), but the shim is there for back-compat, so worth getting right.

`prettier --check` clean, whole-app `tsc --noEmit` exit 0.

On the informational note (tests resolving `@maic/dsl` → `src` via `resolve.alias`): acknowledged and intentional trade-off — kept the alias so `pnpm --filter @maic/dsl test` is standalone on a clean checkout (the standalone-ness was the other review ask). The cost is that a future `exports`-map / `dist` build regression wouldn't be caught by these unit tests. Happy to add a `pretest: build` (so the suite also exercises a fresh `dist`) if you'd prefer belt-and-suspenders — say the word and I'll add it; otherwise leaving as-is per your "fine".

cosarah (approved):
LGTM, verified locally

PR #744, opened 2026-06-15
feat: add LM Studio as a built-in provider

### LINKED ISSUE #730: [Feature]: Ability to select model in app
### Problem or Motivation

When using LMstudio you must specify the model in the env file.  

### Proposed Solution

It would be good if the app displayed the available models and allowed you select in the app.

### Alternatives Considered

_No response_

### Area

API / Backend

### Additional Context

_No response_

athul0rameshan9:
## Summary

Registers **LM Studio** as a built-in, OpenAI-compatible provider — keyless, pointed at its default local server (`http://localhost:1234/v1`), mirroring the existing Ollama and Lemonade built-ins.

Closes #730

## Changes (4 files, +18/−2)

- **`lib/ai/providers.ts`** — add `lmstudio` to `PROVIDERS` (`type: 'openai'`, `requiresApiKey: false`, `defaultBaseUrl: 'http://localhost:1234/v1'`, `models: []`)
- **`lib/server/provider-config.ts`** — add `LMSTUDIO` to `LLM_ENV_MAP`, mark `lmstudio` keyless
- **`lib/types/provider.ts`** — add `'lmstudio'` to `BuiltInProviderId`
- **`.env.example`** — document `LMSTUDIO_BASE_URL` / `LMSTUDIO_MODELS`

## Model discovery

Because `lmstudio` is `type: 'openai'`, model discovery works through the existing generic flow (`app/api/provider/probe-models` + `lib/server/model-fetch.ts`) via the **Fetch Models** button — no provider-specific route or UI needed.

## History

Slimmed down and rebased on latest `main`. The original version's custom `/api/list-models` route, settings-UI changes, i18n keys, and Azure fetch-wrapper have been removed, since `main` now provides generic model discovery and native Azure handling (`lib/ai/azure.ts`). Removing the custom route also resolves the outstanding review blocker (the managed provider's `effectiveBaseUrl` reaching the client) by eliminating that code path.


wyuc (changes requested):
Thanks for this, and welcome — a clean, well-structured first contribution (nice that you covered every locale). A few things to address before merge; two are security-related.

**🔴 Server API key can leak to a client-supplied host** — `app/api/list-models/route.ts`
For a server-managed provider configured with only an API key (e.g. `OPENAI_API_KEY` set, no `OPENAI_BASE_URL`), `resolveBaseUrl()` returns `undefined`, so `resolveBaseUrl(...) || clientBaseUrl` falls back to the **client-supplied** `baseUrl` while `resolveApiKey()` still returns the server secret. The request then sends `Authorization: Bearer <server key>` to an arbitrary client-controlled host. For managed providers, please ignore `clientBaseUrl` entirely and use only the server/default URL when a server key is attached.

**🔴 Missing SSRF guard** — `app/api/list-models/route.ts`
The other routes that fetch a client-supplied URL (`app/api/generate/voice`, `azure-voices`, `generate/video`, `proxy-media`, `parse-pdf`) all call `validateUrlForSSRF()` from `@/lib/server/ssrf-guard` before fetching. This new route fetches after only `new URL()` parsing, so a crafted `baseUrl` can make the server probe internal addresses, and the upstream error body is returned to the caller. Please run the same `validateUrlForSSRF()` before building/fetching `modelsUrl`. This matters most for hosted / multi-tenant deployments.

**🟡 Success-response shape mismatch breaks the feature** — `components/settings/provider-config-panel.tsx`
`apiSuccess({ models })` serializes as `{ success: true, models: [...] }` (the helper spreads, it does not nest under `data`). The client reads `data.data?.models`, which is always `undefined`, so even a successful fetch shows the failure message. Read `data.models` instead.

Once the two security items are addressed the rest looks good. Thanks again for the contribution!

athul0rameshan9:
Thanks for the review! All three items have been addressed:

- 🔴 **API key leak**: Managed providers now ignore all client-supplied credentials (`isServerConfiguredProvider` check, matching the pattern in tts/voice/video routes)
- 🔴 **SSRF guard**: Added `validateUrlForSSRF()` on client-supplied base URLs before fetching
- 🟡 **Response shape**: Client now reads `data.models` instead of `data.data?.models`

Also rebased onto latest main. Ready for re-review!

wyuc:
## Review: `/api/list-models` + LM Studio provider

Thanks for the PR — the model auto-discovery is a nice addition, and the SSRF guard + managed-provider key/baseUrl isolation are good to see. One **blocking** issue and a couple of follow-ups before merge.

### 🔴 P1 (blocking) — Redirect-based SSRF bypass · `app/api/list-models/route.ts`

The outbound `fetch(modelsUrl, …)` uses the default `redirect: 'follow'`, but `validateUrlForSSRF()` only validates the **initial** user-supplied `baseUrl`. An attacker can supply a public `baseUrl` that returns a `3xx` redirect to an internal target (e.g. cloud metadata `http://169.254.169.254/…`, `localhost`, or an internal service); `fetch` transparently follows it and the redirect target is never re-validated — so the SSRF guard is bypassed and the upstream body is returned to the caller.

This is reachable **unauthenticated** on deployments where `ACCESS_CODE` is unset (the default for the public demo and typical self-host).

The repo already has the right pattern for this — please mirror it:
- `app/api/proxy-media/route.ts` uses `redirect: 'manual'` and re-validates **every** `Location` hop with `validateUrlForSSRF` (see #398).
- `app/api/azure-voices/route.ts` and `app/api/verify-pdf-provider/route.ts` use `redirect: 'manual'` and reject `3xx`.
- The `REDIRECT_NOT_ALLOWED` / `TOO_MANY_REDIRECTS` error codes already exist in `lib/server/api-response.ts`.

`/api/list-models` is currently the only user-URL-fetching route missing this protection.

### 🟡 P2 — `defaultBaseUrl` fallback · `app/api/list-models/route.ts` + `components/settings/index.tsx`

Two issues on the same fallback path:
- **Security:** only `safeClientBaseUrl` goes through `validateUrlForSSRF`. The `|| provider?.defaultBaseUrl` branch is fetched without any SSRF check. It's a hardcoded constant today (so low risk), but the guard should run against the **resolved** `baseUrl`, not just the raw client input. (The middle `|| safeClientBaseUrl` term is also dead code — `resolveBaseUrl` already returns it for unmanaged providers.)
- **Correctness:** for a fresh LM Studio setup with the Base URL field left blank, the fetch succeeds via `defaultBaseUrl` but only the **models** are persisted, not the base URL. Since keyless providers are considered configured only when `baseUrl` is set, the provider still won't appear in the active model list until the user manually types the URL. Either persist the effective default base URL, or require it before fetching.

### 🟡 P2 — Upstream error body echoed verbatim · `app/api/list-models/route.ts`

On a non-OK upstream response the route returns `Provider returned ${status}: ${errorText}` (the full raw upstream body) to the client. Combined with the P1 this is the exfiltration channel, but it's also an independent info leak. Other routes (`verify-model`, `verify-image-provider`) return sanitized/generic messages and log the detail server-side — please do the same here.

### 🔵 P3 (nit) — SSRF gating divergence

This route validates unconditionally, whereas most sibling routes gate the check on `NODE_ENV === 'production'`. Unconditional is the more secure choice (and there's the `ALLOW_LOCAL_NETWORKS=true` escape hatch), but note it means a local LM Studio at `localhost:1234` is blocked unless that env var is set — which slightly undercuts the feature's main use case. A short code comment explaining the trade-off would help.

---

### Verified OK
Managed-provider key isolation (client key/baseUrl suppressed for admin-owned providers) ✅ · `AbortController` 10s timeout with `clearTimeout` on all paths ✅ · `localeCompare` guarded by `if (model.id)` ✅.

The P1 is the only hard blocker. Happy to re-review once the redirect handling is in.

wyuc:
**Re-review — second-round fixes look good, one new finding before merge**

Thanks for the quick turnaround. I re-checked the three items from the last round and they're genuinely closed:

- ✅ **Redirect-SSRF** — `redirect: 'manual'` with per-hop `validateUrlForSSRF` re-validation now mirrors `proxy-media` correctly. Break condition, `MAX_REDIRECTS`, and relative-`Location` resolution all check out — no off-by-one, the final response is only ever reached through already-validated URLs.
- ✅ **SSRF on the resolved base URL** — the guard now runs unconditionally on `resolveBaseUrl(...) || defaultBaseUrl`, so the fallback path is covered.
- ✅ **Upstream error body** — now logged server-side only; the client gets a generic message.

One new **blocking** issue introduced by the resolved-URL handling:

### 🔴 P2 (blocking) — Managed provider's server base URL is leaked back to the client · `app/api/list-models/route.ts`

```ts
return apiSuccess({ models, effectiveBaseUrl: baseUrl });
```

For a server-managed provider (configured via env / YAML), the route correctly ignores the client-sent `baseUrl`/key — but it then resolves `baseUrl` to the **server-configured** URL and returns it as `effectiveBaseUrl`. The settings UI persists that value when the local Base URL field is left blank (`components/settings/provider-config-panel.tsx`).

So any caller who can hit this endpoint can recover the admin-owned base URL of a managed provider just by passing its `providerId` — which may expose an internal gateway/proxy address that's intentionally hidden behind the managed-provider boundary. This is the same secrecy boundary the round-1 key/baseUrl isolation was protecting, so it's worth closing here too. Reachable unauthenticated when `ACCESS_CODE` is unset.

**Fix:** don't return `effectiveBaseUrl` for managed providers — only echo it back when it originated from a client/default *local* provider URL (where the client already knows it).

### 🟡 Optional hardening (non-blocking)

- **No body-size cap** — `response.json()` isn't bounded by the 10s `AbortController` (which only covers the fetch), and there's no `MAX_*_BYTES` cap like `proxy-media`. A slow/oversized upstream body can stall the request. Consider capping the response size.
- **Generic 500 echoes `error.message`** — the outer `catch` returns `error.message` to the client, which is inconsistent with the carefully sanitized 502 path just above. Suggest a fixed string + server-side log.
- **Azure host match is a substring check** — `effectiveBaseUrl.includes('.openai.azure.com')` would also match e.g. `https://x.openai.azure.com.example.com`. Low impact (it only appends a query param, destination host is unchanged), but `new URL(effectiveBaseUrl).hostname.endsWith('.openai.azure.com')` is the safer test.

The P2 is the only hard blocker — happy to re-review once `effectiveBaseUrl` no longer exposes managed URLs.

athul0rameshan9:
Slimmed this down and rebased on the latest `main`.

Since I first opened it, `main` gained a generic model-discovery flow (`app/api/provider/probe-models` + `lib/server/model-fetch.ts`, wired into settings for every `type: 'openai'` provider) and native Azure handling (`lib/ai/azure.ts`). Those cover most of what this PR originally carried, so I removed:

- the custom `app/api/list-models` route — this also resolves the last review blocker (the `effectiveBaseUrl` leak) by removing that code path entirely
- the bespoke settings UI and i18n keys
- the Azure fetch-wrapper hunk

What's left is just the built-in `lmstudio` provider registration (4 files, +18/−2), mirroring Ollama/Lemonade. LM Studio is `type: 'openai'`, so it picks up model discovery via the existing **Fetch Models** button automatically. Rebased cleanly on `main` — no conflicts. Closes #730.

lib/ai/providers.ts
```diff
@@ -1275,6 +1280,16 @@ export const PROVIDERS: Record<ProviderId, ProviderConfig> = {
...
   },
+
+  lmstudio: {
+    id: 'lmstudio',
+    name: 'LM Studio',
+    type: 'openai',
+    defaultBaseUrl: 'http://localhost:1234/v1',
+    requiresApiKey: false,
+    icon: '/logos/lmstudio.svg',
+    models: [],
```
wyuc, line 1291: LM Studio has no bundled models, so this PR depends entirely on the existing Fetch Models flow. That flow currently cannot discover models from a normal server-managed LM Studio setup: the settings UI hides the managed base URL and sends this hardcoded default to `/api/provider/probe-models`, while the route neither receives `providerId` nor resolves `LMSTUDIO_BASE_URL` server-side. A custom managed URL is therefore ignored, and the default `localhost:1234` is rejected by the unconditional SSRF guard unless `ALLOW_LOCAL_NETWORKS=true` is enabled. This leaves a fresh setup with no selectable models, contrary to the PR and #730's main goal. Please make probing provider-aware and resolve trusted managed configuration server-side without exposing the URL, or provide an equivalent discovery path. The unmanaged local case should also clearly document the `ALLOW_LOCAL_NETWORKS` requirement.

wyuc (changes requested):
Re-review of the slimmed-down implementation: the previous custom-route security findings are resolved by removing that code path, and the remaining provider registration is clean. One P2 correctness issue still blocks the feature's primary model-discovery flow; details are attached inline. All current CI checks are green.

wyuc:
Thanks @athul0rameshan9 for this contribution. I'm closing it to keep the queue tidy, since there hasn't been an update since the changes requested on 2026-07-24. It isn't a rejection of the idea: if you'd like to pick it back up, you're welcome to reopen it or open a fresh PR against the current `main` that addresses the review, and we'll take another look.

PR #746, opened 2026-06-16
feat(export): add video course export

wangxiaoming1988:
## Summary
- add MP4 video course export with selectable resolution
- include narration audio in ffmpeg composition and show detailed render errors
- cache generated narration audio by text and TTS settings to avoid duplicate synthesis

## Checks
- pnpm exec tsc --noEmit
- pnpm exec eslint app/api/export/video-course/route.ts components/stage/header-controls.tsx lib/export/use-export-video-course.ts
- pnpm run check:i18n-keys
- pnpm run build


danishsshaikh:
This is a feature I also need.

My use case is exporting the full generated classroom playback as a single video file, including:
- scene visuals
- narration/audio
- playback timing
- scene transitions where applicable

This would make it easier for teachers to share generated lessons outside the interactive OpenMAIC app, for example as an MP4 in an LMS or lecture archive.

I’m happy to help test this on self-hosted classrooms or pick up a focused follow-up if there is a smaller subtask that would be useful.

wyuc:
Thanks for putting together a working video export path. This PR is useful because it validates several concrete pieces: resolution selection, rendered scene images, narration audio collection/generation, server-side FFmpeg composition, and better render error reporting.

I opened #854 to consolidate the video export architecture. Based on that, I do not think we should make the current static-frame + custom FFmpeg route the canonical OpenMAIC video backend. It will be difficult to extend cleanly to the full classroom surface: embedded/generated videos, interactive HTML, quiz/PBL/widget/discussion fallbacks, subtitles, and action-level playback timing.

The current maintainer direction is to prefer a Hyperframes-based pipeline:

```text
OpenMAIC classroom data
  -> video export compiler / composition artifact
  -> Hyperframes render backend
  -> MP4/WebM output
```

So I would treat this PR as a valuable prototype/reference unless you want to adapt parts of it toward #854. The pieces most likely worth salvaging are:

- export UX around resolution/options;
- narration audio collection and cache behavior;
- render-progress/error messaging;
- lessons learned from server-side render failure handling.

I want to be explicit here so you do not spend more time polishing a backend path that we are unlikely to standardize on as-is.

cosarah:
Hi @wangxiaoming1988, thank you for this — the end-to-end MP4 export with selectable resolution, narration in the FFmpeg composition, and the TTS audio caching by text + settings is a genuinely useful piece of work, and it pushed the whole video-export effort forward.

We've since consolidated the architecture for classroom video export under #854. The direction we've settled on is to keep OpenMAIC classroom data as the single source of truth and use **Hyperframes** (HTML/CSS + seekable animation → MP4 via deterministic headless-Chrome frame capture) as the rendering backend, with a compiler layer in between. I validated the backend end-to-end with a P0 spike (spotlight mask + dual audio → MP4), and broke the work into tracked phases:

- #861 — persist audio duration on TTS write
- #863 — shared choreography/spec package (single source of truth)
- #864 — `VideoTimeline` IR contract + pure compile pipeline
- #865 — L1 emitter + collection layer + export ZIP (the first user-usable path)
- #866 — service-backed MP4 render
- #867 — polish + higher-fidelity track

Because we're moving the canonical backend to Hyperframes rather than a bespoke static-frame + FFmpeg route, this PR is best treated as a valuable prototype/reference rather than merged as-is — but several ideas in it map directly onto the new phases and we'd love to carry them over: the **resolution-tier selection + progress/error UX** (→ #866/#867) and the **TTS audio caching by text + settings** (→ relevant for #861/#865).

I'd really value your take on the #854 architecture. And if you're up for it, we'd be glad to have you contribute toward these issues — recovering the resolution/progress UX from this PR into #867, or the audio-cache design into #865, would be a great fit. No pressure either way, and thanks again for the contribution. 🙏

wyuc (changes requested):
This branch is valuable as a prototype, but it should not become the canonical backend after the #854 Hyperframes decision. There is also a concrete production-safety blocker in the current route: it accepts an unbounded number of scenes/tracks and unbounded uploaded image/audio sizes, then runs multiple FFmpeg/ffprobe processes, accumulates all segments on disk, and finally reads the entire MP4 into memory. A single request can exhaust CPU, temporary disk, memory, or the 300-second worker window.

Please do not merge this endpoint as-is. Salvage the resolution/progress/audio-cache pieces into the tracked #854 phases, where the render service can enforce authentication, quotas, byte/scene/duration limits, concurrency control, cancellation, and streamed output. The existing static-frame FFmpeg route remains a useful reference implementation.

wyuc:
Closing this prototype now that the canonical video-export architecture is active under #854 and the `VideoTimeline` compile pipeline is being implemented in #913/#864. The static-frame FFmpeg endpoint remains useful reference work, but it is intentionally not the merge path because it diverges from the chosen IR/Hyperframes backend and retains the unbounded render-resource blocker noted in review. The resolution/progress/error UX and audio-cache ideas can still be carried into #865-#867.

PR #747, opened 2026-06-16
feat(model): optional per-stage LLM model routing (#745)

### LINKED ISSUE #745: RFC: Per-stage LLM model configuration
> Route different generation stages (outline, scene content, quiz, in-class chat, PBL …) to different LLMs, instead of one global model for everything.

## Motivation

Today OpenMAIC resolves a **single** language model and uses it for *every* LLM call across the whole app. There is no supported way to say "use a cheap/fast model for outlines, a stronger model for scene content, and a specific provider for the in-class conversation."

This matters for three groups of operators:

1. **Cost** — the heavy stages (scene content) and the light stages (outline, quiz grading, query rewrite) have very different quality requirements but are forced onto the same model. Being able to put a smaller/cheaper model on the light stages is an easy, large saving with no quality loss.
2. **Quality** — some models are great at structured generation but weak at multi-turn teaching dialogue (or vice-versa). Operators want to mix.
3. **Incremental provider migration** — when you want to move a deployment from provider A to provider B, you currently have to flip the *entire* app at once. There is no way to validate B on one low-risk stage first, then expand. Per-stage configuration is exactly the mechanism that makes a gradual, de-risked migration possible.

## Current behavior

Model resolution is a single global knob — `lib/server/resolve-model.ts`:

```ts
const modelString = params.modelString || process.env.DEFAULT_MODEL || 'gpt-5.4-mini';
```

`params.modelString` comes from the `x-model` request header (still a *whole-app* selection, not per-stage). Every route then resolves this one model and hands it to `callLLM` / `streamLLM`.

The important part: **a per-stage label already exists at every call site.** `callLLM(params, source, …)` (`lib/ai/llm.ts`) is already called with a distinct `source` string everywhere, used for logging today:

`outline`, `scene-outlines-stream`, `scene-content`, `scene-actions`, `quiz-content`, `quiz-grade`, `pbl-design`, `pbl-chat`, `agent-profiles`, `chat-adapter`, `generate-classroom`, …

So the routing *key* is already threaded through the code. The only missing piece is a map from that key to a model.

## Proposal

Introduce an **optional** stage → model map, consulted during resolution and falling back to today's behavior when unset. Zero behavior change unless an operator opts in.

Resolution order for a given `source`:

```
stage override (if configured)  →  DEFAULT_MODEL  →  built-in fallback
```

Config-only, in the same spirit as the existing `DEFAULT_MODEL` / `*_MODELS` env conventions. Two equivalent surfaces:

**(a) A JSON map env (single var):**

```bash
# Cheap/fast model as the default; stronger model only for the heavy stages.
DEFAULT_MODEL=openai/gpt-5.4-mini
MODEL_ROUTES='{
  "scene-content": "openai/gpt-5.4",
  "scene-actions": "openai/gpt-5.4",
  "pbl-chat":      "anthropic/claude-sonnet-4",
  "chat-adapter":  "anthropic/claude-sonnet-4"
}'
```

**(b) Or YAML, alongside the existing server-providers config**, e.g. a `modelRoutes:` section keyed by stage.

Unlisted stages fall through to `DEFAULT_MODEL`, so an operator can override just one or two stages and leave the rest alone.

### Suggested stage groups

To keep the surface small and meaningful, stages naturally fall into two buckets that also map to migration risk:

- **Stateless one-shot generation** — `outline`, `scene-content`, `scene-actions`, `quiz-content`, `agent-profiles`. Single request → response. Easy and low-risk to point at a different model and validate.
- **Conversational / agentic** — `pbl-chat`, the in-class chat/director loop, `chat-adapter`. These depend on multi-turn behavior and (for some) eager tool-calling, which varies across providers. Operators will want to validate these per-provider before switching them.

Per-stage routing is what lets an operator move the first bucket immediately while keeping the second on a known-good model until it is validated — i.e. it turns an all-or-nothing switch into an incremental one.

## Scope / Non-goals

- **In scope:** server-side, config-only stage→model overrides with fallback to `DEFAULT_MODEL`; docs + `.env.example`; minimal tests.
- **Non-goals (for this RFC):**
  - No per-stage UI selector in this round — server config is enough to unblock the use cases; UI can follow if there is demand.
  - No automatic/learned routing — the map is operator-defined and static.
  - No change to how providers, keys, base URLs, or thinking config are resolved — only *which model string* a stage gets.
  - TTS/ASR/image/video resolution is unchanged (this is LLM-only).

## Backwards compatibility

Fully backward compatible. If `MODEL_ROUTES` / `modelRoutes` is unset, resolution is identical to today (`DEFAULT_MODEL`, optional `x-model` override). No migration required.

## Open questions

1. **Surface:** JSON env, YAML, or both? (Leaning: JSON env for parity with `*_MODELS`, optionally mirrored in YAML.)
2. **Stage taxonomy:** expose the raw `source` labels as the config keys, or define a smaller, documented, stable set of "stage" names that map to one-or-more sources? Raw labels are zero-effort but less stable across refactors; a curated set is friendlier but needs maintenance.
3. **Resolution site:** resolve per-stage inside the route (pass `source` into `resolveModel`) vs. let `callLLM` consult the map (it already has `source`). The latter centralizes it in one place.
4. Should `x-model` (whole-app override) take precedence over a stage route, or the other way round?

Feedback welcome — especially on the stage taxonomy (Q2) and whether a UI selector is wanted now or later.

---

## 中文摘要

**现状**:OpenMAIC 全程只解析**一个**全局模型(`resolve-model.ts` 的 `DEFAULT_MODEL`,或 `x-model` 头),所有环节共用。

**问题**:无法"简单环节用便宜模型、复杂环节用强模型",也无法在迁移 provider 时只先切一个低风险环节验证——只能整站一次性切换。

**关键事实**:其实每个 `callLLM(params, source, …)` 调用点**已经带了 `source` 标签**(`scene-content`/`scene-actions`/`outline`/`pbl-chat`/`chat-adapter`…),路由的 key 早就在了,**只差一张 source→model 的映射表**。

**提案**:加一个**可选**的 stage→model 映射(env `MODEL_ROUTES` JSON 或 YAML),解析顺序 `环节覆盖 → DEFAULT_MODEL → 内置兜底`;不配置时行为与今天**完全一致**,零迁移。

**分组**:无状态一次性生成(outline/scene/quiz)易切、可即时灰度;对话/agentic(pbl-chat、课堂对话)跨 provider 行为差异大,需逐个验证——**分环节路由正是把"全量切"变成"逐步切"的前提**。

**非目标**:本期不做 per-stage UI 选择器、不做自动路由、不动 provider/key/thinking/TTS/ASR 解析。

wyuc:
Closes #745.

Adds an **optional, config-only** stage → model map so operators can route
different generation stages to different LLMs, instead of one global model for
everything. Unset = identical to today's behavior (no migration required).

## What

- New `lib/server/model-routes.ts`: parses/validates/caches the `MODEL_ROUTES`
  env JSON and exposes a typed `LLM_STAGES` registry + `getStageModel(stage)`.
- `resolveModel` consults the stage during resolution. Resolution order:
  **stage route → x-model (client) → `DEFAULT_MODEL` → builtin fallback.**
- Each `resolveModel` call site threads its stage. `generate-classroom` resolves
  its `web-search-query-rewrite` sub-stage independently (lazily, only when
  routed and web search is enabled).
- **Per scene-content type:** `scene-content` can be routed per scene type via
  composite keys `scene-content:{slide,quiz,interactive,pbl}`. `getStageModel`
  resolves composite `a:b` keys most-specific-first (trimming `:` segments), so
  a type falls back to the base `scene-content` route, then x-model, then
  `DEFAULT_MODEL`. The scene-content route derives its stage from `outline.type`.
- Docs (`.env.example`) + unit tests.

## Config

Model strings use the canonical `provider:model` format. Example — cheap default,
stronger model only for the heavy/conversational stages:

```bash
DEFAULT_MODEL=openai:gpt-5.4-mini
MODEL_ROUTES='{"scene-content":"openai:gpt-5.4","scene-actions":"openai:gpt-5.4","pbl-chat":"anthropic:claude-sonnet-4","chat-adapter":"anthropic:claude-sonnet-4"}'
```

Per scene-content type — route only quiz/interactive, leave the rest on the base:

```bash
MODEL_ROUTES='{"scene-content":"openai:gpt-5.4-mini","scene-content:quiz":"openai:gpt-5.4","scene-content:interactive":"anthropic:claude-sonnet-4"}'
```

Routable stages: `scene-outlines-stream`, `scene-content`
(+ `scene-content:{slide,quiz,interactive,pbl}`), `scene-actions`,
`agent-profiles`, `quiz-grade`, `pbl-chat`, `chat-adapter`, `generate-classroom`,
`web-search-query-rewrite`. Unlisted stages fall through to `x-model`/`DEFAULT_MODEL`.

## Design decisions (re: the RFC open questions)

1. **Surface:** JSON env `MODEL_ROUTES` only this round (parity with
   `DEFAULT_MODEL` / `*_MODELS`); YAML mirror can follow if wanted.
2. **Stage taxonomy:** the existing `callLLM` source labels, frozen into the
   `LLM_STAGES` registry so config keys are validated (unknown keys warn) and a
   mistyped stage at a call site is a compile error.
3. **Resolution site:** in `resolveModel` (the single place that already owns
   provider/key/baseUrl resolution), not `callLLM` (which receives an
   already-built model).
4. **Precedence — stage route wins over `x-model`.** The browser UI always sends
   its saved model as `x-model`; if `x-model` won, `MODEL_ROUTES` would be inert
   for normal UI traffic on exactly the heavy stages this targets. A configured
   route is the operator's deliberate per-stage choice, so it wins; unrouted
   stages still honor the client's `x-model`. When a route overrides the client
   model, the client-sent `apiKey`/`baseUrl`/`providerType` are dropped so a
   routed (possibly cross-provider) model resolves purely from server config.

## Backwards compatibility

Fully compatible. With `MODEL_ROUTES` unset, resolution is identical to today.

## Test plan

- `lib/server/model-routes.ts` and `resolveModel` precedence covered by unit
  tests (parsing/validation, stage > x-model > DEFAULT_MODEL ordering,
  composite scene-content:<type> resolution + fallback, cross-provider routing,
  and dropping client connection params on a routed override).
- `pnpm test` (full suite) green; `tsc --noEmit`, `eslint`, and
  `prettier --check` all clean.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -210,12 +210,12 @@ WEB_SEARCH_MINIMAX_BASE_URL=https://api.minimaxi.com
 DEFAULT_MODEL=
 
 # Optional per-stage model routing (#745). A JSON object mapping a generation
-# stage to a model string (`provider:model`). Consulted between an explicit
-# x-model override and DEFAULT_MODEL: x-model > stage route > DEFAULT_MODEL.
-# Unset = identical to today (everything uses DEFAULT_MODEL). Unlisted stages
-# fall back to DEFAULT_MODEL, so you can override just one or two.
-# Note: this applies on server-managed deployments where the client does not send
-# a model. If the UI selects a model (sent as x-model), that wins over the route.
+# stage to a model string (`provider:model`). Resolution order for a stage:
+# stage route > x-model (client) > DEFAULT_MODEL. A configured route is the
+# operator's deliberate choice and wins even when the browser sends its saved
+# model as x-model. Unset = identical to today (everything uses DEFAULT_MODEL).
+# Unlisted stages fall back to x-model then DEFAULT_MODEL, so you can override
+# just one or two and leave the rest to the client/default.
 # Point a stage only at a server-configured provider (its key resolvable); a
 # route to an unconfigured/invalid model fails at request time for that stage,
 # the same way a bad DEFAULT_MODEL would (no startup validation).
--- a/.env.example
+++ b/.env.example
@@ -209,6 +209,19 @@ WEB_SEARCH_MINIMAX_BASE_URL=https://api.minimaxi.com
 # MiniMax example: minimax:MiniMax-M2.7-highspeed
 DEFAULT_MODEL=
 
+# Optional per-stage model routing (#745). A JSON object mapping a generation
+# stage to a model string (`provider:model`). Consulted between an explicit
+# x-model override and DEFAULT_MODEL: x-model > stage route > DEFAULT_MODEL.
+# Unset = identical to today (everything uses DEFAULT_MODEL). Unlisted stages
+# fall back to DEFAULT_MODEL, so you can override just one or two.
+# The target provider must be server-configured (its key resolvable).
+# Routable stages: scene-outlines-stream, scene-content, scene-actions,
+# agent-profiles, quiz-grade, pbl-chat, chat-adapter, generate-classroom,
+# web-search-query-rewrite.
+# Example: cheap default, stronger model only for the heavy/conversational stages:
+# MODEL_ROUTES='{"scene-content":"openai:gpt-5.4","scene-actions":"openai:gpt-5.4","pbl-chat":"anthropic:claude-sonnet-4","chat-adapter":"anthropic:claude-sonnet-4"}'
+# MODEL_ROUTES=
+
 # LOG_LEVEL=info
 # LOG_FORMAT=pretty
 # LLM_THINKING_DISABLED=false
--- a/.env.example
+++ b/.env.example
@@ -214,7 +214,11 @@ DEFAULT_MODEL=
 # x-model override and DEFAULT_MODEL: x-model > stage route > DEFAULT_MODEL.
 # Unset = identical to today (everything uses DEFAULT_MODEL). Unlisted stages
 # fall back to DEFAULT_MODEL, so you can override just one or two.
-# The target provider must be server-configured (its key resolvable).
+# Note: this applies on server-managed deployments where the client does not send
+# a model. If the UI selects a model (sent as x-model), that wins over the route.
+# Point a stage only at a server-configured provider (its key resolvable); a
+# route to an unconfigured/invalid model fails at request time for that stage,
+# the same way a bad DEFAULT_MODEL would (no startup validation).
 # Routable stages: scene-outlines-stream, scene-content, scene-actions,
 # agent-profiles, quiz-grade, pbl-chat, chat-adapter, generate-classroom,
 # web-search-query-rewrite.
--- a/.env.example
+++ b/.env.example
@@ -227,10 +227,15 @@ DEFAULT_MODEL=
 # scene-content:pbl. A type falls back to the base scene-content route when it
 # has no key of its own (so scene-content:<type> > scene-content > x-model >
 # DEFAULT_MODEL).
+# A route value can be a model string, OR an object {"model","effort"} to also
+# pin the thinking effort for that stage. effort ∈ none|minimal|low|medium|high|
+# xhigh|max ("none" disables thinking). When a stage is routed: a set effort wins
+# over the client's thinking; with no effort the routed model uses its own default
+# and the client's thinking is dropped. Unrouted stages keep the client thinking.
 # Example: cheap default, stronger model only for the heavy/conversational stages:
 # MODEL_ROUTES='{"scene-content":"openai:gpt-5.4","scene-actions":"openai:gpt-5.4","pbl-chat":"anthropic:claude-sonnet-4","chat-adapter":"anthropic:claude-sonnet-4"}'
-# Example: route only the quiz and interactive scene types, leave the rest:
-# MODEL_ROUTES='{"scene-content":"openai:gpt-5.4-mini","scene-content:quiz":"openai:gpt-5.4","scene-content:interactive":"anthropic:claude-sonnet-4"}'
+# Example: per scene type + pinned thinking effort:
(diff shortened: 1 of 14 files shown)


wyuc:
### Review summary — merging

Cross-reviewed for correctness and backward-compatibility across two rounds; CI green (Analyze, E2E, Lint/Typecheck/Unit).

**Verified**
- Resolution order: **stage route > x-model > `DEFAULT_MODEL`**; unrouted stages fall through to the client model then the default.
- Per scene-content **type** routing via composite keys (`scene-content:{slide,quiz,interactive,pbl}`) with most-specific-first fallback to the base `scene-content`.
- Per-stage **thinking** via the full `ThinkingConfig` in the route value (`{model, thinking}`), normalized per the model's capability; routed thinking wins over client thinking, an unrouted stage keeps the client's. All routable stages (incl. `chat-adapter` and the `generate-classroom` sub-stages) honor it.
- A routed (possibly cross-provider) model resolves its connection params (apiKey/baseUrl/providerType) purely from server config — client-sent params for the previous model don't leak onto it.
- **Backward compatible**: with `MODEL_ROUTES` unset, resolution is identical to before.

**One intentional behavior change (flagging explicitly)**
- Removed the hardcoded `gpt-5.4-mini` fallback in `resolveModel`. It now **throws** when a stage resolves to no model (no `MODEL_ROUTES` entry, no `x-model`, no `DEFAULT_MODEL`) — fail-loud instead of silently picking a vendor default. `DEFAULT_MODEL` is therefore effectively required for server-side stages; documented in `.env.example`.

Tests cover string-vs-object route values, invalid-field sanitization, composite fallback/precedence, the thinking arbitration cases, dropped client params on a routed override, and the throw-on-no-model path.

PR #749, opened 2026-06-17
fix(slide): tolerate malformed generated slide data

yipwingtim:
## Summary

  Improve slide renderer robustness when generated slide data contains malformed or missing fields.

  This prevents client-side render failures caused by common bad generated data, such as missing text
  values, invalid table cell spans, or unsafe numeric values.

  ## Changes

  - Normalize text-like values before rendering.
  - Guard table cell text and span values.
  - Avoid invalid shape transform values from malformed dimensions.
  - Keep rendering behavior unchanged for valid slide data.

  ## Verification

  - Cherry-picked onto latest `upstream/main`.
  - TypeScript/lint/build verified in downstream branch before preparing the upstream PR.

### DIFF
--- a/components/slide-renderer/components/element/ShapeElement/index.tsx
+++ b/components/slide-renderer/components/element/ShapeElement/index.tsx
@@ -63,8 +63,23 @@ export function ShapeElement({ elementInfo, selectElement }: ShapeElementProps) 
       defaultColor: '#000000',
     };
     if (!elementInfo.text) return defaultText;
-    return elementInfo.text;
+    return {
+      ...elementInfo.text,
+      content: typeof elementInfo.text.content === 'string' ? elementInfo.text.content : '',
+    };
   }, [elementInfo.text]);
+  const viewBoxWidth =
+    Array.isArray(elementInfo.viewBox) &&
+    Number.isFinite(elementInfo.viewBox[0]) &&
+    elementInfo.viewBox[0] > 0
+      ? elementInfo.viewBox[0]
+      : elementInfo.width || 1;
+  const viewBoxHeight =
+    Array.isArray(elementInfo.viewBox) &&
+    Number.isFinite(elementInfo.viewBox[1]) &&
+    elementInfo.viewBox[1] > 0
+      ? elementInfo.viewBox[1]
+      : elementInfo.height || 1;
 
   // Update text content
   const updateText = useCallback(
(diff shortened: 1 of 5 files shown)


yipwingtim:
This is a defensive renderer fix for malformed AI-generated slide data. It keeps valid slide rendering unchanged and prevents client crashes when generated content misses optional text/span/dimension fields.

wyuc (changes requested):
The direction here is good and I confirmed there's no regression for well-formed data — valid `content`/`colWidths` (including a legit `w: 0`), `colspan`/`rowspan` = 1, and valid `viewBox`/`width`/`height` all render identically, and the numeric/transform crashes (non-string `content`, non-array `data`/`colWidths`/`row`, `width / viewBox[0]` → `NaN`/`Infinity`) are now handled. Nice.

Before merging, though, the table hardening is asymmetric and still crashes on two malformed shapes it looks like it means to tolerate:

**1. A malformed (null / non-array) row still crashes `getHiddenCells`, before the render guard runs.**
`StaticTable` only guards rows at render time (`Array.isArray(row) ? row : []`), but `getHiddenCells(tableData)` runs first and does `data[rowIdx].length` / `data[rowIdx][colIdx]`, which throws on a bad row. So `data = [null]` or `data = [[...], 5]` crashes before the render guard is ever reached. Suggest normalizing rows before `getHiddenCells`, or having `getHiddenCells` skip non-array rows.

**2. A null cell still crashes the render path, even though `getHiddenCells` was hardened for it.**
`getHiddenCells` now uses `cell?.colspan` / `cell?.rowspan` (so it tolerates a null cell), but the render loop still does `cell.style?.backcolor`, `getTextStyle(cell.style)`, `formatText(cell.text)`, `cell.colspan` with no `cell?.` guard. So `data = [[{ id:'a', text:'x' }, null]]` is tolerated by `getHiddenCells` but throws at render. The two paths should agree — guard (or filter) null cells on the render side too.

Minor: `safeCellMinHeight` coerces a legitimately-zero `cellMinHeight` to 40 (`Number.isFinite(cellMinHeight) && cellMinHeight > 0`). Probably fine, just flagging that a valid `0` is altered — `>= 0` (or a null/NaN-only check) would preserve it.

With the row/cell guards made symmetric this fully delivers the "tolerate malformed slide data" goal. Happy to help iterate — thanks!

yipwingtim:
Thanks for the careful follow-up. I updated the table hardening to make the row/cell guards symmetric.

Changes made:
- getHiddenCells now skips malformed non-array rows before reading row.length or row[colIdx].
- StaticTable now skips null/invalid cells in the render loop, matching the getHiddenCells tolerance.
- safeCellMinHeight now preserves a valid 0 value and only falls back for non-finite/invalid values.
- Added a targeted test covering malformed rows and null cells in getHiddenCells.

Verification:
- pnpm vitest run tests/slide-renderer/table-utils.test.ts
- pnpm lint
- pnpm build

wyuc (approved):
Thanks for the update — I re-reviewed the table hardening changes against the previous requested changes.

The blockers I raised are addressed:

- `getHiddenCells()` now skips malformed non-array rows before reading row length/cells.
- `StaticTable` now guards malformed rows and skips null cells in the render loop, so the hidden-cell and render paths are symmetric.
- `safeCellMinHeight` now preserves a valid `0` and only falls back for non-finite/invalid values.
- The added test covers malformed rows and null cells for the hidden-cell path.

CI is green, and the remaining renderer guards look scoped to malformed generated data without changing valid table rendering semantics. Approving.

PR #750, opened 2026-06-17
fix(generation): strip reasoning blocks before JSON parsing

yipwingtim:
## Summary
  Strip model reasoning blocks before parsing JSON responses.

  Some reasoning models may wrap intermediate thoughts in `<think>...</think>` before emitting the final
  JSON. This can prevent otherwise valid JSON extraction and parsing.

  ## Changes

  - Remove `<think>...</think>` blocks before JSON extraction.
  - Apply all existing parsing strategies to the cleaned response.
  - Log parse failure snippets from the cleaned response so diagnostics match the parser input.

  ## Verification

  - `pnpm lint` passes locally with existing warnings.
  - `pnpm exec tsc --noEmit` was attempted locally, but the current working tree has stale `.next` generated type references and unresolved workspace package types unrelated to this one-file change.

### DIFF
--- a/lib/generation/json-repair.ts
+++ b/lib/generation/json-repair.ts
@@ -33,8 +33,51 @@ function logJsonParseError(stage: string, jsonStr: string, error: unknown): void
 }
 
 export function parseJsonResponse<T>(response: string): T | null {
+  const exactParsed = tryParseExactJson<T>(response);
+  if (exactParsed !== null) return exactParsed;
+
+  const cleanedResponse = stripReasoningPrefix(response);
+  if (cleanedResponse !== response.trim()) {
+    const parsedCleaned = parseJsonResponseCandidate<T>(cleanedResponse);
+    if (parsedCleaned !== null) return parsedCleaned;
+  }
+
+  const parsed = parseJsonResponseCandidate<T>(response);
+  if (parsed !== null) return parsed;
+
+  log.error('Failed to parse JSON from response');
+  log.error('Raw response (first 500 chars):', cleanedResponse.substring(0, 500));
+  log.error(
+    'Raw response (last 500 chars):',
+    cleanedResponse.substring(Math.max(0, cleanedResponse.length - 500)),
+  );
+
+  return null;
+}
+
+function tryParseExactJson<T>(response: string): T | null {
+  try {
+    return JSON.parse(response.trim()) as T;
+  } catch {
+    return null;
+  }
+}
+
+function stripReasoningPrefix(response: string): string {
+  const trimmed = response.trim();
+  const matches = [...trimmed.matchAll(/<\/(?:think|thinking|reasoning)>\s*/gi)];
+  const lastMatch = matches.at(-1);
+
+  if (!lastMatch || lastMatch.index === undefined) return trimmed;
+
+  return trimmed.slice(lastMatch.index + lastMatch[0].length).trim();
+}
+
+function parseJsonResponseCandidate<T>(response: string): T | null {
+  const cleanedResponse = response.trim();
+
   // Strategy 1: Try to extract JSON from markdown code blocks (may have multiple)
-  const codeBlockMatches = response.matchAll(/```(?:json)?\s*([\s\S]*?)```/g);
+  const codeBlockMatches = cleanedResponse.matchAll(/```(?:json)?\s*([\s\S]*?)```/g);
   for (const match of codeBlockMatches) {
     const extracted = match[1].trim();
     // Only try if it looks like JSON (starts with { or [)
@@ -49,8 +92,8 @@ export function parseJsonResponse<T>(response: string): T | null {
 
   // Strategy 2: Try to find JSON structure directly in response (no code block)
   // Look for array or object start
-  const jsonStartArray = response.indexOf('[');
-  const jsonStartObject = response.indexOf('{');
+  const jsonStartArray = cleanedResponse.indexOf('[');
+  const jsonStartObject = cleanedResponse.indexOf('{');
 
   if (jsonStartArray !== -1 || jsonStartObject !== -1) {
     // Prefer the structure that appears first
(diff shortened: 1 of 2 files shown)


wyuc (changes requested):
Thanks for this — stripping reasoning blocks before JSON parsing is the right direction. `parseJsonResponse` on `main` currently does nothing about `<think>`, so this is a real gap to close.

Before merging I'd like a more complete/robust version, since `parseJsonResponse` is a shared parser used across scene / outline / quiz / search-query generation. Three concerns:

**1. Unpaired trailing `</think>` isn't handled — and it's the most common case.**
The regex `/<think>[\s\S]*?<\/think>\s*/gi` only matches a *paired* block. Many providers stream reasoning via a separate `reasoning_content` channel; when that gets merged back into the text, only a closing tag survives:
```
…reasoning prose… </think>
{ …json… }
```
That output isn't cleaned at all. Strategy 2 then runs `indexOf('{' / '[')` over the reasoning prose — if the prose contains a `{`/`[` (LaTeX, code, set notation), extraction locks onto the wrong offset and parsing fails or returns wrong data. This is exactly the case the change is meant to fix.

**2. The unconditional global strip can corrupt valid JSON.**
The pre-clean runs on the whole response *before* any parse attempt, so a legitimate JSON whose *string values* contain a literal `<think>…</think>` (e.g. content that explains reasoning tokens, or HTML-ish payloads) is silently mutated before parsing. Low probability, but this is a shared parser, so the blast radius is wide.

**3. Only `<think>` is covered.**
`<thinking>`, `<reasoning>`, and `<|…|>`-style delimiters pass through untouched. Not blocking, but the title says "reasoning blocks" more broadly.

**Suggested direction** (any combination works):
- Strip up to and including the **last** `</think>` (covers both paired and unpaired) instead of only matching paired `<think>…</think>`; and/or
- Apply the pre-clean as a **fallback only** — try to parse first, and strip reasoning wrappers only when the initial parse fails. That removes the corruption risk in (2) entirely; and/or
- Anchor the strip to the reasoning **prefix** (before the first JSON structure) rather than a global replace.

Two unit tests would lock this down: (a) an unpaired trailing `</think>` followed by JSON, and (b) a JSON string value that legitimately contains `<think>`.

Happy to help iterate — thanks again!

yipwingtim:
Thanks for the detailed review. I updated the parser to first try the raw response and only strip
  reasoning prefixes as a fallback, so valid JSON strings containing literal <think>...</think> are
  preserved.

  The fallback now handles trailing closing reasoning tags such as </think>, </thinking>, and </reasoning>
  before the JSON payload, including the unpaired </think> case. I also added tests for both the unpaired
  trailing </think> case and preserving literal <think> tags inside valid JSON strings.

wyuc (changes requested):
Thanks for the update. The raw-first fallback fixes the previous valid-JSON string mutation concern, and the unpaired closing tag case is now covered.

One remaining blocker: the raw parse can still accept JSON from the reasoning prefix before reaching the final answer. For example:

```text
<think>{"draft": true}</think>
{"ok": true}
```

Because `parseJsonResponseCandidate(response)` runs before `stripReasoningPrefix`, Strategy 2 extracts and returns `{"draft": true}` from inside the reasoning block, so the parser never reaches the actual final JSON. `parseJsonResponse<T>` does not validate the returned shape, so a parseable scratch object/code block in model reasoning can become the generated scene/outline/quiz/search payload.

Could you add a regression test where the reasoning prefix contains parseable JSON or a fenced JSON fragment before the closing reasoning tag, followed by the real payload, and adjust the parser so obvious reasoning-prefix responses prefer the payload after the closing reasoning tag while still preserving valid raw JSON strings that contain literal `<think>...</think>` text?

yipwingtim:
Thanks, good catch. I updated the parser so responses that clearly start with a reasoning wrapper now
prefer the payload after the closing reasoning tag before trying the raw extraction path.

This prevents parseable draft JSON or fenced JSON inside `<think>...</think>` from being returned as the
final payload, while still preserving valid raw JSON strings that contain literal `<think>...</think>`
text.

Added regression coverage for both:
- parseable draft JSON inside a reasoning block before the final payload
- fenced draft JSON inside a reasoning block before the final payload

Verification:
- pnpm vitest run tests/generation/json-repair.test.ts
- pnpm lint
- pnpm build

wyuc (changes requested):
Thanks for the follow-up — the paired `<think>...</think>` draft cases are fixed, but one blocker remains on the common unpaired-closing-tag path.

`hasReasoningPrefix()` is false when the response contains only `</think>`, so `parseJsonResponseCandidate(response)` still runs before `stripReasoningPrefix()`. If the reasoning prefix contains parseable scratch JSON, Strategy 2 returns that draft and never reaches the final payload:

```ts
const raw = `reasoning draft {"draft": true} </think>\n{"ok": true}`;
expect(parseJsonResponse(raw)).toEqual({ ok: true });
// current result: { draft: true }
```

The existing unpaired-tag test uses `{not json}`, so the raw candidate fails and allows the fallback to run; it does not cover the wrong-valid-JSON case above.

Please add this regression and make a detected closing reasoning tag prefer the post-tag payload before permissive extraction from the prefix, while still preserving an exact valid JSON response whose string value contains literal `<think>...</think>` tags. One workable order is: preserve an exact whole-response JSON parse first, then prefer the cleaned post-tag candidate when a closing tag is present, then fall back to the existing permissive raw extraction/repair strategies.

yipwingtim:
Thanks for catching this. I updated the parsing order to cover the unpaired closing-tag case with parseable scratch
JSON.

The parser now:
1. First attempts an exact whole-response JSON.parse, preserving valid JSON responses whose string values contain
literal <think>...</think> tags.
2. If the exact parse fails and a closing reasoning tag is present, it prefers the payload after the last </think>, </
thinking>, or </reasoning> tag.
3. Only then falls back to the existing permissive extraction and repair strategies.

I added the requested regression test for:

reasoning draft {"draft": true} </think>
{"ok": true}

The parser now returns the final {"ok": true} payload rather than the scratch JSON.

Verification:
- pnpm vitest run tests/generation/json-repair.test.ts — 7 tests passed
- pnpm lint — 0 errors
- pnpm build — passed
- Rebased onto the latest upstream/main

wyuc (approved):
Re-reviewed head `2f0b1f1`. The parser now preserves exact JSON first, prefers the payload after paired or unpaired reasoning closing tags, and covers the prior wrong-valid-draft regression. Focused local verification: `pnpm vitest run tests/generation/json-repair.test.ts` — 7/7 passed. Current CI is green. LGTM.

PR #759, opened 2026-06-18
feat(outline): make scene type authoritative via derive-on-confirm specializer

wyuc:
## Problem

Changing a scene's `type` in the outline editor (e.g. `slide` → `interactive`) is **silently reverted to `slide`** at generation time. Root cause: downstream routing keys off the *presence of type-specific config* (`widgetOutline` / `pblConfig` / `quizConfig`) rather than the `type` field, and `applyOutlineFallbacks` downgrades any type whose config is missing — with only a server-side `warn`, no UI signal. The user's edit is lost.

## Approach (keep `type` explicit, derive its config on confirm)

Add `lib/generation/outline-specializer.ts`. At scene-content time, when the declared `type` lacks its required config, derive that config from the outline's intent fields (`title` / `description` / `keyPoints`):

- **quiz / pbl** derive purely from common fields (pbl maps title/description/keyPoints; quiz uses sensible defaults).
- **interactive** derives `widgetType` + a minimal `widgetOutline` via one LLM call, defaulting to `diagram` if the model is unavailable or returns something invalid. (Most widget-content variables already fall back to the outline's common fields, so a minimal `{ concept }` is enough to render.)
- **slide** is a no-op.
- A `specializedFor` intent-hash tags the derived config as a **cache**: when intent is unchanged and the config is present, specialization returns the outline untouched — re-confirming costs nothing. The derived outline rides back as the route's `effectiveOutline`.
- The gated **vocational `procedural-skill`** path is preserved (sanitized to a diagram when not enabled).

Wiring:
- `scene-content` route calls `specializeOutline` instead of the silent `applyOutlineFallbacks` revert. `applyOutlineFallbacks` is kept for server-batch / classroom auto-gen paths, where an LLM-produced outline legitimately lacking config should still degrade to slide.
- The outline editor clears stale type-specific config when the user changes a scene's type (`resetConfigForType`), so derivation starts clean.

## Scope

6 files, +469/-3. New: `outline-specializer.ts` + its test. Edited: `generation.ts` (`specializedFor`), `generation-pipeline.ts` (barrel), `scene-content/route.ts` (call site), `outlines-editor.tsx` (type-change reset).

## Tests

25 unit tests covering intent hashing, per-type config checks, quiz/pbl/interactive derivation, the derive-on-confirm cache (hit / stale invalidation), pbl-without-language-model degradation, and procedural-skill gating. `tsc --noEmit` clean on the touched files.

## Status

**Draft** — opening for review of the interface and the derive-on-confirm approach. Only `interactive` incurs an LLM call (and only when its config is missing); quiz/pbl/slide are pure. Default behavior for outlines that already carry matching config is unchanged (cache hit → zero extra calls).

### DIFF
--- a/app/api/generate/scene-content/route.ts
+++ b/app/api/generate/scene-content/route.ts
@@ -9,7 +9,7 @@
 import { NextRequest } from 'next/server';
 import { callLLM } from '@/lib/ai/llm';
 import {
-  applyOutlineFallbacks,
+  specializeOutline,
   generateSceneContent,
   buildVisionUserContent,
 } from '@/lib/generation/generation-pipeline';
@@ -128,8 +128,13 @@ export async function POST(req: NextRequest) {
     };
 
     // ── Apply fallbacks ──
+    // ── Derive-on-confirm: ensure the declared type's config exists ──
+    // The scene `type` is authoritative; if the user changed it in the editor
+    // the matching config is derived here (cached via specializedFor).
     const vocationalActive = resolveVocationalActive(requirements);
-    const effectiveOutline = applyOutlineFallbacks(outline, !!languageModel, {
+    const effectiveOutline = await specializeOutline(outline, {
+      aiCall,
+      hasLanguageModel: !!languageModel,
       allowProceduralSkill: vocationalActive,
     });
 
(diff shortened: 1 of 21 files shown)


wyuc:
Pausing this direction for now — closing. The branch is kept for reference.

PR #769, opened 2026-06-21
fix(generation): don't regenerate a deleted slide on a finished deck

### LINKED ISSUE #768: Deleting a slide in Pro/Edit mode resurrects it — the page is re-generated and appended to the end after returning to the classroom
## Description

After a course finishes generating, opening **Pro mode** (edit mode) and deleting a slide, then returning to the classroom view, causes the deleted slide to be **re-generated and appended as the last page** of the course.

## Steps to reproduce

1. Generate a course with N slides.
2. Open Pro mode and delete one slide (middle or last).
3. Navigate back to the classroom (`/classroom/[id]`).
4. Observe: the deleted slide is regenerated and appended at the end.

**Expected**: a deleted slide stays deleted.
**Actual**: the deleted slide is regenerated and lands as the last page.

## Root cause

A course keeps two parallel structures: the `outlines` plan (persisted for resume-on-refresh) and the actual `scenes`. Resume-generation reconciles them **by `order`**.

1. `lib/store/stage.ts` `deleteScene()` (lines 193–209) removes the scene from `scenes` only — it **never updates `outlines`**, leaving an orphaned outline that is still persisted.
2. `outlines` is persisted for resume-on-refresh via `setOutlines()` (lines 234–242).
3. On returning, the *"Auto-resume generation for pending outlines"* effect in `app/classroom/[id]/page.tsx` (lines 143–189) computes `hasPending = outlines.some(o => !completedOrders.has(o.order))` (line 152). The deleted slide's `order` is now absent from `scenes`, so `hasPending` is true and `generateRemaining()` is called.
4. `lib/hooks/use-scene-generator.ts` recomputes `pending = outlines.filter(o => !completedOrders.has(o.order))` (lines 311–313) and regenerates that outline.
5. `addScene()` (`lib/store/stage.ts:141`) pushes the new scene to the end of `scenes`, and `deleteScene()` does not rebalance `order` (leaving a gap), so the regenerated slide appears as the last page.

## Suggested fix

Have `deleteScene()` also drop the matching outline (by `order`) from `outlines` and persist it. Optionally rebalance scene `order` after deletion to remove the gap. The undo path is unaffected: re-inserting the scene restores its `order`, so `hasPending` stays false without restoring the outline.

wyuc:
## Summary

Fixes #768.

Deleting a slide in Pro/Edit mode removed it from `scenes` but left its entry in the persisted `outlines` plan. On the next classroom mount the resume-generation pass treats the now-missing `order` as a pending outline and regenerates the slide, appending it as the last page.

## Approach

Pruning outlines by `order` was rejected: `order` is rebalanced by Pro-mode insert/duplicate/reorder, so it is not a stable key — pruning could mis-target another slide's outline and would break delete-undo (the recycle entry only restores the scene). Instead, resume is gated on a persisted `generationComplete` flag:

- The flag is recorded (alongside outlines, in the `stageOutlines` record) when generation finishes — the two `generateRemaining` completion paths, the retry-to-completion path (`markGenerationCompleteIfDone`), and the already-materialized-at-mount path. It is restored on load. The flag is written only **after** a verified scene flush, so it can never outrun the deck's scenes.
- A finished deck is frozen for editing: `loadFromStorage` no longer derives pending placeholders from orphaned outlines, the classroom resume effect skips `generateRemaining`, media resume is filtered to outlines that still have a scene, and `isCourseComplete` honours the flag — so a deleted slide neither regenerates nor leaves a phantom pending page.
- Decks generated before the flag existed self-heal: if every outline already has a matching scene, the deck is inferred complete and the flag is persisted, so the first deletion does not regenerate.

`outlines` are never pruned, so order-rebalancing and delete-undo are unaffected. Genuinely interrupted generations (not complete) still resume as before.

## Test plan

- [x] 11 new unit tests for the store flag / persistence / self-heal / completion paths (`tests/store/stage-generation-complete.test.ts`).
- [x] Full suite green (864 tests), lint clean.
- [x] Manual browser E2E (real headless browser, upstream app): on `origin/main`, deleting a middle slide in Pro mode and reopening the course regenerates the deleted slide at the end (bug reproduced); on this branch, the deleted slide stays deleted and is not regenerated after reopen.

### DIFF
--- a/app/classroom/[id]/page.tsx
+++ b/app/classroom/[id]/page.tsx
@@ -154,11 +154,14 @@ export default function ClassroomDetailPage() {
     if (loading || error || generationStartedRef.current) return;
 
     const state = useStageStore.getState();
-    const { outlines, scenes, stage } = state;
+    const { outlines, scenes, stage, generationComplete } = state;
 
-    // Check if there are pending outlines
+    // Check if there are pending outlines. A finished deck is frozen for
+    // editing: deleting a slide leaves its outline orphaned, but that must not
+    // be treated as an interrupted generation and regenerated. Only resume
+    // when generation has not completed.
     const completedOrders = new Set(scenes.map((s) => s.order));
-    const hasPending = outlines.some((o) => !completedOrders.has(o.order));
+    const hasPending = !generationComplete && outlines.some((o) => !completedOrders.has(o.order));
 
     if (hasPending && stage) {
       generationStartedRef.current = true;
@@ -191,7 +194,19 @@ export default function ClassroomDetailPage() {
       // Resume media generation for any tasks not yet in IndexedDB.
       // generateMediaForOutlines skips already-completed tasks automatically.
       generationStartedRef.current = true;
-      generateMediaForOutlines(outlines, stage.id).catch((err) => {
+      // The deck reached the classroom already fully materialized (e.g. a
+      // single-slide course, or a deck whose last slide finished in
+      // generation-preview), so generateRemaining's completion path never
+      // ran. Record completion now so a later edit/delete is not treated as
+      // an interrupted generation. No-op if already complete or not all
+      // outlines have scenes.
+      useStageStore.getState().markGenerationCompleteIfDone();
+      // Resume media only for outlines that still have a scene. On a finished
+      // deck the user may have deleted a slide, leaving an orphaned outline;
+      // generating its media would waste API calls on a slide that is gone.
+      const materializedOrders = new Set(scenes.map((s) => s.order));
+      const materializedOutlines = outlines.filter((o) => materializedOrders.has(o.order));
+      generateMediaForOutlines(materializedOutlines, stage.id).catch((err) => {
         log.warn('[Classroom] Media generation resume error:', err);
       });
     }
--- a/components/edit/PlaybackChromeRoot.tsx
+++ b/components/edit/PlaybackChromeRoot.tsx
@@ -82,6 +82,7 @@ export const PlaybackChromeRoot = forwardRef<PlaybackChromeRootHandle, PlaybackC
       outlines,
     } = useStageStore();
     const failedOutlines = useStageStore.use.failedOutlines();
+    const generationComplete = useStageStore.use.generationComplete();
 
     const currentScene = getCurrentScene();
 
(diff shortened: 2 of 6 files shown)


wyuc:
### Summary

Fixes #768 — a slide deleted in Pro mode reappears (regenerated at the end) after reopening a finished course.

**Root cause:** a course keeps a persisted `outlines` plan alongside the rendered `scenes`. Deleting a slide removed only the scene; on the next mount the resume-generation pass treated the now-missing `order` as "pending" and regenerated the slide, appending it at the end.

**Approach:** rather than prune `outlines` by `order` (which Pro-mode insert/duplicate/reorder rebalances, so it is not a stable key — pruning would mis-target another slide's outline and break delete-undo), resume is gated on a persisted `generationComplete` flag. A finished deck is frozen for editing:
- `loadFromStorage` no longer derives pending placeholders from orphaned outlines;
- the classroom resume effect skips `generateRemaining`;
- media resume is filtered to outlines that still have a scene;
- `isCourseComplete` honours the flag.

The flag is set on every completion path (generation finish, retry-to-completion, already-materialized-at-mount), persisted only after a verified scene flush so it never outruns the deck, self-heals decks generated before the flag existed, and `deleteScene` records completion when deleting from an already-complete deck. `outlines` are never pruned, so order-rebalancing and delete-undo are unaffected; genuinely interrupted generations still resume.

**Verification:** new unit tests cover the flag lifecycle (persistence, flush ordering, self-heal, delete-from-complete-deck); the full suite and lint/format pass. Manually verified in a real browser — the bug reproduces on `main` (delete → reopen → slide regenerated at end) and is fixed on this branch (deleted slide stays, end page intact), with normal and interrupted generation unaffected.

wyuc:
@sophietao20-star Thanks for taking this! A quick way to validate end-to-end:

**Reproduce the bug first (without this PR — e.g. current `main`):**
1. Generate a course and let it finish.
2. Enter Pro mode (专业模式) and delete a middle slide.
3. Go back to home, then reopen the course.
4. **Bug:** the deleted slide is regenerated and appended as the **last page** (slide count returns to the original).

**Confirm the fix (with this PR):**
- Repeat the same steps → the deleted slide **stays deleted** (no regeneration, no phantom pending page), and the **"Course complete" end page still shows**.
- Sanity check: a normal generation still completes, and an interrupted generation still resumes on refresh.

sophietao20-star:
Validated the manual E2E testing.

Reproduced the bug on `main`. Verified the fix on this branch: the deleted slide is no longer regenerated after reopening the course, the "Course complete" page still appears correctly, and no regressions were observed in normal generation or refresh recovery.

sophietao20-star (approved):

PR #770, opened 2026-06-21
feat(maic-agent): fix_interactive_html — surgical bug-fix for interactive scenes

wyuc:
## What

Adds a new editor-agent tool **`fix_interactive_html`** so the Pro-mode "Edit with AI" agent can surgically fix bugs in an **interactive scene's HTML** (a button that does nothing, a control with no effect, an animation that never shows, a layout glitch) instead of refusing — today `regenerate_scene` is slide-only and interactive scenes can't be touched.

The teacher describes the bug in natural language in the AgentBar; the agent reads the page from the trusted scene context and asks the model to **change only the reported problem and return the complete corrected HTML**, then it's applied with one-tap undo.

## How

- **Generator** `fixInteractiveHtml(currentHtml, bugDescription, aiCall)` — one LLM round-trip, then reuses `extractHtml` + `postProcessInteractiveHtml` (same path as fresh widget generation). Returns `null` (fail loud) when the prompt template is missing or no HTML can be extracted; never silently ships the original page.
- **Prompt** `fix-interactive-html` (`system.md` / `user.md`) — instructs a minimal, targeted fix; preserve all other markup/ids/classes/styles/scripts/CDN links verbatim; keep element ids stable (teacher actions reference them); output the full document only.
- **Tool** `fix_interactive_html({ sceneId, bugDescription })` — same trust boundary as `regenerate_scene` (model supplies only id + description; html comes from `getSceneContext`). interactive-only; refuses non-interactive or empty-html scenes with a typed error. Registered in the v0 toolset + allowlist; the editor system prompt routes interactive bug reports here.
- **Client apply/undo** — a `planRegenerateApply` branch snapshots the current `InteractiveContent` and writes the new `html`, preserving `url` / `widgetType` / `widgetConfig` / `teacherActions`. The iframe reloads via `srcDoc`; Restore reverts through the existing snapshot store.
- **UI** — an AgentPanel tool card mirroring `regenerate_scene` (status + one-tap Restore), with `edit.fixHtml` strings (en-US / zh-CN / zh-TW; other locales fall back).

## Model routing (black-box tools)

The editor generation tools are now **self-contained model black boxes**: each resolves its own generation-stage model via `MODEL_ROUTES` rather than sharing the `maic-agent` conversation model. `fix_interactive_html` → `scene-content:interactive`, `regenerate_scene` → `scene-content:slide` (+ `scene-actions`), `regenerate_scene_actions` → `scene-actions`. `deps.aiCall(stage, …)` is resolved per stage in the route (cached per turn). Unrouted stages fall back to the client model, so default behaviour is unchanged; operators can now route editor generation independently of the agent's conversation model. The agent only decides *which* tool to call.

## Scope (v0)

Teacher-driven only; full-HTML rewrite (no diff); interactive scenes only. Out of scope: iframe runtime-error capture / auto-detect, search/replace diff apply, multi-scene batch fix.

## Tests

`vitest` — generator (fix + KaTeX post-process + null on no-HTML), tool (fix / refuse non-interactive / refuse empty / missing context), client apply (writes html, preserves metadata, snapshots for restore), registry + system-prompt capability. Full suite green (1015 tests); `tsc --noEmit` and `eslint` clean.

> Draft, stacked on `feat/maic-editor-agent-v0`.


sophietao20-star (approved):
Approved. 
(on non-interactive slides the agent may regenerate static UI that looks like interactive controls—an expected boundary issue,  not a regression in the interactive HTML fix loop.)

PR #772, opened 2026-06-22
fix(generation): make outline type changes take effect (interactive/PBL no longer downgraded to slide)

wyuc:
## Problem

In the outline editor (`OutlinesEditor`), the type dropdown (`TypePill`) lets you
switch a scene outline between **slide / quiz / interactive / pbl**. But the
handler only set `{ type }` (a shallow merge) and never supplied the new type's
required config, and a config form existed only for `quiz`. So switching a scene
to **interactive** (needs `widgetType` + `widgetOutline`) or **pbl** (needs
`pblConfig`) produced an outline that `applyOutlineFallbacks`
(`lib/generation/outline-generator.ts`) defensively downgrades back to **slide**:

- `interactive` without `interactiveConfig` and without (`widgetType` + `widgetOutline`) → `slide`
- `pbl` without `pblConfig` → `slide`

Net effect: **changing an outline's type to interactive/pbl silently doesn't take
effect** — it renders as a slide. `slide`/`quiz` worked only because they need no
required config (quiz defaults to `{3, medium, [single]}`).

Root cause: the editor could hold a typed outline lacking the config its type
requires (an invalid-by-construction state), with no UI to make it valid.

## Fix

- **`changeOutlineType(outline, newType)`** total constructor
  (`lib/generation/outline-type.ts`): returns a new outline that strips foreign
  per-type config and seeds the target type's required config from the shared
  fields (title / description / keyPoints), so the result is valid by
  construction and survives `applyOutlineFallbacks`. Switching away from a type
  also clears its config so stale data is never persisted.
- `TypePill` now **replaces** the outline via `changeOutlineType` (replace, not
  partial-merge).
- New **`InteractiveConfigDisclosure`** (widget-kind picker + concept) and
  **`PblConfigDisclosure`** (project topic / description / target skills),
  mirroring `QuizConfigDisclosure`, so interactive and PBL scenes — which
  previously had no input UI at all — can be configured.
- i18n keys added across all 8 locales.

## Scope / non-goals

- Deliberately **not** retyping `SceneOutline` into a full discriminated union
  across the codebase: the outline-generation prompts already require valid
  interactive/pbl config, classroom export carries only `type + content` (not
  `SceneOutline`), and `applyOutlineFallbacks` stays as the runtime net.
  Validity is enforced at the single mutation point (the editor type switch). A
  union refactor can be a follow-up.
- `procedural-skill` widget (a gated vocational feature) is intentionally not
  exposed in the manual widget picker.

## Tests

- `tests/generation/outline-type.test.ts` — constructor behavior + the bug-fix
  invariant: `applyOutlineFallbacks(changeOutlineType(o, 'interactive' | 'pbl')).type`
  keeps the chosen type.
- `tests/i18n/outline-review-locales.test.ts` — extended to cover the new keys
  in every supported locale.
- Full `tsc --noEmit` clean; `next build` passes.

Manual e2e screenshots to follow.


wyuc:
Verified locally on a production build (`next build` + `next start`):

- Unit tests green: `changeOutlineType` behavior + the invariant `applyOutlineFallbacks(changeOutlineType(o,'interactive'|'pbl')).type` keeps the chosen type; i18n locale-coverage extended for the new keys.
- Full `tsc --noEmit` clean.
- Manual e2e in the outline editor: switching a scene Slide → Interactive now yields `{type:'interactive', widgetType:'simulation', widgetOutline:{concept:<title>}}` (valid by construction; previously a config-less interactive that got downgraded to slide). The new InteractiveConfigDisclosure (widget kind: simulation/diagram/code/game/3D + concept) and PblConfigDisclosure (topic/description/target skills) render and persist correctly.

Marking ready for review once CI is green.

wyuc:
**Cross-review (Claude `/code-review` + codex `exec review`) — converged after 4 rounds.**

Each round, both reviewers independently audited the full `origin/main..HEAD` diff; findings were merged by consensus, fixed, and re-reviewed until both sides reported zero actionable findings.

- **R1** — preserve an existing `procedural-skill` widget config in `changeOutlineType` instead of replacing it with a seeded simulation; reset `widgetOutline` to the shared `concept` on a widget-kind change so foreign kind fields don't leak; broaden the i18n locale-coverage test to ko-KR/pt-BR; drop a redundant cast; add tests.
- **R2** — surface a preserved `procedural-skill` widget as its own labeled, selected kind (no longer mislabeled "Simulation") and make a same-kind re-select a no-op so its task-engine fields are never clobbered; dedicated `removeSkill` aria-label.
- **R3** — make re-selecting the *current* type a no-op in `changeOutlineType`: a rebuild on same-type selection could drop legacy `interactiveConfig` / a partial `pblConfig`.
- **R4** — both reviewers clean.

Final: `tsc --noEmit` clean, eslint + prettier clean, 11 unit tests pass.

wyuc:
@sophietao20-star could you run an e2e check on this fix? Suggest **reproduce first, then confirm**:

**Reproduce (on `main`):** In the outline review/edit step, open a scene's type dropdown and switch it to **Interactive** or **PBL**. The change silently doesn't take effect — at generation time the scene is downgraded back to a **slide** (because the editor never supplies the required per-type config, and `applyOutlineFallbacks` drops it).

**Confirm (on this branch / PR #772):**
1. Switching a scene to **Interactive** sticks, and a widget-kind + concept config appears as a cascading control to the left of the type pill; the generated scene keeps the interactive type.
2. Switching a scene to **PBL** sticks, and a project topic / description / target-skills form appears; the scene generates as PBL.
3. Switching back to **slide/quiz** still works; re-selecting the *current* type changes nothing (no data loss).

Thanks!

sophietao20-star (dismissed):
The fix is working as expected based on manual E2E testing.

* **Reproduced on `main`:** Changing a scene type to **Interactive** or **PBL** in the outline editor had no effect during generation. 

* **Verified on `fix/outline-type-change`:**
  * Scene type changes are applied correctly and no longer fall back to `slide`.
  * Widget type changes also work as expected.
  * Config persists after a page refresh.

PR #774, opened 2026-06-22
feat(ai): add GLM-5.2 and Kimi K2.7 Code

### LINKED ISSUE #773: [Feature]: Add GLM-5.2 and Kimi K2.7 Code model catalog support
## Summary

Add first-class model catalog support for the latest official GLM and Kimi coding models:

- `glm-5.2`
- `kimi-k2.7-code`
- `kimi-k2.7-code-highspeed`

## Scope

- Add the models to the built-in GLM and Kimi provider catalogs.
- Align model metadata with official API behavior only.
- Keep provider-specific reasoning controls consistent with each model's official API constraints.
- Add focused tests for model availability and request parameter injection.

## Official API notes

- GLM-5.2 uses `model: "glm-5.2"`, supports a 1M context window and 128K maximum output, supports thinking mode, and adds `reasoning_effort` for controlling reasoning intensity.
- Kimi K2.7 Code uses `model: "kimi-k2.7-code"`; `kimi-k2.7-code-highspeed` is the same model family with identical parameter constraints and a faster serving tier.
- Kimi K2.7 Code always has thinking enabled. Passing `thinking.type = "disabled"` should not be exposed for this model.
- Kimi K2.7 Code has fixed sampling parameters. The catalog should avoid introducing OpenMAIC-side controls that imply those values are configurable.

## References

- https://docs.z.ai/guides/llm/glm-5.2
- https://docs.z.ai/guides/overview/concept-param
- https://platform.kimi.ai/docs/guide/kimi-k2-7-code-quickstart
- https://platform.kimi.ai/docs/api/models-overview

wyuc:
## Summary

Closes #773.

Adds first-class catalog support for the latest official GLM and Kimi coding models:

- `glm-5.2`
- `kimi-k2.7-code`
- `kimi-k2.7-code-highspeed`

## Details

- Adds `glm-5.2` to the GLM provider catalog with official 1M context and 128K output metadata.
- Adds both Kimi K2.7 Code variants to the Kimi provider catalog with 256K context and 32K output metadata.
- Models Kimi K2.7 Code thinking as fixed-on, so OpenMAIC does not expose an unsupported disable toggle.
- Adds GLM-5.2 reasoning effort metadata using the official effort values, with `max` as the default.
- Updates GLM OpenAI-compatible request injection so GLM-5.2 sends `reasoning_effort` when the catalog exposes effort-based thinking.

## Validation

- `NODE_OPTIONS=--max-old-space-size=1024 VITEST_MAX_WORKERS=4 pnpm exec vitest run tests/ai`
- `NODE_OPTIONS=--max-old-space-size=1024 pnpm exec tsc --noEmit --pretty false --incremental false`
- `NODE_OPTIONS=--max-old-space-size=1024 pnpm exec eslint lib/ai/providers.ts lib/ai/model-metadata.ts tests/ai/openai-provider.test.ts tests/ai/thinking-config.test.ts`
- `pnpm check`
- `NODE_OPTIONS=--max-old-space-size=1024 VITEST_MAX_WORKERS=4 pnpm exec vitest run`


wyuc:
## Cross-review — ✅ no correctness issues, safe to merge

Two independent model reviews of `origin/main..HEAD`, plus a run of the affected tests: **34/34 pass** (`tests/ai/thinking-config.test.ts`, `tests/ai/openai-provider.test.ts`).

### P1 / P2: none

Both reviews agree on the key risk areas:

- **No regression for existing GLM models.** Splitting `glm` out of the shared `kimi/xiaomi` fallthrough is safe: all non-effort GLM models (`glm-5.1`, `glm-5`, `glm-5v-turbo`, `glm-4.7*`, `glm-4.6*`) use `toggleCapability` (control `'toggle'`) and route through the new non-effort fallback at `lib/ai/providers.ts:1268-1270`, which preserves the old behavior exactly (`disabled→{type:'disabled'}` / `enabled→{type:'enabled'}` / else `undefined`). `glm-5.2` is the only effort GLM.
- **New `glm` effort branch handles partial configs correctly:** `{effort:'minimal'}` with no mode still emits `reasoning_effort:'minimal'`; `{mode:'enabled'}` with no effort backfills `defaultEffort:'max'`; `effort:'none'` and `mode:'disabled'` both short-circuit to `{thinking:{type:'disabled'}}`.
- **`kimi-k2.7-code(-highspeed)` → `fixedThinkingCapability` (`control:'none'`) is coherent:** `getCompatThinkingBodyParams` returns `undefined` for `control==='none'`, so no thinking field is sent — correct for an always-on, non-toggleable model. The tests assert exactly this.

### P3 (optional nits)

1. **`lib/ai/providers.ts:1264` — unreachable clause.** In `if (mode === 'enabled' || effort)`, whenever `mode === 'enabled'` the preceding lines always assign `effort = defaultEffort` (`'max'`, truthy), so the `mode === 'enabled' ||` part can never be the deciding condition. It's equivalent to `if (effort)`. Simplifying makes the "thinking is always paired with a `reasoning_effort`" invariant explicit and removes the only path that could ever emit `thinking:{type:'enabled'}` without an effort.
2. **Divergence from the `deepseek` branch on a fully-empty config.** For `{thinking:{}}` (no `mode`/`enabled`/`effort`), the new `glm` branch returns `undefined` (model uses its server-side default, matching `defaultEnabled:true`), whereas the `deepseek` branch it's modeled on force-emits `{thinking:{type:'enabled'}, reasoning_effort:'high'}`. The GLM behavior is arguably the more correct one — flagging only to confirm the divergence is intentional.
3. **Redundant inline `capabilities.thinking` blocks** on the three new model entries: `applyModelMetadata` overwrites them at runtime from `THINKING_CAPABILITIES`, so they're never read. Values agree and this matches the existing convention for every model in the file, so no action needed.

### Spec values to double-check against provider docs (not code-checkable)

- `glm-5.2` context `1000000` / output `128000`
- `kimi-k2.7-code` / `-highspeed` context `256000` / output `32768`, and `vision:true`
- GLM accepting `reasoning_effort` ∈ {`minimal`, `xhigh`, `max`}

sophietao20-star (approved):
Completed E2E testing.

- [x] GLM-5.2 appears in the model picker; thinking effort control works, with `max` as the default.
- [x] Both Kimi K2.7 Code variants appear in the picker; no thinking-off control (fixed-on, as designed)
- [x] All of the above models successfully generate a course.
- [x] Dev-server logs show expected model IDs on generation (`model=glm:glm-5.2`, …)
- [x] Different GLM-5.2 thinking levels show observable differences as expected in generation (tested: `off` / `medium` / `max`)

Issue #778, opened 2026-06-24
[Task] Publish the @maic/* package family to npm (claim the npm org first) — #720 Phase 1


wyuc:
> Breaks out the "Publish to npm" item from **#720 Phase 1** into an actionable, ordered task. First publish is a one-time, largely-irreversible "claim + foundation" step, so the ordering matters.

## Current state

- None of `@maic/dsl`, `@maic/renderer`, `@maic/importer` are on npm — and the **`@maic` scope itself is unclaimed** (registry returns 404 for the scope).
- `publishConfig` (public, registry.npmjs.org) and `files: [dist, README, LICENSE]` are already set on each package.
- No `prepublishOnly` / build guard — `build` is a manual script, so a publish without an explicit build ships stale/empty `dist`.
- The `dist/index.js` extensionless re-export bug (breaks bare-node ESM resolution) is still open — a package published today would not import cleanly for external ESM consumers, which defeats the point of publishing.

## Key point: claiming the scope ≠ publishing a package

`@maic/*` is reserved by **creating and owning the npm org `maic`**. Creating the org reserves the entire scope — no package publish is required to hold it. So scope-claim is **org provisioning + ownership**, decoupled from shipping working code. This sidesteps the "burned version number / broken package" risk of a rushed publish.

## Prerequisites before a real (`latest`) publish

- [ ] **Create the npm org `maic`**; decide ownership (a project/org account, not a personal token), enable 2FA and publish provenance
- [ ] **Fix the dist ESM resolution bug** (extensionless re-exports → emit `.js`) so packages import cleanly
- [ ] Add `prepublishOnly: build && test` to each package (guard against shipping stale/empty `dist`)
- [ ] `npm pack --dry-run` and review the actual tarball contents before first publish
- [ ] Publish from this repo as the canonical source of the public packages
- [ ] (ideally) a CI publish workflow with provenance, triggered on tag — none exists yet

## Optional: visibly reserve with a placeholder

If a published placeholder is wanted before the real release, publish `0.0.1` under a **non-`latest` dist-tag** (e.g. `--tag placeholder`) so an accidental `npm install` does not pull a not-yet-ready version as the default.

## Versioning

Real first release is `0.1.x` (current package versions). `0.x` allows breaking changes during stabilization, so iterating after first publish is cheap.

Tracks the "Publish `@maic/dsl`, `@maic/renderer`, `@maic/importer` to npm" box under **#720 Phase 1**.

[...]

PR #780, opened 2026-06-24
chore(packages): publish the @openmaic/* SDK family to npm (#778)

wyuc:
Prepares the `@openmaic/{dsl,renderer,importer}` family for its first npm publish, and moves the SDK packages onto the `@openmaic` scope. Closes the "Publish to npm" item under #720 Phase 1; tracks #778.

## Why the scope move (`@maic` → `@openmaic`)

The `@maic` org name is **not claimable** on npm — an unscoped `maic` package already holds the name, and an org name can't collide with an existing package. `@openmaic` matches the project name, the scope is free, and the repo already ships an `@openmaic/docs` package, so the SDK family now lines up with that existing convention.

## What's in here

- **Scope rename** `@maic/*` → `@openmaic/*` across the three packages, the package directory, `pnpm-workspace.yaml`, and all import sites; lockfile regenerated (`--frozen-lockfile` passes, no dependency changes).
- **renderer**: add `publishConfig` (public, registry.npmjs.org) — it was missing, so a scoped publish would default to the wrong registry / restricted access.
- **importer**: add a `files` allowlist (`dist`, `README`, `LICENSE`) and drop the fragile `.npmignore` blacklist that shipped `src/`; add an `exports` map so ESM consumers resolve `dist/index.js` instead of falling back to the `.cjs` `main`.
- **all three**: add a `prepublishOnly` build (+ test/typecheck) guard so a publish can never ship a stale or empty `dist`.
- **CI**: add a tag-triggered publish workflow with npm provenance, **pinned by name** to the three `@openmaic` packages so the vendored forks (`mathml2omml`, `pptxgenjs`) are never published.

## Notes

- The earlier "extensionless re-export breaks bare-node ESM" concern from #778 is already resolved in `dsl` (explicit `.js` re-exports + `NodeNext`); `renderer`/`importer` bundle via rollup so it never applied to them.
- The npm org `openmaic` is already created and owns the scope, so this PR is purely the code-side prerequisites. Actual publish is a follow-up (add `NPM_TOKEN` repo secret, push a `@openmaic/<pkg>@<version>` tag).
- `importer` is at `0.3.0`; `dsl`/`renderer` at `0.1.0`. `dsl` must publish first (the other two depend on it) — the workflow's `pnpm -r` handles that order.

Refs #778, #720.


wyuc:
**Cross-review (Claude + codex, 2 rounds) — converged.**

- Round 1 caught a real break: `@openmaic/renderer` advertised a CJS entry but externalizes ESM-only `@openmaic/dsl`, so `require('@openmaic/renderer')` would throw. Renderer is now ESM-only (dropped the `.cjs` output + `require` conditions).
- Round 2: codex flagged that `@openmaic/importer`'s published bundle throws `XMLHttpRequest is not a constructor` in a pure Node process — it's a browser-targeted build (browser pdf.js). The app consumes it client-side only, so this is **documented as an accepted, browser-only limitation** (prominent README note + `browser` field) rather than blocking the publish. `dsl`/`renderer` are Node-clean.
- Rename completeness (zero `@maic` left), lockfile consistency, the publish-workflow scoping (pinned to the three `@openmaic` names; vendored forks never published), and the `next.config.ts` merge resolution were all verified clean.

PR #777, opened 2026-06-23
feat(maic-agent): promote editor-agent line (v0) — Pro-mode "Edit with AI"

wyuc:
## Summary

Promote the editor-agent line (`feat/maic-editor-agent-v0`) to `main`. This brings the Pro-mode **"Edit with AI"** agent and the surrounding editor surfaces together onto the default branch.

41 commits, 148 files. The pieces landed and were reviewed individually on the feature line; this PR is the promotion of that vetted line as a whole.

## What this includes

- **Editor agent (Pro mode)** — a capability-bounded "Edit with AI" agent that reads a scene, then acts:
  - `read_scene_content` — read-only projection of the current scene for the model.
  - `regenerate_scene` — whole-slide regenerate (content + narration) from a natural-language instruction; slide-only this release.
  - `regenerate_scene_actions` — regenerate narration/actions to match edited content.
  - `edit_interactive_html` — surgical, anchor-based bug-fix for interactive scenes.
- **Routable LLM stage** (`maic-agent`) — the agent is a per-stage model route, independent of the content-generation stages.
- **Agent UI** — sidebar chat with streaming markdown, reasoning panel, tool cards with robust states, and undo ↔ resume.
- **Interactive runtime-error capture** — sandboxed iframes report runtime errors back to the host so the agent can fix the root cause of a blank/broken page instead of guessing.
- **Timeline / script editor** — narration script bar with inline cue badges and canvas spotlight sync.
- `lib/agent/VENDOR.md` records the fork intent for the pinned agent runtime.

## Review

- The constituent PRs on this line were each cross-reviewed before merging.
- Final leak-scan over the full `main...HEAD` delta is clean.
- A cross-vendor review pass over the full delta is being run against this PR before merge.

## Tests / build

- Type-check and the touched test suites pass on the branch tip.

PR #781, opened 2026-06-24
fix(619): keep-alive e2e aligns with #777 edit-mode visibility (+ pi-ai server-external)

wyuc:
Fixes the `interactive-iframe-keepalive-619.spec.ts` e2e that has been red on `main` since #777. Two related #777-fallout fixes.

## 1. The actual e2e cause: test asserted pre-#777 behavior (primary)

#777 deliberately changed `InteractiveIframeHost.tsx` — it dropped the `mode !== 'edit'` guard so the interactive iframe **stays visible in Pro/edit mode**:

```diff
- visible={mode !== 'edit' && entry.owner !== null && sceneId === activeSceneId}
+ visible={entry.owner !== null && sceneId === activeSceneId}
```

with an explicit rationale in the doc comment it added: *"the editor agent fixes interactive HTML, so the teacher must see the live page while editing."* That is the intended behavior — but the #619 keep-alive e2e was never updated, so its `toBeHidden()` at line 189 ("hidden in edit mode") failed on every run since #777.

Fix: update Trigger A to the new contract — after the Pro-mode toggle the iframe is **visible** and its in-iframe counter state is preserved (the real keep-alive proof: neither unmounted nor reloaded). Trigger B (scene switch to a slide → ownership released → hidden) is unchanged.

## 2. Server-side dynamic-import error spam from #777's agent libs (cleanup)

While investigating, the CI logs showed repeated `Cannot find module as expression is too dynamic` unhandled rejections. These come from `@earendil-works/pi-ai` / `pi-agent-core` (added in #777), which lazily load node builtins via a computed `import(specifier)` that webpack can't statically analyze. They are **not** the e2e cause (the error persisted/cleared independently of the test result), but they are real server-side unhandled rejections. Marking both packages `serverExternalPackages` makes Next load them natively so the dynamic import resolves as a real Node call.

Refs #619, #777.

### DIFF
--- a/e2e/tests/interactive-iframe-keepalive-619.spec.ts
+++ b/e2e/tests/interactive-iframe-keepalive-619.spec.ts
@@ -182,11 +182,15 @@ test.describe('#619 interactive iframe keep-alive', () => {
     await resetMutations(page);
 
     // `.first()` because the ~280ms mode cross-fade briefly mounts both chrome
-    // layers (each with its own Pro switch). iframe visibility is the unambiguous
-    // "transition settled" signal.
-    // --- Trigger A: Pro mode toggle (edit chrome unmounts playback chrome) ---
+    // layers (each with its own Pro switch).
+    // --- Trigger A: Pro mode toggle (edit chrome remounts the placeholder) ---
     await page.getByRole('switch').first().click();
-    await expect(iframeEl).toBeHidden(); // hidden in edit mode, not unmounted
+    // Since #777 the interactive iframe stays VISIBLE in edit mode — the editor
+    // agent ("Edit with AI") fixes interactive HTML, so the teacher must see the
+    // live page while editing. The keep-alive proof is that it is neither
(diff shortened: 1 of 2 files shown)


wyuc:
**Cross-review (independent Claude + codex passes over the branch diff) — zero actionable findings.**

- `serverExternalPackages`: the `@earendil-works/pi-*` libs do a computed `import(specifier)` for node builtins that webpack can't statically resolve; externalizing lets Node resolve it natively. Both packages are needed (`pi-agent-core` reaches the bad module through `pi-ai`'s `export *`). Scanned the rest of the #777 dep cluster (`@assistant-ui/*`, `streamdown`, `typebox`) — none use the same pattern, nothing missing. No client-bundle impact (the only client-side import is type-only).
- e2e assertion: matches the post-#777 component behavior (the interactive keep-alive iframe now stays visible in edit mode); the keep-alive proof — no remount + preserved counter state — is intact.

PR #784, opened 2026-06-24
feat(token-plan): one-click token-plan setup + deployment usage dashboard

### LINKED ISSUE #783: [Feature]: Token Plan one-click setup + deployment usage dashboard
### Problem or Motivation

Configuring a multi-modal "token plan" today means hand-filling base URLs, protocols, and model lists across several settings panels (LLM provider, image, video, TTS, web-search) — error-prone and repetitive. There is also no visibility into how many tokens / images / seconds the deployment has actually consumed.

### Proposed Solution

Two related capabilities:

**1. Token Plan one-click setup** (new "Token Plan" section in System Settings)
- Preset picker for token plans / aggregators / vendors. A preset declares, per modality, the target provider id + base URL; applying a key fills every declared modality and lights it up.
- Data-driven: adding a vendor = one entry in `lib/config/token-plan-presets.ts`, no code changes.
- Probes the provider's `/models` endpoint (multi-candidate fallback) to auto-populate the LLM model list; falls back to preset-declared `defaultModels` for endpoints that expose no model list (e.g. Volcengine Agent Plan).
- Ships with: MiniMax (full-set template), 火山方舟 Agent Plan, OpenRouter, SiliconFlow, DeepSeek, GLM, Qwen.

**2. Deployment usage dashboard** (new "Usage" section)
- Captures token usage at the `callLLM`/`streamLLM` chokepoint and non-LLM generation (image/video/TTS) at the generate routes.
- Stores usage as monthly JSONL (no cost/pricing — usage only).
- Per-modality dashboard: by-model / by-day / by-kind aggregation, theme-aware chart.

### Alternatives Considered

- Per-vendor balance/quota query via cloud AK/SK signing — rejected: every vendor needs its own signature scheme (Volcengine SigV4 / Tencent TC3 / Aliyun BSS), heavy and vendor-specific. We track usage, not cost.
- Custom (manual baseURL/protocol) token plan entry — rejected: a one-off provider is better configured directly on the Providers page.

### Area

API / Backend

### Additional Context

Implementation lands on branch `feat/token-plan-usage` (PR to follow). Notable design choices documented in commit messages.

yanpgwang:
## Summary

Adds two related capabilities to System Settings:
1. **Token Plan** — a preset-driven, one-click setup that fills a single key across every modality a plan declares (LLM / image / video / TTS / web-search) and auto-populates the LLM model list.
2. **Usage** — a deployment-wide usage dashboard that captures token/image/second/character usage and aggregates it by model / day / modality. Usage only, no cost.

## Related Issues

Closes #783

## Changes

- **Token Plan presets** (`lib/config/token-plan-presets.ts`, `apply-token-plan.ts`): data-driven presets; applying a key lights up each declared modality. Ships MiniMax (full-set), 火山方舟 Agent Plan, OpenRouter, SiliconFlow, DeepSeek, GLM, Qwen. Adding a vendor = one entry, no code changes.
- **Model probing** (`lib/server/model-fetch.ts`, `app/api/provider/probe-models`): multi-candidate `/models` discovery with SSRF guard; falls back to a preset's `defaultModels` for endpoints with no model list (e.g. Volcengine Agent Plan → `ark-code-latest`).
- **Usage capture & storage** (`lib/usage/normalize.ts`, `lib/server/usage-storage.ts`, `lib/ai/llm.ts`, generate routes): usage captured at the `callLLM`/`streamLLM` chokepoint and the image/tts/video routes; persisted as monthly JSONL.
- **Usage dashboard** (`components/settings/usage-dashboard.tsx`, `app/api/usage`): per-modality by-model / by-day / by-kind aggregation, theme-aware chart.
- **Settings UI** (`token-plan-settings.tsx`, `provider-config-panel.tsx`, `index.tsx`): Token Plan section, Fetch Models button, preset picker.
- Shared helpers (`fetch-with-timeout.ts`, `modelInfoFromId`), i18n keys across 8 locales, Volcengine cloud logo, tests.

## Type of Change

### Steps to reproduce / test

1. System Settings → Token Plan → pick a preset (e.g. 火山方舟 Agent Plan) → enter API key → Apply.
2. Confirm the LLM provider is lit up and its model list is populated.
3. System Settings → Usage → confirm requests/tokens aggregate after some generations.

- Volcengine Agent Plan end-to-end with a real `ark-` key: `/api/plan/v3/chat/completions` returns 200; the plan exposes no `/models` (404), so `ark-code-latest` (auto-routing alias) is seeded and works.
- OpenAI-compatible plans (OpenRouter/DeepSeek/GLM/Qwen) use the standard `/models` probe path.
- NOT verified in this PR: image/video generation via token plans (their adapters hardcode `/api/v3` paths; deferred until a second media-capable plan justifies making adapter paths configurable).

### Evidence

- [x] Manually tested locally

## Checklist


wyuc (commented):
### Overall

Nice direction — the data-driven preset approach (`token-plan-presets.ts` + `apply-token-plan.ts`) and the `verifyModels` probe-on-apply are clean, and the usage dashboard is a useful addition. I do have one scoping concern and a couple of concrete correctness issues found while testing against real MiniMax / Volcengine Ark plan keys.

### 1. Scope "Token Plan" to actual token plans (MiniMax + Volcengine Ark)

A "token plan / 套餐" has a defining property: **one key spans multiple modalities** (LLM + image + video + TTS + …). MiniMax and the Volcengine Ark Agent Plan are exactly that. The other presets shipped here aren't:

- **Aggregators** (OpenRouter, SiliconFlow) and **vendor-direct** (DeepSeek, GLM, Qwen) are all **single-modality (LLM-only)**. They're just normal API providers, already covered by the existing add-provider flow.

Grouping these under a feature literally named **"Token Plan"** conflates "convenience LLM preset" with "multi-modal plan". It muddies the concept, duplicates the existing provider setup, and makes the value proposition ("one key lights up every modality") inconsistent — most entries don't do that.

**Suggestion:** limit this feature to true token plans (MiniMax, Volcengine Ark Agent Plan). If quick LLM-provider presets are still wanted, they fit better as presets inside the regular "add provider" flow rather than under "Token Plan".

### 2. Correctness issues found in testing

**(a) Ark video `defaultModels` contains an invalid id.** In the `volcengine-ark` preset, `video.defaultModels: ['doubao-seedance-2.0', 'doubao-seedance-1.5-pro']` — `doubao-seedance-2.0` is rejected by the Agent Plan endpoint as `UnsupportedModel`. `verifyModels` prunes it so nothing breaks, but it means Seedance 2.0 is **never offered even on a tier that supports it**. The working id is `doubao-seedance-1.5-pro`; for 2.0 use the catalog id (e.g. `doubao-seedance-2-0-260128`) or drop the entry.

**(b) MiniMax token plan lights up "Video", but video generation fails out of the box.** The MiniMax preset advertises the full set incl. video, but `minimax-video` declares `supportedResolutions: ['720p', '1080p']` while the default model `MiniMax-Hailuo-2.3` only accepts `768P`/`1080P`. `normalizeVideoOptions` forces the unsupported `720p`, so the API returns `2013 ... does not support resolution 720P`. (Looks pre-existing in `video-providers.ts`, but this PR is what exposes it as a one-click modality.) Worth fixing the supported-resolution list or making it model-aware so the advertised modality actually works.

### 3. Depth: the Ark Agent Plan covers more than LLM/image/video

If "token plan = one key, all modalities" is the thesis, the Ark Agent Plan also includes **TTS, ASR, and embeddings (vectorization)** under the same plan key/billing — all reachable today. Note the voice endpoints live on a **different host** (`openspeech.bytedance.com/api/v3/plan/tts/…` and `…/sauc/…`) with a **non-OpenAI protocol** (custom chunked/WebSocket + `X-Api-Resource-Id` header), so they'd need a dedicated adapter rather than the OpenAI-compatible path. Reasonable as follow-up, but flagging it since it's where the "token plan" promise gets fully realized for Ark.

### 4. Minor: "lit up" ≠ "generation works"

The result panel marks a modality green once the provider is configured, but configured ≠ generatable (see 2b: MiniMax video lights green yet fails; Ark video depends on plan tier). Consider distinguishing "configured" from "verified", or not implying generation will succeed.

yanpgwang:
Thanks for the thorough review @wyuc — all four points addressed. Summary of what changed, with the evidence behind the two cases where I diverged from the literal suggestion (all probed against a live Ark Agent Plan key).

### 1. Scope to true multi-modal token plans ✅
Dropped the single-modality LLM presets (OpenRouter, SiliconFlow, DeepSeek, GLM, Qwen). Only **MiniMax** and the **Volcengine Ark Agent Plan** remain — the two that actually light up multiple modalities from one key. The picker already hides empty categories, so no dangling sections. (`25e5477`)

### 2a. Ark video model id — kept `doubao-seedance-2.0` ⚠️
Couldn't switch to the catalog id as suggested: on the plan endpoint `doubao-seedance-2-0-260128` returns **`UnsupportedModel` (404)**, while the dotted alias `doubao-seedance-2.0` is the one the official Agent Plan curl example uses and that the gateway accepts. On my key's tier 2.0 is gated (`UnsupportedModel`), but `verifyModels` prunes it on apply and an upgraded tier lights it up with no code change — so the candidate is kept best-first (`2.0` → `1.5-pro`) rather than dropped. Same reasoning widened the image list to a best-first Seedream `5.0/4.5/4.0`. (`d55da6c`)

### 2b. MiniMax video resolution ✅
Fixed. `normalizeVideoOptions` defaulted to `720p`, but Hailuo 2.3 only accepts `768P`/`1080P` (`2013 ... does not support resolution 720P`). MiniMax's mid tier is **768P**, so the shared `720p` now maps to `768P` (the adapter already fell back to 768P). Regression tests lock the mapping. (`536052e`)

### 3. More modalities under the plan key ✅ (TTS + web search) / ⏳ (ASR)
- **Web search**: 豆包搜索 (Custom 版) wired as a one-click modality — the plan key authenticates directly against `open.feedcoopapi.com/search_api/web_search` (verified). (`00d3c1f`)
- **TTS**: Doubao Seed-TTS 2.0 via the HTTP `…/api/plan/tts/unidirectional` endpoint. `generateDoubaoTTS` now picks auth from the key shape (single key → `X-Api-Key`; `appId:accessKey` → app-id/access-key headers), since the two are not interchangeable (a plan key 401s on the normal endpoint; the plan endpoint rejects the app-id pair). (`0177b18`, `d55da6c`)
- **ASR**: deferred to a follow-up. Under the Agent Plan, ASR is **WebSocket-only** (binary framing: header bit-packing + gzip + big-endian seq) — no HTTP path. The HTTP "录音文件识别" line exists but its plan variant (`/api/v3/plan/auc/bigmodel/submit`) is gated behind an undocumented internal `X-Inner-Vendor-Key` and rejects the plan key otherwise. It's a self-contained binary-protocol client, better as its own PR than bloating this one. Embeddings likewise TBD.

### 4. "lit up" ≠ "generation works" ✅
Reworked the result panel: it now renders **only after probing finishes** (so no premature green), with two states — green when a modality is configured/usable, muted when a live probe proved it unavailable on the tier (e.g. video on a tier that doesn't include it). (`c1068f8`)

All green locally: 1163 tests pass, tsc/eslint/prettier clean.

wyuc:
One small UX follow-up on the Apply flow (optional, non-blocking):

Right now Apply spins the button and then renders the whole result panel at once after probing finishes — which reads as dead air while the model probe runs. Consider a **progressive reveal** instead: drop the modality rows in immediately on Apply, each with its own loading spinner, and let each row resolve to lit/muted as its own probe returns.

That's progressive disclosure of *structure*, not a premature green — a row only turns green once its own probe confirms — so it **complements** the "render only after probing" change here rather than reverting it, while giving live progress during the wait. Parallelize the probes, optional small stagger + `prefers-reduced-motion`, and it also clears the "did Apply do anything?" gap. Cheap win — happy to send a PR if useful.

yanpgwang:
Done — implemented the progressive reveal in `5d7495e`.

On Apply the rows now drop in immediately: each modality with a live probe in flight shows a spinner (`pending`) and resolves to lit/muted on its own as its probe returns; non-probe modalities (TTS, web search) show lit right away. Probes run in parallel via `Promise.all` instead of the old sequential loop, and a row only turns green once its own probe confirms — so it's structural disclosure + live progress, not a premature green, which keeps the "no premature green" intent intact.

Kept it minimal: no artificial stagger (the spinners already convey per-row progress and real probe latency varies), so there's nothing motion-wise to gate behind `prefers-reduced-motion`. Happy to add a staggered entrance later if you think it reads better.

wyuc (commented):
Thanks for the token plan work. I found two issues that I think should be addressed before merge:

1. Token Plan models should reuse/enrich the existing model config instead of creating fresh metadata.

In `lib/config/apply-token-plan.ts:110-126` and `components/settings/token-plan-settings.tsx:173-205`, Token Plan builds new `ModelInfo` objects from model ids and only fills basic capabilities via `modelInfoFromId()` (`components/settings/utils.ts:12-21`). This bypasses the existing built-in model metadata/config, so model capabilities drift from the normal provider path.

A concrete user-facing result: models that should support configurable thinking do not get `capabilities.thinking`, so `InlineThinkingControl` is hidden by `components/generation/generation-toolbar.tsx:437-438`. We should keep the Token Plan endpoint/model id where required, but enrich the model object from the existing built-in model config/metadata first, then overlay Token Plan-specific baseUrl/apiFormat/key/model aliases. Otherwise every capability like thinking/context/vision/tools can silently diverge.

2. The generation loading experience needs a frontend design pass.

The current loading UI in `app/generation-preview/page.tsx:1288-1380` is not really progressive disclosure: most state is represented by small dots plus one central visualizer. In the side-pane layout it also feels cramped to one side while leaving a lot of unused empty space, so the generation process does not feel polished or informative.

Could you run this through the `/frontend-design` skill and revise the layout? Ideally the loading screen should use the available pane width, progressively reveal completed/current/upcoming stages, and make the active step, status copy, and visualizer feel like one coherent loading experience rather than a single static card.

wyuc (commented):
I ran another cross-review pass on the latest head. A few issues still look actionable:

1. **[P1] Image/video probing treats almost any non-404 as model support**

   `app/api/provider/probe-chat-models/route.ts:86-88`

   For image/video probes, `verifyModel()` returns `res.status !== 404`. That means transient 500/502 responses, misrouted endpoints, and other unexpected non-auth server errors are recorded as "model supported". In the token-plan flow this can leave unusable media models in the generated config and mark the modality enabled, while the first real generation request fails.

   I think this should only accept known-good / expected invalid-request statuses, e.g. `2xx` and the specific `400` response that proves the model exists, and reject 5xx/unexpected statuses.

2. **[P2] Removing a token plan does not restore or switch away from plan provider state**

   `lib/config/apply-token-plan.ts:225-247`

   For LLMs, applying a plan overwrites the shared built-in provider in place (`baseUrl`, `models`, `name`, `icon`, `isBuiltIn`, etc.), but removal only clears `apiKey`. Because provider config updates are shallow merges, the built-in provider can remain pointed at the token-plan endpoint with plan-specific model IDs and `isBuiltIn: false`.

   For image/video/TTS/web-search, removal clears keys and disables providers but does not switch the currently selected provider away from the removed plan. Some generation paths still use the selected provider and can fail with an empty key until the user manually reselects a provider.

   Please restore the previous/built-in provider config on removal, and clear or replace active selections when the removed provider is currently selected.

3. **[P2] Probed model import loses provider-specific model metadata**

   `components/settings/index.tsx:417-422`

   The fetched-model merge only appends new IDs, so stale probed model IDs remain after a user changes the provider base URL/API key and fetches again.

   Also, this path calls `modelInfoFromId(id)` without passing `pid`. `modelInfoFromId()` only reattaches catalog thinking capability when the provider id is available, so fetched built-in provider models can lose `capabilities.thinking`, which hides the thinking control.

   Please replace old probe-derived models instead of only appending, while preserving explicit manual models, and call `modelInfoFromId(id, pid)`.

4. **[P3] Failed media tier verification can leave a disabled provider selected**

   `components/settings/token-plan-settings.tsx`

   The apply flow optimistically selects image/video providers before verification. If the probe finds no supported models and disables the provider, the active selected provider/model can still point at that rejected provider. The dashboard marks it unavailable, but the app state remains broken until the user manually switches.

5. **[P3] Usage totals can double-count cached input tokens**

   `app/api/usage/route.ts:53` and `app/api/usage/route.ts:83-84`

   For OpenAI-compatible usage, `prompt_tokens` already includes cached tokens, but the dashboard total adds `cacheReadTokens` again. This can overstate displayed totals for cache-hit requests.

6. **[P3] Doubao TTS concatenated-JSON parsing is fragile around braces in strings**

   `lib/audio/tts-providers.ts:919-952`

   The parser splits objects by brace depth without tracking string/escape state. Success chunks are likely fine because base64 has no braces, but error messages containing braces can corrupt parsing or hide the real error.

wyuc (dismissed):
Looks good from my side. The previous review findings are addressed.

Only remaining blocker is the merge conflict in `lib/audio/tts-providers.ts`. Please resolve it by keeping the shared `splitConcatenatedJsonObjects` helper from `lib/audio/json-stream.ts` while preserving the token-plan Doubao TTS auth/endpoint changes, then let CI rerun green and merge.

yanpgwang:
@wyuc conflict resolved (kept the shared `splitConcatenatedJsonObjects` from `lib/audio/json-stream.ts` and preserved the token-plan Doubao auth/endpoint changes). CI is green — ready to merge.

wyuc (approved):
Verified the conflict resolution on the latest head: tts-providers.ts keeps the shared splitConcatenatedJsonObjects from lib/audio/json-stream.ts (no duplicated local copy) and preserves the token-plan Doubao dual-auth/endpoint changes. CI is green on the head. Merging.

PR #785, opened 2026-06-24
fix(editor): stop dumping raw tool-failure blob in AI edit tool cards

wyuc:
## Problem

In the **Edit with AI** panel, a failed tool call renders the tool result's full text inline at the bottom of its `ToolCard`. When the failure is a framework schema-validation error (e.g. an `edit_interactive_html` edit missing `oldText`), that text is a large blob — the validation message **plus the entire received-arguments JSON, including the full interactive-page HTML** — which floods the panel with raw, unactionable content.

## Fix

Drop the inline failure body from `ToolCard`. The card is non-expandable by design (a single status row), so failure now surfaces only through the amber ✗ status mark and its hover tooltip; the model also restates the actionable reason in its own reply.

`failText` is kept as an accepted-but-ignored prop so existing call sites (`edit-interactive-html`, `regenerate-scene`, `regenerate`) compile unchanged.

Pure UI change — no change to the agent runtime or tool behavior.

### DIFF
--- a/components/edit/AgentPanel/edit-interactive-html-tool-ui.tsx
+++ b/components/edit/AgentPanel/edit-interactive-html-tool-ui.tsx
@@ -18,14 +18,12 @@ function EditInteractiveHtmlCard({
   stopped,
   failed,
   sceneId,
-  failText,
   toolCallId,
 }: {
   running: boolean;
   stopped: boolean;
   failed: boolean;
   sceneId?: string;
-  failText?: string;
   toolCallId: string;
 }) {
   const { t } = useI18n();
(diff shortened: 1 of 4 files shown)


wyuc:
## Verification & review

**Verified**
- Rendered the real `ToolCard` in a failed state with a representative oversized validation payload (the kind that previously flooded the panel). The failed card now renders as a single status row (✗ + tooltip) with no inline blob; a control still rendering the old inline body reproduces the original behavior.
- CI green: Analyze, Lint/Typecheck/Unit Tests, E2E.

**Review (two independent passes)**
- Final round: clean — no correctness/type issues. The `failText` prop and all its call-site wiring are fully removed and the doc comments updated to match.
- One design note, accepted as a deliberate trade-off: removing the inline body also drops the *short* backend failure reasons (e.g. missing scene context, edit-anchor "found N occurrences") from the card, not just the oversized schema-validation blob. This is intentional — the card is non-expandable by design; failure is signaled by the amber ✗ and its tooltip, and the model restates the actionable reason in its chat reply. Surfacing a *truncated* reason can be a follow-up if it proves needed.

PR #788, opened 2026-06-24
feat: auto-retry transient failures during scene generation

### LINKED ISSUE #755: [Feature]: Auto-retry failed scene generation with exponential backoff
### Problem or Motivation

When generating course content, any single scene generation failure immediately pauses the entire generation queue. Users must manually click "Retry" to continue, which breaks the flow of teaching - especially problematic given OpenMAIC's core advantage of asynchronous generation (users can teach while slides generate in the background).

Through testing with high-performance models like Zhipu's GLM-5.2, I've observed that generation errors occur randomly and frequently - often happening on nearly every generation attempt. These failures are typically transient (API rate limits, temporary timeouts, network hiccups) and resolve with a single retry. The issue can be even more pronounced with weaker or more cost-effective models that many users rely on.

The current "fail once, stop everything" approach undermines the async generation experience, forcing users to wait mid-lecture instead of just waiting for the first slide.

### Proposed Solution

Implement automatic retry logic similar to mature AI coding tools (Claude Code, Cursor, Codex):

1. **Auto-retry mechanism:**
   - Automatically retry 3-5 times for retryable errors (5xx, 429 rate limits, timeouts)
   - Use exponential backoff strategy (1s → 2s → 4s → 8s → 16s)
   - Distinguish non-retryable errors (401 auth, 400 bad request) - fail immediately for these

2. **Non-blocking feedback (optional):**
   - Show "Retrying (2/5)" status on the sidebar card
   - Display a dismissible banner at the top of sidebar after 3 failed attempts: "Some pages failed to generate, click to review"

This preserves the user's teaching flow while silently handling transient errors in the background.

### Area

Classroom generation

### Additional Context

If maintainers agree this is a reasonable direction, I'm willing to submit a PR implementing the auto-retry mechanism in the next few days :-)

YizukiAme:
## Summary

Scene generation (content, actions, TTS) can fail due to transient errors like rate limits (HTTP 429), server overload (503), timeouts, or network issues. Currently, any single failure immediately pauses the entire generation queue, forcing users to manually click "Retry" — breaking their teaching flow when they are watching a lesson while later scenes generate in the background.

This PR introduces automatic retry with exponential backoff for all scene generation API calls, so transient failures are silently resolved without user intervention.

## Problem

The async generation pipeline is one of OpenMAIC's best features — users only wait for the first slide and can start teaching immediately while remaining scenes generate in the background. However, with models like GLM 5.2 (and especially budget models), random API failures are frequent. When a failure occurs:

1. The generation queue **silently pauses** (`setGenerationStatus('paused')`)
2. The user continues teaching, unaware of the failure
3. When they navigate to the failed scene, they hit a **"Generation Failed" error screen** and must manually retry
4. This **breaks the teaching flow** and defeats the purpose of async generation

These failures are almost always transient — simply retrying the request succeeds. This is similar to how tools like Claude Code and Codex handle LLM API errors: retry silently before surfacing failures to the user.

## Solution

A shared retry helper (`withGenerationRetry`) wraps each API call with:
- Up to **5 retries** (6 total attempts)
- **Exponential backoff with jitter** (1s → 2s → 4s → 8s → 16s)
- **Smart error classification**: retries 429, 5xx, timeouts, network errors; fails fast on 401, 403, 404
- **AI SDK `RetryError` unwrapping**: correctly classifies errors wrapped by the AI SDK's internal retry mechanism
- **AbortError passthrough**: respects user-initiated cancellation

### New files
| File | Purpose |
|------|------|
| `lib/generation/generation-retry.ts` | Shared retry helper with error classification and exponential backoff |
| `tests/generation/scene-api-retry-boundary.test.ts` | Verifies API routes disable AI SDK inner retries |
| `tests/hooks/use-scene-generator-retry.test.ts` | Client-side retry wrapper tests |
| `tests/server/generation-retry.test.ts` | Retry helper unit tests |
| `tests/server/classroom-generation-retry.test.ts` | Server pipeline integration tests |

### Modified files
| File | Change |
|------|------|
| `lib/hooks/use-scene-generator.ts` | Wrap `fetchSceneContent`, `fetchSceneActions`, `generateAndStoreTTS` with retry; export functions for reuse |
| `lib/server/classroom-generation.ts` | Wrap server-side content/actions generation with retry |
| `app/api/generate/scene-content/route.ts` | Set `maxRetries: 0` to disable AI SDK built-in retries |
| `app/api/generate/scene-actions/route.ts` | Set `maxRetries: 0` to disable AI SDK built-in retries |
| `app/generation-preview/page.tsx` | Refactor to reuse exported fetch/TTS functions (eliminates ~100 lines of duplicated code) |

### Architecture

```
                  lib/generation/generation-retry.ts  ← shared retry helper
                         ↙                    ↘
      Server-side (batch)                Client-side (browser)
 classroom-generation.ts            use-scene-generator.ts
 ├─ content → withRetry             ├─ fetchSceneContent → withRetry
 └─ actions → withRetry             ├─ fetchSceneActions → withRetry
                                    └─ generateAndStoreTTS → withRetry
                                         ↓
                                 API Routes (maxRetries: 0)
                                 ├─ scene-content/route.ts
                                 └─ scene-actions/route.ts
```

Key design decision: retry is placed at the **HTTP fetch boundary**, not in the outer generation loop. This means:
- The outer `generateRemaining` loop's abort/epoch/failure logic is **completely unchanged**
- If all 6 attempts fail, the existing `addFailedOutline` + `pause` behavior kicks in as before
- Minimal invasiveness — easy to review and reason about

### Why disable AI SDK `maxRetries`?

The AI SDK's `generateText` has a default `maxRetries: 2` (3 total attempts). If we add our own 5 retries on top, a persistent failure would trigger `6 × 3 = 18` requests. Setting `maxRetries: 0` on the scene API routes ensures our retry helper is the single retry authority.

## Testing

- `vitest` targeted test files: **all tests passed**
- `prettier --check` on touched files: **passed**
- `eslint` on touched files: **passed**
- `tsc --noEmit`: **passed**
- `git diff --check`: **passed**

fix #755 

*Assisted by GPT-5.5 xhigh and audited by a human.*


wyuc:
Cross-reviewed this independently (two reviewers over the same `origin/main..HEAD` diff). The retry core looks solid — exponential backoff with jitter, `RetryError` unwrapping with a cycle guard, and `maxRetries: 0` is correctly threaded through `callLLM` → `generateText`, so there's no 6×3 retry multiplication. Three integration issues seem worth addressing before merge:

**1. Server scene calls lose the per-route thinking config (correctness)**
In `lib/server/classroom-generation.ts`, the existing `aiCall` passes `classroomThinking` as the 4th arg to `callLLM`, but the new `sceneAiCall` omits it. Since scene content/actions now route through `sceneAiCall`, any `generate-classroom` deployment with a configured `thinkingConfig` will silently fall back to the global thinking config for scene generation. Forwarding `classroomThinking` into `sceneAiCall` should restore parity.

**2. AbortError is swallowed into a failure result (correctness)**
In `lib/hooks/use-scene-generator.ts`, `fetchSceneContent` / `fetchSceneActions` wrap the retry in a try/catch that converts *every* error — including the `AbortError` that `withGenerationRetry` correctly rethrows — into `{ success: false }`. The `generation-preview` page then throws a plain `Error` that no longer hits its `AbortError` guard, so cancelling or navigating away during first-scene generation can surface a "generation failed" and clear the session instead of being treated as a cancel. Rethrowing abort errors before returning the failure result would fix this.

**3. Client-side fast-fail on 401/403 doesn't actually trigger**
`app/api/generate/scene-content/route.ts` (and scene-actions) collapse all LLM errors to HTTP 500. On the client, `isRetryableGenerationError` checks the status code first and treats any `>= 500` as retryable, short-circuiting before the message / `isRetryable` checks. So a permanent upstream failure (e.g. a bad key → 401) reaches the client as 500 and gets retried the full 6× (~31s of backoff) rather than failing fast. The fast-fail path only really works for the server pipeline, where the real `APICallError` status / `isRetryable` survives. Either propagate the upstream status from the route, or soften that claim in the description.

Minor:
- First (blocking) scene: on a persistently-failing endpoint the user now waits ~31s of cumulative backoff before the failure surfaces, vs. immediately before — a smaller retry budget for the foreground slide might be worth considering.
- `generateSceneActions` retry has no `shouldRetryResult`, unlike content's `result === null`, so a soft "empty actions" result won't retry (likely fine, just noting the asymmetry).

Overall nice work — placing the retry at the fetch boundary keeps the outer generation loop untouched, which makes the change easy to reason about.

YizukiAme:
Hi @wyuc, thanks for the careful review!! Agreed on all three.

1. Good catch on `classroomThinking`. `sceneAiCall` omitted the 4th `callLLM` argument while disabling `maxRetries`; now forwarded so server-side scene generation keeps honoring the `generate-classroom` thinking route.

2. Confirmed. Added a shared `isAbortError` helper and rethrow aborts from `fetchSceneContent` / `fetchSceneActions` before the `{ success: false }` fallback. Applied the same guard to the two TTS loops and migrated the existing outer `instanceof DOMException` checks to the helper so abort handling is consistent across runtimes.

3. Valid. Added `lib/server/llm-error-response.ts` which unwraps `APICallError` / `RetryError` and preserves the upstream status so the client classifier sees the real 401/403/429/5xx. User-facing messages are mapped to canned strings; raw upstream errors stay in the server logs.

On the minors:

- Reduced the foreground first-scene retry budget to 2 via a dedicated `FOREGROUND_SCENE_RETRY_OPTIONS` constant. Persistent failures now surface in ~3s instead of ~31s on the blocking path; background scenes keep the default budget.
- Leaving the `generateSceneActions` empty-result asymmetry as-is for now: supported scene types already fall back to default actions when parsed actions are empty, so this looks like a different class of issue from transient failure. Happy to revisit if you have a concrete case :-)

wyuc:
Re-reviewed the fix commit (independent two-reviewer pass over the updated diff). All three earlier points are addressed correctly:

- **1. thinking config** — `sceneAiCall` now forwards `classroomThinking` as the 4th `callLLM` arg. ✅
- **2. abort handling** — the shared `isAbortError` helper (robust across `Error` / `DOMException` / plain record) is rethrown ahead of the `{ success: false }` fallback in all three fetch wrappers and the TTS loops. ✅
- **3. upstream status** — `llmApiError` / `lib/server/llm-error-response.ts` unwraps `APICallError` / `RetryError` and preserves the real status, and the routes now return it instead of a blanket 500, so the client classifier sees the true 401/403/429/5xx. Mapping to canned messages while keeping raw errors server-side is a clean way to avoid leaking provider details. ✅

The foreground retry-budget reduction (`FOREGROUND_SCENE_RETRY_OPTIONS`, 2 retries) is wired into the blocking first-scene content/actions/TTS calls. 👍

One remaining item, same abort theme as #2 but on a different code path:

**[P2] Backoff `sleep` isn't abort-aware** — `lib/generation/generation-retry.ts`
`withGenerationRetry` only forwards the signal into `operation` (the fetch); the `await sleep(nextDelayMs)` between attempts is a bare `setTimeout` with no signal. So if a user hits Stop / navigates away *during a backoff delay* after a retryable error, the abort isn't observed until the sleep elapses and the next `operation()` runs — up to the max backoff (~16s on the default background budget; ~1–2s on the foreground budget). The rethrow fix handles abort *during the fetch*, but not *during the wait*. Suggest threading the `AbortSignal` into `withGenerationRetry` and making `sleep` reject on `abort` (attach an `'abort'` listener that clears the timer and rejects with an `AbortError`), so the loop's catch sees a non-retryable abort and exits immediately.

Everything else looks good — thanks for the quick turnaround.

YizukiAme:
Great catch, thanks! Threaded the `AbortSignal` through `withGenerationRetry` and made the default sleep abort-aware (clears the timer and rejects with `AbortError`). The three client wrappers pass their fetch signal in; `ClientRetryOptions` omits it so it can't be supplied externally.

Also added a `throwIfAborted` at the other yield points in the loop so a late abort isn't missed even if `operation` doesn't observe the signal, and the `catch` arm preserves a real `AbortError` rather than synthesising a new one.

Tests cover pre-aborted signal, custom-sleep abort threading, and the default sleep rejecting on abort without waiting for its timer :-)

wyuc:
All review findings are resolved and CI is green (E2E + Lint/Typecheck/Unit). The retry core, the three integration fixes (per-route thinking config forwarding, abort rethrow at the fetch wrappers, and upstream-status preservation for client classification), and the abort-aware backoff all look solid, with tests covering the key paths. Thanks for the thorough iterations — LGTM. Merging.

wyuc (approved):
LGTM — all findings addressed, CI green.

PR #789, opened 2026-06-24
fix(importer): port PPTX shape restoration hotfixes

xuyuanwei678:
## Summary

This PR ports the June 24 PPTX importer hotfixes from the MAIC-Mono comparison workspace into the `@openmaic` package set.

Key fixes include:
- Preserve rotated auto-fit text frames for quarter-turn text so vertical copyright and rotated labels are not clipped by viewport clamping.
- Add support for `textArchUp` / `textArchDown` preset text warps to improve curved text restoration.
- Preserve group-local filled custom geometries and `grpFill` inheritance so grouped shapes do not lose their fills.
- Improve chart restoration for Excel serial date labels and mixed line/area chart data.
- Adjust the renderer slide element wrapper positioning to support the imported text and shape layout fixes.

## Tests

- `PATH=/opt/homebrew/Cellar/nvm/0.39.5/versions/node/v20.20.2/bin:$PATH pnpm -C packages/@openmaic/importer test`
- `PATH=/opt/homebrew/Cellar/nvm/0.39.5/versions/node/v20.20.2/bin:$PATH pnpm -C packages/@openmaic/importer run build`
- `PATH=/opt/homebrew/Cellar/nvm/0.39.5/versions/node/v20.20.2/bin:$PATH pnpm -C packages/@openmaic/renderer run build`


wyuc (commented):
Slide renderer risk: the new `SlideElement` root is now absolutely positioned with `inset: 0`, so every `.slide-element` host covers the whole slide.

The visible geometry still lives in the nested `Base*Element`, but `onElementClick` is attached to the full-slide host. Any consumer that enables `onElementClick` or uses DOM hit-testing against `slide-element-*` can end up with the top z-index element receiving clicks even when the user clicks empty space or a different element. This is especially risky for slide editing / element-pick flows.

Can we keep the root geometry non-interactive, or move the click handler to the actual element box? Another option is `pointerEvents: "none"` on the full-slide wrapper and `pointerEvents: "auto"` on the nested rendered element; the key requirement is that the interactive target should match the element visual bounds.

wyuc:
Did a closer follow-up pass on the importer/renderer changes. One confirmed bug plus a few smaller notes.

## Bug

**`frameHeightPx` is computed in points but consumed as pixels** — `packages/@openmaic/importer/src/serializer/shapeSerializer.ts:1237`

```ts
const frameHeightPx = node.size.h > 0 ? pxToPt(node.size.h) : undefined;
```

`node.size` is in px — the sibling `frameWidthPx = node.size.w` on the line above is passed through unconverted and later compared against pixel margins. Applying `pxToPt` (×0.75) here yields a points value despite the `Px` name. It's then consumed as pixels in `bodyTextHeightPx` (`textSerializer.ts`), which subtracts `emuToPx(tIns + bIns)` and emits `height: ${...}px` for the warp container. Net effect: a 100px-tall frame produces a ~55px arch container instead of ~80px (~30% short), so `textArchUp` / `textArchDown` text is vertically compressed and no longer matches the PPT frame box.

Suggested fix: drop the conversion — `const frameHeightPx = node.size.h > 0 ? node.size.h : undefined;`

## Smaller notes

- **Empty date category labels render as `12月30日`** — `chartSerializer.ts:78-79`. When a date-formatted numeric cache has an empty/whitespace `<c:v/>`, `Number(raw)` is `0` (and `Number.isFinite(0)` is `true`), so the serial-date path emits `12月30日` instead of keeping the cell blank. Guard `raw.trim() === ''` before converting.
- **Warp path skips hyperlink / `fld` handling** — `renderTextWarp` in `textSerializer.ts` resolves `sym` symbol fonts (correctly) but, unlike the non-warp loop, doesn't wrap `hlinkClick` runs in `<a href>` or resolve `<a:fld type="slidenum">`. Minor parity gap for arch titles that are links or slide-number fields.
- **`flatExtent` now also requires `!hasPresetGeometry && !hasAreaFill`** — `shapeSerializer.ts:828-832`. The group-local fill fix is right, but a standalone sub-pixel-height *filled* custom-geometry divider (or a sub-pixel preset rect) is no longer bumped to a visible 1px and could render invisible. Low risk; a regression test on filled sub-pixel dividers would lock it down.

The mixed line+area handling (area series replaced by the line series) reads as intentional per the added test, so noting it as a fidelity limitation rather than a bug. The `SlideElement` hit-target change looks correct — all `Base*Element` roots are absolutely positioned, so the interactive region collapses to the element's visual bounds as intended, and clicks on empty space fall through the `pointer-events: none` root. Thanks for addressing the earlier hit-target concern.

xuyuanwei678:
Follow-up on the importer review:

I re-checked the reported `frameHeightPx` issue against the importer code path. The review is correct: `node.size.h` is already in pixels, but we were converting it with `pxToPt()` before passing it as `frameHeightPx`. That value is later consumed by `bodyTextHeightPx()` as pixels and emitted into a CSS `height: ...px`, so the text warp container was shorter than the original PPT frame. I fixed this by keeping `frameHeightPx` in px, matching `frameWidthPx`.

I also addressed the blank date-category case in `chartSerializer`. Date-formatted numeric cache values now return early when `raw.trim() === ''`, so empty labels stay blank instead of becoming Excel serial `0` / `12月30日`.

Added regression coverage for both cases:
- `textArchUp` / `textArchDown` now verifies a 100px frame produces the expected warp container height after text insets.
- Blank date-formatted category labels remain blank.

Validation run locally:
- `pnpm -C packages/@openmaic/importer exec vitest run test/shapeSerializer.groupLocal.test.ts test/chartSerializer.test.ts`
- `pnpm -C packages/@openmaic/importer test`
- `pnpm -C packages/@openmaic/importer run build`
- `pnpm check`
- `npx tsc --noEmit`

I intentionally limited this patch to the confirmed unit bug and the blank-date edge case. The hyperlink/fld warp parity and filled sub-pixel divider note look reasonable as follow-up hardening items, but I left them out of this pass to avoid expanding the scope of this PR.

wyuc (approved):
Both reported issues are fixed and verified in dcda521: `frameHeightPx` now stays in px (matching `frameWidthPx`), and blank date-formatted category labels return early instead of becoming Excel serial 0. Regression tests cover both. CI green. The hyperlink/fld warp parity and filled sub-pixel divider notes are reasonable follow-up items. LGTM.

PR #796, opened 2026-06-25
Refactor widget actions into the scene action pipeline

### LINKED ISSUE #794: [Refactor] Align widget interaction actions with slide actions; drop the teacherActions → Action compile layer
## Background

Interactive (widget) scenes currently model teacher interactions in **two forms**, with a compile step between them:

- **Authoring form** — `TeacherAction[]` (`lib/types/widgets.ts`), generated by a dedicated LLM call (`generateWidgetTeacherActions` in `lib/generation/scene-generator.ts`) and stored on `InteractiveContent.teacherActions`. Types: `speech | highlight | annotation | reveal | setState`. Each entry bundles a visual op **and** its narration (`content`), plus a UI `label`.
- **Runtime form** — the flat `Action[]` stream (`scene.actions`), consumed by `lib/action/engine.ts` and `lib/playback/engine.ts`. Widget verbs: `widget_highlight | widget_setState | widget_annotation | widget_reveal`.

`convertTeacherActionsToActions` (`scene-generator.ts`) lowers the first into the second **at generation time**, by:
1. splitting each bundled "do + say" `TeacherAction` into a visual `widget_*` action + a separate `speech` action (engine runs them differently: visual is non-blocking, speech is blocking TTS);
2. namespacing `highlight -> widget_highlight`, etc.;
3. reissuing collision-free `nanoid` IDs.

## Problem

By contrast, **slide** scenes have no such layer: `generateSceneActions` emits the final flat `Action[]` directly (e.g. `spotlight` + `speech` as separate entries), persisted as-is, and the engine routes by `type` at playback. No intermediate form, no compile.

The widget path therefore carries avoidable complexity:

- **Dual persistence.** Both `InteractiveContent.teacherActions` (source) and `scene.actions` (compiled) are serialized into the document. At playback only the flat stream is read — `teacherActions` is dead weight there (kept only for regeneration). This is a source + derived-cache duplication with no guard keeping them consistent.
- **Playback semantics baked at generation time.** The "split do+say" decision is an execution concern (it exists because TTS blocks and visuals don't), yet it's frozen into the persisted document by a generation-time transform instead of being a runtime lowering.
- **A parallel action vocabulary** (`TeacherAction` vs `Action`) that differs only by namespacing + splitting, kept in sync by hand.

Note the iframe-target resolution is **already** the engine's job — `executeWidgetHighlight` etc. `postMessage` the selector into the iframe. The compile layer does not touch addressing; it only reshapes/duplicates.

## Proposal

Align widget actions with the slide model:

- The post-HTML LLM call emits the **flat `Action[]` directly** (`widget_highlight` / `widget_setState` / `widget_annotation` / `widget_reveal` / `speech`), instead of `TeacherAction[]`.
- **Remove** `InteractiveContent.teacherActions` (no longer persisted), `convertTeacherActionsToActions`, and the `TeacherAction` type.
- Execution-time routing stays entirely in the engine (unchanged): `spotlight` resolves against the canvas; `widget_*` forwards to the iframe.

Result: a single persisted `Action` stream for every scene type, with the engine as the sole owner of execution semantics.

### What inherently stays

- **Two-stage generation.** Widget actions must reference real selectors in the generated widget HTML, so the actions are still produced by a second LLM call after the HTML exists. Only its **output schema** changes (flat `Action[]`), and the convert step disappears.
- **Engine iframe routing** (`executeWidget*` -> `postMessage`).

### Decisions to make explicitly

1. **Prompt reschema** — the widget-actions prompt outputs the flat `Action` schema (separate visual + `speech` entries) instead of the bundled `TeacherAction` schema. This is the bulk of the work.
2. **Loss of the "one teacher gesture = N actions" grouping.** Slides already operate without it; accept it, or carry a soft `groupId` if an edit UI needs it.
3. **Migration** — a forward migration drops `teacherActions` from existing persisted classrooms (already inert at playback).

## Relationship to #787

This unblocks and simplifies **#787 Part A**. Once widgets emit one flat stream, the question "what enters `@openmaic/dsl`" collapses cleanly: the flat `Action` union — including the now-stable `widget_*` verbs — is the single asset, with no generation-time `teacherActions`/compile coupling to reason about. Recommended ordering: land this app-side refactor first, then bring the widget verbs into the contract under #787.

## Acceptance

- [ ] widget scene generation emits flat `Action[]` directly; `convertTeacherActionsToActions` removed
- [ ] `TeacherAction` type and `InteractiveContent.teacherActions` removed; only `scene.actions` is persisted
- [ ] playback / rendering behavior for widget scenes unchanged (engine routing untouched)
- [ ] widget-actions prompt updated to the flat schema; generation tests updated
- [ ] forward migration drops `teacherActions` from existing documents
- [ ] no regression in the slide / quiz action paths

wyuc:
## Summary

- Remove the widget `teacherActions` generation and conversion layer so interactive widgets generate final `Action[]` entries directly through the scene action pipeline.
- Keep all four widget action types: `widget_highlight`, `widget_setState`, `widget_annotation`, and `widget_reveal`.
- Forward widget action `content` through the iframe payload so highlight, annotation, setState, and reveal can expose local helper text.
- Move access-code token helpers out of the Next route module so route exports remain compatible with production type checking.

Closes #794

## Test Plan

- `pnpm exec vitest run`
- `pnpm exec tsc --noEmit --pretty false`
- `pnpm lint`
- `NEXT_TELEMETRY_DISABLED=1 NODE_OPTIONS=--max-old-space-size=4096 pnpm exec next build --webpack`


cosarah:
Findings:

1. The direct widget-action prompt now only documents stable selector guidance for procedural-skill and simulation widgets. Since generateSceneActions passes only widgetType/widgetConfig and not the generated HTML selectors, diagram/code/game/visualization3d actions can easily target selectors that do not exist, causing widget_highlight/annotation/reveal/setState to silently no-op during playback.

2. The forward migration removes only content.teacherActions. Legacy content.widgetConfig.teacherActions can still remain serialized in existing classrooms, so the dead authoring layer is not fully stripped from persisted widget documents.

cosarah:
Update after latest head ca126a9: the nested widgetConfig.teacherActions migration is fixed.

Remaining finding: the direct widget-action prompt still only documents stable selector guidance for procedural-skill and simulation widgets. For diagram/code/game/visualization3d, generated widget actions may target selectors that do not exist and silently no-op during playback.

wyuc:
Thanks. On the remaining finding: the action generator grounds selectors through the shared widget contract rather than the raw HTML. Each widget's content prompt emits canonical IDs (simulation / visualization3d / procedural-skill embed a `widget-config` script and mandate stable IDs like `#feedback-panel`, `#zoom-in-btn`, `#{key}-slider`), and `widgetConfig` — which the action prompt receives — carries the addressable handles (diagram's `revealOrder` / `nodes[].id`, game element IDs, etc.). The simulation widget also resolves `HIGHLIGHT_ELEMENT` / `ANNOTATE_ELEMENT` generically via `document.querySelector(target)`, and `widget_setState` is config-key based, so playback grounds correctly without passing the HTML — which matches what we see end-to-end.

The fair point is that the action prompt only spelled out the convention for procedural-skill and simulation. `603232b3` now documents the canonical selectors the diagram / game / visualization3d content prompts actually emit, plus a rule to use `widget_setState` or a speech-only beat instead of guessing a `target` when no stable selector is known (e.g. free-form code widgets). That keeps generation grounded for all widget types without bloating the prompt with the full HTML.

cosarah (approved):
LGTM

PR #801, opened 2026-06-26
feat(agent-edit): multi-session conversation history for the AI editor

wyuc:
## Summary

The "Edit with AI" panel (`AgentPanel`) keeps only **one** conversation thread per stage in `localStorage` (`agent-thread-store.ts`), and the "New conversation" button **deletes** it. Once you start a new conversation the previous one is gone with no way back.

This PR turns the single thread per stage into a **per-stage session list** backed by IndexedDB (Dexie), so past conversations are retained and switchable.

## Changes

- **Dexie v11 → v12**: new `agentEditSessions` table (`id, stageId, [stageId+updatedAt]`). Messages keep the existing slim `serialize-thread` projection.
- **`agent-thread-store.ts`** rewritten from a single-thread localStorage store into a Dexie-backed multi-session store: `createSession` / `saveSession` / `loadSession` / `listSessions` / `deleteSession`, plus a one-time `migrateLegacyThread` that imports the old `maic-agent-threads` localStorage entry and drops it.
- **"New conversation"** now **archives** the current session and starts a fresh one instead of deleting; the old session stays in history.
- **History popover** in the panel header lists the stage's sessions (auto-titled from the first user message, current one highlighted) — click to switch, hover to delete.
- **Active session is remembered per stage** (small localStorage pointer) so a refresh restores the session you had open, including a just-cleared empty one (the clean slate survives reload).
- **Correctness hardening**: `saveSession` preserves the original `createdAt` and skips tombstoned ids (a delete wins over an in-flight save); switching stages drops the previous thread synchronously so it can't be sent as history or archived under the new stage; deleting a course also removes its `agentEditSessions` rows.
- **Soft cap** of 30 sessions per stage (oldest pruned); empty sessions are never written.
- **i18n** keys added across all 8 locales.

Pure client-side (IndexedDB) — no backend changes.

## Tests

- `deriveSessionTitle` unit tests (truncation, fallback, whitespace).
- Session store unit tests: create / save / load / list ordering / delete / prune cap / legacy migration idempotency / createdAt preservation / delete-tombstone-no-resurrect / active-id remember-recall.
- Full agent test suite green; `tsc --noEmit`, `eslint`, and `check:i18n-keys` pass.

## Notes

The `agent-thread-store.ts` design comment previously said history/rename/server-sync was "intentionally out of scope"; this PR implements the history list. Rename and server sync remain out of scope.


sophietao20-star (dismissed):
End-to-end testing is completed, and no blocking issue was found.

Two minor suggestions for possible future iterations (non-blocking for v0):

- Generate a short summary for each session title instead of using the first prompt verbatim.
- Stop and flush the current session before switching to another one(archive partial output as cancelled, same as Stop)

PR #806, opened 2026-06-26
Polish interactive simulation framing and iframe layout guards

williamfaulconer26:
## Summary
- harden sandboxed interactive iframes with runtime/storage shims so generated simulations keep rendering under a safer sandbox
- add bounded simulation shell styling to prevent overlap, clipping, and horizontal overflow in classroom embeds
- cover the iframe patch with a focused unit test and an interactive classroom layout e2e scenario

## Verification
- pnpm.cmd exec vitest run tests/utils/iframe.test.ts
- direct local classroom probe against the seeded interactive stage confirmed wrapped badges, no horizontal overflow, and clipped 22px stage framing

## Notes
- the committed e2e test now primes the classroom route first so OpenMAIC creates its IndexedDB schema before the stage seed runs

### DIFF
--- a/components/scene-renderers/interactive-renderer.tsx
+++ b/components/scene-renderers/interactive-renderer.tsx
@@ -9,6 +9,79 @@ interface InteractiveRendererProps {
   readonly sceneId: string;
 }
 
+/**
+ * In-memory localStorage/sessionStorage shim, injected as the FIRST thing in the
+ * document so the page's own scripts see working storage.
+ *
+ * The interactive iframe is sandboxed `allow-scripts` WITHOUT `allow-same-origin`
+ * (intentional — combining them negates the sandbox for LLM-authored HTML). In a
+ * null-origin document, touching `window.localStorage` throws a SecurityError;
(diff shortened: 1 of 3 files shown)


wyuc:
This focused iframe/simulation work still appears independently valuable, so I am not closing it with #807. The branch now has real conflicts with current `main`, however, and its sandbox/runtime/storage shim changes sit on a security boundary. Please rebase and resolve the conflicts on your branch, keep the scope limited to these three files, and rerun the focused unit test plus the real interactive-layout E2E on the exact final SHA. Re-request review after that; we will review both rendering behavior and sandbox/storage isolation.

wyuc:
Thanks @williamfaulconer26, and sorry for the long wait. Hardening the sandboxed interactive iframes is something we'd like to take. The branch is from June and `main` has changed a lot around interactive scenes since, so could you rebase onto the latest `main`? Once it's rebased we'll review it properly — in particular the runtime/storage shims, since they sit on a security boundary.

Issue #815, opened 2026-06-29
[Feature] In-editor authoring of classroom agents (persona / role / avatar + per-scene multi-agent)


wyuc:
## Background

The Pro-mode course editor currently registers scene editors only for `slide`
and `quiz` surfaces (`lib/edit/scene-editor-registry.ts`; surfaces under
`components/edit/surfaces/`). The AI side panel
(`components/edit/AgentPanel/`) is a content-authoring assistant — it helps
revise the **slide content/narration** through chat. It does not edit the
classroom agents (the AI roles that teach/host the lesson) themselves.

Meanwhile the DSL already models agents as first-class course data:

- `GeneratedAgentConfig { id, name, role, persona, avatar, color, priority }`
  — `packages/@openmaic/dsl/src/stage.ts`
- `MultiAgentConfig { enabled, agentIds[], directorPrompt? }`
  — `packages/@openmaic/dsl/src/stage.ts`
- `stage.agentIds` / `stage.generatedAgentConfigs` — `lib/types/stage.ts`
- `scene.multiAgent` — DSL Scene interface

## Problem

Agent personas are fixed at generation time (LLM-authored in
`lib/server/classroom-generation.ts`). After a course is generated, an author
has **no way in the editor** to:

1. Preview or edit any field of `stage.generatedAgentConfigs`
   (name / role / persona / avatar / color).
2. Configure `scene.multiAgent` (enable a discussion, pick participating
   `agentIds`, set a `directorPrompt`) on a per-scene basis.

The global `components/settings/agent-settings.tsx` panel selects agents for
playback but does not edit the agent config carried by the course `stage`, so
it can't fill this gap.

## Proposal

Add agent authoring as a first-class part of the editor, reusing the existing
data contract and the editor's surface-registry pattern.

**Stage-level — edit the agent roster**
- A panel that lists `stage.generatedAgentConfigs` as editable cards.
- Edit per agent: `name`, `role`, `persona`, `avatar`, `color`.
- Add / remove / reorder agents; writes back to the course `stage` and
  persists like other editor edits (undo/redo aware).

**Scene-level — per-scene multi-agent**
- On a scene, toggle `multiAgent.enabled`, choose `agentIds` from the stage
  roster, and optionally set `directorPrompt`.

## Out of scope (for this issue)
- Voice/TTS authoring and avatar asset upload pipelines.
- Changing the agent runtime / director selection logic.

## Open questions
[...]

PR #816, opened 2026-06-29
feat(editor): in-editor authoring of classroom agents (Stage-level roster)

wyuc:
## Summary

Adds an in-editor **Agents** view to the Pro-mode course editor so authors can edit
the course's agent roster (the classroom AI roles) directly, instead of the roster
being fixed at generation time. Implements the Stage-level part of #815.

From a new **Agents** tab in the editor's right sidebar (beside *Edit with AI*), authors can:

- add / remove / reorder agents in the roster
- edit each agent's **name, role (teacher / assistant / student), avatar, persona**
- `role` drives `priority`; `color`/`avatar` are auto-assigned and cycle through the
  existing `AGENT_COLOR_PALETTE` / `AGENT_DEFAULT_AVATARS` constants
- a "keep at least one teacher" guard disables deleting / role-changing the last teacher

The roster is the existing DSL `GeneratedAgentConfig[]` on the stage — no new DSL
fields are introduced.

## How it works

- `lib/edit/agent-ops.ts` — pure reducer for the roster (`agent.add/update/delete/reorder`)
  with an immer-based undo/redo history (mirrors `lib/edit/slide-ops.ts`). Consecutive
  same-field edits coalesce into one history entry.
- `lib/edit/agent-roster.ts` — `materializeRoster()` resolves the editable roster from
  `stage.generatedAgentConfigs`, else from preset `agentIds`, else a default teacher.
- `components/edit/RightRailTabs.tsx` — the editor's right rail is a tabbed panel with
  **Edit with AI** and **Agents** tabs (single `<aside>` owning resize/collapse; the AI
  panel renders body-only via a `naked` prop so there are no stacked headers). Both tabs
  stay mounted so the AI thread survives switching.
- `components/edit/AgentsView/AgentRosterPanel.tsx` — the Agents tab: a narrow vertical
  agent list (up/down reorder, last-teacher delete guard, add) plus an inline
  `AgentInspector`; reuses the `useAgentRoster` controller hook.
- Store/persistence — `setStageAgents` plus a dedicated debounced sync that writes the
  roster to `db.generatedAgents` + the agent registry **only on an actual roster edit**
  (not on tab-open, scene changes, or other saves).
- The slide canvas stays visible while editing agents; the existing slide/quiz editing
  path is untouched.

Scope is intentionally narrow: **Stage-level roster authoring only**. Per-scene
`multiAgent` orchestration, voice/TTS authoring, and avatar upload are out of scope here.

## Testing

- Unit tests for the reducer, materialization, and store (vitest). Full suite green.
- Manual end-to-end in a browser: enter Pro mode → Agents, add / edit / delete / reorder,
  confirmed edits persist across reload (to both the stage snapshot and `db.generatedAgents`)
  and that deletions are reconciled out of the registry.

## Open questions for maintainers (why this is a Draft)

Independent review surfaced a few agent-identity / reference questions that touch runtime
behaviour and would benefit from maintainer input before this is finalized:

1. **Agent id identity on edit.** When a course's roster comes from preset `agentIds`
   (e.g. `default-1`), materialization currently re-mints fresh stage-scoped ids so that
   persisting an edit can't overwrite the global built-in default agents. The trade-off:
   any existing references to the old ids (e.g. a scene's `multiAgent.agentIds` /
   discussion `action.agentId`, or the current playback selection) are **not yet
   reconciled** to the new ids. What's the preferred contract here — preserve ids and
   special-case the global defaults, or re-mint and rewrite references?
2. **Playback selection.** Roster edits are not propagated to
   `useSettingsStore.selectedAgentIds`, so same-session playback reflects add/reorder only
   after a reload. Should editor edits update the active selection directly?
3. **Voice metadata.** The DSL `GeneratedAgentConfig` has no voice fields, so editing the
   roster does not carry `voiceConfig` / `voiceDesign` through. Should the roster preserve
   voice metadata as passthrough, or is voice authoring a separate surface?

Happy to iterate on these based on the preferred direction.

Addresses #815 (Stage-level scope; per-scene `multiAgent` authoring tracked separately).


wyuc:
## Independent review pass — done, findings addressed

Ran an independent review over the branch and fixed everything it surfaced:

- **Roster controller state sync** — removed a stale-closure / local-draft desync so undo/redo and fast (blur-level) edits no longer drop or silently revert changes; undo now steps per committed edit.
- **Agent-id handling** — materialization re-mints fresh ids only for *global default* presets and keeps already-stage-scoped generated ids, so editing a default-preset course no longer overwrites the built-in default agents while existing generated agents keep their ids.
- **Persistence + playback** — roster edits sync to `db.generatedAgents` + the agent registry only on a real edit (not on tab-open or scene advance), and the active playback selection now follows the edited roster, so rename / persona / avatar changes take effect without a reload.
- **Cleanup** — removed dead plumbing from the previous layout (the old view toggle slot, unused hook surface) and re-wired the "Edit with AI" tab visibility gate.

Checks: `tsc --noEmit` clean, unit tests green (1793), lint clean. The editor flow (add / edit / reorder / remove / change avatar / persist-across-reload, plus the accordion + pinned-teacher UI) was verified manually in a browser.

**Out of scope by design** (happy to add on request): per-agent voice authoring (the DSL `GeneratedAgentConfig` has no voice field) and per-scene participation.

Marking ready for review — feedback on the open questions in the description is very welcome.

sophietao20-star (dismissed):
Completed E2E testing, no blocking issues. Just one minor suggestion: default agent names are hardcoded in Chinese while personas are hardcoded in English. It might be worth localizing both before release.

PR #811, opened 2026-06-29
feat(dsl): bring the Action playback verbs into @openmaic/dsl (#787 Part A)

wyuc:
Part A of #787 (Phase 2, tracked under #720): promote the **Action** playback contract into `@openmaic/dsl`, keeping the package zero-runtime-dependency.

## What changed

- **New `packages/@openmaic/dsl/src/action.ts`** — the full Action contract moved out of the app: the `Action` union + all variants (`spotlight`, `laser`, `speech`, the `Wb*` whiteboard family, `play_video`, `discussion`, and the `widget_*` interaction actions), `ActionType`, the `FIRE_AND_FORGET_ACTIONS` / `SLIDE_ONLY_ACTIONS` / `SYNC_ACTIONS` category lists, and the `PercentageGeometry` overlay type. Zero imports → the zero-runtime-dep invariant holds.
- **`src/stage.ts`** — `Scene<TAction>` now **defaults to the standard `Action` union** (was `never`). Skeleton-only consumers (renderers / importers that reject actions) opt out explicitly with `Scene<never, …>`.
- **`src/index.ts`** — re-exports `./action.js`; docs updated so the dependency topology now names the runtime/action layer.
- **`lib/types/action.ts`** — collapses from ~330 lines to a **pure re-export shim** from `@openmaic/dsl` (same pattern as `lib/types/stage.ts`), so all existing `@/lib/types/action` call sites keep working unchanged.
- **README** — module table gains the `action.ts` row; roadmap + Stage/Scene split section updated.

## Scope note — widget actions

The original #787 plan kept the `widget_*` actions app-side (injected via `Scene<TAction>`). After the scene-action unification (#796) those actions are **config-free playback verbs** (`target` selectors / `Record<string, unknown>` state / `content`) with no coupling to widget *configs*. So they graduate into the contract alongside the other generic verbs, and the standard `Action` union now covers them. Widget *configs* (`WidgetType` / `WidgetConfig`) and `PBLProjectConfig` stay app-side as before.

No discriminant guards / validators in this PR — those land with **Part B** (JSON Schema + pure validators + migration registry).

## Verification

- `@openmaic/dsl`: `tsc --noEmit` clean · builds · **vitest 11/11 pass** (incl. updated `Scene` default-flip assertion + new `Scene<never>` opt-out case).
[...]

Issue #787, opened 2026-06-24
[Task] @openmaic/dsl Phase 2: bring the Action DSL into the contract + validators & migration registry


wyuc:
> Breaks out the three still-open **Phase 2 — Grow the DSL** items from #720 into one actionable task: bring the Action DSL into the contract, add JSON Schema + pure validators, and activate the migration registry. (The first Phase 2 box — promoting `Stage` / `Scene` / `SceneContent` — already landed in `@openmaic/dsl`.)

## Current state

- `@openmaic/dsl` owns the object model: `slides` + the `Stage` / `Scene<TAction, TContent>` skeleton + `SlideContent` / `QuizContent` + guards, with `version.ts` a stub. Zero runtime deps.
- The runtime verbs are still outside the contract: `Action` lives app-side in `lib/types/action.ts` (~286 LOC, ~22 types) + `lib/action/engine.ts`. `Scene<TAction = never>` defaults to no actions — a consumer of only `@openmaic/dsl` can describe what a lesson *looks like* but not how it *plays back*.
- No runtime validation (TS types + guards only), despite the "(future) JSON Schema, pure validators" note in `src/index.ts`.
- `version.ts` exports `DSL_MIGRATIONS = []` (stub); `DSL_VERSION` is defined but unused.

## Part A — Bring the Action DSL into the contract

New `@openmaic/dsl/src/action.ts`, reusing the existing "skeleton in, feature surface out" split (slide/quiz in the contract; interactive/pbl injected by the app via `Scene`'s generics).

Suggested partition (for review):

- **Into `@openmaic/dsl` — generic playback primitives:** `ActionBase`, `ActionType`, `PercentageGeometry`, `SpotlightAction`, `LaserAction`, `SpeechAction`, the whiteboard `Wb*` family (`WbOpen`, `WbDrawText`, `WbDrawShape`, `WbDrawChart`, `WbDrawLatex`, `WbDrawTable`, `WbDrawLine`, `WbDrawCode`, `WbEditCode`, `WbClear`, `WbDelete`, `WbClose`), `PlayVideoAction`, `DiscussionAction`
- **Stay app-side — couple to widget configs, injected via `Scene<TAction>`:** `WidgetHighlightAction`, `WidgetSetStateAction`, `WidgetAnnotationAction`, `WidgetRevealAction`

Changes:

- new `src/action.ts` exporting the generic `Action` union + discriminant guards
[...]

PR #814, opened 2026-06-29
fix(editor): keep emptied/zero-action scenes playable, bind outline by stable id, surface meaningless content

wyuc:
Editor empty-state robustness: fix silent slide loss at the playback root, bind a scene's outline by stable id, and surface incomplete content calmly. Five commits.

## 1. `fix(agent-edit)` — resolve a scene's outline by stable id, not mutable order
The editor agent matched an outline to a scene by `outline.order === scene.order`, so after a Pro-mode insert / reorder / delete (which rebalances `scene.order`) it attached **another slide's** outline to the scene being edited. Scenes now carry the stable `outlineId` of the outline they were built from (an app-layer annotation — not part of the `@openmaic/dsl` Scene contract); `resolveSceneOutline` matches by it, falling back to a scene-derived outline for inserted/duplicated/legacy scenes. `allOutlines` (the page list handed to the regenerate tools) is also rebuilt in current scene order, and a duplicated scene clears its inherited `outlineId`.

## 2. `fix(playback)` — dwell on zero-action scenes instead of skipping them
A scene with no actions was skipped, so an emptied / freshly inserted blank slide silently dropped from playback (and forced a meaningless seeded empty-speech clip to stay visible). `resolvePlaybackCursor` now yields one synthetic dwell beat for a zero-action scene, and `PlaybackChromeRoot` builds the engine for a zero-action **slide** scene (non-slide scenes are unchanged) so it actually dwells and auto-play advances past it. `createBlankSlideScene` no longer seeds a fake empty speech.

## 3. `fix(edit)` — clear the cue preview when its glyph unmounts (e.g. on delete)
Hovering a spotlight/laser cue replays it on the canvas; deleting the hovered cue unmounted its glyph without firing `onMouseLeave`, leaving the effect stuck on the slide. Both cue glyphs now clear the preview on unmount.

## 4. `feat(edit)` + `refine(edit)` — surface incomplete content, calmly
`content-validation.ts` (pure, tested) flags incomplete scenes/outlines. Incomplete timeline clips get a soft amber **dashed frame**; a blank outline title gets a small amber **dot**; the slide rail shows an amber dot per incomplete scene; fields keep their normal gray placeholder. The outline → generate gate is a neutral **"N / M sections ready"** counter + meter that jumps to the first blank title (no red error chip). Color stays out of the way of the blue interactive controls. New i18n keys across all 8 locales.

## Review
[...]

PR #817, opened 2026-06-30
feat(dsl): JSON Schema artifacts + pure validators (#787 Part B-1)

wyuc:
Implements **Part B-1** of #787 (Phase 2 — grow the DSL): the validators + JSON Schema half. Builds on #811 (Part A — Action verbs in the contract). The migration-registry half (`version.ts` activation) is a separate follow-up, **Part B-2**.

`@openmaic/dsl` shipped TS types + type-guards only — compile-time safety, nothing enforceable at runtime or across language boundaries. This PR makes the contract enforceable while keeping the **zero-runtime-dependency** invariant intact (new packages are `devDependencies` only).

## What's here

**1. Build-time JSON Schema artifacts** — `ts-json-schema-generator` (devDep, build-only) emits `dist/schema/{stage,scene,action}.schema.json` from the TS types (the single source of truth), shipped under a new `./schema/*` export. Serves non-TS / bring-your-own-validator consumers.
- Roots: `Stage`, `Action`, and `SerializedScene` (a concrete `Scene<Action, SceneContent>` — JSON Schema isn't generic, so a concrete instantiation is needed; kept internal via `src/schema-roots.ts`, not re-exported).

**2. Pure validators** — `validateStage` / `validateScene` / `validateAction` in `src/validate.ts`: hand-written, zero-dependency, fail-loud structural checks layered on the existing guards. They verify discriminants, required fields, and known nested discriminants — a gate for untrusted input (LLM/agent output) and persistence boundaries. Exhaustive per-field validation is delegated to the JSON Schema. `ValidationResult` collects every issue with a JSON-pointer-ish path rather than failing on the first.

**3. Supporting** — a frozen `ACTION_TYPES` set + `isActionType` guard (mirrors `PPT_ELEMENT_TYPES` / `isPPTElementType`); README "Runtime layer" section documenting both consumption modes; minor version bump `0.0.2 → 0.1.0`.

## Tests

- `test/validate.test.ts` — good/bad fixtures through each validator, asserting the offending paths.
- `test/schema.test.ts` — compiles the generated schema with `ajv` (devDep) in-memory and validates fixtures, proving the artifact is real and that `vitest run` stays self-contained on a clean checkout.

`pnpm --filter @openmaic/dsl build && typecheck && test` all green (26 tests).

## Acceptance (this PR — the schema/validator boxes of #787)

- [x] JSON Schema artifacts exposed + pure validators, still zero-runtime-dep
- [x] `test/` covers the action guards and validators
- [x] README updated with the runtime layer + dependency topology

Deferred to **Part B-2**: `DSL_MIGRATIONS` first migration end-to-end + `version.ts` no longer a stub (the `SlideContent.schemaVersion` content-level precedent will be reconciled there).

## Notes for review

- Zero runtime deps preserved: `package.json` has no `dependencies`; generator + ajv are devDependencies.
- The generator parses TS source directly, so codegen works pre-build and the schema test needs no compiled `dist`.

### DIFF
--- a/lib/api/stage-api-scene.ts
+++ b/lib/api/stage-api-scene.ts
@@ -4,7 +4,7 @@
  * Factory function that creates the scene namespace of the Stage API.
  */
 
-import type { Scene, SceneContent } from '@/lib/types/stage';
+import { makeScene, type Scene, type SceneContent } from '@/lib/types/stage';
 import type { StageStore, APIResult, CreateSceneParams } from './stage-api-types';
 import { generateId, validateSceneId, getScene, createDefaultContent } from './stage-api-defaults';
 
@@ -56,17 +56,18 @@ export function createSceneAPI(store: StageStore) {
           content = createDefaultContent(params.type);
         }
 
-        const newScene: Scene = {
-          id: sceneId,
-          stageId: state.stage.id,
-          type: params.type,
-          title: params.title,
-          order,
+        const newScene: Scene = makeScene(
+          {
+            id: sceneId,
+            stageId: state.stage.id,
+            title: params.title,
+            order,
+            actions: params.actions,
+            createdAt: Date.now(),
+            updatedAt: Date.now(),
+          },
           content,
-          actions: params.actions,
-          createdAt: Date.now(),
-          updatedAt: Date.now(),
-        };
+        );
 
         const newScenes = [...state.scenes, newScene].sort((a, b) => a.order - b.order);
 
--- a/lib/api/stage-api-scene.ts
+++ b/lib/api/stage-api-scene.ts
@@ -4,7 +4,7 @@
  * Factory function that creates the scene namespace of the Stage API.
  */
 
-import { makeScene, type Scene, type SceneContent } from '@/lib/types/stage';
+import { makeScene, type Scene, type ScenePatch, type SceneContent } from '@/lib/types/stage';
 import type { StageStore, APIResult, CreateSceneParams } from './stage-api-types';
 import { generateId, validateSceneId, getScene, createDefaultContent } from './stage-api-defaults';
 
@@ -119,7 +119,7 @@ export function createSceneAPI(store: StageStore) {
      * @param updates - Fields to update
      * @returns Whether successful
      */
-    update(sceneId: string, updates: Partial<Scene>): APIResult<boolean> {
+    update(sceneId: string, updates: ScenePatch): APIResult<boolean> {
       try {
         const state = store.getState();
 
@@ -127,9 +127,13 @@ export function createSceneAPI(store: StageStore) {
           return { success: false, error: `Scene not found: ${sceneId}` };
         }
 
-        const newScenes = state.scenes.map((scene) =>
-          scene.id === sceneId ? { ...scene, ...updates, updatedAt: Date.now() } : scene,
-        );
+        const newScenes = state.scenes.map((scene) => {
+          if (scene.id !== sceneId) return scene;
+          // Rebind `type` to the (possibly updated) content's kind so a
+          // content-only or type-only patch can't desync the discriminant.
+          const content = updates.content ?? scene.content;
+          return makeScene({ ...scene, ...updates, updatedAt: Date.now() }, content);
+        });
 
         store.setState({ scenes: newScenes });
 
--- a/lib/api/stage-api-scene.ts
+++ b/lib/api/stage-api-scene.ts
@@ -45,12 +45,22 @@ export function createSceneAPI(store: StageStore) {
         // Determine order
         const order = params.order ?? state.scenes.length;
 
-        // Create default content or use the provided content
+        // Create default content or use the provided content. `params.type` is
+        // authoritative: reject a `content.type` that disagrees, and pin the
+        // merged content's `type` to it so the scene's discriminant can't be
+        // silently overridden by a partial content override.
         let content: SceneContent;
         if (params.content) {
+          if (params.content.type !== undefined && params.content.type !== params.type) {
+            return {
+              success: false,
+              error: `content.type '${params.content.type}' does not match scene type '${params.type}'`,
+            };
+          }
           content = {
             ...createDefaultContent(params.type),
             ...params.content,
+            type: params.type,
           } as SceneContent;
         } else {
           content = createDefaultContent(params.type);
--- a/lib/agent/client/apply-regenerate.ts
+++ b/lib/agent/client/apply-regenerate.ts
@@ -9,7 +9,7 @@
  */
 import { nanoid } from 'nanoid';
 import type { Action } from '@/lib/types/action';
-import type { Scene, SceneContent, InteractiveContent } from '@/lib/types/stage';
+import type { Scene, ScenePatch, SceneContent, InteractiveContent } from '@/lib/types/stage';
 import type { GeneratedSlideContent } from '@/lib/types/generation';
 import { CURRENT_SLIDE_CONTENT_SCHEMA_VERSION } from '@/lib/edit/slide-schema';
 
@@ -79,7 +79,7 @@ export interface RegenerateApplyPlan {
     actionsOnly?: boolean;
   } | null;
   /** Partial scene update to apply, or null if nothing should change. */
-  patch: Partial<Scene> | null;
+  patch: ScenePatch | null;
 }
 
 /**
@@ -127,7 +127,7 @@ export function planRegenerateApply(
       | undefined;
     const existingCanvas = sceneContent?.type === 'slide' ? sceneContent.canvas : undefined;
     const runtime = toRuntimeSlideContent(details.content, existingCanvas);
-    const patch: Partial<Scene> = {
+    const patch: ScenePatch = {
       content: runtime,
       ...(actions.length > 0 ? { actions } : {}),
     };
--- a/lib/agent/client/apply-slide-content.ts
+++ b/lib/agent/client/apply-slide-content.ts
@@ -9,11 +9,11 @@
  */
 import { useStageStore } from '@/lib/store/stage';
 import { useSlideEditSession } from '@/components/edit/surfaces/slide/slide-edit-session';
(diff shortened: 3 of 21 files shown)


wyuc:
## Cross-review conclusion ✅

This branch went through a cross-vendor review loop (Claude + Codex reviewing the same `origin/main..HEAD` diff independently each round) until both converged to zero findings.

| Round | Claude | Codex | Resolution |
|------|--------|-------|------------|
| 1 | multi-angle finders (correctness / cleanup / altitude / conventions) | 1×P2 | Hardened validators: compile-time exhaustiveness assertions for `ACTION_TYPES`/`SCENE_TYPES`, scene↔content discriminant agreement check, moved `SCENE_TYPES` + `isSceneType` into `stage.ts` (matching the `guards.ts` idiom), shared one schema generator, hardened the codegen direct-invocation guard, and made the validators' scope honest in docs. Conventions clean (zero-runtime-dep invariant verified). |
| 2 | clean | **2×P2** | (a) `scene.schema.json` now ties `type` to the matching `content.type` via a discriminated `SerializedScene`, so the schema is consistent with `validateScene` instead of accepting cross-products; (b) pinned `ts-json-schema-generator` to `~2.4.0` so the build stays within the repo's `engines.node >=20.9.0` (2.5+ requires Node ≥22). |
| 3 | clean | clean | **Converged.** |

### Verification
- `pnpm --filter @openmaic/dsl build` + `typecheck` + `test` → green (29 tests).
- Zero runtime dependencies preserved (`package.json` has no `dependencies`; the generator and ajv are devDependencies only).
- Scope limited to `packages/@openmaic/dsl/**` (+ lockfile); CI (Lint/Typecheck/Unit Tests, E2E, CodeQL) green.

Marking ready for review.

cosarah (changes requested):
1. `validateAction` currently checks only object-ness, `id`, and `type`, so it can return `valid` for actions that are still unusable at runtime. For example, `validateAction({ id: "a", type: "spotlight" })` passes today in `packages/@openmaic/dsl/test/validate.test.ts:63`, but runtime consumers require variant-specific fields such as `elementId` (`lib/action/engine.ts:199`) or `topic` for `discussion` (`lib/playback/engine.ts:569`). If `validate*` is intended to be the zero-dependency gate for untrusted input / persistence boundaries, this creates dangerous false positives.

2. The public TypeScript contract and the shipped schema/validator no longer describe the same thing. `Scene` still exposes an independent `type: SceneType` plus `content: TContent` (`packages/@openmaic/dsl/src/stage.ts:201`), so a mismatched pair like `type: "quiz"` with `content.type: "slide"` still type-checks. But `validateScene` rejects that shape (`packages/@openmaic/dsl/src/validate.ts:95`), and `scene.schema.json` is generated from the stricter internal `SerializedScene` alias instead (`packages/@openmaic/dsl/src/schema-roots.ts:23`). That means the real contract has effectively moved from the exported TS types to an internal hidden type, which conflicts with the goal that the TS types remain the single source of truth.

3. The two advertised runtime validation modes operate on different domains. `validateScene` intentionally accepts envelope-level `interactive` / `pbl` scenes because `SceneType` includes them (`packages/@openmaic/dsl/src/stage.ts:20`) and the validator explicitly leaves app-owned shapes to the consuming app (`packages/@openmaic/dsl/src/validate.ts:101`). But the shipped `scene.schema.json` only covers `SlideScene | QuizScene` via `SerializedScene` (`packages/@openmaic/dsl/src/schema-roots.ts:23`). In the app, the actual scene type is the widened four-way union (`lib/types/stage.ts:88`). So the README currently describes these two enforcement modes as complementary, while in practice they will disagree on valid app scene objects.

wyuc:
Thanks @cosarah — you're right that the public TS type, `validateScene`, and `scene.schema.json` had drifted into three different notions of a valid scene. The root cause was an earlier change that made the schema/validator *stricter* than the public type, which pushed the real rule into a hidden internal alias. I've realigned everything to the public type (8c1db543):

**3 — same-domain consistency.** `validateScene` and `scene.schema.json` now cover the same domain: both allow any `SceneType` for the scene-level `type`, and both restrict `content` to the contract-owned kinds (`slide`/`quiz`), rejecting `interactive`/`pbl` *content*. The README no longer calls the two modes "complementary"; it now frames them as one authoritative layer + one cheap subset (see below).

**2 — TS types stay the single source of truth.** Reverted `SerializedScene` back to the plain default `Scene<Action, SceneContent>` and dropped `validateScene`'s `type`↔`content.type` agreement check. The schema is now generated from (effectively) the public default type and encodes no constraint the public `Scene` doesn't express — nothing hidden in an internal stricter type.

**1 — no dangerous false-positive framing.** `validateAction` stays a cheap, zero-dep *structural pre-check*. The authoritative, exhaustive per-field validation (an action's `elementId`, `discussion`'s `topic`, etc.) is the shipped **JSON Schema** — `action.schema.json` already rejects e.g. `{id, type:'spotlight'}` with no `elementId`. The docs (validate.ts header + README) now say this explicitly: `validate*` is a strict **subset** of the schema, not a separate gate; reach for the schema at real trust boundaries.

One deliberate non-change: I did **not** make the public `Scene` a discriminated union to enforce `type`↔`content` agreement at the type level. That would ripple into the app / renderer / importer and is bigger than this PR's scope — happy to take it as a follow-up if you'd like that guarantee promoted into the contract.

Cross-reviewed (Claude + Codex, zero findings) and CI is green. Re-requesting your review.

cosarah:
I still see one blocking issue in the latest revision:

`scene.schema.json` and `validateScene` now accept mismatched scene/content pairs again because `SerializedScene` was relaxed back to plain `Scene<Action, SceneContent>` (`packages/@openmaic/dsl/src/schema-roots.ts:16-19`), and the new test explicitly locks that behavior (`packages/@openmaic/dsl/test/validate.test.ts:49-56`). I verified locally that both Ajv and `validateScene` accept shapes like `{ type: "pbl", content: { type: "slide", ... } }`.

That is still a real contract bug, not just a typing compromise, because downstream code consumes `scene.type` and `content.type` as if they agree. For example, `summarizeScenes` counts by `scene.type` (`lib/classroom/complete-summary.ts:13-20`), PPTX export selects slides by `scene.content.type` (`lib/export/use-export-pptx.ts:1106-1107`), and slide APIs branch on `scene.type === "slide"` and then treat `scene.content` as `SlideContent` (`lib/api/stage-api-canvas.ts:35-49`). So a schema-valid mismatch can now be counted as PBL, exported as a slide, or routed into slide-only mutation code with the wrong content shape.

I think the right fix is still to tighten the exported `Scene` contract so that scene/content discriminants stay aligned, rather than weakening the schema/validator to match the current hole.

wyuc:
You're right, and thanks for the concrete consumer evidence — `stage-api-canvas` casting `content as SlideContent` after a `scene.type === 'slide'` check (and `complete-summary` doing the same for quiz) is exactly the load-bearing invariant. I've taken **Direction B**: tighten the exported contract instead of loosening the validators (94c20dd5, 4ea5fa8a).

**Public `Scene` now binds `type` to `content`.** It's a distributive discriminated type — `SceneCore<TAction> & { type: TContent['type']; content: TContent }` distributed over `TContent` — so TS itself rejects a mismatched scene, while generic widening (the app's four-kind `AppScene`) still works per-member. `validateScene` and the generated `scene.schema.json` (now a spelled-out discriminated `SerializedScene`) follow from it: type, validator, and schema all bind `type<->content` and agree.

**`validateAction` is now a real gate.** It checks each variant's required fields for presence *and type* (`{ type:'spotlight', elementId: 123 }` is rejected — your point 1). A bidirectional test pins the hand-written field map to the generated `action.schema.json` (the TS-derived source of truth) — it can't under-list, over-list, or mis-type a field. Validators are repositioned in the docs as the zero-dep in-process boundary producers rely on; the JSON Schema is the cross-language mirror + exhaustive value checker.

**App consumers updated.** The discriminated `Scene` surfaced the sites that built/patched scenes with independent `type`/`content`; a new `ScenePatch` (non-distributive partial) + `makeScene(core, content)` helper (one isolated cast, `type` always derived from `content.type`) routes them all. Both public scene-construction paths — `stageApi.scene.create()` and `.update()` — and the store/migrate/deserialize paths now rebind `type` to content. This closes two latent desync foot-guns (a type-only patch, and drifted persisted records).

Verified: root `tsc --noEmit` clean, full unit suites green (dsl + app + importer), and cross-reviewed (Claude + Codex) to zero findings across two fix rounds. Re-requesting your review.

lib/api/stage-api-scene.ts
```diff
@@ -56,18 +56,19 @@ export function createSceneAPI(store: StageStore) {
           content = createDefaultContent(params.type);
         }
 
-        const newScene: Scene = {
-          id: sceneId,
-          stageId: state.stage.id,
-          type: params.type,
-          title: params.title,
-          order,
+        const newScene: Scene = makeScene(
```
cosarah, line 69: CreateSceneParams still exposes an independent type plus content?: Partial<SceneContent>, so makeScene(...) now lets content.type silently override params.type. A call like api.scene.create({ type: 'slide', content: { type: 'interactive', html: '...' } }) now succeeds as an interactive scene, and because we still seed defaults from params.type above it can also produce hybrid payloads with the wrong scene kind defaults mixed in. We should either reject mismatched kinds here or strip content.type so the public type parameter remains authoritative.

wyuc:
Good catch — fixed in the latest commit. `scene.create()` now keeps `params.type` authoritative: it rejects a `content.type` that disagrees with `params.type` (fail-loud), and pins the merged content's `type` to `params.type` so a partial `content` override can't silently flip the scene's discriminant (or mix in wrong-kind defaults). So `create({ type: 'slide', content: { type: 'interactive', … } })` now returns `{ success: false }` instead of producing a hybrid interactive scene. Added a test both ways. (Also rebased onto the latest `main` — the #814 `outlineId` now flows through `makeScene`.)

cosarah (approved):
LGTM

PR #820, opened 2026-06-30
feat(playback): add seek bar and resume position

### LINKED ISSUE #818: [Feature]: Add playback seek bar and resume position support for classroom slides
### Problem or Motivation

When using the classroom playback experience, it is difficult to control playback position.

Currently, if I move from one slide/scene to another and then return, playback does not reliably continue from the point where I left off. There is also no seek/progress control that allows me to move forward or backward within the current slide/scene playback.

This becomes frustrating when:

- I miss part of the narration and want to go back a few seconds
- I already watched part of a slide and want to skip ahead
- I navigate to another slide/scene and then return
- A generated classroom has long narration or actions, but I cannot resume from the previous position

Expected behavior:

- The classroom player should show current playback progress
- Users should be able to seek forward/backward within the current slide/scene
- When returning to a previous slide/scene, playback should resume from the last known position instead of starting over
- Existing play/pause and slide navigation behavior should continue to work normally

Related issue: #73 mentions similar pain around playback control during discussion, but this request is specifically for classroom slide/scene playback seek and resume behavior.

### Proposed Solution

Add a seek/progress bar to the classroom playback UI.

Proposed UX:

- Show a progress bar/slider for the active slide or scene
- Display current time and total duration where possible
- Allow users to drag the slider to seek forward/backward
- Store the last playback position for each slide/scene during the current classroom session
- When the user returns to a slide/scene, resume from the last stored position
- Keep the control minimal so it does not interfere with the existing classroom UI

Possible implementation direction:

- Add playback progress state to the classroom playback layer
- Track progress by slide/scene ID
- Expose a `seekTo`-style action in the playback state/controller
- Sync seeking with narration/audio and visual classroom actions
- Use existing UI primitives if available instead of adding a new dependency
- Add i18n keys for any visible labels/tooltips instead of hardcoding user-facing text

Possible files/areas to inspect:

- `app/classroom/[id]/`
- `lib/playback/`
- `lib/action/`
- `lib/store/`
- `components/`
- `messages/`

I am happy to work on this feature and open a PR if the maintainers agree with the direction.

### Alternatives Considered

Alternative 1: Only add previous/next slide controls.

This does not solve the main issue because users still cannot jump to a specific timestamp inside a long slide/scene.

Alternative 2: Always restart the slide from the beginning.

This is the current behavior and is frustrating for long generated lessons.

Alternative 3: Add only play/pause.

Play/pause helps temporarily, but it does not help users go back to something they missed or skip forward to a specific point.

### Area

Classroom generation

### Additional Context

_No response_

danishsshaikh:
## Summary

Adds a seek/progress bar to classroom playback so users can move forward/backward within the active slide/scene and resume from the last known position when returning to a slide/scene.

Generated TTS audio supports precise seeking through the shared audio player. Browser Native TTS uses a best-effort sentence/chunk fallback because the Web Speech API does not expose precise media-style seeking APIs.

## Related Issues

Closes #818

## Changes

- Added scene-level playback progress tracking in the playback engine
- Added `seekTo` support for classroom playback
- Added generated-audio seeking through the shared audio player
- Added a compact seek/progress slider to the classroom playback toolbar
- Stores last playback position per scene in `sessionStorage`, scoped to the current classroom/stage
- Restores the previous scene position when returning to a slide/scene
- Added i18n strings for seek/progress accessibility labels
- Added playback seek tests

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Open a generated classroom in playback mode.
2. Start playback and verify the progress slider updates while the scene is playing.
3. Pause playback and verify the progress position freezes.
4. Drag the seek bar forward/backward and verify playback resumes from the selected position.
5. Navigate to another slide/scene, then return to the previous scene and verify the saved playback position is restored.
6. Verify existing play/pause and slide navigation behavior still works.

- Verified the seek bar appears during classroom playback
- Verified playback progress updates while playing
- Verified pause freezes progress
- Verified seeking forward/backward works
- Verified returning to a previous scene restores the saved playback position
- Verified existing play/pause and slide navigation still work
- Verified generated-audio seek behavior through the shared audio player
- Verified Browser Native TTS uses a best-effort fallback path
- Verified the feature on my server with real classroom playback
- `pnpm lint` reports 0 errors. The remaining warnings appear unrelated to this PR.

### Evidence

- [x] Manually tested locally

Screenshots:

1. Seek/progress bar displayed during classroom playback

Screenshot 2026-06-30 at 3 47 20 PM

2. Playback position updated after seeking

Screenshot 2026-06-30 at 3 47 33 PM

3. Previous scene resumes from the saved playback position

Screenshot 2026-06-30 at 3 47 38 PM
Screenshot 2026-06-30 at 3 46 37 PM

Local verification run:

```bash
pnpm check
pnpm lint
npx tsc --noEmit
pnpm test tests/playback/engine-seek.test.ts
pnpm build
```

Results:

- `pnpm check` passed
- `pnpm lint` completed with 0 errors and existing warnings
- `npx tsc --noEmit` passed
- `pnpm test tests/playback/engine-seek.test.ts` passed: 4 tests
- `pnpm build` passed
- Manually tested on server with real classroom playback

## Checklist

### DIFF
--- a/components/canvas/canvas-area.tsx
+++ b/components/canvas/canvas-area.tsx
@@ -43,6 +43,8 @@ export function CanvasArea({
   onTogglePresentation,
   showStopDiscussion,
   onStopDiscussion,
+  playbackProgress,
+  onSeek,
   hideToolbar,
   isPendingScene,
   isCourseComplete,
@@ -267,6 +269,8 @@ export function CanvasArea({
           onTogglePresentation={onTogglePresentation}
           showStopDiscussion={showStopDiscussion}
           onStopDiscussion={onStopDiscussion}
+          playbackProgress={playbackProgress}
+          onSeek={onSeek}
         />
       )}
     </div>
--- a/components/canvas/canvas-toolbar.tsx
+++ b/components/canvas/canvas-toolbar.tsx
@@ -20,6 +20,8 @@ import { cn } from '@/lib/utils';
 import { useStageStore } from '@/lib/store';
 import { useI18n } from '@/lib/hooks/use-i18n';
 import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from '@/components/ui/tooltip';
+import { Slider } from '@/components/ui/slider';
+import type { PlaybackProgress } from '@/lib/playback';
 
 export interface CanvasToolbarProps {
   readonly currentSceneIndex: number;
@@ -50,6 +52,8 @@ export interface CanvasToolbarProps {
   readonly onToggleAutoPlay?: () => void;
   readonly playbackSpeed?: number;
   readonly onCycleSpeed?: () => void;
+  readonly playbackProgress?: PlaybackProgress | null;
+  readonly onSeek?: (timeMs: number) => void;
 }
 
 /* Compact control button */
@@ -62,6 +66,13 @@ const ctrlBtn = cn(
 /* Subtle separator */
 function CtrlDivider() {
   return <div className="w-px h-3 bg-gray-200/80 dark:bg-gray-700/60 mx-0.5 shrink-0" />;
+}
+
+function formatPlaybackTime(timeMs: number): string {
+  const totalSeconds = Math.max(0, Math.floor(timeMs / 1000));
+  const minutes = Math.floor(totalSeconds / 60);
+  const seconds = totalSeconds % 60;
+  return `${minutes}:${seconds.toString().padStart(2, '0')}`;
 }
 
 /* Volume icon based on level */
@@ -108,6 +119,8 @@ export function CanvasToolbar({
   onToggleAutoPlay,
   playbackSpeed = 1,
   onCycleSpeed,
+  playbackProgress,
+  onSeek,
 }: CanvasToolbarProps) {
   const { t } = useI18n();
   const canGoPrev = currentSceneIndex > 0;
@@ -120,6 +133,7 @@ export function CanvasToolbar({
 
   // Volume slider hover state
   const [volumeHover, setVolumeHover] = useState(false);
+  const [seekDraft, setSeekDraft] = useState<number | null>(null);
   const volumeTimerRef = useRef<ReturnType<typeof setTimeout>>(undefined);
   const volumeContainerRef = useRef<HTMLDivElement>(null);
 
@@ -138,6 +152,19 @@ export function CanvasToolbar({
   // Effective volume for display
   const effectiveVolume = ttsMuted ? 0 : ttsVolume;
   const presentationLabel = isPresenting ? t('stage.exitFullscreen') : t('stage.fullscreen');
+  const hasPlaybackProgress = !!playbackProgress;
+  const seekable =
+    !!playbackProgress?.seekable && playbackProgress.durationMs > 0 && !!onSeek && !isLiveSession;
+  const seekValue = Math.min(
+    seekDraft ?? playbackProgress?.currentTimeMs ?? 0,
+    playbackProgress?.durationMs ?? 0,
+  );
+  const currentTimeLabel = formatPlaybackTime(seekValue);
+  const durationLabel = formatPlaybackTime(playbackProgress?.durationMs ?? 0);
+  const playbackTimeLabel = t('stage.playbackTime', {
+    current: currentTimeLabel,
+    duration: durationLabel,
+  });
 
   return (
     <div className={cn('flex items-center gap-2', className)}>
@@ -168,7 +195,7 @@ export function CanvasToolbar({
       <CtrlDivider />
 
       {/* ── Center: unified playback controls ── */}
-      <div className="flex-1 flex items-center justify-center min-w-0">
+      <div className="flex-1 flex items-center justify-center min-w-0 gap-2">
         <div
           className={cn(
             'inline-flex items-center gap-0.5 px-1 h-7',
@@ -394,6 +421,33 @@ export function CanvasToolbar({
             )}
           </button>
         </div>
+
+        {hasPlaybackProgress && (
+          <div className="hidden sm:flex items-center gap-2 min-w-[140px] w-[min(28vw,260px)]">
+            <Slider
+              value={[seekValue]}
+              min={0}
+              max={Math.max(1, playbackProgress?.durationMs ?? 1)}
+              step={250}
+              disabled={!seekable}
+              aria-label={t('stage.playbackSeekLabel')}
+              aria-valuetext={playbackTimeLabel}
+              onValueChange={([value]) => {
+                if (typeof value === 'number') setSeekDraft(value);
+              }}
+              onValueCommit={([value]) => {
+                setSeekDraft(null);
+                if (typeof value === 'number' && seekable) onSeek?.(value);
+              }}
+              className={cn(!seekable && 'opacity-45')}
+            />
+            <span className="w-[72px] shrink-0 text-[10px] leading-none tabular-nums text-gray-400 dark:text-gray-500 select-none">
+              {currentTimeLabel}
+              <span className="opacity-35 mx-px">/</span>
+              {durationLabel}
+            </span>
+          </div>
+        )}
       </div>
 
       {/* ── Right: fullscreen + chat toggle ── */}
--- a/components/edit/PlaybackChromeRoot.tsx
+++ b/components/edit/PlaybackChromeRoot.tsx
@@ -714,7 +714,11 @@ export const PlaybackChromeRoot = forwardRef<PlaybackChromeRootHandle, PlaybackC
         // Restore this scene's session-only position, but never auto-play.
         const savedPositionMs = scenePlaybackPositionsRef.current[currentScene.id] ?? 0;
         if (savedPositionMs > 0) {
-          updatePlaybackProgress(engine.seekTo(savedPositionMs));
+          void engine.seekTo(savedPositionMs).then((progress) => {
+            if (engineRef.current === engine) {
+              updatePlaybackProgress(progress);
+            }
+          });
         } else {
           updatePlaybackProgress(engine.getProgress());
         }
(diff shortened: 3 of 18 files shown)


wyuc:
**Review — the feature is wanted (#818 is a real need) and the UI/persistence layers are in good shape, but there are two P1s in the engine's seek model that need fixing before this can merge**

What's already solid: progress correctly freezes on pause (and the test covers it), the browser-TTS chunk-offset fallback handles the edge cases (empty text, offset past end, single chunk), `audio-player.seekTo` clamps past-end once metadata loads, sessionStorage access is try/catch-guarded with shape validation, the slider has proper `aria-label`/`aria-valuetext` with all locales present, and unmount cleanup saves position and clears timers correctly.

**🔴 P1 — Seek repositions the cursor but never reconstructs state from non-speech actions**

`seekTo()` (`lib/playback/engine.ts:180-189`) cancels playback, sets `actionIndex`, and resumes — nothing re-executes or undoes the actions in between, and non-speech actions can't even be seek targets (`INSTANT_ACTION_DURATION_MS = 0`, engine.ts:538):

- *Forward*: in `[speech, wb_open, wb_draw_text, speech2]`, seeking into `speech2` never runs the whiteboard actions — narration talks over a whiteboard that was never opened/drawn. Same for `widget_setState`/`widget_reveal` and `play_video`.
- *Backward*: `cancelActivePlayback()` only clears spotlight/laser overlays; whiteboard elements persist in the store, and `addElement` (`lib/api/stage-api-whiteboard.ts:155-166`) appends unconditionally with no id dedupe — play → seek to 0 → play again draws every element twice, doubling on each repeat.

Fix direction: a fast-forward pass that executes non-speech actions from a checkpoint up to the target index with animations suppressed, plus a state reset when seeking backward. If that's too big for this PR, an honest v1 scope cut is to enable the slider only for scenes without stateful actions.

**🔴 P1 — No playback generation token: seeking during a pending `play()` races and corrupts playback**

`processNext()` (`engine.ts:718-747`) starts speech with `audioPlayer.play(...).then(...)`, and the continuation never checks whether its action is still the active one. `play()` awaits an IndexedDB read (`audio-player.ts:52`) — a real async window. Seek inside it and:

- a stale `play()` resolving `true` calls `stop()` (killing the new action's audio) and plays the *old* action's clip — engine on action B, audio from action A, `onEnded` advances from B;
- a stale `play()` resolving `false` calls `scheduleReadingTimer()` with the old action's duration, overwriting `this.speechTimer` while the previous timeout keeps running → both fire `processNext()` → double-advance and skipped lines. Rapid consecutive seeks stack more.

Fix: a monotonically increasing generation captured before `play()` and checked in `.then`/`.catch` and in the reading-timer callback, bumped by `seekTo`/`pause`/`stop`/`cancelActivePlayback`.

**🟠 P2 — Seeking while playing permanently freezes the lecture transcript** *(both reviewers flagged this independently)*

`handleSeek` (`PlaybackChromeRoot.tsx:915-917`) unconditionally calls `pauseBuffer(lectureSessionId)`, but `resumeBuffer` only ever runs from the pause→resume branch of `handlePlayPause`. Seek during playback → engine keeps playing, audio and slides continue, but the chat transcript stops revealing until the user happens to toggle pause/resume. `handleSeek` should resume (or not pause) the buffer when the engine stays `'playing'`.

**🟠 P2 — Stale audio element corrupts the timeline, progress, and precise seek**

The `ended` listener never nulls `this.audio` (`audio-player.ts:42-44`), and `play()` returning `false` (the normal "no TTS generated" path) returns without `stop()`. In `[speech A (audio), speech B (no audio)]`: `computeProgress()` reports A's final position as B's elapsed, `getActionDurationMs()` returns A's duration as B's, and the precise-seek path "succeeds" against the dead element so the user's drag visibly does nothing. Fix: null the element in `ended` (revoke first) and/or `stop()` on the missing-record path.

**🟠 P2 — The `targetMs >= durationMs` branch mishandles both seek-to-end and restore**

Two sides of the same branch (`engine.ts:155-163`):

- Committing the slider at the end stops the engine without firing the normal completion callback — UI ends up idle-but-not-completed, skipping the usual lecture cleanup.
- On remount, scene duration is estimate-only (~150 ms/char), so a saved position recorded against real (longer) audio can exceed it → restore hits this branch, pins the cursor at the end, and with auto-advance on, the scene **skips past itself**, silently discarding the user's real position. Stale entries after content regeneration (positions keyed by `scene.id` only) collapse the same way. Consider routing end-seeks through the completion path, and clamping restores to the last action boundary below `durationMs` (or dropping positions ≥ ~95%).

**🟡 P3** — `seekTo` has no `'live'`-mode guard (the slider disable is UI-convention only, and Roundtable's gate at `roundtable/index.tsx:632` omits the `!!chatSessionType` term the chrome-level gate has); a cheap early-return in the engine would make it robust.

**🟡 P3** — The 250 ms progress interval runs while idle/paused and re-renders the chrome tree 4×/s, plus a `JSON.stringify` + `sessionStorage.setItem` per tick while playing; skip ticks outside `'playing'` and throttle storage writes to ~1–2 s (teardown flush already exists).

The two P1s are the blockers — happy to re-review quickly once they're addressed, whichever scope you choose for the first one.

danishsshaikh:
Thanks for the detailed review @wyuc, this makes sense.

I’m going to investigate the fuller fix path for the first P1 before falling back to a v1 scope cut. Specifically, I’ll look into whether we can reset/reconstruct state for non-speech actions and fast-forward stateful actions up to the seek target with animations suppressed, so seeking forward/backward does not skip or duplicate whiteboard/widget/video state.

For the async playback race P1, I’ll add a monotonically increasing playback generation token and check it across async audio `play()` continuations, error paths, and reading-timer callbacks, bumping it on seek/pause/stop/cancel paths.

I’ll keep the investigation focused on the playback engine and report back if full deterministic reconstruction turns out to be too large for this PR, in which case I’ll use the safer v1 scope cut you suggested :)

danishsshaikh:
I did a deeper pass on the unsafe action families and the limited-speech-seek option.

For the current #820 fix, I’m going to keep the behavior conservative:

- enable full scene seek only for seek-safe scenes
- support deterministic reconstruction where the engine can safely reset/replay state, mainly speech and whiteboard/canvas-style actions
- disable the scene-wide seek slider for scenes containing action families that do not currently have a deterministic reset/replay contract, such as `widget_*`, `discussion`, and `play_video`
- add an engine-level guard so unsafe `seekTo()` calls no-op even if the UI path is bypassed
- add the playback generation-token fix for async audio/timer races

I also investigated whether disabled scenes could still support a limited “current speech only” seek mode.

Technically, the engine can identify the active speech action and already has a same-action generated-audio seek path. So we may be able to support seeking within the currently active narration/audio clip without crossing widget/discussion/video boundaries.

However, the current toolbar exposes a scene-wide slider (`0 → durationMs`). If we leave that slider enabled for current-speech-only seek, it would imply that the whole scene is seekable, and rejecting/snap-backing when the user drags outside the active speech segment would be confusing.

So my plan for this PR is:

- keep #820 focused on correct scene-level seek semantics
- disable scene seek for unsafe scenes for now
- leave current-speech-only seek for unsafe scenes as a possible follow-up, likely after `PlaybackProgress`/UI can expose something like `seekMode` or `seekableRange`

If you think the limited current-speech-only behavior should be included in #820 despite the UI caveat, I’m happy to adjust, but I’ll proceed with the conservative scene-level guard first.

danishsshaikh:
I pushed the first pass addressing the two P1 engine blockers.

This update includes:

- scene seek-safety detection
- scene-wide seek disabled/guarded for `widget_*`, `discussion`, and `play_video`
- engine-level `seekTo()` guard for unsafe scenes
- whiteboard/canvas reset + silent replay for seek-safe scenes
- playback generation tokens for stale async audio/timer callbacks
- AudioPlayer request-token cleanup and stale audio cleanup
- additional playback/audio tests

Automated checks are passing, but I’m still doing server/manual validation before considering this ready for re-review.

During manual testing I noticed a couple of behaviors I want to inspect before asking for final review:

- spotlight/laser can sometimes appear duplicated or retriggered around seek boundaries
- progress/duration can drift slightly after seeking in some scenes

I’m investigating those now and will push a follow-up if needed. I’ll update again once server validation is complete.

wyuc:
**Update after an end-to-end pass — direction change: we're not going to pursue a continuous-time scrubber for classroom playback**

We ran this branch end-to-end in a real generated classroom (real TTS audio, Chromium). Credit where due — the core works as built: play/pause, both seek directions, and the immediate seek feedback (jump + transcript sync + audio pickup) all behave; per-scene resume is genuinely solid (position survives scene switches to the millisecond).

But the pass also surfaced three user-visible failures, and they're the deciding factor:

1. **Seek during playback permanently freezes the transcript** — audio kept narrating for 10+ seconds while the chat transcript never updated again (the `pauseBuffer`-without-resume path from the earlier review, confirmed live).
2. **Seeking to the end bricks the scene** — progress pins at full, the transcript bubble empties, and pressing Play does nothing; the only way out is dragging the slider back.
3. **The total duration visibly jumps around** (0:45 ↔ 0:52 ↔ 0:57 for the same scene) as real audio durations replace per-line estimates.

Stepping back, these aren't three independent bugs — they're the cost of grafting a continuous-media metaphor onto a discrete action engine. The timeline has to fake continuity (estimates → jumping totals), seeks have to fake arbitrary positions (boundary snapping), and the classroom has to fake random access (whiteboard/widget state is cumulative — the state-reconstruction P1 from the review). Fixing all of that buys an interaction model that still doesn't match what learners actually do, which is "replay that sentence" / "skip this part".

So the direction we want for #818 is **sentence/action-level navigation** instead (detailed on the issue):

- click a transcript line to jump to it; prev/next-sentence controls in the toolbar
- progress shown per-line/segment (no estimated continuous timeline)
- keep this PR's per-scene resume idea, recorded as an action index rather than milliseconds

Two concrete paths forward, both welcome:
1. **Split the resume-position work into its own small PR** — that part is wanted as-is and shouldn't be held hostage by the navigation paradigm.
2. **Rework this PR (or open a fresh one) toward line-level navigation** per #818. Note the two engine findings from the earlier review still apply to line jumps (non-speech action fast-forward/reset, and a playback generation token), so that work isn't wasted.

Leaving this open in case you want to take it in that direction — and thanks for the thorough build; the verdict here is about the interaction paradigm, not the execution quality.

danishsshaikh:
Thanks for the detailed end-to-end pass and for the clear direction :)

That makes sense.

I’ll keep this branch as an exploration/reference branch, since the generation-token and non-speech reset/fast-forward findings are still useful for future action-level navigation work.

For the wanted part, I’ll open a fresh smaller PR, focused only on per-scene resume, with action-level resume rather than millisecond-based resume.

After that, I can follow up separately on the #818 direction:

- click transcript line to jump to that action
- prev/next sentence controls
- progress per line/segment instead of a continuous timeline

danishsshaikh:
I opened the smaller resume-only PR here: #836 

That PR preserves the wanted per-scene resume behavior, changes it to action-level resume, and removes the continuous-time scrubber scope.

I’ll close this PR to avoid splitting review attention, but I’m keeping the seekbar branch in my fork as an exploration/reference branch because the generation-token and non-speech reset/fast-forward findings are still useful for future action-level navigation work.

Thanks again for the detailed end-to-end feedback.

PR #825, opened 2026-07-01
feat(dsl): activate the migration registry + runner (#787 Part B-2)

wyuc:
Implements **Part B-2** of #787 (Phase 2 — grow the DSL): activate the migration registry. Builds on #811 (Action verbs in the contract) and #817 (JSON Schema + validators). This is the last open box under #787.

`version.ts` shipped a stub — `DSL_MIGRATIONS = []`, `DSL_VERSION` defined but unused, no runner. This PR fills in the **document-level migration mechanism** while keeping `@openmaic/dsl` **zero-runtime-dependency** (pure functions only, no new deps).

## What's here

**1. The `migrate` runner** (`src/version.ts`)
- `migrate(doc)` reads the document's `dslVersion` envelope field (absent ⇒ treated as legacy/`0.0.0`), walks the `DSL_MIGRATIONS` ladder up to `DSL_VERSION`, applying each pure transform and stamping the result.
- **Idempotent** — a document already at `DSL_VERSION` is returned by reference.
- **Forward-compatible** — a document stamped *newer* than `DSL_VERSION` is returned untouched, never silently downgraded (mirrors the app's `migrateSlideContent`).
- **Fail-loud** — a gap in the ladder throws rather than yielding a half-migrated document.
- **Pure** — never mutates its input; each step returns a fresh stamped object.
- Plus `dslVersionOf` / `needsMigration` readers, the `DslVersioned` envelope type, and `UNVERSIONED_DSL_VERSION` / `DSL_VERSION_KEY` constants.

**2. First ladder entry** — `{ from: '0.0.0', to: '0.1.0', migrate: (doc) => doc }`: a no-op *transform* that stamps legacy (pre-`dslVersion`) documents up to the current version. Promoting `Action` into the contract (#811) and adding validators (#817) did **not** change any serialized document shape, so the current on-disk shape already *is* `0.1.0`. This entry wires the pipeline end to end and gives real documents a version to migrate forward from — without faking a shape bump.

**3. Two independent version axes (documented in the README)**
- `DSL_VERSION` (serialized-contract shape) stays `0.1.0` — no on-disk shape changed.
- the **npm package** version minor-bumps `0.1.0 → 0.2.0` for the new API surface.
- (the finer app-side `SlideContent.schemaVersion`, which versions the PPTist canvas *inside* a slide, is orthogonal and untouched.)

**4. Tests** (`test/version.test.ts`) — ladder invariants (contiguous chain ending at `DSL_VERSION`), stamp/idempotence/purity, forward-compat, fail-loud, non-object passthrough.

## Where the version field lives

The runner only needs the `dslVersion` envelope field to read/stamp; **which aggregate carries it** (a whole `Stage`, a Scene row, or a bundle) is deliberately left open — that's decided when a normalized store first consumes this pipeline, matching the open question in #787. So this PR lands the *mechanism* + proves it end to end via tests; app-side store wiring is a follow-up.

## Verification

- `pnpm --filter @openmaic/dsl test` → 86 passed (incl. new `version.test.ts`)
- `pnpm --filter @openmaic/dsl typecheck` → clean
- `pnpm --filter @openmaic/dsl build` → `dist/schema/*.json` still generated
- `prettier --check` → clean

## Acceptance (#787)

- [x] `DSL_MIGRATIONS` runs a first migration end to end; `version.ts` is no longer a stub
- [x] zero-runtime-dep invariant held (pure functions, no new deps)
- [x] `test/` covers the migration runner
- [x] README updated with the version/migration mechanism + the version-axis topology

### DIFF
--- a/packages/@openmaic/dsl/src/version.ts
+++ b/packages/@openmaic/dsl/src/version.ts
@@ -29,9 +29,19 @@ export type DslVersion = typeof DSL_VERSION;
 /**
  * The version a document is treated as when it carries no {@link DSL_VERSION_KEY}
  * stamp: everything written before the version field existed. The first
- * migration lifts these legacy documents to {@link DSL_VERSION}.
+ * migration lifts these legacy documents forward.
  */
 export const UNVERSIONED_DSL_VERSION = '0.0.0' as const;
+
+/**
+ * The first shipped serialized-contract version — a **pinned literal**, not the
+ * moving {@link DSL_VERSION}. Migration endpoints must be immutable: they name a
+ * fixed point in the ladder, so they cannot reference `DSL_VERSION` (which moves
+ * every time the shape changes). It equals `DSL_VERSION` today; the two diverge
+ * the moment the first real shape change bumps `DSL_VERSION` and appends a step
+ * from here.
+ */
+export const INITIAL_DSL_VERSION = '0.1.0' as const;
 
 /**
  * Envelope property that carries the serialized-contract version on a document.
@@ -66,18 +76,22 @@ export interface DslMigration {
 
 /**
  * Ordered migration ladder. Each entry's `to` is the next entry's `from`, and
- * the last entry's `to` is {@link DSL_VERSION} (both checked by a test).
+ * the last entry's `to` is {@link DSL_VERSION} (both checked by a test). Every
+ * `from` / `to` is a **pinned literal** — never the moving `DSL_VERSION`
+ * constant — so appending a future step can't retroactively re-target an
+ * existing one.
  *
- * The first entry stamps legacy (pre-`dslVersion`) documents up to the current
- * contract version. It is intentionally a no-op *transform*: bringing `Action`
- * into the contract (#811) and adding validators (#817) did not alter any
- * serialized document, so the current on-disk shape already *is* 0.1.0. The
+ * The first entry stamps legacy (pre-`dslVersion`) documents up to
+ * {@link INITIAL_DSL_VERSION}. It is intentionally a no-op *transform*: bringing
+ * `Action` into the contract (#811) and adding validators (#817) did not alter
+ * any serialized document, so the current on-disk shape already *is* 0.1.0. The
  * entry exists to wire the pipeline end to end and to give real documents a
  * version stamp to migrate forward from. When the serialized shape first
- * changes, bump {@link DSL_VERSION} *then* and append a real transform.
+ * changes, bump {@link DSL_VERSION} *then* and append a real transform from
+ * `INITIAL_DSL_VERSION` to the new pinned version.
  */
 export const DSL_MIGRATIONS: readonly DslMigration[] = [
-  { from: UNVERSIONED_DSL_VERSION, to: DSL_VERSION, migrate: (doc) => doc },
+  { from: UNVERSIONED_DSL_VERSION, to: INITIAL_DSL_VERSION, migrate: (doc) => doc },
 ];
 
 function isObject(v: unknown): v is Record<string, unknown> {
--- a/packages/@openmaic/dsl/src/version.ts
+++ b/packages/@openmaic/dsl/src/version.ts
@@ -1,13 +1,24 @@
 /**
- * DSL version + migration scaffolding.
+ * DSL version + migration registry.
  *
  * The DSL version is independent of the npm package version: it identifies the
- * shape of the serialized slide contract so that persisted documents can be
- * migrated forward as the schema evolves.
+ * shape of the *serialized* slide contract so that persisted documents can be
+ * migrated forward as the schema evolves. A package release can bump for
+ * code/API reasons (new exports, refactors) without touching the serialized
+ * shape — in which case {@link DSL_VERSION} stays put; conversely the first
+ * breaking change to the on-disk shape bumps {@link DSL_VERSION} and appends a
+ * migration, regardless of where the package version happens to be.
  *
- * This is intentionally a stub: the real migration registry is filled in once
- * the first breaking schema change lands. The types below pin down the shape
- * that registry will take so callers can already code against it.
+ * This module owns the *mechanism*: the ordered {@link DSL_MIGRATIONS} ladder,
+ * plus the pure {@link migrate} runner that walks a document from whatever
+ * version it was written at up to {@link DSL_VERSION}. It carries no runtime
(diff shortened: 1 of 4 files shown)


wyuc (commented):
E2E regression run for PR #825 (`a375114`), compared with base `6538e718`.

Scope tested:
- Fixed `.maic.zip` import regression via the homepage file chooser, using `江苏卷第16题：碰撞与质心全拆解.maic.zip`
- Imported IndexedDB state verification
- Imported classroom open/render
- Real classroom next-scene navigation into the interactive scene
- Same script rerun against base `main` for baseline comparison

Checks that passed on this PR:
- Import completed through the UI.
- IndexedDB state matched the expected fixture shape:
  - 1 stage
  - 10 scenes
  - 4 generated agents
  - 63 audio records
  - 1 generated media record
  - 115 actions
  - scene types: `slide`, `interactive`
- Imported classroom opened successfully.
- First slide rendered with title `高考压轴题：江苏卷第16题`.
- Scene 2 rendered the interactive iframe for `场景识别：拆解四个模型`.

Focused tests also passed:
- `pnpm --filter @openmaic/dsl test` → 4 files / 86 tests
- `pnpm test tests/export/classroom-zip.test.ts tests/export/export-classroom-inline.test.ts tests/export/inline-assets-importmap.test.ts` → 3 files / 20 tests

Not a PR-specific regression, but the full fixed import E2E is not green:
- The interactive iframe's visible `下一步` button could not be clicked by Playwright because `button#nextBtn` is intercepted by the same `#controls` subtree.
- The iframe reports `Identifier 'config' has already been declared`.
- The same iframe click failure and runtime error reproduce on base `main` (`6538e718`) with the same script and fixture.
- The home page also reports a React hydration mismatch in dev mode; this also appears on both PR and base.

Conclusion: the DSL migration registry/runner change does not appear to introduce a new import/render regression. The remaining failing E2E item is a pre-existing fixed-fixture interactive iframe issue, not caused by this PR.

wyuc:
## Cross-review (Claude + Codex) — converged, 0 findings

Two independent reviewers evaluated the same `origin/main..HEAD` diff.

**Round 1**

| Severity | Raised by | Finding | Verdict |
|---|---|---|---|
| P2 | Codex | The first `DSL_MIGRATIONS` entry used `to: DSL_VERSION`. Once a future shape change bumps `DSL_VERSION` and appends a real step, that entry's `to` would move with it, so legacy (`0.0.0`) documents would be stamped straight to the new version and **skip the appended transform** — silent data corruption. Migration endpoints must be immutable. | Confirmed, fixed |

Fixed in the follow-up commit: added a pinned `INITIAL_DSL_VERSION` literal and pointed the first entry at it instead of the moving `DSL_VERSION` constant, plus a regression assertion (`DSL_MIGRATIONS[0].to === INITIAL_DSL_VERSION`).

**Round 2** (post-fix re-review): Codex — *"No actionable correctness issues found"*. Consensus: zero findings.

**Local verification**: `pnpm --filter @openmaic/dsl test` → 86 passed · `typecheck` clean · `build` regenerates `dist/schema/*.json` · `prettier --check` clean.

packages/@openmaic/dsl/src/version.ts
```diff
@@ -29,7 +75,104 @@ export interface DslMigration {
...
-export const DSL_MIGRATIONS: readonly DslMigration[] = [];
+export function dslVersionOf(doc: unknown): string {
+  if (isObject(doc) && typeof doc[DSL_VERSION_KEY] === 'string') {
+    return doc[DSL_VERSION_KEY] as string;
+  }
+  return UNVERSIONED_DSL_VERSION;
+}
+
+/** True when `doc` is written at an older version than {@link DSL_VERSION}. */
+export function needsMigration(doc: unknown): boolean {
```
cosarah, line 151: `needsMigration()` and `migrate()` disagree on non-object inputs. `dslVersionOf(42)` maps to `"0.0.0"`, so `needsMigration(42)` is `true`, but `migrate(42)` returns `42` unchanged because the stamping path is a no-op for non-objects. A caller that checks `needsMigration()` before or after migrating will think the value still needs migration forever. It would be better to either fail fast on non-objects or make `needsMigration()` mirror the same no-op semantics.

packages/@openmaic/dsl/src/version.ts
```diff
@@ -29,7 +75,104 @@ export interface DslMigration {
...
+export const DSL_MIGRATIONS: readonly DslMigration[] = [
+  { from: UNVERSIONED_DSL_VERSION, to: INITIAL_DSL_VERSION, migrate: (doc) => doc },
+];
+
+function isObject(v: unknown): v is Record<string, unknown> {
+  return typeof v === 'object' && v !== null && !Array.isArray(v);
+}
+
+/** Parse an `x.y.z` version into numeric parts; non-numeric parts become 0. */
+function parseVersion(v: string): [number, number, number] {
```
cosarah, line 107: This parser currently treats malformed version stamps as valid comparison inputs. For example, `"1"`, `"0.1"`, and `"0.1.0-beta"` all compare as >= `"0.1.0"`, so `needsMigration()` returns `false` and `migrate()` leaves them untouched. That seems to violate the documented `x.y.z` / fail-loud contract and would let invalid persisted data bypass migration entirely. Can we validate the stamp format before comparing, or treat any non-`x.y.z` value as an error?

packages/@openmaic/dsl/src/version.ts
wyuc: Fixed in 921a3ba. `needsMigration` and `migrate` now agree on non-objects: both treat a non-object as "nothing to migrate" (`needsMigration` → `false`, `migrate` → input unchanged), so `while (needsMigration(x)) x = migrate(x)` can't loop forever. Added a test asserting the agreement invariant across non-object inputs.

packages/@openmaic/dsl/src/version.ts
wyuc: Fixed in 921a3ba. Added an `isValidVersion` (`x.y.z`) guard; `dslVersionOf` now **fails loud** on a present-but-malformed stamp (e.g. `"1"`, `"0.1"`, `"0.1.0-beta"`) instead of letting `parseVersion` coerce it into a comparable value that reads as ≥ DSL_VERSION and bypasses migration. Absent stamp / non-object still map to the unversioned baseline. Test covers the malformed-stamp fail-loud path.

cosarah (approved):
LGTM

PR #829, opened 2026-07-01
feat(interactive-actions): feed real HTML element inventory into the prompt

### LINKED ISSUE #824: Include HTML element inventory when generating widget teacher actions, to fix interactive action/UI mismatch
## Problem

For interactive (widget) scenes, teacher actions (`highlight` / `setState` / `annotation`) frequently **fail to drive the UI** during playback — the action `target` selectors don't match the real elements in the widget HTML.

## Root cause

When generating teacher actions, the model **only sees `widgetConfig`, not the widget HTML**. In `lib/generation/scene-generator.ts`:

- `generateWidgetContent` (L1184-1331) generates the HTML and then calls `extractWidgetConfig` (L1306, L1336-1347), which only pulls the JSON out of the embedded `<script id=\"widget-config\">` — it does **not** include element IDs / classes.
- That config is then passed to `generateWidgetTeacherActions` (L1309-1315, L1352-1376), whose prompt receives `JSON.stringify(widgetConfig)` (L1363). The HTML never reaches the prompt.

As a result, the model can only rely on the hard-coded \"selector conventions\" in `lib/prompts/templates/widget-teacher-actions/system.md` (`#{var}-slider`, `#n1` / `#edge-n1-n2`, `#step-1-control`, etc.) to **guess** the target, and these guesses often don't line up with the elements actually produced in the generated HTML.

Examples:
- HTML has `#score-val` but the model emits `#score-display` (doesn't exist) → highlight is a no-op.
- Game HTML has `.pairing-rules` / `#active-zone`, but the model produces different selectors from the convention → spotlight / annotation lands on nothing.

## Proposal

When generating widget teacher actions, extract an inventory of interactable elements from the generated HTML (id + key class + role + associated text label) and feed it into the prompt, so the model **selects** targets from real elements instead of guessing from a convention.

Concrete entry points:
- In `lib/generation/scene-generator.ts` `generateWidgetContent`, in addition to `extractWidgetConfig`, add an `extractInteractiveElements(html)` step and pass the result to `generateWidgetTeacherActions`.
- In `lib/prompts/templates/widget-teacher-actions/user.md`, add an `{{elementInventory}}` variable and instruct the model to pick `target` from that inventory.
- Keep the existing selector conventions as a fallback.

Expected outcome: teacher-action `target` selectors go from \"guessed by convention\" to \"chosen from the real HTML\", eliminating no-op highlights / `setState`s caused by non-existent selectors, and significantly improving the action-to-UI sync in interactive scenes.

yanpgwang:
## Summary

- Fixes THU-MAIC/OpenMAIC#824. Teacher actions (`widget_highlight` / `widget_setState` / `widget_annotation` / `widget_reveal`) frequently misfire because the interactive-actions prompt gives the model only `widgetConfig` and forces it to guess `target` selectors from prompt conventions (`#{var}-slider`, `#start-btn`). The generated widget HTML routinely uses real ids the conventions don't cover (`#score-val`, `#launch-btn`, `#active-zone`), so highlights land on nothing and the widget doesn't animate along with the narration. Game widgets are hit hardest because their conventions cover almost nothing.
- Extract an element inventory (ids + notable classes + `aria-label` / `role` / `data-step-id` / `name` / `type`) from the generated HTML and thread it into `INTERACTIVE_ACTIONS` under a new `{{elementInventory}}` variable. The system prompt now instructs the model to pick `target` from the inventory first and fall back to the naming conventions only when nothing there fits.
- Persist the inventory on the runtime `InteractiveContent` so pro-mode `regenerate_scene_actions` sees the real selectors, not a stale guess. Refresh it inside `edit_interactive_html` so a subsequent action regen picks up the post-edit dom. And compute it on the fly in `generateSceneActions` when a legacy scene lacks it, so courses generated before this change benefit without a data migration.
- Filter Tailwind / utility class tokens so the 30-slot notable-classes cap keeps semantic hooks like `.pairing-rules`, `.dna-card`, `.btn-launch` from being crowded out by layout tokens.

## Empirical check

Same DNA base-pairing game widget, one shot with inventory vs. one without (same html, same prompt otherwise):

| input | target hits real element | action count |
|---|---|---|
| no inventory (old behaviour) | 0/1 (`#start-btn` — not in html) | 1 highlight, rest fell back to speech-only |
| with inventory | 3/3 (`#active-zone`, `#card-a`, `#check-btn`) | 3 highlights |

Matches the ratio reported in the linked issue.

## Test plan

- [x] `pnpm test tests/lib/agent tests/generation tests/prompts` (46 files / 259 tests pass, incl. two new files: `extract-interactive-elements.test.ts`, `interactive-inventory-fallback.test.ts`)
- [x] `tsc --noEmit` clean for changed paths
- [x] End-to-end: generate an 8-scene interactive course locally, inspect the emitted teacher-action `target`s against the widget HTML — inventory selectors are the ones consumed
- [ ] Pro-mode smoke: open the ProMode toggle on an existing course, ask the agent to `edit_interactive_html` then `regenerate_scene_actions`, confirm the new actions target ids from the edited HTML (feature-flag `NEXT_PUBLIC_MAIC_EDITOR_ENABLED=true`)


wyuc:
**Review — the direction is exactly right; one design change before merge (and it dissolves three of the findings below at once)**

First: this is the right fix for #824. Grounding `target` selection in what the HTML actually contains, instead of hoping naming conventions hold, is the approach we want — and the empirical 0/1 → 3/3 check plus the no-migration legacy fallback are appreciated. The extractor itself is carefully built (quoted-`>` handling, unquoted attrs, the explicit `(no interactive elements detected)` sentinel, cap logic) and the tests cover the right things.

The findings all cluster around one design choice: **persisting the derived inventory instead of computing it where it's consumed.**

**🟠 Main ask — drop the persisted `elementInventory` field; compute it at prompt-build time**

The only consumer is `generateSceneActions`, which already holds `content.html` and already contains the on-the-fly extraction as a fallback. Making that the *only* path (always `extractInteractiveElements(content.html)`) is strictly simpler — it deletes the schema fields on `InteractiveContent`/`GeneratedInteractiveContent`, the threading through `edit_interactive_html` details and `apply-regenerate` — and it eliminates by construction three findings that exist today:

1. **Same-turn staleness (the real P2 here):** `sceneContextMap` is built once per request and never written back (`app/api/agent/edit/route.ts`), so `edit_interactive_html` → `regenerate_scene_actions` *in the same agent turn* reads the pre-edit html **and** pre-edit inventory — and the persisted-inventory-wins precedence in `generateSceneActions` guarantees the stale inventory reaches the prompt. The comments in `edit-interactive-html.ts` and `apply-regenerate.ts` currently claim the opposite ("a subsequent `regenerate_scene_actions` sees the post-edit selectors"). With recompute-at-use the inventory at least always matches whatever html the tool actually reads. (The html half of the staleness is pre-existing — a `setSceneContext`-style write-back would fix it properly if you want to go further, otherwise please correct the comments to "on the next turn".)
2. **Born-stale mismatch:** `generateWidgetContent` persists `postProcessInteractiveHtml(html)` but extracts the inventory from the *raw* html — the LaTeX delimiter rewrite can make the two disagree from birth. Recomputing from the stored html makes this impossible.
3. **Server path already loses the field:** `createSceneWithActions` persists `InteractiveContent` without `elementInventory` (the parallel `scene-builder.ts` path keeps it), so for server-generated classrooms the persistence half of the feature is silently defeated today and every later regen takes the fallback anyway — i.e. production would already be running the recompute-only design.

Timing matters: once this merges, courses start persisting the field and removing it becomes a data-compat question. Cheap to change now, expensive later.

**🟠 P2 (both reviewers flagged independently) — strip HTML comments; guard unterminated `<script>`**

`<!-- <button id="old-btn"> -->` currently becomes a phantom inventory entry, and the system prompt tells the model to *prefer* inventory entries — inverting the PR's purpose for exactly the commented-out markup LLM widgets love to emit. Add `.replace(/<!--[\s\S]*?-->/g, '')` alongside the script/style strip, and consider dropping everything after an unmatched `<script` open so a truncated generation doesn't inventory ids inside `innerHTML` template strings.

**🟡 P3 — `attrAt` re-scans the flat attribute string**

For `<button id="go" aria-label="try name=alpha or id=fake">`, `attrAt(attrs, 'name')` matches inside the aria-label value and fabricates `name=alpha`. The outer tag regex already tokenizes attributes correctly — iterate that grammar into a name→value map instead of per-attribute regexing.

**🟡 P3 — interpolated values need whitespace collapsing and a length cap**

Quoted attribute values can span newlines and are interpolated verbatim into `user.md`, so an odd/hostile aria-label can forge extra inventory lines or fake prompt sections. Collapse whitespace and truncate each value (~120 chars) when building lines.

**🟡 P3 — the utility-class filter false-drops semantic hooks these widgets actually use**

`grid-`, `fill-`, `text-`, `content-`, `row-`, `col-`, `select-`, `ring-` kill `.grid-cell`, `.fill-blank`, `.text-input`, `.content-area`, `.row-label`, `.select-btn`, `.ring-carbon` — precisely the id-less repeated elements the class section exists for. Consider keeping a class when `.name` appears in the page's own `<style>` block, or only filtering when the suffix looks like a Tailwind value token (digit/scale/color).

Happy to re-review quickly — the core of this is wanted, and with the persistence dropped the diff gets smaller, not bigger.

yanpgwang:
Addressed in 269a51e.

- **Persisted `elementInventory` dropped.** `InteractiveContent` / `GeneratedInteractiveContent` no longer carry the field; `generateSceneActions` always recomputes from `content.html`. This dissolves the born-stale mismatch (post-processed html vs raw-html extraction), same-turn staleness (`sceneContextMap` is not written back so a same-turn `regenerate_scene_actions` reads pre-edit html + pre-edit inventory), and the silent divergence where `createSceneWithActions` was already dropping the field on the server path. Comment in `apply-regenerate.ts` corrected to note the same-turn staleness of `sceneContextMap`.
- **Extractor: strip HTML comments** (`.replace(/<!--[\s\S]*?-->/g, '')`) and **drop content after an unmatched `<script` open** so commented-out markup and template-string ids can't forge phantom inventory rows.
- **`attrAt` false matches fixed.** Attributes are now parsed into a name→value map by walking the tag grammar (`ATTR_TOKEN_REGEX`), so `aria-label="try name=alpha or id=fake"` no longer lifts `name=alpha` / `#fake`.
- **Attribute values are cleaned before interpolation.** Whitespace is collapsed and each value truncated to ~120 chars, so a multi-line or hostile aria-label can't fake prompt sections.
- **Utility-class filter refined.** Semantic hooks (`.grid-cell`, `.fill-blank`, `.text-input`, `.select-btn`, `.ring-carbon`, etc.) survive when they are declared in the page's own `<style>` block — we collect styled class names before stripping and skip the filter for them.

Test coverage added for each item (`extract-interactive-elements.test.ts`), and the persistence path tests updated to reflect the new always-recompute invariant. Full suite green apart from the unrelated pre-existing `provider-config.test.ts` TTS failure.

wyuc:
**Re-review of `269a51e` — all six items verified fixed; one last gap surfaced by the second pass**

Verified each fix empirically against the new head (extractor run directly + full suites: 45 files / 261 tests green):

- ✅ **Persistence dropped** — `elementInventory` is gone from all five threading sites (types, edit tool details, apply-regenerate, scene-builder), and `generateSceneActions` recomputes from `content.html` with the sentinel fallback. The corrected `sceneContextMap` staleness comment reads right now.
- ✅ **Comments / unterminated `<script>`** — `<!-- <button id="old-btn"> -->` and ids inside a truncated script's template strings no longer produce inventory rows.
- ✅ **Attribute map** — `aria-label="try name=alpha or id=fake"` no longer fabricates `name=alpha` / `#fake`.
- ✅ **Value cleaning** — newlines collapse (a hostile aria-label can't forge an "Elements with id:" section) and long values truncate at ~120 chars with an ellipsis.
- ✅ **Styled-class heuristic** — `.grid-cell` survives when declared in the page's `<style>` block while `w-4` is still dropped.

**🟠 One remaining P2 — data-only stable targets never enter the inventory**

Elements that carry `data-step-id` / `data-action` but no `id` (and no surviving semantic class) are invisible: the extractor only surfaces those attributes as decoration on id rows. That's exactly the shape of the documented procedural-skill contract — `interactive-actions/system.md:74` and `procedural-skill-content/system.md:264` both specify `[data-step-id="step-1"]` step rows, which typically have no id. A procedural widget whose rows are all data-only can even get the `(no interactive elements detected)` sentinel while the system prompt says to prefer the inventory — degrading the one widget family the conventions already served well.

Suggested fix: a third inventory section, e.g.

```
Stable data attributes:
[data-step-id="step-1"] <li>
[data-action="check"] <button>
```

capped like the others, deduped on the attribute value. (A test with an id-less `<li data-step-id="step-1">` would lock it in.)

That's the last item from our side — once it's in, this is good to merge.

yanpgwang:
Addressed in 149c42b — new **Stable data attributes** section for id-less `data-step-id` / `data-action` targets.

- A new `Stable data attributes:` block emits `[data-step-id="step-1"] <li>` / `[data-action="check"] <button>` lines for each id-less element, deduped on the attribute value and capped at 30 entries (same shape as the id / class caps).
- Elements that already carry an id keep those data-* attributes as row decoration only — no duplicate row, so the section stays scoped to the actual gap.
- Test coverage in `extract-interactive-elements.test.ts`: an id-less `<li data-step-id="step-1">` now shows up, an id + `data-action="reset"` element still uses only the id row, and an all-data-only page no longer hits the `(no interactive elements detected)` sentinel.

wyuc (approved):
Round-3 verified: the new **Stable data attributes** section behaves exactly as intended against direct extractor runs — id-less `<li data-step-id="step-1">` rows surface as `[data-step-id="step-1"] <li>`, id-carrying elements aren't duplicated into the new section, dedup and the 30-slot cap hold. CI (unit + E2E + CodeQL) is green on the merged-with-main head. (For the record: a local 3-test failure I hit during verification was an artifact of my checkout borrowing a stale `@openmaic/dsl` build — the failing paths are untouched by this PR and pass on CI's proper workspace build.) Three rounds, fast turnarounds, and the diff got simpler each time — thanks for the excellent collaboration. Merging.

PR #830, opened 2026-07-01
fix(editor): show per-line loading while the batch "regenerate all TTS" runs

wyuc:
## Problem

In the timeline editor, clicking the batch **"regenerate all TTS"** control (the "voice all" button) only spun the button and set a global in-flight flag. Each speech line's status row (preview / regenerate) kept showing its previous status, so the batch gave **no per-line feedback** — even though the *single-line* regenerate path correctly shows a "generating" state.

## Fix

- Track the ids being regenerated in `ActionsBar` (`regeneratingIds`), set up-front before the first `await` and cleared in `finally`.
- Thread a `regenerating` flag down through `SpeechClip` → `SpeechTtsBar`. While set, the line shows the generating state (spinner + disabled preview/regenerate buttons), and the single-line regenerate is disabled during the batch (no concurrent double-regen).
- To avoid a ~1-frame flash back to the "not voiced" status at batch end (the parent flag clears before the async `audioExists` re-check resolves), each line latches a `batchPending` flag on the rising edge of `regenerating` (adjusted during render, not in an effect) and clears it inside its own audio re-check effect. The clear is guarded by `!regenerating` (and `regenerating` is an effect dep, so a stale pre-batch check is cancelled at batch start), so **only the batch-end re-check clears the latch** — a stale/mid-batch check can't drop it early. The re-check is wrapped so an IndexedDB rejection still clears the latch (the row can never wedge in the generating state), and `ttsRefresh` is always bumped in the `finally` so every line re-checks at batch end, including the all-fail batch case.

Net: clicking the batch control now puts every queued line into the generating state for the batch duration, then each transitions cleanly to voiced / not-voiced with no flash and no stuck state.

## Scope

- Pure UI-state change in `components/edit/ActionsBar/ActionsBar.tsx` (+62 / −11). No changes to synthesis, storage keys, or the audio pipeline.
- Empty-text and id-less speech lines are excluded from the batch (as before) and are never marked regenerating.

## Pre-existing behavior (unchanged, noted for reviewers)

- If the *entire* batch fails (e.g. managed TTS disabled), no explicit error is surfaced — lines return to their prior status (no longer any stuck loading). A dedicated error state/toast would be a separate product decision.

## Verification

- `tsc --noEmit`: no new errors.
- `eslint` on the changed file: clean.
- Reviewed to convergence across independent reviewers (3 rounds): two P2 findings surfaced and fixed (end-of-batch flash; stuck generating-state on IndexedDB rejection / stale mid-batch check), final round zero findings.


wyuc:
## Ready for review

Fix: in the timeline editor, clicking the batch "regenerate all TTS" ("voice all") control did not put each speech line's status row into the generating/loading state (previously only the single-line regenerate did).

**Change** (only `components/edit/ActionsBar/ActionsBar.tsx`, +62/−11):
- `ActionsBar` tracks the ids being regenerated and passes a `regenerating` flag down to each `SpeechTtsBar` → during the batch the line shows the generating state (spinner + disabled buttons); the single-line regenerate is disabled during a batch to avoid concurrent double-regen.
- Each line uses a `batchPending` latch (set on the rising edge of `regenerating` during render, cleared by that line's own audio re-check effect) to remove the ~1-frame flash back to "not voiced" at batch end. The clear is guarded by `!regenerating` with `regenerating` in the effect deps, so only the batch-end re-check clears the latch; the re-check has error handling so an IndexedDB read failure can't wedge the row in the generating state.

No changes to synthesis, storage keys, or the audio pipeline; empty-text / id-less lines are excluded from the batch.

**Verification**: `tsc --noEmit` no new errors; `eslint` clean; reviewed to convergence across independent reviewers over 3 rounds (two P2 findings fixed along the way: end-of-batch flash, and a stuck generating-state on rejection).

Ready for review 🙏

sophietao20-star (commented):
Tested locally, all behaves as expected.
- “Regenerate all TTS” sets queued lines to generating state
- disables per-line preview / regenerate during batch
- returns cleanly to voiced / not-voiced state without flash or stuck loadingTTS failure can leave a speech line stuck in the generating state.
- TTS failures can return rows to their prior voiced / not-voiced status without adding new failure UI.

sophietao20-star (approved):
Tested locally, all behaves as expected for the stated scope.

wyuc:
Thanks for testing and approving! 🙏

On the note about a TTS failure leaving a line stuck in the generating state — I traced this and wanted to align on the exact path:

- A **normal synthesis failure** (the request errors, or returns null because managed TTS is off) is caught per-line in the batch loop, and the batch always re-checks every line in its `finally` (the `ttsRefresh` bump runs even when nothing synthesized). So a failed line resolves back to "not voiced" rather than staying in the generating state — this is the case your local test exercised.
- The one path that *can* wedge a row is a synthesis request that **never settles** (a hung `POST /api/generate/tts` — there's currently no client-side timeout on it). Because the batch awaits each line sequentially, a hang blocks the loop and the `finally` never runs, so the queued lines stay in the generating state. This is a pre-existing unbounded await (the single-line regenerate has the same exposure); this PR just makes it more visible as a per-line generating indicator instead of only a spinning button.

If you have a specific repro where a *normal* (settling) failure keeps a line stuck, please share it and I'll fix it here. Otherwise I'd suggest bounding each synthesis with a timeout/abort so a hang can't wedge the row — happy to fold that into this PR, or do it as a small follow-up, whichever you prefer.

sophietao20-star:
Thank you for the attention! It's my fault that the earlier conclusion was not accurate. The actual behavior is that after the batch TTS calls fail in the backend, there can be an ~3 min delay before rows return to their prior voiced / not-voiced status, which made it look like the frontend was stuck in generating. From a product perspective, I think this delay is acceptable, so I consider this PR good to go.

wyuc:
Thanks for digging into it and confirming — agreed the ~3-min backend-failure delay is a separate, acceptable concern. Appreciate the thorough review! I'll squash-merge once the post-update CI finishes.

PR #832, opened 2026-07-01
feat(dsl): own element-level normalization & defaults; wire the generator (#787)

wyuc:
Picks up the remaining open item under **#787** — element-level normalization / defaults — the companion to the #817 validators (**B-1**) and #825 migration registry (**B-2**). Where `validate*` *reports* on a document, `normalize*` *repairs* one, so producers stop carrying their own imperative "fix up the output" pass (the generator's `fixElementDefaults` being the clearest example).

## What lands

- **`@openmaic/dsl/src/normalize.ts`** — pure, zero-dep `normalizeElement` / `normalizeSlide` / `normalizeScene` / `normalizeStage` + a canonical `ELEMENT_DEFAULTS`. It:
  - fills the required **content** fields a producer may have left off (text `defaultFontName`/`defaultColor`/`content`, image/shape `fixedRatio`, shape `fill`, line `style`/`color`/`points`);
  - derives geometry-dependent fields (a line's `start`/`end`, a shape's `viewBox`/`path`) from the element box;
  - **fails loud** on a present-but-wrong-typed field instead of silently resetting it.
  - Pure, non-mutating, idempotent.
- **Static defaults ship on the JSON Schema.** `ELEMENT_DEFAULTS` is the single source of truth; the same values ride out as `@default` annotations on the generated schema (so non-TS consumers get them too), pinned to the constant by a lockstep test.
- **Generator wired.** `fixElementDefaults` drops its per-type fixups for `normalizeElement`. Image aspect-ratio reconciliation stays there — it needs the resolved asset's real dimensions, a producer concern the DSL deliberately does not own (the "media-specific reconciliation" carve-out noted on #787).

## Scope boundary

`normalize*` owns element **content** + derivable geometry. It does **not** fill or check the base identity/geometry every element shares (`id`, `left`/`top`/`width`/`height`/`rotate`) — those are producer-supplied (the `id` is typically assigned downstream of this pass) and left to `validate*` / the JSON Schema. normalize and validate are complementary.

## Two latent bugs fixed while canonicalizing

- **shape `viewBox`**: the old missing-`viewBox` fallback emitted a *string* `"0 0 w h"`, but every consumer reads `viewBox[0]`/`viewBox[1]` as an array — normalize emits the contract's `[w, h]` pair.
- **line `start`/`end`**: the old fallback derived **absolute** `[left, top]` coordinates, but the renderer positions the line container at `(left, top)` and reads `start`/`end` as **local** offsets (`getLineElementPath` / `getElementRange`), double-offsetting the line. normalize derives them in the local frame.

## Behavior change

A malformed (present-but-wrong-typed) element now **throws** from `normalizeElement` rather than being silently coerced. The generator wraps the call so a bad element degrades gracefully (log + keep raw) instead of aborting a whole scene/course — it runs on unreliable model output and on legacy slides the old pass itself produced (e.g. the string `viewBox` above). Direct callers of `normalizeElement` get the strict, fail-loud contract.

## Notes

- `@openmaic/dsl` stays **zero-runtime-dep**; the schema `@default` emission is build-time only.
- Minor bump `@openmaic/dsl` `0.2.0 → 0.3.0`.
- `test/normalize.test.ts` covers defaults, derivation (incl. the local-frame line regression), fail-loud coercion, purity/idempotency, generic app-widened `normalizeScene`, schema conformance of normalized output, and the `ELEMENT_DEFAULTS ↔ schema` lockstep.

Closes the normalize/defaults item under **#787**.


lib/generation/scene-generator.ts
```diff
@@ -542,112 +552,46 @@ function fixElementDefaults(
...
+    // pass was best-effort and never threw). Direct callers of `normalizeElement`
+    // still get the strict, fail-loud contract.
+    let normalized: PPTElement;
+    try {
+      normalized = normalizeElement(el);
+    } catch (err) {
+      log.warn(
+        `Element normalization failed, keeping raw element: ${err instanceof Error ? err.message : String(err)}`,
+      );
+      return el;
```
cosarah: Catching every `normalizeElement()` error and returning the raw element keeps malformed line payloads alive. If the model emits `start: null` or `end: null`, this path preserves the bad element instead of repairing or dropping it, and downstream code still does direct index access (`getElementRange()`, `BaseLineElement`, export helpers) on `start[0]` / `end[0]`. That means the 'graceful degradation' path can still crash playback/export for a single bad line.

packages/@openmaic/dsl/src/normalize.ts
```diff
@@ -0,0 +1,309 @@
...
+}
+
+function normalizeShape(el: Raw): PPTShapeElement {
+  return {
+    ...el,
+    viewBox: pair(el, 'viewBox', () => [geom(el, 'width'), geom(el, 'height')]),
+    path: strOrDerive(el, 'path', () => rectPath(geom(el, 'width'), geom(el, 'height'))),
+    fill: str(el, 'fill', ELEMENT_DEFAULTS.shape.fill),
+    fixedRatio: bool(el, 'fixedRatio', ELEMENT_DEFAULTS.shape.fixedRatio),
+  } as PPTShapeElement;
```
cosarah, line 195: `normalizeShape()` only repairs top-level shape fields. A payload like `text: {}` still normalizes successfully, but downstream code assumes `el.text.content` is a string and passes it into `formatHTML()` during PPTX export. Since this PR positions `normalize*` as the repair boundary for required content fields, nested `ShapeText` looks like a remaining hole here.

cosarah (changes requested):
I found two issues that should be addressed before merge.

lib/generation/scene-generator.ts
wyuc: Good catch — fixed in f6b762fe. The catch no longer keeps the raw element; the policy is now repair-or-drop: JSON nulls are stripped first (a model's `start: null` becomes an absent field, which normalize then derives), and an element that still fails normalization is discarded with a warning instead of being passed through. No malformed payload reaches the unguarded `start[0]` readers anymore.

packages/@openmaic/dsl/src/normalize.ts
wyuc: Fixed in f6b762fe. `normalizeShape` now normalizes a present `text` overlay: `content` / `defaultFontName` / `defaultColor` / `align` get the same missing→default, wrong-typed→throw semantics as the element's own required fields (align defaults to 'middle', matching the renderers). `text: {}` now normalizes to a fully-defaulted ShapeText, so `formatHTML(el.text.content)` always sees a string. The new defaults ship on the schema as `@default` and are pinned by the lockstep test.

cosarah (approved):
LGTM, verified locally

PR #833, opened 2026-07-01
fix(quiz): render formulas in quiz text

### LINKED ISSUE #771: [Bug]: quiz中题目公式渲染问题
### Bug Description

在quiz的部分，有一些题目的公式是不能被正确渲染的，比如说会出现：“a(x-2)+b(2-x)^2=a(x-2)+b(x-2)^2=(x-2)(a+b(x-2))”。

### Steps to Reproduce

1、限制模型的输出，强制输出为latex格式。2、quiz的渲染兼容文本格式转换为latex格式

### Expected Behavior

都能被渲染为latex格式

### Actual Behavior

直接呈现为文本格式了

### Deployment Method

### Browser

_No response_

### Operating System

_No response_

```shell

```

### Additional Context

_No response_

dpersek:
## Summary

- Render math in quiz question text, options, submitted answers, AI comments, and analysis blocks instead of displaying formula-like strings as plain text.
- Add a focused quiz math-text helper that supports explicit LaTeX delimiters and conservative delimiter-free standalone algebra detection.
- AI-assisted contribution; self-review and independent Codex review completed before submission.

## Related Issues

Fixes #771.

## Changes

- Add quiz math-text segmentation/rendering with KaTeX fallback behavior.
- Wire readonly quiz display fields through the renderer while preserving answer state, grading, and persistence behavior.
- Add focused tests for explicit delimiters, delimiter-free algebra, malformed math fallback, currency/prose preservation, and subtraction in dollar-delimited math.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Render a quiz scene with formula-like question text such as `a(x-2)+b(2-x)^2=a(x-2)+b(x-2)^2=(x-2)(a+b(x-2))`.
2. Verify the quiz display renders math instead of raw formula text.
3. Verify ordinary quiz prose, currency-like text, and malformed math remain readable as text.

- `pnpm exec prettier --check lib/quiz/math-text.ts tests/quiz/math-text.test.ts components/scene-renderers/quiz-view.tsx`
- `pnpm vitest run tests/quiz/math-text.test.ts` (11 tests)
- `pnpm lint` (0 errors; 15 existing warnings outside touched files)
- `pnpm exec tsc --noEmit`
- `pnpm test` (223 files, 1844 tests)
- `pnpm check`
- `git diff --check`

### Evidence

- [x] CI-style local checks passed (`pnpm check`, `pnpm lint`, and TypeScript type checking)
- [ ] Manually tested locally

Note: I did not attach browser screenshots or recordings yet; verification is from helper coverage plus lint, TypeScript, and Vitest. I can add visual evidence before marking this draft ready if maintainers prefer.

## Checklist

## Scope / non-goals

- Does not change quiz generation, grading, answer persistence, or model prompts.
- Does not introduce new user-facing copy.
- Display-mode math is enabled only for question stems and analysis blocks; option labels, submitted answers, and comments render inline to avoid compact-layout breaks.

### DIFF
--- a/components/scene-renderers/quiz-view.tsx
+++ b/components/scene-renderers/quiz-view.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useState, useMemo, useCallback, useEffect, useRef } from 'react';
+import { memo, useState, useMemo, useCallback, useEffect, useRef } from 'react';
 import { motion, AnimatePresence } from 'motion/react';
 import {
   PieChart,
@@ -23,6 +23,7 @@ import type { QuizQuestion } from '@/lib/types/stage';
 import { useDraftCache } from '@/lib/hooks/use-draft-cache';
 import { SpeechButton } from '@/components/audio/speech-button';
 import { gradeChoiceQuestions, isShortAnswer, type QuestionResult } from '@/lib/quiz/grading';
+import { renderQuizMathText } from '@/lib/quiz/math-text';
 import {
   clearSubmitted,
   draftKey,
@@ -40,6 +41,43 @@ interface QuizViewProps {
   readonly questions: QuizQuestion[];
   readonly sceneId: string;
 }
+
+const QuizMathText = memo(function QuizMathText({
+  text,
+  className,
+  allowDisplayMode = false,
+}: {
+  text: string;
+  className?: string;
+  allowDisplayMode?: boolean;
+}) {
+  const segments = useMemo(() => renderQuizMathText(text), [text]);
+  if (segments.length === 1 && segments[0].type === 'text') {
+    return <span className={className}>{segments[0].value}</span>;
+  }
+
+  return (
+    <span className={className}>
+      {segments.map((segment, index) => {
+        if (segment.type === 'text') {
+          return <span key={index}>{segment.value}</span>;
+        }
+
+        return (
+          <span
+            key={index}
+            className={cn(
+              allowDisplayMode && segment.displayMode
+                ? 'block my-1 overflow-x-auto [&_.katex-display]:!my-0'
+                : 'inline-block align-baseline [&_.katex-display]:!my-0',
+            )}
+            dangerouslySetInnerHTML={{ __html: segment.html }}
+          />
+        );
+      })}
+    </span>
+  );
+});
 
 /** Call /api/quiz-grade for a single short-answer question. */
 async function gradeShortAnswerQuestion(
@@ -255,7 +293,7 @@ function SingleChoiceQuestion({
                   isReview && !isCorrectOpt && !selected && 'text-gray-400 dark:text-gray-500',
                 )}
               >
-                {opt.label}
+                <QuizMathText text={opt.label} />
               </span>
               {isReview && isCorrectOpt && (
                 <CheckCircle2 className="w-5 h-5 text-emerald-500 shrink-0" />
@@ -360,7 +398,7 @@ function MultipleChoiceQuestion({
                   isReview && !isCorrectOpt && !isSelected && 'text-gray-400 dark:text-gray-500',
                 )}
               >
-                {opt.label}
+                <QuizMathText text={opt.label} />
               </span>
               {isReview && isCorrectOpt && (
                 <CheckCircle2 className="w-5 h-5 text-emerald-500 shrink-0" />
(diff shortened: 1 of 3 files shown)


wyuc:
**Review — solid core, but the standalone-math heuristic has one content-corrupting bug and a couple of false-positive classes to fix before merge**

The overall design is good: the delimiter segmentation is correct (adjacent `$a$$b$`, `$$` vs `$` tie-breaking, escaped `\$`, unclosed delimiters, currency like `costs $5 and $10` all behave right, and the tests cover those), XSS is handled (only KaTeX output reaches `dangerouslySetInnerHTML`, `trust` stays off, `throwOnError` falls back to plain text), and the `memo`/`useMemo` integration in `quiz-view.tsx` is clean. The findings below are all in `isLikelyStandaloneMathText` / the delimiter-free path. Each rendered output quoted below is from running the string through `katex.renderToString` with the PR's exact options.

**🔴 P1 — `%` silently truncates rendered content**

`%` is in `FORMULA_CHAR_RE`, but in LaTeX it starts a comment, and `strict: false` suppresses KaTeX's `commentAtEnd` warning — so nothing throws and everything after the first `%` is dropped:

- `50% = 1/2` → renders as **`50`**
- `25%*4 = 100%` → renders as **`25`**

Percentage/fraction conversions are exactly what math-quiz options contain, and the corruption is silent. Fix: escape `%` → `\%` before `katex.renderToString` on the delimiter-free path (or drop `%` from `FORMULA_CHAR_RE` so such strings stay plain text).

**🟠 P2 — short prose prompts render as smashed-together math (two variants, both confirmed)**

1. The `words.length > 2` prose guard passes single-instruction-word prompts: `Solve x=2` → **`Solvex=2`**, `Simplify x^2` → **`Simplifyx2`** — the instruction word becomes juxtaposed italic variables and the space disappears.
2. Any `\command` match waives the word cap entirely, so prose *mentioning* a command is classified as math: `Use \sqrt{2} not 2 here` → **`Use√2not2here`**. AI comments and analysis text (both now flow through `QuizMathText`) are the realistic source of this shape.

Suggested direction: keep the word cap unconditionally (commands shouldn't waive it, maybe just raise it), and/or require that letters mostly appear as single-char variables adjacent to operators rather than as words.

**🟡 P3 — code-like options render as italic math**

`i = i + 1`, `Answer = True` pass the check (≤2 long words + `=`) and render in math italics with math spacing — wrong typography for programming quizzes. Cosmetic; rejecting `;`, `==`, `+=`-style operators would cover most of it.

**🟡 P3 — the #771 pattern still isn't covered when the algebra is embedded in a sentence**

`FORMULA_CHAR_RE` requires the *whole* field to be formula characters, so `解方程 a(x-2)+b(2-x)^2=a(x-2)+b(x-2)^2` (bare algebra inside prose — the original issue's symptom) still renders raw. That conservatism is defensible for this PR, but worth stating the scope in the PR description, and the robust long-term fix is prompt-side: require the generator to emit `$…$` delimiters so the heuristic is only a fallback.

The P1 and both P2 variants would make good additions to `tests/quiz/math-text.test.ts` as regression cases. Happy to re-review quickly once they're in.

dpersek:
Thanks for the detailed review. I addressed all four buckets in follow-up commit `b043cd9`:

- Escaped literal `%` before KaTeX rendering so delimiter-free inputs like `25%*4 = 100%` no longer get truncated by LaTeX comment semantics.
- Tightened standalone detection so short prose prompts stay text, including `Solve x=2`, `Simplify x^2`, and `Use \sqrt{2} not 2 here`.
- Added code-like assignment guards so options such as `i = i + 1` and `Answer = True` stay plain text.
- Added embedded compact-algebra rendering so prose with a #771-like formula renders only the formula segment and preserves the surrounding sentence text.

Proof run locally:

- `pnpm exec prettier --check lib/quiz/math-text.ts tests/quiz/math-text.test.ts components/scene-renderers/quiz-view.tsx`
- `pnpm vitest run tests/quiz/math-text.test.ts` (19 tests)
- `git diff --check`
- `pnpm lint` (0 errors; same pre-existing warnings outside this patch)
- `pnpm exec tsc --noEmit`
- `pnpm test` (223 files / 1852 tests)
- `pnpm check`

wyuc (approved):
Re-verified `b043cd9` empirically — all four buckets are fixed:

- `50% = 1/2` and `25%*4 = 100%` now render in full (percent escaping works, no more comment truncation)
- `Solve x=2` / `Simplify x^2` / `Use \sqrt{2} not 2 here` stay plain text
- `i = i + 1` / `Answer = True` stay plain text
- The embedded path handles the #771 shape inside prose: `解方程 a(x-2)+b(2-x)^2=…` renders only the formula segment — and URLs, currency, dates, ratios, and `Score: 85/100` all stay text

Legit math still renders (`E = mc^2`, `x^2 + 2x = 3`, the full #771 expression). One known trade-off worth recording: bare trig prose like `sin(x)+cos(x)=1` now stays plain text because `sin`/`cos` count as prose words — that's the right conservative default (generators should emit `\sin` or `$…$` delimiters), just noting it's intentional.

Tests pass (40/40 in `tests/quiz/`). Merging — thanks for the fast, thorough turnaround!

PR #835, opened 2026-07-02
fix(docker): fix `postinstall` script failure due to missing `sync-maic-importer.mjs` file in docker build

Lee-Flier:
**… building docker image**

**## Summary**
Fixes a Docker build failure caused by the `postinstall` script crashing due to a missing `sync-maic-importer.mjs` file. Adds a `COPY scripts/ ./scripts/` step to the `deps` stage to ensure required scripts are present during `pnpm install`.

**## Changes**

- **Dockerfile Modification**: Added `COPY scripts/ ./scripts/` to the `deps` stage to copy the `scripts/` directory (containing `sync-maic-importer.mjs`) into the container during the dependencies installation phase.

**## Type of Change**

- Bug fix (non-breaking change that fixes an issue)

**## Verification**

**### Steps to reproduce / test**

1. Build the Docker image using the modified `Dockerfile`:
2. Check the build logs for the `deps` stage to confirm `pnpm install` completes successfully without errors.
3. Verify that the `sync-maic-importer.mjs` file exists in `/app/scripts` within the built container:

**### What you personally verified**

- Confirmed that the Docker build succeeds end-to-end without encountering the `MODULE_NOT_FOUND` error.
- Manually inspected the container file system to confirm the presence of `sync-maic-importer.mjs` in the expected location.
- Tested the build on both Linux and macOS environments to cover potential platform differences.
- Verified that all CI checks (e.g., `pnpm check`, `pnpm lint`, `tsc --noEmit`) pass locally.

**### Evidence**

- CI passes (`pnpm check && pnpm lint && npx tsc --noEmit`)
- Manually tested locally

**## Checklist**

- My code follows the project's coding style
- I have performed a self-review of my code
- I have added/updated documentation as needed *(Not required for this change)*
- My changes do not introduce new warnings
- I have tested all new and existing functionality locally


wyuc (changes requested):
**Confirmed the root cause — but the fix is incomplete for clean-clone builds; one more line needed.**

The `COPY scripts/ ./scripts/` addition correctly fixes the `deps` stage: `postinstall` ends with `node scripts/sync-maic-importer.mjs`, which today isn't in the copied context, so `pnpm install --frozen-lockfile` crashes. ✅

However, on a fresh clone the build now fails one stage later:

1. In `deps`, the postinstall sync step writes the importer bundle to `/app/public/vendor/maic-importer` (`scripts/sync-maic-importer.mjs`).
2. The `builder` stage only carries `node_modules` and `packages` over from `deps`; `public/` comes from `COPY . .` — the build context.
3. `/public/vendor/maic-importer` is gitignored, so a clean checkout's context doesn't contain it.
4. `pnpm build` runs `scripts/assert-vendor-maic-importer.mjs` first, which exits 1 when `public/vendor/maic-importer/index.js` is absent.

Your local end-to-end build likely passed because your working tree had run `pnpm install`, so `public/vendor/` existed in the docker context (`COPY . .` picks it up; it's not in `.dockerignore`).

This also matters at runtime, not just for the assert: the runner stage serves `public/` copied from `builder`, and the PPTX importer is URL-imported from `/vendor/maic-importer/index.js` — so the bundle genuinely needs to be in `builder`.

Suggested fix — carry the deps-stage artifact over, after `COPY . .` in the builder stage:

```dockerfile
COPY --from=deps /app/public/vendor ./public/vendor
```

(Alternatively `RUN node scripts/sync-maic-importer.mjs` after `COPY . .` also works, since `packages/@openmaic/importer/dist` is carried over from deps — but the COPY is cheaper.)

Could you verify with a fresh clone (or a `git clean -xdf` tree) and `docker build .`? With that extra line the build should pass end-to-end and the bundle should land in the final image.

Lee-Flier:
@wyuc Thank you so much for your thorough and insightful feedback! Your analysis pinpointed the incomplete fix in the original PR. I've carefully verified your suggested fix and implemented the changes. I've updated the PR with these additional changes and verification steps. The complete modified Dockerfile sections are now included for clarity.

wyuc (approved):
Re-reviewed the updated Dockerfile change.

I verified the fix against a clean Docker context where `public/vendor/maic-importer` is absent locally:

- `COPY scripts/ ./scripts/` lets the deps-stage `postinstall` run `scripts/sync-maic-importer.mjs` successfully.
- `COPY --from=deps /app/public/vendor ./public/vendor` carries the generated importer bundle into the builder stage, so `scripts/assert-vendor-maic-importer.mjs` passes before `next build`.
- A full `docker build --progress=plain -t openmaic-pr-835-review:local .` completed successfully.
- I also tested a local merge of this PR onto current `origin/main`; that merged state also completed a full Docker build successfully.
- The final runner image contains a non-empty `/app/public/vendor/maic-importer/index.js` readable by the runtime user.

No further blocking issues from my side.

PR #836, opened 2026-07-02
feat(playback): resume scene playback position

danishsshaikh:
## Summary

This splits the per-scene resume behavior out of #820 into a smaller focused PR.

After maintainer feedback, this intentionally avoids the continuous-time scrubber model. Resume is aligned with the discrete classroom playback engine by storing an action-level position rather than a millisecond timestamp.

## Behavior

- Saves the last meaningful playback action position per scene.
- Restores that action-level position when returning to a scene during the same browser session.
- Uses `sessionStorage` only.
- Validates stored resume data before applying it.
- Ignores stale/invalid resume entries safely.
- Avoids timestamp-based resume, duration estimates, and continuous progress scrubbing.
- Clears saved resume for a scene once that scene completes.

## Safety

This PR intentionally keeps resume conservative for cumulative/live action families.

Resume is skipped for saved positions that would require reconstructing prior unsafe state, including:

- `widget_*`
- `discussion`
- `play_video`
- `wb_*` before the saved point

The scene can still be revisited and played normally. The conservative behavior only means this PR does not resume from the middle of those stateful scenes yet.

This avoids duplicating widgets/discussion sessions/video state or reconstructing partial whiteboard state without the reset/fast-forward infrastructure planned for the later sentence/action-level navigation work.

## Not included

- No scene-wide seek slider.
- No continuous-time scrubbing.
- No total duration timeline.
- No transcript-line click navigation yet.
- No prev/next sentence controls yet.
- No millisecond/audio-currentTime resume.

Transcript-line jumps and prev/next sentence navigation can be handled in a follow-up aligned with #818.

## Validation

Automated:

- `git diff --check` passed
- `pnpm test tests/playback/resume-position.test.ts` passed
- `pnpm test tests/lib/playback/engine-cursor.test.ts` passed
- `pnpm exec tsc --noEmit` passed
- `pnpm build` passed
- `pnpm lint` passed with existing warnings only
- `pnpm check` passed

Manual server validation:

- Safe speech-only scene resumes from the saved action boundary.
- Completed scenes clear saved resume and start from the beginning.
- Widget scenes do not resume into the middle of widget state.
- Discussion scenes do not resume into the middle of live/session state.
- Video scenes do not restore video `currentTime` or resume mid-video.

### DIFF
--- a/components/edit/PlaybackChromeRoot.tsx
+++ b/components/edit/PlaybackChromeRoot.tsx
@@ -18,7 +18,13 @@ import { SceneSidebar } from '@/components/stage/scene-sidebar';
 import { Header } from '@/components/header';
 import { CanvasArea } from '@/components/canvas/canvas-area';
 import { Roundtable } from '@/components/roundtable';
-import { PlaybackEngine, computePlaybackView } from '@/lib/playback';
+import {
+  PlaybackEngine,
+  clearPlaybackResumePosition,
+  computePlaybackView,
+  loadPlaybackResumePosition,
+  savePlaybackResumePosition,
+} from '@/lib/playback';
 import type { EngineMode, TriggerEvent, Effect } from '@/lib/playback';
 import { ActionEngine } from '@/lib/action/engine';
 import { createAudioPlayer } from '@/lib/utils/audio-player';
@@ -27,7 +33,6 @@ import { useWidgetIframeStore } from '@/lib/store/widget-iframe';
 import type { AudioIndicatorState } from '@/components/roundtable/audio-indicator';
 import type { Action, DiscussionAction, SpeechAction } from '@/lib/types/action';
 import { cn } from '@/lib/utils';
-// Playback state persistence removed — refresh always starts from the beginning
 import { ChatArea, type ChatAreaRef } from '@/components/chat/chat-area';
 import { agentsToParticipants, useAgentRegistry } from '@/lib/orchestration/registry/store';
 import type { AgentConfig } from '@/lib/orchestration/registry/types';
@@ -80,6 +85,7 @@ export const PlaybackChromeRoot = forwardRef<PlaybackChromeRootHandle, PlaybackC
       setCurrentSceneId,
       generatingOutlines,
       outlines,
+      stage,
     } = useStageStore();
     const failedOutlines = useStageStore.use.failedOutlines();
     const generationComplete = useStageStore.use.generationComplete();
@@ -457,6 +463,7 @@ export const PlaybackChromeRoot = forwardRef<PlaybackChromeRootHandle, PlaybackC
       // on a scene's first visit and silently drop every widget action. Looking it
       // up per-send always sees the live registration.
       const sceneIdForWidget = currentScene.id;
+      const playbackResumeScopeId = stage?.id ?? currentScene.stageId;
       const widgetSendMessage = (type: string, payload: Record<string, unknown>) =>
         useWidgetIframeStore.getState().getSendMessage(sceneIdForWidget)?.(type, payload);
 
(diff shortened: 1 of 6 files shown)


danishsshaikh:
One design clarification/question on the conservative handling for `widget_*`, `discussion`, `play_video`, and `wb_*` scenes:

In this PR I intentionally skip saved resume positions that would require reconstructing prior stateful actions. So the user can still return to those scenes and play them normally, but playback resumes from the beginning rather than jumping back into the middle of widget/discussion/video/whiteboard state.

I wanted to confirm whether this is the behavior you’d prefer for the resume-only PR.

My reasoning was that even though this PR does not implement transcript-line click navigation yet, action-level resume is still effectively a jump to a saved action boundary. For simple speech scenes that is safe, but for stateful/live actions:

- `widget_*` may require iframe setup/state to already exist
- `discussion` may involve live session/model/message side effects
- `play_video` would require media `currentTime` restoration, which this PR intentionally avoids
- `wb_*` before the saved point would require deterministic reset/silent replay to avoid missing or duplicated whiteboard state

So I kept the resume behavior conservative and deferred those cases to the later action-navigation work where reset/fast-forward and generation-token handling can be designed more deliberately.

Should the resume-only PR keep this conservative behavior, or would you prefer some limited resume behavior for these scene types as well?

wyuc:
Thanks for calling out the design question explicitly.

For a resume-only PR, yes: the conservative behavior is the right choice. I would not try to restore mid-state for `widget_*`, `discussion`, `play_video`, or `wb_*` here. Those cases need reset/fast-forward semantics, media `currentTime` handling, discussion/session behavior, and generation-token guards, which belong in the broader action-navigation work rather than this focused slice.

That said, this also means the standalone UX value of #836 is intentionally limited. Since stateful scenes correctly fall back to start-from-beginning, this PR mainly helps simple narration/action sequences. That is useful as playback plumbing, but I would not expand this PR as an end-user playback UX feature.

My preference is to keep this very narrow: either keep only the small reusable resume foundation needed by the later action-navigation work, or fold this into the #843 follow-up once the broader reset/fast-forward model is settled. Please do not add visible controls or more complex state reconstruction in this PR.

danishsshaikh:
Thanks for the clarification here, @wyuc 

That makes sense to me. I’ll keep this PR narrow and won’t expand.

Since the broader action-navigation work is now in #843, I’ve addressed the related resume/restore issue there instead:

* fixed the refresh restore case where a saved later speech action could be overwritten by the first speech during mount
* added regression coverage for the `speech -> whiteboard actions -> speech` repro shape
* kept the conservative guards for `widget_*`, `discussion`, and `play_video`
* kept whiteboard reset / prefix replay and stale async callback protections in the broader action-navigation path
* removed the visible bottom-toolbar sentence controls from #843 for now, so the core plumbing can be reviewed separately from final UX

I’ll treat #836 as the smaller resume-only foundation, but avoid adding more here. If #843 becomes the preferred path for the combined playback/navigation plumbing, I’m fine closing this PR as superseded by #843.

wyuc (changes requested):
The conservative unsafe-action guard leaves a stale safe resume point behind.

For example, after saving speech B, playback can advance through a `wb_*`/widget/video/discussion action. `savePlaybackResumePosition` then returns early for the unsafe boundary but never deletes the previously stored speech-B position. On revisit, `loadPlaybackResumePosition` still accepts that old safe prefix and resumes in the middle instead of starting the stateful scene from the beginning as this PR promises.

Please clear the scene's stored position as soon as an unsafe boundary is crossed, and add a sequential regression test (save safe speech -> save unsafe/later snapshot -> load returns null). The branch now conflicts with current playback work, and your own note says #843 contains the broader corrected restore path; please rebase this narrow foundation or close it as superseded rather than having maintainers resolve it.

danishsshaikh:
Thanks @wyuc :)

 The stale safe resume position you identified is a real issue in this narrow implementation: after playback crosses an unsafe/stateful action, the previously stored position should be cleared rather than remaining available for a later mid-scene resume.

Since #843 has already merged the broader action-navigation and resume/restore path, and this branch now conflicts with the current playback implementation, I’m going to close #836 as superseded rather than rebase and maintain duplicate playback plumbing.

Future resume/restore fixes and regression coverage should go through the merged #843 path. Thanks for catching this.

PR #837, opened 2026-07-02
feat(document): rename to Document Parsing, show supported formats, extend MinerU support

### LINKED ISSUE #838: [Feature]: Rename Settings to Document Parsing, show supported formats, and extend MinerU support to match actual capabilities
### Problem or Motivation

After #741 landed multi-format course-material upload, three gaps remain in the Settings and upload UX, plus one significant finding during implementation:

**1. Settings UI still called "PDF Parsing".** The sidebar entry and copy in `components/settings/` still say "PDF 解析" / "PDF Parsing", but the section now configures a **document** extraction pipeline that handles DOCX/PPTX/TXT/MD alongside PDF. Users can't tell they're in the right place to configure non-PDF extraction.

**2. No visibility into what each provider supports.** `unpdf`, `MinerU`, and `MinerU (Cloud)` support **different** file formats, but the Settings panel doesn't show this at all. A user configuring `unpdf` has no way to know it only handles PDF; a user configuring `MinerU` doesn't know DOCX/PPTX/XLSX/images are options.

**3. Homepage upload picker doesn't filter by selected provider.** `components/generation/generation-toolbar.tsx` uses a hardcoded `COURSE_MATERIAL_ACCEPT` string listing every format the union of extractors supports. When the user has `unpdf` selected, the native file dialog still lets them pick a `.docx` — which is then rejected only at generation time.

**4. MinerU whitelist is significantly narrower than actual capabilities.** While probing MinerU Cloud during (3) we found #741's whitelist misses most of MinerU's supported formats. Empirically verified via `/file-urls/batch`:

| Format | MinerU Cloud | Currently allowed |
|---|---|---|
| PDF | ✅ | ✅ |
| DOCX / PPTX / XLSX | ✅ | DOCX, PPTX only |
| DOC / PPT / XLS (legacy OLE) | ✅ | ❌ |
| Images: PNG / JPG / WebP / GIF / BMP / JP2 | ✅ | ❌ |
| TXT / MD | ❌ (handled locally by `plain-text` extractor) | (correctly local) |

MinerU self-host (v3.1+) supports the same set **except** legacy OLE (`.doc/.ppt/.xls`), which is cloud-only.

### Proposed Solution

- Rename "PDF Parsing" → "Document Parsing" in the sidebar. Keep the old `pdfSettings` i18n key to avoid breaking downstream.
- Add a "Supported Formats" badge row above the existing "Features" row in the settings panel.
- Derive `<input accept>` from the currently-selected provider + always-on `plain-text` extractor, so the native file dialog grays out unsupported formats and drag-drop rejects them at runtime.
- Extend the MinerU whitelists to match actual capability, split into a self-host list and a cloud list.
- Consolidate the provider→MIME/extension lookup tables into a single source of truth (currently duplicated across `mime.ts`, `extractors/pdf.ts`, and `mineru-cloud.ts` with a "keep in sync" comment). Add a drift-guard test.
- Ship a probe script (`scripts/probe-mineru-cloud.mjs`) so future format additions can be validated against real Cloud behavior with one command.

### Alternatives Considered

- **Extend the accept list statically to the union across all providers**, keeping picker permissive. Rejected because it hides real per-provider capability from the user until generation-time, exactly the current pain.
- **Only fix (1)–(3), leave the MinerU whitelist as-is**. Rejected once probing revealed the extent of the gap — most users would keep hitting rejects on formats MinerU can actually handle.
- **Take the intersection of self-host + cloud so both providers advertise the same list**. Rejected: cloud legitimately handles `.doc/.ppt/.xls` and users on cloud should get that benefit.

### Area

UI / UX

### Additional Context

- Follow-up to #741 (multi-format upload).
- MinerU official docs: https://opendatalab.github.io/MinerU
- Implementation in PR #837.

yanpgwang:
## Summary

Rename the Settings sidebar entry from "PDF Parsing" to "Document Parsing", surface each provider's supported formats as badges, filter the upload picker's `accept` string by the currently-selected extractor, and extend the MinerU whitelist to match actual capabilities (verified by probing `mineru.net/file-urls/batch`).

## Related Issues

Closes #838
Follow-up to #741

## Changes

- Add "Supported Formats" badge row in `components/settings/pdf-settings.tsx` (unpdf: PDF; MinerU: PDF/DOCX/PPTX/XLSX + images; MinerU Cloud: same + DOC/PPT/XLS).
- Sidebar label switches to `documentParsingSettings` i18n key (kept `pdfSettings` untouched for downstream compat).
- `components/generation/generation-toolbar.tsx`: `<input accept>` derives from `[pdfProviderId, 'plain-text']`; drag-drop uses `isMimeSupportedByProviders`.
- Drop attached course-material file when the user switches to a provider that no longer supports it (prevents stale file from failing only at generate-time).
- Extend `MINERU_CLOUD_SUPPORTED_EXTENSIONS` (in `lib/pdf/mineru-cloud.ts`) and per-provider MIME lists (in `lib/document/extractors/pdf.ts`) with `.doc/.ppt/.xls/.xlsx` and images (`png/jpg/webp/gif/bmp/jp2`). Self-host and cloud lists split — legacy OLE is cloud-only.
- Consolidate the provider→MIME/extension whitelists into a single `DocumentFormat[]` registry in `lib/document/mime.ts`; `lib/document/extractors/pdf.ts` and `lib/pdf/mineru-cloud.ts` import from it. `types.ts`: `supportedMimeTypes` widened to `readonly string[]`.
- `normalizeDocumentMimeType`: map known alias MIMEs (e.g. `image/jpeg2000`, `text/x-markdown`) to canonical, and fall back to the extension when the reported MIME is unknown but the extension resolves cleanly.
- Derive MinerU zip fallback image regex from the shared image-MIME list so newly added image formats can't slip through.
- i18n: 8 locales get new `documentParsingSettings` + `supportedFormats` keys; `courseMaterialSizeLimit` and `unsupportedCourseMaterial` copy updated to be provider-agnostic.
- Add `scripts/probe-mineru-cloud.mjs` — utility script that uploads one sample per format and reports outcome. Requires `MINERU_CLOUD_API_KEY`.
- Tests: add drift-guard (PROVIDER_SUPPORTED_MIME_TYPES vs registry `supportedMimeTypes`) + alias/unknown-MIME normalization; update existing tests to new whitelists.

## Type of Change

### Steps to reproduce / test

1. Open Settings → sidebar shows "文档解析" / "Document Parsing".
2. Click each provider (unpdf, MinerU, MinerU (Cloud)) — "Supported Formats" badges reflect actual capability.
3. Homepage upload popover: select `unpdf` → picker filters to `.pdf/.txt/.md`; switch to `MinerU (Cloud)` → picker gains `.docx/.pptx/.xlsx/.doc/.ppt/.xls` + images.
4. Attach `.docx` under `MinerU (Cloud)`, then switch to `unpdf` → file is cleared and error shown inline.
5. Run `MINERU_CLOUD_API_KEY=… node scripts/probe-mineru-cloud.mjs` to verify Cloud support end-to-end.

- All 14 formats accepted by MinerU Cloud (via probe script; `.txt`/`.md` correctly rejected as unsupported by MinerU and handled locally by `plain-text`).
- Dynamic `accept` string changes as user toggles extractor in the popover (verified via Playwright).
- Provider-switch effect drops attached file and shows error (verified via Playwright).
- Drag-drop with unsupported MIME (`application/msword` under unpdf) fires the runtime toast.
- Drift-guard test fails intentionally when I flip one MIME in either file — pinning works.
- **Not verified**: MinerU self-host support for `.xlsx` and images. The environment had no self-host deployment reachable. Support is asserted from official docs (MinerU v3.1+); to empirically verify, someone with a v3.1+ deployment can point `PDF_MINERU_BASE_URL` at it and run the probe script (needs a small tweak to hit `/file_parse` instead of the Cloud endpoint).
  - **Update (2026-07-06)**: Verified end-to-end against a MinerU self-host deployment by hitting `/api/extract-document` with real files. `.docx` returns extracted text; `.xlsx` returns an HTML table faithful to the source; `.png` returns OCR'd text plus embedded image data. Self-host handles all the previously-doc-asserted formats.

### Evidence

- [x] Manually tested locally (Playwright + probe script)
- [x] Screenshots attached below

<details><summary>Settings panel — supported formats (MinerU Cloud)</summary>

Badges: PDF · DOCX · PPTX · XLSX · PNG · JPG · WebP · GIF · BMP · JP2 · DOC · PPT · XLS

</details>

<details><summary>MinerU Cloud probe summary</summary>

```
✅ sample.pdf: accepted
✅ sample.doc: accepted
✅ sample.docx: accepted
✅ sample.ppt: accepted
✅ sample.pptx: accepted
✅ sample.xls: accepted
✅ sample.xlsx: accepted
✅ sample.png / .jpg / .jpeg / .webp / .gif / .bmp / .jp2: accepted
❌ sample.txt: batch-rejected — unsupported file type
❌ sample.md: batch-rejected — unsupported file type
```

</details>

## Checklist

- [x] I have performed a self-review of my code (see `/code-review` findings addressed in the diff)
- [x] I have added/updated documentation as needed (in-code comments; no external docs)


wyuc:
@sophietao20-star could you help run an independent E2E validation for this PR when you have bandwidth?

Suggested focus:
- Settings -> Document Parsing shows each provider's supported formats.
- MinerU Cloud, once configured, accepts Office/image formats such as `.docx`.
- Switching the extractor back to `unpdf` removes unsupported formats and clears an already-attached `.docx` course material.

I ran a local pass, but an independent maintainer E2E check would be helpful before this lands.

wyuc (changes requested):
Thanks for pushing this forward. The Settings-side provider/format display and the toolbar accept behavior are generally heading in the right direction, but I found one correctness issue that should be fixed before merge.

[P2] `normalizeDocumentMimeType()` now trusts the filename extension for any unknown non-empty MIME.

At `lib/document/mime.ts`, the new fallback path normalizes any unknown MIME to the MIME implied by the filename extension. For example, an upload with:

```ts
{ mimeType: 'application/x-msdownload', fileName: 'lesson.pdf' }
```

normalizes to `application/pdf` and passes `isMimeSupportedByProviders(..., ['unpdf'])`. The backend `/api/extract-document` then uses that normalized MIME to select a document extractor, so this expands the upload whitelist beyond the intended browser fallback cases.

Please restrict extension-based fallback to generic browser MIME fallbacks such as `application/octet-stream` / zip-style Office fallbacks, plus curated known aliases like `application/x-msword`; otherwise preserve/reject the unknown MIME instead of treating it as a supported document type.

Lower-priority note: `components/generation/generation-toolbar.tsx` derives upload `accept` and file validation only from `pdfProviderId`. If a previously selected provider becomes unavailable after credentials/server config change, the toolbar can still accept that provider's formats until a later server-side failure. This is less important than the MIME issue because store sync recovers some stale states, but it is worth tightening if you touch this area.

Validation I ran locally:
- `pnpm test tests/document/mime.test.ts tests/document/extractor-registry.test.ts tests/document/mineru-cloud.test.ts tests/document/extract-document-route.test.ts` -> 4 files / 22 tests passed.
- Browser E2E on `http://127.0.0.1:3003`: Document Parsing settings show provider formats; MinerU Cloud accepts `.docx`; switching back to `unpdf` removes `.docx` from `accept` and clears the already-attached `.docx` with an unsupported-format error.

yanpgwang:
@wyuc P2 addressed in e20b19a: `normalizeDocumentMimeType` no longer trusts the filename extension for arbitrary unknown MIMEs.

Changes:
- Moved `application/x-msword` into `DocumentFormat.aliasMimes` so it round-trips through the existing curated `canonicalFromAlias` path.
- Extension-based fallback is now restricted to the generic upload MIMEs (`octet-stream` / zip family) — the original Office-ZIP fallback case is preserved.
- Any other unknown MIME is returned verbatim so provider whitelists can reject it.
- Added a regression test: `{application/x-msdownload, lesson.pdf}` stays as `application/x-msdownload` and fails `isMimeSupportedByProviders(..., ['unpdf'])`.
- End-to-end check: `curl -F 'file=@spoof.pdf;type=application/x-msdownload' -F providerId=unpdf` against `/api/extract-document` now returns `INVALID_REQUEST: No document extractor supports MIME type "application/x-msdownload"`.

Test/lint status: `pnpm test tests/document/*.test.ts` (the 4 files you ran) → 23/23 passing (was 22, +1 for the new spoof test); `npx tsc --noEmit` clean; `pnpm lint` 0 errors; `pnpm check` (prettier) clean.

The lower-priority toolbar note (accept derived only from `pdfProviderId`) — agreed it's worth tightening; I'll leave it for a follow-up PR rather than expanding this one's scope.

yanpgwang:
@wyuc Thanks for the review — the P2 is fixed in `e20b19a`.

**Change:** removed the "unknown MIME + matching extension → trust the extension" fallback in `normalizeDocumentMimeType()`. Normalization now follows three paths only:
1. Missing / generic upload-fallback MIME (octet-stream, zip-family) → use the extension;
2. Known alias (canonical + the registry's curated `aliasMimes`) → map to the canonical MIME;
3. Otherwise → return the reported MIME verbatim, letting the provider whitelist reject it.

Legit cross-platform aliases (e.g. `application/x-msword` → `.doc`) now live in `DocumentFormat.aliasMimes`, so they still round-trip. Your `{application/x-msdownload, lesson.pdf}` case now keeps its original MIME, and `isMimeSupportedByProviders(..., ['unpdf'])` returns `false` — pinned with a regression test (`tests/document/mime.test.ts`, 6/6 passing).

The lower-priority note (toolbar deriving `accept`/validation only from `pdfProviderId`) is deferred to a follow-up PR and not included here. Re-requested review — PTAL.

sophietao20-star:
Tested locally, no blocking issues was found. Verified:
- Settings -> Document Parsing shows each provider's supported formats for `unpdf`, `MinerU`, and `MinerU Cloud`.
- MinerU Cloud, once configured, accepts Office/image formats including `docx` `pptx` `pdf` `png`
- Attached a `.docx` course material in the browser, then switched the extractor back to `unpdf`; the file was cleared and the UI prompted to switch extractors(as expected). 
- TXT/Markdown parse successfully end-to-end through the local plain-text extractor.
- Self-hosted MinerU was configured via `PDF_MINERU_BASE_URL`, and DOCX/PPTX/XLSX parse successfully end-to-end through `/api/extract-document`.
- Unsupported/spoofed uploads are rejected: `.docx.exe` and unknown MIME + `.pdf` do not bypass validation.

Minor UX notes:
- self-hosted MinerU Supported Formats may over-promise PDF/image support for lightweight `mineru-api` installs unless `mineru[pipeline]` or `mineru[core]` is installed.
- missing-pipeline failures currently surface raw MinerU JSON(Screenshot below); a friendlier app-level error would be easier to act on. e.g: 当前 MinerU 服务未安装 pipeline/core 依赖，无法解析 PDF/图片。截屏2026-07-09 14 21 11

yanpgwang:
Thanks for the thorough testing! Addressed both minor UX notes in 9e6e72f:

- **Raw MinerU JSON on missing-pipeline failures** — self-hosted MinerU errors now go through `describeSelfHostedMinerUError`, which detects the missing-dependency signatures (`ModuleNotFoundError` / `ImportError` / `Device string must not be empty`) and surfaces an actionable message instead of the raw traceback/JSON: it tells the operator to install `mineru[pipeline]` or `mineru[core]` (or switch to MinerU Cloud). Unknown errors keep their detail but are truncated so the UI stays readable.
- **Self-hosted 'Supported Formats' over-promising PDF/image** — added a caveat under the format badges (shown only for the self-hosted MinerU provider) noting that PDF/image parsing requires the `mineru[pipeline]`/`mineru[core]` extras, and that lightweight `mineru-api` installs may only handle Office documents.

Added unit tests for the error describer, and `tests/document` + `tests/i18n` pass.

Note: the underlying backend default (`hybrid-auto-engine` → configurable `PDF_MINERU_BACKEND`, defaulting to `pipeline`) is being handled separately in #844, so this PR stays focused on the UX layer — the two are complementary.

wyuc (dismissed):
Approved from the code-review side.

I rechecked the latest head (`9e6e72f`): the MIME spoofing blocker is fixed, the self-hosted MinerU UX follow-up is reasonable, and the relevant local checks pass on my side:
- `pnpm test tests/document/*.test.ts`
- `pnpm test tests/i18n/*.test.ts`
- `pnpm lint`
- `npx tsc --noEmit`

Please fix the remaining CI failure before merging. The failing `Lint, Typecheck & Unit Tests` job is currently only Prettier on `components/settings/pdf-settings.tsx`; `pnpm exec prettier components/settings/pdf-settings.tsx --write` should address it.

wyuc (approved):
Re-approved after the latest update.

The follow-up commit only fixes Prettier formatting in `components/settings/pdf-settings.tsx`, and all checks are green now.

PR #843, opened 2026-07-03
feat(playback): add action-level navigation

danishsshaikh:
## Summary

Implements sentence/action-level playback navigation for classroom slides.

This follows the #818 direction: transcript/action-boundary navigation instead of a continuous-time scrubber. Users can click transcript lines, use previous/next sentence controls, and see compact `Line i / N` progress without introducing timestamp-based seeking.

## Related Issues

Related to #818  
Follow-up to #836

## Changes

- Added action-level transcript navigation.
- Added previous/next sentence controls in the playback toolbar.
- Added compact `Line i / N` progress.
- Added action-level `jumpToAction` playback engine support.
- Added playback generation-token guards for stale async audio/timer/TTS callbacks after jumps.
- Added deterministic whiteboard reset/silent replay for safe action jumps.
- Cleared transient spotlight/laser effects on jumps and skipped them during silent replay.
- Guarded unsafe jumps across `widget_*`, `discussion`, and `play_video`.
- Preserved paused action position across scene switches.
- Added tests for transcript navigation, action jumps, stale callback handling, whiteboard replay, spotlight boundaries, unsafe guards, and resume behavior.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Open a generated classroom with multiple narrated lines.
2. Start playback and click different transcript/notes lines.
3. Use previous/next sentence controls while paused and while playing.
4. Verify `Line i / N` updates as playback moves.
5. Test a scene with whiteboard actions and jump backward/forward across those actions.
6. Test a scene with spotlight/laser effects and jump back before the effect.
7. Test scenes containing `widget_*`, `discussion`, or `play_video` actions and confirm unsafe jumps are guarded.
8. Jump to the last transcript line and let playback finish.

### What i personally verified

- Transcript line clicks jump to the expected action/line.
- Previous/next sentence controls work and respect boundaries.
- `Line i / N` updates correctly.
- Playing jumps continue from the target line.
- Paused jumps position the cursor without unexpected autoplay.
- Paused scene switch restores the saved action boundary.
- Completed scenes clear/restart correctly.
- Whiteboard jumps reset/replay without duplicate drawings.
- Spotlight does not retrigger early after jumping back.
- Spotlight still appears during normal playback when its own action is reached.
- Unsafe widget/discussion/video jump targets are guarded.
- Normal playback still works through widget/discussion/video scenes.
- Jumping to the last line completes normally.
- Rapid jump/pause/play interactions did not produce stale audio, double-advance, or transcript freeze.

### Evidence

- [x] Manually tested locally/server

Screenshot 2026-07-03 at 2 23 55 PM

Automated validation run:

- `git diff --check` passed
- `pnpm test tests/playback/action-navigation.test.ts` passed: 23 tests
- `pnpm test tests/chat/lecture-notes-navigation.test.ts` passed
- `pnpm test tests/lib/playback/engine-cursor.test.ts` passed
- `pnpm test tests/audio/audio-player-leak.test.ts` passed
- `pnpm exec tsc --noEmit` passed
- `pnpm build` passed
- `pnpm lint` passed with existing warnings only
- `pnpm check` passed

## Checklist

### DIFF
--- a/components/canvas/canvas-area.tsx
+++ b/components/canvas/canvas-area.tsx
@@ -44,7 +44,6 @@ export function CanvasArea({
   showStopDiscussion,
   onStopDiscussion,
   hideToolbar,
-  actionNavigation,
   isPendingScene,
   isCourseComplete,
   isGenerationFailed,
@@ -268,7 +267,6 @@ export function CanvasArea({
           onTogglePresentation={onTogglePresentation}
           showStopDiscussion={showStopDiscussion}
           onStopDiscussion={onStopDiscussion}
-          actionNavigation={actionNavigation}
         />
       )}
     </div>
--- a/components/canvas/canvas-area.tsx
+++ b/components/canvas/canvas-area.tsx
@@ -44,6 +44,7 @@ export function CanvasArea({
   showStopDiscussion,
   onStopDiscussion,
   hideToolbar,
+  actionNavigation,
   isPendingScene,
   isCourseComplete,
   isGenerationFailed,
@@ -267,6 +268,7 @@ export function CanvasArea({
           onTogglePresentation={onTogglePresentation}
           showStopDiscussion={showStopDiscussion}
           onStopDiscussion={onStopDiscussion}
+          actionNavigation={actionNavigation}
         />
       )}
     </div>
--- a/components/canvas/canvas-toolbar.tsx
+++ b/components/canvas/canvas-toolbar.tsx
@@ -6,6 +6,8 @@ import {
   ChevronRight,
   Play,
   Pause,
+  SkipBack,
+  SkipForward,
   PencilLine,
   LayoutList,
   MessageSquare,
@@ -50,6 +52,14 @@ export interface CanvasToolbarProps {
   readonly onToggleAutoPlay?: () => void;
   readonly playbackSpeed?: number;
   readonly onCycleSpeed?: () => void;
+  readonly actionNavigation?: {
+    currentLine: number;
+    totalLines: number;
+    canGoPrev: boolean;
+    canGoNext: boolean;
+    onPrev: () => void;
+    onNext: () => void;
+  };
 }
 
 /* Compact control button */
@@ -108,6 +118,7 @@ export function CanvasToolbar({
   onToggleAutoPlay,
   playbackSpeed = 1,
   onCycleSpeed,
+  actionNavigation,
 }: CanvasToolbarProps) {
   const { t } = useI18n();
   const canGoPrev = currentSceneIndex > 0;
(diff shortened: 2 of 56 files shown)


wyuc (changes requested):
Thanks for pushing action-level playback control forward. I think the underlying model is valuable: action-index navigation, safe-jump guards, whiteboard-prefix reconstruction, and the focused tests are all moving in the right direction.

I ran a local E2E pass on this PR with an imported deterministic classroom covering:
- `speech -> wb_open -> wb_draw_text -> speech -> speech`
- toolbar next/previous sentence navigation
- notes-row click navigation
- whiteboard prefix reconstruction before the target speech
- an unsafe `discussion` boundary where the later speech should not be jumpable

The core jump behavior worked in that flow, and the unsafe guard behaved as expected. I also ran:

```bash
pnpm exec vitest run tests/playback/action-navigation.test.ts tests/chat/lecture-notes-navigation.test.ts
```

`2 files / 25 tests` passed.

However, I do not think this is ready to merge yet.

## Blocking: resume-on-refresh does not restore the saved action position

Repro from the local E2E run:

1. Open a classroom with speech actions separated by whiteboard actions.
2. Jump to the second speech line.
3. Confirm `sessionStorage` has the expected action-level resume position, e.g. `actionIndex: 3`, `actionId: "speech-2"`.
4. Reload the classroom.

Expected: the classroom restores to the saved speech line (`Line 2 / 3`) without autoplay.

Actual: the UI comes back at `Line 1 / 3`, and the resume state is overwritten back to the first speech (`actionIndex: 0`, `actionId: "speech-1"`). This breaks the resume behavior introduced by the PR and needs a regression test around mount/refresh restore.

## UX: please split or feature-flag the visible controls

Separately, I do not think the current visible toolbar UX should merge as-is. The `Line x / y` sentence controls in the bottom toolbar feel bolted onto the existing classroom UI and duplicate part of what the notes panel is already doing. They also introduce a third visible navigation layer next to scene navigation and playback controls, which makes the hierarchy harder to understand.

Could you split this so the core playback/navigation plumbing can be reviewed independently from the final visible UX? Concretely, I would suggest:

1. keep the action navigation helpers, engine jump plumbing, safe-jump guards, and tests;
2. fix the resume-on-refresh issue above;
3. remove or feature-flag the new visible toolbar controls for now;
4. iterate the final UX separately so it fits the existing classroom / roundtable / notes interaction model.

I would be happy to re-review once the resume bug is fixed and the visible UX surface is narrowed.

danishsshaikh:
Thanks for the detailed review and for calling out the restore issue :)

I addressed the feedback in a follow-up commit: `3d9aeda fix(playback): address action navigation review feedback`.

Changes made:

* Fixed the resume-on-refresh issue. The playback chrome now reads and validates the saved action-level resume cursor before resetting scene UI state, so a saved later speech action is not overwritten by the default first-speech cursor during mount.
* Added regression coverage for the repro shape you described: `speech -> whiteboard actions -> speech`, ensuring the saved action cursor is preserved during mount/refresh.
* Removed the visible bottom-toolbar previous/next sentence controls and `Line i / N` display for now.
* Kept the core action-level navigation plumbing, transcript/notes click navigation, safe-jump guards, whiteboard prefix replay, and stale async callback guards.

Validation run:

* `pnpm test tests/playback/action-navigation.test.ts`
* `pnpm test tests/chat/lecture-notes-navigation.test.ts`
* `pnpm test tests/lib/playback/engine-cursor.test.ts`
* `pnpm test tests/audio/audio-player-leak.test.ts`
* `pnpm exec tsc --noEmit`
* `pnpm lint`
* `pnpm check`
* `git diff --check`

I’ll also run `pnpm build` and a local browser QA pass before asking for another review.

wyuc (approved):
Thanks for the follow-up. I re-reviewed the latest head (`02dbd840067ea7f9daa2d6158b00798fa5585654`) and the changes address my blocking feedback.

What I checked:

- The visible bottom-toolbar previous/next sentence controls are gone from the final PR diff; action navigation is now limited to the notes-line click path.
- The saved action resume cursor is read before scene UI reset and used as the initial action cursor, so the mount/refresh path no longer overwrites a later saved speech action with the first speech.
- The restore path still keeps the conservative guards for discussion / video / widget actions and only replays deterministic whiteboard-prefix actions before a target speech line.
- The regression coverage now includes the `speech -> whiteboard actions -> speech` resume/restore shape.

Also ran locally:

```bash
pnpm exec vitest run tests/playback/action-navigation.test.ts tests/chat/lecture-notes-navigation.test.ts
```

Result: 2 files / 26 tests passed.

Looks good to me for the narrowed core action-navigation plumbing.

wyuc:
@sophietao20-star could you help run a quick browser E2E pass on this PR before we merge?

I reviewed the code and focused regressions, and the latest head addresses the previous blocker. What still needs an independent UI pass is the actual playback interaction:

- open/import a classroom with `speech -> whiteboard actions -> speech` in a scene;
- start playback and confirm the lecture notes highlight follows the current speech action;
- click a later speech line in Notes and confirm playback jumps there, with the whiteboard prefix reconstructed correctly;
- refresh the classroom and confirm it restores to the saved later speech action instead of resetting to the first line;
- confirm the bottom toolbar does not expose the old previous/next sentence controls.

If this browser pass looks good, I think the narrowed core action-navigation plumbing is ready to merge.

wyuc:
Product E2E follow-ups from @sophietao20-star :

- Add a visible "Replay current page" control near play/pause. Today users can replay the current page by opening Notes and clicking the first line, but that path is not discoverable enough.
- Improve the hover affordance for clickable Notes rows. Instead of relying on the browser-native "Jump to this line" tooltip, use a lightweight hover state (for example a subtle purple background while keeping normal text color), while reserving the stronger purple background/text for the active playback line.

I do not think these should block this narrowed PR. The core UX change here is still limited to Notes click-to-jump plus action-level resume/restore, and the visible bottom-toolbar sentence controls were already removed. Let's track the two items above as follow-up UX polish.

danishsshaikh:
Thanks for the review, QA pass, and merge :) @wyuc @sophietao20-star

PR #160, opened 2026-03-20
feat: support multiple PDF uploads per generation

YizukiAme:
## Summary

Adds support for uploading multiple PDFs per generation session. Currently only a single PDF can be uploaded — this PR extends the full pipeline to handle up to 5 PDFs with parallel parsing, intelligent content aggregation, and graceful failure recovery.

Closes #148

### Upload Layer
- `pdfFile: File | null` → `pdfFiles: SelectedPdf[]`
- Toolbar supports multi-select, drag-and-drop of multiple files
- File deduplication by `name + size + lastModified` signature
- Validation: single file 50MB, max 5 files, total 150MB

### Document Aggregation (`lib/pdf/document-aggregator.ts` — NEW)
- **Text budget allocation**: "base guarantee + proportional" split of `MAX_PDF_CONTENT_CHARS` (50000) across files, after dynamically deducting header/separator framing chars
- **Stable image ID rewriting**: `img_${fileOrder}_${localIndex}` intermediate IDs → flattened to sequential global IDs. Pure function, no module-level state, retry-safe
- **Vision pool**: round-robin selection across files (1 per file first, then fill remaining), sorted by `visionPriority`

### Generation Preview
- Parallel PDF parsing with configurable concurrency (default: 2)
- Per-file status display: `pending → parsing → success/failed`
- Partial failure recovery: "Retry failed" or "Continue with successful files"
- Backward compatible with legacy single-file sessions

### Downstream
- `sortPdfImagesForVision()` integrated into outline-generator, scene-generator, and outlines-stream route
- No changes to `/api/parse-pdf` protocol — all multi-file semantics handled client-side

## Files Changed (12)

| File | Change |
|------|--------|
| `lib/pdf/document-aggregator.ts` | **NEW** — core aggregation logic (248 lines) |
| `lib/types/generation.ts` | Added `SelectedPdf`, `SessionPdfSource`, `PdfParseStatus` |
| `lib/utils/image-storage.ts` | Added `storePdfFiles()` helper |
| `lib/i18n/common.ts` | Added multi-PDF toolbar strings (zh/en) |
| `lib/i18n/generation.ts` | Added parsing progress, failure, retry strings |
| `app/page.tsx` | Multi-file state management + session construction |
| `components/generation/generation-toolbar.tsx` | Multi-select UI, file list, validation |
| `app/generation-preview/page.tsx` | Parallel parse + aggregate + failure gate |
| `app/generation-preview/types.ts` | Added `pdfSources` to session state |
[...]

Issue #148, opened 2026-03-20
feat: support multiple PDF uploads per generation


YizukiAme:
## Summary

Currently only a single PDF can be uploaded per generation. Users want to upload multiple PDFs and have them merged as course material.

## Current Behavior

- `pdfFile: File | null` — single file state
- One `pdfStorageKey` in IndexedDB
- One OCR parsing pass in generation-preview
- One `pdfText` fed to the generation pipeline

## Desired Behavior

- Upload multiple PDF documents
- Merged content is used as course material for generation
- UI shows a file list with individual remove buttons

## Design Decisions Needed

- **Content merging strategy**: concatenate in upload order? per-file section headers?
- **Token budget**: how to distribute `MAX_PDF_CONTENT_CHARS` across files — equal split, proportional to length, or sequential truncation?
- **OCR images**: per-file limit or shared `MAX_VISION_IMAGES` pool?
- **Max file count**: cap at N files to avoid overwhelming the LLM context?
- **UI**: drag-to-reorder? total size indicator?

## Affected Layers

| Layer | Change |
|-------|--------|
| Toolbar UI | `File \| null` → `File[]`, list display |
| IndexedDB storage | Multiple storage keys or array storage |
| Session state | Multi-file metadata |
| Generation preview | Loop OCR parsing + content merge |
| Generation pipeline | Combined text with truncation |

suzaku007:
may i know when this feature will go live?

wyuc:
Completed by #844, which landed the multi-file course-material bundle flow on `main` (ordered uploads, per-source storage/parsing, bounded text merging, and cross-document image handling). Closing this request as delivered.

PR #844, opened 2026-07-03
Feat/document bundles milestone 3

jackefn:
## Summary

Implement MAIC ETL Milestone 3 by extending course material upload from a single extracted document to a multi-file document bundle.

This PR lets users attach multiple course material files, stores each source independently, parses them through the existing document extractor boundary, and merges the extracted text/images into one generation-compatible bundle. The bundle preserves source order and source metadata, normalizes image IDs across documents, and applies shared text/vision budgets so generation can use multiple uploaded materials without breaking the existing PDF-compatible pipeline.

## Related Issues

Related to #148, #160

## Changes

- Add document bundle aggregation helpers for multi-file extracted materials.
- Add upload/session types for selected course materials and persisted document sources.
- Update the homepage upload flow from single file state to ordered multi-file material selection.
- Store each uploaded material independently in IndexedDB before navigating to generation preview.
- Preserve backward compatibility with legacy single-document `pdfStorageKey` sessions.
- Parse each uploaded source through the existing document extraction/PDF-compatible route.
- Merge extracted document text with source headers, upload order, MIME type, and page metadata.
- Allocate bounded text budgets across documents so smaller sources are not starved by larger files.
- Normalize image IDs across documents to avoid collisions like repeated `img_1`.
- Preserve original image/source metadata for citation and downstream image selection.
- Assign cross-document vision priorities so image budget is distributed across sources.
- Update prompt/image formatting paths to carry source-document metadata.
- Update generation preview/session logic to operate on course material bundles.
- Update toolbar UI and locale strings for multi-file course material upload.
- Add focused document bundle tests for text budgeting, source ordering, image ID normalization, and vision priority.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Start OpenMAIC locally and upload multiple course material files from the homepage upload control.
2. Confirm the selected files are listed, can be removed, and retain deterministic upload order.
3. Generate a classroom and confirm generation preview parses all uploaded materials before outline generation.
4. Upload materials containing images and confirm image references remain stable after bundle merging.
5. Generate with a mix of text-heavy and small documents and confirm source headers from each document appear in the bounded extracted context.

- Multiple uploaded materials are stored as independent session document sources.
- Old single-file sessions still hydrate through the legacy `pdfStorageKey` compatibility path.
- Extracted text is merged in source order with per-source section headers.
- Duplicate extractor image IDs are rewritten into stable bundle-level IDs.
- Vision image priority is distributed across source documents before taking extra images from the same source.
- Bundle tests cover text budget allocation, source ordering, image ID normalization, and cross-document vision prioritization.

### Evidence

- [ ] Manually tested locally

Local verification:

```text
pnpm vitest run tests/document/bundle.test.ts tests/document/extractor-registry.test.ts
```

## Checklist


app/generation-preview/page.tsx
```diff
@@ -256,117 +289,112 @@ function GenerationPreviewContent() {
...
-        );
+            text: parseResult.data.text as string,
+            rawTextLength: (parseResult.data.text as string).length,
+            pageCount: parseResult.data.metadata?.pageCount,
+            images,
+          });
+        }
+
+        const bundle = buildDocumentBundle(parsedParts);
+        const imageStorageIds = await storeImages(bundle.images);
```
yanpgwang, line 440: **Silent storageId misalignment.**

`storeImages` (lib/utils/image-storage.ts:55-77) wraps each base64→Blob conversion in try/catch and only pushes on success — so `storedIds.length < bundle.images.length` whenever any single image fails (bad data URL from a MinerU response, `atob` padding error, IndexedDB quota, …).

The `bundle.images.map((img, i) => ({ storageId: imageStorageIds[i] }))` below zips by index, so after the first failure every subsequent `storageId` points to the wrong image and the tail gets `undefined`. Vision mode then attaches picture A while the prompt says `img_k` → the LLM hallucinates content.

Old single-doc code had the same shape but bundle mode multiplies image counts across N sources, so this is far more likely to manifest. Suggest either making `storeImages` fail-fast (throw on the first failure) or return `Array<string | null>` so callers explicitly opt in to gaps.

app/generation-preview/page.tsx
```diff
@@ -256,117 +289,112 @@ function GenerationPreviewContent() {
...
-          });
-          throw new Error(t('generation.courseMaterialLoadFailed'));
-        }
-
-        // Wrap as a File to guarantee multipart/form-data with correct content-type
-        const pdfFile = new File([pdfBlob], currentSession.pdfFileName || 'document.pdf', {
-          type: currentSession.documentMimeType || pdfBlob.type || 'application/pdf',
-        });
+        for (const source of [...documentSources].sort((a, b) => a.order - b.order)) {
+          const documentBlob = await loadDocumentBlob(source.storageKey);
```
yanpgwang: **Extraction is fully serial across N sources.**

Each iteration awaits `loadDocumentBlob` + `POST /api/extract-document` before starting the next. For 5 PDFs at ~15s MinerU parse each, the user waits ~75s instead of ~15s. Extraction is idempotent per file and both client & server accept concurrency.

Suggested shape:

```ts
const parsedParts = await Promise.all(
  [...documentSources]
    .sort((a, b) => a.order - b.order)
    .map(async (source) => { /* per-source loadBlob + fetch */ }),
);
```

Order is re-established by `buildDocumentBundle` (it sorts `parts` by `source.order` internally), so the pre-sort here becomes optional. If you're worried about MinerU cost/rate limits, wrap in a small p-limit style pool.

yanpgwang (commented):
Overall the multi-file bundle refactor is a nice piece of work — the budget allocator, cross-source vision fairness, and legacy-session compat are all thoughtful. Below are the findings from a high-effort review. The ones I'd prioritize before merging are the silent image-misalignment bugs (1, 3, 4, 5) and the sequential extraction (2) since it's a very visible UX regression.

## Correctness

1. **`storeImages` failures misalign `imageStorageIds` → wrong storage key attached to every subsequent image.** See inline on `app/generation-preview/page.tsx`.
2. **Extraction loop is fully serial** — N files ≈ N × parseTime instead of ~1 × maxParseTime. See inline on `app/generation-preview/page.tsx`.
3. **`.slice()` on the id-rewritten text can cut mid-token**, e.g. `img_11` → `img_1` (or split a UTF-16 surrogate pair). See inline on `lib/document/bundle.ts`.
4. **`img.pageNumber || 1`** treats a legitimate `0` as missing. MinerU parser at `lib/pdf/mineru-parser.ts:81` emits `pageNumber: 0` when `meta` is absent. Use `??`.
5. **`sortDocumentImagesForVision` tiebreaks with `id.localeCompare`** → `img_10` < `img_2` in the tail (priority-0) images; once >9 images exist the non-vision order is arbitrary.
6. **`providerConfig` (containing `apiKey`) is duplicated into every `SessionDocumentSource`** — a 5-file upload writes the apiKey 6+ times to sessionStorage. Store once at session root and look up by `providerId`.
7. **`Promise.all(courseMaterials.map(async → storeDocumentBlob))`** short-circuits on first rejection while sibling writes have already committed → orphan blobs in IndexedDB until the 24h `cleanupOldImages` sweep. Repeated retries hit quota.
8. **Scene-content vision path (`lib/generation/scene-generator.ts:681`) does NOT call `sortDocumentImagesForVision`** — the outline path (`outline-generator.ts:60`, `scene-outlines-stream/route.ts:330`) does. So `visionPriority` is honored during outline generation and dropped during scene-content generation; the two paths pick different images for the same scene once `assignedImages.length > MAX_VISION_IMAGES`. (Not shown inline because the file isn't in this PR's diff.)
9. **Bundle limits (`MAX_DOCUMENT_BUNDLE_FILES=5`, `MAX_DOCUMENT_BUNDLE_TOTAL_SIZE_BYTES=150MB`) are enforced only in the toolbar**, not on `/api/extract-document`. A resumed / hand-edited session can carry N>5 sources and the server processes them one-by-one.

## Cleanup (worth a follow-up)

10. **File-fingerprint dedup key `${name}:${size}:${lastModified}` is duplicated** in `handleFilesSelect` (toolbar) and `addCourseMaterials` (app/page.tsx). One helper.

## Other observations (not blocking)

- `replaceImageIds` is O(images × textLength) and runs twice per bundle build (stableIdMap → finalIdMap). A single alternation regex + composed map cuts this to one pass.
- The two-phase id rewriting (`doc_N_img_M` → `img_K`) has a subtle collision hazard: if any original extractor id equals a target `doc_N_img_M` for another image in the same part, the second `String.replace` iteration cascades over the first. A structured placeholder scheme at the extractor boundary (e.g. `{{IMAGE:<extractorId>}}`) would avoid the free-text search-and-replace entirely.
- Session shape now carries two representations of the document set (`documentSources` + legacy `pdfStorageKey/pdfFileName/documentMimeType`), reconciled at read sites via `legacySourceFromSession`. Consider migrating at the session-load boundary and dropping the legacy fields.
- `PdfImage` type retains its PDF-centric name but now carries multi-source metadata; the prompt formatters fall back to a literal `" from PDF"` string when `sourceDocumentName` is missing. Renaming to `DocumentImage` and making `sourceDocumentName` required (default at bundle-build time) would remove that branch.
- `storeDocumentBlob = storePdfBlob` and `loadDocumentBlob = loadPdfBlob` are const-aliases — rename the definitions and keep the old names as `@deprecated` aliases if any external caller depends on them.

lib/document/bundle.ts
```diff
@@ -0,0 +1,235 @@
...
+    textContentBudget,
+  );
+
+  const flattenedImages = stableParts.flatMap((part) => part.images);
+  const finalIdMap = new Map<string, string>();
+  flattenedImages.forEach((image, index) => finalIdMap.set(image.id, `img_${index + 1}`));
+
+  const text = stableParts
+    .map((part, index) => {
+      const boundedText = replaceImageIds(part.text, finalIdMap).slice(0, textBudgets[index]);
```
yanpgwang: **`.slice` can cut through an image id (or a UTF-16 surrogate pair).**

By the time `.slice(0, textBudgets[index])` runs, `part.text` has already been rewritten to contain the final `img_N` ids. Once a bundle has ≥10 images, ids span `img_1..img_11+`. If the budget lands mid-token in a passage like `... see img_11.`, the trailing character is cut and the emitted bundle contains `... see img_1` — silently rewiring the reference to a completely different image.

Same class also cuts multi-byte UTF-16 (emoji, some CJK combined forms) → lone surrogate that later breaks JSON serialization of the SSE body.

Fix ideas:
- Slice on `part.text` first, then run `replaceImageIds` on the sliced text (image ids won't cross the boundary because they weren't inserted yet).
- Or after slicing, walk back to a whitespace/punctuation boundary.
- Or use `Intl.Segmenter` for a grapheme-safe cut.

app/generation-preview/page.tsx
```diff
@@ -256,117 +289,112 @@ function GenerationPreviewContent() {
...
-                description?: string;
-                width?: number;
-                height?: number;
-              }) => ({
+          const rawImages = parseResult.data.metadata?.pdfImages;
+          const images = rawImages
+            ? rawImages.map((img: ParsedDocumentResponseImage) => ({
                 id: img.id,
                 src: img.src || '',
                 pageNumber: img.pageNumber || 1,
```
yanpgwang: **`img.pageNumber || 1` treats a legitimate 0 as missing.**

`lib/pdf/mineru-parser.ts:81` returns `pageNumber: meta ? meta.pageIdx + 1 : 0` — a real `0` from that path gets rewritten to `1`, tying cover/unknown-page images with genuine page-1 images. `compareImagesForVision` (bundle.ts:106) then treats them as equal-priority, and prompt formatters print `from ... page 1` for cover-page figures, so the LLM cites the wrong page.

Use `img.pageNumber ?? 1`.

lib/document/bundle.ts
```diff
@@ -0,0 +1,235 @@
...
+}
+
+export function sortDocumentImagesForVision<
+  T extends Pick<PdfImage, 'visionPriority' | 'pageNumber' | 'id'>,
+>(images: T[]): T[] {
+  return [...images].sort((a, b) => {
+    const priorityDiff = (b.visionPriority ?? 0) - (a.visionPriority ?? 0);
+    if (priorityDiff !== 0) return priorityDiff;
+    if (a.pageNumber !== b.pageNumber) return a.pageNumber - b.pageNumber;
+    return a.id.localeCompare(b.id);
```
yanpgwang, line 177: **Lexicographic tiebreaker on `img_N` ids is wrong.**

All images beyond the top `MAX_VISION_IMAGES` get `visionPriority = 0`. On the same page they collapse to `a.id.localeCompare(b.id)`, which sorts `img_10` before `img_2` (and similarly for larger counts). Both `outline-generator.ts:60` and `scene-outlines-stream/route.ts:330` slice by `MAX_VISION_IMAGES` against this order, so once a bundle has >9 images the tail ("text-only") descriptions are effectively randomized.

Sort by an extracted numeric suffix (`Number(id.match(/img_(\d+)/)?.[1] ?? 0)`), or expose a stable `sourceDocumentOrder` + `insertionIndex` on the image and tiebreak on those.

app/page.tsx
```diff
@@ -329,6 +357,25 @@ function HomePage() {
...
+              size: item.size,
+              lastModified: item.lastModified,
+              mimeType: normalizeDocumentMimeType({
+                mimeType: item.file.type,
+                fileName: item.file.name,
+              }),
+              order: index + 1,
+              storageKey: await storeDocumentBlob(item.file),
+              providerId: pdfProviderId,
+              providerConfig: pdfProviderConfig,
```
yanpgwang: **`providerConfig` (with `apiKey`) is duplicated per-source.**

Each `SessionDocumentSource` gets its own `{ providerId, providerConfig: { apiKey, baseUrl } }`, in addition to the session-level `pdfProviderConfig`. A 5-file upload persists the apiKey 6+ times in sessionStorage.

Beyond storage bloat, this widens the surface for accidental logging/dumping and duplicates the credential in any code that serializes the session object. Consider storing `providerConfig` once at the session root and looking it up by `providerId` in the extract loop.

app/page.tsx
```diff
@@ -329,6 +357,25 @@ function HomePage() {
             baseUrl: providerCfg.baseUrl,
           };
         }
+
+        documentSources = await Promise.all(
```
yanpgwang: **Partial-failure leak in Promise.all + storeDocumentBlob.**

If any file's `arrayBuffer()` rejects (quota, corrupt file, revoked File), `Promise.all` short-circuits and `handleGenerate` throws — but sibling `storeDocumentBlob` calls that already resolved have already written `pdf_<nanoid>` records to `db.imageFiles`. Nothing references them because the session was never created, so they sit until the 24h `cleanupOldImages` sweep. Repeated retries accumulate multi-hundred-MB orphan blobs and eventually hit IndexedDB quota.

Either use `Promise.allSettled` and delete the resolved blobs on any rejection, or write blobs sequentially and roll back on error.

components/generation/generation-toolbar.tsx
```diff
@@ -114,17 +121,50 @@ export function GenerationToolbar({
...
     }
+
+    const existing = new Set(
+      courseMaterials.map((file) => `${file.name}:${file.size}:${file.lastModified}`),
+    );
+    const dedupedFiles = supportedFiles.filter(
+      (file) => !existing.has(`${file.name}:${file.size}:${file.lastModified}`),
+    );
+    if (dedupedFiles.length === 0) return;
+
```
yanpgwang, line 189: **Bundle limits are only enforced client-side.**

`MAX_DOCUMENT_BUNDLE_FILES=5` and `MAX_DOCUMENT_BUNDLE_TOTAL_SIZE_BYTES=150MB` are enforced here in the toolbar and nowhere else. `app/generation-preview/page.tsx` re-issues `POST /api/extract-document` for each `documentSources[i]` from a stored session — a resumed session (or one edited via devtools/session-storage-editor) can carry N>5 sources or arbitrary total size, and the server processes them one by one.

Either add a bundle-aware endpoint (`/api/extract-documents`) that receives all files and enforces count/size server-side, or minimally re-validate `documentSources.length` and aggregate size in the extract loop before hitting the network.

components/generation/generation-toolbar.tsx
```diff
@@ -114,17 +121,50 @@ export function GenerationToolbar({
...
     }
-    if (file.size > MAX_COURSE_MATERIAL_SIZE_BYTES) {
+    if (supportedFiles.some((file) => file.size > MAX_COURSE_MATERIAL_SIZE_BYTES)) {
       onPdfError(t('upload.fileTooLarge'));
       return;
     }
+
+    const existing = new Set(
+      courseMaterials.map((file) => `${file.name}:${file.size}:${file.lastModified}`),
+    );
```
yanpgwang: **Dedup key duplicated in two places.**

The `${file.name}:${file.size}:${file.lastModified}` composite key is built here and also in `addCourseMaterials` (app/page.tsx ~L290). Both build a Set from current `courseMaterials` and filter incoming files against it. If someone later changes the fingerprint (add mimeType, use a content hash, ignore lastModified for pasted files), one call site will be updated and the other forgotten.

Extract a single `dedupeCourseMaterials(existing, incoming)` helper (or a lower-level `fileFingerprint(file)`) and have both entry points call it. The toolbar can then focus on MIME/size/count validation and pass all `supportedFiles` down.

yanpgwang (commented):
Re-reviewed `905cafa fix(document): address bundle review findings` against the earlier high-effort review. All 10 prioritized findings plus the scene-content vision issue (#8) are addressed, with regression tests for the two subtle ordering/truncation ones. Verification below.

**Local verification (on PR HEAD):**
- `vitest` bundle + registry tests → 13 passed (incl. 2 new regression tests)
- `tsc --noEmit` → clean
- `eslint` on changed files → clean (the one `exhaustive-deps` warning in `app/page.tsx` is pre-existing, present in the parent commit)
- `prettier --check` → clean

### Blocking findings — all addressed

1. **storeImages misalignment** — `storeImages` now wraps the whole loop in one try/catch, rolls back partial writes via `Promise.allSettled(delete)`, and throws → `imageStorageIds[i]` is guaranteed 1:1 with `bundle.images[i]`. ✅
2. **Serial extraction** — replaced with `Promise.all(sortedSources.map(...))`; parallel, and N≤5 so no concurrency cap needed. ✅
3. **slice mid-token** — new `truncateTextAtBoundary` runs on the stable-id text before the final id-rewrite, uses `Array.from` (surrogate-safe) and backs off trailing `[\p{L}\p{N}_-]`. Regression test added. ✅
4. **`pageNumber || 1`** — now `?? 1`; no residual `|| 1`. ✅
5. **`localeCompare` tiebreak** — numeric `img_(\d+)` parse before `localeCompare`. Regression test added. ✅
6. **apiKey duplicated per source** — `providerConfig` removed from `SessionDocumentSource`; read once from `currentSession.pdfProviderConfig`, with a narrow legacy per-source fallback. ✅
7. **Promise.all orphan blobs** — `app/page.tsx` switched to a sequential loop tracking `storedDocumentKeys`, rolled back on throw. ✅
8. **Scene-content vision path** — `sortDocumentImagesForVision` now called in both `scene-content/route.ts` and `scene-generator.ts` (vision and text branches), so outline and scene-content select the same images. ✅
9. **Bundle limits enforcement** — see note below. ✅
10. **Dedup key duplication** — extracted to `dedupeCourseMaterialFiles`/`courseMaterialFingerprint` in new `lib/document/course-materials.ts`, used in both call sites. ✅

The five non-blocking observations (double `replaceImageIds` pass, two-phase id-rewrite collision hazard, legacy-field migration, `PdfImage` → `DocumentImage` rename, const-alias rename) were intentionally deferred — consistent with them being flagged "worth a follow-up."

### Two notes (neither blocking)

- **Finding 9 is enforced client-side, not in the route.** The guard (`validateDocumentSources`) lives in `generation-preview`, where the extraction loop is, rather than in `/api/extract-document`. That does close the described threat (a resumed/hand-edited session with N>5 no longer fires N requests), since the count/total-size limit is inherently bundle-level and the endpoint is per-file. A direct caller of `/api/extract-document` is still unbounded per-request — fine as long as the endpoint keeps its own per-file size limit.
- **Finding 1's fix changed failure semantics.** `storeImages` previously logged-and-skipped a bad image and continued; it now aborts the whole bundle on any single image failure. That's the correct trade for the alignment bug (a wrong storage key silently attached to every later image is worse than a clean failure), but it is a behavior change worth being aware of.

Net: the blocking set is cleared. LGTM pending the two notes above if you want to act on them.

yanpgwang (commented):
Re-reviewed `36d044f fix(document): harden extraction limits and storage errors`. This closes both of the non-blocking notes from the previous round.

**Local verification (on PR HEAD `abacea4`):**
- `vitest` document suite → 20 passed (incl. the new 413 route test)
- `tsc --noEmit` → clean
- `eslint` / `prettier --check` on changed files → clean

### Both notes resolved

- **Note on #9 (server-side limit):** `/api/extract-document` now rejects any file over 50MB with a `413` *before* extraction runs. The new per-file cap (`MAX_EXTRACT_DOCUMENT_FILE_SIZE_BYTES = 50MB`) exactly matches the client toolbar limit (`MAX_COURSE_MATERIAL_SIZE_MB = 50`), so it's clean defense-in-depth with no client/server mismatch. Test asserts the parser is not called on oversize input. ✅
- **Note on #1 (storage failure semantics):** `storeImages` now tracks `currentImageId` and throws a wrapped `Error(message, { cause: error })` naming the offending image, instead of the bare error — clearer diagnostics while keeping the atomic rollback. ✅

LGTM — no outstanding blocking or non-blocking items from my side.

yanpgwang:
### Local testing summary

Ran an extensive local pass on this PR's head (worktree), using the self-hosted MinerU extractor + `unpdf`, with the ModelVerse LLM gateway.

**Toolbar / upload UI — all pass**
- Multi-file upload with source ordering (numbered `1..N`, "merged in display order")
- Dedup by `(name, size, lastModified)` fingerprint
- 5-file cap (exact boundary; whole batch rejected on overflow)
- Remove + automatic re-numbering
- Per-file 50MB limit (rejected client-side)
- 150MB total-size limit (blocked at upload)
- Provider-scoped format filtering (docx/png rejected under `unpdf`; `plain-text` always available for txt/md)
- Switching provider drops only the now-incompatible files

**Bundle aggregation — verified by inspecting the outline request payload**
- 3 sources merged in order with `## Source Document N` headers (name / order / MIME / pages)
- Global image-ID normalization to continuous `img_1..N`; no `doc_x_img_x` leftovers, no cross-document collisions
- Every image carries `sourceDocumentName / order / originalId / storageId`
- Vision priority assigned round-robin across sources (a 3-image doc is not starved by a 38-image doc)

**End-to-end generation**
- Real classroom generation; a figure extracted from the PDF rendered correctly into a slide (vision path via the multimodal model)
- Cross-format bundle via MinerU: `txt + md + docx + png + pdf` (5 files = cap) all extracted concurrently and merged; the image-format file (png) was also parsed by MinerU

**Automated**
- Unit tests `bundle.test.ts` + `extract-document-route.test.ts`: 12/12 pass
- CI green (Lint/Typecheck/Unit + E2E)

**Not exercised locally (low risk)**
- Backward-compat with legacy single-PDF sessions (covered by the `legacySourceFromSession` fallback)
- Storage rollback on mid-bundle failure (hard to induce; logic reviewed)

@wyuc do you need to run any additional testing on your side? If not, I think this is good to **approve & merge**.

wyuc (commented):
Checked a real local mixed-document generation run on the tested head (`5e08efc`):

- one bundle containing `xlsx + png + pptx + 2 PDF`
- all five `/api/extract-document` requests started in parallel and returned `200`; the parsing stage completed with the slowest file (~17.8s)
- outline generation, 5 agent profiles, 6 scenes, and TTS all completed successfully

The document parsing/generation path itself looks good.

One UX copy follow-up before merge: the preview step still derives a single file type and renders copy such as “Analyzing PDF file”. That is misleading for a mixed-format bundle. Please make this step format-agnostic — `Analyzing documents` / `解析文档` — remove the unused `{{type}}` interpolation, and update all locales consistently.

The branch has now been updated with the latest `main` in `064cf27`; GitHub reports it mergeable and both CI checks are green, so the earlier conflict/block appears resolved.

wyuc (approved):
Re-reviewed head `720da83`. The requested format-agnostic preview copy is implemented consistently across all locales, with no parsing/generation behavior changes. The previously verified mixed-document generation path remains valid, and both current CI checks pass. LGTM.

PR #845, opened 2026-07-03
feat(importer): normalize slides at the pipeline output boundary (#787 follow-up)

wyuc:
Follow-up to #787 (now closed): wire the DSL contract's `normalize*` pass at the **importer's** output boundary — the counterpart of the generator wiring that landed with `normalize` itself in #832. With this, both producers (generator + importer) hand consumers contract-normalized slides.

A cross-review round on the initial wiring surfaced two contract-level issues, so the PR also touches `@openmaic/dsl` (0.3.0 → 0.4.0).

## What changed

**`@openmaic/importer`**

- **`normalizeImportedSlides(slides)`**: `parsedToSlides` (and therefore `importPptx`, the app's entry) now returns slides run through the DSL's `normalizeSlide`. Matching the importer's degrade-not-fail stance, it uses `onInvalid: 'drop'`: an element normalization cannot repair (a present-but-wrong-typed required field) is dropped with a `console.warn` rather than failing the whole import or reaching consumers that read it unguarded.
- Exported from the package root and documented in the README: consumers calling `transformParsedToSlides` directly (bundler-constrained integrations) should apply it themselves to get the same output contract.
- Tests: default-filling, empty-fill preservation, drop-and-warn, input purity, and a `parsedToSlides` end-to-end regression for the fill case below.

**`@openmaic/dsl` (0.3.0 → 0.4.0)**

- **`normalizeShape` preserves an explicit `fill: ''`** (new `strKeepEmpty`). The empty string is a *meaningful* fill — "no solid fill" (transparent, or gradient/pattern carried by sibling fields) — and the renderer maps it to `none` (`element.fill || 'none'`). Defaulting it would have painted every imported gradient / image-filled / unfilled shape with the canonical `#5b9bd5`. Found by the review round; regression-tested at three levels.
  - Behavior note: this also reaches the generator path (`fixElementDefaults` → `normalizeElement`) — a model-emitted `fill: ""` now renders as intended no-fill instead of being rescued to the default color. Absent (`undefined`) fills still get the default.
- **`normalizeSlide` gains an optional `{ onInvalid: 'throw' | 'drop', onDropped }` policy**, so producers normalizing wild-world input own one shared repair-or-drop pass instead of hand-rolling the `try / normalizeElement / catch` idiom per producer (the importer was about to become the second copy alongside the generator's). Default stays `'throw'`; `onDropped` observes each drop so the loss is never silent. `normalizeScene` / `normalizeStage` now call `normalizeSlide` via an explicit arrow so `map`'s index can't land in the options parameter.

## Review

Cross-vendor review ran twice (Claude multi-angle + Codex `exec review`):

- Round 1: consensus P1 — the `fill: ''` clobbering above (Codex caught it; Claude finders confirmed impact across gradient / image / unfilled shapes). Consensus P2 — the drop policy belonged in the DSL, not hand-rolled per producer. Both fixed at the contract level.
- Round 2 (full diff, fresh context): no correctness findings; two P3 doc gaps (README API table, generator behavior note) addressed in the follow-up commits.

## Notes

- No behavior change for well-formed decks: the transform already emits complete DSL `Slide` objects, so the pass is a no-op safety net there.
- The app consumes the importer via the build-time vendor copy (`sync-maic-importer.mjs`), which re-bundles from `dist` — no vendored artifact to update.

Closes the "where to wire validation — importer output" open question from #787; tracked under #720 (next-steps comment).


cosarah (changes requested):
I found two issues that should be fixed before this lands:

1. `@openmaic/importer` now has a new public export and a user-visible behavior change (`parsedToSlides()` / `importPptx()` now normalize at the output boundary), but `packages/@openmaic/importer/package.json` is still `0.0.2`. Our publish workflow is version-bump driven, and `pnpm publish` skips versions that are already on npm, so if `0.0.2` has already been published these importer changes will never ship. Please bump the importer version in the same PR.

2. Adding `NormalizeSlideOptions` as the second parameter to `normalizeSlide` is a TypeScript breaking change for existing `slides.map(normalizeSlide)` call sites. The repo already had to patch its own internal uses to explicit arrow wrappers in `packages/@openmaic/dsl/src/normalize.ts`, which is a strong signal that external consumers will hit the same problem. I reproduced it with a minimal `tsc` check: `slides.map(normalizeSlide)` now fails because `Array.prototype.map` passes the numeric index into the second parameter. If this break is intentional, it needs to be called out explicitly in the release notes / API docs; otherwise I would keep the old unary-call-site ergonomics via an overload or a separate helper.

wyuc:
Both points addressed in d6e02b0 — thanks, the map-index reproduction was exactly right.

**1. Importer version:** bumped `@openmaic/importer` 0.0.2 → 0.1.0 in this PR (new public export + output-boundary behavior change warrants the minor).

**2. `normalizeSlide` ergonomics:** agreed the internal arrow-wrapper patching was the tell, so I kept the old unary contract instead of documenting a break: `normalizeSlide(slide)` is unary again and `slides.map(normalizeSlide)` type-checks as before (now pinned by a test). The invalidity policy moved to a curried **`normalizeSlideWith(options)`**, which returns a map-safe unary function — and returns `normalizeSlide` itself when no drop policy is requested, so the two can't drift:

```ts
const normalize = normalizeSlideWith({ onInvalid: 'drop', onDropped: log });
const clean = slides.map(normalize);
```

`normalizeScene` / `normalizeStage` are back to plain `.map(normalizeSlide)`, and the importer builds its dropper once and maps it. Both packages' tests and builds are green (dsl 126, importer 32).

cosarah (approved):
LGTM

PR #850, opened 2026-07-04
Added ComfyUi to has Image Provider

PhillLittlewood:
## Summary

This adds ComfyUI as an Image Provider.

## Related Issues

NA

## Changes

Added ability for user to select ComfyUI as an image provider and select the desired workflow.
- Created `app/api/comfyui-workflows/route.ts` to handle backend API routing for ComfyUI workflows.
- Created `lib/media/adaptors/comfyui-image-adapter.ts` to implement the core image generation adaptor logic.
- Integrated the new ComfyUI provider options into the front-end settings component (`components/settings/image-settings.tsx`).

## Type of Change

### Steps to reproduce / test

1. Navigate to the application settings.
2. Go to the Image Settings section and locate the image provider dropdown.
3. Select "ComfyUI" as the active image provider.
4. Configure or select the desired ComfyUI workflow template.
5. Trigger an image generation task to verify successful integration and output retrieval.

- Verified that the new frontend configuration components render correctly without layout shifts.
- Confirmed the API route successfully handles payload handshakes for ComfyUI workflows.
- Verified standard execution flow using local environment testing.

### Evidence

- [x] Manually tested locally

## Checklist


wyuc (changes requested):
Thanks for adding ComfyUI support. I ran this through a cross-review pass (Claude report-only + Codex subagent, both against the same PR diff) and found a few issues that should be fixed before merge.

## Blocking

**[P2] The ComfyUI workflow id is used as a server-side file path without validating it against the workflow list.**

`generate/image` forwards the client-supplied `x-image-model` header as `config.model`, and `loadWorkflow()` turns that into `public/${config.model}`. That means the generation endpoint trusts the model header as a workflow filename. This should be constrained to the same set of workflow filenames returned by `/api/comfyui-workflows` (or at least to a safe basename in that directory), otherwise callers can request unintended JSON files rather than one of the listed workflows.

Relevant paths: `app/api/generate/image/route.ts`, `lib/media/adapters/comfyui-image-adapter.ts`.

**[P2] The documented server-side ComfyUI env config is not wired to runtime.**

`.env.example` adds `IMAGE_COMFYUI_BASE_URL` and `IMAGE_COMFYUI_WORKFLOW_PATH`, but `lib/server/provider-config.ts` has no `IMAGE_COMFYUI` entry in `IMAGE_ENV_MAP`, and the image server config only treats `lemonade` as a keyless image provider. As a result, operators cannot server-configure ComfyUI using the advertised env var. `IMAGE_COMFYUI_WORKFLOW_PATH` is also never read by the route or adapter, and it contradicts the current implementation/docs that load workflows from `public/`.

Please either wire these env vars through fully (including keyless server-managed ComfyUI and a clear workflow path/list story) or remove the dead env documentation.

**[P2] The default ComfyUI model id points to a non-existent workflow path.**

The provider registry defines `models: [{ id: "comfyui-image", ... }]`. When the user switches to the ComfyUI provider, the settings store can resolve `imageModelId` to `comfyui-image`; the adapter then treats that as a workflow filename and tries to load `public/comfyui-image`, which does not exist. Selecting a real workflow later works, but the default provider state fails immediately.

The provider should either default to an actual available workflow file, use an explicit workflow field instead of overloading `model`, or require/select a valid workflow before generation.

**[P2] Selecting a ComfyUI workflow in Settings can change the global image model without activating the ComfyUI provider.**

In the Settings dialog, the image provider list changes only the local panel selection (`selectedImageProviderId`), not the active `imageProviderId`. The new ComfyUI workflow list then calls only `setImageModelId(workflow.id)`. So a user can open Settings, view ComfyUI, click `comfyui-*.json`, and leave the active provider as Seedream/OpenAI/etc.; the next generation sends that workflow filename as the other provider's model id.

Please make workflow selection scoped to the ComfyUI provider, or activate ComfyUI atomically when selecting a workflow.

## Also worth fixing

- **[P3] Server-managed image providers now show editable API key/base URL fields even though the server ignores client overrides for managed providers.** This regresses the previous managed-provider UX and can persist inert values while displaying a request URL that is not actually used. Please restore the read-only/hidden credential behavior for server-configured providers.
- **[P3] The new ComfyUI settings UI has hardcoded English strings** (`Workflows`, `Refresh`, empty state/help text, workflow load error) while the rest of this panel uses i18n keys.
- **[P3] `loadWorkflow()` has unreachable duplicate code** after returning the fetched JSON.
- **[P3] The setup doc says the default max resolution is `1920x1920`, but the provider registry sets `1024x1024`.**

Review note: CI is green, but the issues above are runtime/configuration and UX-path problems that the current checks do not cover.

PhillLittlewood:
Thank you for the excellent and thorough report.  I have made all the changes you suggested and also removed comfy from env.example

wyuc:
## Code Review — ComfyUI Image Provider

Reviewed the full diff independently with two separate reviewers and merged the results by consensus. The security-sensitive surface is genuinely solid: workflow-id validation is well-layered (basename check → live allowlist → `path.resolve` containment as defense-in-depth), and SSRF is already handled upstream of this code by `validateUrlForSSRF` on both routes. The findings below are **functional correctness**, not security holes.

### 🔴 P2 — Default workflow file doesn't exist; feature fails whenever no workflow id is passed
`lib/media/adapters/comfyui-image-adapter.ts:49-50` + `lib/media/image-providers.ts:128-142`

`DEFAULT_WORKFLOW_FILENAME = 'comfyui-workflow.json'`, but the only file shipped is `public/comfyui-ernie.json`. When `config.model` is empty, `loadWorkflow` throws `workflow file not found`. Confirmed failure paths:
- **Classroom / autonomous media** (`lib/server/classroom-media-generation.ts:103`): `model = providerConfig?.models?.[0]?.id` → `undefined` for ComfyUI (`models: []`). **Every** ComfyUI image generation through this path fails, with no way to select `comfyui-ernie.json`.
- **User selects ComfyUI but hasn't clicked a workflow yet**: `imageModelId=''` → empty `x-image-model` header → same failure.

The comment at `image-providers.ts:137-140` actively claims this fallback works. Suggested fix: default to the first entry from `listComfyuiWorkflowFilenames()` when `config.model` is empty (and correct the comment), or ship a `comfyui-workflow.json`.

### 🟠 P2 — Route `maxDuration = 60` vs. 5-minute adapter polling
`app/api/generate/image/route.ts:37` hardcodes `maxDuration = 60`, but the adapter polls up to `GENERATION_TIMEOUT_MS = 300_000` (5 min), and the new setup guide notes workflows can take 3–5 min. On platforms that enforce `maxDuration`, the request is killed at ~60s while the adapter keeps polling, so generation fails after about a minute. (On a self-hosted Node server `maxDuration` is ignored, so impact depends on deploy target.) Separately, `app/api/verify-image-provider/route.ts` sets no `maxDuration` at all.

### 🟠 P2 — Portrait aspect ratios exceed the declared max resolution
`comfyui-image-adapter.ts:222` → `aspectRatioToDimensions('9:16', 1920)` returns **1920×3413** (the helper always scales height *up* for portrait), which exceeds the provider's declared `maxResolution: 1920×1920`. This oversized value is patched straight into the latent-size node, so portrait generations may OOM or fail. Suggested fix: use a bounding-box clamp, or reuse the route's already-clamped dimensions.

### 🟡 P3
- **Runtime execution errors → 5-min hang + misleading "timed out"** — `comfyui-image-adapter.ts` poll loop only checks `entry.status.completed`. A workflow that errors *during execution* records `completed: false, status_str: "error"`, so the loop polls the full 5 minutes then throws a generic timeout, masking the real error. Suggested fix: check `status_str === 'error'` and throw immediately.
- **No `AbortSignal`/timeout on any `fetch`** — `queuePrompt` / `fetchImageAsBase64` / `pollHistory` / `testComfyuiImageConnectivity`. The 5-min bound is on the polling *loop*, not the awaited calls. `testComfyuiImageConnectivity` (via `verify-image-provider`, no `maxDuration`) can hang if a host accepts the connection but stalls the response body. Suggested fix: `AbortSignal.timeout(...)`.
- **Server-side base64 via per-byte loop** — `fetchImageAsBase64` builds a multi-MB string with `for (const byte of bytes) binary += String.fromCharCode(byte)` then `btoa`. This runs server-side; suggested fix: `Buffer.from(buffer).toString('base64')`.
- **Local ComfyUI blocked in production** — provider defaults to `localhost:8188`, but a client-supplied `x-base-url` to localhost/private ranges is rejected (403) by `validateUrlForSSRF` in production. This matches existing local-provider behavior and may be intentional, but the localhost default is misleading — worth documenting the intended deployment topology in `comfyui-setup-instructions.md`.

### ⚪ Minor
- The browser branch of `loadWorkflow` (`comfyui-image-adapter.ts:166-172`) validates with `isComfyuiWorkflowFilename` but omits the `listComfyuiWorkflowFilenames()` allowlist check the server branch enforces. Currently unreachable (generation runs server-side), but a latent asymmetry.
- `providerComfyUIImage` is added to all locale files but appears unused — provider display names come from `IMAGE_PROVIDERS[p.id].name`. (All 8 `comfyui*` keys actually referenced by the UI are present in all 8 locales — no missing-key issue.)
- `patchWorkflow` assigns `promptNode.inputs['value']` directly; a malformed workflow node lacking an `inputs` object throws a `TypeError`.

---

The nonexistent-default-workflow bug is the must-fix — it breaks the classroom media path entirely. The other two P2s were each verified as well. Nice work on the traversal/allowlist hardening.

PhillLittlewood:
I have made all the requested changes

wyuc (approved):
Re-reviewed at 7796e03 after updating onto current main. CI is green (lint/typecheck/unit and E2E). Targeted ComfyUI smoke also passed: connectivity, explicit workflow selection, default workflow fallback, 9:16 clamping to 1080×1920, real queue execution, PNG retrieval, and 200 response. Approving.

Issue #851, opened 2026-07-05
[Task] @openmaic/renderer v2: the editing surface (#720 Phase 3)


wyuc:
Breaks out **Phase 3 — renderer v2** from #720 into reviewable parts, the way #787 did for Phase 2. This is the largest remaining item and the epic's critical path; the goal of this issue is to fix the part boundaries and the package API design before code moves.

## Current state

- `@openmaic/renderer` v1 is **read-only by design**: `SlideCanvas` + per-type element renderers under `src/elements/`, consumed by the app's three thumbnail surfaces since #707.
- The app still owns the entire editing surface, duplicated against the package's read-only renderers:
  - `components/slide-renderer/Editor/` (~39 files): the editable canvas — `EditableElement`, the `Operate/` handles (drag / resize / rotate), `MouseSelection`, `AlignmentLine`, `GridLines`, `Ruler`, `ElementCreateSelection`, `ShapeCreateCanvas`, plus playback overlays (`SpotlightOverlay`, `LaserOverlay`, `HighlightOverlay`, `ScreenCanvas`).
  - `components/slide-renderer/components/element/` (~41 files): the full-size element renderers, including `ProsemirrorEditor` for inline text editing.
- Editing state (selection, active tool, undo) lives in app stores; the editable canvas reaches into them directly.

## Part A — editing core: selection + operate surface

Promote the interaction machinery into `@openmaic/renderer`:

- selection model (single / multi / marquee via `MouseSelection`), keyboard modifiers
- `Operate` handles: drag, resize (8-point), rotate; `AlignmentLine` snapping, `GridLines`, `Ruler`
- element creation flows (`ElementCreateSelection`, `ShapeCreateCanvas`)
- a controlled API: the package renders and emits **edit intents** (`onElementsChange`, `onSelectionChange`, …); the host owns the document and undo. No store coupling inside the package.

## Part B — editable element renderers + inline text

- absorb the full-size renderers (`components/slide-renderer/components/element/`) as the editable counterparts of the v1 read-only set, sharing geometry/style code with v1 instead of duplicating it
- ProseMirror inline editing for text / shape-text (the `ProsemirrorEditor` stack becomes package-internal; ProseMirror enters the package dependency set — renderer already has runtime deps, unlike `@openmaic/dsl`)
- media stays resolved by the host (same `gen_*` reactive-resolution pattern the thumbnail wrappers use today)

## Part C — app slimdown

[...]

PR #853, opened 2026-07-05
chore(renderer): machine-enforce the @openmaic/renderer import boundary

wyuc:
Adds a package-scoped ESLint boundary for `@openmaic/renderer`: a `files`-scoped `no-restricted-imports` rule that bans host-app imports via the `@/…` path alias inside `packages/@openmaic/renderer/**`, enforced by the existing `pnpm lint` CI step.

## Why

`@openmaic/renderer` depends only on `@openmaic/dsl` and its declared peers. Host concerns (document/undo ownership, media resolution, i18n, hotkeys) are injected via props/callbacks, not imported. This is the machine-enforced counterpart of the renderer v2 editing-surface boundary (#720 Phase 3, scoped in #851): with the rule in place, a deadline can't punch a "temporary" store dependency through the package API later.

This is gate item 2 of #851 ("package-side lint/CI rule rejects app imports (`@/…`)").

## Scope

The package imports zero `@/…` today, so the rule is purely additive — no source changes required.

## Verification

- In-package `@/…` imports (deep `@/lib/store` and shallow `@/foo`) now fail lint.
- App-land files importing `@/…` are unaffected — the rule is scoped to `packages/@openmaic/renderer/**`.
- Existing package source lints clean under the rule.

Refs #851, #720.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -49,11 +49,18 @@ const eslintConfig = defineConfig([
     },
   },
   // Package boundary (machine-enforced): @openmaic/renderer is a standalone,
-  // app-agnostic package that depends only on @openmaic/dsl plus its declared
-  // peers. It must never reach back into the host app through the `@/…` path
-  // alias, so a deadline can't punch a "temporary" store/undo/media dependency
-  // through the package API. Host concerns (document + undo ownership, media
-  // resolution, i18n, hotkeys) are injected via props/callbacks instead.
+  // app-agnostic package. It must never reach back into the host app through
+  // the `@/…` path alias, so a deadline can't punch a "temporary"
+  // store/undo/media dependency through the package API. Host concerns
+  // (document + undo ownership, media resolution, i18n, hotkeys) are injected
+  // via props/callbacks instead.
+  //
+  // Coverage: `no-restricted-imports` handles static `import` / `export … from`
+  // (including `import type`); `no-restricted-syntax` closes the dynamic
+  // `import('@/…')` and `require('@/…')` bypasses, which the base rule does not
+  // catch (it registers no ImportExpression/CallExpression handler). Relative
+  // parent escapes (`../../app`) are outside this alias rule's scope; the
+  // package's own standalone `tsc` (rootDir: src) rejects those at build time.
   {
     files: ['packages/@openmaic/renderer/**/*.{ts,tsx,js,jsx,mjs,cjs}'],
     rules: {
@@ -67,6 +74,19 @@ const eslintConfig = defineConfig([
                 '@openmaic/renderer must not import from the host app (@/…). Depend only on @openmaic/dsl and declared peers; inject host concerns (stores, undo, media resolution, i18n, hotkeys) via props/callbacks.',
             },
           ],
+        },
+      ],
+      'no-restricted-syntax': [
+        'error',
+        {
+          selector: 'ImportExpression[source.value=/^@\\//]',
+          message:
+            '@openmaic/renderer must not dynamically import from the host app (@/…). Inject host concerns via props/callbacks.',
+        },
+        {
+          selector: "CallExpression[callee.name='require'][arguments.0.value=/^@\\//]",
+          message:
+            '@openmaic/renderer must not require() from the host app (@/…). Inject host concerns via props/callbacks.',
         },
       ],
     },
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -98,7 +98,8 @@ const eslintConfig = defineConfig([
             '@openmaic/renderer must not require() from the host app (@/…). Inject host concerns via props/callbacks.',
         },
         {
-          selector: "CallExpression[callee.name='require'][arguments.0.quasis.0.value.cooked=/^@\\//]",
+          selector:
+            "CallExpression[callee.name='require'][arguments.0.quasis.0.value.cooked=/^@\\//]",
           message:
             '@openmaic/renderer must not require() from the host app (@/…). Inject host concerns via props/callbacks.',
         },
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -48,6 +48,29 @@ const eslintConfig = defineConfig([
       ],
     },
   },
+  // Package boundary (machine-enforced): @openmaic/renderer is a standalone,
+  // app-agnostic package that depends only on @openmaic/dsl plus its declared
+  // peers. It must never reach back into the host app through the `@/…` path
+  // alias, so a deadline can't punch a "temporary" store/undo/media dependency
+  // through the package API. Host concerns (document + undo ownership, media
+  // resolution, i18n, hotkeys) are injected via props/callbacks instead.
+  {
+    files: ['packages/@openmaic/renderer/**/*.{ts,tsx,js,jsx,mjs,cjs}'],
+    rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: [
+            {
+              group: ['@/*', '@/**'],
+              message:
+                '@openmaic/renderer must not import from the host app (@/…). Depend only on @openmaic/dsl and declared peers; inject host concerns (stores, undo, media resolution, i18n, hotkeys) via props/callbacks.',
+            },
+          ],
+        },
+      ],
+    },
+  },
 ]);
 
 export default eslintConfig;
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -61,9 +61,11 @@ const eslintConfig = defineConfig([
   //
   // - `no-restricted-imports` covers static `import` / `export … from`
   //   (including `import type`).
-  // - `no-restricted-syntax` covers the dynamic call forms the base rule can't
-  //   see — `import()`, `require()`, `require.resolve()` — on single string- or
-  //   template-literal specifiers.
+  // - `no-restricted-syntax` covers the resolve/import call forms the base rule
+  //   can't see — `import()`, `require()`, `require.resolve()`,
+  //   `import.meta.resolve()` — on single string- or template-literal specifiers.
+  //   Together with the static forms above, this is the complete set of ways to
+  //   reference a module by a single literal specifier.
   //
   // Deliberately out of scope (not decidable by lint, and evasion-only): a
   // specifier assembled from non-literal parts — `import('@/lib/' + x)`,
@@ -119,6 +121,18 @@ const eslintConfig = defineConfig([
             "CallExpression[callee.object.name='require'][callee.property.name='resolve'][arguments.0.quasis.0.value.cooked=/^@\\//]",
           message:
             '@openmaic/renderer must not require.resolve() a host-app path (@/…). Inject host concerns via props/callbacks.',
+        },
+        {
+          selector:
+            "CallExpression[callee.object.type='MetaProperty'][callee.property.name='resolve'][arguments.0.value=/^@\\//]",
+          message:
+            '@openmaic/renderer must not import.meta.resolve() a host-app path (@/…). Inject host concerns via props/callbacks.',
+        },
+        {
+          selector:
+            "CallExpression[callee.object.type='MetaProperty'][callee.property.name='resolve'][arguments.0.quasis.0.value.cooked=/^@\\//]",
+          message:
+            '@openmaic/renderer must not import.meta.resolve() a host-app path (@/…). Inject host concerns via props/callbacks.',
         },
       ],
     },
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -55,12 +55,21 @@ const eslintConfig = defineConfig([
   // (document + undo ownership, media resolution, i18n, hotkeys) are injected
   // via props/callbacks instead.
   //
-  // Coverage: `no-restricted-imports` handles static `import` / `export … from`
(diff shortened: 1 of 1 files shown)


wyuc:
## Cross-review summary

This change was independently cross-reviewed (two reviewers, same diff), then looped to convergence.

**Round 1**
- Reviewer A: clean.
- Reviewer B found a real gap: the base `no-restricted-imports` rule only registers `ImportDeclaration` / `ExportNamedDeclaration` / `ExportAllDeclaration` handlers, so dynamic `import('@/…')` and `require('@/…')` slipped past the boundary and resolved at runtime — exactly the "temporary punch-through under deadline" the rule is meant to prevent.

**Fix** (2nd commit, `8cb0896e`)
- Closed dependency-free with `no-restricted-syntax` selectors for `ImportExpression[source.value=/^@\//]` and `require()` `CallExpression[...]`, and tightened the comment to describe what lint actually enforces.

**Round 2**
- Both reviewers converged with zero actionable findings.

**Verification**
- Static `import '@/…'`, `import type … from '@/…'`, dynamic `import('@/…')`, and `require('@/…')` all error.
- Legitimate `import('@openmaic/dsl')` / `import('echarts')` and non-`@/` requires are unaffected (no over-match).
- Existing package source lints clean; all CI checks green (`Lint, Typecheck & Unit Tests`, `E2E Tests`, `Analyze`).

**Out of scope (documented in the config comment)**
- Relative parent escapes (`../../app`) are not covered by this alias rule; the package's own standalone `tsc` (`rootDir: src`) rejects those at build time. A future `eslint-plugin-import` `no-restricted-paths` zone could machine-enforce that vector too if desired.

Ready for review.

cosarah:
I found two remaining bypasses in the new boundary rule:

1. `eslint.config.mjs:82` — `ImportExpression[source.value=/^@\\//]` only matches string-literal dynamic imports. `await import(`@/lib/foo`)` is still allowed, so switching quotes to a no-substitution template literal bypasses the rule.

2. `eslint.config.mjs:87` — `CallExpression[callee.name=require][arguments.0.value=/^@\\//]` only catches direct `require("@/...")`. `require.resolve("@/lib/foo")` still passes, so the package can still reach app-owned paths through the alias.

If the goal is to machine-enforce the boundary, these should be covered as well.

wyuc:
Both correct — confirmed and fixed in `e6760a54`.

- Template-literal dynamic import: the selector matched `source.value` (Literal), so a `TemplateLiteral` source slipped past. Added `ImportExpression[source.quasis.0.value.cooked=/^@\//]` (also covers `` require(`@/…`) ``).
- `require.resolve(...)`: the selector matched `callee.name==='require'` (Identifier), so the `MemberExpression` callee was missed. Added `CallExpression[callee.object.name='require'][callee.property.name='resolve'][arguments.0.value=/^@\//]` (+ its template-literal form).

The rule now covers all statically-analyzable `@/` forms — string-literal and template-literal — across `import`, `import()`, `require()` and `require.resolve()`. Verification matrix (all 6 error; legitimate imports and bare strings untouched):

```
import('@/lib/a')          → error      import('@openmaic/dsl')      → ok
import(`@/lib/b`)          → error      import(`echarts`)           → ok
require('@/lib/c')         → error      require('fs')               → ok
require(`@/lib/d`)         → error      require.resolve('path')     → ok
require.resolve('@/lib/e') → error      const s = '@/not-an-import' → ok
require.resolve(`@/lib/f`) → error
```

One honest boundary worth stating: genuinely runtime-computed specifiers — `import(someVar)`, `import('@' + '/x')`, an aliased `const r = require; r('@/x')` — are undecidable by any lint rule and stay out of scope. If we want enforcement that doesn't depend on the literal syntax at all, the resolved-path approach (`eslint-plugin-import`'s `no-restricted-paths`) is the follow-up option; it would also close the relative-parent escape (`../../app`) that this alias rule leaves to the package's own `tsc` `rootDir`. Happy to add that as a separate change if you'd prefer the stronger guarantee here.

cosarah:
I found one remaining bypass in the boundary rule:

`eslint.config.mjs:79-116` still does not fully machine-enforce the `@/…` boundary. The selectors cover string literals and no-substitution template literals, but they do not match string concatenation expressions. I verified that both `await import("@/lib/" + "foo")` and `require.resolve("@/lib/" + "foo")` still pass lint inside `packages/@openmaic/renderer`.

If the goal is to make the boundary machine-enforced, this case still needs to be blocked as well.

wyuc:
Fixed in `533e7386` — you're right, the shape-specific selectors couldn't keep up with concatenation.

Rather than add yet another call-shape selector, I switched the approach: match the `@/…` **string prefix wherever it is authored** (`Literal[value=/^@\//]` + `TemplateElement[value.cooked=/^@\//]`), instead of matching `import` / `require` / `require.resolve` shapes. That flags the `'@/lib/'` operand in `import('@/lib/' + 'foo')` directly, and subsumes dynamic `import()` / `require()` / `require.resolve()` in any quote style in two selectors instead of six. The package has zero legitimate `@/…` strings, so there are no false positives.

Verified now blocked: static / `import type` / `export … from`, `import('@/x')`, `` import(`@/x`) ``, `require('@/x')`, `require.resolve('@/x')`, and both `import('@/lib/' + 'foo')` and `require.resolve('@/lib/' + 'bar')`. Clean: `import('@openmaic/dsl')`, `require('fs')`, `require.resolve('path')`, and plain strings.

One honest limit worth stating explicitly: a specifier assembled entirely from non-`@/` pieces — `import('@' + '/x')`, `import(someVar)`, an aliased `const r = require; r('@/x')` — is undecidable by any lint rule (resolver-based rules such as `eslint-plugin-import`'s only see statically-resolvable string sources too) and is reachable only by deliberate evasion, which `// eslint-disable` allows regardless. So lint is the accidental-mistake guard; the airtight backstop is that the package builds and publishes standalone with only `@openmaic/dsl` + declared peers external — any *resolvable* `@/` dependency breaks that isolated build. If you'd prefer that guarantee made explicit in CI, I'm happy to add `pnpm --filter @openmaic/renderer build` (or a package typecheck) as a follow-up.

cosarah:
I found one remaining issue in the latest version of the boundary rule:

`eslint.config.mjs:96-103` is no longer scoped to import contexts. `Literal[value=/^@\\//]` and `TemplateElement[value.cooked=/^@\\//]` now flag any authored `@/…` string in `@openmaic/renderer`, even when no app import occurs, and they also double-report static `import` / `export ... from` cases that are already covered by `no-restricted-imports`.

I verified both of these behaviors:
- `import x from "@/lib/foo"` now produces two boundary errors.
- `const example = "@/lib/foo"` also fails, even though it does not cross the package boundary.

If the goal is specifically to reject app imports from `@/…`, this looks broader than intended and will create avoidable false positives.

wyuc:
You're right — that was a regression I introduced when I broadened the rule to catch concatenation. Fixed in `5a8fcf65`.

The two findings you raised pull in opposite directions (catch concatenation ⇒ broaden ⇒ false-positives), so I've drawn an explicit line rather than keep widening:

- `no-restricted-syntax` is back to **import contexts only** — dynamic `import()` / `require()` / `require.resolve()` on single string- or template-literal specifiers. Static `import` / `export … from` stays with `no-restricted-imports`.
- Both behaviors you flagged are gone: a bare `const x = '@/lib/foo'` no longer errors, and a static import is reported once, not twice. Verified with probe files (bare string → 0 boundary errors; static import → exactly 1).

- String concatenation (`import('@/lib/' + x)`), computed/variable specifiers, and relative parent escapes are now **explicitly documented as out of scope** for this rule (in the config comment). They're not decidable by lint in general and are evasion-only; the real guarantee for them is that the package builds/publishes in isolation with only `@openmaic/dsl` + declared peers external. If you'd like that made a CI gate, I'm glad to add a package build step as a follow-up.

Does treating concatenation/computed as build-backstop territory (rather than lint) match what you'd want here? Happy to reconsider if you'd rather keep it in lint despite the precision cost.

cosarah:
I found one remaining gap in the latest version of the boundary rule:

`eslint.config.mjs:88-123` still misses the ESM-native resolution path `import.meta.resolve("@/…")`. The rule now covers `import()`, `require()`, and `require.resolve()`, but not `import.meta.resolve()`.

I verified that both of these still pass lint inside `packages/@openmaic/renderer`:
- `const x = import.meta.resolve("@/lib/foo")`
- `const x = import.meta.resolve(`@/lib/foo`)`

This seems in scope for the rule, because `packages/@openmaic/renderer/package.json:5` declares the package as ESM-only (`"type": "module"`), so `import.meta.resolve()` is the direct ESM counterpart to the `require.resolve()` case that is already blocked.

wyuc:
Right again — `import.meta.resolve()` was a real in-scope gap (ESM counterpart to `require.resolve()`). Fixed in `fa4374e9` (string- and template-literal forms).

Rather than keep patching one form at a time, here's the **complete** enumeration of every way to reference a module by a single-literal specifier, with coverage — so we can see whether anything else is open:

| # | Form | Blocked by | Status |
|---|------|-----------|--------|
| 1 | `import x from '@/…'` (static) | `no-restricted-imports` | ✅ |
| 2 | `import '@/…'` (side-effect) | `no-restricted-imports` | ✅ |
| 3 | `export … from '@/…'` | `no-restricted-imports` | ✅ |
| 4 | `import x = require('@/…')` (TS import-equals) | `no-restricted-imports` | ✅ (verified) |
| 5 | `export import x = require('@/…')` | `no-restricted-imports` | ✅ (verified) |
| 6 | `import('@/…')` (dynamic) | `no-restricted-syntax` | ✅ |
| 7 | `require('@/…')` | `no-restricted-syntax` | ✅ |
| 8 | `require.resolve('@/…')` | `no-restricted-syntax` | ✅ |
| 9 | `import.meta.resolve('@/…')` | `no-restricted-syntax` | ✅ (this commit) |

Each string-literal form also has its no-substitution template-literal variant covered. Verified empirically (probe files + `npx eslint`): every form above errors for a single `@/` specifier; a bare `const x = '@/…'` string, and any of these forms with a non-`@/` specifier (e.g. `import.meta.resolve('@openmaic/dsl')`), stay clean, and no form double-reports.

Considered and deliberately excluded:
- `module.require('@/…')` — `module` is not defined in an ESM package (`"type": "module"`), so this is a `ReferenceError`, not a path to app code.
- Specifiers assembled from non-literal parts — `import('@/lib/' + x)`, `import(dynamicVar)`, an aliased `const r = require; r('@/x')` — and relative `../` escapes. Undecidable by lint and evasion-only; covered by building/publishing the package in isolation (only `@openmaic/dsl` + declared peers external).

If you spot a module-reference form missing from that table, I'll add it — but I believe this is now the complete statically-decidable surface for an ESM package.

cosarah:
I found one remaining gap in the latest version of the boundary rule:

`eslint.config.mjs:113-133` still only matches dot-property `resolve` calls. The new selectors block `require.resolve("@/…")` and `import.meta.resolve("@/…")`, but computed-property variants still pass lint.

I verified that all of these still pass inside `packages/@openmaic/renderer`:
- `require["resolve"]("@/lib/foo")`
- `require["resolve"](`@/lib/foo`)`
- `import.meta["resolve"]("@/lib/foo")`
- `import.meta["resolve"](`@/lib/foo`)`

These are still direct single-literal module-resolution paths, so if the goal is to machine-enforce the boundary for literal specifiers, this case still needs to be covered as well.

wyuc:
Correct again — `require["resolve"]` / `import.meta["resolve"]` (computed property) slipped past the dot-property selectors. Fixed in `c42cbe71`, but I've changed the **approach** rather than adding a 7th selector, because I don't think per-shape selectors converge here.

Six rounds in, each covered shape has invited the next variant (dynamic → template → require.resolve → import.meta.resolve → computed property), and the shape space is open-ended — after computed property comes optional chaining (`require?.resolve?.(…)`), a parenthesized callee, `Reflect.apply`, and so on. Matching call *shapes* is a losing game.

So `c42cbe71` matches the **`@/` string prefix wherever it appears** (`Literal` / `TemplateElement`), as the sole rule. That closes the entire single-literal class at once — static / `import type` / `export … from`, `import()`, `require()`, `require.resolve()`, `import.meta.resolve()`, **their computed-property and template variants**, and concatenation operands (`'@/lib/' + x`) — all in one selector, one report per violation (no more double-reporting). Verified against all of the above.

**The tradeoff — and it's your call, because it reverses your earlier point:** this also flags a bare `const x = '@/foo'` that isn't an import — the "false positive" you flagged when I last went broad. My argument for accepting it: `@/` is *exclusively* the host-app path alias, the package authors zero `@/…` strings of any kind, so treating "no `@/` strings at all" as the policy is precise *for this package* — a bare `@/` string is either an import or a mistake, never legitimate.

If you'd rather keep lint strictly precise (only flag actual imports, never a bare string), the coherent alternative is: revert to the narrow per-shape rule and get completeness from a **build-in-isolation CI check** instead — building the package with only `@openmaic/dsl` + declared peers external fails on any *resolvable* `@/` reference regardless of syntax. That puts precision in lint and completeness in the build, where each fits.

Which do you prefer — the broad rule as pushed, or narrow-lint + build-backstop? Either ends the per-variant cycle; I just want your call on where the precision/completeness line sits rather than guessing it again.

cosarah (approved):
LGTM

PR #855, opened 2026-07-05
feat(renderer): scaffold @openmaic/renderer/editing subpath (v2 editing surface, Stage 0)

wyuc:
Stage 0 of the renderer v2 editing surface (#851, #720 Phase 3): the packaging skeleton for the `@openmaic/renderer/editing` subpath. **Draft — companion to the editing-surface RFC on #851; not for merge until the RFC is accepted.**

## What this is

Packaging decision A from the RFC: editing ships behind a dedicated `./editing` entry, so the read-only entry (`@openmaic/renderer`) never pulls the editing bundle.

## Contents (scaffold only — no interaction logic yet)

- `src/editing/types.ts` — the **L1** edit-intent contract: `EditIntent` (the bounded canvas gesture vocabulary), `Selection`, `EditableSlideCanvasProps`. The agent tool surface (L2) and the canonical change representation (L0, in `@openmaic/dsl`) are deliberately out of scope here — see the layering note on #851.
- `src/editing/EditableSlideCanvas.tsx` — a shell that renders through the v1 read-only `SlideCanvas` and supports click-to-select only. Operate handles (drag/resize/rotate), alignment snapping, ProseMirror inline editing, and `onElementsChange` intent emission arrive in Part A / Part B.
- `package.json` + `rollup.config.js` — wire the `./editing` entry (mirrors the existing `./snapshot` / `./elements` subpaths).

## Intentionally deferred

- Text-command handle, `onTextInput`, `resolveMedia`, and ProseMirror optional-peer deps — they depend on RFC open questions and land with Part B.
- Intent emission + undo wiring — Part A.

## Verification

- Editing subtree typechecks clean (0 errors, with `@openmaic/dsl` resolved).
- `pnpm check` (Prettier) and `pnpm lint` pass (0 errors).
- Read-only entry unchanged; the new entry is purely additive.

Part of the extraction sequence scoped in #851; follows #853 (the package import boundary).

### DIFF
--- a/packages/@openmaic/renderer/rollup.config.js
+++ b/packages/@openmaic/renderer/rollup.config.js
@@ -40,6 +40,7 @@ const entries = {
   'elements/index': 'src/elements/index.ts',
   'types/index': 'src/types/index.ts',
   'snapshot/index': 'src/snapshot/index.ts',
+  'editing/index': 'src/editing/index.ts',
 };
 
 // ESM-only: @openmaic/dsl (kept external) is ESM-only, so a CJS renderer bundle
--- a/packages/@openmaic/renderer/rollup.config.js
+++ b/packages/@openmaic/renderer/rollup.config.js
@@ -1,6 +1,7 @@
 import { nodeResolve } from '@rollup/plugin-node-resolve';
 import commonjs from '@rollup/plugin-commonjs';
 import typescript from '@rollup/plugin-typescript';
+import preserveDirectives from 'rollup-plugin-preserve-directives';
 
 const external = [
   /^@openmaic\/dsl($|\/)/,
@@ -33,6 +34,11 @@ const plugins = [
     declarationMap: false,
     rootDir: 'src',
   }),
+  // Preserve module-level directives (e.g. 'use client') into the built output.
+  // Rollup strips them by default (see the MODULE_LEVEL_DIRECTIVE warning silenced
+  // in onwarn above), which drops the client-boundary marker from dist/ — breaking
+  // Next App Router server-component consumers of the editing entry.
+  preserveDirectives(),
 ];
 
 const entries = {
@@ -55,6 +61,10 @@ const buildBundle = () => ({
     format: 'es',
     entryFileNames: '[name].js',
     chunkFileNames: 'chunks/[name]-[hash].js',
+    // Emit one file per source module so per-module directives ('use client')
+    // survive into dist (rollup-plugin-preserve-directives requires this); a
+    // bundled build would mix client and non-client modules and drop the marker.
+    preserveModules: true,
     preserveModulesRoot: 'src',
     sourcemap: true,
   },
--- a/packages/@openmaic/renderer/src/editing/index.ts
+++ b/packages/@openmaic/renderer/src/editing/index.ts
@@ -1,3 +1,5 @@
+'use client';
+
 export { EditableSlideCanvas } from './EditableSlideCanvas';
 export { EMPTY_SELECTION } from './types';
 export type {
--- a/packages/@openmaic/renderer/src/editing/index.ts
+++ b/packages/@openmaic/renderer/src/editing/index.ts
@@ -1,0 +1,10 @@
+export { EditableSlideCanvas } from './EditableSlideCanvas';
+export { EMPTY_SELECTION } from './types';
+export type {
+  EditableSlideCanvasProps,
+  EditIntent,
+  Selection,
+  SnappingOptions,
+  ReorderCommand,
+  AlignCommand,
+} from './types';
--- a/packages/@openmaic/renderer/package.json
+++ b/packages/@openmaic/renderer/package.json
@@ -98,6 +98,7 @@
     "@types/react-dom": "^19",
     "@types/tinycolor2": "^1.4.6",
     "rollup": "^4.35.0",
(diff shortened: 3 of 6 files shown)


packages/@openmaic/renderer/src/editing/index.ts
```diff
@@ -0,0 +1,10 @@
+export { EditableSlideCanvas } from './EditableSlideCanvas';
```
cosarah, line 3: This still is not a reliable client entry for Next App Router. The new subpath is imported through `@openmaic/renderer/editing` -> `src/editing/index.ts`, but `'use client'` only exists in `EditableSlideCanvas.tsx`. On top of that, `packages/@openmaic/renderer/rollup.config.js` currently ignores `MODULE_LEVEL_DIRECTIVE`, and the built `dist/editing/index.js` drops the directive entirely. So a Server Component importing `@openmaic/renderer/editing` can still be treated as importing a server module, and function props like `onSelectionChange` may still fail serialization. Please move `'use client'` to the public editing entry and ensure the published bundle preserves it.

wyuc:
You're right, and I verified it rather than hand-waving this time. Root cause and evidence:

- Rollup drops module-level directives. A minimal `preserveModules` build of a `'use client'` module emits `(!) Module level directives cause errors when bundled, "use client" … was ignored`, and the output starts with the code, no directive.
- `rollup.config.js`'s `onwarn` silences exactly that `MODULE_LEVEL_DIRECTIVE` warning, so the drop happens quietly.
- Ground truth: the **published `@openmaic/renderer@0.0.2` dist** has no `'use client'` in its JS at all (the only `use client` hits are inside `.js.map` source content; e.g. `dist/chunks/SlideCanvas-*.js` starts with `import { jsxs … }`).

So this is pre-existing and package-wide — v1's published bundle already ships without the directive — not something the editing entry introduced. It happens not to bite v1 today because those surfaces are consumed from client components, but it's a real gap for an explicitly interactive entry.

I pushed the part that's correct at the source (`2e35ddc2` adds `'use client'` to the public `./editing` entry), but on its own it's dropped by the build just like the component's directive. The effective fix is **preserving module-level directives in the rollup build** — either `rollup-plugin-preserve-directives`, or a small inline `renderChunk` plugin that re-emits the directive for chunks whose source had it — which also fixes v1's published output.

Two questions before I land that:

1. Scope: it changes the whole package's published bundles (v1 included) and adds a build dependency, so it feels larger than this Stage 0 scaffold. Prefer it here, or as a dedicated build-fix PR that this scaffold then depends on?
2. I'd rather land it verified against a real package build than push it blind. (My local checkout is missing the workspace build deps, so I couldn't run the full `pnpm build` here — I confirmed the drop with a standalone rollup build and the published tarball instead.)

Either way, thanks for pushing on this — I initially waved it off as matching the v1 barrel convention, which was wrong: v1's barrels lack the directive *and* the build strips it everywhere.

packages/@openmaic/renderer/rollup.config.js
```diff
@@ -40,6 +40,7 @@ const entries = {
   'elements/index': 'src/elements/index.ts',
   'types/index': 'src/types/index.ts',
   'snapshot/index': 'src/snapshot/index.ts',
+  'editing/index': 'src/editing/index.ts',
```
cosarah, line 49: Wiring the new `editing/index` entry is not enough on its own. Even with `'use client'` added to `src/editing/index.ts`, the current build still strips module-level directives, so the emitted `dist/editing/index.js` does not retain `'use client'`. That means consumers of `@openmaic/renderer/editing` still resolve a file that looks like a server module to Next App Router, and function props like `onSelectionChange` can still fail serialization. This needs the build output to preserve directives, not just the source entry to declare them.

wyuc:
Fixed the build to actually preserve the directive — `3db9bac7`. Verified in the built `dist`, not just asserted:

```
dist/editing/index.js               -> "use client";   ✓
dist/editing/EditableSlideCanvas.js -> "use client";   ✓
dist/SlideCanvas.js  (v1)           -> "use client";   ✓ (was dropped before)
dist/index.js  (read-only barrel)   -> export { SlideCanvas } …   (no directive, stays clean)
```

Approach: enabled `preserveModules: true` (the config already had `preserveModulesRoot: 'src'`, inert without it) and added `rollup-plugin-preserve-directives`, so each source module's `'use client'` survives per-file. This is precise — only the modules that declare it get it; the read-only `.` barrel re-exports the client `SlideCanvas.js` and stays directive-free, so the read-only path is unchanged and still doesn't pull the editing bundle. It also fixes v1, which shipped without the directive.

One thing worth your eyes, since it's a judgment call: **this changes the published output layout** from bundled chunks to one file per source module (that's what the plugin requires). It's arguably the originally-intended structure given `preserveModulesRoot` was already there, and it tends to improve consumer tree-shaking — but it's a package-wide change to the emitted files, so I wanted to flag it rather than slip it in. If you'd prefer a smaller-footprint fix, the alternative is an `output.banner: "'use client';"` (keeps the current chunked layout but marks *every* emitted file client, including the type-only entry) — less precise but structurally minimal. Happy to switch if that's your preference.

(Note: the package build isn't run in CI, so this is verified locally against a real `pnpm build`, not by the CI checks.)

cosarah (approved):
LGTM

Issue #857, opened 2026-07-06
[Task] @openmaic/storage foundation — package skeleton + KVStore + AssetProvider (#779 Part 1)


wyuc:
Tracking issue: #779. This is the first executable slice of the `@openmaic/storage` RFC — the foundation that unblocks the rest, scoped to what does **not** depend on any open DSL item.

> Naming note: package scope is `@openmaic/*`, matching `@openmaic/renderer` (#851) and `@openmaic/exporter` (#852). The RFC (#779) was drafted before that convention settled and has been updated from `@maic/*` to `@openmaic/*` accordingly.

## Scope

Establish the package and land the two components that have no open prerequisite, plus the shared contract-test harness every later backend will reuse.

Dependency rule (same acyclic constraint as #720): `@openmaic/storage` → `@openmaic/dsl` only. Enforced the same way as the renderer boundary in #853.

### In scope

1. **Package skeleton** `packages/@openmaic/storage` — pure, zero runtime deps beyond `@openmaic/dsl`, `sideEffects: false`, `exports`/`types`, machine-enforced import boundary (mirror #853).
2. **`StorageProvider` interface into `@openmaic/dsl`** (small prerequisite). The RFC places this interface in the DSL (renderer already consumes media resolution; exporter will consume the asset manifest). `put(blob, meta?) → ref` / `resolve(ref) → url` / `remove(ref)`. This is the only DSL change in Part 1.
3. **`KVStore`** — `device` / `account` scopes; `get` / `set` / `remove` / `keys(prefix?)`.
   - Browser backend over `localStorage` (namespaced). `device` scope is defined as never leaving the device (invariant holds for every future backend).
   - Route `zustand` `persist` through a `KVStore`-backed `StateStorage` shim, removing the split-brain between `persist` and ad-hoc `localStorage` keys. Store business logic unchanged — only the storage source and scope change.
4. **`AssetProvider`** — implements the DSL `StorageProvider`. Browser backend: IndexedDB bytes + object URLs. Documents reference a **stable ref**, never a raw URL; `AssetProvider` resolves refs to URLs at read time.
5. **Contract-test harness** — one implementation-agnostic suite, run against a backend via a small conformance runner. Seed it with the `KVStore` and `AssetProvider` suites so later backends (HTTP, SQL) prove equivalence against the same cases. `device`-scope-stays-local and ref-round-trip are explicit cases.

### Explicitly deferred (later Parts, tracked under #779)

[...]

Issue #779, opened 2026-06-24
[RFC] @openmaic/storage: a pluggable persistence layer over the document / runtime / asset model


wyuc:
## Goal

Add a persistence layer to the `@maic` family so the app stops hard-wiring its
own storage stack and instead persists through a documented contract that
depends only on `@openmaic/dsl`. This is the storage counterpart to the epic in
#720: just as `renderer` / `importer` / `exporter` became packages that depend
only on the DSL, persistence should too.

Dependency rule (same acyclic constraint as #720): `@openmaic/storage` → `@openmaic/dsl`
only. The DSL owns *what* persists (shape + validation + migration + the blob
`StorageProvider` interface); `@openmaic/storage` owns *where/how* it persists
(backends, queries, and the aggregate↔table adapter).

## Where we are today

Client persistence is spread across three unrelated mechanisms with no shared
contract:

- IndexedDB via Dexie (`lib/utils/database.ts`) — the local object store
  (stages, scenes, chat sessions, outlines, generated media metadata, playback
  cursor, etc.).
- Ad-hoc `localStorage` keys.
- `zustand` `persist` middleware writing directly to `localStorage`.

Two consequences worth calling out:

1. **No backend seam.** The store is wired to the browser. There is no
   documented contract a server-backed deployment could implement.
2. **Migrations are trapped in the engine.** Schema migrations live in Dexie's
   versioned `upgrade()` callbacks, so they are IndexedDB-specific and cannot be
   reused by any other backend. Migration is therefore the thing that actually
   blocks backend pluggability.

## The model: four layers, one of them out of scope

Treating the persisted schema as the de-facto serialization boundary, the data
splits into four layers along two orthogonal axes — *does the DSL own its
shape?* and *does the instance travel with the core document?*

| Layer | Shape owner | Travels with the document? | Examples |
|---|---|---|---|
| **document** (aggregate) | `@openmaic/dsl` | yes | stage metadata, scenes (+ actions), agents, outline, quiz |
| **runtime** | `@openmaic/dsl` | no — exportable as a replay asset | chat sessions |
| **asset** (blob) | `StorageProvider` interface in `@openmaic/dsl` | by reference only | generated images / audio / video bytes |
| **device-local / cache** | not the DSL | no | playback cursor, local voice config, undo snapshots |

Out of scope for this layer entirely: business / operations / analytics /
billing data. Those belong to the consuming application, not to a portable
storage contract.

## Proposed components

[...]

PR #858, opened 2026-07-06
feat(storage): scaffold @openmaic/storage — KV + asset primitives (browser) (#857)

wyuc:
First executable slice of the `@openmaic/storage` RFC (#779) — **Part 1**, tracked in #857. A pure, app-agnostic persistence package that depends only on `@openmaic/dsl`, mirroring how `@openmaic/renderer` / `@openmaic/importer` sit on the DSL.

The DSL owns *what* persists; this package owns *where/how*. The pluggable seam is the **backend**, not the DB driver: this PR ships the browser backends (zero server) + the primitive contracts.

## What's in it

- **`StorageProvider` asset seam added to `@openmaic/dsl`** — `put(blob)→ref` / `resolve(ref)→url` / `remove(ref)`. Typed against a structural `BinaryBlob` (`size`/`type`/`arrayBuffer()`) so the pure DSL keeps `lib: ES2022` (no DOM) while owning the contract. A document embeds only a stable ref; a raw URL would bake in a provider + expiry and break portability.
- **`KVStore`** (`device` / `account` scopes) + **`BrowserKVStore`** over `localStorage`. `device` values (theme/locale/layout) never leave the device — the scope is part of the primitive, so it survives a future server backend.
- **`BrowserAssetProvider`** — content-addressed (`sha256-<hex>`) bytes in IndexedDB, resolved to object URLs; identical bytes de-duplicate to one asset/one ref.
- **`kvPersistStorage`** — adapts a `KVStore` into a zustand `persist` storage, so a store's own `version`/`migrate`/`merge` stay untouched while its bytes move off raw `localStorage`. Pure util here; **wiring the app's stores is a follow-up** (see below).
- **Implementation-agnostic contract suites** (`test/kv-contract.ts`, `test/asset-contract.ts`) run against the browser backends — the coming HTTP backend proves equivalence against the same cases.
- **Machine-enforced import boundary** in eslint (no `@/…` host-app paths), mirroring the `@openmaic/renderer` boundary (#853).

Backends take their `Storage` / `IDBFactory` by **injection**, so the package is testable without a browser (Node + `fake-indexeddb`).

## Notes for review

- **`BinaryBlob` vs `Blob` in the DSL.** I used a minimal structural type rather than pulling `"DOM"` into the DSL's `lib`, to preserve its env-agnostic purity. Happy to switch to `Blob` + DOM lib if you'd rather.
- **Naming.** `@openmaic/*` scope (RFC #779 has been updated from `@maic/*`).
- **Two-PR split.** This is the pure package with **zero app behavior change**. The zustand/`localStorage` reroute is deliberately *not* here: it renames the live `settings-storage` / `user-profile-storage` / `agent-registry-storage` keys, so it needs a legacy-key compat migration to avoid dropping existing users' settings — that carries real regression risk and deserves its own review.

## Deferred (later Part-1 / later parts, tracked in #857)

- Wire the app's zustand stores + ad-hoc `localStorage` through `KVStore` (with the legacy-key compat migration).
- `DocumentStore` (aggregate ↔ normalized adapter, migrate-on-read via the DSL migration registry, validation gate) — Part 2.
- `RuntimeStore` — needs a DSL-native runtime message type (still open in #720).
- HTTP backend + reference server + one HTTP contract.

## Verification

- `@openmaic/storage`: `pnpm test` → 21 passing (10 KV + 6 asset + 5 persist-adapter); `pnpm typecheck` + `pnpm build` clean.
- `@openmaic/dsl`: `pnpm test` → 126 passing (the `storage.ts` addition is purely additive); `pnpm build` regenerates schema unchanged.
- Repo `pnpm check` (prettier) clean; `pnpm lint` clean on the new package; the import-boundary rule verified to fire on a `@/…` import.

### DIFF
--- a/packages/@openmaic/storage/src/asset/browser.ts
+++ b/packages/@openmaic/storage/src/asset/browser.ts
@@ -1,0 +1,99 @@
+import type { AssetMeta, AssetRef, BinaryBlob, StorageProvider } from '@openmaic/dsl';
+
+export interface BrowserAssetProviderOptions {
+  /** IndexedDB factory. Defaults to the ambient `indexedDB`. Injectable for tests. */
+  indexedDB?: IDBFactory;
+  /** Database name. Defaults to `maic-assets`. */
+  dbName?: string;
+}
+
+const STORE = 'assets';
+
+interface StoredAsset {
+  bytes: ArrayBuffer;
+  contentType: string;
+}
+
+function toHex(buffer: ArrayBuffer): string {
+  const view = new Uint8Array(buffer);
+  let hex = '';
+  for (let i = 0; i < view.length; i++) hex += view[i].toString(16).padStart(2, '0');
+  return hex;
+}
+
+/**
+ * Browser `StorageProvider` backend: bytes live in IndexedDB, and `resolve`
+ * hands back a `blob:` object URL for use as a media `src`. Refs are
+ * content-addressed (`sha256-<hex>`), so identical bytes de-duplicate to one
+ * stored asset and one stable ref.
+ */
+export class BrowserAssetProvider implements StorageProvider {
+  private readonly idb: IDBFactory;
+  private readonly dbName: string;
+  private dbPromise?: Promise<IDBDatabase>;
+  /** ref → object URL, so repeated `resolve` returns one stable URL per asset. */
+  private readonly urls = new Map<AssetRef, string>();
+
+  constructor(options: BrowserAssetProviderOptions = {}) {
+    this.idb = options.indexedDB ?? globalThis.indexedDB;
+    this.dbName = options.dbName ?? 'maic-assets';
+  }
+
+  private openDb(): Promise<IDBDatabase> {
+    this.dbPromise ??= new Promise((resolve, reject) => {
+      const req = this.idb.open(this.dbName, 1);
+      req.onupgradeneeded = () => {
+        if (!req.result.objectStoreNames.contains(STORE)) req.result.createObjectStore(STORE);
+      };
+      req.onsuccess = () => resolve(req.result);
+      req.onerror = () => reject(req.error);
+    });
+    return this.dbPromise;
+  }
+
+  private async tx<T>(
+    mode: IDBTransactionMode,
+    run: (store: IDBObjectStore) => IDBRequest<T>,
+  ): Promise<T> {
+    const db = await this.openDb();
+    return new Promise<T>((resolve, reject) => {
+      const transaction = db.transaction(STORE, mode);
+      const req = run(transaction.objectStore(STORE));
+      req.onsuccess = () => resolve(req.result);
+      req.onerror = () => reject(req.error);
+      transaction.onabort = () => reject(transaction.error);
+    });
+  }
+
+  private async computeRef(data: BinaryBlob): Promise<{ ref: AssetRef; bytes: ArrayBuffer }> {
+    const bytes = await data.arrayBuffer();
+    const digest = await crypto.subtle.digest('SHA-256', bytes);
+    return { ref: `sha256-${toHex(digest)}`, bytes };
+  }
+
+  async put(data: BinaryBlob, meta?: AssetMeta): Promise<AssetRef> {
+    const { ref, bytes } = await this.computeRef(data);
+    const asset: StoredAsset = { bytes, contentType: meta?.contentType ?? data.type ?? '' };
+    await this.tx('readwrite', (store) => store.put(asset, ref));
+    return ref;
+  }
+
+  async resolve(ref: AssetRef): Promise<string | null> {
+    const cached = this.urls.get(ref);
+    if (cached) return cached;
+    const asset = await this.tx<StoredAsset | undefined>('readonly', (store) => store.get(ref));
+    if (!asset) return null;
+    const url = URL.createObjectURL(new Blob([asset.bytes], { type: asset.contentType }));
+    this.urls.set(ref, url);
+    return url;
+  }
+
+  async remove(ref: AssetRef): Promise<void> {
+    await this.tx('readwrite', (store) => store.delete(ref));
+    const url = this.urls.get(ref);
+    if (url) {
+      URL.revokeObjectURL(url);
+      this.urls.delete(ref);
+    }
+  }
+}
--- a/packages/@openmaic/storage/src/asset/browser.ts
+++ b/packages/@openmaic/storage/src/asset/browser.ts
@@ -92,7 +92,22 @@ export class BrowserAssetProvider implements StorageProvider {
     const { ref, bytes } = await this.computeRef(data);
     const asset: StoredAsset = { bytes, contentType: meta?.contentType ?? data.type ?? '' };
     await this.tx('readwrite', (store) => store.put(asset, ref));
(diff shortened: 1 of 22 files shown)


wyuc:
## Cross-review outcome

Reviewed independently by Codex (`codex exec review`) and a Claude multi-angle pass (correctness / cleanup+altitude / conventions+packaging) over the same `origin/main..HEAD` diff. Findings merged by consensus.

### Fixed in `fix(storage): address cross-review findings`

Correctness / CI:
- **`resolve` on `transaction.oncomplete`, not `request.onsuccess`** — a write that succeeds as a request can still abort at commit (QuotaExceeded); the old code reported that as durable success.
- **`openDb` no longer memoizes a rejected open** — a transient IDB failure (private-mode, one-off VersionError) bricked the provider for the whole session; it now retries.
- **`resolve` memoizes per ref** — concurrent `resolve(sameRef)` calls previously each minted an object URL, orphaning one (unrevocable leak). They now share one.
- **`KVStore.set` guards non-JSON values** — `set(k, undefined)` stored the literal `"undefined"`, which threw on the next `get` and would reject zustand rehydration; it's now treated as a removal.
- **Contract Blobs built from strings** — a `Uint8Array` BlobPart fails the root `tsconfig` typecheck (`next build`) under TS 5.7+ typed-array generics. (This one only shows under the root config, not the package's own `tsc`.)
- Tests: per-test object-URL registry reset; regression tests for `set(undefined)`, concurrent-resolve, and actual de-dup (one stored row).

Packaging:
- **Dropped `@openmaic/storage` from the root `postinstall` chain** — nothing imports it yet, so building it on every install added latency and `&&`-coupled its build failures to importer/renderer/sync. It re-joins when app code consumes it.

### Deferred — design calls for a maintainer (not silently dropped)

1. **Interface name `StorageProvider` collides** with the existing app-internal `lib/storage/types.ts` `StorageProvider` (different shape: `upload/getUrl`, Buffer/hash-in). The DSL one pairs with `AssetRef`/`AssetMeta` — worth renaming to `AssetStore` / `AssetProvider`? (Touches the RFC #779 wording.)
2. **Ref format `sha256-<hex>` vs bare hex.** The existing server layer (`lib/storage/client.ts`, `/api/storage/upload`) keys by bare hex. `AssetRef` is opaque, but the **Part-2 HTTP backend must map** or dedup breaks against that endpoint. Flagging for Part 2.
3. **Second IndexedDB store** (`maic-assets`) sits parallel to the app's Dexie DB during migration → transient double quota. Convergence is the point of this layer; noting the interim.
4. **Single `.` export barrel is DOM-typed.** When the server/HTTP backend lands, a Node consumer of the pure `kvPersistStorage`/KV types would pull the DOM-typed browser backends. Consider `./kv` / `./asset` / `./zustand` / `./server` subpaths then (as `@openmaic/renderer` does).
5. **`KVStore.get` can't distinguish stored `null` from missing** — acceptable given the `T | null` signature; noted.

Post-fix: package `pnpm test` 24 passing, `typecheck` + `build` clean, root-config typecheck of the package clean, repo `pnpm check` (prettier) + `pnpm lint` clean.

wyuc:
### Cross-review converged

A second Codex pass on the fixed branch caught that the per-ref `resolve` memoization had reintroduced the same rejected-promise-caching bug just fixed in `openDb` (a transient IDB failure inside `resolve` would be replayed on every later call). Fixed in `fix(storage): don't cache a rejected resolve promise` with a regression test that fails a resolve's IDB open once and asserts recovery (verified it fails without the fix).

Third pass: **no correctness findings** — converged. Remaining items are the design calls listed above (interface naming, ref format, subpaths), left for a maintainer decision. Final state: 25 tests passing, package + root-config typecheck clean, `pnpm check` / `pnpm lint` clean.

packages/@openmaic/storage/src/asset/browser.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  private async computeRef(data: BinaryBlob): Promise<{ ref: AssetRef; bytes: ArrayBuffer }> {
+    const bytes = await data.arrayBuffer();
+    const digest = await crypto.subtle.digest('SHA-256', bytes);
+    return { ref: `sha256-${toHex(digest)}`, bytes };
+  }
+
+  async put(data: BinaryBlob, meta?: AssetMeta): Promise<AssetRef> {
+    const { ref, bytes } = await this.computeRef(data);
+    const asset: StoredAsset = { bytes, contentType: meta?.contentType ?? data.type ?? '' };
```
cosarah, line 93: Because the ref is derived only from the bytes, this write can mutate `contentType` for an existing ref while `resolve()` keeps returning a previously cached object URL. I reproduced: `put(blob(type="")) -> resolve(ref) -> put(same bytes, type="image/png")`; the same provider still returns the old URL (`Blob.type === ""`), while a fresh provider resolves the same ref as `image/png`. That makes MIME behavior depend on cache warmth. Either make metadata part of the identity, make first-write win, or invalidate the cached URL when an existing ref is overwritten.

packages/@openmaic/storage/vitest.config.ts
```diff
@@ -0,0 +1,24 @@
+import { defineConfig } from 'vitest/config';
+import { fileURLToPath } from 'node:url';
+
+// Resolve `@openmaic/storage` self-imports and the `@openmaic/dsl` dependency to
+// their package sources, so `pnpm test` is standalone on a clean checkout (no
+// `dist` build required). Consumers still resolve via each package's `exports`
```
cosarah, line 6: The test aliasing here is fine, but it also masks a workspace bootstrap gap: this package only exports `dist/*`, and the repo-level install/build flow still does not build `@openmaic/storage`. On a clean checkout, `pnpm install` leaves this package without `dist`, so any workspace consumer resolving `@openmaic/storage` through the normal exports path will fail until someone runs `pnpm --filter @openmaic/storage build`. Can we either add this package to the root bootstrap build chain or document a source-alias strategy for local consumers?

packages/@openmaic/storage/src/asset/browser.ts
wyuc: Good catch — went with cache invalidation (your third option). `put` now revokes and drops any cached object URL for the ref after the write, so `resolve` reflects the latest write and a warm provider matches a fresh one. Kept the write itself last-write-wins (so a corrected `contentType` for the same bytes takes effect rather than being pinned to the first, possibly empty, type). Added a regression test (`put type="" -> resolve -> put same bytes type="image/png" -> resolve reports image/png`; verified it fails without the fix). 26b40cb5

packages/@openmaic/storage/vitest.config.ts
wyuc: Agreed — restored `@openmaic/storage` to the root `postinstall` build chain (right after `@openmaic/dsl`, which it depends on), matching `dsl`/`importer`/`renderer`. A clean `pnpm install` now builds its `dist`, so a workspace consumer resolving it through `exports` works without a manual `--filter` build. (I'd removed it in an earlier round to avoid building a not-yet-consumed package, but you're right that the half-built state is the worse trade-off.) 26b40cb5

cosarah (approved):
LGTM, verified locally

PR #859, opened 2026-07-06
feat(renderer/editing): selection + drag-to-move core (@openmaic/renderer v2, #851 Part A)

wyuc:
## Summary

First slice of the `@openmaic/renderer/editing` surface (#851, #720 Phase 3): the pure gesture core plus a controlled `EditableSlideCanvas` that does selection and single-element drag-to-move, emitting one `EditIntent` per completed gesture. Nothing in the app imports `@openmaic/renderer/editing` yet, so this merges inert — no runtime behavior change.

## What's included

- `editing/core/geometry.ts` — pure bounds/rotation helpers (`getElementRange`, `getElementListRange`, `uniqAlignLines`, `AlignLine`, px↔canvas-unit). React-free, `@openmaic/dsl`-only.
- `editing/core/snapping.ts` — alignment lines from element edges/centers plus the six viewport lines; nearest-in-range correction; transient `Guide` data.
- `editing/core/drag.ts` + `intent.ts` — single-element drag math (delta, axis-lock, snapping) and `element.update` intent derivation.
- `editing/handles/` — presentational `BorderLine` + `SelectionOverlay`.
- `EditableSlideCanvas` — fills the Stage-0 shell: controlled `selection`/`onSelectionChange`, drag-to-move with an internal working copy (60fps feedback), emitting one `element.update` on pointer-up (a click emits selection only). Renders a separate sibling hit/selection overlay so the v1 read-only `SlideCanvas` is left untouched, and mirrors its centering offset so hit-testing stays aligned in letterboxed containers.
- Test infra: jsdom + Testing Library, `.tsx` test include, a `ResizeObserver` polyfill.

## Deferred to later slices (not in this PR)

Resize / rotate / line endpoints / shape keypoints / multi-select; visual guide/grid/ruler rendering; ProseMirror inline text editing (Part B); app-side surface wiring + registry swap (the parity flip).

## Tests & boundary

- 18 tests across the editing subpath; full package suite green; `tsc` clean; `dist/editing/*` builds.
- Respects the `@openmaic/renderer` import boundary (#853): no host `@/…` imports.


cosarah (commented):
I found three behavioral issues in the current implementation:

1. `packages/@openmaic/renderer/src/editing/useEditGesture.ts:59` accepts a controlled `selection`, but the hook never uses it when deciding what a drag should do. Dragging an unselected element does not first update selection, and dragging one member of a multi-selection still emits a single-element move intent. That can leave the controlled selection out of sync with the element that actually moved.

2. `packages/@openmaic/renderer/src/editing/EditableSlideCanvas.tsx:114` filters line elements out of the hit layer, and `packages/@openmaic/renderer/src/editing/handles/SelectionOverlay.tsx:18` skips them as well. With overlapping content, a pointer-down on a line will fall through to the box element underneath and select or move the wrong thing. Even if line editing is deferred, we still need a blocker or a dedicated hit area so the event does not route to another element.

3. In select-only hosts (`onSelectionChange` without `onElementsChange`), `packages/@openmaic/renderer/src/editing/useEditGesture.ts:121` still publishes a live `workingSlide` during `pointermove`. The element visibly follows the pointer and then snaps back on `pointerup`, which makes a read-only/select-only surface look draggable even though nothing can commit. I would gate live movement on `onElementsChange` as well.

wyuc:
Thanks for the review — all three points are addressed:

**1. Selection sync on drag.** The gesture now selects the target element on pointer-down, so a drag always ends with the dragged element selected and the controlled `selection` stays in sync. Moving all members of a multi-selection *together* is out of scope for this slice (it belongs to the multi-select work); for now, dragging a member collapses the selection to that element and moves just it.

**2. Line click routing.** Line elements now render a dedicated inert hit blocker shaped like the actual stroke — an SVG `<path>` reusing the same `getLineElementPath` the renderer draws, with `pointer-events: stroke` and a grab band of `max(10px, el.width * scale)`. Pointer-downs on the visible line — straight, elbow (`broken`/`broken2`) or curved (`curve`/`cubic`) — are consumed instead of routing to elements underneath, while the off-stroke bbox area stays click-through. One gap I've deferred (with a code note): endpoint markers (arrow/dot) can paint beyond the stroke and aren't part of the hit target yet — the only fall-through case is a *marked* line overlapping other content exactly at an endpoint. Since line editing itself is deferred, I planned to cover marker extents when that lands, but I'm happy to pull it in now if you'd prefer.

**3. Select-only hosts.** Live movement during a drag is now gated on `onElementsChange`. A select-only surface no longer shows the element following the pointer and snapping back on release — a pointer interaction just updates the selection.

Pushed; CI is green. Re-requesting your review.

cosarah (commented):
I think the latest line-blocker change introduced a selection regression for line elements.

`packages/@openmaic/renderer/src/editing/EditableSlideCanvas.tsx:137` now renders an inert SVG blocker for `type === "line"`, but its `onPointerDown` only calls `stopPropagation()`, so clicking a line no longer emits `onSelectionChange`. On top of that, `packages/@openmaic/renderer/src/editing/handles/SelectionOverlay.tsx:18` still filters line elements out, so even a host-controlled line selection has no visible feedback.

Before this series, `EditableSlideCanvas` delegated selection to `SlideCanvas`'s `onElementClick`, which meant line elements were still selectable even though line editing was not implemented yet. The new blocker correctly prevents fall-through to an overlapping box element, but it also makes lines effectively unselectable. I think the blocker should stay drag-inert while still selecting the line, and this should get a regression test.

wyuc:
Good catch — the blocker had made lines unselectable. Fixed: line elements are now **selectable but still not draggable**.

- The line's stroke blocker selects the line on pointer-down (`onSelectionChange({ elementIds:[id], primaryId:id })`), matching how box elements select, while staying drag-inert (no gesture armed, no `element.update`) and still consuming the pointer so it never falls through to an overlapping box.
- `SelectionOverlay` now draws a border for a selected line. Its bounds are derived from the **actual rendered path** (`getLineElementPath`), so straight, elbow (`broken`/`broken2`) and curved (`curve`/`cubic`) lines are each enclosed correctly — including the `broken2` one-axis case, where the border stays flat instead of spanning the raw control point.
- Two consistency touch-ups from a second pass: a **locked** line is inert like a locked box (blocks, does not select), and re-clicking an already-selected line no longer re-emits `onSelectionChange`.
- Added regression tests: click-selects-a-line, selected-line-shows-a-border, curved/`broken2` bounds, locked-line-not-selectable.

Two things I deliberately deferred (documented in code), both tied to work that isn't in this slice — happy to pull either in if you'd rather:
- **Endpoint markers** (arrow/dot) can paint beyond the stroke; their hit extents aren't in the blocker yet — deferred with line editing.
- Line selection bypasses the gesture hook's active-pointer guard, so a second finger on a line during an in-flight box drag could change selection — a multi-touch edge, deferred with the multi-select/multi-touch work; single-pointer is unaffected.

Pushed; CI is green. Re-requesting your review.

cosarah (commented):
I think `getLineBounds()` is still overstating the selection box for curved lines.

In `packages/@openmaic/renderer/src/editing/core/geometry.ts:35`, the implementation says it derives the bounds of the “actual drawn path”, but for `Q` / `C` paths it only regex-parses the numbers from `getLineElementPath()` and takes min/max over the endpoints plus control points. That is not the Bézier curve's true bounding box: control points can sit outside the rendered curve.

For example, `M0,0 Q50,80 100,0` only reaches `y = 40`, but the current code returns `maxY = 80`, so `SelectionOverlay` ends up drawing a selection border much taller than the visible line. The new expectations in `packages/@openmaic/renderer/test/editing/core/geometry.test.ts:50` and `packages/@openmaic/renderer/test/editing/handles/SelectionOverlay.test.tsx:70` are currently locking in that oversized-box behavior.

If we want the border to match the rendered stroke, this needs real quadratic/cubic Bézier bbox math (or an equivalent path-bounds helper), not control-point min/max.

wyuc:
You're right — control-point min/max overstated the box. `getLineBounds` now computes the **true** curve bounding box.

- New pure helpers (`quadraticBounds`, `cubicBounds`, `getPathBounds` in `editing/core/geometry.ts`) parse the rendered path and, for each `Q`/`C` segment, take the endpoints plus the actual extrema (`B'(t)=0` roots in `(0,1)`) — never the raw control points. So `M0,0 Q50,80 100,0` now bounds to `maxY = 40` (the real peak), not 80; cubics use the derivative roots likewise.
- The two expectations you flagged (`geometry.test.ts` and `SelectionOverlay.test.tsx`) are updated to the correct tight values, and I added direct unit tests for the quadratic peak (40) and hand-computed cubic extrema (e.g. `C10,-30 90,50 100,0` → `minY ≈ -6.944`, `maxY = 16.875`; a symmetric case → `67.5`, not the `90` control point).
- Also hardened the path tokenizer to accept exponent-form coordinates (e.g. `7.1e-15` from a rotated/imported line) so they parse as one number rather than being split.

Pushed; CI green. Re-requesting review.

cosarah (commented):
I think the latest line-bounds fix is still missing the rendered stroke width.

`packages/@openmaic/renderer/src/editing/core/geometry.ts:227` now computes a tight bbox for the line path centerline, which fixes the old control-point oversizing for `Q` / `C` curves. But `packages/@openmaic/renderer/src/elements/line/BaseLineElement.tsx:114` renders the actual line with `strokeWidth={elementInfo.width}`, and that width is never reflected in `getLineBounds()`.

As a result, `packages/@openmaic/renderer/src/editing/handles/SelectionOverlay.tsx:35` can still draw a selection border that is visibly too small for thick lines. For example, a thick horizontal line still gets `minY === maxY` from `getLineBounds()`, even though the visible stroke extends by `width / 2` above and below the centerline. So the current implementation does not yet match the stated goal of wrapping the visible stroke.

I think `getLineBounds()` needs to inflate the geometric path bbox by half the rendered stroke width on both axes (and probably keep markers as a separate deferred concern). There also is not a regression test yet for a nontrivial `elementInfo.width`.

wyuc:
Good point — the tight path bbox only bounded the centerline. `getLineBounds` now inflates the path bounding box by **half the rendered stroke width** (`line.width`, what the v1 renderer draws as `strokeWidth`) on every side, so the border wraps the visible stroke. A thick horizontal line now gets a `width`-tall border instead of a zero-height one (`minY = -width/2`, `maxY = +width/2` around the flat centerline), and the same half-width padding applies on all sides for every line shape.

Added a regression test for a nontrivial `width` (a `width:10` horizontal line → bounds inflated by 5 on each side). Endpoint markers stay a separate deferred concern as noted.

Also brought the branch up to date with `main` (it was BEHIND). Pushed; CI green. Re-requesting review.

cosarah (commented):
I think the latest fix is still solving this in local patches instead of modeling the full rendered geometry of a line.

Concretely, `packages/@openmaic/renderer/src/editing/core/geometry.ts:227` now inflates the path bbox by `strokeWidth / 2`, which fixes thick straight lines, but it can still under-bound thick polyline joins. In `packages/@openmaic/renderer/src/elements/line/BaseLineElement.tsx:114`, the browser is rendering a stroked SVG path with the default SVG stroke behavior, including joins. For a thick `broken` line such as `M0,0 L50,50 L100,0`, the mitered corner can extend past the current `+/- width/2` inflation, so `SelectionOverlay` can still clip the visible stroke.

More importantly, I do not think this should keep being fixed case by case. The sequence so far has been: skip lines entirely, then add a blocker, then make lines selectable, then bound path coordinates, then compute Bézier extrema, then parse exponent-form numbers, then inflate by stroke width. The reason this keeps resurfacing is that we still do not have a single model of the *actual rendered line geometry*. `getLineElementPath()` gives the centerline path, but the visible output in `BaseLineElement` is centerline + stroke width + join/cap behavior + endpoint markers. As long as selection bounds are derived from partial approximations of that stack, more edge cases will keep appearing.

I see two viable directions:

1. Introduce a real rendered-geometry helper, e.g. `getRenderedLineBounds()`, and make it the single source of truth for line selection bounds. That helper should model the rendered path together with stroke width, stroke join/cap behavior, and eventually endpoint markers. Then both `SelectionOverlay` and its tests can depend on that one contract.

2. Narrow the scope of this PR explicitly: keep line handling limited to selectable / non-draggable / no click-through, but document that the line selection border is only approximate for now and defer pixel-tight visual fidelity to the dedicated line-editing slice.

Without one of those two choices, I expect this area to continue accumulating one more geometry patch per newly discovered line shape.

wyuc:
You're right that the bbox-derivation kept accumulating patches because it was approximating the rendered stack instead of modeling it. Rather than keep refining that, I've **replaced the whole approximate-bbox approach with a real line editor**, ported from the app's `LineElementOperate` + `useDragLineElement`:

- A selected line now shows draggable **endpoint handles** (`start`/`end`, plus a control handle for `broken`/`broken2`/`curve` and two for `cubic`) that reshape the line — pointer-drag → `computeLineDrag` (a faithful, unit-tested port of the app's move + bbox-normalization + endpoint-straighten-snap math) → one `element.update` intent (`{left, top, start, end, +control}`). Selectable/not-draggable/no-click-through still hold; a locked line stays inert.
- **Selection chrome is now the exact rendered stroke, not a computed box.** A selected line highlights its actual `getLineElementPath` stroke (the same path the renderer draws) and, when editable, layers the handles on top. That gives feedback in every state — editable, read-only (no-callback), and locked — with zero bbox approximation.
- **Removed `getLineBounds` and the entire quadratic/cubic-bbox chain** — it's superseded by the real editor + stroke highlight, so there's no longer a partial-approximation of the rendered geometry to keep chasing.

Deferred as documented refinements (they match the app but aren't core to a usable editor): snapping a dragged endpoint to *other* elements' anchor points, the Ctrl/Shift "keep the control point in place" modifier, and endpoint-marker hit extents — all noted for the follow-up. The line-endpoint drag and the box gesture each keep their own single-pointer guard (multi-touch out of scope, as before).

Net: lines are now a first-class editor alongside the other elements, and the selection geometry is sourced from the actual rendered path instead of an approximation. Pushed; CI green. Re-requesting review.

packages/@openmaic/renderer/src/editing/EditableSlideCanvas.tsx
```diff
@@ -1,41 +1,370 @@
...
+
+          {/* Line handles: a selected, unlocked line's endpoint/control handles
+              are its selection chrome (SelectionOverlay no longer draws a line
+              border). Rendered above the stroke blocker so a handle pointer-
+              down hits the handle, not the blocker beneath. Each handle reads
+              the WORKING line (via `displayElements`), so during a reshape drag
+              the handles track the previewed geometry. Handles use absolute
+              SCREEN coordinates (they bake in `viewportStyles`), so they sit as
+              direct children of the overlay, NOT inside the offset container. */}
+          {interactive &&
```
cosarah: I think the current state model is still conflating "selectable" with "editable".

`interactive` becomes true even when the host only passes `onSelectionChange`, so a select-only mount still renders `LineHandles` here. But `useLineHandleGesture` intentionally no-ops without `onElementsChange`, and the regression test now locks that behavior in. The result is a surface that shows draggable endpoint / control handles even though dragging them can never preview or commit anything.

That is the same class of problem the earlier select-only box-drag fix addressed: the UI is advertising an edit affordance the host cannot actually perform. I think line handles should be gated on editability (`onElementsChange`) rather than generic interactivity, with the selected stroke highlight remaining as the selection-only feedback.

packages/@openmaic/renderer/src/editing/EditableSlideCanvas.tsx
```diff
@@ -1,41 +1,370 @@
...
+                      fill="none"
+                      // Selected -> a visible accent highlight (feedback in
+                      // read-only/locked/editable alike); unselected -> a
+                      // transparent fat grab band. Either way the path stays
+                      // inert (no drag armed); pointer-events is disabled
+                      // entirely in a read-only mount so a selected-but-
+                      // read-only line never captures the pointer.
+                      stroke={isSelected ? '#3b82f6' : 'transparent'}
+                      strokeOpacity={isSelected ? 0.7 : undefined}
+                      strokeWidth={isSelected ? highlightCanvas : grabCanvas}
```
cosarah: I think this still introduces a hit-testing regression for selected lines.

Right now the same `<path>` is doing double duty as both the visible selection chrome and the interaction blocker. When `isSelected`, `strokeWidth` switches from `grabCanvas` to `highlightCanvas`, so the effective `pointer-events: stroke` band shrinks from `max(10px, el.width * scale)` down to the much thinner highlight width. For a thin selected line over a box, clicks inside the old blocker band but outside the highlight can now fall through to the box underneath, which breaks the earlier no-click-through guarantee as soon as the line becomes selected.

I do not think this should be fixed case by case. The underlying problem is that hit geometry and visual chrome are still being modeled as one stroke. Two viable directions seem reasonable here: either (1) split this into two paths — an invisible fat blocker that always keeps the stable hit region, plus a visible highlight path with `pointer-events: none`; or (2) introduce one rendered-line-affordance helper that returns both the visual chrome geometry and the hit geometry so selected / unselected / locked states all consume the same contract.

wyuc:
Both addressed:

**1. Handles gated on editability.** `LineHandles` now renders only when `onElementsChange` is provided (plus selected + unlocked), not on generic interactivity. A select-only mount (`onSelectionChange` without `onElementsChange`) shows just the stroke highlight as selection feedback — no draggable handles the host can't commit. (Test updated to lock this in.)

**2. Split hit geometry from visual chrome.** Took your direction (1): the line now renders **two** paths off the same `getLineElementPath`:
- an invisible **blocker** — always `stroke: transparent`, `pointer-events: stroke`, `strokeWidth = max(10px, el.width*scale)`. Its hit band no longer changes with selection, so the no-click-through region is stable whether or not the line is selected.
- a separate visible **highlight** (`pointer-events: none`, `strokeWidth = max(2px, el.width*scale)`), rendered only when selected — purely visual, never affects hit-testing.

So the stable hit region and the selection chrome are now independent contracts. Added a test asserting a selected thin line keeps its full grab band (selection doesn't shrink hit-testing) and the highlight is non-interactive.

Pushed; CI green. Re-requesting review.

cosarah (approved):
LGTM, verified locally

PR #860, opened 2026-07-06
feat(storage): DocumentStore — normalized document aggregate (browser backend)

wyuc:
Part 2 of the `@openmaic/storage` rollout (RFC #779, part breakdown #857), following the KV / asset primitives in #858.

## What

Adds `DocumentStore` — persistence for the `document` aggregate (a course: stage metadata + scenes + embedded agents / quiz / actions + an opaque outline snapshot) — with a browser (IndexedDB) backend. Depends only on `@openmaic/dsl`; no Dexie dependency (raw injected `IDBFactory`, matching the Part 1 asset backend).

Implements the three mechanisms from #779:

1. **Aggregate ↔ normalized adapter.** The portable embedded document splits into per-entity rows on write (diffing so only changed scenes are written; removed scenes deleted) and reassembles on read. Scenes key by `[stageId, id]` so documents stay isolated even if two courses reuse a scene id.
2. **Migrate-on-read/write.** Each document is root-stamped with `dslVersion`. Reads run the DSL migration ladder forward — a newer-versioned document passes through untouched (never downgraded). `saveDocument` normalizes stale input forward (or fails loud on a version with no ladder path) and refuses to overwrite a newer stored document; the incremental `putScene` / `deleteScene` require an already-current document so they can never strand unmigrated sibling scenes above the read line. The opaque outline rides alongside — persisted verbatim, never validated or migrated.
3. **Validation gate.** Writes run `validateStage` / `validateScene` and fail loud before any row is written (atomic). The store additionally enforces its own key/sort invariants — a scene belongs to its stage, `id` is a string, `order` is a number, no duplicate ids — independently of the content validator.

## Public API (DSL-typed repositories)

`saveDocument` / `loadDocument` / `listDocuments` / `deleteDocument`, plus incremental `putScene` / `getScene` / `deleteScene`.

The store is **generic over the scene type**: `DocumentStore<TScene>` defaults to the DSL `Scene` (universal `slide` / `quiz`). An app that widens `Scene` with its own kinds (`interactive` / `pbl`, content the DSL does not own) parameterizes the store over its scene union and injects a matching `validateScene`, so those scenes persist and the gate stays fail-loud for the app's shapes.

## Open questions from #779, resolved

- **Public API shape (#779 Q1):** DSL-typed repositories (not a generic `collection<T>`).
- **Outline update policy (#779 Q2):** original-intent snapshot — persisted verbatim, opaque to the store.

## Tests

- One **backend-agnostic contract suite** (`test/document-contract.ts`), run against the browser backend, so a future HTTP backend is proven equivalent by the same suite.
- Backend-seeded cases for migrate-on-read, forward-compat (no downgrade), version-guard rejections, and the app-widened scene union.
- Pure adapter unit tests for split / reassemble.

## Scope / non-goals (this part)

HTTP backend + reference server, `RuntimeStore`, asset-manifest wiring, and routing the app's existing storage through this store are later parts.

### DIFF
--- a/packages/@openmaic/storage/README.md
+++ b/packages/@openmaic/storage/README.md
@@ -39,8 +39,8 @@ a browser.
   would bake in a provider + expiry and break portability).
 - **Document normalization.** The DSL `document` is a portable embedded
   aggregate; `DocumentStore` normalizes it into per-entity rows so scene-level
-  saves stay cheap (writes diff and touch only changed scenes) and reassembles
-  it on read. Each document is stamped with a `dslVersion`; reads run the DSL
+  writes (`putScene`) stay cheap, and reassembles it on read. Each document is
+  stamped with a `dslVersion`; reads run the DSL
   migration ladder forward, and writes are validated against the DSL gate
   (`validateStage` / `validateScene`) so schema drift fails loud. The outline is
   an opaque, app-owned snapshot carried alongside — persisted verbatim, neither
--- a/packages/@openmaic/storage/README.md
+++ b/packages/@openmaic/storage/README.md
@@ -20,13 +20,14 @@ No dependency on React, zustand, or any host app. Backends take their `Storage`
 / `IDBFactory` by injection, so the package is app-agnostic and testable without
 a browser.
 
-## What's in here (Part 1)
+## What's in here
 
 | Export | Role | Browser backend |
 | --- | --- | --- |
 | `KVStore` | small `device` / `account`-scoped values not owned by the DSL | `BrowserKVStore` over `localStorage` |
 | `StorageProvider` (from `@openmaic/dsl`) | the asset seam: `put(blob) → ref`, `resolve(ref) → url`, `remove(ref)` | `BrowserAssetProvider` over IndexedDB + object URLs |
 | `kvPersistStorage` | adapt a `KVStore` into a zustand `persist` storage | — |
+| `DocumentStore` | persist the DSL `document` aggregate (stage + scenes + embedded agents / quiz / actions + an outline snapshot) | `BrowserDocumentStore` over IndexedDB (normalized `stages` / `scenes` / `outlines`) |
 
 - **Scopes.** `account` values are user data a server-backed deployment syncs
   across devices; `device` values (theme, locale, layout) never leave the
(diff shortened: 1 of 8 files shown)


wyuc:
**Review notes — correctness hardening**

This change went through independent adversarial review focused on data-safety and the version model. Summary of the invariants that were probed and are enforced + tested, to help focus review:

**Scene / stage integrity**
- A scene whose `stageId` disagrees with its document is rejected — it would otherwise land in a compound-key partition no read path for that document can reach (silent data loss). Enforced independently of the content validator, together with `id`-is-string, `order`-is-number, and no-duplicate-scene-ids.

**Version model (every write path guarded)**
- `saveDocument` rejects a document written by a newer client (never downgrades), normalizes stale input forward through the DSL migration ladder (or fails loud when there is no path), and — inside the write transaction, so there is no TOCTOU — refuses to overwrite a stored document newer than this client.
- `putScene` / `deleteScene` require the stored document to be current, so an incremental mutation can never mark a document current while leaving unmigrated sibling scenes stranded above the migrate-on-read line, nor downgrade a newer one.
- `deleteDocument` is intentionally *not* version-guarded — a whole-document removal is a deliberate coarse action.
- The opaque `outline` is excluded from `migrate()` entirely, so a future migration transform cannot silently drop it; scenes stay inside the migrated core because the DSL owns the scene contract.

**Atomicity**
- All validation / integrity / version checks run before any write (the stored-row version check runs inside the transaction but prior to any mutation); a rejection leaves the stores untouched.

**Generic over scene type**
- `DocumentStore<TScene>` + an injectable `validateScene` let an app persist a widened scene union without the store assuming slide/quiz-specific fields, and the injected validator cannot weaken the store's own key/sort invariants.

All of the above are covered by the backend-agnostic contract suite plus backend-seeded cases (migrate-on-read, forward-compat, version-guard rejections, malformed stamp, app-widened scenes). Package tests green; typecheck / prettier / eslint clean.

cosarah (commented):
1. `packages/@openmaic/storage/src/document/browser.ts:277`

The scene diff in `saveDocument` uses `JSON.stringify(prev) !== JSON.stringify(scene)`, but the store contract says everything below `SceneLike` is opaque. IndexedDB can persist structured-clone values that JSON flattens or drops, so this can silently skip real edits for app-owned scene payloads. I reproduced it with a custom `interactive` scene whose `content.data` changed from `new Map([['k', 1]])` to `new Map([['k', 2]])`; the second `saveDocument` was skipped and `getScene` still returned `1`. We need a comparison that preserves opaque payloads here, or we should always rewrite existing scene rows.

2. `packages/@openmaic/storage/src/document/browser.ts:310`

`listDocuments` reads raw stage rows and returns them without touching `dslVersionOf` / `needsMigration`. That means malformed `dslVersion` stamps are treated as hard errors by `loadDocument`, `getScene`, and `saveDocument`, but still show up as healthy documents here. I restamped a saved row to `dslVersion = "not-a-version"`: the other APIs threw, while `listDocuments` still returned the document normally. This should apply the same corruption/version guard, or explicitly document that `listDocuments` intentionally tolerates invalid rows.

wyuc:
Thanks — both are good catches. Addressed in 1463717c.

**1. Scene diff skipping opaque edits.** You're right: `JSON.stringify` equality became unsafe once the store went generic over opaque scene content, because IndexedDB persists structured-clone values (`Map`/`Set`/`Date`, `NaN`, `undefined` members) that JSON flattens or drops. Replaced it with a `structuralEquals` that recognizes those kinds and — critically — **declines to prove equality for anything it doesn't recognize** (class instance, typed array, Blob), so it can only ever bias toward a harmless spurious write, never a false positive that drops an edit. Recursion is depth-bounded so cyclic/deep content can't overflow. Your `Map(k→1)`→`Map(k→2)` repro is now a test (`re-saves a scene whose opaque structured-clone content changed`), and it fails on the old code / passes on the new.

**2. `listDocuments` version tolerance.** I went with the "explicitly document it" option rather than the guard, and I'd value your read on it: `listDocuments` returns only version-independent fields (id/name/timestamps/sceneCount) and never migrates or reads content, so applying the fail-loud guard would make a single corrupt row break the entire listing (an unusable picker), whereas a broken document already surfaces fail-loud the moment it's actually opened. I documented the intentional tolerance on the `DocumentStore.listDocuments` contract + the impl, and pinned it with a test (`listDocuments tolerates a malformed version stamp`). Happy to switch to the guard (or to skipping/flagging corrupt rows) if you'd rather it not surface them at all.

All tests green + typecheck/prettier/eslint clean; the fix was cross-checked for false-positive equality paths.

cosarah (commented):
`packages/@openmaic/storage/src/document/browser.ts:135`

`structuralEquals` fixes the old `JSON.stringify` bug, but the `Map` / `Set` branches still treat those types as unordered membership checks. `saveDocument` uses this comparator to decide whether it can skip rewriting a scene, so an order-only edit like `new Map([['a', 1], ['b', 2]])` -> `new Map([['b', 2], ['a', 1]])` (same issue for `Set`) is still considered equal and the old order stays on disk. I reproduced this locally. Since insertion order is part of both types' observable semantics and IndexedDB preserves it, this still violates the “opaque app content” contract. These branches should either compare iteration order too, or conservatively return `false`.

wyuc:
Good catch again — and it pushed me to reconsider the whole approach rather than keep patching. Fixed in 51b92b14.

I first made `structuralEquals` order-sensitive for `Map`/`Set`/object keys, but red-teaming it surfaced that a hand-rolled deep-equal over opaque scene content can't be made reliably correct: faithfully matching what IndexedDB persists means reproducing structured-clone equality across iteration order, sparse-array holes vs explicit `undefined`, array expando properties, null vs `Object` prototype, and shared-reference topology — each a false-positive where a wrong "equal" silently drops an edit.

So I took your other suggestion from the first pass: **drop the per-scene skip.** `saveDocument` now writes every incoming scene and deletes the ones no longer present — no content comparison at all. A redundant write is harmless, and the cheap per-scene incremental path is `putScene` (which #779's normalization is really there to enable). The tests that a real edit is persisted (value change, and Map/Set/object-key reorder) are kept as regression guards. Net: the comparator is gone (~55 fewer lines) and the class of bug is closed.

All tests green + typecheck/prettier/eslint clean.

cosarah (approved):
LGTM

Issue #854, opened 2026-07-05
[Task] Hyperframes-based classroom video export architecture


wyuc:
## Context

#290 tracks the user-facing request to export generated classrooms as video, with narration/audio and eventually subtitles. Several related issues and PRs now touch the same area, so we should consolidate the architecture before more parallel implementations diverge.

Related product requests and architecture tasks:

- #290 — main request: export generated classrooms as video/audio/subtitles.
- #203 — video and external media embedding; this affects video export fidelity for embedded/generated videos.
- #204 — multi-language generation and subtitles; this affects caption/subtitle output.
- #539 / #540 — generated video manifest refs and `mediaRef` handling; this is the media reference foundation video export should reuse.
- #818 / #836 / #843 — playback seek/resume/action-level navigation; relevant for later mapping classroom actions onto a video timeline.
- #821 — narration/script editing and TTS regeneration; relevant for exportable narration quality.
- #852 — `@openmaic/exporter` extraction for PPTX / HTML / offline ZIP. Video export should align with this direction, but it may need a separate phase/backend because rendering MP4 has heavier runtime requirements.

Active implementation PRs:

- #746 — direct MP4 export prototype using rendered scene PNGs, narration audio, and server-side FFmpeg composition.
- #849 — video-frame export foundation: ordered PNG frames, scene JSON, cached audio/media, HTML sidecars, and a manifest.

## Proposed maintainer direction

Prefer Hyperframes as the primary rendering backend for classroom video export.

Hyperframes is an HTML/CSS/media/seekable-animation to MP4 pipeline built around deterministic headless-Chrome frame capture plus FFmpeg encoding. That model fits OpenMAIC better than growing a custom static-frame + FFmpeg implementation because OpenMAIC classrooms already contain slide DSL, HTML interactive scenes, generated media, narration audio, and action timing.

References:

- https://github.com/heygen-com/hyperframes
- https://hyperframes.heygen.com/introduction
- https://hyperframes.heygen.com/packages/producer

The intended source of truth should remain OpenMAIC classroom data: stage, scenes, slide DSL, actions, audio/media records, generated-agent metadata, and media refs. Hyperframes should be the render backend, not a replacement for the OpenMAIC classroom model.

## Proposed architecture

Add a compiler layer:

```text
OpenMAIC classroom data
  -> OpenMAIC video export compiler
  -> Hyperframes composition/project
  -> Hyperframes render worker / CLI / producer
  -> MP4/WebM output
```

[...]

PR #862, opened 2026-07-07
feat(audio): persist audio duration when storing TTS (#861)

### LINKED ISSUE #861: [Task] Persist audio duration when storing TTS (prerequisite for video export)
## Context

Sub-issue / prerequisite of #854 (Hyperframes-based classroom video export architecture).

The video-export compiler needs to map each narration/TTS audio clip onto a timeline segment (scene/action → timeline segment). That mapping depends on the **actual duration** of each audio clip. Today, duration is not written when TTS audio is persisted:

- `AudioFileRecord.duration?` is already declared in the IndexedDB schema (`lib/utils/database.ts:88`) but is never populated.
- The client persistence path `generateAndStoreTTS` (`lib/hooks/use-scene-generator.ts:292`) only writes `id / blob / format / createdAt` — no `duration`.
- The server persistence path (`lib/server/classroom-media-generation.ts:281`) writes the audio file and stamps `audioUrl / audioId` on the speech action, also without duration.

Without a stored duration, the export compiler would have to decode every audio clip at render time to measure its length — slow, and impossible to validate as a pure manifest in tests without FFmpeg/Chrome (which is exactly what #854's acceptance criteria call for: "the artifact is inspectable in tests without requiring FFmpeg/Chrome").

## Goal

Compute and persist audio duration (in seconds) at TTS persistence time, so `AudioFileRecord.duration` becomes a reliable source of truth that the later video-export manifest can consume directly.

## Scope

1. **Client path**: before persisting in `generateAndStoreTTS`, decode the generated audio blob to obtain its duration and write it into `duration`.
2. **Server path**: in `classroom-media-generation.ts`, compute the audio duration when writing/stamping, and attach it to the `SpeechAction` (e.g. a new optional `audioDuration` field) or emit it alongside an audio manifest for the export compiler to read.
3. Keep the unit consistent (seconds) and tolerate decode failures (log a warning and leave `duration` empty rather than aborting persistence).
4. Add unit tests: assert persisted records carry a duration of the correct magnitude; assert graceful degradation (no throw) on decode failure.

## Non-goals

- The video-export compiler itself, Hyperframes composition, or MP4 rendering (later phases of #854).
- Fine-grained subtitle / timeline alignment logic.

## Acceptance criteria

- Newly generated TTS audio has a correct duration field in its persisted record (client `AudioFileRecord` and/or server speech action / manifest).
- Decode failure does not break normal audio persistence or playback.
- Unit tests cover both the duration-write path and the degradation path.

cosarah:
## Summary

Persist audio duration when the client stores TTS audio, as a prerequisite for #854 (Hyperframes-based classroom video export). Closes #861.

The video-export compiler needs each narration clip's duration to map it onto a timeline segment. `AudioFileRecord.duration` already exists in the IndexedDB schema but the TTS store path never populated it — so export/import (which round-trip duration through the media manifest) had nothing to carry, and the compiler would have to re-decode every clip at render time (slow, and impossible to validate in a pure-manifest test without FFmpeg/Chrome).

## Design: duration is an asset property

Duration belongs to the audio **file**, not the speech **action** — re-generating a line (new voice/speed) keeps the text but changes the duration, and it travels with the blob. This matches the existing export/import convention:

- Export reads `af.record.duration` into `mediaIndex[].duration` (`use-export-classroom.ts:160`).
- Import writes `meta.duration` back into `AudioFileRecord.duration` (`use-import-classroom.ts:129`).

So this PR populates `AudioFileRecord.duration` and does **not** add a field to the `@openmaic/dsl` `SpeechAction` contract.

## Scope

Client persistence path only. The server generation path serializes scenes to JSON with no `AudioFileRecord` and no media index, so it has no asset-side home for duration; that's left to a later slice (e.g. when the server emits a media manifest, or at #854's compiler stage).

## Changes

- **`lib/audio/audio-duration.ts`** (new): dependency-free `measureAudioDuration` — a RIFF chunk walk for WAV and a Xing/Info-or-CBR estimate for MP3 (the two formats OpenMAIC's TTS providers emit). Pure byte parsing, no DOM, returns `null` for anything it can't parse so callers degrade gracefully.
- **`lib/hooks/use-scene-generator.ts`** (`generateAndStoreTTS`): populate `AudioFileRecord.duration` when storing to IndexedDB.
- **`tests/audio/audio-duration.test.ts`**: WAV/MP3 parsing, word-alignment/ID3 edge cases, format dispatch + sniffing, graceful degradation on empty/unsupported input.

## Degradation

Decode failure never blocks persistence: `measureAudioDuration` returns `null`, `duration` is left undefined, and the audio is still stored and playable.

## Testing

- `pnpm vitest run tests/audio/audio-duration.test.ts` — 11 passing.
- `pnpm exec tsc --noEmit` — clean.

### DIFF
--- a/lib/audio/audio-duration.ts
+++ b/lib/audio/audio-duration.ts
@@ -1,0 +1,200 @@
+/**
+ * Dependency-free audio duration measurement.
+ *
+ * Video export (#854) maps each narration/TTS clip onto a timeline segment,
+ * which needs the clip's duration. Rather than decode audio at render time
+ * (slow, and impossible to validate in a pure-manifest test without
+ * FFmpeg/Chrome), we measure duration once when the client persists TTS audio
+ * and store it on the `AudioFileRecord` alongside the blob.
+ *
+ * This module parses the container headers directly (no DOM / native audio
+ * API), so it's unit-testable in plain Node and stays usable from either the
+ * browser or a server context. It covers the two formats OpenMAIC's TTS
+ * providers actually emit — WAV and MP3 — and returns `null` for anything it
+ * cannot parse so callers degrade gracefully (store the audio, leave duration
+ * undefined) instead of failing.
+ */
+
+/** Coerce common inputs into a Uint8Array view without copying when possible. */
+function toBytes(input: Uint8Array | ArrayBuffer): Uint8Array {
+  return input instanceof Uint8Array ? input : new Uint8Array(input);
+}
+
+function readAscii(bytes: Uint8Array, offset: number, length: number): string {
+  let out = '';
+  for (let i = 0; i < length; i++) out += String.fromCharCode(bytes[offset + i] ?? 0);
+  return out;
+}
+
+function readUint32LE(bytes: Uint8Array, offset: number): number {
+  return (
+    (bytes[offset] |
+      (bytes[offset + 1] << 8) |
+      (bytes[offset + 2] << 16) |
+      (bytes[offset + 3] << 24)) >>>
+    0
+  );
+}
+
+function readUint32BE(bytes: Uint8Array, offset: number): number {
+  return (
+    ((bytes[offset] << 24) |
+      (bytes[offset + 1] << 16) |
+      (bytes[offset + 2] << 8) |
+      bytes[offset + 3]) >>>
+    0
+  );
+}
+
+/**
+ * WAV: walk the RIFF chunk list to find `fmt ` (byte rate) and `data` (size).
+ * duration = dataSize / byteRate. Walking chunks — rather than assuming fixed
+ * offsets — tolerates optional chunks (LIST, fact, …) some encoders insert
+ * before `data`.
+ */
+export function measureWavDuration(input: Uint8Array | ArrayBuffer): number | null {
+  const bytes = toBytes(input);
+  if (bytes.length < 12) return null;
+  if (readAscii(bytes, 0, 4) !== 'RIFF' || readAscii(bytes, 8, 4) !== 'WAVE') return null;
+
+  let byteRate = 0;
+  let dataSize = 0;
+  let offset = 12;
+  while (offset + 8 <= bytes.length) {
+    const chunkId = readAscii(bytes, offset, 4);
+    const chunkSize = readUint32LE(bytes, offset + 4);
+    const bodyOffset = offset + 8;
+    if (chunkId === 'fmt ' && bodyOffset + 16 <= bytes.length) {
+      byteRate = readUint32LE(bytes, bodyOffset + 8);
+    } else if (chunkId === 'data') {
+      // Some streams write 0 / 0xFFFFFFFF for an unknown-length data chunk;
+      // fall back to the actual remaining bytes in that case.
+      const declared = chunkSize;
+      const remaining = bytes.length - bodyOffset;
+      dataSize = declared > 0 && declared <= remaining ? declared : remaining;
+      break;
+    }
+    // Chunks are word-aligned: an odd size is padded with one byte.
+    offset = bodyOffset + chunkSize + (chunkSize % 2);
+  }
+
+  if (byteRate <= 0 || dataSize <= 0) return null;
+  return dataSize / byteRate;
+}
+
+// MPEG audio version → layer → bitrate table (kbps). Index by version bits then
+// layer bits, then the 4-bit bitrate index from the frame header.
+const MP3_BITRATES: Record<string, number[]> = {
+  // MPEG 1
+  '1-1': [0, 32, 64, 96, 128, 160, 192, 224, 256, 288, 320, 352, 384, 416, 448],
+  '1-2': [0, 32, 48, 56, 64, 80, 96, 112, 128, 160, 192, 224, 256, 320, 384],
+  '1-3': [0, 32, 40, 48, 56, 64, 80, 96, 112, 128, 160, 192, 224, 256, 320],
+  // MPEG 2 / 2.5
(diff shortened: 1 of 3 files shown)


wyuc:
## Cross-review result

Reviewed by two independent report-only paths against the same PR diff. No files were modified during review.

### Summary

No P1/P2 blocker was independently reported by both reviewers.

There is one accepted P2 robustness issue around trusting the reported audio format before sniffing bytes. The remaining findings are P3 accuracy/test-coverage gaps around MP3 duration parsing.

### Findings

#### [P2] Supported-but-wrong format hints bypass content sniffing

`lib/audio/audio-duration.ts:191`

`measureAudioDuration` immediately dispatches on a supported `format` hint. If the provider reports `mp3` while the bytes are actually WAV, the function never sniffs the RIFF header and will either return `null` or, in a worse false-sync case, an incorrect MP3 duration.

This can happen because `getAudioResponseFormat` falls back to `mp3` when `Content-Type` is missing or unrecognized (`lib/audio/tts-providers.ts:357-363`), while some providers/request modes can still be configured or defaulted to produce WAV. For example, Lemonade requests `response_format: config.format || 'wav'` but then trusts response `Content-Type` to report the format.

Recommendation: in `measureAudioDuration`, sniff the magic bytes first when the hint conflicts with the content, or retry sniffing when the hinted parser returns `null`. Add a regression test where `format='mp3'` but bytes begin with a valid RIFF/WAVE header.

#### [P3] VBRI MP3s are not implemented despite the comment

`lib/audio/audio-duration.ts:116-174`

The comment says the MP3 parser prefers `Xing/Info (or VBRI)`, but the implementation only checks `Xing`/`Info`. VBRI-only MP3s fall through to the CBR byte-count estimate using the first frame bitrate, which can produce a plausible but wrong duration.

Recommendation: either implement VBRI frame-count parsing, or remove the claim and return `null` when the file appears to be unsupported VBR rather than estimating as CBR.

#### [P3] Xing/VBR path lacks direct tests

`tests/audio/audio-duration.test.ts`

The new tests cover WAV, CBR MP3, ID3v2 skip, sniffing, and unsupported input, but do not cover the Xing/Info frame-count path. A bug in the version/channel offset calculation would silently persist wrong durations for VBR MP3.

Recommendation: add a minimal MP3 test vector with a Xing or Info header, `flags=0x01`, and a known frame count.

#### [P3] CBR estimate includes trailing non-audio metadata

`lib/audio/audio-duration.ts:172-174`

The CBR fallback uses `bytes.length - frame`, which counts trailing ID3v1/APEv2 tags or appended non-audio data. This is usually a small overestimate but can be larger with unusual trailers.

Recommendation: ignore known trailing tags before calculating CBR duration, or add a test documenting the accepted tolerance.

#### [P3] Non-WAV/MP3 TTS formats still have unknown duration

`lib/audio/audio-duration.ts:190-199`

The PR intentionally returns `null` for unsupported formats. That is graceful, but provider settings expose formats such as `opus`, `pcm`, `flac`, `webm`, and `ogg`; those generated clips will persist without duration and may still be problematic for the video-export timeline.

Recommendation: for the video export epic, decide whether non-WAV/MP3 TTS formats are unsupported for export, should be converted, or need additional duration parsers.

### Verification

- GitHub checks were green at review time: Lint/Typecheck/Unit Tests, CodeQL, and E2E.
- Local tests were not rerun because the disposable review worktree did not have dependencies installed.

cosarah:
## Addressed review feedback (`6e3ff911`)

### Fixed in this PR

**[P2 — accepted] Format hint bypassed content sniffing**
`measureAudioDuration` used to dispatch on the reported `format` first. Since `getAudioResponseFormat` falls back to `mp3` when `Content-Type` is missing/unrecognized, a WAV clip could reach the MP3 parser and false-sync into a wrong duration. It now **sniffs the magic bytes first and only falls back to the `format` hint when the content is unrecognizable** — content always wins over a mismatched hint, so WAV bytes never go through the MP3 parser (or vice versa).

**[P3] VBRI not implemented despite the comment**
Implemented VBRI frame-count parsing (header at `frame + 4 + 32`, frame count at byte 14). The parser comment's "Xing/Info (or VBRI)" claim is now accurate.

**[P3] CBR estimate included trailing non-audio metadata**
The CBR fallback now excludes a trailing ID3v1 tag (last 128 bytes, `TAG` magic) before computing `audioBytes`, so the tag can't inflate the estimate.

**[P3] Xing/VBR path lacked direct tests**
Added regression tests: mismatched hint (both directions), Xing frame count, Info (CBR) header frame count, VBRI frame count, and trailing ID3v1. **17 passing (was 11)**; `tsc --noEmit` clean.

### Deliberately out of scope

**[P3] Non-WAV/MP3 TTS formats still have unknown duration** — Formats like `opus`, `pcm`, `flac`, `webm`, `ogg` continue to return `null` (audio still persists, duration left undefined). As the review notes, whether these are unsupported-for-export / should be converted / need extra parsers is a decision for the video-export epic (#854), not this client-persistence slice.

wyuc (approved):
Re-reviewed the latest head `6e3ff911`.

The previous review feedback is addressed:

- `measureAudioDuration` now sniffs recognized WAV/MP3 content before trusting the reported format hint, so a missing/incorrect `Content-Type` no longer routes WAV bytes through the MP3 parser.
- VBRI frame-count parsing is implemented.
- CBR fallback excludes trailing ID3v1 tags.
- Regression coverage was added for mismatched hints, Xing/Info frame counts, VBRI frame counts, and trailing ID3v1.

I also rechecked the integration point in `generateAndStoreTTS`; the remaining non-WAV/MP3 formats are intentionally left as video-export epic scope. GitHub CI is green. I did not rerun local Vitest because the disposable review worktree has no installed dependencies.

Issue #869, opened 2026-07-08
[RFC] @openmaic/storage: DSL-native runtime layer — record envelope + RuntimeStore (#779)


wyuc:
## Goal

The storage RFC (#779) defers `RuntimeStore` until a DSL-native runtime type exists. This proposal defines that contract: a common **runtime record envelope** owned by `@openmaic/dsl`, plus the `RuntimeStore` component that persists it — completing the fourth quadrant of the #779 layer model (document / **runtime** / asset / kv).

"Runtime" here means what a learner produces while taking a course: conversations, quiz attempts, project progress, playback position. It does not travel with the document, but it must survive reloads, be exportable as a replay, and — in any deployment where more than one learner opens the same course — be isolated per learner.

## Where runtime data actually lives today (inventory)

Four families, three unrelated storage mechanisms, no shared contract:

| Data | Storage today | Anchoring | Learner dimension | Timestamps |
|---|---|---|---|---|
| Chat sessions | Dexie `chatSessions` (`lib/utils/database.ts`) | `sceneId`, `lastActionIndex` | none | epoch ms |
| Quiz attempts | `localStorage` `quizDraft:` / `quizAnswers:` / `quizResults:<sceneId>` (`lib/quiz/persistence.ts`) | `sceneId` | none | **none** |
| PBL v2 learner state | **inside the document** — `scene.content.projectV2` (`lib/pbl/v2/types.ts`) | scene content | none | ISO strings |
| Playback progress | Dexie `playbackState` (`lib/utils/playback-storage.ts`) — `sceneIndex` / `actionIndex` / `consumedDiscussions` | `sceneId` guard | none | epoch ms |

Problems this causes, all observable in the browser-only app today:

1. **No learner dimension anywhere.** Two learners sharing a browser overwrite each other's quiz answers and playback position. A server-backed deployment cannot even express "this learner's session on that course" — the primary key is missing.
2. **PBL runtime state is embedded in the document.** `PBLProjectV2` mixes design-time fields (scenario, milestone briefings, completion criteria) with learner state (milestone/microtask `status`, `submissions`, `evaluations`, `engagementEvents`, `proficiencyAssessment`, `pendingHandover`). Every learner interaction becomes a *document save*; `mergePreservingPBL` in `lib/store/stage.ts` exists purely to stop an ordinary content save from wiping learner progress. The `runtimeEvents?` field comment already anticipates a "runtime-split" — this RFC is that split.
[...]

PR #870, opened 2026-07-08
feat(dsl): runtime envelope — RuntimeSession + RuntimeRecord + core-kind skeletons (#869 Part A)

wyuc:
Part A of #869: the DSL-native runtime layer contract — pure types, guards, and validators in `@openmaic/dsl`, no storage code (the `RuntimeStore` browser backend + contract suite is Part B). Mirrors how #858/#860 delivered the document side.

## What

- **`RuntimeSession`** — the unit of learner-runtime identity and lifecycle: owns `(stageId, learnerKey, kind)` and the `active → completed → archived` status ladder. Extends `RuntimeVersioned`: sessions stamp their own `runtimeDslVersion` envelope field and ride a dedicated runtime version line (`RUNTIME_DSL_VERSION` + `RUNTIME_DSL_MIGRATIONS` + `migrateRuntime`), mechanically disjoint from the document ladder (proved by test).
- **`RuntimeRecord<TPayload>`** — one ordered fact inside a session. The store-assigned monotonic `seq` is the sole replay ordering key (never client timestamps); `sceneId`/`actionIndex`/`subAnchor` are documented best-effort anchors into the editable document timeline. Payload internals are app-owned via the generic, exactly like `DocumentStore<TScene>` (#860).
- **`RuntimeRecordInit`** — the creation shape a producer hands to `append`: structurally omits `seq`, since only the store may assign it.
- **Core-kind skeletons** for `chat` (roles + content) and `quizAttempt` (phase + answers map), with guards. `playback` is a recognized kind name with a deliberately app-owned payload.
- **Envelope validators** `validateRuntimeSession` / `validateRuntimeRecord` — the store write gate: well-formed `x.y.z` version stamps, ISO-8601 timestamps validated by pure calendar arithmetic (same rule `migrate` enforces, via the newly exported `isWellFormedDslVersion`), non-negative-integer `seq`/`actionIndex`, non-empty required identity strings, payload presence.

## Design decisions refining the RFC text

- Sessions are first-class and records reference them by `sessionId` — identity/lifecycle lives once on the session instead of being duplicated per record, resolving the session/record granularity #869 left open.
- Replay ordering moved from timestamps to store-assigned `seq` (multi-tab and clock-skew safe); timestamps remain display metadata, ISO 8601 by design (divergence from the document aggregate's epoch-ms is called out in the module docs).
- Compile-time exhaustiveness links between each union type and its `as const` array, following the `SCENE_TYPES` house pattern.

## Tests

153 passing (`pnpm --filter @openmaic/dsl test`); typecheck, build, prettier, eslint all clean.

> [!NOTE]
> Scope grew slightly during review: a dedicated runtime version line with its own `runtimeDslVersion` envelope field (review thread on migration-ladder coupling), pure component-level ISO-8601 calendar validation (review thread on `Date.parse` normalization), and the `@openmaic/dsl` test suite now runs in PR CI.

### DIFF
--- a/packages/@openmaic/dsl/src/runtime.ts
+++ b/packages/@openmaic/dsl/src/runtime.ts
@@ -30,12 +30,18 @@
  * Versioning: a session carries its OWN version field, `runtimeDslVersion`
  * (distinct from a document's `dslVersion`), and rides its OWN version line —
  * `RUNTIME_DSL_VERSION` + `migrateRuntime`, not the document's `DSL_VERSION` +
- * `migrate`. The two lines stamp different envelope fields, so neither ladder
- * reads the other's version — but disjoint fields alone would still let a
- * misrouted session be lifted (as unversioned) on the document line. The
- * cross-line guard in `runLadder` closes this: a runner throws on any
- * aggregate that carries the sibling line's stamp but not its own, surfacing
- * the misroute instead of guessing — and vice versa (see `version.ts`).
+ * `migrate`. The stamp is **required** on a session: sessions are born stamped
+ * at write time and the runtime line has **no unversioned epoch** (nothing
+ * legitimately predates this brand-new envelope, unlike real pre-versioning
+ * documents). So an unstamped object reaching the runtime line is a bug — a
+ * misrouted legacy document or an unstamped producer write — and fails loud
+ * (`noRuntimeEpochError`) rather than being lifted like a legacy document. The
+ * two lines stamp different envelope fields, so neither ladder reads the other's
+ * version — but disjoint fields alone would still let a misrouted session be
+ * lifted (as unversioned) on the document line. The cross-line guard in
+ * `runLadder` closes this: a runner throws on any aggregate that carries the
+ * sibling line's stamp but not its own, surfacing the misroute instead of
+ * guessing — and vice versa (see `version.ts`).
  *
  * No runtime dependencies. Pure types + plain data constants only.
  */
@@ -177,20 +183,31 @@ export function isCoreRuntimeKind(value: unknown): value is CoreRuntimeKind {
  * `(stageId, learnerKey)` plus a `kind`; a learner may hold several sessions
  * of the same kind on one stage (e.g. repeated quiz attempts).
  *
- * Extends {@link RuntimeVersioned}, so a session carries an optional
- * `runtimeDslVersion` serialized-contract stamp (absent on legacy data) — a
- * DIFFERENT envelope field from a document's `dslVersion`. Stamping +
- * migrate-on-read run on the runtime line only: a session is stamped with
- * `RUNTIME_DSL_VERSION` and migrated by `migrateRuntime`, independent of the
- * document's `DSL_VERSION` / `migrate`. The two stamps live on distinct fields
- * so neither ladder reads the other's; the cross-line guard in `runLadder`
- * turns a misrouted migration into a loud error rather than corruption — a
- * runner throws on any aggregate stamped on the sibling line but not its own,
- * and `validateRuntimeSession` rejects a stray `dslVersion` at the door
+ * Extends {@link RuntimeVersioned} but **overrides its stamp to required**: a
+ * session carries a `runtimeDslVersion` serialized-contract stamp — a DIFFERENT
+ * envelope field from a document's `dslVersion`. Sessions are **born stamped**:
+ * producers write `RUNTIME_DSL_VERSION` at creation and there is no unversioned
+ * epoch on the runtime line (nothing legitimately predates the envelope), so a
+ * stored session always carries the stamp; an absent one is a transient
+ * in-memory state at most, never valid stored data. Stamping + migrate-on-read
+ * run on the runtime line only: a session is migrated by `migrateRuntime`,
+ * independent of the document's `DSL_VERSION` / `migrate`. The two stamps live on
+ * distinct fields so neither ladder reads the other's; the cross-line guard in
+ * `runLadder` turns a misrouted migration into a loud error rather than
+ * corruption — a runner throws on any aggregate stamped on the sibling line but
+ * not its own — while a wholly-unstamped object hitting the runtime line throws
+ * `noRuntimeEpochError` (no legacy lift), and `validateRuntimeSession` rejects
+ * both a missing `runtimeDslVersion` and a stray `dslVersion` at the door
  * (see `version.ts`).
  */
 export interface RuntimeSession extends RuntimeVersioned {
   id: string;
+  /**
+   * Runtime-contract version this session was written at. Required: sessions are
+   * born stamped with `RUNTIME_DSL_VERSION`; the runtime line has no unversioned
+   * epoch, so this narrows {@link RuntimeVersioned}'s optional field to mandatory.
+   */
+  runtimeDslVersion: string;
   /** {@link CoreRuntimeKind} or an app-defined kind. */
   kind: string;
   stageId: string;
--- a/packages/@openmaic/dsl/src/runtime.ts
+++ b/packages/@openmaic/dsl/src/runtime.ts
@@ -79,8 +79,20 @@ export function isRuntimeSessionStatus(value: unknown): value is RuntimeSessionS
  * zone designator (`Z` or `±hh:mm`). Runtime timestamps are display metadata
  * whose only cross-tab guarantee is a comparable, unambiguous instant, so the
  * zone is not optional here — a zoneless string names no instant.
+ *
+ * Capture groups feed the component-level range check in {@link isIsoTimestamp}:
+ * 1 year, 2 month, 3 day, 4 hour, 5 minute, 6 second, 7 optional offset hours,
+ * 8 optional offset minutes (groups 7/8 are undefined for the `Z` form).
  */
-const ISO_TIMESTAMP_RE = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d+)?(Z|[+-]\d{2}:\d{2})$/;
+const ISO_TIMESTAMP_RE =
+  /^(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})(?:\.\d+)?(?:Z|[+-](\d{2}):(\d{2}))$/;
+
+/** Days in a Gregorian month, applying the full leap-year rule for February. */
+function daysInMonth(year: number, month: number): number {
+  const isLeap = year % 4 === 0 && (year % 100 !== 0 || year % 400 === 0);
+  const lengths = [31, isLeap ? 29 : 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
+  return lengths[month - 1];
+}
 
 /**
  * True when `value` is a well-formed ISO-8601 timestamp per the runtime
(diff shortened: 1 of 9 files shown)


wyuc:
**Cross-model review summary** (pre-submission, two independent reviewers over the full branch diff: Claude Code multi-angle review + Codex `exec review` at high effort, findings merged by consensus):

- **Round 1** surfaced 11 actionable findings. The one both reviewers hit independently (P2): `validateRuntimeSession` accepted any string as a `dslVersion` stamp while `migrate()`/`dslVersionOf()` throw on malformed stamps — the validator now enforces the same well-formed `x.y.z` rule via the newly exported `isWellFormedDslVersion`. The rest of the round: compile-time exhaustiveness links for all four union↔array pairs (the `SCENE_TYPES` house pattern), `RuntimeSession extends DslVersioned` instead of re-declaring the stamp field, a shared `checkFields` helper replacing three copy-pasted validator loops, `seq`/`actionIndex` narrowed to non-negative integers (NaN/Infinity/fractional passed a bare `typeof` check), required identity strings must be non-empty (an empty `learnerKey` would merge every broken client into one partition), `payload: undefined` no longer passes the presence check, `RuntimeRecordInit` added as the store-append creation shape, skeleton guards (`isChatMessageSkeleton`/`isQuizAttemptSkeleton`), and doc corrections (`playback` has no DSL-owned skeleton by design; ISO-8601 timestamps are a deliberate divergence from the document aggregate's epoch-ms). Fixed in caded6aa.
- **Round 2**: re-review caught a P2 introduced by round 1 itself — `isQuizAttemptSkeleton` accepted `Map`/`Date`/class instances for `answers`; now requires a plain record (prototype `Object.prototype` or `null`). Fixed in 24289e9d.
- **Final verification pass**: no actionable findings. 153 tests passing; typecheck, build, prettier, eslint clean; CI green.

Known deferrals, called out for reviewers: the `kind`↔payload type binding intentionally lives in Part B's typed accessors (per #869 open question 1) rather than as a discriminant on the record envelope; JSON-schema emission for the runtime envelope is deferred alongside the HTTP backend.

packages/@openmaic/dsl/src/validate.ts
```diff
@@ -206,3 +251,97 @@ export function validateAction(doc: unknown): ValidationResult {
...
+/**
+ * Validate a runtime session envelope (#869). Payloads live on records, so
+ * this is a pure envelope check; `kind` is an open string by design.
+ */
+export function validateRuntimeSession(doc: unknown): ValidationResult {
+  const errors: ValidationIssue[] = [];
+  if (!isObject(doc)) {
+    return { valid: false, errors: [{ path: '/', message: 'runtime session must be an object' }] };
+  }
+  checkFields(doc, RUNTIME_SESSION_REQUIRED_FIELDS, {}, errors);
```
cosarah, line 275: This currently treats `createdAt` / `updatedAt` as "any string", so `validateRuntimeSession({ ..., createdAt: 'not-a-date', updatedAt: 'still-bad' })` still returns `valid: true`. The contract docs in `runtime.ts` call these ISO 8601 timestamps, and unlike stage/scene/action there is no generated schema artifact for the runtime envelope to catch the format elsewhere. I think this needs an explicit ISO-8601 check here, and the same check for `RuntimeRecord.createdAt` below.

packages/@openmaic/dsl/src/runtime.ts
```diff
@@ -0,0 +1,266 @@
...
+ * The unit of learner-runtime identity and lifecycle. Sessions are keyed by
+ * `(stageId, learnerKey)` plus a `kind`; a learner may hold several sessions
+ * of the same kind on one stage (e.g. repeated quiz attempts).
+ *
+ * Extends {@link DslVersioned}, so a session carries the same optional
+ * `dslVersion` serialized-contract stamp (absent on legacy data) and rides the
+ * same stamping + migrate-on-read ladder as the document aggregate — sharing
+ * one envelope definition instead of re-declaring the field here.
+ */
+export interface RuntimeSession extends DslVersioned {
```
cosarah: Extending `DslVersioned` couples runtime sessions to the document migration ladder, but `DslMigration` is still defined as an arbitrary whole-document transform. That is safe today only because the sole migration step is a no-op stamp. Once we add a real `Stage`/`Scene` migration, `migrate(session)` will run through code that was authored against a different aggregate shape and can either mutate runtime envelopes incorrectly or throw. I think runtime data needs its own versioning path, or `migrate()` needs to be narrowed so future document migrations cannot accidentally consume `RuntimeSession`.

packages/@openmaic/dsl/src/runtime.ts
```diff
@@ -0,0 +1,266 @@
...
+  /** ISO 8601. */
+  updatedAt: string;
+}
+
+/**
+ * One ordered fact inside a session. Identity, learner and lifecycle live on
+ * the parent {@link RuntimeSession}; the record carries only ordering,
+ * anchoring, and the app-owned payload.
+ */
+export interface RuntimeRecord<TPayload = unknown> {
```
cosarah: There is a type/runtime mismatch here: with the default `TPayload = unknown`, `const record: RuntimeRecord = { ..., payload: undefined }` type-checks, but `validateRuntimeRecord()` rejects the same value. That makes the static contract looser than the runtime contract and pushes the failure to append-time. Could we exclude `undefined` from `payload` at the type level, or otherwise align the interface with the validator?

cosarah (commented):
I left a few inline comments on correctness and contract boundaries.

packages/@openmaic/dsl/src/validate.ts
wyuc: Fixed in 30be1725. Added a pure `isIsoTimestamp` guard in `runtime.ts` (regex for the ISO-8601 shape *plus* `Date.parse` for calendar validity — the regex alone accepts impossible dates like `2026-13-01`, and `Date.parse` alone accepts many non-ISO forms, so both halves are required) and wired it into `validateRuntimeSession` (`createdAt`/`updatedAt`) and `validateRuntimeRecord` (`createdAt`). `'not-a-date'` and month-13 now fail; offset forms like `+08:00` pass. Agreed this is the format enforcement point given the runtime envelope has no generated schema artifact.

packages/@openmaic/dsl/src/runtime.ts
wyuc: Fixed in 30be1725 by giving runtime data its own version line. `version.ts` now ships `RUNTIME_DSL_VERSION`, a dedicated `RUNTIME_DSL_MIGRATIONS` ladder, and `migrateRuntime`, all walked by a shared `runLadder` runner parameterized on ladder + target version (so `migrate` stays byte-identical — the existing document-migration tests pass unmodified). A `RuntimeSession` still shares the `dslVersion` envelope *field* (`extends DslVersioned`), but is stamped/migrated on the runtime line only, so a future `Stage`/`Scene` migration can no longer run over — or accidentally consume — a runtime envelope.

packages/@openmaic/dsl/src/runtime.ts
wyuc: Fixed in 30be1725. Added `type RuntimePayload = NonNullable<unknown> | null` (any value except `undefined`; `null` remains a legal stored payload, matching the validator) and constrained `RuntimeRecord<TPayload extends RuntimePayload>` and `RuntimeRecordInit` with it. `payload: undefined` is now a compile error, aligning the static type with `validateRuntimeRecord`; a `@ts-expect-error` test pins the exclusion.

packages/@openmaic/dsl/src/validate.ts
cosarah: One remaining hole here: the `Date.parse` half does **not** reject all calendar-impossible dates. In JS, strings like `2026-02-30T00:00:00.000Z`, `2026-02-29T00:00:00.000Z`, and `2026-04-31T00:00:00.000Z` are normalized to real instants (March 2 / March 1 / May 1) instead of returning `NaN`, so `isIsoTimestamp()` still accepts some invalid Gregorian dates while the docstring says impossible dates are rejected. I think this needs a stricter component-level check (or a parse-and-round-trip check against the original Y/M/D/H/M/S fields), plus a regression test for a non-leap-year `02-29` / `02-30` case.

wyuc: Good catch — confirmed the normalization hole. On V8 a full ISO datetime like `2026-02-30T00:00:00.000Z` (and `2026-01-01T24:00:00.000Z`) doesn't return NaN; the engine rolls it into a neighbouring real instant, so a `Date.parse` gate can't pin calendar validity, and the verdict is engine-dependent besides. Fixed in 09e8246f by dropping `Date` entirely: the regex now captures year/month/day/hour/minute/second and the numeric offset, and a pure arithmetic check validates month 1–12, day-in-month under the full Gregorian leap rule (÷4, except centuries unless ÷400), hour ≤ 23, minute/second ≤ 59 (leap-second `:60` deliberately not accepted), and offset hours ≤ 23 / minutes ≤ 59. Regression tests cover Feb 30, non-leap Feb 29, leap Feb 29, Apr 31, hour 24, offset +25:00, and century non-leap 2100-02-29.

packages/@openmaic/dsl/src/runtime.ts
wyuc: Follow-up hardening in 5cf37db2: the separation is now mechanical rather than a call-site convention. Runtime sessions stamp their own envelope field `runtimeDslVersion` (`RuntimeVersioned`, with `runtimeDslVersionOf`/`needsRuntimeMigration` counterparts), disjoint from the document line's `dslVersion`. A document migration walked over a session now finds no `dslVersion` stamp at all — misrouted data is inert instead of silently reinterpreted, in either direction. Same commit also wires the @openmaic/dsl test suite into PR CI (it previously ran only on publish).

cosarah: I do not think the separate key fully delivers the safety claim here. `migrate()` still treats a `RuntimeSession` as an *unversioned document* whenever `dslVersion` is absent, because `versionOf(doc, DSL_VERSION_KEY)` falls back to `0.0.0`. The new test in `version.test.ts` even locks that behavior in by asserting that `migrate({ runtimeDslVersion: RUNTIME_DSL_VERSION })` picks up a document-line `dslVersion` as well. That is only harmless today because `DSL_MIGRATIONS` is still a no-op stamp. The first real Stage/Scene transform on the document ladder will still execute against runtime envelopes and can still corrupt/throw, which is exactly the class of bug this thread was trying to rule out. To make misrouted data actually inert, I think the runner needs an aggregate/line guard (for example: if the other line's stamp is present and this line's stamp is absent, return unchanged), not just disjoint stamp keys.

cosarah: Rechecked on the current head (`596daa5`). The only delta since `09e8246` is two extra negative `isIsoTimestamp` cases in `packages/@openmaic/dsl/test/runtime.test.ts`; `packages/@openmaic/dsl/src/version.ts` is unchanged, so the concern above still stands. `migrate()` still falls back to `0.0.0` when `dslVersion` is absent and will still run the document ladder over a runtime envelope that only has `runtimeDslVersion`. I still think this needs a line/aggregate guard in the runner, not just separate stamp keys.

wyuc: Agreed — disjoint stamp keys keep each ladder from *reading* the other's version, but they don't stop a misrouted aggregate from being lifted on the wrong line: lacking its own key it just reads as unversioned there, so `migrate()` on a runtime session walked the document ladder and stamped a foreign `dslVersion` (symmetric for `migrateRuntime()` on a document). Inertness now comes from a cross-line guard in the shared `runLadder`, with three cases: (1) own line's stamp present → migrate normally on the own line, regardless of the other key; (2) both stamps absent → genuine legacy, walk the own ladder; (3) own stamp absent but the sibling line's stamp present → return the input unchanged. The two ladder-independence tests that previously asserted the foreign stamp gets added are inverted into inertness regressions (misrouted input comes back unchanged, no foreign key), plus new both-stamps and neither-stamp cases. Comments and README now credit the guard rather than the key split. Fixed in 52f8e1b5.

wyuc: Follow-up in 18480142: the guard also has to live in the *predicates*. `needsMigration` / `needsRuntimeMigration` still read an absent own-line stamp as 0.0.0, so a correctly paired `while (needsMigration(x)) x = migrate(x)` over a misrouted runtime-line object would report "needs migration" while the guarded runner returns it untouched — a non-terminating loop. Both predicates now apply the same own-stamp-absent + sibling-stamp-present check, restoring the predicate/runner agreement invariant the docs promise, with tests asserting the pair (`needsX(doc) === false` and `migrateX(doc) === doc`) on misrouted data in both directions, and that doubly-stamped data still migrates on its own line.

cosarah: One edge still looks off to me: the new guard treats raw sibling-key presence as proof that the object belongs to the sibling line. But `validateRuntimeSession()` explicitly ignores a stray `dslVersion`, and an unversioned session is still considered valid runtime data. So a session like `{ ...session, dslVersion: DSL_VERSION }` now becomes permanently ineligible for the *runtime* line: `needsRuntimeMigration()` returns `false` and `migrateRuntime()` returns it unchanged, even though it still lacks `runtimeDslVersion`. That can also happen with data produced by the earlier wrong-ladder behavior if `migrate()` touched a still-unstamped session first. I think this guard needs stronger evidence than `doc[otherKey] !== undefined` (or the validator needs to reject stray sibling version keys), otherwise we lock in an orphaned one-line-stamped state.

wyuc: Fixed in db4a6046, taking both of your suggested directions at once, because the state is genuinely undecidable from the envelope: `{ ...session, dslVersion: DSL_VERSION }` is byte-identical to a document misrouted into the runtime runner, so no stronger *evidence* exists at this layer — any silent choice locks in one failure mode (migrating mangles a misrouted sibling aggregate; returning unchanged permanently orphans own-line data, exactly the state you constructed).

So the guard stops guessing: both runners **and** both `needs*Migration` predicates now throw the same error on own-stamp-absent + sibling-stamp-present, treating the ambiguous envelope like a malformed stamp (fail loud at read time, same rationale as `versionOf`). And per your alternative, `validateRuntimeSession` now rejects any present `dslVersion` on a session — the one deliberate exception to the structural-subset rule — so the state can't enter storage through a validated write path either.

On the migration-history worry: the earlier wrong-ladder behavior never shipped anywhere (the package is unpublished and this PR is the first release of the runtime envelope), so no stored data can carry that artifact; the throw is there to keep it that way.

cosarah: I think there is still one semantic gap here: the new fail-loud rule only lives in `needs*Migration()` / `migrate*()`, while `dslVersionOf()` / `runtimeDslVersionOf()` still treat the same sibling-stamp-only envelope as *unversioned* because `versionOf()` ignores the sibling field. That is not just a doc mismatch; there are already downstream call sites that use the plain readers as authoritative version reads. For example, `packages/@openmaic/storage/src/document/browser.ts` uses `dslVersionOf(stageRow)` in the `putScene()` / `deleteScene()` guards and user-facing errors, so an ambiguous row now gets reported as DSL version `0.0.0` with a "load and save it" remediation, even though the load path will immediately throw the new cross-line error. If sibling-stamp-only envelopes are now considered corrupt/ambiguous, I think the plain `*VersionOf()` helpers need the same cross-line check as the runner/predicates, or the downstream callers/docs need to stop treating them as authoritative reads.

wyuc: Fixed in d0aec6a3 — the cross-line throw now lives in the shared `versionOf` reader itself, so `dslVersionOf` / `runtimeDslVersionOf`, the `needs*Migration` predicates, and the runners give one answer per envelope; the duplicated guard sites in the runner/predicate collapsed into the reader. The storage call sites you pointed at get the corrected behavior for free: an ambiguous row in `putScene`/`deleteScene` now surfaces the cross-line error (naming both stamps and the repair options) instead of reporting DSL version 0.0.0 with a load-and-save remediation that the load path would refuse. Verified `@openmaic/storage`'s suite against the new semantics (73 passing) — its guards only ever read stage rows, which legitimately never carry `runtimeDslVersion`, so only genuinely corrupt rows can hit the throw.

cosarah:
One remaining review point: the new cross-line check makes sibling-stamped envelopes fail loud, but the wrong-ladder problem still exists for completely unstamped input. `packages/@openmaic/dsl/src/version.ts` still treats "neither stamp present" as `0.0.0` for both lines, and `packages/@openmaic/dsl/test/version.test.ts` now explicitly asserts that `migrateRuntime({ id: "z", name: "course" })` adds `runtimeDslVersion`. An unstamped legacy document is byte-identical at this layer, so once `RUNTIME_DSL_MIGRATIONS` contains a real session transform, the runtime ladder can still run over document-shaped data and corrupt/throw. Since the runtime envelope is new, I do not think the runtime line should keep an implicit `0.0.0 -> 0.1.0` lift for arbitrary unstamped objects unless there is some aggregate discriminator that proves the input is actually a `RuntimeSession`.

wyuc:
Fixed in d465f651, taking the stronger form of your suggestion: rather than adding a shape discriminator (a heuristic that widens as apps inject their own payloads), the runtime line now has **no unversioned epoch at all**. The runtime envelope is brand new — nothing legitimately predates it, and the store stamps `runtimeDslVersion` at write time — so "unstamped" is never legacy runtime data; it is a misrouted legacy document or an unstamped producer write, both bugs.

Concretely:
- `RUNTIME_DSL_MIGRATIONS` drops the implicit `0.0.0 → 0.1.0` lift and is empty until the first real session shape change appends a step from the pinned initial version. The test you cited asserting `migrateRuntime({ id, name })` stamps the object is flipped to assert the throw.
- `migrateRuntime` / `needsRuntimeMigration` / `runtimeDslVersionOf` all throw on an unstamped object (same one-answer-per-envelope rule as the cross-line guard; the shared reader takes a per-line legacy version, `null` meaning no epoch). The document line keeps its legacy lift — real pre-versioning documents exist there.
- `RuntimeSession` now **requires** `runtimeDslVersion`, in both the type and `validateRuntimeSession`, so an unstamped session can neither type-check as a stored value nor pass a validated write path — closing the same defer-to-read-time gap at the door instead of only at migrate time.

Full matrix now: own stamp present → normal walk; sibling stamp only → cross-line throw; neither stamp → document line lifts (legacy epoch exists), runtime line throws (no epoch). dsl 189 + storage 73 tests green.

cosarah:
One follow-up review point: the new runtime-line design says the *first real* runtime migration must append from `INITIAL_RUNTIME_DSL_VERSION`, but `packages/@openmaic/dsl/test/version.test.ts` no longer pins that invariant. The current future-facing test only checks contiguity and that the last `to` reaches `RUNTIME_DSL_VERSION`; a mistaken first step from `0.0.0` or `0.0.5` would still pass. Since the whole point of the new "no unversioned epoch" model is that the runtime line starts at the pinned initial version rather than a legacy origin, I think the test should also assert `RUNTIME_DSL_MIGRATIONS[0].from === INITIAL_RUNTIME_DSL_VERSION` whenever the ladder becomes non-empty.

wyuc:
Fixed in e9d480e1: the future-facing invariant test now also pins `RUNTIME_DSL_MIGRATIONS[0].from === INITIAL_RUNTIME_DSL_VERSION` whenever the ladder is non-empty, alongside the contiguity and terminal checks — a copy-pasted 0.0.0 legacy lift can no longer pass. (A local fault-injection edit used to exercise exactly that mistaken origin was accidentally swept into e9d480e1 and briefly broke the branch — the new test caught it, which is at least a live demonstration of the invariant working; restored in e114f7e0, suite green.)

cosarah (approved):
LGTM

Issue #871, opened 2026-07-08
Widget content prompts for diagram/game/code miss the postMessage listener contract → scene actions silently no-op


yanpgwang:
## Summary

The scene action pipeline dispatches widget actions (`SET_WIDGET_STATE`, `HIGHLIGHT_ELEMENT`, `ANNOTATE_ELEMENT`, `REVEAL_ELEMENT`) into the widget iframe via `postMessage` — see `lib/action/engine.ts:691-717`. For a widget to respond, its generated HTML must register `window.addEventListener('message', ...)` and handle those message types. The content-generation prompt is what instructs the model to emit that listener.

Only some widget types' prompts include the contract:

| widgetType        | prompt includes postMessage contract       | effect when platform sends `SET_WIDGET_STATE` |
|-------------------|---------------------------------------------|-----------------------------------------------|
| simulation        | ✅ `simulation-content/system.md:32`         | responds                                      |
| visualization3d   | ✅                                           | responds                                      |
| procedural-skill  | ✅                                           | responds                                      |
| **diagram**       | ❌ (`diagram-content/system.md`, 54 lines)   | **silent no-op**                              |
| **game**          | ❌ (`game-content/system.md`)                | **silent no-op**                              |
| **code**          | ❌ (`code-content/system.md`)                | **silent no-op**                              |

The widget HTML is deterministic input to the model (the model only authors the actions on top of it), so a widget type whose prompt lacks the contract produces widgets that *never* implement the handler — every scene action targeting those widgets is a no-op. Presence/absence of the contract section in the prompt lines up exactly with whether generated widgets can be driven.

## Why this looks like an oversight rather than intended

PR #796 ("Refactor widget actions into the scene action pipeline") introduced platform-driven widget actions and, in the same change, added the `## CRITICAL: postMessage Listener for Widget Actions` section to the `simulation`, `visualization3d`, and `procedural-skill` content prompts. The `diagram` and `game` prompts were not touched, and the `code` prompt received only a small edit without the listener section. So the three affected types share a single root cause: the contract was rolled out to some widget prompts but not all.

## Point 1 (primary) — diagram / game / code miss the contract

Copy the `## CRITICAL: postMessage Listener for Widget Actions` section (and the small Element Naming Convention note) from `simulation-content/system.md:32-108` into the `diagram`, `game`, and `code` content prompts, adapting the handler examples to each widget type's DOM. This makes all interactive widget types uniformly drivable.

[...]

PR #872, opened 2026-07-08
fix(prompts): add postMessage listener contract to diagram/game/code widgets

yanpgwang:
## Summary

Widget actions are delivered to the iframe via `postMessage` (`lib/action/engine.ts:691-717`), so a widget must register a `message` listener to be driven by scene actions. The `simulation`, `visualization3d`, and `procedural-skill` content prompts document this contract; `diagram`, `game`, and `code` did not — so widgets of those types never emit the handler and every scene action targeting them silently no-ops.

This adds the `## CRITICAL: postMessage Listener for Widget Actions` section (handling `SET_WIDGET_STATE` / `HIGHLIGHT_ELEMENT` / `ANNOTATE_ELEMENT` / `REVEAL_ELEMENT`) to the three prompts, adapting the `SET_WIDGET_STATE` handler to each widget type's DOM and mirroring the existing simulation prompt for the highlight/annotate/reveal handlers. Each prompt also gets an Element Naming Convention note so those actions have stable targets.

The message shape (`{ type, target, state, content }`) matches what the iframe host posts (`components/scene-renderers/InteractiveIframeHost.tsx:109` sends `{ type, ...payload }`) and what the existing simulation handler already consumes.

Addresses point 1 of #871. The simulation select-variable issue (point 2) is left for separate discussion.

## Test plan

- Generate one widget of each affected type (`diagram`, `game`, `code`) and confirm the output HTML registers a `window` `message` listener handling the four action types.
- Replay each action against a generated widget and confirm it responds (highlight appears, state applies, annotation shows, element reveals) instead of no-op.

### DIFF
--- a/lib/prompts/templates/code-content/system.md
+++ b/lib/prompts/templates/code-content/system.md
@@ -139,7 +139,10 @@ window.addEventListener('message', function(event) {
     case 'SET_WIDGET_STATE':
       // e.g. { code: "...", run: true } — set editor contents, optionally run.
       if (state && typeof state.code === 'string') {
-        if (typeof editor?.setValue === 'function') editor.setValue(state.code);
+        // Guard the identifier itself: `editor?.setValue` still throws
+        // ReferenceError when no `editor` variable is declared (e.g. a
+        // textarea-only widget). `typeof editor` is safe for that case.
+        if (typeof editor !== 'undefined' && typeof editor.setValue === 'function') editor.setValue(state.code);
         else { const ta = document.getElementById('code-input'); if (ta) ta.value = state.code; }
       }
       if (state && state.run && typeof runCode === 'function') runCode();
--- a/lib/prompts/templates/code-content/system.md
+++ b/lib/prompts/templates/code-content/system.md
@@ -123,6 +123,76 @@ async function runCode() {
 - Ensure editor has minimum height of 200px on mobile
 - Test cases should be collapsible on small screens
 
+## CRITICAL: postMessage Listener for Widget Actions (REQUIRED)
+
+The platform drives this widget by posting messages into the iframe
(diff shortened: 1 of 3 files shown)


wyuc (changes requested):
Found one small but blocking issue in the new code-widget prompt example.

In `lib/prompts/templates/code-content/system.md`, the `SET_WIDGET_STATE` sample uses `typeof editor?.setValue === 'function'` before falling back to `#code-input`. If a generated code widget uses a textarea or a differently named editor variable, `editor` is undeclared and this expression throws `ReferenceError: editor is not defined`; the fallback path never runs. That means the newly documented handler can still crash/no-op for valid code widgets, which cuts against this PR's goal.

Please make the sample guard undeclared editors explicitly, e.g. `typeof editor !== 'undefined' && typeof editor.setValue === 'function'`, then fall back to `#code-input`. The diagram/game prompt additions look fine to me.

yanpgwang:
Good catch — confirmed and fixed in 4575be3.

You're right: `editor?.setValue` only guards a null/undefined *value*, but when `editor` is an undeclared identifier (a textarea-only code widget), evaluating the expression throws `ReferenceError: editor is not defined` before optional chaining can help, so the `#code-input` fallback is unreachable.

Changed the sample to guard the identifier first:

```js
if (typeof editor !== 'undefined' && typeof editor.setValue === 'function') editor.setValue(state.code);
else { const ta = document.getElementById('code-input'); if (ta) ta.value = state.code; }
```

Verified both paths in a real browser by posting `SET_WIDGET_STATE` into a textarea-only iframe (no `editor` declared):
- old guard → `ReferenceError: editor is not defined`, fallback never runs
- new guard → no error, falls back to the textarea and sets the code

`runCode` on the next line was already safe (`typeof runCode === 'function'` on an undeclared identifier doesn't throw), so I left it as-is.

wyuc (approved):
The undeclared editor guard is fixed in 4575be3, the textarea fallback is now reachable, and both CI suites are green. The diagram/game/code prompt additions otherwise look good.

Issue #879, opened 2026-07-08
[Task] @openmaic/storage RuntimeStore — session/record persistence, browser backend + contract suite (#869 Part B)


wyuc:
Tracking: #869 (RFC), #779 (layer model). Part A (#870, merged) landed the DSL-native runtime envelope; this issue is the executable slice for the store that persists it.

## Scope

`RuntimeStore` in `packages/@openmaic/storage`: interface + browser backend + backend-agnostic contract suite — a mirror of the DocumentStore delivery (#860), reusing the established package patterns (#858): dependency rule `@openmaic/storage` → `@openmaic/dsl` only, injected native `IDBFactory` (no Dexie), `fake-indexeddb` in tests, `runXStoreContract(name, makeStore)` factory suites.

## Contract points already fixed by Part A

- **Sessions are born stamped**: the store stamps `runtimeDslVersion: RUNTIME_DSL_VERSION` at write time — the runtime line has no unversioned epoch, and `validateRuntimeSession` rejects an unstamped session.
- **Migrate-on-read** via `migrateRuntime` (never the document line's `migrate`); ambiguous cross-line envelopes fail loud by design, and the store must surface those errors, not swallow them.
- **Ordering is store-owned**: `RuntimeRecord.seq` is a per-session monotonic key assigned on append (`RuntimeRecordInit` omits it structurally); timestamps are display metadata only.
- **Multi-tenancy**: sessions are keyed by `(stageId, learnerKey)` plus `kind`; `learnerKey` is an opaque exact-match partition key.

## In scope

1. **Session CRUD**: create (stamping version + validating), get, update status ladder, list by `(stageId, learnerKey)` and by `stageId`.
2. **Record append + reads**: append assigns `seq` atomically per session; get/list by session in `seq` order; query by `sceneId` anchor.
3. **`mergeLearner(fromKey, toKey)`**: anonymous→signed-in identity migration; idempotent, never clobbers existing records of the target key.
4. **Per-learner deletion** (a learner clears own sessions without touching the document) and **document-cascade deletion** (deleting a stage's runtime removes all learners' sessions for it).
5. **Core kinds** `chat` / `quizAttempt` / `playback` flow through with the Part A skeleton guards; app-defined kinds need no new interface (one store + typed accessors, per the RFC's leaning on open question 1).
[...]

PR #880, opened 2026-07-08
feat(storage): RuntimeStore — sessions + append-only records, browser backend (#869 Part B)

wyuc:
Part B of #869: the `RuntimeStore` persistence contract for the learner-runtime layer — interface + browser (IndexedDB) backend + backend-agnostic contract suite in `@openmaic/storage`, mirroring the DocumentStore delivery (#860). Builds directly on the Part A envelope (#870).

## What

- **`RuntimeStore`** (`src/runtime/types.ts`) — sessions are multi-tenant by `(stageId, learnerKey)`; records are append-only ordered facts under an **active** session, with the store-assigned per-session monotonic `seq` as the sole replay ordering key. Operations: session create/get/list/status/delete, `appendRecord`/`listRecords` (with best-effort `sceneId` narrowing), `mergeLearner` (the anonymous-learner-signs-in re-key, global by design), and the two deletion cascades (`deleteLearnerRuntime`, `deleteStageRuntime` — the document-deletion hook).
- **`BrowserRuntimeStore`** (`src/runtime/browser.ts`) — raw IndexedDB via an injected `IDBFactory` (no Dexie), the document backend's exact idioms: memoized open, transactions resolving on commit, fail-loud validation before the write transaction opens.
- **Version-line duty**: sessions are **born stamped** — the store writes `runtimeDslVersion: RUNTIME_DSL_VERSION` at `createSession`; there is no code path that persists an unstamped session (the runtime line has no unversioned epoch, per the Part A review). Reads migrate via `migrateRuntime`; writes reject future-stamped rows and migrate stale rows in place (a documented divergence from `putScene`: records carry no version of their own, and the interface has no full-session save for a caller-side remedy).
- **Per-kind payload validators** injected like `SceneValidator` (#860), defaulting to the DSL skeleton guards for `chat`/`quizAttempt`; `playback` and app-defined kinds carry app-owned payloads.
- **Contract suite** `runRuntimeStoreContract(name, makeStore)` + backend-specific version-skew tests (stale / future / corrupt / sibling-stamped seeded rows).
- **CI**: adds `pnpm --filter @openmaic/storage test` — the package's suite (including the pre-existing DocumentStore/KV/asset contracts) previously never ran in CI.

## Behavior decisions worth reviewer attention

- **Listings tolerate corrupt rows by omission** (the `listDocuments` precedent): one poison row must not make a whole partition unenumerable. Direct `getSession` on such an id stays fail-loud, and the delete paths are the cleanup tool.
- **`mergeLearner` takes every write gate**: key validation, per-row future-guard, migrate-before-re-key, envelope validation — any throw aborts the whole merge atomically. Self-merge is a no-op returning 0.
- **The contract pins gapless `seq` from 0** — intentional strictness: a backend that cannot serialize appends (e.g. a future HTTP backend without server-side sequencing) should fail the suite, not silently diverge.

## Deliberately out of scope (noted for the record)

- Replay-asset export (#779 open question 3) — follow-up once the export shape is settled.
- Bulk-import / seq-backfill primitives for migrating the app's existing Dexie/localStorage data — they belong to the app-side strangler migration, which will also need a permissive `payloadValidators` override for legacy shapes.
- Read paths do not persist migrations (only writes do) — re-reads re-run the ladder; acceptable while migrations are cheap, revisit if a heavy migration ever lands.

## Tests

103 passing (`pnpm --filter @openmaic/storage test`); typecheck, build, prettier, eslint clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,9 @@ jobs:
       - name: Unit Tests (dsl)
         run: pnpm --filter @openmaic/dsl test
 
+      - name: Unit Tests (storage)
+        run: pnpm --filter @openmaic/storage test
+
   e2e:
     name: E2E Tests
     runs-on: ubuntu-latest
--- a/packages/@openmaic/storage/README.md
+++ b/packages/@openmaic/storage/README.md
@@ -52,8 +52,10 @@ a browser.
   store over its scene union and injects a matching `validateScene`, so those
   scenes persist and the gate stays fail-loud for the app's shapes.
 - **Runtime layer.** `RuntimeStore` is partitioned by `(stageId, learnerKey)`:
-  a stage has many sessions — one or more per learner — so every query is
-  partition-scoped, never global. Sessions are **born stamped**: the store
+  a stage has many sessions — one or more per learner — so every listing is
+  partition-scoped (there is deliberately no global listing; single-session
+  operations are id-keyed, and `mergeLearner` is the one deliberate
+  cross-stage sweep). Sessions are **born stamped**: the store
   writes `runtimeDslVersion` itself at `createSession`, and the runtime line
   has no unversioned epoch, so an unstamped row fails loud instead of being
   lifted like a legacy document. Records are **append-only** ordered facts
--- a/packages/@openmaic/storage/README.md
+++ b/packages/@openmaic/storage/README.md
@@ -28,6 +28,7 @@ a browser.
 | `StorageProvider` (from `@openmaic/dsl`) | the asset seam: `put(blob) → ref`, `resolve(ref) → url`, `remove(ref)` | `BrowserAssetProvider` over IndexedDB + object URLs |
 | `kvPersistStorage` | adapt a `KVStore` into a zustand `persist` storage | — |
 | `DocumentStore` | persist the DSL `document` aggregate (stage + scenes + embedded agents / quiz / actions + an outline snapshot) | `BrowserDocumentStore` over IndexedDB (normalized `stages` / `scenes` / `outlines`) |
+| `RuntimeStore` | persist what a learner produces while taking a course — sessions + append-only records (chat, quiz attempts, playback facts) | `BrowserRuntimeStore` over IndexedDB (`sessions` / `records`) |
 
 - **Scopes.** `account` values are user data a server-backed deployment syncs
   across devices; `device` values (theme, locale, layout) never leave the
@@ -50,11 +51,27 @@ a browser.
   kinds (`interactive` / `pbl`, content the DSL does not own) parameterizes the
   store over its scene union and injects a matching `validateScene`, so those
   scenes persist and the gate stays fail-loud for the app's shapes.
+- **Runtime layer.** `RuntimeStore` is partitioned by `(stageId, learnerKey)`:
+  a stage has many sessions — one or more per learner — so every query is
+  partition-scoped, never global. Sessions are **born stamped**: the store
+  writes `runtimeDslVersion` itself at `createSession`, and the runtime line
+  has no unversioned epoch, so an unstamped row fails loud instead of being
+  lifted like a legacy document. Records are **append-only** ordered facts
+  under an **active** session; the store assigns the per-session monotonic
+  `seq` on append — the sole replay ordering key, never timestamps. Record
+  payloads are gated per kind by injectable validators, defaulting to the DSL
+  skeleton guards for `chat` / `quizAttempt` (`playback` and app-defined kinds
+  carry app-owned payloads). `mergeLearner` re-keys an anonymous learner's
+  sessions to a signed-in key across all stages; `deleteLearnerRuntime`
+  cascades one learner's sessions + records on one stage, and
+  `deleteStageRuntime` clears a whole stage — the hook a document deletion
+  cascades through.
 
 ## Backend equivalence
 
 Each primitive has one implementation-agnostic contract suite
-(`test/kv-contract.ts`, `test/asset-contract.ts`, `test/document-contract.ts`).
+(`test/kv-contract.ts`, `test/asset-contract.ts`, `test/document-contract.ts`,
+`test/runtime-contract.ts`).
 Every backend is proven by running the same suite against it, so a new backend
 (the coming HTTP one) cannot silently diverge from the primitive's semantics.
 
@@ -65,8 +82,9 @@ Every backend is proven by running the same suite against it, so a new backend
 - [x] implementation-agnostic contract suites
 - [x] `DocumentStore` (aggregate ↔ normalized adapter, migrate-on-read via the
       DSL migration registry, validation gate) + browser backend
+- [x] `RuntimeStore` (sessions + append-only records, runtime version line,
+      per-kind payload gate) + browser backend
 - [ ] wire the app's zustand stores + ad-hoc `localStorage` through `KVStore`
-- [ ] `RuntimeStore`
 - [ ] HTTP backend + reference server + one HTTP contract
 
 ## License
--- a/packages/@openmaic/storage/src/index.ts
+++ b/packages/@openmaic/storage/src/index.ts
@@ -33,6 +33,8 @@ export type {
 } from './document/types.js';
 export { BrowserDocumentStore, type BrowserDocumentStoreOptions } from './document/browser.js';
 
+export type { RuntimeStore, RuntimeSessionInit, RuntimePayloadValidator } from './runtime/types.js';
+
 // Re-export the DSL-owned asset contract for convenience, so consumers can get
 // the interface and a backend from one import without reaching into the DSL.
 export type { AssetRef, AssetMeta, BinaryBlob, StorageProvider } from '@openmaic/dsl';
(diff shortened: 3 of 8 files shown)


wyuc (commented):
**Cross-review report** (4 independent reviewers over the full branch diff: Codex `exec review` at high effort + three task-shaped Claude reviewers — state-space matrix, claim falsification + invariant-test audit, evolution probe; findings merged by consensus). Tracking task: #879.

**Verified holding by construction** (stating explicitly since these are the load-bearing claims): seq assignment + `records.add` share one readwrite transaction over `[SESSIONS, RECORDS]`, so concurrent appends serialize — monotonicity is not a check-then-write race; `mergeLearner` is non-clobbering structurally (global `id` keyPath makes source/target collisions impossible), atomic (single `txRun`), idempotent (tested twice-over); cascade deletes are record-complete on all three paths (tested); no `switch(kind)` anywhere — app-defined kinds flow through per the RFC; exports/CI/import-boundary wiring is sound.

## P2

1. **Records have no migrate-on-read; the first real runtime migration strands them.** `listRecords` returns raw rows; `migrateRuntime` only ever touches session rows. `src/runtime/types.ts` justifies diverging from #860's stale-sibling guard with *"records carry no version, so there are no unmigrated siblings to strand"* — but once `RUNTIME_DSL_MIGRATIONS` gains a real `0.1.0 → 0.2.0` step touching the record envelope (`sceneId`/`actionIndex`/`subAnchor`/`createdAt` are DSL-owned), `appendRecord` migrates + re-`put`s the parent session while its stored records stay old-shaped and unversioned — mixed-shape replay under a current-stamped session, unmigratable by construction. The session is the versioned row; its records are the stranded siblings. Either document the record envelope as frozen, or add a record-migration hook before the first real transform lands.
2. **IDB upgrade is keyed on store-existence, not `oldVersion`.** The `onupgradeneeded` body guards with `if (!db.objectStoreNames.contains(...))` and never reads `event.oldVersion` / `req.transaction`. A later Part adding an index to the existing `sessions` store (v1→v2) skips the guarded block entirely for every returning user — `.index('by-…')` then throws `NotFoundError` precisely for existing-data users. Needs an `oldVersion`-keyed ladder. (Same idiom exists in `document/browser.ts` from #860 — pre-existing there, worth fixing at least here.)

## P3

3. **`appendRecord` is the only write path that persists a session without `validateRuntimeSession`** *(two reviewers independently)*: the migrate-in-place branch `put`s the migrated parent unvalidated, while `setSessionStatus`/`mergeLearner` assert before every `put`. Latent today (empty ladder), becomes live with the first real migration — a buggy migration output would persist unvalidated on this path only. Same root cause makes entry points disagree on one input: a doubly-stamped parent (stray `dslVersion` + valid `runtimeDslVersion`) is appendable but rejected by `setSessionStatus`/`mergeLearner`.
4. **`listSessions` "ordered by createdAt" is breakable by forward-compat rows**: future-stamped sessions are deliberately passed through, but their `createdAt` never re-passed `isIsoTimestamp` on this client — `Date.parse` → `NaN` gives the comparator an inconsistent order that can scramble the *valid* rows too. The adjacent "safe here" comment is false for exactly the rows the store lets through.
5. **Delete paths are version-agnostic while every mutation path guards future-stamped rows** — documented as intended; flagging for explicit ratification that an older client destroying a newer client's sessions is the chosen semantics.
6. **Contract suite over/under-constrains a second backend** *(two reviewers independently on the second half)*: byte-identical assertions (`toEqual(created)`, `'runtimeDslVersion' in r0 === false`) reject conformant HTTP/SQL backends that echo server-managed columns; meanwhile there is **no concurrent-append case** (`Promise.all` interleave), so a remote backend that duplicates `seq` under concurrency would pass — the semantic most likely to diverge on a networked backend is the one untested.
7. **Test gaps**: merge-preserves-records never asserted (append → merge → `listRecords` intact); present-but-malformed stored stamp (`runtimeDslVersion: 'legacy'`) not seeded against any write path (fail-loud behavior unpinned).
8. **Doc drift**: `src/index.ts` still says *"the `RuntimeStore` follow in later parts"* right above its export; `package.json` description names only KV/asset.

Codex reported no discrete correctness issues (consistent with the invariants-hold-by-construction picture; the findings above are state-space/evolution class).

wyuc:
**Cross-model review summary** (pre-submission; two independent reviewers over the full branch diff: a Claude Code multi-angle pass — including a version state-space matrix over {unstamped, stale, current, future, sibling-stamped, malformed} × every public method, docstring falsification, and evolution probes — plus Codex `exec review` at high effort; findings merged by consensus):

- **Round 1** — 10 findings fixed in 21cac07c. The two both reviewers hit independently (P2): `mergeLearner` was the only write path with neither envelope validation nor a version guard (it could re-key poison rows into clean partitions and mutate future-stamped sessions — it now takes every write gate and aborts atomically), and `listSessions` sorted by timestamp *text*, mis-ordering valid ISO offsets (now sorts by instant). The state-space matrix also surfaced that one corrupt row made a whole partition unenumerable — listings now tolerate corrupt rows by omission (the `listDocuments` precedent) while direct `getSession` stays fail-loud. Docstring-truth fixes: self-merge now returns 0 (idempotency held only for the non-degenerate case), `appendRecord` now validates the *actual* completed envelope (with the store-assigned `seq`) inside the transaction, corrupt-stamp throw reasons documented and pinned by tests, `seq` derived via `openKeyCursor` (no full-payload deserialization), and the `IDBKeyRange` shim moved to `test/setup.ts`.
- **Round 2** — re-review found one P2: `mergeLearner` re-keyed the *raw* stored row, skipping the migrate-before-mutate step every other write path takes; fixed in e73c53db (future-guard → migrate → re-key → validate → put).
- **Final verification pass**: no findings. 103 tests passing; typecheck, build, prettier, eslint clean; CI green (including the storage suite, which this PR wires into CI for the first time).

Known deferrals are listed in the PR description (replay-asset export, bulk-import primitives for the app-side migration, read paths not persisting migrations); the contract's gapless-`seq` pin is intentional strictness for future backends.

cosarah (commented):
1. packages/@openmaic/storage/src/runtime/browser.ts:241-273 claims corrupt rows are omitted from listings and direct getSession stays fail-loud, but the implementation only runs migrateSession() on read and never re-validates the envelope. A row with a valid runtimeDslVersion but invalid createdAt/status/etc. is returned as-is. I reproduced this by rewriting createdAt to "not-iso" via raw IDB; both getSession() and listSessions() surfaced the bad row. If this fail-loud/omit-corrupt contract is intended, reads need post-migration validateRuntimeSession() gating and listSessions() should drop rows that fail it.\n\n2. packages/@openmaic/storage/src/runtime/types.ts:103-122 ships an append-only API (appendRecord/listRecords only), but RFC issue #869 still says RuntimeStore should support "record append/upsert within a session". If append-only is the intended design pivot, the RFC/spec should be updated before merge; otherwise Part B is still missing an interface operation the issue promised.

wyuc:
Both points addressed:

1. **Read-path envelope validation** — fixed in 5eb7a911. Reads now gate stored rows through the same `validateRuntimeSession` as the writes, post-migration: `getSession` fails loud on a row whose stamp resolves but whose envelope is corrupt (your `createdAt: "not-iso"` repro is now a pinned regression test), and `listSessions` omits such rows under the same per-row tolerance it applies to version corruption — one poison row still can't make a partition unenumerable. `setSessionStatus`/`appendRecord` were already covered: their post-update `assertValid` rejects pre-existing corruption before any write.

2. **RFC vs append-only** — the divergence was a deliberate pivot, and you're right that the spec should say so: #869's RuntimeStore section is now amended (append/upsert → append-only, with a note that mutable learner state is modeled as events to fold rather than records to rewrite). That is also the shape Part C intends for the PBL learner-state split — milestone/microtask status changes arrive as appended facts and fold at read time, which keeps replay lossless; an upsert API would have destroyed the history replay depends on.

cosarah (commented):
Follow-up on the latest head (5eb7a911): the new read-path envelope validation closes the corrupt-row gap I flagged earlier; I rechecked the raw-IDB reproduction and direct getSession now fails loud while listSessions omits the bad row. Remaining concern: RuntimeStore still exposes an append-only record API (appendRecord/listRecords in packages/@openmaic/storage/src/runtime/types.ts), while RFC issue #869 still describes Part B as supporting record append/upsert within a session. If append-only is the final design, the issue/spec should be updated before merge; otherwise this PR is still missing part of the promised interface.

cosarah (commented):
Remaining concern: in packages/@openmaic/storage/src/runtime/types.ts, RuntimeStore still exposes an append-only record API (appendRecord/listRecords only), while RFC issue #869 still describes Part B as supporting record append/upsert within a session. If append-only is the intended design change, the issue/spec should be updated before merge; otherwise this PR is still missing part of the promised public interface.

wyuc:
Our messages crossed — the #869 RFC body was amended right as your follow-up reviews landed. The RuntimeStore operations sentence now reads:

> Operations: session CRUD, append-only record append within a session, query by anchor (`sceneId`). *(Amended during Part B review: the original "append/upsert" wording is dropped — records are immutable ordered facts, and mutable learner state is modeled as events to fold at read time, not records to rewrite. See the discussion on #880.)*

So the spec and this PR's interface now agree. Thanks for confirming the read-path fix closes the corrupt-row gap.

cosarah (approved):
LGTM

PR #881, opened 2026-07-08
feat(renderer/editing): 8-point resize + rotate operate handles (#851 Part A)

wyuc:
Continues #851 Part A after the selection + drag-to-move slice (#859): the next two deferred capabilities of the `@openmaic/renderer` editing surface. Nothing in the app imports these yet, so this merges inert — no runtime behavior change.

## What

- **`core/resize.ts`** — `computeResize`: the unrotated path snaps the moving edge (before min-size clamping, with the aspect re-lock applied after a snap), the rotated path rotates the pointer delta into element-local axes and applies the opposite-point correction so the handle diagonally opposite the drag stays visually fixed. Per-type minimum sizes; aspect lock from a held modifier (Ctrl/Shift/Meta) or the element's `fixedRatio`, corners only; rotation-bucketed resize cursors; per-kind handle gates (text exposes only its width axis — left/right, or top/bottom when `vertical`; code gets none; line is excluded — its chrome is the landed `LineHandles`).
- **`core/rotate.ts`** — `computeRotate`: signed `(-180, 180]` angle from the pointer around the element center, snapped to 45° multiples within range 5; `canRotate` gate (chart/video/audio excluded, per the app's operate dispatch).
- **`useResizeGesture` / `useRotateGesture`** + **`ResizeHandles` / `RotateHandle`**, wired into `EditableSlideCanvas` behind the existing editability gates. Handle positions are computed analytically in screen space from the same rotated-points math as the resize correction (not nested in a rotated CSS frame). All the landed gesture invariants hold: one `element.update` intent per completed gesture, pointercancel reverts, single-pointer guard, 2px threshold, working-copy live preview.
- Operate handles render only for a **single-element** selection — group scaling is a later slice, and per-element handles on a multi-selection would misread as it.

## Deferred (stated in code comments)

Content-space recomputation stays host-side or comes in later slices: shape `pathFormula`/`viewBox` refresh, table row-height redistribution, image clip mode, group (multi-element) scaling, and guide-line rendering (guides are computed, matching the landed move gesture, but not yet drawn).

## Review

Cross-model review before pushing (Codex `exec review` high effort + an independent red-team pass over the geometry against the in-repo reference implementation, the kind × handle × rotate × lock state matrix, and the gesture protocol): no P1/P2. The two accepted P3s are fixed in the second commit (no-op resize intents after a snap-back no longer emitted; handles gated to single selection). Known accepted deviations: the aspect modifier also accepts Meta (Cmd on macOS); resize snap candidates include sibling center lines (the shared `buildAlignLines`), which the in-app resize did not use.

## Tests & boundary

134 tests across the editing subpath (≈70 new: pure geometry with hand-computed expectations incl. the rotated opposite-point invariant, both gesture hooks, handle components, canvas wiring); package build emits `dist/editing`; `tsc` clean; respects the #853 import boundary (no host `@/…` imports).

### DIFF
--- a/packages/@openmaic/renderer/src/editing/EditableSlideCanvas.tsx
+++ b/packages/@openmaic/renderer/src/editing/EditableSlideCanvas.tsx
@@ -437,8 +437,12 @@ export function EditableSlideCanvas(props: EditableSlideCanvasProps) {
               absolute SCREEN coordinates (baking in `viewportStyles`) and are
               gated on EDITABILITY (`onElementsChange`), not generic
               `interactive`: without a mutation channel the gestures no-op, so
-              a select-only mount shows no draggable handles. */}
+              a select-only mount shows no draggable handles. Shown only for a
+              SINGLE-element selection: these gestures transform one element,
+              and per-element handles on a multi-selection would misread as
+              group scaling (which is a later slice). */}
           {Boolean(onElementsChange) &&
+            activeSelection.elementIds.length === 1 &&
             elements.map((el) => {
               if (el.type === 'line') return null;
               if (!activeSelection.elementIds.includes(el.id) || el.lock) return null;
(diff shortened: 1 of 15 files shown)


cosarah (approved):
Manually verified locally. I also confirmed that table elements in thumbnails, which were visibly distorted before this PR, now render correctly.

PR #885, opened 2026-07-09
feat(runtime): device-anonymous learner identity + RuntimeStore app bootstrap (#869 Part C, step 1)

wyuc:
First step of #869 Part C (the PBL runtime split), deliberately tiny and behavior-neutral: the runtime layer's identity and storage get wired into the app, unused. This is the foundation the strangler migration (chat / quiz / playback / PBL learner state) will build on.

## What

- **`lib/runtime/learner-key.ts`** — `getLearnerKey()`: a device-anonymous `anon:<uuid>` minted once and persisted through the KV **`device`** scope (explicitly not `account`: a synced "device" key would merge two people's runtime into one partition). Minting is race-safe: cross-tab mutual exclusion via the Web Locks API (with a documented fallback for environments without `navigator.locks`), in-flight promise memoization within a bundle, and read-after-write convergence — once a key exists it is never overwritten. When sign-in lands, `RuntimeStore.mergeLearner(anonKey, accountKey)` is the purpose-built migration for this key.
- **`lib/runtime/store.ts`** — lazy app-wide `BrowserRuntimeStore` singleton (its own `maic-runtime` IndexedDB), plus `deleteStageRuntimeSafely()`: the stage-deletion cascade, bounded by a 5s timeout and never throwing — a broken or hung runtime DB must not brick stage deletion in the main app DB (orphaned runtime rows are inert; a startup sweep is deliberately deferred until the store gains real readers in the next step).
- **Cascade wired into the live deletion path** (`deleteStageData` in `lib/utils/stage-storage.ts` — the flow the classroom UI actually calls) and into `deleteStageWithRelatedData`.
- First app-side dependency on `@openmaic/storage`.

## Notes for reviewers

- Zero user-visible behavior change: nothing reads or writes runtime data yet; the cascade only ever deletes rows that cannot exist until a writer lands.
- Both `lib/runtime` modules are client-only (they lazily touch `localStorage`/`indexedDB`); server code must inject stores explicitly.
- Found during review, out of scope here: `deleteStageData` and `deleteStageWithRelatedData` overlap but delete *different* table sets — worth a follow-up to converge them.

## Tests

2001 passing at root (12 new under `tests/runtime/`), typecheck/prettier/eslint clean, `@openmaic/storage` suite untouched (105).

### DIFF
--- a/lib/runtime/learner-key.ts
+++ b/lib/runtime/learner-key.ts
@@ -1,0 +1,23 @@
+/**
+ * Device-anonymous learner identity for the runtime layer (#869).
+ *
+ * `learnerKey` partitions all learner-runtime data (RuntimeStore sessions).
+ * Until sign-in exists it is a per-device anonymous key: minted once and kept
+ * in the KV `device` scope, which never syncs across devices — a synced key
+ * would merge two people's runtime into one partition. When sign-in lands,
+ * `RuntimeStore.mergeLearner(anonKey, accountKey)` is the migration path.
+ */
+import { BrowserKVStore, type KVStore } from '@openmaic/storage';
+
+export const LEARNER_KEY_KV_KEY = 'runtime.learnerKey';
+
+let defaultKv: KVStore | undefined;
+
+export async function getLearnerKey(kv?: KVStore): Promise<string> {
+  const store = kv ?? (defaultKv ??= new BrowserKVStore());
+  const existing = await store.get<string>(LEARNER_KEY_KV_KEY, 'device');
+  if (existing) return existing;
+  const minted = `anon:${crypto.randomUUID()}`;
+  await store.set(LEARNER_KEY_KV_KEY, minted, 'device');
+  return minted;
+}
--- a/lib/runtime/learner-key.ts
+++ b/lib/runtime/learner-key.ts
@@ -6,18 +6,45 @@
  * in the KV `device` scope, which never syncs across devices — a synced key
  * would merge two people's runtime into one partition. When sign-in lands,
  * `RuntimeStore.mergeLearner(anonKey, accountKey)` is the migration path.
+ *
+ * Client-only: the default KV store lazily touches `localStorage`. Server
+ * code must not import this without injecting its own `KVStore`.
  */
 import { BrowserKVStore, type KVStore } from '@openmaic/storage';
 
 export const LEARNER_KEY_KV_KEY = 'runtime.learnerKey';
 
 let defaultKv: KVStore | undefined;
+let defaultInFlight: Promise<string> | undefined;
 
-export async function getLearnerKey(kv?: KVStore): Promise<string> {
-  const store = kv ?? (defaultKv ??= new BrowserKVStore());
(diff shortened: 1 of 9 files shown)


wyuc:
**Cross-model review summary** (pre-submission; Claude Code multi-angle + Codex `exec review` high effort, findings merged by consensus, two fix rounds to closure):

- **Round 1** (fixed in fb350e18): the deletion cascade was originally wired into `deleteStageWithRelatedData`, which has no callers — the classroom UI deletes through `deleteStageData`; the cascade now sits on the live path (both, actually). The learner-key mint had a read-then-write race (both reviewers independently); first pass added in-flight memoization + read-after-write convergence. Also: `crypto.randomUUID` now guarded per the codebase's existing convention, the live wiring got an integration test, and the "retryable" doc claim was removed (no retry path exists; orphaned rows are inert).
- **Round 2** (fixed in 93a64c18): re-review showed read-after-write still loses one interleaving cross-tab — minting now runs under the Web Locks API (`navigator.locks`, serializing fake in tests; documented fallback where unavailable), and once a key exists it is never overwritten. The cascade `await` was unbounded — a hung runtime DB could block stage deletion; it is now raced against a 5s bound, with late rejections kept off the unhandled-rejection path (test-pinned).
- **Final verification pass**: no findings. 2001 root tests green; CI green.

Known overlap flagged for a follow-up (out of scope here): `deleteStageData` vs `deleteStageWithRelatedData` delete different table sets.

cosarah (commented):
Medium: the new live deletion cascade now always touches the runtime IndexedDB even though this PR explicitly says the runtime layer still has no writers yet. `deleteStageData()` in `lib/utils/stage-storage.ts:152` now calls `deleteStageRuntimeSafely()`, and that helper unconditionally falls through to `getRuntimeStore().deleteStageRuntime(stageId)` in `lib/runtime/store.ts:51`. `BrowserRuntimeStore.deleteStageRuntime()` then opens `maic-runtime` through `indexedDB.open()` in `packages/@openmaic/storage/src/runtime/browser.ts:149`, creating the database if it does not exist, before doing the delete at `packages/@openmaic/storage/src/runtime/browser.ts:443`. So for a user who has never written any runtime data, deleting a classroom now still pays the cost of opening or creating a second IndexedDB database; in degraded environments (private mode, broken IDB, hung IDB) this path can also burn the full 5s timeout for no cleanup value at all. I think the safer behavior here is to only cascade if the runtime DB is already known to exist, or to defer wiring the live delete path until Part C2 introduces real runtime writers.

wyuc:
Good catch — fixed in a1fa1fde. `deleteStageRuntimeSafely` now probes with `indexedDB.databases()` first and returns immediately when no `maic-runtime` database exists, so deletion on a device that never wrote runtime data no longer creates the DB (or burns the timeout) just to clean nothing. Where the probe API is unavailable (older Firefox) it falls through to the existing bounded cascade, which shares the same timeout/catch envelope so a hanging `databases()` cannot block deletion either.

cosarah (approved):
LGTM

PR #887, opened 2026-07-09
feat(extraction): media (audio/video) extraction + AliDocMind provider

### LINKED ISSUE #886: MAIC ETL: media (audio/video) extraction + AliDocMind provider
## Summary

Add the **media (audio/video) extraction** stage of MAIC ETL (#621), plus **AliDocMind** (Aliyun Document Mind, LLM version) as a new extraction vendor that covers both the document and media paths.

This is the sibling to the document extraction work landed in #741 (jackefn). Document extraction handles file-like input (pdf/docx/pptx/…) and produces `DocumentArtifact`; media extraction handles audio/video and produces a timestamp-anchored `MediaArtifact`.

## Why a separate `MediaArtifact`

Media's native anchor is a timestamp (`transcript[].startMs/endMs`, `keyframes[].timeMs`), not a page/block. Squeezing audio/video into `DocumentArtifact.blocks` loses the timeline. AliDocMind's own API confirms this shape: its document result is `layouts[]`, its media result is `segments[]` with `audio_frames`/`video_frames`. We keep two artifact types and a shared success/failure envelope (`ExtractionResult`).

## Scope

**Document side (AliDocMind as a new provider):**
- `alidocmind` entry in `PDF_PROVIDERS`, selectable anywhere MinerU is.
- Supports pdf/docx/pptx/xlsx + images through one submit→poll→get flow.
- AK/SK credentials (not a single apiKey), with env fallback.

**Media side (new domain):**
- `lib/media-parse/` mirroring `lib/pdf/` (types/constants/providers).
- `MediaArtifact` + `MediaExtractorProvider` + media registry + `extractMedia()` entry, symmetric to the document side.
- AliDocMind maps `segments[]`/`audio_frames`/`video_frames` → transcript + keyframes.

## Design notes

- AliDocMind is one **vendor** exposing **two capabilities**; it appears as one entry in each registry (document + media), the same way Qwen/MinerU appear across multiple provider registries. Not one-folder-per-vendor.
- Shared submit→poll→get SDK wrapper via `@alicloud/docmind-api20220711`.
- UI is intended to stay a **single "Document Parsing" panel** (post-#837): AliDocMind is just a provider that supports more formats than MinerU, including audio/video. One credential input, mime-based dispatch at the route layer.

## Tasks

- [x] `PDFParserConfig` / `DocumentExtractorConfig` gain `accessKeyId`/`accessKeySecret`
- [x] `alidocmind` in `PDF_PROVIDERS` + `parseWithAliDocMind()` (file mode)
- [x] `lib/media-parse/` domain (types/constants/providers)
- [x] `MediaExtractorProvider` interface + media registry + `extractMedia()`
- [x] Env-gated smoke test (real PDF + real video through AliDocMind)
- [ ] Document Parsing settings panel: AliDocMind AK/SK inputs (after #837 merges)
- [ ] Mime-based dispatch in `/api/extract-document` (file → `DocumentArtifact`, media → `MediaArtifact`); route-level media entry

## Related

- RFC #621 (MAIC ETL)
- #741 (document extraction — the sibling landed by jackefn)
- #837 (Document Parsing panel rename + format registry — UI work builds on this)

yanpgwang:
## Summary

Adds the **media (audio/video) extraction** stage of MAIC ETL and **AliDocMind** (Aliyun Document Mind, LLM version) as a new extraction vendor covering both the document and media paths.

Sibling to the document extraction landed in #741. Document extraction → `DocumentArtifact` (page/block anchored); media extraction → `MediaArtifact` (timestamp anchored).

Closes #886. Part of #621. Builds on #837 (Document Parsing panel + format registry).

## What's in this PR

**Document side — AliDocMind as a new provider:**
- `alidocmind` entry in `PDF_PROVIDERS`; selectable anywhere MinerU is, flows through the existing document extractor registry. Supports pdf/docx/pptx/xlsx + images.
- `parseWithAliDocMind()` maps `layouts[]` → `ParsedPdfContent`. Table/chart content is read from `llmResult`; figure/picture images are downloaded from their (short-lived) OSS URLs and inlined as base64 — parity with unpdf/MinerU, so downstream `storeImages` → slide works unchanged.
- AK/SK credentials (not a single apiKey), with an opt-in env fallback gated to server-managed use.

**Media side — new domain:**
- `lib/media-parse/` mirrors `lib/pdf/`. `parseMedia()` maps `segments[]`/`audio_frames`/`video_frames` → `MediaArtifact` (transcript + keyframes).
- `MediaExtractorProvider` + media registry + `extractMedia()`, symmetric to the document side.

**Single Document Parsing panel (reuses upload + generation path):**
- AliDocMind appears in the post-#837 panel as a peer of unpdf/MinerU. One credential entry (AK/SK), Test Connection, format badges (incl. audio/video).
- Audio/video uploads dispatch through the same `/api/extract-document` route → `extractMedia()`, flattened to the text shape the generation pipeline already consumes. No separate upload area, no downstream changes.

**Shared:**
- `lib/pdf/alidocmind-client.ts` — submit → poll → get flow via `@alicloud/docmind-api20220711`.

## Hardening (from an adversarial self-review)

- **SSRF**: the media branch runs the same `validateUrlForSSRF` on a client `baseUrl` as the document branch.
- **Credential verify**: `verifyAliDocMindCredentials` whitelists success signals (probed against the real API: valid creds + bogus job → no-throw `BizIdNotExist`; invalid → thrown `InvalidAccessKeyId.NotFound`), so an unreachable endpoint / localized error / throttling reports failure instead of a false "connection successful".
- **Env-fallback gate**: env creds are opt-in (`allowEnvFallback`), enabled only for a server-managed provider — an unauthenticated client can't silently run on the server account.
- **Error handling**: `getDocParserResult` throws on a non-200 result envelope instead of returning empty text as success.
- **Media pagination**: stops after the first page when `segments[]` are present (layoutNum/layoutStepSize address layout blocks, not segments).
- **pageCount**: AliDocMind `pageNum`/`pageCountEstimate` are 0-based (verified); pageCount no longer undercounts by one.

## Verification

- `tsc --noEmit` clean; `pnpm lint` 0 errors; `pnpm test` + package tests pass; CI green.
- Env-gated smoke test (`tests/document/alidocmind.smoke.test.ts`, not in CI) run with a real key:
  - PDF (6 MB) → text + **24 images inlined as base64** in `images[]`/`metadata.pdfImages`; no expiring OSS URLs left in text.
  - Video (Sintel trailer) → `MediaArtifact` with transcript/keyframes; `durationMs` asserted to be in ms. (Content is non-deterministic server-side — see the test header note.)
- Settings UI verified via Playwright: AliDocMind panel renders AK/SK inputs + format badges; Test Connection returns success through the real Aliyun API.

## Follow-ups (not blocking)

- URL-passthrough for images (skip the download) would need the downstream generation pipeline to accept URL assets — a separate refactor once that side supports it.

@wyuc — this is the media half of the extraction split (RFC #621); would appreciate your review, especially the artifact boundary (`MediaArtifact` vs `DocumentArtifact`) and the single-panel/route reuse.


wyuc (changes requested):
## Cross-review + real E2E — requesting changes

Reviewed head: `d0a47372d2fc31c72dbe0e8d2e3f0b957d6123bc` (37 files). Claude and Codex independently reviewed the same 2,708-line GitHub diff in report-only/read-only mode; neither saw the E2E result or the other review. I then merged the findings by consensus and validated the single-reviewer items against the code.

### Real E2E result

The extraction main path works at this head:

- **PDF:** HTTP 200; 5,229 text characters; 14 pages; 24/24 images inlined as base64; no temporary OSS URL remained in the emitted text.
- **MP4:** HTTP 200; 2,752 text characters; transcript, keyframes and synopsis present.
- The run used the real browser upload flow, real `/api/extract-document`, and real AliDocMind. Downstream scene generation was intentionally intercepted, so this proves extraction E2E, not full course-generation E2E.

However, the negative credential case is blocking: an otherwise valid AK/SK with OSS access but **without DocMind permission** was shown as “Connection successful”; real PDF/video extraction then returned `NoPermission`.

### P1 — blocking

1. **Body-level auth failures are green-lit** — `lib/pdf/alidocmind-client.ts:253-257` returns `{ ok: true }` for every non-throwing SDK response and never inspects `res.body.code/message`. This exactly explains the E2E `NoPermission` false positive. Only an explicit job-not-found/expired probe response should be accepted; `NoPermission` must fail. Add a deterministic `no throw + NoPermission` test.

2. **Verifier endpoint crosses the credential/endpoint trust boundary** — `app/api/verify-pdf-provider/route.ts:33-50` lets an unmanaged request fall through to server env AK/SK and passes client `baseUrl` without the SSRF guard; even a managed request does not ignore the client endpoint. That can issue a server-signed request to an attacker-controlled/private host. For managed providers, use only server-owned credentials/default endpoint. For unmanaged providers, never fall back to env credentials and validate the endpoint before constructing the SDK client.

3. **Provider-returned image URLs create a second-order SSRF / memory-DoS path** — `lib/pdf/pdf-providers.ts:376-385,464-466` fetches every returned URL with automatic redirects, no SSRF validation, no byte cap, and unbounded `Promise.all`. A client-selected public custom endpoint can return internal-network or oversized image URLs. Validate each hop/URL, cap downloaded bytes and image count/concurrency, or restrict to trusted Aliyun OSS hosts.

4. **User-managed AliDocMind cannot be selected in the normal generation UI** — `components/generation/generation-toolbar.tsx:245-250` considers a provider available only via `cfg.apiKey`/server config. AliDocMind stores `accessKeyId` + `accessKeySecret`, so valid user credentials still leave the selector disabled. Use a provider-specific/shared readiness predicate that recognizes the AK/SK pair.

### P2 — correctness and integration

5. **Server-managed/env-only config is internally inconsistent** — `lib/server/provider-config.ts:98-102` has no AliDocMind PDF env/config mapping, while extraction allows env fallback only when `isServerConfiguredProvider()` is true (`app/api/extract-document/route.ts:144-168,254-258`). Verification can use env credentials while extraction rejects them. Add an explicit server-config representation for the AK/SK pair and make verify/extract resolve it identically.

6. **Polling ignores business error codes** — `lib/pdf/alidocmind-client.ts:141-165` reads only `data.status`. A body-level non-200 response without a status can retry for the full 15 minutes. Check `body.code/message` before polling state, as already done for result retrieval.

7. **All extracted images are attributed to page 1** — `lib/pdf/pdf-providers.ts:442-472` discards each layout's `pageNum` while collecting URLs, then hard-codes `pageNumber: 1`. Downstream prompts identify images by source page, so multi-page documents receive incorrect image associations. Preserve `{url, pageNum}` through fetch/filtering.

8. **Empty media extraction is returned as success** — `app/api/extract-document/route.ts:172-184` turns an artifact with no synopsis/transcript/keyframes into `text: ''` and HTTP 200. Return a parse/provider error instead of silently generating from nothing.

9. **The advertised format matrix exceeds the provider contract** — `lib/document/mime.ts:165-170,187-196` advertises WebP, JP2 and M4A, while the official AliDocMind format list supports JPG/JPEG/PNG/BMP/GIF and MP4/MKV/AVI/MOV/WMV/MP3/WAV/AAC. Remove unsupported entries or add a tested conversion layer: https://help.aliyun.com/zh/document-mind/developer-reference/document-parsing-large-model-version

10. **Invalid provider/media combinations become opaque 500s** — `app/api/extract-document/route.ts:142-170` forwards a document-only `providerId` such as `unpdf` into the media registry for an MP4 upload. Validate provider capability first and return a clear 4xx.

### P3 / follow-up

- Upload help copy still lists only documents/slides/spreadsheets/images/plain text even though audio/video are accepted.
- `formatTimestamp()` emits `[75:03]` for a 75-minute video; use `HH:MM:SS` once duration exceeds an hour.
- Credential failures should not be classified as `INTERNAL_ERROR` 500.
- Add deterministic fixtures for layout mapping, segment mapping, pagination, empty media, body-level error envelopes, and page/image association. The current real E2E covers the successful PDF/MP4 extraction paths but not these edge cases.

Please rerun the credential negative case plus one real PDF and MP4 extraction after the blockers are fixed.

yanpgwang:
Thanks for the thorough cross-review + real E2E — every finding was accurate. All P1/P2 are fixed in d6fcb58, plus the two quick P3 items.

### P1 (blocking)
1. **Body-level auth green-lit** — `verifyAliDocMindCredentials` now inspects the no-throw `body.code`. Probed the real API: a DocMind-capable key returns `BizIdNotExistOrResultExpired`, an OSS-only key returns `NoPermission` (both no-throw). Only a success/`200` or the job-not-found code is accepted; everything else (incl. `NoPermission`) fails. Added a deterministic mocked test covering exactly the no-throw+`NoPermission` case (`tests/document/alidocmind-verify.test.ts`).
2. **Verifier trust boundary** — managed now uses server-owned AK/SK + default endpoint only (client values ignored); unmanaged uses client creds only with **no env fallback**, and the client endpoint is SSRF-validated before the SDK client is built.
3. **Image-fetch SSRF / DoS** — restricted to Aliyun OSS hosts, `redirect: 'error'`, per-image byte cap (10 MB), image-count cap (200), and bounded concurrency (6) instead of unbounded `Promise.all`.
4. **Not selectable in generation UI** — availability predicate now recognizes the AK/SK pair, not just `apiKey`.

### P2
5. **Server-config inconsistency** — added an explicit AK/SK server-config representation (`applyAliDocMindFallback` + `resolveManagedAliDocMindCredentials`); verify and extract resolve managed/env identically now.
6. **Polling ignores business codes** — poll loop checks `body.code` before `status`, so a body-level error fails fast instead of polling 15 min.
7. **All images page 1** — `{url, pageNum}` is threaded through fetch/filter; `pdfImages[].pageNumber` reflects the source page.
8. **Empty media = success** — an artifact with no synopsis/transcript/keyframes returns `422 PARSE_FAILED` instead of 200 + empty text.
9. **Format matrix** — trimmed to the official contract: images JPG/JPEG/PNG/BMP/GIF (dropped WebP/JP2), media MP4/MKV/AVI/MOV/WMV/MP3/WAV/AAC (dropped M4A). Added a regression assertion.
10. **Opaque 500 on provider/media mismatch** — a document-only provider + media upload now returns a clear 4xx.

### P3
- Credential-verify failures now return `INVALID_CREDENTIALS` 4xx (not `INTERNAL_ERROR` 500).
- `formatTimestamp` emits `HH:MM:SS` past one hour.
- Deterministic fixtures: added the verify body-code test; layout/segment/pagination/empty-media fixtures still open — noted as follow-up.
- Upload help copy (audio/video) — follow-up (i18n copy across 8 locales).

### Re-run (real key, head d6fcb58)
- verify: valid DocMind key → `{ok:true}`; invalid AK/SK → `{ok:false, InvalidAccessKeyId.NotFound}`.
- PDF: 200; text + 24 images inlined as base64 with correct per-page association; no OSS URLs left in text.
- MP4: 200; transcript + keyframes + synopsis.

I couldn't reproduce the exact OSS-only-key negative against the live API (I only have a DocMind-capable key), but the mocked unit test pins that classification deterministically. If you can re-run your OSS-only-key case against this head that'd be the ideal confirmation.

yanpgwang:
Re-tested end-to-end at head `02b9de6` with a real key, covering the three cases you asked for — all pass.

### Credential negative case ✅
A DocMind-**restricted** AK/SK (OSS access, no DocMind permission) now correctly fails verification:
```
POST /api/verify-pdf-provider → 400
verifyAliDocMindCredentials → { ok: false, error: "NoPermission: You are not authorized to perform this operation." }
```
This is the exact case that previously showed "connection successful" then failed at extraction. The fix inspects the no-throw `body.code` and only accepts a success/job-not-found probe response. (A DocMind-capable key still verifies `200`.)

### Real extraction ✅
- **PDF**: `POST /api/extract-document 200` (23.6s) → full generation chain (outlines → scene-content → actions → tts) all 200. 24 images inlined as base64 with correct per-page association; no OSS URLs left in text.
- **MP4**: job `success` → extraction 200 → full generation chain 200.
- **MP3** (audio track extracted from the sample video, real speech): `POST /api/extract-document 200` (77s) → transcript flattened to text → full generation chain 200.

No `NoPermission` / `PARSE_FAILED` / sharp errors in the logs on the happy paths.

### Also in this push (found while manual-testing)
- `ensureBuiltInPDFProviders`: backfill AliDocMind into persisted settings on rehydrate, so it's selectable and picks up the server-configured flag (previously a stale persisted store hid it).
- Analysis step label: audio/video no longer render as "解析 document 文件" — documents keep their precise format token, media uses a dedicated localized string (8 locales).

All P1/P2 from your review are addressed (commit `d6fcb58`), plus these two UX fixes (`02b9de6`). Ready for another look.

wyuc (changes requested):
Reviewed at `02b9de6c992789eb6308247b0dbc2806cac46884`.

Most findings from the previous review have been addressed. Focused validation on this head passed:

- AliDocMind verifier / extractor registry / MIME tests: 28/28 passed
- Relevant ESLint checks passed
- i18n key validation passed for all 8 locales
- GitHub `Lint, Typecheck & Unit Tests` and `E2E Tests` checks are green

There are still three issues to resolve before approval:

1. **[P1] YAML-managed AliDocMind credentials can verify successfully but fail during extraction.**

   `app/api/verify-pdf-provider/route.ts` resolves server-managed credentials through `resolveManagedAliDocMindCredentials()`, including YAML configuration. However, both extraction paths in `app/api/extract-document/route.ts` (around lines 177-184 and 272-284) clear the managed AK/SK and enable environment fallback. The lower-level fallback only reads `ALIDOCMIND_*` environment variables. A deployment configured only through server YAML can therefore pass verification and then fail when extracting a PDF or media file.

   Please use the shared managed-credential resolver for both document and media extraction, and add a regression test for YAML-only managed credentials with no `ALIDOCMIND_*` environment variables.

2. **[P1] The 10 MB image limit is checked after the complete response has already been buffered.**

   In `lib/pdf/pdf-providers.ts` (around lines 413-420), the code checks `Content-Length`, then calls `res.arrayBuffer()`, and only afterward checks the actual byte length. If `Content-Length` is absent or false, an oversized remote response is fully loaded into memory before rejection, so the limit does not protect the server from memory exhaustion.

   Please read the response as a stream, maintain a cumulative byte count, and cancel/abort as soon as the limit is exceeded. Restricting this path to HTTPS and a precise expected OSS host allowlist would also strengthen the boundary.

3. **[P2] An empty verification response is still treated as positive authorization.**

   `lib/pdf/alidocmind-client.ts` (around lines 274-276) includes `codeStr === ''` in the success whitelist. A malformed or custom endpoint returning an empty body can therefore be reported as a successful credential verification, contrary to the verifier's positive-signal requirement.

   Please remove the empty-code success case and add an empty-body regression test.

There is also a current merge blocker: this head conflicts with `main` in `package.json`. When resolving it, please retain the PR's AliCloud dependencies together with the current `main` dependencies (including `@openmaic/storage` and `fake-indexeddb`) without duplicating or relocating existing package entries incorrectly.

@sophietao20-star, after the author fixes the items above and updates the branch from `main`, could you please rerun the AliDocMind extraction/full-generation E2E against the exact final head and report the tested SHA?

Please cover:

1. A DocMind-capable AK/SK verifies successfully (`200`).
2. An OSS-only AK/SK without DocMind permission is rejected during verification (`400`, `NoPermission`).
3. YAML-only server-managed AK/SK, with no `ALIDOCMIND_*` environment variables, both verifies and successfully extracts at least one PDF and one MP4.
4. Full browser flows complete for PDF, MP4, and MP3: upload -> extraction -> downstream generation.
5. Extracted images retain correct page associations, no raw OSS URLs leak into generated content, and there are no related browser-console or server errors.

Please also add deterministic automated coverage for the YAML-managed extraction path, empty verification bodies, and streaming overflow rejection.

yanpgwang:
All three follow-ups fixed + rebased onto `main`. Head is now `02d4b55a4dc87bd45871081affc2a5e237447026`.

**1. [P1] YAML-managed creds reach extraction.** Both extract paths (document + media) now resolve server-owned AK/SK via the shared `resolveManagedAliDocMindCredentials()` (env **or** YAML), the same resolver the verifier uses — a YAML-only deployment no longer verifies-then-fails. Regression test added: YAML-only `pdf.alidocmind` creds with no `ALIDOCMIND_*` env resolve correctly (`tests/server/provider-config.test.ts`).

**2. [P1] Image size cap enforced while streaming.** The body is now read chunk-by-chunk with a cumulative byte counter that aborts the moment it passes the cap, so a missing/false `Content-Length` can't buffer an oversized response into memory. Host allowlist tightened to `oss-*.aliyuncs.com`. Tests: non-OSS host refused without fetching, no-`Content-Length` overflow aborts, declared-oversize rejected (`tests/document/alidocmind-image-fetch.test.ts`).

**3. [P2] Empty verification body rejected.** Removed the `codeStr === ''` success branch — a working key always returns the job-not-found business code for the bogus probe id, so an empty/absent body is now a failure. Regression test added for empty and absent bodies.

**Merge conflict** resolved in the rebase: kept `@openmaic/storage` + `fake-indexeddb` (main) together with the AliCloud deps, no relocated entries.

Verification:
- New/updated deterministic tests: 60/60 pass (verify incl. empty-body, YAML-managed resolver, streaming overflow).
- Full `tests/document` + `tests/server`: 216 pass.
- `tsc`, `pnpm check`, `pnpm lint` (0 errors), i18n all clean.
- Real key: PDF + MP4 extraction still pass end-to-end after the managed-creds refactor.

@wyuc / @sophietao20-star — ready for the final E2E pass against `02d4b55a4dc87bd45871081affc2a5e237447026`. The YAML-only managed path (item 1) now has automated coverage; if you can still run it against a live YAML-configured deployment that'd be the ideal confirmation.

wyuc (changes requested):
Reviewed at `02d4b55a4dc87bd45871081affc2a5e237447026`.

The streaming image-size cap and empty verification-response fixes look correct. I also ran the focused AliDocMind/provider tests locally: 82/82 passed. Both GitHub CI jobs are green, and `main` is now an ancestor of this head.

One managed-configuration blocker remains:

**[P1] YAML AliDocMind credentials are still dropped when the entry also specifies `baseUrl`.**

`loadEnvSection()` treats every PDF provider with a YAML `baseUrl` as configured and creates `pdfConfig.alidocmind`, but that generic entry copies only `apiKey`, `baseUrl`, `models`, and `proxy` (`lib/server/provider-config.ts`, around lines 185-199). It does not preserve `accessKeyId` or `accessKeySecret`. `applyAliDocMindFallback()` then returns immediately when that entry already exists (around line 291), so it never adds the AK/SK.

I reproduced this with a YAML entry containing all three fields:

```yaml
pdf:
  alidocmind:
    accessKeyId: review-ak
    accessKeySecret: review-sk
    baseUrl: https://docmind-api.cn-hangzhou.aliyuncs.com
```

The resulting state is `managed: true`, while `resolveManagedAliDocMindCredentials()` returns `undefined`. Verification and extraction therefore both lose the configured credentials. The newly added regression test does not catch this because its YAML fixture omits `baseUrl`, causing the generic loader to skip the entry and allowing the fallback to run.

Please merge the AliDocMind AK/SK into an existing generic entry instead of returning early, and add a regression case for YAML `baseUrl + accessKeyId + accessKeySecret` together.

@sophietao20-star, after this is fixed, please run the requested live AliDocMind E2E against the exact final SHA and report the tested SHA. GitHub's automated E2E is green, but the final-head live credential/extraction/full-generation verification is still outstanding.

yanpgwang:
Fixed. Head is now `3c76ab9ded314dc3b6c7f9663b85ca0866c01015`.

**[P1] YAML AK/SK dropped when the entry also has `baseUrl`** — spot on, and the root cause was exactly as you described: `loadEnvSection()` creates the `pdf.alidocmind` entry from the `baseUrl` (pdf requires one) but copies only apiKey/baseUrl/models/proxy, then `applyAliDocMindFallback()` returned early on the existing entry and never added AK/SK.

Fix: `applyAliDocMindFallback()` now **merges** the AK/SK into any existing entry (preserving its baseUrl/models/proxy) instead of returning early. Regression test added with all three fields together:

```yaml
pdf:
  alidocmind:
    accessKeyId: review-ak
    accessKeySecret: review-sk
    baseUrl: https://docmind-api.cn-hangzhou.aliyuncs.com
```
→ `isServerConfiguredProvider('pdf','alidocmind') === true` **and** `resolveManagedAliDocMindCredentials()` returns all three. (Confirmed the test fails against the old early-return and passes now.)

Verification: `tests/server` + `tests/document` 217 passed; `tsc` / `pnpm check` / `pnpm lint` clean.

@sophietao20-star — ready for the live AliDocMind E2E against `3c76ab9ded314dc3b6c7f9663b85ca0866c01015` whenever convenient.

wyuc:
@sophietao20-star I have completed the code review on current head `3c76ab9ded314dc3b6c7f9663b85ca0866c01015`; the last code blocker is fixed and the focused regressions pass locally.

Please run the live AliDocMind E2E against the exact final SHA, covering the previously requested credential positive/negative cases, YAML-managed extraction, and the PDF/MP4/MP3 upload -> extraction -> full-generation flows. Please report the tested SHA and results on this PR.

If the live E2E passes and the required GitHub checks are green, you may directly **Approve** and **merge this PR**. If the head changes before testing completes, please rerun against the new final SHA before approving.

yanpgwang:
Heads up — the branch advanced past the head you last reviewed. New head is `af0484c9f8e040af84e7d66bd227d792021e7922`.

After your YAML+baseUrl fix landed, I ran an additional self-review pass (an automated adversarial sweep over the credential/config path, since that area had produced several edge cases). It surfaced two more, both now fixed with tests. Neither changes previously-reviewed behavior — they extend it into cases your review didn't cover:

1. **[MED] `baseUrl`-only YAML entry with no AK/SK** — the symmetric case of the bug you caught. A `pdf.alidocmind` entry with `baseUrl` but no `accessKeyId`/`accessKeySecret` (and no env AK/SK) made the generic loader create the entry → `isServerConfigured=true` (managed) → but `resolveManagedAliDocMindCredentials()` returned `undefined`, so the provider was locked out **and** any client-entered AK/SK were silently dropped. `applyAliDocMindFallback()` now deletes a credential-less entry so the provider stays **unmanaged** (clients supply their own creds). Your `baseUrl + AK/SK` case (fixed in the prior commit) is unaffected. Regression test added.

2. **[LOW] verify/poll asymmetry on a `code: "success"` status body** — verify accepted `"success"`/`"200"`, but the extraction poll loop threw on any non-`"200"` code, so a success-shaped status would pass verification then fail extraction. The poll loop now treats `"200"`/`"success"` as benign, matching the verifier. (Latent — no evidence the status endpoint emits `"success"` today.)

Verification: `tests/server` + `tests/document` 218 pass; `tsc`/`pnpm check`/`pnpm lint` clean; real-key PDF + MP4 extraction still pass end-to-end.

@sophietao20-star — please target `af0484c` for the live E2E rather than the earlier SHA.

yanpgwang:
**Holding this PR pending #844 (Document Bundles M3).**

#844 and this branch touch a large overlapping set of core files, so merging them independently would produce nontrivial (logic-level, not just textual) conflicts. Overlapping files:

- `app/api/extract-document/route.ts` (both restructure the route)
- `lib/pdf/pdf-providers.ts`, `lib/document/extractors/pdf.ts`, `lib/document/index.ts`
- `app/page.tsx`, `app/generation-preview/{page,types}.tsx`, `components/generation/generation-toolbar.tsx`
- all 8 `lib/i18n/locales/*.json`

Plan: land #844 first, then rebase this branch onto the updated `main` and resolve the conflicts here. The AliDocMind review is essentially complete (all P1/P2 resolved, latest self-review edge cases fixed at `af0484c`), so the remaining work is purely the post-#844 rebase — no outstanding functional changes on this side.

@sophietao20-star — suggest deferring the live E2E until after the rebase, so it runs against the final conflict-resolved head rather than a SHA that will change.

wyuc:
@yanpgwang #844 has now been merged into `main`. Please rebase this PR onto the latest `main`, resolve the overlapping route/provider/UI/i18n changes carefully, and push the conflict-resolved final head. Once updated, please report the exact final SHA and rerun the required GitHub checks.

@sophietao20-star, after the rebase is complete and CI is green, please run the previously requested live AliDocMind E2E against that exact SHA and report the results here. I will then review the post-#844 conflict-resolution diff and update my review status.

yanpgwang:
Rebased onto `main` now that #844 (Document Bundles M3) is merged. New head: `3fead05c8ea1dd79bf5d1b36203ae88daae128cd`. No conflicts remain; the branch is green again.

Conflicts were in 3 files, all resolved keeping both sides' intent:
- **`lib/document/extractors/pdf.ts`** — combined #844's self-host MinerU routing (all types, incl. pdf, go through `parseWithMinerUDocument`) with the AliDocMind branch (single `parsePDF` flow for pdf/docx/pptx/xlsx/images).
- **`app/generation-preview/page.tsx`** — kept #844's `source`-based multi-document upload structure and added the AK/SK FormData fields onto its resolved `providerConfig`.
- **`app/generation-preview/types.ts`** — #844 dropped the per-type analysis label (now generic "Analyzing documents"); kept that, and retained the audio/video branch so media uploads still read "Analyzing audio/video" rather than "…documents".

Verification on `3fead05c8ea1dd79bf5d1b36203ae88daae128cd`:
- `tests/document` + `tests/server`: 232 pass; `tsc` / `pnpm check` / `pnpm lint` clean; i18n aligned.
- Real key: PDF + MP4 extraction still pass end-to-end.

@sophietao20-star — this is the conflict-resolved final head; good to target for the live AliDocMind E2E now.

sophietao20-star (commented):
Completed the live AliDocMind E2E review, no blocking issues was found.

Verified:

- A DocMind-capable AK/SK verifies successfully (200).
- An OSS-only AK/SK without DocMind permission is rejected during verification (400, NoPermission).
- YAML-only server-managed AliDocMind works with no `ALIDOCMIND_*` environment variables set: Settings no longer asks for client-side credentials and shows`此提供方由服务端托管，无法在此修改。如需使用自己的凭据，请新增一个提供方。`; a full PDF and a full MP4 flow completed successfully via the server-managed configuration.
- Full browser flows complete for PDF, MP4, and MP3: upload -> extraction -> downstream generation.
- the large media file (50 MB MP4) parsed correctly, downstream classroom content was generated, and the processing time was acceptable.
- Focused runtime coverage passes for AliDocMind provider config, verification, MIME/provider routing, media/document artifacts, extraction route handling, and settings server sync.

Minor UI/UX issues:

- Retry after generation failure preserves the prompt history but clears the uploaded file; product-wise, the file should likely be retained to align with the prompt-retention behavior.
- Secret visibility toggle uses the opposite state icon: the secret is visible while the slashed-eye icon is shown, screenshot as below:

Image

- The `NoPermission` verification error could be simplified into a more actionable user-facing message instead of exposing low-level provider details, now is as below:

Image

wyuc (approved):
Reviewed the post-#844 rebase at `3fead05c8ea1dd79bf5d1b36203ae88daae128cd`. The conflict resolutions preserve both Document Bundles and AliDocMind/media extraction behavior; focused CI is green, and the live PDF/MP4/MP3 plus credential/YAML-managed E2E passed on this exact head. The remaining UI/UX notes are non-blocking follow-ups. Approved.

PR #888, opened 2026-07-09
feat(renderer/editing): marquee multi-select + multi-element drag (#851 Part A)

wyuc:
Continues #851 Part A after #859 (selection + drag) and #881 (resize + rotate): the multi-select half of the Part A selection model. Merges inert — nothing in the app imports `@openmaic/renderer/editing` yet.

## What

- **Marquee (rubber-band) selection** — `core/marquee.ts` normalizes the start/current pointer pair into one min/max rect (collapsing the reference implementation's four quadrant branches into a single `contain`/`intersect` predicate pair), hit-tests rotation-aware AABBs, excludes locked elements, and applies all-or-nothing group cohesion. Lines are tested against an AABB that includes their `broken`/`curve`/`cubic` control points — a bent line's visible path can bow outside its start/end chord (conservative, by the Bézier convex-hull argument). No modifier → full containment; Ctrl/Shift/Meta → intersection (reference parity).
- **Blank-canvas capture surface + `MarqueeBox` chrome + `useMarqueeGesture`** — previously a blank click did nothing at all; now sub-threshold (5 canvas units, both axes) clears the selection and a past-threshold drag replaces it. Editable mounts only (select-only mounts keep native touch panning), main button only, cancel-safe (clear-on-up), and release decisions read the live selection rather than the pointer-down closure.
- **Click modifier table, one resolver for boxes AND lines** — `core/selection.ts`: outside the selection a plain click selects (a grouped element expands to its whole group), a modifier click adds; inside, a modifier click removes (guarded against emptying; whole group as the unit; `primaryId` preserved when it survives), and a plain pointer-down keeps the selection, sets the primary, and arms a multi-drag.
- **Multi-element drag** — `computeMultiDragMove`: rigid translation of the selected set; snapping computed on the union bounding box with the corrected delta applied to every member; selected elements excluded from snap candidates; locked members never move; lines are carried by their `left`/`top`. Shift held mid-drag engages the dominant-axis lock (shift at pointer-down remains a selection modifier). Commit emits ONE `element.updateMany` intent — the batch member of the `EditIntent` union, previously declared but unused — so a multi-drag is one host undo entry; single-element drags keep emitting `element.update`.
- Multi-selections still show no resize/rotate handles (group scaling is a later slice; the #881 gate is re-asserted in tests).

## Review

Cross-model review before pushing (Codex `exec review` high effort + an independent red-team pass over the selection-transition matrix, event layering, multi-drag geometry, and gesture protocol): no P1. All actionable findings were fixed in the second commit — the consensus one (locked elements translated when a host-controlled selection contains them), the curved-line marquee miss, lines bypassing the modifier table, group cohesion holding only at the marquee entry point, the touch-panning regression from a full-bleed capture surface, a right-click clearing the selection, the stale-closure release path, and the dead axis-lock plumbing. A second Codex pass over the full diff reports clean. Known deferral (documented in code): the single-pointer guarantee is per-hook; cross-hook pointer arbitration rides the existing multi-touch caveat.

## Tests & boundary

203 tests across the editing subpath (+69 in this PR); package build emits `dist/editing`; `tsc` clean; respects the #853 import boundary.

### DIFF
--- a/packages/@openmaic/renderer/src/editing/EditableSlideCanvas.tsx
+++ b/packages/@openmaic/renderer/src/editing/EditableSlideCanvas.tsx
@@ -11,6 +11,7 @@ import { LineHandles } from './handles/LineHandles';
 import { ResizeHandles } from './handles/ResizeHandles';
 import { RotateHandle } from './handles/RotateHandle';
 import { MarqueeBox } from './handles/MarqueeBox';
+import { isSelectionModifier, resolveClickSelection } from './core/selection';
 import { useEditGesture } from './useEditGesture';
 import { useLineHandleGesture } from './useLineHandleGesture';
 import { useMarqueeGesture } from './useMarqueeGesture';
@@ -115,8 +116,9 @@ export function EditableSlideCanvas(props: EditableSlideCanvasProps) {
   // Marquee (rubber-band) gesture. It arms from a blank-canvas pointer-down on
   // the capture surface below the element hit targets, tracks a live rectangle,
   // and on release REPLACES the selection with whatever it covers (or clears it
-  // on a sub-threshold blank click). Selection-only, so it is gated on
-  // `onSelectionChange`.
+  // on a sub-threshold blank click). The capture surface only mounts on
+  // editable hosts (`onElementsChange` — see the surface comment below), and
+  // the hook itself additionally requires `onSelectionChange` to publish.
   const { marqueeRect, onCanvasPointerDown } = useMarqueeGesture({
     slide,
     scale: canvasScale,
@@ -189,9 +191,13 @@ export function EditableSlideCanvas(props: EditableSlideCanvasProps) {
               while a pointer-down on empty canvas falls to this full-bleed layer
               and arms a rubber-band select. It is a sibling of the element hit
               divs (not an ancestor), so an element pointer-down never bubbles
-              into it. Selection-only, so it is gated on `onSelectionChange`;
-              without it the surface is absent and blank clicks do nothing. */}
-          {Boolean(onSelectionChange) && (
+              into it. Gated on EDITABILITY (`onElementsChange`), not on
+              `onSelectionChange`: this full-bleed `touchAction: 'none'` layer
+              would swallow native touch panning across the whole canvas, which
+              is only acceptable in a real editor. A select-only/viewer-ish
+              mount renders no surface — element tap-select still works via the
+              per-element hit targets; blank taps neither marquee nor clear. */}
+          {Boolean(onElementsChange) && (
             <div
               data-marquee-surface=""
               onPointerDown={onCanvasPointerDown}
@@ -313,34 +319,41 @@ export function EditableSlideCanvas(props: EditableSlideCanvasProps) {
                         // bypasses the `activePointerRef` multi-pointer guard.
                         // A second pointer-down on a line during an in-flight
                         // box drag can therefore change the selection mid-
-                        // gesture. This is deferred together with multi-touch/
-                        // multi-select support; single-pointer/mouse use (only
-                        // one active pointer at a time) is unaffected.
+                        // gesture. More generally, the single-pointer guarantee
+                        // is PER-HOOK: each gesture hook (move/marquee/line-
+                        // handle/resize/rotate) guards only its own active
+                        // pointer, so a second pointer can still arm a
+                        // DIFFERENT hook's gesture concurrently. Cross-hook
+                        // arbitration is deferred together with multi-touch
+                        // support; single-pointer/mouse use (only one active
+                        // pointer at a time) is unaffected.
                         //
                         // Always consume the pointer to block fall-through to
                         // an overlapped box beneath (even with no selection
                         // callback, and even when the line is locked). When a
-                        // selection callback is provided, also select the line
-                        // — on pointer-down, for parity with box elements
-                        // (which select via onElementPointerDown). A line is
-                        // selectable but NOT draggable here: no working copy is
-                        // armed and no move intent is ever emitted (line
-                        // editing deferred).
+                        // selection callback is provided, resolve the click
+                        // through the SAME modifier/selection table as box
+                        // elements ({@link resolveClickSelection}) so a
+                        // modifier click adds/removes a line from a multi-
+                        // selection too. A line is selectable but NOT draggable
+                        // here: `armDrag` is ignored, no working copy is armed,
+                        // and no move intent is ever emitted (line editing
+                        // deferred).
                         e.stopPropagation();
                         // A locked line is inert like a locked box: it blocks
                         // fall-through (stopPropagation above) but must not be
                         // selected.
                         if (el.lock) return;
-                        // Idempotent re-selection: skip the emit when this line
-                        // is already the sole primary selection, mirroring the
-                        // box `alreadySolePrimary` guard in useEditGesture — a
-                        // re-click on an already-selected line must not re-emit.
-                        const alreadySolePrimary =
-                          activeSelection.primaryId === el.id &&
-                          activeSelection.elementIds.length === 1 &&
-                          activeSelection.elementIds[0] === el.id;
-                        if (alreadySolePrimary) return;
-                        onSelectionChange?.({ elementIds: [el.id], primaryId: el.id });
+                        const { next } = resolveClickSelection({
+                          element: el,
+                          elements: slide.elements,
+                          selection: activeSelection,
+                          modifier: isSelectionModifier(e),
+                        });
+                        // `next` is null when nothing changes (e.g. a re-click
+                        // on the current primary, or a subtractive click that
+                        // would empty the selection) — no redundant re-emit.
+                        if (next) onSelectionChange?.(next);
                       }}
                       style={{ cursor: 'default', touchAction: 'none' }}
                     />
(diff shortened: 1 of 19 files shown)


packages/@openmaic/renderer/src/editing/EditableSlideCanvas.tsx
```diff
@@ -167,6 +185,25 @@ export function EditableSlideCanvas(props: EditableSlideCanvasProps) {
...
+              while a pointer-down on empty canvas falls to this full-bleed layer
+              and arms a rubber-band select. It is a sibling of the element hit
+              divs (not an ancestor), so an element pointer-down never bubbles
+              into it. Gated on EDITABILITY (`onElementsChange`), not on
+              `onSelectionChange`: this full-bleed `touchAction: 'none'` layer
+              would swallow native touch panning across the whole canvas, which
+              is only acceptable in a real editor. A select-only/viewer-ish
+              mount renders no surface — element tap-select still works via the
+              per-element hit targets; blank taps neither marquee nor clear. */}
+          {Boolean(onElementsChange) && (
```
cosarah: This makes marquee selection depend on `onElementsChange`, even though `useMarqueeGesture` itself only needs `onSelectionChange`. A select-only host can still tap individual elements, but it can no longer marquee-select or blank-click empty canvas to clear the controlled selection. That feels like the selection API is getting coupled to mutability here.

packages/@openmaic/renderer/src/editing/core/drag.ts
```diff
@@ -99,3 +99,85 @@ export function computeDragMove(input: DragInput): DragResult {
...
+  let dx = 0;
+  let dy = 0;
+  let guides: Guide[] = [];
+
+  const resolved = resolveSnapping(snapping);
+  if (resolved) {
+    const shifted = selected.map(
+      (el) => ({ ...el, left: el.left + dx0, top: el.top + dy0 }) as PPTElement,
+    );
+    const targetRange = getElementListRange(shifted);
```
cosarah: Multi-drag snapping is still using `getElementListRange()`, which bottoms out in `getElementRange()` for each selected element. For `line`, that helper only looks at `start/end`, so a bent `broken`/`curve`/`cubic` line contributes an undersized bbox to the union. That means the same bowed line can be included by marquee selection but then snap incorrectly during multi-drag. Can we share the control-point-aware line bounds between marquee and drag so these two paths stay consistent?

packages/@openmaic/renderer/src/editing/core/selection.ts
```diff
@@ -0,0 +1,118 @@
...
+      selection.primaryId !== undefined && remaining.includes(selection.primaryId)
+        ? selection.primaryId
+        : remaining[remaining.length - 1];
+    return { next: { elementIds: remaining, primaryId }, armDrag: false, dragIds: unit };
+  }
+
+  // Already selected, no modifier: keep the whole selection, re-point the
+  // primary at the clicked element. Skip the emit when it is already the
+  // primary so a plain click doesn't double-emit.
+  return {
```
cosarah, line 140: This branch assumes the incoming controlled selection is already group-cohesive. In the `inSelection && !modifier` case, though, `dragIds` is just `selection.elementIds` as-is. If a host ever restores or programmatically passes a partial group selection like `[g1]` for a `g1/g2` group, a plain click on `g1` will arm a drag for only `g1`, even though this helper documents group clicks as all-or-nothing. Should this path normalize the live selection back to the full group unit before returning?

packages/@openmaic/renderer/src/editing/EditableSlideCanvas.tsx
wyuc: Fixed in abcfb6c6. The surface now renders for any selection-capable mount — marquee and blank-click-clear are pure selection features, as you say — and only the `touchAction: 'none'` opt-out is tied to editability. Net behavior: select-only hosts get mouse/pen marquee and blank-clear with native touch panning preserved; touch-driven marquee requires an editable mount (documented on the surface). The test asserting select-only mounts had no surface is replaced by tests pinning both halves.

packages/@openmaic/renderer/src/editing/core/drag.ts
wyuc: Fixed in abcfb6c6 — the control-point-aware line bounds moved up into the editing geometry layer as `getEditingElementRange` / `getEditingElementListRange` (deliberately not touching `utils/element.ts`, which the v1 read-only surface shares), and marquee hit-testing, the multi-drag union snap box, and `buildAlignLines` candidates all read it now. Added a test where a bowed curve inside a multi-drag selection snaps by its bent bbox where the chord bbox would not.

packages/@openmaic/renderer/src/editing/core/selection.ts
wyuc: Fixed in abcfb6c6 — `resolveClickSelection` now group-closes the incoming controlled selection (every known member expands to its full group unit; unknown ids are preserved in place), so a host-restored partial group like `[g1]` plain-clicked arms a drag over `[g1, g2]`. The no-op emit optimization survives only when the incoming selection was already cohesive and the primary already set; the never-splits-a-group invariant is stated on the resolver and pinned by tests.

packages/@openmaic/renderer/src/editing/core/snapping.ts
```diff
@@ -39,7 +44,7 @@ export function buildAlignLines(
 
   if (opts.toElements) {
     for (const el of others) {
-      const { minX, maxX, minY, maxY } = getElementRange(el);
+      const { minX, maxX, minY, maxY } = getEditingElementRange(el);
```
cosarah: Using the conservative control-point hull here fixes missed bends, but it also means snapping now exposes guide lines from geometry the curve never actually renders. For example, a quadratic line with control point `y = 80` now contributes a bottom snap line at `y = 130`, even though the visible curve only reaches `y = 90`. Would it make sense to keep the conservative range for marquee / multi-drag union, but derive snapping guides from a visual-bounds helper instead?

packages/@openmaic/renderer/src/editing/EditableSlideCanvas.tsx
```diff
@@ -167,6 +185,29 @@ export function EditableSlideCanvas(props: EditableSlideCanvasProps) {
...
+              panning, while touch-driven marquee requires an editable mount. */}
+          {Boolean(onSelectionChange) && (
+            <div
+              data-marquee-surface=""
+              onPointerDown={onCanvasPointerDown}
+              style={{
+                position: 'absolute',
+                inset: 0,
+                pointerEvents: 'auto',
+                touchAction: onElementsChange ? 'none' : undefined,
```
cosarah: This only relaxes touch suppression on the blank-canvas surface. The per-element hit targets below still use `touchAction: none`, so a select-only mount only preserves native touch panning when the gesture starts on empty canvas. If the swipe starts on a box or line, the editing overlay still intercepts it.

packages/@openmaic/renderer/src/editing/core/snapping.ts
wyuc: Fixed in b80b0a94, taking your split: align-line candidates now come from a new `getVisualElementRange` — exact rendered bounds for lines (polyline elbows are real vertices; quadratic/cubic curves contribute their Bézier axis extrema from the derivative roots), so no guide is ever emitted for geometry the curve doesn't render. Marquee hit-testing and the multi-drag union box keep the conservative hull; the split is documented at the helpers (hull answers "could this be here" — selection must never miss; visual answers "what the user sees aligns"). Your y=80 control-point example is pinned in a test: the candidate sits at the Bézier extremum, not at 130.

packages/@openmaic/renderer/src/editing/EditableSlideCanvas.tsx
wyuc: Fixed in b80b0a94 — the same editability condition now gates `touchAction` on every editing hit layer: box hit targets, locked-element blockers, and the SVG line stroke paths, not just the blank surface. On a select-only mount a swipe starting anywhere (blank, box, or line) pans natively, and tap-to-select still works; suppression applies only on editable mounts. Pinned for both a box and a line in the canvas tests.

packages/@openmaic/renderer/src/editing/core/geometry.ts
```diff
@@ -31,6 +31,144 @@ export interface ElementRange {
...
+
+  const xs = [el.start[0], el.end[0]];
+  const ys = [el.start[1], el.end[1]];
+  const start: Point = el.start;
+  const end: Point = el.end;
+
+  if (el.broken) {
+    xs.push(el.broken[0]);
+    ys.push(el.broken[1]);
+  } else if (el.broken2) {
```
cosarah, line 152: I think the `broken2` branch is still overestimating the visible bounds. The renderer does not actually draw through `(broken2.x, broken2.y)`: in `getLineElementPath()`, `broken2` becomes an orthogonal polyline through either `(broken2.x, startY/endY)` or `(startX/endX, broken2.y)`, depending on orientation. For example, `start=[0,0], end=[120,20], broken2=[80,140]` renders with `maxY = 70`, but this helper returns `maxY = 190`. Since snapping now uses `getVisualElementRange()`, that would still expose invisible snap guides for `broken2` lines. Can we mirror the actual elbow coordinates from the renderer path construction here?

wyuc: Fixed in 43a01a91 — the `broken2` visual branch now mirrors `getLineElementPath`'s elbow construction exactly (documented to stay in lockstep with the renderer), instead of including the raw handle point. Your example checked out precisely: the path renders `M0,0 L80,0 L80,20 L120,20`, so visual bounds give maxY=70 where the old branch reported 190; that case is pinned as a regression test against both the path string and the AABB. The raw point still participates only in the conservative hull (marquee/union), where never-missing is the goal.

cosarah (commented):
Reviewed the latest delta (b80b0a94..43a01a91). No new issues found in this update. The broken2 visual bounds now track the rendered orthogonal elbows instead of the control handle, and the targeted renderer suite passed locally (19 files / 220 tests).

cosarah (approved):
LGTM

PR #868, opened 2026-07-07
feat(choreography): scaffold @openmaic/choreography — shared orchestration spec (#863)

cosarah:
## Summary

Introduces `@openmaic/choreography`, a small **pure-TS spec package** that both app playback and the (future) classroom video exporter consume, so orchestration semantics — timing, the index→time action timeline, and animation descriptors — have exactly **one home**. Neither side owns the semantics; both interpret the same spec, so the exported video cannot silently drift from real playback when the app is tuned.

The first step is a **behavior-neutral refactor**: the timing constants/formulas that lived in the app engines are moved into the package, the app imports them, and the local literals are deleted. Removing the duplicated literals is what makes drift structurally impossible.

## Related Issues

Closes #863
Sub-issue of #854 (Hyperframes-based classroom video export). Independent of #861/#862 — see note under Changes.

## Changes

- **New package `packages/@openmaic/choreography`** (mirrors `@openmaic/dsl`): pure TS, single runtime dep `@openmaic/dsl`. No React / DOM / GSAP / framer-motion — enforced by a DOM-free `tsconfig` `lib`, an eslint `@/` import-boundary block, and isolated build, so "runs in a pure Node environment" is verifiable on the dependency graph.
- **`timing.ts`** — timing constants (`EFFECT_AUTO_CLEAR_MS`, `DISCUSSION_TRIGGER_DELAY_MS`, the `WB_*`/widget durations, `MAX_VIDEO_WAIT_MS`, `wbDrawCodeMs`/`wbClearMs`) + the deterministic no-audio `estimateSpeechDurationMs`, moved **verbatim** from the app engines.
- **`cursor.ts`** — `resolvePlaybackCursor` + `EMPTY_SCENE_DWELL`, moved from `lib/playback/engine-cursor.ts` (typed on `SceneCore` so an app-widened `Scene` assigns without casting).
- **`timeline.ts`** — new pure `resolveActionTimeline`, the index-domain → time-domain expansion (blocking cursor-advance vs fire-and-forget visual duration), keyed off the DSL action categories. Takes an **optional** audio-duration resolver and falls back to the estimate, so it does **not** depend on the #861/#862 audio-duration work.
- **`descriptors/`** — versioned, schema-validated animation descriptors `spotlight.v1` + `laser.v1` (declarative: property / from / to / duration / easing — no implementation); JSON Schema generated at build time.
[...]

Issue #889, opened 2026-07-09
[Task] Migrate spotlight/laser overlays to read choreography descriptors (de-dup animation values)


cosarah:
## Context

Follow-up to #863, which introduced `lib/choreography/` as the single source of truth for orchestration semantics. That task fully de-duplicated the **timing** dimension — the constants/formulas were deleted from the app engines, which now import them from `lib/choreography/timing.ts`, so there is exactly one copy.

The **animation** dimension is only half done. #863 added versioned, schema-validated descriptors (`spotlight.v1`, `laser.v1`) in `lib/choreography/descriptors/`, but the effect components still hardcode their own animation values:

- `components/slide-renderer/Editor/SpotlightOverlay.tsx`
- `components/slide-renderer/Editor/LaserOverlay.tsx`

So right now the spotlight/laser animation values exist in **two places** — the descriptor and the component — which is exactly the drift risk #863 set out to eliminate. The descriptor currently only serves the (future) video exporter, which interprets it directly; it does not yet de-duplicate the app side.

#863 explicitly scoped this migration out as incremental / non-blocking, but left it untracked — this issue is that tracking.

## Goal

Make `SpotlightOverlay` / `LaserOverlay` **read** their animation from the descriptors, so the hardcoded values are deleted and the descriptor becomes the single source for the animation dimension too (matching what #863 achieved for timing).

## Scope

1. A small `descriptor → motion/react` adapter that resolves a descriptor's layers/tracks (property, from/to, duration, delay, easing, repeat) and geometry-/corner-derived values into the `initial` / `animate` / `exit` / `transition` props the components pass to `motion`.
2. Refactor `SpotlightOverlay` and `LaserOverlay` to build their `motion` props from `spotlight.v1` / `laser.v1` via that adapter, and **delete the local hardcoded animation literals**.
3. Resolve `{param}` placeholders (dimness, color) from descriptor `params`, overridable by the existing store options (`spotlightOptions.dimness`, `laserOptions.color`).

## Non-goals

- Adding new descriptors (`highlight`, `zoom`, etc.) — separate work.
- The video exporter's own descriptor interpreter (it implements from the descriptors directly; that lives with the exporter phase).

## Acceptance criteria

- `SpotlightOverlay` / `LaserOverlay` contain no hardcoded animation durations / easings / keyframe values — they derive them from the descriptors.
- The rendered spotlight/laser animations are visually unchanged (manual before/after check, since this is not a pure-data refactor).
[...]

PR #890, opened 2026-07-09
feat(choreography): shared orchestration spec in lib/choreography (#863)

### LINKED ISSUE #863: [Task] Shared choreography spec in lib/ (single source of truth for orchestration semantics)
## Context

Sub-issue of #854 (Hyperframes-based classroom video export architecture). Follows the audio-duration prerequisite #861.

To keep the video exporter faithful to real playback **and** decoupled from render backends, the orchestration semantics that today live in the app must become a single shared source of truth. Right now they are split across three places:

- **Timing** — constants and formulas in the playback / action engines (e.g. effect auto-clear duration, the no-audio speech-duration estimate, scene/effect timing).
- **Animation** — duration / easing / keyframes embedded inside the effect components (spotlight, laser).
- **Geometry** — element positioning (already partly pure via `findElementGeometry`).

If the exporter re-implements any of these, we get a **second source of truth**: whenever the app is tuned, the exported video silently drifts, with no mechanism to detect it.

## Goal

Introduce a small set of **pure-TS choreography modules under `lib/`** (proposed `lib/choreography/`) that both playback and the exporter consume, so orchestration semantics have exactly one home. Neither side "owns" the semantics; both interpret the same spec.

### Why `lib/` and not a package

An earlier attempt (PR #868) extracted these into a standalone `@openmaic/choreography` package. We are **not** taking that route:

- The choreography spec **co-evolves with the playback engine** — every playback timing/animation tuning touches it. A package evolves slower than the app by design (separate build, `postinstall`, version boundary), which adds friction to exactly the code that changes most often.
- The exporter/compiler will also live **in the app** (`lib/video-export/` or similar, driven from a Node route/script), not as a standalone package. So both consumers can share `lib/choreography/` via ordinary imports — no cross-package path mapping or reverse app→package dependency needed.

Packages remain the right home for genuinely stable, slow-moving surfaces (`@openmaic/dsl`, `@openmaic/exporter` for PPTX/HTML per #852); the choreography spec is not one of those yet.

## Scope

1. **Pure-TS modules under `lib/choreography/`** — no React / DOM / GSAP / framer-motion / any render-backend concept, so "the exporter runs in a pure Node environment" holds. Because these files sit under the app's DOM-enabled tsconfig, purity is **enforced by an eslint import-boundary rule** on `lib/choreography/` (no-restricted-imports blocking react/react-dom/gsap/framer-motion/dom-only APIs) rather than by a package's isolated build.
2. Contents:
   - **Timing spec** — timing constants + the deterministic no-audio speech-duration function, moved verbatim from the current engines.
   - **Animation descriptors** — versioned, declarative descriptors (e.g. `spotlight.v1`) describing *what property, from what value to what value, over how long, with what easing* — no implementation. Values match current behavior.
   - **`resolveActionTimeline`** — the index-domain → time-domain expansion (cursor advance vs fire-and-forget visual duration), formalizing current playback semantics.
3. **First step is a behavior-neutral refactor**: move the existing constants/formulas out of the app engines into `lib/choreography/` and have the engines import them, **deleting the local literals**. Removing the duplicated literals is what makes drift structurally impossible.
4. Overlay components can migrate to reading descriptors incrementally (not a blocker for the exporter, which implements from the descriptors directly).

## Non-goals

- The IR / compiler (that's the next phase).
- Any Hyperframes / render-backend concept.
- The asset-collection or emitter layers.
- Extracting choreography into a standalone package (explicitly rejected — see "Why `lib/`").

## Acceptance criteria

- Pure-TS modules under `lib/choreography/` with no React/DOM/render-backend imports, **enforced by an eslint import-boundary rule** on that directory (CI fails on violation).
- App playback imports timing/spec values from `lib/choreography/`; the previous local literals are deleted (behavior-neutral — existing playback regression passes).
- `resolveActionTimeline` has golden-case unit tests (blocking actions, fire-and-forget visual duration, effect auto-clear, scene boundaries).
- Animation descriptors are versioned and schema-validated.

Depends on #861. Blocks the IR/compiler phase.

cosarah:
## Summary

Introduces **`lib/choreography/`**, the single source of truth for the orchestration semantics a faithful classroom-video exporter needs from playback — so the app runtime and the (future) exporter interpret **one spec** instead of each re-implementing, and silently drifting from, the other.

**Why `lib/`, not a package.** An earlier attempt (#868) extracted these into a standalone `@openmaic/choreography` package; that PR is closed. These semantics **co-evolve with the playback engine** (every timing/animation tuning touches them), so a package's separate build / `postinstall` / version boundary just adds friction to the most-changed code. And the exporter/compiler will also live **in the app**, so both consumers share `lib/choreography/` via ordinary imports — no cross-package path mapping or reverse app→package dependency. Packages remain right for stable, slow-moving surfaces (`@openmaic/dsl`); this spec isn't one yet.

**Purity is still machine-enforced.** `lib/choreography/**` depends only on `@openmaic/dsl` (types + the fire-and-forget partition) and `zod`. An eslint boundary on the directory blocks (1) any `@/…` host-app path (mirrors the `@openmaic/renderer` / `@openmaic/storage` boundaries) and (2) `react` / `react-dom` / `gsap` / `framer-motion` / `motion` runtime imports — so "the exporter runs in pure Node" holds without a package build. Negative-tested: dropping a `react` or `@/…` import into the directory fails lint.

## Related Issues

Closes #863. Sub-issue of #854 (Hyperframes-based classroom video export). Supersedes the closed #868 (package approach). Follow-up #889 tracks migrating the overlay components to *read* the descriptors.

## Changes

- **`lib/choreography/timing.ts`** — timing constants (`EFFECT_AUTO_CLEAR_MS`, `DISCUSSION_TRIGGER_DELAY_MS`, the `WB_*`/`WIDGET_MS` durations, `MAX_VIDEO_WAIT_MS`, `wbDrawCodeMs`/`wbClearMs`) + the deterministic no-audio `estimateSpeechDurationMs`, moved **verbatim** from the engines.
- **`lib/choreography/cursor.ts`** — `resolvePlaybackCursor` + `EMPTY_SCENE_DWELL`, moved from `lib/playback/engine-cursor.ts` (typed on dsl `SceneCore` so an app-widened `Scene` assigns without casting).
- **`lib/choreography/timeline.ts`** — new pure `resolveActionTimeline`: the index-domain → time-domain expansion (blocking cursor-advance vs fire-and-forget visual duration), keyed off the DSL's `FIRE_AND_FORGET_ACTIONS`. Takes optional audio/video/clear-count resolvers and falls back to the estimate, so it does **not** depend on the #861 audio-duration work.
- **`lib/choreography/descriptors/`** — versioned, **zod-schema-validated** animation descriptors `spotlight.v1` + `laser.v1` (declarative: property / from / to / duration / easing — no implementation), pinned verbatim to the current `SpotlightOverlay` / `LaserOverlay`. Schema authored with zod; types inferred via `z.infer` (single source); the exporter can reuse `AnimationDescriptorSchema` to validate what it interprets.
- **Behavior-neutral engine refactor** — `lib/action/engine.ts` and `lib/playback/engine.ts` import from `lib/choreography`; the local literals are **deleted**, so the timing dimension now has exactly one copy.
- **eslint boundary** on `lib/choreography/**` (see above).

## Type of Change

## Verification

1. `pnpm --filter ... ` n/a — no new package; `npx vitest run tests/lib/choreography` → **39 tests pass**.
2. `npx tsc --noEmit` and `pnpm lint` at root — clean; the new import boundary passes and was negative-tested (a `react`/`@/…` import in the directory fails lint).
3. Engine-adjacent regression: `tests/playback/action-navigation.test.ts` + `tests/action/widget-actions.test.ts` → 28 pass.

- Ran the dev server and confirmed playback behaves identically after the engines were switched to import the moved constants / estimate / cursor (speech dwell, whiteboard/widget timings, spotlight/laser auto-clear). Home + a classroom route compile and return 200.
- `resolveActionTimeline` golden cases: blocking speech (audio vs estimate fallback, playback-speed scaling), fire-and-forget effects with `advancesCursorMs = 0`, effect auto-clear, `wb_*` durations, discussion trigger-delay dwell, empty-scene dwell, and `startMs` accumulation across scenes.
- Descriptors validate against the zod schema; `spotlight.v1` / `laser.v1` values re-checked against the **current** `SpotlightOverlay` / `LaserOverlay` source (durations, easings, dimness 0.7 / z-100, laser default `#ff0000` from `lib/store/canvas.ts`, ring pulse, fly-in/exit).
- Confirmed the engine diff only swaps literals for imports (no logic change) and that no moved literals remain in the engines.

## Known gap (tracked in #889)

This PR fully de-duplicates the **timing** dimension (literals deleted from the engines, one copy). The **animation** dimension is half done: descriptors exist and the exporter can implement from them, but `SpotlightOverlay` / `LaserOverlay` still hardcode their own animation values — they do not yet *read* the descriptors. So the spotlight/laser animation values currently live in two places. Migrating the components (a `descriptor → motion/react` adapter; not a behavior-neutral change, needs a visual before/after check) is scoped out here per #863 and tracked in **#889**.


cosarah:
## Cross-review (Claude `/code-review` + codex `exec review`)

Ran an independent two-reviewer cross-review on this PR — Claude Code's `/code-review` and OpenAI codex's `exec review --base origin/main` (gpt-5.4, high effort) each reviewed the same diff independently, then fix → re-review to convergence. **8 review rounds, 7 fix commits.** Both reviewers were present in every round (no reviewer absent). All fixes are contained in `lib/choreography/`.

### Findings fixed

| Round | Severity | Fix |
|---|---|---|
| 1 | **P1** | `resolveActionTimeline` didn't model the engine's implicit whiteboard auto-open. `ActionEngine.execute()` awaits `ensureWhiteboardOpen()` (`WB_OPEN_MS`) before the first `wb_*` mutation on a closed board, so every later segment started ~2s early. Added a synthetic `IMPLICIT_WB_OPEN` beat + a `whiteboardOpen` flag carried across scenes and toggled by `wb_open`/`wb_close`. |
| 1 | P2 | Speech real-audio duration is now divided by `playbackSpeed` (matches the live `AudioPlayer.setPlaybackRate`), keeping it in lockstep with the no-audio estimate. |
| 1 | P2 | Spotlight mask relationship made expressible: `LayerSchema` gained `role: 'content' \| 'mask'` + `maskedBy: {layerId, mode: 'subtract' \| 'intersect'}`; the cutout is a mask layer the dim layer subtracts. |
| 1 | P3 | `wb_clear` on an empty board is 0ms (engine early-returns on `elementCount === 0`), not `wbClearMs(0)`. |
| 2 | **P1** | Fire-and-forget effect lifetime is not a flat `EFFECT_AUTO_CLEAR_MS`. The engine clears effects at each scene boundary / on completion, and `scheduleEffectClear` uses one shared timer each new effect resets. Added `clampFireAndForgetLifetimes`: `durationMs = min(next-scene-boundary/completion, shared-timer deadline chained through later same-scene effects)`. |
| 2 | P2 | Spotlight `dimness` default is **0.5** (`executeSpotlight`: `dimOpacity ?? 0.5`; DSL documents 0.5), not the component's unreachable `?? 0.7` fallback. |
| 2 | P2 | Modeled the spotlight wrapper's enter/exit opacity fade. |
| 3 | P2 | Pinned the spotlight fade to Motion's default 300ms tween (was implicit → a literal consumer would pop it on/off). |
| 3 | P2 | Added the laser dot's static geometry (center-anchor `translate(-50%,-50%)`, `rounded-full` ring/core) so a literal renderer doesn't draw an offset square. |
| 4 | P2 | Skipped discussions (consumed / agent not selected) and no-op whiteboard draws (empty `wb_draw_text`, empty `wb_draw_table`) now cost 0ms — via an `isDiscussionSkipped` resolver + action-determinable checks. |
| 5 | P2 | Effect chain break at an **exact** 5s boundary (the earlier clear timer fires first): changed the chain guard from `> deadlineMs` to `>= deadlineMs`. |
| 5 | P2 | Spotlight dim layer full-screen geometry (`x/y/width/height` = 100×100 at origin) made explicit so the descriptor is self-contained. |
| 6 | P2 | `wb_edit_code` no-op (missing/non-code target, stale line refs → engine skips the delay) via an `isEditCodeNoop` resolver. |
| 7 | P2 | Descriptor layer inheritance: added `inheritsFrom: {parentId, props}` to `LayerSchema`. Laser ring/core inherit the animated `dot`'s `left/top/opacity`; the spotlight border inherits the wrapper `opacity` fade — so a literal consumer follows the nested-wrapper motion instead of leaving child layers at a static origin / an outline lingering after the dim clears. |

### Rejected as false-positive (verified against source)

- **Round 8 — "`wb_delete` stale target should be 0ms."** codex claimed `executeWbDelete` returns before its delay for a missing element. Source (`lib/action/engine.ts:712-721`) shows `deleteElement()` is a silent no-op for a missing id but the `await delay(WB_DELETE_MS)` runs **unconditionally**, so charging `WB_DELETE_MS` is faithful. Not changed.

### Deliberately out of scope

- **Empty-scene `speech` dwell → lecture bubble.** `EMPTY_SCENE_DWELL` (`{type:'speech', text:''}`) is a verbatim move from `origin/main`'s `lib/playback/engine-cursor.ts`, unchanged here; it pushes empty text, and it only affects the live app path, not the exporter's `resolveActionTimeline` (which handles the synthetic dwell correctly). Pre-existing behavior, not a regression from this move-only change.
- **KaTeX-failure / other runtime-state-dependent no-ops.** Consistent with the resolver pattern (`getVideoDurationMs` / `isDiscussionSkipped` / `isEditCodeNoop`): a low-frequency runtime-dependent edge a consumer can supply a resolver for when needed.
- **#889** — migrating the overlay components to *read* these descriptors.

### Verification

- `tests/lib/choreography`: **60 pass** (was 39). New coverage: implicit wb-open (incl. reopen after close, cross-scene carry), effect-lifetime (completion-cut / boundary-cut / full-lifetime / shared-timer extension / exact-5s break / last-action 0ms), speed-scaled audio, no-op draws/edits/skipped-discussion, mask relation, dim full-screen geometry, layer inheritance.
- Engine regression (`tests/playback/action-navigation`, `tests/action/widget-actions`): 28 pass.
- `tsc --noEmit` + `pnpm lint`: clean (eslint import boundary intact).

_Pure review + fix cycle; no behavior change to the engines beyond the round-1 literal→import extraction._

wyuc (changes requested):
Thanks for pushing this direction into `lib/choreography`; keeping the playback/exporter semantics shared in-app makes sense to me.

I found two timing issues that look blocking for faithful video export:

1. `discussion` currently contributes only `DISCUSSION_TRIGGER_DELAY_MS` to the timeline. In live playback, after that delay the `ProactiveCard` is shown in playback mode and has its own 5s countdown before it calls `onSkip()`; only then does playback continue. So unattended playback/export will start every following action 5s too early after a non-skipped discussion, and any spotlight/laser lifetime spanning that interval will also be computed too short. Please either share/model the auto-skip duration in choreography (e.g. `DISCUSSION_TRIGGER_DELAY_MS + DISCUSSION_AUTO_SKIP_MS`) or make discussion an explicit pause/unresolved boundary unless the caller supplies a policy. A regression like `spotlight -> discussion -> speech` would catch this.

2. `play_video` falls back to `0` when `getVideoDurationMs` is missing. Live playback treats `play_video` as blocking and waits until the video finishes or the 5-minute cap fires, so silently producing a zero-length segment will shift all later actions early whenever exporter metadata is missing. Please make unresolved video duration explicit: require the resolver when a timeline includes `play_video`, emit an unresolved/error segment, or expose a caller-chosen fallback policy. The important part is not silently treating an unknown blocking video as 0ms.

Non-blocking but worth tightening in this PR: the `lib/choreography` purity guard is not yet a true allowlist. It blocks `@/` and some render packages, but relative escapes such as `../store/canvas`, unrelated bare imports, and dynamic imports are still possible. Since the PR claims purity is machine-enforced, an import allowlist or isolated build/lint check for `@openmaic/dsl`, `zod`, and in-folder relatives would make that guarantee much stronger.

cosarah:
Thanks @wyuc — all three addressed in 6db69f92. Confirmed each against the runtime source before changing.

**1. `discussion` auto-skip duration (blocking) — fixed.**
You're right: in playback mode the `ProactiveCard` runs its own 5s countdown (`proactive-card.tsx` — `duration = 5000`, auto-`onSkip()` at `progress <= 0`) after the `DISCUSSION_TRIGGER_DELAY_MS` show delay, so a non-skipped discussion blocks for both. I took the "share the constant" route: added `DISCUSSION_AUTO_SKIP_MS = 5000` to the timing spec, and a non-skipped discussion now charges `DISCUSSION_TRIGGER_DELAY_MS + DISCUSSION_AUTO_SKIP_MS`. `ProactiveCard` now imports that same constant (both the countdown and the seconds label derive from it), so the card and the timeline can't drift. Your `spotlight -> discussion -> speech` case is covered — with the discussion now ~8s, `clampFireAndForgetLifetimes` extends the preceding spotlight across the full interval (it was under-computing before). A `discussion` the engine skips outright (consumed / agent not selected) still contributes 0 via `isDiscussionSkipped`.

**2. `play_video` unknown duration (blocking) — fixed, no longer silent.**
Unresolved video duration no longer silently becomes 0ms. Added an `onUnresolvedVideoDuration` policy that **defaults to `'throw'`** — an exporter that forgets to supply `getVideoDurationMs` fails loudly instead of shifting every later action early. `'cap'` (assume `MAX_VIDEO_WAIT_MS`) and `'zero'` (opt back into no-dwell explicitly) are the caller-chosen alternatives. A resolved value is still capped at `MAX_VIDEO_WAIT_MS`.

**3. Purity guard → true import allowlist (non-blocking) — done.**
Turned the `lib/choreography/**` boundary into an allowlist rather than a blocklist. On top of the existing `@/…` + render-package blocks, `no-restricted-syntax` now rejects, on the import/export source node:
- parent-escape imports/re-exports (`../store/canvas`, `export * from '../playback/engine'`),
- any bare specifier other than `@openmaic/dsl`(/subpaths) and `zod`,
- dynamic `import()` and `require()` (which would bypass the static check).

Negative-tested — all four (`../store/canvas`, a stray bare package, `export * from '../playback/…'`, `import('react')`) now fail lint; the real files pass clean. This is regex-on-AST rather than a real resolver (no `eslint-plugin-import` in the toolchain), but it closes the escape hatches you listed. Happy to migrate to `import/no-restricted-paths` if you'd prefer we pull that plugin in.

Verification: `tests/lib/choreography` 60 pass, engine regression (`action-navigation` + `widget-actions`) 28 pass, `tsc --noEmit` + `pnpm lint` clean.

wyuc (approved):
LGTM. The two blocking timing issues are addressed: unattended discussion dwell now includes the shared auto-skip interval, and unresolved play_video duration fails explicitly by default instead of silently becoming 0ms. I also verified the latest-main integration locally with the focused choreography/engine suite (88 tests), typecheck, and focused lint; the updated CI, E2E, and CodeQL checks are green. Non-blocking follow-up: the choreography import guard should eventually reject normalized '..' path segments too, since forms such as './../ can bypass the current prefix allowlist.

PR #891, opened 2026-07-09
Feat/interactive slide repair core

atiqbitstream:
## Summary

This PR adds a small, self-contained toolkit for detecting and repairing broken interactive slides: a layered validation harness that classifies a slide's LLM-authored HTML as pass/warn/fail, and a deterministic SEARCH/REPLACE applier for making targeted edits without regenerating a whole scene.

Generated interactive slides are arbitrary HTML/JS and occasionally ship subtly broken — malformed markup, a syntax error in an inline <script>, or a runtime error on load — in ways that render blank with no clear signal to the pipeline. These primitives give the project a programmatic way to (a) tell whether a slide is broken and why, and (b) apply a minimal, verifiable fix.

## Scope note (please read)

These are intentionally standalone primitives — pure functions with no coupling to app state, routing, storage, or auth. They are the reusable core of a larger "slide self-repair" capability I've been running downstream; the orchestration and any UI around them are not in this PR, because in my downstream build they were tied to a bespoke persistence/auth layer that doesn't match this project's storage abstraction. I've deliberately kept this PR to the pieces that stand on their own and are worth having independently. If you'd prefer I also wire runValidation into an existing flow (e.g. a generation-time QA check or an e2e assertion) before merging, I'm happy to , I left that out to keep the surface small and reviewable, but I don't want to contribute code with no call site if you'd rather it be integrated first.

### feat(repair) — layered validation harness lib/repair/validate.ts, validators/*, types.ts)

Runs three independent layers and combines them into one ValidationReport:

- static-html — structural markup check via html-validate, configured for widget HTML (catches unclosed/mismatched tags, duplicate ids, void-element misuse; does not flag valid-but-minimal HTML for missing doc-level boilerplate).
- lint-js — parse-only syntax check of each embedded <script> via the Function constructor (throws on invalid JS without executing it).
- headless — renders the patched HTML in headless Chromium (already a dev dependency for e2e) and collects pageerror / console.error / requestfailed. Infra failures are advisory warn), never a content fail.
### feat(repair) — deterministic SEARCH/REPLACE applier lib/repair/search-replace.ts)

Parses <<<<<<< SEARCH / ======= / >>>>>>> REPLACE blocks and applies them to a source string. Strict by design to avoid silent corruption: a zero-match SEARCH is an error, a multi-match SEARCH is an error (ambiguous — it refuses to guess), and the replacement is applied literally $ is not a regex reference). Returns a discriminated { ok, result } | { ok, error }.

## Dependencies

Adds *html-validate** ^11.5.3) for the static-HTML layer. Playwright (used by the headless layer) is already a dev dependency.

## Testing

- Unit tests for the static-HTML and JS-lint layers (pass/warn/fail classification) and the full SEARCH/REPLACE surface (single/multi block, not-found, ambiguous, literal$). All green.
- Verified against a clean install: pnpm test green, tsc --noEmit 0 errors, Prettier + ESLint clean, pnpm build compiles successfully.
- The headless layer is intentionally not exercised in the vitest unit run (it needs a browser); it is covered structurally and is safe-by-default (advisory on infra failure).
## Checklist

- [x] Code follows the project's style (Prettier + ESLint clean)
- [x] Self-reviewed
- [x] Pure, dependency-light, independently testable primitives
- [x] No new warnings (tsc / eslint / build clean)
- [ ] Wired into an existing call site — deliberately omitted; see "Scope note" above, happy to add if preferred

### DIFF
--- a/lib/repair/search-replace.ts
+++ b/lib/repair/search-replace.ts
@@ -1,0 +1,33 @@
+type Result = { ok: true; result: string } | { ok: false; error: string };
+
+const BLOCK_RE = /<<<<<<< SEARCH\n([\s\S]*?)\n=======\n([\s\S]*?)\n>>>>>>> REPLACE/g;
+
+function countOccurrences(haystack: string, needle: string): number {
+  if (needle === '') return 0;
+  let count = 0;
+  let idx = haystack.indexOf(needle);
+  while (idx !== -1) {
+    count += 1;
+    idx = haystack.indexOf(needle, idx + needle.length);
+  }
+  return count;
+}
+
+export function applySearchReplace(source: string, blocks: string): Result {
+  const matches = [...blocks.matchAll(BLOCK_RE)];
+  if (matches.length === 0) return { ok: false, error: 'no SEARCH/REPLACE blocks found' };
+  let working = source;
+  for (const m of matches) {
+    const search = m[1];
+    const replace = m[2];
+    const n = countOccurrences(working, search);
+    if (n === 0) return { ok: false, error: `SEARCH block not found:\n${search.slice(0, 120)}` };
+    if (n > 1)
+      return {
+        ok: false,
+        error: `SEARCH block ambiguous (${n} matches):\n${search.slice(0, 120)}`,
+      };
+    working = working.replace(search, () => replace);
+  }
+  return { ok: true, result: working };
+}
--- a/lib/repair/types.ts
+++ b/lib/repair/types.ts
@@ -1,0 +1,18 @@
+/**
+ * Shared types for the interactive-slide validation harness.
+ *
+ * A generated interactive slide is validated across independent layers; each
+ * returns a {@link ValidationLayer} and they are combined into a single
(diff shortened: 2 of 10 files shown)


wyuc (changes requested):
There is a concrete compatibility blocker before this standalone core can land.

The lockfile resolves `html-validate@11.5.5`, whose engine is `^22.22.0 || >=24.8.0`, while OpenMAIC declares `node >=20.9.0`. This silently narrows the supported install/runtime range for a toolkit that currently has no call site.

Please either select a dependency version compatible with the repository's declared Node range or propose the Node support change separately. Also wire the validator into one explicit supported execution boundary (CI/offline tool/server flow) and exercise the headless layer there: `runValidation` currently imports dev-only `@playwright/test`, launches a new Chromium per call, has no caller, and the headless path is not tested. That boundary is needed to judge packaging, lifecycle, and failure semantics rather than merging dormant production-library code.

wyuc:

PR #893, opened 2026-07-09
feat(pbl): runtime-event outbox + dual-write into the RuntimeStore (#869 Part C, step 2)

wyuc:
Step 2 of #869 Part C: every PBL learner-state mutation now emits a typed runtime event, and the client copies those events into the RuntimeStore — the dual-write half of the strangler. Reads stay 100% on projectV2 (adversarially verified — see below); the read flip and fold are the next step.

## Architecture: an outbox that crosses the stateless boundary

The PBL API routes are stateless (the client posts the whole project; the server mutates a copy), so the only transport-safe emission point is inside the project itself:

- **Pure ops emit** — every mutation family appends a typed `PBLRuntimeEvent` to `project.runtimeEvents` (ring-capped at 500, dedup by id): messages, submissions, evaluations, status changes across project/milestone/microtask/uiPhase, handover and task-completion gates, proficiency updates, and a `project_reset` epoch marker. Emission lives inside the pure mutators — no I/O, no storage imports.
- **Advance patches carry server-minted deltas** — events emitted server-side during `advanceMicrotask` (e.g. `handover_staged`) ride the patch; both sides mint deterministic `patch:*` ids for the same transition, so echoes collapse under id-dedup with the server authoritative. Deterministic `norm:*` ids likewise collapse repeated normalization repairs from cloned copies. Both id families lead with a **reset epoch** (the count of visible `project_reset` events) so a learner repeating the same transition after a reset is recorded, never deduped against the pre-reset event — the epoch derives from the same visible window the dedup scans, so ring eviction cannot desynchronize the two.
- **The client drains** — after each persisted project change, a bounded, never-throwing drainer copies both ledgers (`runtimeEvents` + `engagementEvents`) into the learner's `pbl` RuntimeSession, watermarked per `(stageId, sceneId, learnerKey)` through the device-scoped KV. First-session creation is race-free (deterministic session id + already-exists fallback). Corrupt watermarks degrade to a redrain and self-repair. Stage deletion clears its watermarks alongside the runtime rows.

## Semantics reviewers should weigh

- **At-least-once, by design**: an evicted watermark redrains the visible ledger; duplicates carry the same event id and downstream folds dedup by id. Cross-tab drains can duplicate but never lose.
- **Ring-overflow latent loss accepted for this step**: >500 events appended between two drains would evict undrained events. The drain fires on every persisted change, making that window near-unreachable — and projectV2 remains the source of truth throughout dual-write; the read flip will backfill from a projectV2 snapshot, which also recovers anything this window could lose.
- **Fold-baseline contract** (documented in types): folds initialize `status: 'active'` / `uiPhase: 'hero'` from design-time defaults; generation-time packaging deliberately emits no events.
- **Zero read-side change** was verified adversarially: no reader iterates `runtimeEvents` generically, no prompt builder serializes the whole project (no token bloat), `mergeSceneContentForUpdate` is untouched.

## Tests

2035 passing at root (~30 new across ops emission, drain watermark/race/corruption cases, and an end-to-end run driving real reducers into a real BrowserRuntimeStore over fake-indexeddb); typecheck, prettier, eslint clean.


wyuc:
**Cross-model review summary** (pre-submission; Claude multi-angle red-team + Codex `exec review` high effort, four rounds to closure):

- **Round 1** (fixed in 8bfa3d48): watermarks were stage-scoped — multi-scene stages and multi-learner devices would cross-contaminate (Codex); first-session creation could race and split one history across two sessions (Codex; now a deterministic session id + already-exists fallback); a corrupt watermark value threw inside the KV read *before* the shape-normalizer ran and was never rewritten — permanently blocking that scope's drain, the one true loss cell in the state-space sweep (Claude red-team); stage deletion now clears its watermarks; `resetProjectProgress` emits a `project_reset` epoch marker so folds never resurrect pre-reset submissions from the preserved ledger; normalization repairs use deterministic ids so cloned-copy echoes collapse.
- **Round 2** (ced254a3): two real event-loss holes on the server-patch path — `applyAdvanceProjectPatch` computed status-change `from` values after the milestone snapshot was already assigned (suppressing every real Instructor-advance event as from===to), and server-minted events like `handover_staged` never crossed the SSE boundary (patches now carry the operation's event delta; deterministic `patch:*` ids collapse both sides' emissions, server authoritative).
- **Round 3** (34658f33): deterministic ids gained a reset-epoch prefix — a learner repeating the same transition after a reset is recorded instead of deduped against the pre-reset event; the epoch derives from the same visible window the dedup scans, so ring eviction cannot desynchronize the two.
- **Also verified adversarially**: zero read-side change (no reader iterates `runtimeEvents` generically; no prompt builder serializes the whole project — no token bloat; `mergeSceneContentForUpdate` untouched), operation purity, and emission coverage (the three deliberate exclusions are documented in commit e5a18f4a's scope notes).
- **CI note**: the first run tripped a pre-existing store-test timer leak (`setStageAgents` debounces firing after environment teardown) that this branch's larger import graph made deterministic; fixed test-side with fake-timer hygiene in 68c82677, including two sibling suites carrying the same latent race.

Final Codex pass: no actionable findings. 2035 tests green; CI green.

cosarah (commented):
I found a few blocking issues in the runtime outbox implementation:

1. `lib/pbl/v2/operations/runtime-events.ts:35`
`runtimeEventEpoch()` counts only currently visible `project_reset` events. Once the 500-entry ring evicts an older reset marker while same-epoch `status_changed` events are still present, the next reset reuses the same `patch:/norm:` ids and `appendRuntimeEvent()` will collapse a real post-reset transition as a duplicate.

2. `lib/pbl/v2/agents/instructor.ts:1678` + `lib/pbl/v2/operations/advance-patch.ts:48`
The server-side advance delta is captured by array index (`runtimeEventStartIndex`) and later sliced from the capped ring buffer. If the ledger is near 500 and the advance emits multiple runtime events, eviction can happen before `slice(startIndex)`, so `patch.runtimeEvents` may drop some newly emitted server events.

3. `lib/pbl/v2/runtime/drain.ts:209`
The drainer appends all `runtimeEvents` first and all `engagementEvents` second. But `RuntimeRecord.seq` is documented as the sole replay ordering key, so this permanently rewrites the event order for any fold/replay that consumes `RuntimeStore`. The current end-to-end test also locks in that reordered behavior.

4. `components/scene-renderers/pbl-renderer.tsx:119` + `lib/store/stage.ts:219` + `lib/store/stage.ts:522`
The PR description says draining happens after each persisted project change, but the implementation drains immediately after `updateScene()`. `updateScene()` only mutates Zustand state and schedules a debounced save 500ms later, so the dual-write can succeed even if the source-of-truth `projectV2` save never persists.

wyuc:
All four addressed in 0bd2092e:

1. **Epoch vs ring eviction** — you're right that deriving the epoch from visible `project_reset` events breaks once an old marker is evicted while same-epoch ids remain. The epoch is now a monotonic `runtimeResetEpoch` field on the project, incremented by `resetProjectProgress` and never derived from the buffer; the `project_reset` event stays as the fold's boundary marker, it just no longer does the counting.
2. **Index-based delta capture** — advance patches now compute their event delta as an id-set difference (events whose id was not present before the operation, in ledger order), immune to cap eviction mid-operation.
3. **Cross-ledger ordering** — the drainer merges undrained events from both ledgers into a single timestamp-ordered stream (stable within each ledger, runtime-first on ties, documented) before appending, so `seq` now preserves global chronology; the end-to-end test interleaves ledgers with distinct timestamps and asserts seq order matches.
4. **Drain after persistence** — draining now subscribes to a post-save signal emitted after `saveToStorage` actually commits (new `lib/store/stage-save-signal.ts`, unsubscribed on unmount), replacing the post-`updateScene` hook — the dual-write can no longer succeed ahead of the debounced source-of-truth save, and the PR description's claim is now literally true.

Full suite 2039 green, typecheck/prettier/eslint clean; CI running on the new head.

cosarah (commented):
I still see one blocking issue in the latest patch:

1. `components/scene-renderers/pbl-renderer.tsx:63`, `components/scene-renderers/pbl-renderer.tsx:136`, `lib/store/stage-save-signal.ts:1`, `lib/store/stage.ts:393`
The new post-save drain hook is keyed only by `stageId`, and it drains `latestProjectV2Ref.current` for whichever PBL scene happens to be mounted when the save finishes. That means if the user edits PBL scene A and then navigates away before the 500ms debounced `saveToStorage()` runs, the save still persists scene A's updated `projectV2`, but the signal is either handled by scene B or by nobody. Scene A's runtime/engagement outbox never drains, so the dual-write can silently miss persisted events. The same mismatch can happen if a newer unsaved snapshot replaces `latestProjectV2Ref.current` while an older save is still in flight.

wyuc:
Fixed in 4b54eb1f. The save signal now carries the PBL scenes exactly as `saveToStorage` persisted them (`{ stageId, pblScenes: [{ sceneId, project }] }`), and the drain moved out of the component into a module-level singleton subscription (`lib/pbl/v2/runtime/drain-wiring.ts`, installed once, intentionally outliving any scene's mount). Both failure modes close: navigating away before the debounced save fires still drains the saved scene (unit-tested with no component mounted), and an in-flight save can no longer drain a newer unsaved snapshot because the drained project comes from the signal payload, not a live ref.

cosarah (commented):
I only have one remaining non-blocking note on the latest revision:

1. `tests/store/stage-agents.test.ts:45`, `tests/store/stage-generation-complete.test.ts:74`, and `tests/store/stage-insert-scene-after.test.ts:52`
These fake-timer cleanup changes still look unrelated to the PBL runtime outbox work itself. `CONTRIBUTING.md` asks to keep PRs focused, so I would prefer to split them into a separate cleanup PR unless there is a strong reason to keep them bundled here.

wyuc:
Fair point on focus — the reason they're bundled: those three suites carry a pre-existing timer leak (`setStageAgents`' debounced saves firing after environment teardown) that was latent on main but becomes **deterministically red on this branch's CI** — this PR grows `lib/utils/stage-storage.ts`'s import graph (it now reaches `lib/pbl/v2/runtime/drain.ts` → `@openmaic/storage`), which shifts module-load timing enough to trip it on every run (first CI run on this PR failed exactly there). So the hygiene fix is a prerequisite for this PR's CI rather than drive-by cleanup.

That said, happy to split it if you prefer: I'd extract the test-hygiene commit into its own PR, land it first, and rebase this one on top — say the word and I'll do that; otherwise I'll leave it bundled with this explanation on record.

cosarah (commented):
I only have one remaining non-blocking note on the latest revision:

1. `tests/store/stage-agents.test.ts:45`, `tests/store/stage-generation-complete.test.ts:74`, and `tests/store/stage-insert-scene-after.test.ts:52`
These fake-timer cleanup changes still look unrelated to the PBL runtime outbox work itself. `CONTRIBUTING.md` asks to keep PRs focused, so I would prefer to split them into a separate cleanup PR unless there is a strong reason to keep them bundled here.

wyuc:
Done — split into #897 as requested. #893 stays as is for now (dropping the commit here would need a force-push and would leave this branch's CI red until the fix lands); once #897 merges, I'll merge main back into this branch and the three test files will drop out of this PR's diff automatically. #893 is therefore sequenced behind #897.

cosarah (commented):
I only have one remaining non-blocking note on the latest revision:

1. `tests/store/stage-agents.test.ts:45`, `tests/store/stage-generation-complete.test.ts:74`, and `tests/store/stage-insert-scene-after.test.ts:52`
These fake-timer cleanup changes still look unrelated to the PBL runtime outbox work itself. `CONTRIBUTING.md` asks to keep PRs focused, so I would prefer to split them into a separate cleanup PR unless there is a strong reason to keep them bundled here.

wyuc:
#897 merged — origin/main is now merged back into this branch (7d04b949), and the three fake-timer cleanup files no longer appear in this PR's diff. The only remaining `tests/store/` file is `stage-save-signal.test.ts`, which belongs to this PR (it tests the post-save drain signal introduced here). The PR is now scoped to the runtime outbox work only.

wyuc:
The post-merge CI failure here is the same latent-race class in a second escape hatch (an un-awaited async debounce callback whose dynamic import lands after teardown — the audio-chain error in the log is that import's lazy tail). Fixed test-side and split out as its own PR per the convention we established: this PR is sequenced behind it; once it merges I'll merge main back here and CI should go green.

wyuc:
#899 merged and origin/main merged back — CI should now be green; this PR is fully unblocked and scoped to the outbox work only.

cosarah (approved):
LGTM

PR #897, opened 2026-07-10
test(store): drain debounced saves before environment teardown

wyuc:
Split out of #893 per review feedback (keeping that PR focused on the PBL runtime outbox).

`setStageAgents` schedules `debouncedSave` + `debouncedSaveAgents`; store tests that finished without flushing left those timers firing after vitest tore the environment down, lazily importing the settings/providers chain into a dead runtime (`EnvironmentTeardownError`). Latent on main today, it becomes deterministically red for any branch that grows `lib/utils/stage-storage.ts`'s import graph — #893 trips it on every CI run, which is why these fixes exist and why #893 depends on this landing first.

The three store suites now run fake timers, drain pending debounced saves in `afterEach`, and assert zero pending timers at teardown; the same latent race in the insert-scene-after and generation-complete suites is fixed in the same style.

Tests: the three suites pass 5x consecutively; full root suite green.

### DIFF
--- a/tests/store/stage-agents.test.ts
+++ b/tests/store/stage-agents.test.ts
@@ -42,6 +42,7 @@ function makeAgentConfig(id: string): GeneratedAgentConfig {
 }
 
 beforeEach(() => {
+  vi.useFakeTimers();
   useStageStore.setState({
     stage: makeStage(),
     scenes: [],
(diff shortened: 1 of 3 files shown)


cosarah (commented):
Reviewed in .worktrees/pr-897. I don’t have any blocking findings. Verified locally with: pnpm exec vitest run tests/store/*.test.ts

cosarah (approved):
LGTM

PR #899, opened 2026-07-10
test(store): settle dynamic imports before environment teardown

wyuc:
Follow-up to #897, same latent-race class, second escape hatch: draining fake timers is not enough when the drained callback is an **un-awaited async function** — its `await import('@/lib/store/settings')` can still land after vitest tears the environment down (the settings store lazily reaches the audio module chain, which is where the `EnvironmentTeardownError` points). Deterministically red on #893's CI after its latest main merge; latent everywhere else.

Fix (test-side, one file): the stage-agents suite now mocks the settings store and awaits `vi.dynamicImportSettled()` after draining timers, so no dynamic import escapes teardown.

Split out per the review convention established on #897; #893 is sequenced behind this landing.

### DIFF
--- a/tests/store/stage-agents.test.ts
+++ b/tests/store/stage-agents.test.ts
@@ -1,5 +1,9 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
+const { setSelectedAgentIds } = vi.hoisted(() => ({
+  setSelectedAgentIds: vi.fn(),
+}));
+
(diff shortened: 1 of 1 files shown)


cosarah (commented):
Reviewed in .worktrees/pr-899. I don't have any blocking findings. Verified locally with: pnpm exec vitest run tests/store/stage-agents.test.ts and pnpm exec vitest run tests/store/*.test.ts

cosarah (approved):
LGTM

Issue #874, opened 2026-07-08
[Refactor] Collapse the media provider adapters: runPolledTask driver + probeAuth helper


cosarah:
## Context

From an architecture review focused on turning **shallow** modules into **deep** ones. The media/audio provider layer follows a consistent "union type + registry object + dispatch switch + per-provider adapter" template, which has produced two deep drivers that are currently hand-rolled inside every adapter.

### 1. `runPolledTask` — one submit-then-poll-until-terminal driver
The identical submit → poll-loop → timeout control structure is reinvented inline in 6 video adapters:
- `lib/media/adapters/seedance-adapter.ts:243`
- `lib/media/adapters/kling-adapter.ts:228`
- `lib/media/adapters/grok-video-adapter.ts:173`
- `lib/media/adapters/veo-adapter.ts:218`
- `lib/media/adapters/happyhorse-adapter.ts:180`
- `lib/media/adapters/minimax-video-adapter.ts:164`

Only the constants, the submit/poll payload shaping, and the success/failure status strings differ. A `MediaTaskAdapter<TOptions,TResult>` interface was even **designed for exactly this** (`lib/media/types.ts:320`) and then **never implemented anywhere** — delete the phantom interface, or make this driver its implementation.

Introduce `async function runPolledTask<T>({ submit, poll, intervalMs, maxAttempts, label }): Promise<T>` where `poll` returns `{done,result} | {failed,message} | pending`. Each adapter shrinks to a `submit` closure + a `poll` closure. **This PR has the smallest conflict surface — do it first.**

### 2. `probeAuth` — one connectivity/auth-probe helper (fixes #574)
All 13 adapters repeat `try { fetch(minimal request); if (401||403) return {success:false, "<Provider> auth failed"}; return {success:true}; } catch { ... }` (e.g. `seedream-adapter.ts:62`, `qwen-image-adapter.ts:37`, `kling-adapter.ts:127`), dispatched by two more switches (`image-providers.ts:139`, `video-providers.ts:133`). Introduce `probeAuth({ request, providerName })` that owns the 401/403-vs-else classification and message strings; adapters supply only the request builder.
- **Fixes / relates to #574** (availability check false positives → unreachable providers assigned to agents). Link it.

### 3. (Deferred) registry-as-dispatch + `resolveProviderCredentials`
[...]

PR #900, opened 2026-07-10
refactor(media): share submit-poll task driver

YizukiAme:
## Summary

Part of #874 (sub-task 1).

All six video adapters hand-rolled the same submit → poll-until-terminal → timeout control flow. This PR extracts it into a shared `runPolledTask` driver (`lib/media/polled-task.ts`) and migrates Seedance, Kling, Grok, Veo, HappyHorse and MiniMax onto it. Each adapter now supplies a `submit` closure, a `poll` closure and its existing constants.

Strictly behavior-preserving:

- Polling cadence unchanged: wait before the first poll, exactly `maxAttempts` polls, per-provider intervals/limits stay local to each adapter.
- Every provider's failure and timeout message is preserved byte-for-byte (task/request IDs, MiniMax's `last status`, Veo's fixed "10 minutes" wording) — pinned by tests.
- Veo: an operation already completed at submit time short-circuits with zero waits and zero polls, matching current behavior — `submit` may return a terminal result for exactly this case.
- MiniMax: the extra file-URL retrieval stays inside the success path.
- Existing public exports (`submitSeedanceTask`, `pollSeedanceTask`, HappyHorse equivalents) keep their signatures and behavior.
- Network/HTTP/JSON errors thrown by adapter code propagate unwrapped.

Also removes the never-implemented `MediaTaskAdapter` interface from `lib/media/types.ts`: its `poll → result | null` shape cannot express submit-time completion or failure detail, and `runPolledTask` supersedes it. The "how to add a provider" guidance in `types.ts` now points at the driver.

No changes to connectivity probes, provider registries/dispatch, or request payloads (those are sub-tasks 2/3 of #874).

## Tests

- `tests/media/polled-task.test.ts` — driver semantics: immediate terminal submit (zero wait / zero poll), wait-before-first-poll, explicit task-id passing, exact `maxAttempts`, unwrapped error propagation, timeout context (default + custom formatter).
- `tests/media/video-polled-adapters.test.ts` — six-adapter regression: submit-once, success, explicit-failure and pending-to-timeout paths asserting the exact legacy message strings; Veo zero-wait fast path; MiniMax secondary file query.
- Full local run: media suite 47/47, unit suite, importer/DSL/storage package tests, Playwright e2e, `tsc --noEmit`, lint, i18n keys — all green (re-verified after rebasing onto current `main`).

### DIFF
--- a/lib/media/adapters/grok-video-adapter.ts
+++ b/lib/media/adapters/grok-video-adapter.ts
@@ -20,15 +20,12 @@ import type {
   VideoGenerationOptions,
   VideoGenerationResult,
 } from '../types';
+import { runPolledTask } from '../polled-task';
 
 const DEFAULT_MODEL = 'grok-imagine-video';
 const DEFAULT_BASE_URL = 'https://api.x.ai/v1';
 const POLL_INTERVAL_MS = 10_000; // 10 seconds
 const MAX_POLL_ATTEMPTS = 60; // 10 minutes max
-
-function delay(ms: number): Promise<void> {
-  return new Promise((resolve) => setTimeout(resolve, ms));
-}
 
 /** Dimension defaults per aspect ratio */
 function getDimensions(aspectRatio?: string): {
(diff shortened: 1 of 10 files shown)


wyuc (approved):
Reviewed at c1953fc after updating onto current main. The updated patch-id exactly matches the cross-reviewed change. Independent Claude and Codex reviews found no correctness issues; the shared driver and all six adapter migrations preserve polling cadence/count, task IDs, terminal/timeout/error semantics, and exception propagation. Media regressions pass 47/47, and latest full CI (lint/typecheck/unit + E2E) is green. Approving.

PR #902, opened 2026-07-11
feat(pbl): read flip — fold learner state from RuntimeStore with snapshot backfill (#869 Part C, step 3)

wyuc:
Step 3 of #869 Part C, built on the Part A envelope (#870), the Part B RuntimeStore (#880), learner identity (#885), and the step-2 dual-write (#893).

## What

On stage load, PBL learner state is now reconstructed by **folding RuntimeStore records over the design-time template**, instead of trusting only the persisted document. The #893 dual-write and document persistence stay untouched — the document remains authoritative whenever the fold disagrees, so this is the strangler's read flip, not the write cutover.

- **`PBLLearnerState` split** (`lib/pbl/v2/runtime/learner-state.ts`): pure `extractLearnerState` / `applyLearnerState` / `stripToDesignTemplate` codify the document/runtime boundary of #869 Part C. A field-inventory test forces every future `PBLProjectV2` field to be classified (learner state / design template / excluded transient).
- **Drain enrichment** (`record-payloads.ts`, `drain.ts`): the document outbox stays id-only, but the drainer — which holds the full project — writes **self-sufficient records**: original event + content attachment (message body, submission, evaluation, proficiency, handover, status detail). Missing content records `attachment: null` with a reason instead of throwing. Payloads are versioned.
- **Snapshot records**: a `pbl_snapshot` payload seeds sessions for pre-existing learners (backfill) and self-heals divergence; snapshots supersede earlier records, including their gaps, and respect `project_reset` epochs.
- **Fold** (`fold.ts`): seq-ordered, id-deduped, epoch-aware, engagement ring-cap identical to the document's, gap diagnostics instead of throws, and compile-time exhaustiveness so a future event kind cannot be silently ignored.
- **Hydration flip** (`hydration.ts`, `lib/store/stage.ts`, classroom server fallback): catch-up drain → fold → compare. Equal → the fold result is used (the flip). Mismatch or gaps → document wins, dev-mode warning, deduped self-heal snapshot so the next load converges. Hydration never throws into the load path, scenes hydrate in parallel, and both load paths re-check stage identity after the async window so a stale load cannot clobber navigation.
- **Drain hardening**: overlapping drains for the same (stage, scene, learner) are serialized and individually time-bounded, so duplicates aren't appended and a hung append can't block the chain.

## Equivalence guarantee

The core invariant is tested property-style: randomized reducer runs drive a real `BrowserRuntimeStore` (fake-indexeddb), and after every persisted change + drain, `fold(records)` must deep-equal `extractLearnerState(project)` on all covered fields.

During review this surfaced one real coverage gap: instructor thread compaction mutates a per-request server copy after the last streamed patch, so compacted thread state never reaches the persisted client document. A `thread_compacted` event was prototyped and then removed as dead code; `earlierSummary` stays extracted for forward-compat and the exclusion is documented in `learner-state.ts`.

## Explicitly deferred (next steps of #869 Part C)

- Stopping document persistence of learner state and removing the preserving merge shim (the write cutover).
- Incremental fold (seq-cursor `listRecords`) — today every load folds full scene history.
- Migrating chat/quiz/playback kinds off Dexie/localStorage (Part B follow-up).

## Tests

2105 passing (root suite; +70 over main): fold unit matrix (seq order, dedup, epochs, snapshot supersession, gaps), learner-state round-trip + field inventory, enrichment shapes, hydration integration (pre-existing learner backfill, fresh learner, happy flip, mismatch self-heal, ring eviction, duplicate drains, stale-load races), and the randomized equivalence property test. Prettier, ESLint, and tsc clean.


cosarah (commented):
I found two issues in the new read-flip / hydration paths:

1. `app/classroom/[id]/page.tsx:53-74`
`setStage(stage)` clears `scenes` and schedules a debounced `saveToStorage()`. Because this path now awaits `hydrateClassroomFallbackScenes(...)` before repopulating the scenes array, a slow hydration can persist `{ stage, scenes: [] }` to IndexedDB. The later `useStageStore.setState({ scenes: hydrated, ... })` does not trigger another save, so the empty snapshot can become durable and the next classroom load will stop at the local empty snapshot instead of falling back to the server.

2. `lib/pbl/v2/runtime/drain.ts:292-325`
The new per-key serialization is queueing `withTimeout(drainProjectRuntimeWork(...))`, not the underlying drain work itself. Once the timeout wins, the queued promise rejects and the next drain is allowed to start, but the timed-out append is still running in the background. If it eventually succeeds, it can interleave with the retry and reintroduce duplicate / overlapping appends for the same `(stageId, sceneId, learnerKey)`.

wyuc:
Thanks for the review — both findings were real, fixed in 4fb2e4c2.

**1. Empty-scenes snapshot on the classroom fallback**: exactly right — `setStage` clears scenes and schedules the debounced save, so a slow hydration let `{ stage, scenes: [] }` become a durable snapshot that would shadow the server copy on the next load. The fallback now hydrates **before** touching the store and applies stage + scenes in the same synchronous update (`applyHydratedClassroomFallbackScenes`), so no save can ever observe the empty window. Regression: `does not persist a server fallback stage before its hydrated scenes are ready`.

**2. Serialization queued the timeout, not the work**: also right — once the race resolved, the next drain could interleave with the still-running append. The per-key chain now awaits the **actual work promise** (overlapping same-key appends are impossible), and the work takes a cooperative deadline it checks before each `appendRecord` and before persisting the watermark, so a slow drain bails out promptly instead of wedging the chain — callers keep the same bounded-wait, never-throw behavior. Regressions: `keeps same-key drain work serialized after the caller timeout wins`, plus the existing no-permanent-block test updated to the cooperative-deadline model.

Full suite 2107 passing; prettier/eslint/tsc clean.

cosarah (commented):
One follow-up issue remains in the latest fix:

`lib/pbl/v2/runtime/drain.ts:271-281` and `lib/pbl/v2/runtime/drain.ts:311-346` still let a truly hung `appendRecord()` wedge the same-key drain chain forever. The overlap bug is fixed because the queue now holds the underlying work promise, but the timeout only wraps the caller-facing await. If one append never resolves, `work` never settles, `inFlightPblDrains` never clears, and every later drain for the same `(stageId, sceneId, learnerKey)` just times out waiting behind the stuck predecessor. The new tests cover a slow append that is eventually released, but not a never-resolving append, which is the head-of-line-blocking case this hardening is supposed to prevent.

cosarah (commented):
One regression remains in the latest store fix:

`lib/store/stage.ts:450` still lets an older load win over a later navigation. The new guard only checks whether `latestStageId` changed away from `entryStageId`, but that is not sufficient when two different target loads overlap from the same resident stage. A concrete repro is:

1. the store is currently on stage B
2. `loadFromStorage('stage-c')` starts
3. before C finishes hydrating, `loadFromStorage('stage-d')` starts

Both calls capture `entryStageId = 'stage-b'`. If C resolves first, it commits stage C. When D resolves, it now sees `entryStageId !== latestStageId` and `latestStageId !== stageId`, so it aborts as a stale load, leaving the user on stage C even though D was the later navigation. I reproduced this locally with a minimal Vitest case.

wyuc:
Good catch again — fixed in bfe86051 (and the unit-test regression you saw on CI was fixed in 76af531d; CI is green on that head).

**Hung append wedging the chain**: correct — the chain linked on the raw work promise, so a never-resolving `appendRecord` held the per-key slot forever. The chain link now races the work against a hard cap (2× the drain timeout) and releases the slot when the cap fires. This is safe because the cooperative deadline already guarantees an expired drain can neither append further records nor touch the watermark — the worst a straggler can do is land the single already-in-flight record, which carries a deterministic event id and dedups downstream, while the successor owns the watermark. That invariant is now documented at the race site. Regression: `releases same-key drains after a hung append and keeps late duplicate state consistent` — asserts the successor proceeds after the cap, the watermark advances to the successor's event and never regresses, and the late resolution adds exactly one dedupable duplicate.

Suite 2109 passing, prettier/eslint/tsc clean.

wyuc:
Right again — and after three rounds of patching that predicate, it clearly wanted the request-token design you suggested in your first review. Done in 644ada48.

Every async scene-committing load (`loadFromStorage` and the classroom server fallback — one shared counter) claims a monotonic token at entry and commits only if it is still the newest request at commit time, so the latest-started navigation wins under every interleaving and resolution order. `setStage`/`clearStore` also advance the token, so synchronous navigation invalidates any pending async load. The old stage-id predicates are gone; the only survivor is the same-stage in-memory-scenes recheck, which covers non-load writers (generation's `setScenes`) that don't claim tokens.

Regressions: your exact repro in both resolution orders (`keeps the later overlapping storage load when the older/newer hydration resolves first`), cross-entry-point interleaves (`lets a later classroom fallback apply beat an earlier storage load`, `does not apply fallback scenes after a newer stage request starts`), plus all prior stale-load and legacy-completion tests. Suite 2112 passing, prettier/eslint/tsc clean.

cosarah (commented):
One follow-up race remains in the latest load-token fix:

`app/classroom/[id]/page.tsx:41-50` and `lib/classroom/pbl-fallback-hydration.ts:30-35` still let an older server fallback overwrite a newer classroom navigation when both classrooms miss IndexedDB. A concrete repro is:

1. `loadFromStorage('stage-a')` starts and finds no local data.
2. The user navigates to stage B, so `loadFromStorage('stage-b')` claims a newer load token.
3. Stage A's `/api/classroom` response arrives before stage B reaches fallback hydration.
4. `applyHydratedClassroomFallbackScenes()` claims a fresh token at line 30 and applies stage A, even though stage B is already the newer request.

The shared token fixes overlap once fallback hydration has already started, but claiming freshness inside the fallback helper is too late for the "older fallback response arrives before the newer fallback starts" case. I think the token needs to be claimed once at classroom-load start and threaded through both the IndexedDB and server-fallback paths (or otherwise bind this helper to the original request token) instead of re-claiming it here.

The current tests cover overlapping storage loads and a later fallback beating an earlier storage load, but they do not cover this older-fallback-vs-newer-navigation sequence.

wyuc:
Fixed in 4646988a. You diagnosed it exactly: claiming a fresh token inside the fallback helper laundered a stale request into the newest one. The token is now claimed once at classroom-load start and threaded through both paths of that navigation — the IndexedDB attempt (loadFromStorage accepts the pre-claimed token; other callers keep the internal claim) and the server fallback, which now only VALIDATES its origin token at commit. A newer navigation claims a newer token at its own load start, atomically invalidating every pending path of the older one.

Regressions cover your sequence and its neighbors: 'does not let an older classroom fallback override a newer storage navigation', 'does not let an older classroom fallback apply before the newer fallback resolves', 'does not let an older classroom fallback override a newer fallback navigation', plus the same-navigation happy path. Suite 2117 passing, prettier/eslint/tsc clean.

cosarah (commented):
One stale-navigation bug still remains in the latest classroom-load fix:

`app/classroom/[id]/page.tsx:41-136` now threads `loadToken` through `loadFromStorage()` and server fallback hydration, but after that first `await` there is no further cancellation check for the rest of the async flow. If classroom A becomes stale after `loadFromStorage('a', loadTokenA)` yields, the old request can still run:

- `restoreFromDB(classroomId)` at `app/classroom/[id]/page.tsx:95`, which merges the old stage's media tasks back into the shared task map (`lib/store/media-generation.ts:160-201`). That undoes the explicit cleanup at `app/classroom/[id]/page.tsx:147-152`, and the comment there already calls out that placeholder IDs are not globally unique across classrooms.
- `loadGeneratedAgentsForStage(classroomId)` at `app/classroom/[id]/page.tsx:99`, which first clears every generated agent from the registry before loading the requested stage's roster (`lib/orchestration/registry/store.ts:336-369`). So a stale A request can wipe out B's generated agents after B has already won the navigation.
- `setLoading(false)` in the `finally` block at `app/classroom/[id]/page.tsx:135-136`, which can also drop the newer navigation out of its loading state early.

A concrete repro is: A starts loading, B navigation starts before A finishes the later awaits, and B has already populated the current classroom (from storage or fallback). A's old `loadClassroom()` then resumes and repopulates A's media tasks / generated agents because the page-level async work was never cancelled.

I think this needs a page-scoped cancellation guard after each async phase (or a request identity that is independent from `setStage()`), not just the stage/fallback token checks.

wyuc:
Fixed in 39fd3b8c. You're right that the token only protected the scene commit — the rest of loadClassroom ran to completion for a superseded navigation. I extracted the whole flow into `runClassroomLoad` with an `isCurrent()` gate that combines the effect-cleanup cancel flag with `isCurrentStageSceneLoadToken(loadToken)`, checked after every await and before every shared-store write. A superseded run now performs zero further side effects.

Guarded write inventory (each post-await store mutation now gated): fallback scene apply; `restoreFromDB` media-task merge (split into DB read + `applyRestoredMediaTasks`, gate before the merge); generated-agent save (freshness checked before/after DB awaits and before registry mutation); generated-agent restore (split read + `applyGeneratedAgentRecordsToRegistry`, gated); agent selection/settings setters; auto-resume media kickoff (before `generateRemaining`/`markGenerationCompleteIfDone`/`generateMediaForOutlines`); and `setError`/`setLoading(false)` only when current, so a superseded run never drops the newer navigation's loading state.

Tests drive the extracted orchestrator with injected fakes and supersede at each phase boundary (after storage, after fetch, after fallback apply, after the media/agent DB reads) plus unmount cancellation and the happy path. Suite 2124 passing, prettier/eslint/tsc clean.

wyuc:
Heads-up on CI: the page-load cancellation work (isCurrent/load-token) I added while addressing your review had a regression that turned E2E red for a few commits — fixed in a03a800f, all checks green now.

Root cause (from the CI Playwright trace + code): the generation-resume effect had picked up a `cancelled`-flag guard around its kickoff. `generateRemaining` gets a fresh identity every render (`useSceneGenerator`'s `useCallback` depends on the `options` object, and the classroom passes a new `{ onComplete }` literal each render), so that effect re-runs every render and its cleanup trips `cancelled`; the kickoff runs inside an async `loadImageMapping().then(...)` that resolves after a re-run, so generation was silently skipped — the deck stayed at Generating… and the Edit-course toggle stayed disabled, timing out the two slide specs. Restored the generation-resume effect to its prior unguarded form (navigation-away already stops the generator via the load effect's `stop()` cleanup). The scene-load-token guards on the load/commit paths are untouched, so the races from your five review rounds stay fixed.

Green and back with you for review.

cosarah (commented):
One follow-up data-loss bug remains in the new `load-classroom` extraction:

`runClassroomLoad()` now calls `saveGeneratedAgents(...)` for server-fallback classrooms at `lib/classroom/load-classroom.ts:106-109`, and `saveGeneratedAgentsForCurrentLoad()` implements that write at `lib/classroom/load-classroom.ts:256-269`. The problem is that the replacement write is now cancellable in the middle:

1. `await db.generatedAgents.where('stageId').equals(stageId).delete()`
2. `if (!isCurrent()) return []`
3. `await db.generatedAgents.bulkPut(records)`

If classroom A becomes stale while step 1 is in flight, the function returns at step 2 after the old records have already been deleted, so stage A loses its persisted generated-agent records entirely. A later visit to that classroom will then behave like it never had generated agents in IndexedDB.

The new tests cover cancellation between the higher-level phases, but not cancellation inside this destructive write path. I think the delete + bulkPut needs to be one non-cancellable replacement (for example a transaction, or at least no early-return branch between the delete and the replacement insert) so cancellation cannot leave the DB half-updated.

wyuc:
Fixed the generated-agent persistence race in `3efadbfe` by making the delete and replacement insert one atomic IndexedDB transaction. A stale navigation can no longer leave the stage with no persisted generated agents.

Codex review also found that stale media restores leaked object URLs. Fixed in `431e35dc` by explicitly revoking URLs for restored tasks that never enter the store.

Full local suite passes: 258 test files and 2127 tests. The latest full-branch Codex review found no actionable correctness issues.

cosarah:
## AI-assisted code review (Standards + Spec axes)

Reviewed the diff `origin/main...HEAD` on two axes. **No blocking items** — the core fold-equivalence invariant is covered by the property test, and nothing from the DEFERRED list leaked in (verified: `mergePreservingPBL` is absent on both `main` and this branch, `listRecords` still takes only `{ sceneId }` so there is no incremental fold, and chat/quiz/playback kinds are untouched). This is a clean read flip. Below are non-blocking findings worth a response.

### Spec

**1. The field-inventory test doesn't force what the PR body claims (partial).**
The PR says *"A field-inventory test forces every future `PBLProjectV2` field to be classified."* But `tests/pbl/v2/learner-state.test.ts:184-262` iterates `Object.keys(project)` over a **fixture instance**, not the `PBLProjectV2` type. A future optional field that isn't added to `makeProject` produces no key and silently escapes the assertion — so it only enforces classification of fields already present on the fixture, not every future field. Worth tightening since it guards the PR's central equivalence guarantee.

**2. `stripToDesignTemplate` doesn't clear `proficiencyAssessment`, contradicting the field classification.**
`stripToDesignTemplate` (`lib/pbl/v2/runtime/learner-state.ts:110-143`) clears the runtime fields but leaves `proficiencyAssessment`, even though the inventory test classifies it as learner state (`learner-state.test.ts:231`). As a result `foldPBLRuntime`'s `baseline` (`lib/pbl/v2/runtime/fold.ts:327-328`) inherits the live `proficiencyAssessment` from the document rather than reconstructing it purely from records. The equivalence tests pass only because the template is derived from the same `args.project`. This weakens the "fold from records" guarantee for `proficiencyAssessment` and is inconsistent with its own classification. (`resetProjectProgress` in `progress.ts:203` deliberately preserves proficiency across a reset, so document and fold happen to agree — but for a reason the fold doesn't model.)

### Standards

**3. One-concern-per-PR (judgement call).**
Beyond the read-flip fold, the branch bundles a large `load-classroom.ts` extraction plus several classroom generation/media fixes not tied to the fold (`431e35d` stale restored media, `3efadbf` atomic generated-agent replacement, `a03a800`/`5a56630` generation-resume behavior). Threading `loadToken` through the server fallback arguably motivates the extraction, but the media-cleanup and generation-resume changes exceed the read-flip's stated scope. Borderline, not a hard breach — flagging for focus.

**4. Minor.**
- i18n (`CONTRIBUTING.md`): `lib/classroom/load-classroom.ts:154` surfaces a hardcoded `'Failed to load classroom'` in the page error state. Pre-existing (moved verbatim from `page.tsx`), but the diff now owns it.
- Duplicated code + data clump: `saveGeneratedAgentsForCurrentLoad` (`load-classroom.ts:284-311`) and `applyGeneratedAgentRecordsToRegistry` (`322-345`) share the same delete-generated-then-re-add shape and the same `{ allowedActions, isDefault, isGenerated, boundStageId, createdAt, updatedAt }` field group; a shared `populateGeneratedAgents` helper would remove it.

Findings 1 and 2 are the substantive ones; both are cheap to address and both concern the equivalence guarantee this PR is built around.

wyuc:
Thanks, both substantive findings were valid and are fixed.

1. The field inventory is now type-exhaustive: `PROJECT_FIELD_BOUNDARY` uses `satisfies Record<keyof PBLProjectV2, ...>`, so adding any future required or optional top-level field fails TypeScript until it is classified. The runtime assertion still checks that the representative fixture and the classification stay aligned.

2. `stripToDesignTemplate` now clears `proficiencyAssessment`, and the empty-history fold test explicitly starts from a document carrying an assessment and verifies that the baseline does not inherit it.

While re-reviewing the complete branch with fresh Codex-only sessions, I found and fixed several adjacent contract issues in the same runtime boundary:

- hydration waits for the real same-key drain before reading records or writing a snapshot;
- that barrier is bounded across the entire queue plus execution, and timeout falls back to document state without writing a snapshot;
- classroom fallback rechecks effect cancellation immediately before applying hydrated scenes;
- `project_reset` clears project progress while preserving the proficiency model, matching `resetProjectProgress`.

The fixes are in `5e0be2d3`, `c919a072`, `7b913a86`, and `3f5e9c18`. Each issue has a regression test. Fresh local gates pass: Prettier, ESLint (0 errors; 15 pre-existing warnings), TypeScript, and 269 test files / 2265 tests. The final fresh Codex review of the last queue-level fix reported no findings.

cosarah:
## Re-review after latest push (`5d4ccc3 → 3f5e9c1`)

Both substantive findings from my earlier review are resolved. Verified against the code, the affected test suite, and a clean `tsc --noEmit`.

**✅ Field-inventory test (was: only enforced fixture keys, not the type).**
`learner-state.test.ts` now declares `PROJECT_FIELD_BOUNDARY` typed `satisfies Record<keyof PBLProjectV2, ProjectFieldBoundary>`. This is a real fix — a future `PBLProjectV2` field now fails to **compile** (missing key in the record) rather than silently passing the assertion. The map covers every current field and typecheck is clean.

**✅ `stripToDesignTemplate` no longer leaks `proficiencyAssessment`.**
`stripToDesignTemplate` now sets `template.proficiencyAssessment = undefined` (`lib/pbl/v2/runtime/learner-state.ts:116`), so the fold baseline no longer inherits it from the document. Proficiency is reconstructed purely from records via the `proficiency_updated` event's `proficiency` attachment (`fold.ts:282-292`, `record-payloads.ts:183-186`). The reset path (`fold.ts:369-372`) explicitly carries proficiency across an epoch reset to mirror `resetProjectProgress`'s deliberate preservation — the fold now *models* that behavior instead of accidentally agreeing with it.

**Drain/hydration hardening (`drainProjectRuntimeFully`) looks sound.**
The bounded completion barrier waits for in-flight appends, runs with an uncapped cooperative deadline under a hard wall-clock `withTimeout`, and swallows late rejections after fallback. `activePblDrainWork` entries are cleaned up in `.finally`, so no leak. The `isCurrent()` guard threaded into the fallback-apply path tightens the stale-load check. Intent — "hydration must observe the completed drain or abort to document fallback; never snapshot while an append is pending" — is well served.

**Verification:** 55 tests pass across `fold` / `learner-state` / `drain` / `hydration` / `pbl-fallback-hydration`; `tsc --noEmit` clean.

**Still open (non-blocking, unchanged):** the one-concern-per-PR point (the classroom-load refactor + generation/media fixes riding along) and the minor i18n string / generated-agent duplicated-code helper. None are blockers.

cosarah (approved):
LGTM

PR #903, opened 2026-07-11
refactor(media): share auth probe across matching adapters

YizukiAme:
## Summary

Part of #874 (sub-task 2) · Related to #574 (whose root cause lives in the TTS voice-availability path, so this PR does not close it)

Extracts the shared auth-probe skeleton into `probeAuth` (`lib/media/probe-auth.ts`) and migrates the seven connectivity probes that genuinely share the auth-only policy: Seedream, Qwen Image, Grok Image, Grok Video, Seedance, Kling, HappyHorse. The helper has a hook-free `{ providerName, request }` interface; each adapter shrinks to its provider name plus a request closure. Kling's key parsing and JWT generation stay inside the closure, so its error boundary is unchanged.

**Scope: deliberately 7 of the 13.** As audited in #874, the other six probes follow different verdict policies — four strict `response.ok` probes (OpenAI Image, Lemonade, MiniMax Image/Video) and two Google two-request-fallback probes (Veo, Nano Banana). Wrapping them would move their classification logic behind per-adapter hooks rather than hide it, so they intentionally stay as-is, pinned by sentinel tests documenting their distinct semantics. ComfyUI's probe (keyless local health check with its own timeout and logging) is also out of scope.

## Behavior preservation

- Commit `test(media): characterize provider auth probes` adds characterization tests against the **pre-migration** code: exact request shape (URL/method/headers/body) and the success / auth-failure / network-error verdicts and message strings for all seven probes, plus sentinels for the six non-matching ones.
- Commit `refactor(media): share auth probe across matching adapters` migrates the seven adapters. The characterization tests run unchanged, providing regression evidence that the covered verdicts and messages are preserved.

## Rebased after #900

#900 has landed. This branch is now rebased onto current `main`; the PR contains only the two Task 2 commits above.

## Tests

Media suite 84/84 (the 28 characterization tests verified to pass on both sides of the migration), full unit suite, importer/DSL/storage package tests, Playwright e2e, `tsc --noEmit`, lint — all green.

### DIFF
--- a/lib/media/adapters/grok-image-adapter.ts
+++ b/lib/media/adapters/grok-image-adapter.ts
@@ -18,6 +18,7 @@ import type {
   ImageGenerationOptions,
   ImageGenerationResult,
 } from '../types';
+import { probeAuth } from '../probe-auth';
 
 const DEFAULT_MODEL = 'grok-imagine-image';
 const DEFAULT_BASE_URL = 'https://api.x.ai/v1';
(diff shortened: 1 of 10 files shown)


wyuc (approved):
Cross-review completed with no actionable findings from either independent reviewer. The seven adapter migrations preserve request construction, 401/403 classification, messages, and error boundaries; distinct provider policies remain excluded and protected by tests. After updating onto the latest main, the PR diff remained byte-for-byte identical to the reviewed diff, and the refreshed Lint/Typecheck/Unit Tests and E2E Tests both passed.

Issue #864, opened 2026-07-07
[Task] VideoTimeline IR contract + pure compile pipeline (first architecture slice)


cosarah:
## Context

Sub-issue of #854 (Hyperframes-based classroom video export architecture). This is **the first architecture slice** — it satisfies the acceptance criteria stated in #854. Depends on the shared spec package #863 (and audio duration #861).

The compiler artifact — a `VideoTimeline` **IR** — is the system contract; the Hyperframes composition is a downstream rendering detail. Making the IR the contract gives us testability, subtitle derivation, and human-diffable review, and lets a thin emitter absorb Hyperframes' pre-1.0 API churn. Crucially, the IR must be produced and inspected **without FFmpeg/Chrome**, per #854's AC.

## Goal

Define the `VideoTimeline` IR schema and a set of **pure compile passes** that turn classroom data into that IR, fully unit-testable with no browser/FFmpeg.

## Scope

1. **IR schema** — `VideoTimeline` (version, stage, canvas, config, total duration, scenes, subtitles, assets, diagnostics). Per-scene segments carry base/narration/effects/videos/markers. Two deliberate choices:
   - Effect segments **reference an animation descriptor id** (e.g. `spotlight.v1`) and carry the effective action-level overrides needed to reproduce authored behavior (e.g. spotlight dimness or laser color). They do **not** inline descriptor tracks/keyframes: "what/where happened" plus authored overrides live in the IR, while reusable animation behavior lives in the spec package (#863).
   - **Diagnostics are first-class** — estimated durations, skipped media, unresolved elements, unsupported scenes are all recorded; the manifest doubles as an export report.
2. **Compile passes (pure functions)**:
   - `normalize` — stable scene ordering + action validation / implicit-rule assertions.
   - `probe` — obtain audio/video durations via an injected `TimingProbe` (real impl provided by the app; stubs in tests).
   - `timeline` — call the spec package's `resolveActionTimeline` to expand absolute times per scene.
   - `geometry` — resolve `elementId` → viewBox coords via the pure `findElementGeometry`; failure → degrade + diagnostic.
   - `assets` — dedup + zip layout/naming plan (reuse #849's planner/collector logic).
   - `emit` — IR → manifest JSON (the emitter/composition HTML itself is the next phase).
3. **Dependency-injection boundary** — `TimingProbe` / `AssetSource` interfaces so the compiler stays pure; app provides Dexie/DOM impls, tests provide stubs.
[...]

PR #913, opened 2026-07-13
feat(video-export): VideoTimeline IR contract + pure compile pipeline (#864)

cosarah:
## Summary

First architecture slice of #854 (Hyperframes classroom-video export), closing #864.

Makes a `VideoTimeline` **IR** the system contract for video export, produced by a set of **pure compile passes** with no FFmpeg / Chrome / DOM. Live app state enters only through injected `TimingProbe` / `AssetSource` interfaces, so the whole compiler runs — and is unit-tested — in plain Node. The Hyperframes composition becomes a downstream consumer of the IR, added in a later phase (P1d).

Two deliberate modeling choices from the issue:
- **Effect segments reference an animation descriptor id** (`spotlight.v1`) rather than inlining animation params — *what/where happened* lives in the IR; *what it looks like* lives in the shared `lib/choreography` descriptors (#863).
- **Diagnostics are first-class** — estimated durations, missing/skipped media, unresolved elements, and unsupported scene families are all recorded, so the manifest doubles as an export report.

## What's included

New module `lib/video-export/`:
- `ir.ts` — the `VideoTimeline` IR, authored as a zod schema with types inferred from it (single source). Envelope + per-scene `base / narration / effects / videos / markers`, plus `subtitles`, `assets`, `config`, and `diagnostics`.
- `deps.ts` — the DI boundary: **synchronous** `TimingProbe` / `AssetSource` interfaces + the `CompilerScene` input type.
- `geometry.ts` — pure `elementId → percentage-geometry` resolver (faithful to the runtime's fixed 1000×562.5 base).
- `passes/` — the six pure passes: `normalize` (deterministic ordering + action validation), `probe` (adapts `TimingProbe` into the choreography option shape), `timeline` (folds `lib/choreography`'s `resolveActionTimeline` into per-scene buckets + subtitles), `geometry` (degrade-on-miss), `assets` (dedup + zip naming plan), `emit` (IR → validated manifest JSON).
- `compile.ts` — the `compileVideoTimeline` orchestrator.

`eslint.config.mjs` — a machine-enforced purity boundary on `lib/video-export/**` (allows `@openmaic/dsl`, `zod`, `../choreography`, and relative siblings; bans the `@/` host alias and React/DOM/GSAP/motion), mirroring the `lib/choreography` boundary.

## Faithful to the single source of truth

The `timeline` pass delegates **all** timing / blocking-vs-fire-and-forget / effect-lifetime-clamping / whiteboard-auto-open semantics to `resolveActionTimeline` in `lib/choreography` (#863). It re-derives none of them — avoiding the second-source-of-truth drift that #854/#863 exist to prevent.

## Acceptance criteria (#864 / #854 first slice)

- [x] Documented compiler contract from classroom data → composition artifact (the IR + manifest).
- [x] Deterministic scene/action ordering.
- [x] Explicit audio/media reference mapping.
- [x] Unsupported scene families (quiz/interactive/pbl) represented intentionally as markers + diagnostics, never silently dropped.
- [x] Artifact inspectable in tests **without FFmpeg/Chrome**.
- [x] Follow-up render execution can be added without rewriting the compiler contract.

## Testing

- **107 tests pass**: 47 new golden/stub-driven specs in `tests/video-export/` (no browser/FFmpeg) + 60 `lib/choreography` regression tests.
- `tsc --noEmit` clean; `eslint` clean.
- Purity boundary verified to block `@/` and `react` imports while allowing `@openmaic/dsl` / `zod` / `../choreography` / in-module relatives.

The changes were cross-reviewed by an adversarial review agent (verdict: faithful, no critical/major defects); the review's minor findings — capped-video `durationSource` labeling, video null-vs-absent-media diagnostics, eslint message wording, and coverage of the previously-untested branches — are addressed in the second commit.

## Out of scope (follow-ups)

- App-side Dexie/DOM `TimingProbe` / `AssetSource` implementations + wiring `compileVideoTimeline` into the export menu (P1d).
- The Hyperframes composition HTML / GSAP emitter and actual MP4 rendering.
- Reconciling with #849's `video-frame-*` artifact once that PR merges.

Depends on #861, #863 (both merged).


cosarah:
## Self-review (adversarial cross-review)

Before opening this PR the diff was put through an adversarial cross-review against the `lib/choreography` source of truth (#863), the full `lib/video-export` diff, the DSL action definitions, the runtime geometry helper, the eslint boundary, and all tests.

**Verdict: solid and faithful — no critical or major defects.** The findings below were all minor/nits and are addressed in commit `511a68a` ("address cross-review findings").

### Confirmed correct (checked, not assumed)
- **No second source of truth.** The `timeline` pass delegates all timing / blocking-vs-fire-and-forget / effect-lifetime-clamping / whiteboard-auto-open semantics to `resolveActionTimeline`; it re-derives none of them.
- **Index alignment.** `normalize` re-indexes kept actions `0..n`; `resolveActionTimeline`, `geometry` (`applyGeometry(timeline.scenes, normalized.scenes)`), and `assets` (`sourceScene.actions[seg.actionIndex]`) all index the same normalized array — no off-by-one, no pass reaches back to the raw input.
- **Guard order in `dispatchSegment`.** `EMPTY_SCENE_DWELL` (a `type: 'speech'` empty-text beat) is diverted to an `empty-scene` marker by the `action.id` guards that run *before* the type switch — so it never becomes a narration segment.
- **Scene/total math.** scene start = first-segment `startMs`; duration = next-scene-start-or-total − start; total = `max(startMs + advancesCursorMs)` (cursor completion). A trailing spotlight-only scene having duration 0 is faithful to the engine's per-scene `clearEffects`.
- **Geometry math** in `lib/video-export/geometry.ts` is equivalent to the runtime `lib/utils/geometry.ts` (fixed 1000×562.5 base, 0.5625 ratio).
- **Purity / DI.** Only `@openmaic/dsl`, `zod`, `../choreography`, and relatives; no DOM/React/GSAP; `normalize` does not mutate input; all state enters via `TimingProbe` / `AssetSource`. Runs in plain Node.

### Findings addressed
1. **[minor] Capped video mislabeled `durationSource: 'stored'`.** When `getVideoDurationMs` returned a value above `MAX_VIDEO_WAIT_MS`, `resolveActionTimeline` clamped `durationMs` to the cap but the pass still labeled it `'stored'`, so a consumer would read the cap as the clip's true length. → now labeled `'capped'` when the resolved value exceeds the cap.
2. **[minor] Untested branches.** `durationSource` `'capped'`/`'zero'`, the `AssetPlanner.unique()` collision suffixing, and cross-elementId video dedup were unpinned. → added assertions for all (107 tests total).
3. **[nit] Video null-media vs absent-bytes collapsed.** `media(...) === null` (no asset referenced) and `present: false` (referenced but bytes missing) emitted the same diagnostic; audio already distinguishes `missing-audio` vs `skipped-media`. → video now emits distinct messages for the two cases.
4. **[nit] eslint message contradicted the rule.** The allowlist message claimed "no parent-escape into the app" while the regex intentionally permits relative paths (documented tradeoff). → message reworded to match the intended rule (the `@/` ban + render-backend ban are the purity teeth).

### Known latent (not changed)
5. **[nit] Dedup assumes consistent `present` per `assetId`.** If two references to one `assetId` disagreed on `present`, the first owns the path and the dedup entry could claim a different `present`. In practice `AssetSource` returns consistent metadata per id, so this is latent only; noted for the P1d collector.

wyuc (changes requested):
Reviewed commit `511a68a`.

Requesting changes because the current IR is not yet sufficient as the standalone input to the P1d emitter in #865.

Blocking findings:

1. **[P1] `VideoSegment` has no placement geometry.** It keeps `elementId`, timing, and `assetRef`, but omits position, size, and rotation. The P1d emitter therefore cannot place the real video at the authored element position using only the IR; it would have to reread the scene DSL.

2. **[P1] An unavailable video can become a five-minute blank segment instead of being skipped.** An unresolved duration is converted to the five-minute cap before asset availability is evaluated. When asset planning later finds no media or `present: false`, it only emits a diagnostic and leaves the timeline unchanged, shifting every subsequent action by up to 300 seconds. This conflicts with #854's `skip + diagnostic` failure behavior.

3. **[P1] Per-action spotlight/laser values are lost.** Timeline compilation stores the effect descriptor ID but drops `SpotlightAction.dimOpacity` and `LaserAction.color`, both of which playback honors. Descriptor defaults cannot recover authored overrides, so an IR-only emitter renders different behavior.

Additional findings:

4. **[P2] Referenced-but-missing video is not represented structurally.** Unlike audio, `present: false` video gets neither an `AssetPlanEntry` nor `assetId/present` state on the segment. Downstream code can distinguish it from “no media association” only by parsing free-form diagnostics.

5. **[P2] Asset extensions need sanitization.** Manifest `format` is inserted directly into the target path. A value such as `../../escape` produces `media/v../../escape`; the ZIP-writing stage should receive only allowlisted/sanitized extensions and traversal-free paths.

6. **[P3] The purity lint boundary allows parent-relative escape imports.** The selector permits every source beginning with `.`, so a pass can import `../../components/foo` without an ESLint error. Restrict relative imports to the video-export module plus the explicit choreography dependency.

Verification: 107 related tests, `tsc --noEmit`, scoped ESLint, Prettier, and the repository CI/E2E checks pass. The findings above are contract counterexamples not covered by the current tests.

cosarah:
Thanks @wyuc — all six findings addressed in `317b191`. Summary:

**P1**
1. **`VideoSegment` placement.** `VideoSegment` now carries `geometry` (0–100 space) + `rotate` (degrees), resolved by the geometry pass (new `findElementPlacement` / `resolveVideoPlacement`). The emitter can place the clip from the IR alone. A missing element degrades to `geometry: null, rotate: 0, degraded: true` with an `unresolved-element` diagnostic.
2. **Unavailable video no longer becomes a 5-minute blank.** Video availability is now resolved up front from the `AssetSource` (in `compile.ts`) and passed to the timeline options, so an unavailable `play_video` gets a **0ms** dwell (`durationSource: 'skipped'`) *before* `resolveActionTimeline` fixes timing — later actions are not shifted. This is the `skip + diagnostic` behavior from #854. (Availability must be decided pre-timeline because asset planning runs after layout; the fix threads it through rather than patching durations afterward.)
3. **Per-action effect overrides.** `EffectSegment` now carries an effective `params` map: descriptor defaults merged with the authored overrides playback honors (`SpotlightAction.dimOpacity` → `dimness`, `LaserAction.color`). An IR-only emitter reproduces authored behavior, not just defaults.

**P2**
4. **Referenced-but-missing video is structural.** A `present: false` video now gets both an `assetId` + `present` on the segment **and** an `AssetPlanEntry`. Downstream code distinguishes "referenced but missing" (has `assetId`) from "no media association" (no `assetId`) without parsing diagnostics — mirroring the audio path.
5. **Extension sanitization.** Asset extensions are now reduced to a bare, lowercased, traversal-free token before entering the zip path, so a hostile `format` like `../../escape` yields `media/clip.escape`, not `media/clip../../escape`.

**P3**
6. **Lint boundary tightened.** The relative-import allowlist no longer permits any `.`-prefixed path. Now only `./…`, a single-level `../…` that stays inside the module, and the explicit `../../choreography` are allowed; a `../../` escape into the rest of the app (e.g. `../../components/foo`) is rejected. Verified the boundary bites on that exact case while `../../choreography` still passes.

**Tests:** +7 specs pinning each fix (video placement + rotation, skip-without-timeline-shift, effect param merge, structural missing-video, extension sanitization, degraded video placement). 115 pass; `tsc --noEmit`, scoped ESLint, and Prettier all clean.

wyuc (changes requested):
Re-reviewed commit `317b191`. Four of the six original findings are fixed, but two reproducible issues remain.

1. **[P1] Video availability is conflated by global `action.id`.** `resolveAvailableVideos` stores availability in a `Set<string>` keyed only by action ID, while the DSL and normalization pass do not enforce stage-wide action-ID uniqueness. With two scenes whose `play_video` actions both have ID `duplicate` — the first media missing with unknown duration, the second media present — the second marks that ID available. The first segment then retains a 300000ms dwell. Asset planning later stamps it `present: false` and `durationSource: skipped` without changing `durationMs`, producing a contradictory IR and shifting the next scene by five minutes. Key availability by action identity or `sceneId + actionIndex`, or validate global uniqueness explicitly.

2. **[P2] The purity lint rule still permits escapes from module-root files.** Allowing every single-level `../...` is correct for files under `lib/video-export/passes`, but from `lib/video-export/compile.ts` it leaves the module. I verified that `import "../action/engine";` in a virtual `compile.ts` produces zero ESLint messages. The rule needs directory-depth-specific overrides (or an equivalent path-aware boundary), so root files and nested passes do not share the same relative-path allowance.

Verified on the exact head: 115 related tests pass, along with `tsc --noEmit`, scoped ESLint, Prettier, and remote CI/E2E. The duplicate-ID regression above fails against the current implementation; the temporary test was removed and the worktree is clean.

cosarah:
Thanks for the re-review @wyuc — both remaining findings fixed in `bde2428`.

**1. [P1] Video availability keyed by action identity, not id.** You're right that action IDs aren't stage-unique. `resolveAvailableVideos` now stores the availability `Set` keyed by the **action object reference** rather than `action.id`. `resolveActionTimeline` receives these same normalized action objects and passes each back to `getVideoDurationMs` by reference, so the identity lookup is exact — two scenes sharing an id (e.g. `dup`) can no longer conflate. Added a regression test with your exact scenario (scene 0 `dup` media absent, scene 1 `dup` media present): the first clip is skipped to 0ms and scene 2 starts at 0, not 300000.

**2. [P2] Lint boundary is now depth-specific.** A single `../…` genuinely means different things at different depths, as you noted. The boundary is split into two disjoint file scopes:
- **Root** (`lib/video-export/*`): relative imports may only be `./…` or `../choreography`. Any other `../…` (e.g. `../action/engine`) is rejected.
- **Passes** (`lib/video-export/passes/**`): `./…`, a single in-module `../…`, and `../../choreography`; a `../../` escape is rejected.

Verified on virtual files: root `import '../action/engine'` → error; passes `import '../../action/engine'` → error; while `../choreography` (root), `../ir` + `../../choreography` (passes), and all real imports pass.

Verification: 116 tests pass (+1 duplicate-id regression), `tsc --noEmit`, scoped ESLint, and Prettier all clean.

wyuc (approved):
Re-reviewed commit `bde2428`. Both remaining findings are resolved: video availability is keyed by action identity, with a regression test covering duplicate IDs across scenes, and the purity boundary now has depth-specific root/pass ESLint rules that reject both verified escape paths while preserving valid internal imports. Verified 116 related tests, `tsc --noEmit`, scoped ESLint, Prettier, and remote CI/E2E. The #864 contract wording was also aligned to distinguish effective action-level overrides from non-inlined descriptor tracks/keyframes.

PR #914, opened 2026-07-13
feat(chat): add experimental Pi classroom runtime behind a flag

wyuc:
## Summary
- add an experimental server-side Pi director loop for classroom chat
- preserve the existing `/api/chat` path unless `NEXT_PUBLIC_PI_CHAT_ENABLED=true`
- add soft-close, follow-up, and interrupted-lecture auto-resume behavior
- add bounded tool/runtime guards and classroom action coverage

## Feature flag
Disabled by default. Set `NEXT_PUBLIC_PI_CHAT_ENABLED=true` at build time to enable both the Pi client path and server route.

## Validation
- `pnpm test` (273 files, 2288 tests)
- `pnpm exec tsc --noEmit`
- `pnpm check:i18n-keys`
- `pnpm build`


wyuc (commented):
Codex-only review loop complete (no Claude reviewer).

Multiple independent review/fix/re-review passes found and addressed:
- serialized async whiteboard actions and made flush single-flight/cancellation-aware;
- retired interrupted buffers and waited for in-flight ActionEngine state mutations before new requests or scene transitions;
- hardened whiteboard edit/delete/draw/table validation and runtime/ledger consistency;
- rejected malformed Pi `agentIds` with 400 responses;
- guarded async scene switching against stale continuations.

Final fresh Codex pass: **no actionable findings**.

Validation on head `76d0e9b1`:
- `pnpm check`: pass
- TypeScript: pass
- focused regression suite: 51/51 pass
- full local suite: 2302 pass; the only local failure is an environment-specific provider-config fixture
- GitHub CI: CodeQL, E2E, and Lint/Typecheck/Unit Tests all pass

Conclusion: ready for human end-to-end smoke/review.

wyuc:
Final-head follow-up: `9c4deed1` also routes live-session error cleanup through the same action-retirement barrier, so an immediate retry cannot overtake a stale whiteboard mutation. The prior Codex-only review conclusion is unchanged: **no actionable findings**.

Final GitHub checks are all green: CodeQL, E2E Tests, and Lint/Typecheck/Unit Tests.

wyuc (commented):
Codex-only review loop completed (Claude was intentionally not used). The final independent Codex pass found no actionable issues.

Issues found and fixed during the loop covered: async action/flush ordering; interruption, error, soft-close, and scene-transition retirement; cancellable background video playback; whiteboard edit/delete/table validation; stable element and code-line IDs across server/runtime/client/ledger; initial-snapshot ledger replay; truncated SSE handling; and strict agentIds validation. Historical whiteboard actions remain persisted in directorState but are no longer replayed twice in child prompts.

Validation:
- 138 focused Pi chat/playback tests passed
- TypeScript, ESLint, Prettier, and git diff checks passed
- GitHub CodeQL, Lint/Typecheck/Unit, and E2E checks passed on 4c1ce24c
- Earlier local full suite: 2302/2303 passed; the sole failure was the environment-only provider-config fixture caused by local provider configuration

Conclusion: ready for maintainer review. Residual risk is limited to timing-sensitive real-browser media/SSE races beyond the existing unit and CI E2E coverage.

cosarah (commented):
I found two substantive issues in this PR.

1. Same-request whiteboard follow-up actions are not actually usable across child agents. `buildVirtualWhiteboardContext` keeps `elementId` in memory, but the prompt text it generates does not expose those IDs back to later child agents, and code blocks only include a short line preview. At the same time, the runtime validators for `wb_delete` and `wb_edit_code` require exact `elementId` / `lineId` values. In practice this means one child agent can draw something, but a later child agent in the same Pi loop usually cannot reliably edit or delete it because the identifiers are no longer available in prompt-visible form. The relevant paths are `lib/orchestration/summarizers/whiteboard-ledger.ts`, `lib/chat/pi/prompts.ts`, and `lib/chat/pi/tools/classroom-actions.ts`.

2. `directorState.whiteboardLedger` now accumulates across turns without a clear bound, but the Pi path itself explicitly replays only the mutations produced during the current request on top of the request-start store snapshot. That means historical ledger entries are still persisted into the session and sent back on follow-up requests, but they are not actually needed for the Pi child prompts. For code-heavy whiteboard usage, this can grow the stored session state and request payloads unnecessarily because `wb_draw_code` / `wb_edit_code` carry code content and line-id metadata forward every turn. The relevant paths are `lib/chat/pi/director-loop.ts` and `components/chat/use-chat-sessions.ts`.

I could not run the targeted tests locally in this worktree because the repo does not have `node_modules` installed here, so I am basing this review on the code paths and the existing test coverage in the PR.

LING-6150:
@sophietao20-star  Two issues identified in previous testing have been fixed in the latest commit.  Please focus on verifying that:
Chat bubbles always contain clean, complete natural-language text with no JSON residue.
Teacher explanations are not cut off midway, and agents do not switch prematurely.
Multi-agent handoffs or prompts for user input happen naturally only after a meaningful explanation is completed.
Whiteboard actions still execute correctly.
Please have the agent check out and start the latest HEAD of [OpenMAIC PR #914](https://github.com/THU-MAIC/OpenMAIC/pull/914). Do not use main or any previously cached checkout.
Before starting, please confirm that:
The repository is THU-MAIC/OpenMAIC.
The current branch is codex/experimental-pi-chat.
The current HEAD includes the latest fix commit I just pushed.
The existing working .env.local is reused without modifying or overwriting any keys.
Pi runtime is enabled.
After startup, please provide the local access URL.

LING-6150:
@cosarah  Thanks for the review. Both issues have been addressed in the latest commita ：5c7719 (2cc66d5..a5c7719):
The Pi whiteboard context now exposes element IDs and code line IDs to later child agents, with shared character budgets to keep the prompt bounded.
directorState.whiteboardLedger now returns only the current request’s mutations; historical board state continues to come from the request-start storeState snapshot.
Regression coverage was added for cross-agent targeting context, bounded multi-block/long-ID rendering, and retaining current-turn ledger actions while dropping carried history. Relevant tests pass locally (115/115). Could you please take another look?

cosarah (commented):
I found two remaining issues after the latest fix.

1.  still spends the shared code-line budget in snapshot order. Because  seeds pre-existing whiteboard code blocks first and new turn mutations are appended later, a large existing block can consume the content/id budget before this turn's newly drawn block is rendered.  still requires exact visible line ids (), so a later child agent can still fail to edit the new block in the same Pi loop when the board already has a large code block. The new test at  currently codifies that behavior.

2. The new code-line cap is not end-to-end for the actual child prompts.  is bounded now, but both  and  still prepend , and  still dumps up to 10 raw code lines per code block from  with no shared budget or per-line truncation. So a code-heavy board from prior turns can still expand the prompt substantially even if the new "Whiteboard Changes This Round" section stays capped. The updated tests only cover , so they would not catch this path.

I couldn't run the targeted tests locally in this worktree because  is missing here, so this review is based on the current code paths.

cosarah (commented):
I found two remaining issues after the latest fix.

1. lib/orchestration/summarizers/whiteboard-ledger.ts:190-323 still spends the shared code-line budget in snapshot order. Because getInitialWhiteboardElements() seeds pre-existing whiteboard code blocks first and new turn mutations are appended later, a large existing block can consume the content/id budget before this turn's newly drawn block is rendered. wb_edit_code still requires exact visible line ids (lib/chat/pi/tools/classroom-actions.ts:483-527), so a later child agent can still fail to edit the new block in the same Pi loop when the board already has a large code block. The new test at tests/lib/orchestration/whiteboard-ledger.test.ts:216-249 currently codifies that behavior.

2. The new code-line cap is not end-to-end for the actual child prompts. buildVirtualWhiteboardContext() is bounded now, but both lib/chat/pi/prompts.ts:132-134 and lib/orchestration/prompt-builder.ts:153-154 still prepend buildStateContext(), and lib/orchestration/summarizers/state-context.ts:59-70,220-226 still dumps up to 10 raw code lines per code block from stage.whiteboard with no shared budget or per-line truncation. So a code-heavy board from prior turns can still expand the prompt substantially even if the new Whiteboard Changes This Round section stays capped. The updated tests only cover buildVirtualWhiteboardContext, so they would not catch this path.

I couldn't run the targeted tests locally in this worktree because node_modules is missing here, so this review is based on the current code paths.

LING-6150:
@cosarah   Thanks for the follow-up review. Both remaining issues have now been addressed in the latest commit.
Code-line budget allocation now prioritizes code blocks most recently drawn or edited in the current Pi request. Large snapshot blocks and earlier same-request blocks can no longer consume the budget before the latest actionable block exposes its elementId and lineIds.

The request-start whiteboard summary produced by buildStateContext() now uses the same shared code-line character budget as buildVirtualWhiteboardContext(). This bounds the code-line payload across the actual child prompt path, not only the per-request ledger section.

Both summarizers now share one truncation implementation to prevent their behavior from drifting apart.
Regression coverage now includes:
Large snapshot code plus current-request code
Multiple current-request code blocks with most-recently-touched priority
Shared bounds across multiple snapshot blocks
End-to-end child prompt growth from 5,000 to 50,000 snapshot lines
Preservation of current-request elementId and lineIds
Local Pi/orchestration regression tests pass (121/121), along with TypeScript, Prettier, and diff checks.
Could you please take another look at the latest HEAD?

cosarah (commented):
I found one remaining issue after the latest update.

1. Cross-request code editability is still biased toward older snapshot blocks. The Pi loop now intentionally returns only the current turn's whiteboard ledger (lib/chat/pi/director-loop.ts:165-170), so on the next request the only prompt-visible source for previously drawn code blocks is buildStateContext() (lib/chat/pi/prompts.ts:132-134 and lib/orchestration/prompt-builder.ts:153-154). That summary now uses the bounded renderer, but it still spends the shared budget in raw element order (lib/orchestration/summarizers/state-context.ts:93-99,230-235), and persisted whiteboard elements are appended in creation order (lib/api/stage-api-whiteboard.ts:155-167). In practice, an older large code block can still consume the budget before a newer block from a prior request gets its line ids rendered, even though wb_edit_code still requires exact line ids at runtime (lib/chat/pi/tools/classroom-actions.ts:483-527). The new tests cover this-round priority and bounded size, but I do not see coverage for the persisted snapshot case.

I could not run the targeted tests locally because node_modules is missing in this checkout, so this review is based on the current code paths.

LING-6150:
@cosarah  Thanks, this remaining case is fixed in the latest commit.
Persisted whiteboard code blocks now receive the shared code-line budget in reverse creation order, matching the append-only element storage path. Display order remains unchanged, and slide summaries still use source order.
I also added coverage for the cross-request path with an empty ledger: a newer persisted code block remains editable even when an older large block exists.
Local Pi/orchestration tests pass (123/123), together with TypeScript and formatting checks. Could you please take another look?

sophietao20-star:
Verified both bugs using the same course and prompt under controlled conditions. Unable to reproduce either issue; fix validated.

LING-6150:
@wyuc      A quick follow-up on your earlier whiteboard lifecycle comment: the previous change correctly removed automatic clearing for soft-closing follow-ups, so continued questions within the same discussion preserve the board.
During further product testing, I found that the complementary lifecycle rule for a genuinely new discussion had not yet been implemented deterministically. After a manual stop, especially when switching slides, a new discussion could reuse the old stage-level whiteboard unless the model happened to issue wb_clear.
Commit fc3111e completes that lifecycle behavior: same-session follow-ups still preserve the board, while a new session after manual stop receives a guarded whiteboard boundary. It does not restore the old follow-up auto-clear mechanism.

wyuc (commented):
@LING-6150 Thanks for investigating and documenting this case. My product preference is **not** to introduce a deterministic `manual stop -> guarded clear before first draw` rule.

A manual stop or a new UI session is not necessarily a semantic boundary: the next discussion may still need to reference, extend, or edit the existing whiteboard. The intended behavior is to provide the current whiteboard state, current slide, and session-boundary signal to the child agent, then instruct the model to call `wb_clear` when it is semantically appropriate, such as when the topic is unrelated, the existing board would cause confusion, or there is insufficient space. Otherwise it should preserve the board.

Execution-safety mechanisms such as stale-action and fingerprint guards are still useful, but they should guard an explicit model-selected action rather than inject a clear that the model did not request. Please remove the deterministic auto-clear behavior from `fc3111e` and handle this choice through context + system prompt instead.

LING-6150:
@wyuc  Thanks for the clarification. I’ve revised Round 2b to match your intended behavior.
Removed the deterministic manual stop → first draw auto-clear path, including the injected wb_clear, fingerprint latch, and guarded-clear metadata.
Added a read-only, first-request session-boundary signal to the child-agent context.
The child agent now receives the current slide and persisted whiteboard state, and decides whether to explicitly call wb_clear.
The prompt preserves the board when the discussion references, extends, or edits existing content, and suggests clearing only when the topic is unrelated, the board would be confusing, or space is insufficient.
A new UI session or slide change alone no longer implies clearing.
The existing explicit whiteboard-action validation remains intact, and the soft-closing behavior is unchanged. The branch has also been merged with the latest main; full tests pass (2898 passed, 4 skipped). Could you please take another look at the latest head?  The fix is commit 5494340ef7e8d215f743f40c8db5857464aa2cbb

cosarah (approved):
LGTM

PR #917, opened 2026-07-13
fix(roundtable): keep the non-presentation input panel inside its card

### LINKED ISSUE #906: [Bug]: Chat panel bubbles collapse line breaks in user messages (missing whitespace-pre-wrap)
### Bug Description

User messages typed with line breaks (Shift+Enter) in the classroom's roundtable input lose their line breaks when displayed in the right-side chat panel — multi-line messages render as one flattened line.

Image

Image

The line breaks actually survive the whole data path; they are only lost at render time:

1. **Input already supports newlines.** The roundtable input is a `<textarea>` (`components/roundtable/index.tsx` ~L740 and ~L1255). `onKeyDown` only `preventDefault()`s a bare Enter to send; Shift+Enter inserts a newline as expected.
2. **The send path doesn't touch the text.** `handleSendMessage` → `onMessageSend` → `sendMessage` (`components/chat/use-chat-sessions.ts` ~L996) passes `content` through unchanged (`trim()` is only used for the empty check).
3. **The newline is dropped by the chat-panel bubble.** In `MessageBubble` (`components/chat/chat-session.tsx` ~L119) the text is rendered inside `<span className="break-words">` with no `whitespace-pre-wrap`, so HTML whitespace collapsing turns `\n` into a space.

Notably, the on-stage roundtable bubble already renders with `whitespace-pre-wrap break-words` (`components/roundtable/index.tsx` ~L1682), so the two surfaces are inconsistent: the same message shows line breaks on stage but not in the side panel.

### Steps to Reproduce

1. Open a classroom and start playback.
2. Open the text input (T) and type a message containing Shift+Enter line breaks.
3. Send it, then look at the message in the right-side chat panel (Chat tab).

### Expected Behavior

Line breaks in the user's message are preserved in the chat-panel bubble, same as in the on-stage roundtable bubble.

### Actual Behavior

The message renders as a single line; `\n` is collapsed into spaces.

### Suggested Fix

Add `whitespace-pre-wrap` to the text wrapper in `MessageBubble` (`components/chat/chat-session.tsx` ~L119). This also lets intentional paragraph breaks in agent/teacher replies render properly in the panel.

### Related: input panel outgrows the bottom bar and gets clipped at the top

While testing multi-line input we hit a second, separate bug in the non-presentation roundtable input:

Image

**Root cause (corrected after live measurement — earlier revisions of this section blamed `field-sizing`/`max-height`, which turned out to be innocent):**

- The non-presentation roundtable root has a fixed height (`h-[192px]`, `components/roundtable/index.tsx` ~L1056); the card inside it measures ~151px and has `overflow-hidden`.
- The input panel adds ~32px of its own padding/border around the textarea and is anchored `bottom-4` (16px) inside that card.
- The textarea's height cap was `max-h-[120px]` — so at 6+ lines the panel reaches ~152px + 16px offset, which **exceeds the 151px card**. The card's `overflow-hidden` clips the panel's top: the top border and the first line(s) of the draft become invisible/unreachable (measured: panel top 18.5px above card top in Chrome 148).
- `field-sizing: content` and `max-height` both behave correctly; the cap simply never fit the fixed-height card it lives in.

Expected: the input panel (textarea + its chrome + bottom offset) always fits fully inside the bottom bar card; once content exceeds the cap, the textarea scrolls internally.

### Optional follow-up

Consider collapsing excessive blank lines before sending (`content.replace(/\n{3,}/g, '\n\n')`) so a run of empty Enters can't produce a giant mostly-blank bubble.

### Deployment Method

Local development (pnpm dev)

YizukiAme:
Fixes #906 (task 2 of 2 — the input-panel clipping bug; task 1, the chat-panel line-break fix, was merged in #908).

## Root cause

The non-presentation roundtable root has a fixed height (`h-[192px]`); the card inside it measures ~151px and has `overflow-hidden`. The input panel wraps the textarea in ~32px of padding/border and is anchored `bottom-4` (16px) inside that card. With the textarea capped at `max-h-[120px]`, a 6+-line draft makes the panel ~152px + 16px offset — taller than the 151px card — so the card clips the panel's top: the top border and the first line(s) of the draft become invisible (measured: panel top 18.5px above card top in Chrome 148).

Notably, `field-sizing: content` and `max-height` were **not** at fault — the cap worked, it just never fit the fixed-height card around it.

## Changes

- Cap the non-presentation textarea at **100px** (5 lines at 20px line-height): panel 132.5px + 16px offset = 148.5px ≤ 151px card, so it always fits. The constant `NON_PRESENTATION_INPUT_MAX_HEIGHT_PX` and the `max-h-[100px]` class carry a sync comment.
- Replace `field-sizing: content` with the effect-based JS autosize already used by the PBL v2 chat input (`height = min(scrollHeight, MAX)`), with `overflow-y-auto` for internal scrolling once capped. The effect's deps (`inputValue`, `isInputOpen`, `isPresenting`) also cover remounts: the panel unmounts on Escape while keeping the draft, so reopening must re-apply the height to a fresh DOM node.
- Add `data-testid`s to the card / input stage / panel so the regression test can assert geometry robustly.
- The presentation-mode (fullscreen) input is untouched.

## Verification

- New Playwright test (reusing the existing classroom IndexedDB fixture) drives the real flow: open input with `T`, fill a 24-line draft, and assert the load-bearing behavior — **the panel stays fully inside the overflow-hidden card** (`panelTop >= cardTop`) — plus: capped at 100px, internal scroll active, height shrinks when content is deleted, and the draft persists across Escape/reopen with the height re-applied. Metrics are attached to the test report.
- Negative control: replicating the old 120px cap makes the panel-containment assertion fail with exactly the observed clipping (panel top 543 < card top 561, the 18px overflow).
- Live headed-browser check on this branch: 6 long wrapped lines → panel fully visible inside the bar (20px headroom), 720px of content scrolling inside a 100px textarea, `field-sizing` gone.
- `vitest` (2205 tests), `tsc --noEmit`, and `eslint` are clean.

### DIFF
--- a/components/roundtable/index.tsx
+++ b/components/roundtable/index.tsx
@@ -93,6 +93,9 @@ interface RoundtableProps {
   readonly fullscreenContainerRef?: React.RefObject<HTMLDivElement | null>;
 }
 
+// This must stay in sync with the non-presentation textarea's max-h-[100px] class.
+const NON_PRESENTATION_INPUT_MAX_HEIGHT_PX = 100;
+
 const VOICE_WAVE_BARS = [
   { peak: 18, duration: 0.55 },
   { peak: 24, duration: 0.72 },
@@ -191,11 +194,24 @@ export function Roundtable({
   const [isVoiceOpen, setIsVoiceOpen] = useState(false);
   const [inputValue, setInputValue] = useState('');
   const [userMessage, setUserMessage] = useState<string | null>(null);
+  const nonPresentationInputRef = useRef<HTMLTextAreaElement>(null);
   const agentScrollRef = useRef<HTMLDivElement>(null);
   const bubbleScrollRef = useRef<HTMLDivElement>(null);
   const teacherAvatarRef = useRef<HTMLDivElement>(null);
   const studentAvatarRefs = useRef<Map<string, HTMLDivElement>>(new Map());
   const userMessageClearTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  useEffect(() => {
+    if (isPresenting) return;
+    const textarea = nonPresentationInputRef.current;
+    if (!textarea) return;
+
+    textarea.style.height = 'auto';
+    textarea.style.height = `${Math.min(
+      textarea.scrollHeight,
+      NON_PRESENTATION_INPUT_MAX_HEIGHT_PX,
+    )}px`;
+  }, [inputValue, isInputOpen, isPresenting]);
 
   // End flash visible state (Issue 3)
   const [endFlashVisible, setEndFlashVisible] = useState(false);
@@ -1225,6 +1241,7 @@ export function Roundtable({
           </AnimatePresence>
 
           <div
+            data-testid="roundtable-non-presentation-card"
             onClick={() => {
               if (isInputOpen || isVoiceOpen) {
                 setIsInputOpen(false);
@@ -1239,6 +1256,7 @@ export function Roundtable({
               {isInputOpen && (
                 <motion.div
                   key="input-stage"
+                  data-testid="roundtable-non-presentation-input-stage"
                   initial={{
                     opacity: 0,
                     scale: 0.95,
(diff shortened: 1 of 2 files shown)


wyuc (approved):
Reviewed head `7a43e5c`. The textarea autosize cap now fits the fixed roundtable card, preserves drafts across close/reopen, and the new geometry-focused Playwright regression covers the observed clipping failure. Current CI is green. LGTM.

Issue #43, opened 2026-03-16
feat: Question Bank Reference for course generation


YizukiAme:
## Summary

Add a **Question Bank Reference** option in the Advanced Settings panel during course generation. This allows users to provide reference questions (via file upload or pasted text), which the AI agents can then use to:

- **Directly include** selected questions from the reference bank
- **Reference the question style/format** to generate new, similarly structured questions

## Motivation

Educators often have existing question banks, past exams, or curated problem sets that reflect their desired difficulty level, topic coverage, and question format. Currently, OpenMAIC generates all questions from scratch, which means the output may not align with the instructor's established assessment style.

By allowing users to provide reference questions, the AI can:
- Maintain consistency with existing course materials
- Match the expected difficulty and format of real exams
- Save time by reusing proven, well-crafted questions where appropriate

### Input
- A new section in **Advanced Settings** (alongside existing options like web search)
- Supports:
  - **File upload** (PDF, TXT, MD — reusing existing upload infrastructure)
  - **Paste text** directly into a text area
  - Multiple reference sources allowed

### Behavior
- When reference questions are provided, they are injected as additional context into the AI generation prompt
- The AI naturally decides whether to draw directly from the bank or use the questions as stylistic/structural references
- No complex mode selection needed — the AI handles it intelligently based on the provided material

### UI Placement
- Within the generation page's advanced/optional configuration area
- Toggle to enable/disable the feature
- Clear indicator showing how many reference questions/files are loaded

## Additional Context

This feature complements the existing web search integration — while web search provides broad topic context, question bank reference provides specific assessment-level guidance.

wyuc:
Thanks for the suggestion. As decided when #44 was closed, we're favouring editing after generation (e.g. adjusting quizzes in Pro mode) over adding a question-bank input before generation.

Closing as out of scope for the project; thanks for understanding.

PR #920, opened 2026-07-14
feat(document): add transform pipeline foundation

jackefn:
## Summary

Implement the foundation of MAIC ETL Milestone 4 by introducing a provider-neutral document transform pipeline between document extraction and bundle/generation assembly.

This PR adds a small OpenMAIC-owned transform contract, execution history, failure policies, cancellation support, and a registry for composing preprocessing steps. It also provides the first deterministic transforms for content normalization and conservative repeated-noise removal.

The pipeline operates on `DocumentArtifact` and preserves the existing extraction, document bundle, PDF compatibility, and generation behavior. It establishes a reusable extension point for future document-processing capabilities without defining document structure, generating outlines, or adding AI/UI features in this focused foundation PR.

The original document artifact remains the source of truth. Any future outline, chapter tree, chunk plan, summary, or index should be introduced separately as an optional, versioned, and rebuildable projection with explicit evaluation criteria and a concrete consumer.

## Related Issues

Related to #52, #335, #43

## Changes

- Extend `DocumentArtifact` with optional transform execution history.
- Add a provider-neutral `DocumentTransform` contract with declared capabilities and transform context.
- Add `transformDocument()` for ordered transform execution over cloned artifact state without mutating caller-owned input.
- Support `fail-fast` and `best-effort` failure policies.
- Record transform version, status, timestamps, input/output counts, options, and diagnostics on the output artifact.
- Preserve accumulated transform history even when a transform returns a newly constructed artifact.
- Add cancellation checks through `AbortSignal` before and between transform steps.
- Add aggregate text and asset metrics for the complete pipeline.
- Add `DocumentTransformRegistry` and a deterministic default transform order:
  - `normalize`
  - `remove-noise`
- Add a normalization transform that:
  - removes invalid control characters and normalizes whitespace and newlines;
  - preserves indentation-significant Markdown content;
  - removes empty text and Markdown blocks while retaining structured blocks;
  - merges compatible adjacent text blocks;
  - remaps citation block references after merging;
  - removes citation references to blocks removed during normalization.
- Add a conservative noise-removal transform that:
  - removes bare page numbers only when explicitly identified with a `page-number` role;
  - removes decorated page-number forms only with header/footer positional or role evidence;
  - removes repeated headers and footers only when they repeat across enough document pages;
  - does not remove repeated body content without header/footer evidence;
  - removes stale citation references when a noise block is removed.
- Export the transform APIs from the existing `lib/document` boundary.
- Add focused tests for execution order, input immutability, transform history, failure policies, cancellation, registry behavior, citation remapping, normalization behavior, and conservative noise removal.
- Keep structure detection, outline generation, logical fallback sections, and numbered-heading detection out of this PR.

## Type of Change

- [x] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Install the repository dependencies from the lockfile.
2. Run the four focused document-transform test files.
3. Run TypeScript type checking.
4. Run the repository formatting checks.
5. Confirm that existing extraction, compatibility, bundle, and generation callers compile without adopting the optional transform history field.
6. Confirm that the current GitHub CI checks pass.

- Transforms execute in declared order over cloned artifacts without mutating caller-owned block or metadata state.
- Successful, skipped, and failed transform steps produce inspectable execution records.
- Transform history is accumulated by the pipeline and survives transforms that return newly constructed artifacts.
- Best-effort mode preserves the last successful artifact and continues after a failed transform.
- Fail-fast mode stops before later transforms run.
- An already-aborted signal prevents pipeline execution.
- Cancellation is checked between transform steps.
- The default registry executes normalization before conservative noise removal.
- Normalization removes invalid control characters and normalizes text whitespace and newlines.
- Markdown normalization preserves indentation-significant content.
- Empty text and Markdown blocks are removed without leaving dangling citation references.
- Compatible adjacent text blocks are merged and citation block references are remapped to the surviving block.
- Blocks carrying HTML payloads are not merged in a way that would discard their HTML content.
- Repeated headers, footers, and page numbers are removed only with sufficient positional or role evidence.
- Repeated body content remains untouched when header/footer evidence is absent.
- The focused transform suite passes all 17 tests across 4 files.
- TypeScript type checking passes.
- Prettier checks pass.
- The current CI checks pass for the revised scope.

### Evidence

- [x] Focused transform tests pass (17 tests across 4 files)
- [x] TypeScript type checking passes (`pnpm exec tsc --noEmit --pretty false`)
- [x] Formatting passes (`pnpm check`)
- [x] Current GitHub CI checks pass
- [ ] Manually tested through a product UI flow (not applicable: the pipeline is not wired into a product flow)
- [ ] Screenshots / recordings attached (not applicable: no UI changes)

Local verification:

```text
pnpm vitest run \
  tests/document/transform-pipeline.test.ts \
  tests/document/transform-registry.test.ts \
  tests/document/normalize-transform.test.ts \
  tests/document/noise-removal-transform.test.ts

pnpm exec tsc --noEmit --pretty false
pnpm check
````

## Compatibility

* The only new `DocumentArtifact` state introduced by this PR is optional transform execution history.
* `DocumentArtifact` does not gain an outline or canonical structure field.
* Existing extractor providers do not need to emit transform history or adopt the transform pipeline.
* Existing PDF compatibility adapters continue to consume and produce document blocks and assets as before.
* Existing document bundle and generation paths are not automatically routed through transforms in this PR.
* Existing artifacts and sessions remain valid without migration.
* Provider-specific raw output remains opaque and is not made part of the transform standard.
* The default transform registry contains only normalization and conservative noise removal.
* No chapter, outline, chunking, or semantic-structure behavior is introduced into existing product flows.

## Known Limitations / Follow-ups

* This PR does not detect document structure or generate a canonical outline.
* This PR does not create logical fallback sections.
* This PR does not implement numbered-heading or chapter-heading detection.
* This PR does not add chapter-selection UI or persistence.
* This PR does not call an LLM or implement hierarchical long-document summarization.
* This PR does not score or filter images with a vision model.
* This PR does not parse or shape question banks or reference materials.
* Transform execution is not yet wired into the generation preview, API, bundle, or other product flows.
* Future outlines, chapter trees, chunk plans, summaries, and indexes should be treated as optional, versioned, and rebuildable projections rather than as the canonical document representation.
* A future structure proposal should define explicit source information, confidence, stable content anchors, evaluation criteria, and a concrete consumer before being added to the default pipeline.
* Numbered-heading detection should include regression coverage ensuring body clauses such as `1、甲方同意……`, `1. 乙方负责……`, and `1) 设备安装……` are not incorrectly promoted to chapters.

## Checklist

* [x] My code follows the project's coding style
* [x] I have performed a self-review of my code
* [x] I have added/updated documentation as needed
* [x] My changes do not introduce new warnings

### CODE REVIEW

File: lib/document/transforms/noise-removal.ts [hunk 1]
```diff
@@ -0,0 +1,119 @@
...
+    const repeatedTexts = removableRepeatedTexts(artifact);
+    const removedBlockIds = new Set<string>();
+
+    artifact.blocks = artifact.blocks.filter((block) => {
+      const text = normalizedBlockText(block);
+      const standalonePageNumber =
+        Boolean(text) &&
+        PAGE_NUMBER.test(text) &&
+        (isExplicitHeaderOrFooter(block) || block.metadata?.role === 'page-number');
+      const repeatedHeaderOrFooter = Boolean(text) && repeatedTexts.has(text);
```
yanpgwang: **Repeated header/footer text set is matched against all blocks, not just header/footer blocks.**

`removableRepeatedTexts()` only collects text from blocks that pass `isExplicitHeaderOrFooter` (line 48). But this removal check applies `repeatedTexts.has(text)` to *every* block without re-checking header/footer evidence.

Failure case: a 3-page doc with a repeated header `Course Overview` (role `header`) on each page, plus a page-1 body paragraph whose text is exactly `Course Overview` (a section title, no bbox/role). `removableRepeatedTexts` collects `course overview` from the headers, and this line then deletes the body block too.

This contradicts the PR's stated invariant: *"does not remove repeated body content without header/footer evidence."*

Suggested fix: gate `repeatedHeaderOrFooter` on `isExplicitHeaderOrFooter(block)` as well.

lib/document/transforms/normalize.ts
```diff
@@ -0,0 +1,109 @@
...
+import type { DocumentTransform } from './types';
+import { cloneDocumentArtifact } from './utils';
+
+export function normalizeDocumentText(value: string): string {
+  return value
+    .replace(/\u0000/g, '')
+    .replace(/[\u0001-\u0008\u000B\u000C\u000E-\u001F\u007F]/g, ' ')
+    .replace(/\r\n?/g, '\n')
+    .replace(/[ \t]+\n/g, '\n')
+    .replace(/[ \t]{2,}/g, ' ')
```
yanpgwang, line 11: **Space/tab collapse corrupts indentation-significant markdown.**

`normalizeDocumentText` runs `.replace(/[ \t]{2,}/g, ' ')` over `markdown` block text (line 66 calls it for both `text` and `html`). Indented code blocks, fenced snippets, and ASCII tables lose their leading whitespace, producing syntactically broken code that then flows into generation/summaries.

Example: `    def f():\n        return 1` collapses to single-space indentation.

Consider skipping the space-collapse for `markdown` blocks (or preserving fenced code regions).

lib/document/transforms/structure.ts
```diff
@@ -0,0 +1,196 @@
...
+      id: `outline_${index + 1}`,
+      title: candidate.title,
+      level: candidate.level,
+      order: index + 1,
+      parentId,
+      blockIds: Array.from(new Set(sectionBlocks.map((block) => block.id))),
+      pageStart: candidate.block.pageNumber,
+      pageEnd: lastSectionBlock.pageNumber ?? candidate.block.pageNumber,
+      startOffset: candidate.startOffset,
+      endOffset:
```
yanpgwang: **A heading's `endOffset` can extend over its own child section.**

When two headings of increasing depth live in one block (e.g. `# Intro` at offset 0, then `## Details` at offset 20), the outer heading finds no same-or-shallower `nextPeer`, so this expression falls through to the `sectionBlocks.length === 1` branch and sets `endOffset` = full text length. The Intro node then claims `[0, len]`, which overlaps and swallows the Details subsection. Any consumer slicing text by `[startOffset, endOffset]` gets Details duplicated under both nodes.

lib/document/transforms/normalize.ts
```diff
@@ -0,0 +1,109 @@
...
+
+function remapArtifactReferences(
+  artifact: DocumentArtifact,
+  blockIdMap: ReadonlyMap<string, string>,
+): void {
+  artifact.citations = artifact.citations?.map((citation) => ({
+    ...citation,
+    blockId: citation.blockId ? (blockIdMap.get(citation.blockId) ?? citation.blockId) : undefined,
+  }));
+  artifact.outline = artifact.outline?.map((node) => ({
```
yanpgwang: **Block merge remaps `blockId`s but leaves outline `startOffset`/`endOffset` stale.**

When a provider-supplied outline carries `startOffset`/`endOffset` into a text block, and `normalize` later merges that block with a preceding one via `[prev.text, block.text].join('\n\n')`, the merged block's character positions shift. `remapArtifactReferences` updates the outline's `blockId` to the survivor but not its offsets, so downstream offset-based slicing reads the wrong span. Either recompute offsets on merge or drop offsets for remapped nodes.

lib/document/transforms/normalize.ts
```diff
@@ -0,0 +1,109 @@
...
+      if (!previous || !canMergeTextBlocks(previous, block)) {
+        merged.push(block);
+        continue;
+      }
+
+      const mergedSourceIds = [
+        ...((previous.metadata?.sourceBlockIds as string[] | undefined) ?? [previous.id]),
+        block.id,
+      ];
+      previous.text = [previous.text, block.text].filter(Boolean).join('\n\n');
```
yanpgwang, line 99: **`html` is dropped when merging two blocks that both carry it.**

`canMergeTextBlocks` passes as long as both blocks have string `text`, but the merge only concatenates `text`. If the merged-away `block` also has `html`, that payload is silently discarded (no diagnostic). Either merge `html` too, or refuse to merge blocks that carry `html`.

lib/document/transforms/structure.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  for (const [index, candidate] of candidates.entries()) {
+    const nextPeer = candidates.slice(index + 1).find((item) => item.level <= candidate.level);
+    const endBlockIndex = nextPeer?.blockIndex ?? blocks.length;
+    const sectionBlocks = blocks.slice(
+      candidate.blockIndex,
+      Math.max(candidate.blockIndex + 1, endBlockIndex),
+    );
+    const lastSectionBlock = sectionBlocks.at(-1) ?? candidate.block;
+    const parentId = parentAtLevel.get(candidate.level - 1);
```
yanpgwang: **Level-skipping headings are orphaned (altitude).**

`parentId` is looked up strictly at `candidate.level - 1`. Real Markdown that jumps `#` → `###` leaves the `###` heading with `parentAtLevel.get(2)` undefined, so it becomes a second top-level node instead of a child of the `#` heading — a flatter/incorrect hierarchy for downstream chapter selection. A nearest-shallower-ancestor lookup handles this generally rather than assuming contiguous levels.

yanpgwang (commented):
Code review of the transform pipeline foundation. Solid structure overall (clean provider-neutral contract, immutable artifact copies, good test coverage). I found a few correctness issues in the noise-removal and normalization heuristics where the implementation doesn't fully honor the invariants described in the PR. Details inline; #1 and #2 are the ones I'd consider blocking.

yanpgwang:
Re-reviewed after `902930dd` (fix: preserve transform content invariants). All six findings from my earlier review are addressed, and the fixes look correct:

1. **Repeated header/footer over-removal** — `repeatedHeaderOrFooter` now also requires `isExplicitHeaderOrFooter(block)`, so body blocks that merely share text with a header are no longer deleted. ✅
2. **Markdown indentation collapse** — `markdown` blocks now go through `normalizeMarkdownText` (control-char + newline normalization only), preserving significant leading whitespace. ✅
3. **Heading `endOffset` overlap** — `endOffset` now uses the immediately-following heading, so a section no longer extends over its nested child. ✅
4. **Stale outline offsets after merge** — nodes referencing a merged-away block now drop `startOffset`/`endOffset`. ✅
5. **`html` lost on merge** — `canMergeTextBlocks` now refuses to merge blocks carrying `html`. ✅
6. **Level-skipping headings orphaned** — parent lookup now walks to the nearest shallower ancestor instead of assuming contiguous levels. ✅

Verification: the 5 transform test files pass (20/20) locally, and targeted repro tests for #1/#2/#3/#6 confirm the corrected behavior. LGTM.

lib/document/transforms/normalize.ts
```diff
@@ -0,0 +1,129 @@
...
+              : normalizeDocumentText(block.text)
+            : undefined,
+        html: typeof block.html === 'string' ? normalizeDocumentText(block.html) : undefined,
+      }))
+      .filter((block) => !isEmptyTextBlock(block));
+
+    const removedEmptyBlocks = artifact.blocks.length - normalized.length;
+    const blockIdMap = new Map<string, string>();
+    const merged: DocumentBlock[] = [];
+    let mergedBlockCount = 0;
```
yanpgwang, line 86: **Removing empty blocks leaves dangling citation/outline references.**

`normalized` filters out empty text/markdown blocks (`!isEmptyTextBlock`), but `remapArtifactReferences` only handles the *merge* map (`blockIdMap`). A block that is dropped by the empty-filter is never in `blockIdMap`, so any `citation.blockId` or `outline.blockIds[]` still points at the removed block ID.

Failure: a block `{id:'empty', text:'   '}` referenced by a citation `{blockId:'empty'}` and an outline node `{blockIds:['empty','keep']}` — after normalize, the block is gone but both references still contain `'empty'`. Downstream lookups by that ID resolve to nothing.

Note `noise-removal.ts` already handles this correctly via `cleanReferences(artifact, removedBlockIds)`. `normalize` should collect the removed empty-block IDs and clean them the same way. (Reproduced with a unit test.)

lib/document/transforms/pipeline.ts
```diff
@@ -0,0 +1,102 @@
...
+        inputAssetCount,
+        outputAssetCount: output.artifact.assets.length,
+        options:
+          typeof context.options?.[transform.id] === 'object' && context.options[transform.id]
+            ? (context.options[transform.id] as Record<string, unknown>)
+            : undefined,
+        diagnostics,
+      };
+
+      current = cloneDocumentArtifact(output.artifact);
```
yanpgwang, line 59: **Pipeline can lose prior transform history.**

History is appended onto `output.artifact` (`current = clone(output.artifact); current.transforms = [...(current.transforms ?? []), record]`). This assumes every transform copies `input.transforms` into its returned artifact. The built-in transforms happen to satisfy this because they `cloneDocumentArtifact(input)`, but the `DocumentTransform` contract does not require it.

Failure: a transform that returns a freshly-constructed artifact (`{ metadata, blocks, assets }`) without copying `input.transforms` — the pipeline then appends onto an empty history, so all prior transform records are lost. Verified: a 2-step pipeline where step 2 rebuilds the artifact yields history `['step2']` instead of `['step1','step2']`.

Fix: accumulate history on the pipeline's own `current` (carry `current.transforms` forward) rather than trusting `output.artifact.transforms`. (Reproduced with a unit test.)

yanpgwang (commented):
Follow-up after testing this pipeline end-to-end against real extractions (unpdf + a real AliDocMind call on an 8-page paper). Two parts: (A) an architectural concern about whether the provider-neutral post-processing model holds up, and (B) two concrete data-integrity bugs that are independent of that discussion and would be wrong on any provider.

---

**A. Architectural concern — post-processing quality is entirely provider-dependent**

Different providers emit fundamentally different shapes: AliDocMind returns per-block `layout[]` with an explicit heading `level` and `type:'title'`; MinerU returns a single markdown blob; unpdf returns plain text with no structure at all. A single deterministic rule set has to be designed for the weakest of these, which produces three symptoms:

- **Rich structure from strong providers is discarded.** AliDocMind already computes a heading `level` per block, but it is dropped along the way: `aliDocMindLayoutsToParsedPdf` does not read `l.level`, `ParsedPdfContent.layout` has no `level` field, and `pdf-compat` only carries `layoutType` into block metadata. `structure.ts` then re-guesses levels by counting `#` / dotted numbers — even though `headingFromMetadata` already reads `metadata.headingLevel` and would trust it at `confidence: 1`. So the authoritative signal exists but is thrown away and re-derived less accurately.
- **Weak providers fall back to heuristics that cannot be trusted.** Numbered-heading detection only inspects the first line of each block; the logical fallback slices at a fixed 12k chars, cutting mid-section.
- **Multi-source structure is not reconciled.** On a real 8-page paper, AliDocMind produced **110 outline nodes, 56 unique, 40 duplicated** — the same heading appears both as `#` inside the text block and as a separate `layoutType:'title'` block, and both detection paths emit a node. (Reproduced end-to-end with a real AliDocMind call.)

Net: the "neutral" layer's output quality tracks the provider, and strong providers get pulled down to the weakest common denominator. Suggestion: have structure detection trust and reuse provider-native structure per-adapter (e.g. thread AliDocMind's `level` through to `metadata.headingLevel`) rather than re-deriving it generically, and add a dedup/merge step when multiple sources describe the same heading. Worth deciding whether this belongs in the foundation before chapter-selection UI is built on top of the outline.

---

**B. Two independent data-integrity bugs (must-fix regardless of A)**

Both are deterministic, reproduced with unit tests, and would be wrong on any provider. Details inline.

yanpgwang:
Could you add a cross-provider comparison report to this PR? The concern in (A) above is really about `structure.ts`, and the clearest way to confirm or refute it is to run the **full pipeline** (`transformDocument`: normalize → detect-structure → remove-noise) on the **same** document through each provider and compare the resulting outline.

Suggested shape — pick one document with a clear chapter/section structure, then for each of `unpdf` / `mineru` / `alidocmind` report:

- total outline nodes / unique titles / duplicated titles
- source breakdown (`heading` / `heuristic` / `logical` / `provider`)
- whether the detected levels match the real document hierarchy
- whether the logical fallback triggered, and where it cut

For reference, my end-to-end run against a real AliDocMind call on an 8-page paper produced **110 nodes / 56 unique / 40 duplicated** (same heading emitted once from the `#` in the text block and once from the `layoutType:'title'` layout block). It would be useful to see the `mineru` and `unpdf` numbers side by side — that comparison is the real test of whether a single provider-neutral rule set produces consistent structure across providers.

jackefn:
Implemented the follow-up review in `27cf97b`.

### Data-integrity fixes

- Normalization now tracks block IDs removed by the empty-block filter, removes citations that point only to those blocks, removes those IDs from outline nodes, drops outline nodes with no surviving block, clears invalid parent links, and invalidates offsets whenever referenced blocks change.
- The pipeline now carries transform history and prior diagnostics from its own `current` state. A transform may return a freshly constructed artifact without erasing earlier transform records.

### Provider-native structure and fallback behavior

- AliDocMind `layout[].level` is preserved through `ParsedPdfContent.layout[].level` and `pdf-compat` as `metadata.headingLevel`.
- If explicit provider heading levels exist, structure detection uses those authoritative layout headings and does not also emit headings parsed from the provider's markdown/text blob.
- Repeated native headings are deduplicated by page, level, and normalized title. This covers the reported AliDocMind shape where the same heading is represented in both text and title layout data.
- Layout titles without an authoritative level do not override a richer Markdown hierarchy (important for MinerU-style markdown).
- Numbered-heading detection scans all lines of a plain-text block instead of only its first line.
- Logical fallback still has a hard 12k ceiling, but now prefers a paragraph, line, or sentence boundary before that ceiling.

### Cross-provider comparison

I added a reproducible end-to-end comparison test for one three-section document, represented in the actual shapes consumed from unpdf (plain text), MinerU (single Markdown blob), and AliDocMind (Markdown plus native title layouts). Every case runs the full default pipeline: normalize -> detect structure -> remove noise.

| Provider shape | Nodes | Unique titles | Duplicate occurrences | Source | Levels | Hierarchy | Logical fallback |
|---|---:|---:|---:|---|---|---|---|
| unpdf plain text | 3 | 3 | 0 | heuristic: 3 | 1 -> 2 -> 1 | matches | no |
| MinerU Markdown | 3 | 3 | 0 | heading: 3 | 1 -> 2 -> 1 | matches | no |
| AliDocMind native layout | 3 | 3 | 0 | provider: 3 | 1 -> 2 -> 1 | matches | no |

The test also verifies the level-2 node is a child of the first level-1 node and that the final level-1 node is top-level. AliDocMind includes a duplicated native title in the input to exercise the merge path.

Important limitation: this table is a controlled same-document adapter-shape comparison, not a new cloud extraction benchmark. This environment has no AliDocMind or MinerU credentials, so I could not honestly rerun the referenced 8-page paper against both remote services. The AliDocMind adapter itself has a separate end-to-end unit test from its raw `layouts` response through the outline, and the regression mechanism that produced `110 / 56 / 40` is covered. A credentialed rerun of that paper would still be useful as an external quality check.

### Validation

- `pnpm vitest run tests/document`: 18 passed, 1 skipped; 86 tests passed, 2 skipped
- `pnpm exec tsc --noEmit`
- ESLint on all changed source/test files
- `pnpm vitest run --maxWorkers=1`: 322 passed, 3 skipped; 2924 tests passed, 4 skipped

yanpgwang:
First off — thanks @jackefn for the substantial work here, and for being so responsive across the earlier rounds. The framework side (transform contract, pipeline, registry, `normalize`, `remove-noise`) is genuinely solid, and the fixes from the last rounds all hold up under testing:

- Empty-block removal now cleans dangling citation/outline refs — verified 0 dangling refs across every document I tried.
- Transform history accumulates correctly (`priorTransforms`), verified end-to-end.
- AliDocMind's `level` now flows through to `metadata.headingLevel`.

I spent some time running the full pipeline against real extractions from all three providers (unpdf, self-hosted MinerU, a real AliDocMind call) across a few documents — an English arXiv paper, a Chinese rental contract, a `.docx`, a `.doc`. I ran into a handful of things that I *think* might be problems, but honestly I'm not sure what the right answer is, so I'd rather surface them and let the maintainers weigh in on direction than pretend I know.

## Things I noticed (roughly by how much they worried me)

**Possibly significant — structure detection can silently drop sections depending on the provider.** With the same Chinese rental contract (18 numbered articles, 第一条…第十八条):
- **MinerU → 19 outline nodes**, all articles present.
- **AliDocMind → 11 nodes**; 第八条 through 第十五条 (8 articles) don't appear in the outline at all, even though they're clearly in the body text (`第..条` shows up 24× in the raw text).

As far as I can tell this comes from `prefersProviderHeadings = providerHeadings.length > 0` — if the provider tags any heading, text-based detection is skipped entirely, so when AliDocMind only tags 11/19 blocks as `title`, the other 8 never get picked up. There's no error or fallback, which is what makes me a little nervous about it for the #335 "pick a chapter" flow — a chapter could just be missing from the picker. But I might be misreading the intent here.

**Smaller things I'm less sure about:**
- The `page:level:title` dedup key seems to behave oddly at the edges — it can drop distinct same-name sections when `pageNumber` is undefined (e.g. `.docx` blocks), and doesn't dedup a running title repeated across pages (the English paper still shows ~7 duplicate nodes).
- AliDocMind's heading levels seem to arrive as 0/1 and get clamped to 1, so the contract outline comes out flat (11 L1 nodes, no nesting).
- On the `.docx`, some plain body paragraphs ("Lorem ipsum…") ended up as outline nodes.
- Minor: the `PAGE_NUMBER` regex looks like it'd also match a footer "2024" or "512" and remove it as a page number.

## The thing I'm genuinely unsure about (would love a maintainer's read)

More than any single item above, I found myself wondering about the overall approach, and I really don't know the right call here. Right now `structure.ts` handles every provider through one shared set of heuristics (count `#`, numbered/chapter regexes, prefer-provider-if-any, the dedup key, level clamping, ~12k slicing, bbox-based footer detection). It's careful work, but in testing it seemed sensitive to which provider/format the document came from — swap the provider and sections can vanish, swap the format and body text can become headings — and I worry that fixing these one at a time is just more patching on the same base.

The alternative I kept coming back to (and I'm not at all sure it's better — this is where I'd defer to you all) is whether it might be cleaner to lean into the fact that providers genuinely differ in capability: let each provider's adapter map *its own* structure (AliDocMind already emits levels/title blocks; MinerU gives clean markdown; unpdf gives nothing structural and we'd just accept that / fall back), and then normalize those into a single outline for generation. That would move "how do we get structure out of this provider" to the layer that actually understands the format, rather than a shared layer guessing for all of them. But the hard part (merging heterogeneous results into one outline, and the no-structure case) is real, and I don't have a strong opinion on whether it's worth the churn versus continuing to refine what's here.

To be clear I think the feature itself is very much worth building — long-document chapterization is a real need (#52, #335). I just wasn't confident the current structure-detection model is the right foundation, and I'd rather ask than assume.

@wyuc would really value your steer on the direction here — per-provider structure mapping + normalization, or is there a reason the provider-neutral heuristic layer was the intended design? Happy to help prototype whichever way you'd prefer, and again, thanks @jackefn for pushing this forward.

jackefn:
# Follow-up: MAIC ETL Milestone 4 Structure Direction

First off — thanks @yanpgwang for testing this against real unpdf, MinerU, AliDocMind, DOCX, and DOC extractions. The concrete comparison was especially useful because it exposed a problem that the controlled fixtures did not: provider-native structure can be authoritative without being complete.

The earlier framework and data-integrity fixes remain unchanged:

- empty-block removal cleans dangling citation and outline references;
- transform history is owned by the pipeline and survives rebuilt artifacts;
- AliDocMind heading metadata flows through the compatibility layer;
- normalization and noise removal remain provider-neutral transforms.

This follow-up focuses on the structure boundary and the real-provider cases raised in the review.

## What changed after the review

### Partial provider structure no longer hides text sections

The review correctly identified the main failure mode:

```ts
const candidates = prefersProviderHeadings ? providerHeadings : textHeadings;
```

That choice treated the presence of any provider heading as evidence that the provider outline was complete. On the Chinese rental contract, AliDocMind supplied only 11 title blocks, so the remaining eight articles were silently excluded even though they were still present in the extracted text.

The transform now collects and reconciles both streams:

```text
provider heading candidates
              +
Markdown / numbered-text candidates
              |
              v
ordered occurrence reconciliation
              |
              v
canonical DocumentOutlineNode[]
```

Provider and text candidates with the same normalized title are matched in document order. A matched pair produces one node: the explicit provider level wins, while the text block and offset remain the content anchor. Candidates found by only one source are retained.

A regression reproduces the reported contract shape: one document title plus 18 numbered articles, with AliDocMind metadata for only 11 of the 19 headings. The resulting outline contains all 19 nodes; the eight missing provider headings are retained as `heuristic` nodes.

The transform also emits a warning when provider structure is partial, including provider, text, matched, and unmatched candidate counts. Missing provider coverage is therefore observable rather than silent.

### Markdown and numbered headings can coexist

Previously, finding any Markdown heading in a block stopped numbered-heading detection for the rest of that block. That could preserve a Markdown document title while still missing plain `第一条` through `第十八条` lines below it.

Both detectors now run over text/Markdown blocks, and `条` is included in the conservative Chinese numbered-heading pattern. Layout blocks are not reparsed as text, because they already carry provider metadata and would otherwise create a second copy of the same heading.

### Deduplication now uses occurrence evidence

The old `page:level:title` key had two opposite edge cases:

- when `pageNumber` was unavailable, distinct same-name DOCX sections could collapse into one;
- a running title repeated on several pages remained several outline nodes.

The global title set has been removed. Adjacent provider blocks are collapsed only when page, level, normalized title, and adjacent block order all agree. Same-name sections without page information remain distinct.

Running headings are handled separately from semantic title deduplication. A page-addressable provider title repeated on at least three pages and at least 60% of the document pages is excluded as a running heading. This safeguard does not apply to same-name sections whose pages are unknown.

### AliDocMind levels are normalized in its adapter

The real response contained levels `0` and `1`. Clamping both values in the shared transform flattened the hierarchy to L1.

The AliDocMind adapter now treats a response containing level `0` as zero-based and converts its title levels to positive, one-based levels. Responses that already use one-based levels remain unchanged. The shared transform validates canonical levels but does not decide what an AliDocMind level means.

This is consistent with the official Doc-JSON example, whose root hierarchy node uses level `0`: [Aliyun Document Mind Doc-JSON structure](https://help.aliyun.com/zh/document-mind/developer-reference/docstructure).

### Sentence-like DOCX paragraphs are filtered conservatively

Provider metadata is strong evidence, but real Office extraction showed that it is not infallible. A layout block that looks like a multi-sentence or sufficiently long sentence-like body paragraph is no longer accepted as a heading solely because it was labeled as a title.

This remains deliberately conservative. It is not intended to reconstruct Word typography or infer headings from font size. Explicit Office style mapping belongs in the adapter when the provider exposes that information.

### Bare numeric footers are preserved

Footer position alone no longer turns a bare number such as `2024` or `512` into a page number.

- a bare number is removable when the provider explicitly assigns `role: 'page-number'`;
- positional header/footer evidence can remove decorated forms such as `Page 12` or `— 12 —`;
- repeated explicit headers and footers continue to use the existing coverage threshold.

The default transform order is now:

```text
normalize -> remove-noise -> detect-structure
```

This prevents removable page noise from first becoming outline structure.

## Where I landed on the architectural question

I agree with the concern that one universal heuristic should not be responsible for interpreting every provider's raw response. At the same time, I do not think each provider should independently implement the complete final outline algorithm. That would give us several subtly different definitions of hierarchy, ranges, deduplication, and fallback.

The direction implemented here is a middle ground:

```text
Provider raw output
  -> provider / format adapter
  -> canonical blocks + provider-aware structure evidence
  -> provider-neutral transforms
       normalize
       remove conservative noise
       collect Markdown / numbered fallback evidence
       reconcile ordered candidates
       build canonical outline
       create logical sections only when no heading survives
```

In other words:

- the adapter answers “what does this provider field mean?”;
- the shared transform answers “how do multiple canonical pieces of evidence become one outline?”

The public `DocumentArtifact`, transform contract, and `DocumentOutlineNode` remain provider-neutral. Provider-specific field meanings stay in the adapter and, when retained, behind the existing `providerRaw` escape hatch.

## Responsibility split

### Provider and format adapters should own

- mapping provider title/layout types to canonical metadata;
- converting provider heading levels to positive, one-based levels;
- preserving page, position, confidence, subtype, and raw data when available;
- mapping explicit DOCX/DOC heading styles when the extraction result exposes them;
- declaring that a provider has no authoritative structure rather than inventing it.

Concretely:

- AliDocMind maps its title blocks and level semantics;
- MinerU supplies Markdown, whose heading markers become strong text evidence;
- unpdf supplies plain text and therefore uses only conservative numbered headings or logical fallback.

### The provider-neutral structure transform should own

- collecting canonical candidate streams;
- matching provider and text occurrences in document order;
- retaining unmatched candidates from both sources;
- selecting the stronger level/confidence for a matched occurrence;
- building node IDs, parent relationships, block references, page ranges, and offsets;
- reporting conflicts, partial coverage, deduplication, and fallback through diagnostics.

## Invariants this direction is intended to protect

1. A partial provider result cannot suppress a distinct heading still present in extracted text.
2. Matching provider and text evidence creates one outline occurrence, not two.
3. Two legitimate same-name sections are retained when occurrence evidence differs or is unavailable.
4. Provider levels are interpreted in the adapter, not guessed in the shared transform.
5. Text is never deleted merely because it was not recognized as a heading.
6. Logical fallback is used only when no provider, Markdown, or numbered candidate survives.
7. Partial coverage and reconciliation decisions are visible in diagnostics.

## What this still does not try to solve

This remains a deterministic foundation. It does not add:

- LLM chapter detection;
- chapter-selection UI;
- semantic long-document summarization;
- provider-specific indexes or retrieval;
- a guarantee that every visually styled heading can be reconstructed from plain text.

The controlled tests now cover the failure shapes from the review, but a sanitized fixture from each real extraction would still be valuable. In particular, the AliDocMind contract, English paper, DOCX, and DOC fixtures would let CI measure known-heading recall, duplicate occurrences, hierarchy accuracy, source coverage, and fallback ranges without requiring cloud credentials.

## Maintainer steer requested

@wyuc — does this responsibility split match the direction you want for MAIC ETL?

The specific proposal is **provider-aware evidence mapping plus provider-neutral reconciliation**, rather than either of these extremes:

1. one shared heuristic interpreting every provider's raw fields; or
2. every provider independently generating its own final outline semantics.

The main points where maintainer guidance would still be useful are:

- whether the ordered reconciliation policy belongs in this foundation PR or should become a separate structure package;
- whether the running-heading threshold should remain a deterministic default or be provider-configurable later;
- whether canonical structure evidence should eventually become an explicit public type instead of remaining internal candidates derived from blocks;
- whether low-confidence `heuristic` and `logical` nodes should be shown differently in the future chapter-selection UI.

@yanpgwang — the real-provider results were the basis for these changes. If you can share sanitized versions of those extraction shapes, I would be happy to turn the controlled regressions into direct fixture comparisons and report the post-fix node/duplicate/hierarchy numbers side by side.

Thanks again to both of you for helping clarify the boundary before chapter selection and long-document transforms build on it.

yanpgwang:
Tested v4 (`a6075b52`) against two publicly available documents (no privacy concerns, freely reproducible):

**1. Public Chinese arXiv paper (2401.00001)** — this looks great.

| Provider | outline nodes | strategy | matched / unmatched |
|---|---|---|---|
| AliDocMind | 28 | hybrid | 27 matched, 0 unmatched provider, 1 unmatched text |
| MinerU | 27 | heading | all clean markdown headings |
| unpdf | 3 | logical | honest fallback |

AliDocMind's 27/27 provider headings were all matched cleanly. No duplicates, no missing sections. The reconcile logic works well here.

**2. Public Wikipedia "中国" page** — all three providers struggle (the export PDF has complex layouts + incomplete CJK font embedding), so this is a document-quality issue, not a code issue. Can be ignored for our purposes here.

One small note: on documents with numbered body sub-clauses (e.g. legal texts, contracts), I noticed `NUMBERED_HEADING` is now scanning all lines, which can pull in body-item lines like `1、甲方同意…` / `2、乙方租用…` as headings alongside the real `第一条`–`第十八条`. You might want to add a length cap or require the matched line to end before a paragraph separator to avoid treating numbered *clauses* as numbered *headings*. Given the broader approach direction, I don't think this is a blocker — just a regex threshold to dial in. Totally your call to handle here or in a follow-up.

But overall the reconcile approach feels like the right direction, and the public-arxiv results above are solid. Thanks again for the thoughtful response and the work on this — really appreciate it.

wyuc (changes requested):
Thanks for the thoughtful iteration and the real-provider validation. I agree with the responsibility split you arrived at: provider adapters should interpret provider-specific fields, while shared code may reconcile canonical evidence. The transform contract, history, normalization, conservative noise removal, and cancellation/failure handling are useful foundations.

I do not think we should freeze the current outline implementation as OpenMAIC's canonical document structure yet. OpenMAIC should remain agent-first: the original document is the source of truth, while outlines, chunks, summaries, and indexes are optional, versioned, rebuildable projections. We should wait for a real consumer, such as chapter selection or repeated corpus retrieval, before standardizing the final structure contract.

There is also a correctness issue in the current range construction. `parsedPdfToDocumentArtifact()` stores both the aggregate `document-text` block and duplicated layout blocks. Reconciled text headings are anchored to block 0, while unmatched provider headings remain anchored to later layout blocks. `candidatesToOutline()` then assumes candidate `blockIndex` values are monotonic when slicing section blocks. In a partial-match document, this can produce overlapping `blockIds` or make a later section contain the entire document. The logical fallback can similarly create sections from the aggregate text and then create them again from duplicated layout blocks.

My requested scope change is:

1. Keep the generic transform pipeline, lineage, normalization, and conservative noise removal in this PR.
2. Make structure detection opt-in or move it to a follow-up PR instead of including it in `DEFAULT_DOCUMENT_TRANSFORMS`.
3. Treat any generated outline as an experimental structure proposal with source, confidence, and stable content anchors, not as the canonical document representation.
4. Keep the running-heading threshold deterministic for now. Do not add provider-specific configuration until real failure cases justify it.
5. Leave different UI treatment for `heuristic` and `logical` nodes to the chapter-selection work.
6. Tighten numbered-heading detection here or in the structure follow-up so body clauses such as `1、甲方同意...` are not promoted to chapters.

This preserves the valuable foundation in the PR while postponing the irreversible structure decision until we have a real consumer and evaluation criteria.

jackefn:
Thanks for the thoughtful review and the real-provider validation.

I agree with the responsibility split: provider adapters should interpret provider-specific fields, while shared code may reconcile canonical evidence where appropriate.

I have narrowed the PR scope accordingly:

- Kept the generic transform contract and ordered pipeline.
- Kept transform lineage/history, normalization, cancellation, and failure policies.
- Kept conservative repeated-header, footer, and page-number removal.
- Removed structure detection and logical fallback sections from this PR.
- Removed outline fields from `DocumentArtifact`.
- `DEFAULT_DOCUMENT_TRANSFORMS` now contains only normalization and conservative noise removal.
- No provider-specific structure configuration was added.
- No chapter-selection UI or heuristic/logical node distinction was introduced.

The document artifact remains the source of truth. Any future outline, chapter tree, chunk plan, summary, or index will be treated as an optional, versioned, rebuildable projection. A follow-up structure proposal will include explicit source, confidence, and stable content anchors before it is consumed by chapter selection or retrieval.

Running-heading behavior remains deterministic through fixed shared thresholds. Numbered-heading detection is intentionally deferred to the structure/chapter follow-up, where it will include regression coverage for body clauses such as:

```text
1、甲方同意……
1. 乙方负责……
1) 设备安装……
```

These clauses must not be promoted to chapters.

Validation completed for the revised scope:

- 4 Transform test files passed, 17 tests passed.
- TypeScript type checking passed.
- Prettier checks passed.

This keeps the reusable transform foundation in the PR while postponing the canonical structure decision until there is a concrete consumer and evaluation criteria.

wyuc:
Thanks for narrowing the implementation. I reviewed the revised diff and focused regression coverage, and the code now matches the requested boundary: the PR keeps the generic transform pipeline, lineage, normalization, and conservative noise removal, while structure detection and canonical outline fields have been removed. The current CI and 17 focused tests are sufficient for this foundation because it is not yet wired into an existing product flow.

There is one final housekeeping item before approval: the PR description still documents the previous scope. It currently mentions structure detection, logical fallback sections, outline fields and references, structure tests, and 23 tests across 7 files. Please update the Summary, Changes, Verification, Compatibility, and Known Limitations sections to reflect the current 13-file scope and 17 transform tests.

Once the PR description matches the revised implementation, this is ready for final approval.

jackefn:
Thanks! I've updated the original PR description in the initial comment to match the revised implementation and current scope. The outdated references to structure detection, canonical outlines, logical fallback sections, and structure-specific tests have been removed.

wyuc (approved):
The PR description now matches the revised implementation and scope. I also rechecked the current head after the latest merge from `main`: the PR diff remains limited to the intended 13 document-transform and test files, and both current CI workflows pass.

The requested scope changes have been fully addressed. Approving.

wyuc:
Merged, thank you. The source branch lives in your fork, so maintainers cannot remove it from the upstream repository. Please delete `feat/document-transform-pipeline` from `jackefn/openmaic-vocational-edu` when convenient.

PR #922, opened 2026-07-14
feat(pbl): cut learner state writes over to RuntimeStore (#869 Part C, step 4)

wyuc:
## Summary

- make RuntimeStore synchronization a durability gate before saving or exporting design-only PBL documents
- make the RuntimeStore fold authoritative after cutover, with a recoverable cutover marker and serialized per-learner runtime transactions
- strip learner state from both `projectV2` and the compatibility `projectConfig`
- restore the Planner-authored proficiency in persisted templates
- remove the old fire-and-forget post-save drain wiring

## Migration behavior

Legacy documents with embedded learner state remain document-authoritative for one hydration and backfill RuntimeStore.

Once the write cutover marker exists, RuntimeStore remains authoritative even if the first stripped document write was interrupted.

A RuntimeStore synchronization failure aborts the stripped document save or export instead of risking learner-state loss.

## Scope

This change is limited to the PBL write cutover. Incremental folding and the chat, quiz, and playback runtime migrations remain deferred.

Part of #869.

## Validation

- `pnpm exec tsc --noEmit`
- changed-file ESLint
- `pnpm check`
- full Vitest suite: 281 files passed, 2400 tests passed, 1 skipped
- fresh Codex-only review with no remaining actionable findings


wyuc:
## Codex-only review loop

This PR went through three Codex-only review passes. No Claude review was used.

1. Pass 1 found learner-state leakage through the compatibility `projectConfig`. The persisted compatibility payload is now rebuilt from the stripped design template.
2. Pass 2 found three data-integrity risks:
   - overlapping saves could let a stale snapshot overwrite newer learner progress
   - an interrupted first cutover could let the legacy document become authoritative again
   - a dynamically adapted proficiency tier could leak into the persisted design baseline

   These were fixed by serializing the full runtime authority transaction per learner and scene, using the first cutover snapshot as a durable authority marker, and restoring the Planner-authored proficiency when stripping the document.
3. Pass 3 found no remaining actionable issues.

Each actionable finding received a failing regression test before the implementation fix.

Validation after the final pass:

- `pnpm exec tsc --noEmit`
- changed-file ESLint
- `pnpm check`
- full Vitest suite: 281 files passed, 2400 tests passed, 1 skipped

Conclusion: the Codex-only review loop is complete with no remaining actionable findings. This is ready for maintainer review.

cosarah (approved):
LGTM

PR #923, opened 2026-07-14
fix(storage): store image files as array buffers

### LINKED ISSUE #910: [Bug]: UnknownError: Error preparing Blob/File data to be stored in object store — fails on all WebKit browsers
### Bug Description

Attaching a PDF and starting classroom generation fails with:
UnknownError: Error preparing Blob/File data to be stored in object store

Root cause: storePdfBlob() in lib/utils/image-storage.ts stores the
uploaded file as a raw Blob directly in the Dexie imageFiles table. This
is a known WebKit bug — storing Blob/File objects via IndexedDB's
structured clone algorithm is unreliable in Safari/WebKit and throws this
exact error. Dexie, PouchDB, and localForage users have all hit the
identical issue; the standard fix is to store the file as an ArrayBuffer
instead of a Blob, then reconstruct the Blob on read.

### Steps to Reproduce

1. Go to the OpenMAIC home page
2. Attach a PDF file
3. Enter a course generation prompt
4. Click "Enter Classroom" to start generation
5. See error: "Error preparing Blob/File data to be stored in object store"

### Expected Behavior

Generation should proceed normally after attaching a PDF, without an
IndexedDB storage error.

### Actual Behavior

Generation fails immediately with:
UnknownError: Error preparing Blob/File data to be stored in object store

### Deployment Method

### Browser

Safari 17 (macOS), Chrome (iOS — WebKit-based)

### Operating System

_No response_

```shell

```

### Additional Context

Confirmed this fails identically on both Safari macOS and Chrome iOS
(iOS Chrome uses WebKit under Apple's platform requirements, so this is
effectively all WebKit engines). Not yet confirmed on non-WebKit desktop
browsers (Chrome/Firefox on Windows).

Image

YizukiAme:
## Summary

- persist new `imageFiles` PDF and page-image payloads as `ArrayBuffer`
- reconstruct `Blob` values on read while remaining compatible with historical Blob records
- keep the existing Dexie schema/version and leave other Blob-backed tables unchanged

## Root cause

WebKit can fail while serializing Blob/File values for IndexedDB with `UnknownError: Error preparing Blob/File data to be stored in object store`. The document storage path still wrote Blob values into `imageFiles`, even after copying uploaded files into memory. Storing structured-clone-compatible `ArrayBuffer` values avoids that WebKit Blob serialization path.

## Implementation

- widen `ImageFileRecord.blob` to `Blob | ArrayBuffer`
- decode image data URLs once into bytes and MIME type
- store PDFs and parsed page images as `ArrayBuffer` with byte-accurate sizes
- normalize both new ArrayBuffer records and legacy Blob records back to Blob on reads
- retain document aliases, cleanup behavior, error propagation, and storage accounting

No database migration is required because `blob` is not an indexed field. Audio, generated-media, voice-profile, and voice-cache Blob fields are intentionally out of scope.

## Verification

- `pnpm exec vitest run tests/utils/image-storage.test.ts --reporter=dot` — 12 passed
- `pnpm exec eslint lib/utils/image-storage.ts lib/utils/database.ts tests/utils/image-storage.test.ts` — passed
- `pnpm test -- --reporter=dot` — 2,285 passed, 1 skipped
- Safari 26.5 normal window: uploaded a PDF containing an embedded page image, completed document extraction, stored page images, and entered the generated classroom without the IndexedDB Blob error

Safari 17 and iOS Chrome still need confirmation on the affected devices. If a later write to another Blob-backed table fails on those devices, that should be tracked as a separate follow-up rather than expanding this fix.

Fixes #910

### DIFF
--- a/lib/utils/database.ts
+++ b/lib/utils/database.ts
@@ -99,7 +99,7 @@ export interface AudioFileRecord {
  */
 export interface ImageFileRecord {
   id: string; // Primary key
-  blob: Blob; // Image binary data
+  blob: Blob | ArrayBuffer; // Image binary data
   filename: string; // Original filename
   mimeType: string; // image/png, image/jpeg, etc.
   size: number; // File size (bytes)
--- a/lib/utils/image-storage.ts
+++ b/lib/utils/image-storage.ts
@@ -2,7 +2,7 @@
  * Image Storage Utilities
  *
  * Store PDF images in IndexedDB to avoid sessionStorage 5MB limit.
- * Images are stored as Blobs for efficient storage.
+ * Images are stored as ArrayBuffers for cross-browser IndexedDB compatibility.
  */
 
 import { db, type ImageFileRecord } from './database';
@@ -13,9 +13,12 @@ const log = createLogger('ImageStorage');
 const SESSION_ID_LENGTH = 10;
 
 /**
- * Convert base64 data URL to Blob
+ * Decode a base64 data URL into its binary payload and MIME type.
  */
-function base64ToBlob(base64DataUrl: string): Blob {
+function decodeBase64DataUrl(base64DataUrl: string): {
+  buffer: ArrayBuffer;
+  mimeType: string;
+} {
   const parts = base64DataUrl.split(',');
   const mimeMatch = parts[0].match(/:(.*?);/);
   const mimeType = mimeMatch ? mimeMatch[1] : 'image/png';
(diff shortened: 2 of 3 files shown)


wyuc (approved):
Verified the focused storage change and its backward-compatibility path. Also tested the affected user flow locally in Safari 26.4: PDF upload, IndexedDB persistence/readback, document extraction (including one embedded image), generation preview, and classroom rendering all completed without the WebKit Blob/File IndexedDB error. Focused tests and CI are green.

PR #926, opened 2026-07-14
feat(runtime): cut chat sessions over to RuntimeStore

wyuc:
## Summary

- persist chat sessions as learner-scoped RuntimeSessions with append-only message and state records
- migrate legacy Dexie chat rows on first load and clear them only after a durable RuntimeStore sync
- preserve interrupted-session, pending-tool-call, and 200-message retention behavior
- keep chat replacement and deletion scoped to chat runtime sessions without affecting PBL data

## Safety

- ignore stale session writes using logical timestamps with sequence tie-breaking
- retain legacy rows when initial migration fails
- fail loudly when RuntimeStore becomes unavailable after cutover
- preserve the lecture `lastActionIndex: -1` sentinel in chat state without writing an invalid runtime action anchor

## Testing

- `pnpm exec tsc --noEmit`
- changed-file ESLint
- `pnpm check`
- full Vitest suite: 2564 passed, 3 skipped

Part of #869.


cosarah (commented):
I found two correctness issues that seem worth fixing before this chat RuntimeStore cutover lands.

1. `saveChatSessions` can delete a newer chat session from another tab. In `lib/utils/chat-storage.ts:361`, `syncSessions(..., deleteOmitted: true)` removes every existing chat session whose runtime id is not present in the caller's current snapshot. That would be fine if the caller always had an authoritative full session set, but stage saves are broader than chat writes: many ordinary stage mutations debounce into `saveToStorage()` and persist the current `chats` array as part of the stage snapshot (`lib/store/stage.ts:176`, `lib/store/stage.ts:295`). So if tab A is still holding an older `chats` snapshot while tab B creates a new chat session and saves it, tab A's next unrelated autosave can erase B's new runtime chat session just because it was omitted locally. The stale-write guard in `syncOne` only protects per-session content/state updates; it does not protect this omitted-session deletion path. I think deletion here needs either an explicit tombstone / user-delete signal, or at least a recency check so an older snapshot cannot remove a newer session it never observed.

2. The first write / migration path still has an unhandled `createSession` race. In `lib/utils/chat-storage.ts:305`, `syncOne` lists sessions and then calls `store.createSession(...)` when the runtime chat session does not exist. `BrowserRuntimeStore.createSession` is fail-loud on duplicate ids (`packages/@openmaic/storage/src/runtime/browser.ts:220`), so two tabs performing the first migration or first save for the same chat session can still lose the race with an `already exists` error. The per-store `enqueue` queue only serializes calls inside one JS context; it does not protect cross-tab writers. The PBL runtime path already treats this as a real race and relists on `already exists` in `lib/pbl/v2/runtime/drain.ts:154`; I think chat needs the same retry/relist pattern here, plus a regression test for the cross-tab first-write case.

I could not run the targeted Vitest file locally in this worktree because `vitest` is not installed here, so I am basing this review on the code paths and the existing tests in the PR.

wyuc:
Thanks for the careful review. Both issues are fixed in the latest push.

1. Omitted-session deletion now only retires chat IDs that the current RuntimeStore instance has previously observed in that learner/stage partition. A stale tab can no longer delete a session created by another tab merely because it is absent from the stale snapshot. Cross-tab regression coverage exercises the unseen-session case.

2. Production chat writes now serialize through Web Locks across tabs, and session creation uses a create-or-get retry path that relists after a duplicate-id race. Regression coverage exercises concurrent first writes from separate RuntimeStore instances.

The follow-up cross-review loop also found and fixed:

- queued writes being overtaken by database clear
- a partition/global lock-name collision while preserving compatibility with already-open tabs
- completed streamed text losing to a stale partial snapshot
- backup export succeeding after a runtime read fallback and omitting runtime-only chats

Verification on the final branch:

- TypeScript, ESLint, Prettier, and diff checks pass
- full Vitest suite: 2,651 passed, 3 skipped
- final independent Claude and Codex full-diff reviews: clean

CI is running on the latest head. Once it is green, this is ready for another review.

cosarah (commented):
I found one remaining compatibility concern in this PR.

`saveStageData` still routes every persisted `chats` array, including `[]`, through `saveChatSessions` (`lib/utils/stage-storage.ts:78`). The default chat path marks the Dexie-backed legacy store as requiring Web Locks, and `enqueue` now throws when `navigator.locks` is unavailable (`lib/utils/chat-storage.ts:159`, `lib/utils/chat-storage.ts:139`). Because stage autosave is triggered from broad non-chat mutations like `setStage`, `setScenes`, `setCurrentSceneId`, and `setChats` (`lib/store/stage.ts:176`, `lib/store/stage.ts:190`, `lib/store/stage.ts:290`, `lib/store/stage.ts:295`), a browser without Web Locks will start failing ordinary stage saves even when the stage has no chat sessions at all. Elsewhere the runtime layer still degrades gracefully on older browsers (for example `lib/runtime/learner-key.ts` falls back when Web Locks are missing), so this looks like a real compatibility regression unless the supported-browser policy has explicitly changed. If dropping no-Web-Locks browsers is intentional, I think that needs to be documented explicitly; otherwise this path probably still needs a non-lock fallback for empty or otherwise non-migrating chat partitions.

I could not rerun the targeted Vitest files locally in this worktree because `vitest` is not installed here, so I am basing this comment on the current code and the added tests in the PR.

wyuc:
Thanks, this exposed a real compatibility gap. I’ve kept document-only stage saves and read-only legacy chat recovery working when Web Locks are unavailable, while retaining fail-loud behavior for non-empty chat writes and restore. I also hardened partial-write atomicity, conflict-clock reconciliation across tabs and clock rollback, stream reveal timestamps, and runtime-wide clear/delete coordination for chat and PBL writers. The latest full suite passes, and a final independent Codex and Claude closure review found no remaining actionable issues.

cosarah (commented):
I found one remaining behavioral gap worth fixing before this lands.

`loadChatSessions` now intentionally falls back to a read-only legacy snapshot when Web Locks are unavailable (`lib/utils/chat-storage.ts:1061`), and that snapshot is loaded straight into stage state (`lib/store/stage.ts:508`). But ordinary autosaves still persist the full `chats` array from stage state (`lib/store/stage.ts:413`), while `saveStageData` only suppresses `ChatStorageLockUnavailableError` when `data.chats.length === 0` (`lib/utils/stage-storage.ts:84`). In practice that means a browser without `navigator.locks` can still open a stage with pre-cutover legacy chats, but the next unrelated stage edit will start failing persistence just because those fallback chats are present. The new tests appear to encode that behavior too (`tests/runtime/database-chat-cutover.test.ts:477`, `tests/runtime/database-chat-cutover.test.ts:495`).

Unless the intended product behavior is that stages with chats become effectively read-only on no-Web-Locks browsers, I think the save path still needs one more compatibility branch here, or the UI needs an explicit unsupported-state guard instead of silently letting autosaves fail.

wyuc:
Thanks, you were right. The no-Web-Locks fallback snapshot now remains usable for ordinary document-only stage autosaves. An unchanged chat snapshot is treated as a safe no-op, while actual chat creation, modification, deletion, and backup restore still fail loud when cross-tab serialization is unavailable.

The follow-up review loop also fixed transient read recovery, durable deletion tombstones, interrupted backup restore recovery, stage-scoped restore staging, same-ID chats across stage partitions, and staged-chat cleanup during stage deletion.

The branch is now updated with the latest main. TypeScript, ESLint, Prettier, targeted runtime tests, GitHub CI, E2E, and CodeQL all pass. The final Codex review found no remaining actionable issues.

Ready for another review when you have time.

cosarah (approved):
## Approving ✅

Ran a two-axis review (standards + spec) against `origin/main...HEAD`. No blocking issues: no implementation errors, no documented-standard violations, and the spec is implemented faithfully — stale-write ordering, migrate-then-clear-after-durable-sync, legacy retention on failure, the `lastActionIndex: -1` sentinel handling, and 200-message/interrupted/pending-tool-call preservation all check out.

A few non-blocking notes for follow-up (all judgement calls, none need to hold the merge):

### Code quality
- **Data Clumps** — `(store, stageId, learnerKey)` travel together through nearly every function in `chat-storage.ts` (`runtimeViews`, `syncOne`, `syncSessions`, `loadRuntimeSessions`, `clearRuntimeChatSessionsUnlocked`, …). Worth bundling into a single `ChatPartition` context type.
- **Duplicated tie-break comparator** — the `updatedAt → generation → id.localeCompare` comparison is re-implemented in ~5 places (`newestRuntimeCandidate`, `highestGeneration`, `retireRuntimeCandidates`, `loadRuntimeSessions`, the `syncSessions` deletion guard). Extract one shared comparator. This is the highest-value cleanup.
- **Primitive Obsession on runtime IDs** — IDs are encoded/parsed via ad-hoc string concat + `split`/`slice` (`generationRuntimeSessionId`, `chatRuntimeIdentity`, `deletionMarkerId`, `restoreMarkerPrefix`). A parsed ID type would remove the scattered decode-guard logic.
- **Middle Man** — `chat-storage-lock.ts` exports pure "compatibility aliases" (`withChatStorageSharedLock = withRuntimeStorageSharedLock`, exclusive variant) that add indirection for a brand-new module; consider dropping them.
- **Naming** — `async function context(options)` is too generic for what it does (resolve store/learnerKey/legacyStore); e.g. `resolveChatContext`.
- **Type casts** — `withPartitionLocks` uses stacked `as unknown as Promise<T>` around `navigator.locks.request`; a typed wrapper would be cleaner.

### Scope creep (acceptable, just flagging)
- **PBL drain/hydration rewrite** — `drainProjectRuntimeFully` → `withDrainedProjectRuntime` and hydration wrapped in `withRuntimeStorageSharedLockUntilSettled` touch PBL runtime paths despite the "without affecting PBL data" goal. It only changes serialization (so the deletion cascade can quiesce all kinds), not PBL data content — justified supporting infra, but the largest out-of-chat change.
- **`clearDatabase` now clears all runtime kinds** via the new `deleteAllRuntime` — reasonable for a whole-DB reset now that chat lives in runtime, but note it also clears PBL runtime.

### One thing worth confirming
The spec says "fail loudly when RuntimeStore becomes unavailable after cutover." The chat-storage layer does throw on read/write, but the stage read path (`stage-storage.ts:131-134`) deliberately swallows the error and returns empty chats to keep the document loadable. It's data-safe (`restoreMarker: undefined` makes a later empty autosave a no-op, and a real chat write still throws), so it reads as intentional — just confirming it is.

Issue #895, opened 2026-07-09
[Task] AI element edits — natural-language → EditIntent vertical, the editing surface's first real consumer (#646 / #851)


wyuc:
Tracking: #646 (AI editing direction), #851 (editing surface). Related: #720 Phase 3.

## Motivation

The `@openmaic/renderer` editing surface now covers selection, drag, resize, rotate, line reshape, and marquee multi-select with a controlled `EditIntent` API (#859 / #881 / #888) — but its only consumers are tests. #851's own gate says the API should be exercised by a second, real consumer before it freezes. Meanwhile the app's editor agent (`lib/agent/`) edits at exactly one granularity: wholesale — `regenerate-scene-actions` rewrites a scene's actions, `edit-interactive-html` rewrites an interactive document. The per-element direction ("make this title blue", "move that figure down") called out in #851's layering note has no implementation.

This task is the smallest vertical that serves both: a natural-language **element edit** flows through the same intent vocabulary as a pointer gesture.

## Scope

1. **An `edit_elements` agent tool** (`lib/agent/tools/`, app-side like its siblings): takes the instruction plus the current slide's element inventory (ids, kinds, box props, and the style props the DSL owns) and the active selection as context; the model returns proposed element updates.
2. **A pure intent-mapping gate** between model output and the canvas: validate and coerce into `element.update` / `element.updateMany` — reject unknown element ids, locked elements, out-of-contract props, and malformed values; clamp geometry the same way the gesture cores do. Malformed model output degrades to a refusal, never a partial apply.
3. **Apply through the host's existing `onElementsChange` path** — one instruction = one intent batch = one undo entry, exactly like a completed pointer gesture. No new mutation channel.
4. **Chrome**: a minimal entry in the existing AgentBar flow (the editor chrome stays app-side per #851); selection-aware ("this title" resolves against the current selection first).
5. **Evaluation cases**: a small fixture suite of instruction → expected-intent pairs (color/position/size/rotation/mixed-target cases, plus refusal cases for locked/unknown targets).

## Explicitly out of scope

- Wholesale regeneration (exists), text content rewriting, and any new `EditIntent` kind — if the vertical genuinely needs a new intent, that is a finding for #851, not something to add here ad hoc.
- ProseMirror / inline text editing (Part B), group scaling and the remaining #851 Part A slices.
- Server-side agent changes beyond registering the tool.

## Why this ordering

[...]

PR #927, opened 2026-07-14
feat(agent): add validated JSON Patch element edits

### LINKED ISSUE #925: [Task] AI element edits v2: validated JSON Patch backend (#646 / #851)
Tracking: #646. Builds on #895 / #896. Related: #851.

## Motivation

#895 established the first user-visible element-editing vertical, but its model boundary is a bespoke `{ id, props }` proposal format backed by a large property-specific mapping gate. Extending that format to text content, structured data, media, and element lifecycle operations would require more parallel tool contracts and validators.

This task replaces that proposal boundary with one validated JSON Patch path. The agent gets one element-editing tool; the host keeps authority over validation, atomic application, undo, persistence, and stale-state protection.

## Initial scope

1. Keep `edit_elements` as the single public agent tool.
2. Change its model-facing input to direct JSON Patch operations against the trusted slide canvas document.
3. Require the agent to inspect the scene before patching. Indexed element paths must be guarded by an element-id `test` operation.
4. Apply the complete patch batch to an isolated copy before committing anything.
5. Validate the patched document against the generated `@openmaic/dsl` schema.
6. Retain policy checks that schema validation cannot express:
   - locked elements cannot be changed;
   - element identity and type cannot be replaced directly;
   - stale target fingerprints refuse the whole batch;
   - accepted edits must map to renderer-visible behavior.
7. Commit the whole patch batch through the existing host edit session as one undo entry.
8. Feed structured apply success or refusal back into the tool card and subsequent agent history.
9. Preserve the existing #895 geometry and style fixture suite while adding text and shape-label content edits.

## Initial operation policy

Supported:

- `test`
- `add`, only for an absent optional property on an existing element; it cannot create or insert an element
- `replace`
- Existing-element geometry and renderer-visible style paths already supported by #895
- Text-element content
- Shape-label content

Deferred:

- Element add, delete, duplicate, move, and reorder
- Chart and table data editing
- Media source replacement
- Element type conversion
- Interactive-scene HTML editing
- Renderer v2 extraction work tracked by #851

These capabilities should later extend the same patch tool rather than introduce element-type-specific public tools.

## Acceptance

- [ ] A single `edit_elements` call can change an existing title's wording and color in one atomic patch batch.
- [ ] Indexed paths guarded by a mismatched element-id `test` are refused without changing the slide.
- [ ] Invalid paths, invalid DSL values, locked targets, stale targets, and renderer-shadowed edits refuse the entire batch.
- [ ] One successful patch batch produces exactly one undo entry.
- [ ] Successful edits survive save and reload.
- [ ] Apply-time refusal is visible to the user and available to the agent on its next turn.
- [ ] Existing #895 geometry, style, group, stale-state, and renderer-contract tests remain green.
- [ ] No new element-type-specific public agent tool is introduced.

wyuc:
## Summary

- Replace the nested natural-language proposal boundary with direct guarded JSON Patch for `edit_elements`.
- Support atomic geometry, renderer-visible style, text-content, and shape-label edits.
- Add paged element manifests and exact target-record reads for large slides.
- Preserve the existing host edit session, stale-state checks, persistence, and one-step undo behavior.

## Contract and safety

- Every indexed mutation requires a preceding element-id `test`.
- `add` is limited to absent optional properties on existing elements; it cannot insert elements.
- Patch batches are cloned, validated against the generated DSL schema, and refused atomically.
- Values that would be clamped or normalized are refused to preserve exact patch semantics.
- Rich-text HTML uses a strict ProseMirror/importer-compatible allowlist with CSS injection protection.
- Duplicate ids, locked elements, oversized content, stale targets, unsupported paths, and renderer-invisible edits are refused.

## Verification

- `pnpm exec tsc --noEmit`
- Changed-file ESLint
- `pnpm check`
- Full Vitest: 294 files passed, 2 skipped; 2588 tests passed, 3 skipped
- Five-round Codex-only independent review loop; all code findings closed

Closes #925
Tracking #646
Related #851
Builds on #895 and #896


wyuc:
Codex-only review loop completed over the full branch diff.

Five independent review rounds covered the JSON Patch contract, stored-content security, client atomicity/staleness/undo, large-slide projection, importer/ProseMirror compatibility, and model-facing usability.

Findings fixed during the loop included stored HTML/CSS injection, optional-property semantics, lossy long-content projection, legacy value normalization, duplicate ids, manifest paging, exact-record budgets, oversized replacement values, and apply-state contract gaps.

Final result: zero remaining code P0-P2 findings. Full Vitest, TypeScript, ESLint, Prettier, and diff checks pass.

cosarah (commented):
I found two substantive issues in this PR.

1. `filters` patches will be wrongly refused on existing slides that still store filter units. `read_scene_content` exposes raw filter values from the slide (`lib/agent/tools/read-scene-content.ts:159`), so the model will naturally patch against values like `blur: "2px"` or `contrast: "90%"`. But `mapElementJsonPatchToEditIntents` builds the proposal from the fully patched element (`lib/agent/tools/edit-elements-patch.ts:255`) and then exactness-checks it against the normalized gate output (`lib/agent/tools/edit-elements-patch.ts:412`). Since `clampUpdateProps` canonicalizes the whole filter map to unitless strings (`lib/agent/tools/edit-elements-gate.ts:759`), touching one leaf such as `/elements/0/filters/brightness` also normalizes untouched siblings and trips the `would be normalized` refusal. The host path explicitly treats unit-suffixed filter values as valid existing data (`tests/lib/agent/client/apply-edit-elements.test.ts:342`). That means a class of legitimate image-filter edits will fail on older slides. I think the patch layer needs to compare only the requested leaf changes, or canonicalize existing sibling filter values before the exactness check.

2. The implementation advertises guarded JSON Patch with `replace`, but whole-object replaces on `outline`, `shadow`, and `filters` do not actually behave like replace. The tool contract says to change only the requested paths (`lib/agent/tools/edit-elements.ts:93`), but once the patch mapper flattens the modified element back to top-level props (`lib/agent/tools/edit-elements-patch.ts:255`), the client apply path merges those structured props instead of replacing them (`lib/agent/client/apply-edit-elements.ts:45`). So `replace /elements/N/filters` or `replace /elements/N/outline` silently preserves omitted sibling keys, and because `remove` is unsupported there is no way for the model to actually clear them through this contract. If the intent is a renderer-visible patch DSL rather than real RFC 6902 replace, the contract and prompt should be narrowed. Otherwise the host apply path needs enough information to distinguish whole-object replace from leaf merge.

I could not run the targeted tests locally in this worktree because `vitest` is unavailable here, so I am basing this review on the code paths and the existing test coverage in the PR.

wyuc:
Addressed both review findings:

- Nested structured-style patches now validate and emit only the requested leaves, so legacy unit-bearing filter siblings remain untouched.
- Whole-object replacements for `outline`, `shadow`, and `filters` now remove omitted siblings before applying the replacement.

I also covered replacement identity values, marker ordering, grouped content edits, and sequential root-add-then-nested-patch semantics. The full local suite and all GitHub checks pass.

cosarah (commented):
I found one remaining compatibility issue in the latest revision.

 still appends CSS units unconditionally. However, this PR now intentionally preserves legacy unit-suffixed sibling filter values when applying partial image-filter edits, as covered in . For example, after a partial update we can now persist .

The app-side renderer already handles that legacy shape safely in  by checking  before appending a suffix. The package renderer does not, so any read-only surface that still goes through  can produce invalid CSS such as  or .  is one concrete example that still renders through the package surface.

I think  should use the same suffix guard as the app renderer, and this should probably get a package-side render/filter test so the two surfaces stay aligned.

cosarah (commented):
I found one remaining compatibility issue in the latest revision.

packages/@openmaic/renderer/src/elements/image/useFilter.ts:16 still appends CSS units unconditionally. However, this PR now intentionally preserves legacy unit-suffixed sibling filter values when applying partial image-filter edits, as covered in tests/lib/agent/client/apply-edit-elements.test.ts:403. For example, after a partial update we can now persist filters like blur=2px, contrast=90%, and brightness=120%.

The app-side renderer already handles that legacy shape safely in components/slide-renderer/components/element/ImageElement/useFilter.ts:16 by checking whether the value already ends with the expected unit before appending a suffix. The package renderer does not, so any read-only surface that still goes through @openmaic/renderer can produce invalid CSS such as blur(2pxpx) or contrast(90%%). components/slide-renderer/SlideThumbnail.tsx:147 is one concrete example that still renders through the package surface.

I think @openmaic/renderer should use the same suffix guard as the app renderer, and this should probably get a package-side render/filter test so the two surfaces stay aligned.

wyuc:
Thanks, fixed in `dbd1894f`.

The package renderer now uses the same suffix guard as the app renderer, so legacy values such as `blur: "2px"` and `contrast: "90%"` no longer become `blur(2pxpx)` or `contrast(90%%)`.

I added:
- package-side canonical and legacy filter tests;
- a `BaseImageElement` render test covering the actual read-only path;
- a cross-renderer contract test that feeds both renderers the same canonical and legacy filter maps and requires identical CSS.

Self-checks passed: package tests (223), package typecheck/build, root tests (2604), root TypeScript, ESLint, Prettier, and diff check.

cosarah (approved):
LGTM

Issue #866, opened 2026-07-07
[Task] Service-backed MP4 render (isolated producer container + in-app one-click export)


cosarah:
## Context

Sub-issue of #854 (Hyperframes-based classroom video export architecture). Depends on the export-ZIP phase #865, which already makes local CLI rendering possible. This phase adds in-app one-click MP4 export by moving render execution to a controlled service.

Constraint: the Hyperframes producer requires **Node >= 22** (repo engines is >= 20.9) and needs Chrome + FFmpeg. Rendering must **not** run in the Next.js route (heavy runtime, memory pressure, maxDuration limits). It must be isolated.

## Goal

Run the Hyperframes producer's HTTP server (POST /render + SSE progress) in an isolated Node 22 container, and wire the app to upload the export ZIP, stream progress, and receive the MP4.

## Scope

1. **Render service**: `@hyperframes/producer` HTTP server (POST /render + SSE) in a dedicated container (Node 22 + chromium + ffmpeg), isolated from the main runtime.
2. **App integration**: upload ZIP → SSE progress → download MP4; graceful failure messaging.
3. **Deployment shape**: for self-hosted users this is an **opt-in capability** — when not configured, degrade to the local-CLI path (P1d) with a clear prompt.
4. Resource/concurrency sizing informed by the memory/compute measurements (see the sizing task).
5. A small golden/smoke render fixture before broad UI exposure (sample key timestamps, tolerance-compare against baseline frames) — as an isolated CI job (Chrome+FFmpeg+Node22), not part of the default lint chain.

## Non-goals

- Any hosted-deployment-specific billing/quota concerns (out of scope for the upstream repo).
- Richer fidelity / caption polish (P3).

## Acceptance criteria

- In-app one-click export produces an MP4 via the containerized producer service.
- When the render service is unconfigured, export degrades cleanly to the local-CLI ZIP path with clear messaging.
- A smoke-render CI fixture guards regressions in an isolated job.

Depends on #865.

PR #931, opened 2026-07-15
feat(video-export): L1 Hyperframes emitter + browser collection + export ZIP (#865)

### LINKED ISSUE #865: [Task] L1 composition emitter + browser collection layer + self-contained export ZIP
## Context

Sub-issue of #854 (Hyperframes-based classroom video export architecture). Depends on the IR/compiler #864. This phase makes #290 actually usable: a user can export and render a video.

The IR from #864 is emitted into a **self-contained Hyperframes composition project (ZIP)**, and a browser-side collection layer gathers the real assets. Rendering happens out-of-app via the Hyperframes CLI (`npx hyperframes render`).

Target fidelity is **L1**: static scene base + spotlight/laser effect layers + real video clips + narration + SRT/VTT subtitles — this already covers all deterministic authored content of a slide classroom.

## Goal

Ship the emitter + browser collection layer + packaging so users can `npx hyperframes preview` / `render` a self-contained export ZIP to MP4.

## Scope

1. **Emitter (IR → Hyperframes project)**:
   - Root composition with scenes laid out sequentially; per-scene nested composition with tracks mapped to app z-order (base / video / spotlight / laser / subtitle / narration).
   - Effect descriptors (`spotlight.v1` etc. from #863) → generated **paused GSAP timelines** registered on `window.__timelines`.
   - `play_video` → a real `<video>` clip **at the element's own position** (mirrors playback, which plays the video in place at the video element's slot); original audio mixed at encode time.
   - Determinism red-line: **vendor GSAP locally (no CDN)**, no `Date.now`/`Math.random`/network at render time, explicit root `data-duration`. Enforce via `hyperframes lint` in CI.
2. **Browser collection layer** (thin app-side adapter, feeds the DI interfaces from #864):
   - Slide snapshots via `slideToPng`; audio/video duration probing; asset collection (reuse #849's collectors); live-mode `ossKey` audio fetched at **compile time** (never at render time) so the ZIP is self-contained.
   - Streaming ZIP write to control memory.
3. **Subtitles**: derive SRT/VTT from the IR's narration segments (single source).
4. **Export UX**: reuse the existing export menu; resolution selection + progress/error UX (recover from #746).
5. **Artifact layout**: `index.html`, `openmaic-video-manifest.json`, `subtitles.srt/.vtt`, `assets/{frames,audio,media,vendor}`, `README.md` with render instructions.

## Non-goals

- Service-backed rendering (next phase, P2).
- Placeholder-card visual polish / burn-in subtitle styling (P3).
- Interactive/quiz/PBL/widget real rendering (research, later).

## Acceptance criteria

- From a generated classroom, the app produces a self-contained ZIP that `npx hyperframes render` turns into an MP4 with correct narration audio, spotlight/laser effects, embedded video playback, and SRT/VTT.
- `hyperframes lint` passes on the emitted project (no CDN, no non-deterministic APIs, explicit durations).
- Emitter output has HTML string-snapshot tests; subtitle derivation has unit tests.
- Memory stays bounded on a large classroom (streaming ZIP; objectURL released after each snapshot).

Depends on #864 (and reuses #849). Precedes the service-backed render phase.

cosarah:
## Summary

Second slice of the Hyperframes classroom-video architecture (#854). Consumes the `VideoTimeline` **IR** landed in #913 (closing #864) and produces a **self-contained Hyperframes composition project (ZIP)** that `npx hyperframes render` turns into an MP4.

This PR stops at producing a valid Hyperframes input — the actual MP4 render/service execution is **#866**.

The emitter is a pure downstream consumer of the IR (HTML/GSAP as strings, under the `lib/video-export/**` purity boundary); everything that touches Dexie/DOM lives in a new `lib/video-export-app/`. Because every IR time is already absolute on the global playback clock, the whole classroom emits as **one flat composition driven by one paused GSAP timeline** on `window.__timelines`.

## Related Issues

Closes #865. Depends on #913 (#864). Precedes #866 (render execution). Reuses slide-snapshot logic from #849.

## Changes

- **`lib/video-export/emit-hyperframes/`** — pure IR → Hyperframes project emitter: base-frame / video / narration clips laid out with `data-start`/`data-duration`; spotlight & laser overlays expanded from `lib/choreography` descriptors into paused-timeline GSAP tweens.
- **`lib/video-export/subtitles.ts`** — SRT/VTT serialization of the IR's first-class subtitle track.
- **`lib/video-export-app/`** — impure app glue: `timeline-deps.ts` (sync `TimingProbe`/`AssetSource` over Dexie, video-duration probing), `collect.ts` (slide snapshots via `slideToPng` + object-URL lifecycle lifted from #849; byte-fetch over the asset plan), `package-zip.ts` (streaming ZIP), `use-export-video.ts` (orchestration hook).
- **Vendored GSAP** at `public/vendor/gsap.min.js` — determinism red-line (no CDN at render time).
- **Export UX** — "Export Video" menu entry with a resolution picker (720p/1080p/4K) in `header-controls.tsx`, plus i18n keys across all 8 locales.
- Artifact layout: `index.html`, `openmaic-video-manifest.json`, `subtitles.srt/.vtt`, `assets/{frames,audio,media,vendor}`, `README.md`.

## Type of Change

### Steps to reproduce / test

1. Open a generated slide classroom.
2. Use the export menu → **Export Video** → pick a resolution.
3. Confirm a `<name>-video.zip` downloads.
4. Unzip and verify `index.html`, `openmaic-video-manifest.json`, `subtitles.srt/.vtt`, `assets/{frames,audio,vendor}`.
5. `npx hyperframes lint` on the unzipped project passes with no errors.

- **End-to-end against the real `hyperframes` CLI (v0.7.58)** on an actual classroom export (`给我讲讲讲可计算性`, 5 slides / 44 narration clips): `hyperframes lint` → **0 errors, 4 informational warnings**.
- The first-pass lint surfaced real correctness gaps that unit tests could not (media elements need `id`, `<video>` needs `data-has-audio`, GSAP exit fades need `tl.set` hard-kills, laser must animate `x/y` transforms not `left/top`) — all fixed; re-lint clean.
- **Rendered a slice to a valid MP4** (H.264 1280×720 + AAC stereo) to confirm the emitted project drives Chrome capture + audio mux correctly. Full-length render is out of scope (that's #866).
- Full test suite: **2577 passed / 4 skipped**, incl. 24 new emitter/subtitle specs + a gated `hyperframes lint` determinism proxy. `tsc`, `eslint` (both purity boundaries), and `prettier` clean.
- Not verified in this PR: the app-side hook running in a live browser (typecheck only), and the MP4 render execution path (#866).

### Evidence

- [x] Manually tested locally

Rendered frame at t=3s (real classroom export → MP4, proving visuals + timing):
first slide "什么是可计算性? (Computability)" renders correctly with narration audio embedded.

## Checklist


cosarah:
## Self-review fixes (AI-assisted) — pushed 9cbd7c1

Ran a two-axis code review (standards + spec) and addressed the findings.

### 🔴 Fixed — live-mode `ossKey` audio not fetched at compile time
The spec requires a **self-contained** ZIP ("live-mode ossKey audio fetched at compile time"), but collection read only the local Dexie `blob`. Records can be LRU-evicted under storage pressure, so a live classroom with dropped blobs would export with silent gaps.

- `collect.ts`: new `resolveBytes()` prefers the local blob and falls back to the record's CDN URL — `ossKey` for audio/media, `posterOssKey` for posters.
- `timeline-deps.ts`: presence checks widened to `blob.size > 0 || !!ossKey` so the compiler keeps these entries instead of marking them absent (otherwise the fallback would never run). Video-duration probing stays local-only — an evicted video just gets its dwell capped by the compiler, and the bytes are still fetched at collect time for the render.
- Added `tests/video-export/collect.test.ts` (local-first, ossKey fallback, fetch 404/throw, video+poster).

### 🧹 Fixed — emitter duplication
Extracted one shared `escapeHtml` + `sec` into `emit-hyperframes/format.ts`, removing two divergent HTML escapers (one omitted `>`), two `sec()` definitions with different rounding, and a middle-man `escapeAttr`. Output is byte-identical — the HTML snapshot is unchanged.

### ⏸️ Deferred — "streaming ZIP write to control memory"
Collection still buffers asset blobs in memory and `JSZip.generateAsync` materializes the archive, so this is **not** truly streaming. Deliberately leaving it as-is for now:

- This is **browser-tab memory**, not server memory — the whole export pipeline is client-side (Dexie → JSZip → `saveAs`); the server isn't involved.
- Measured peak is small relative to the tab itself: a 10-slide classroom with lots of image assets produces a ~70 MB ZIP, while the classroom page already sits at ~200–300 MB. Streaming would add a File System Access API / service-worker (StreamSaver-style) dependency for a marginal saving.
- The higher-value half of that requirement — releasing each snapshot's objectURLs immediately so frames don't accumulate — is already implemented.

Not opening a follow-up issue; we can revisit if a real out-of-memory case shows up on a very large classroom.

### Verification
- `tests/video-export/`: 77 passed / 1 skipped (the skipped one needs `HF_E2E_DIR` for the real-CLI E2E).
- `eslint` (incl. the `lib/video-export/**` purity boundary) + `prettier`: clean.
- `tsc --noEmit`: no new errors (the pre-existing `use-import-pptx.ts` `@openmaic/importer` error is unrelated to this branch).

wyuc (changes requested):
There is one remaining blocker in the new OSS fallback path.

**[P1] Restore evicted generated media before rendering slide base frames.** `resolveBytes()` is now used when collecting standalone audio/video/image/poster entries, but the frame path still calls `resolveGeneratedMedia()`, which only accepts a non-empty local `record.blob`. When a live-mode record has been LRU-evicted (`blob.size === 0`, valid `ossKey`), lines 105-108 clear an image's `src`; video and poster URLs are likewise not restored. `renderFrame()` then snapshots that incomplete slide. The ZIP may contain the separately fetched media asset while its base-frame PNG is permanently missing the generated image/video.

Please resolve the media bytes (local blob first, then `ossKey` / `posterOssKey`) before creating the object URLs used by `slideToPng`, and add a frame-collection test covering an evicted generated image or video. The newly added tests currently exercise only standalone asset entries, so they do not catch this path.

The feature-flag and emitter-dedup updates otherwise look sound. I could run the feature-flag tests locally; the collect suite could not start in this review worktree because the shared dependency checkout did not contain a built `@openmaic/renderer/snapshot`. CI should remain the authoritative run for that suite.

cosarah:
## Addressed the P1 blocker — pushed e6fc120

@wyuc thanks, you're right — the ossKey fallback only covered the standalone asset entries; the slide base-frame path was still local-blob-only.

**Fix:** `resolveGeneratedMedia()` now resolves bytes through the same `resolveBytes()` helper (local blob first, then `ossKey` / `posterOssKey`) before creating the object URLs handed to `slideToPng`. So an evicted generated image/video is restored into the base-frame snapshot instead of having its `src` cleared. The type-mismatch clear (`element.src = ''`) and the immediate objectURL revoke are preserved.

**Test:** added a `frame base restores evicted generated media` block to `collect.test.ts`:
- evicted generated image → restored via `ossKey` (asserts the snapshotted slide's `src` is a `blob:` URL, not `''`, and the objectURL is revoked after);
- evicted generated video + poster → restored via `ossKey` / `posterOssKey`;
- image with no `ossKey` → still cleared to `''`.

The renderer snapshot is mocked (`vi.mock('@openmaic/renderer/snapshot')`) so the frame path runs in plain Node — this also sidesteps the missing built `@openmaic/renderer/snapshot` you hit in the review worktree, so the frame path is now covered without needing the real renderer.

Full `tests/video-export/` suite: 80 passed / 1 skipped (the skip needs `HF_E2E_DIR`). eslint + prettier clean.

wyuc (approved):
Re-review passed: the OSS fallback now restores evicted generated image/video/poster bytes before slide snapshotting, with focused frame-path regression coverage. The update from main does not touch the reviewed video-export files. Unit/typecheck and CodeQL pass; E2E is still running.

PR #937, opened 2026-07-15
feat(video-export): service-backed MP4 render + in-app one-click export (#866)

cosarah:
## Summary

Adds the last mile of classroom video export (#854): turning the self-contained Hyperframes project ZIP that #865 already builds in the browser into an **MP4**, one-click in-app, via an **isolated render service**. This is the **Part A** scope of #866 — the render service, app integration, docker-compose wiring, and graceful degrade. The isolated smoke-render CI job is deferred (export fidelity will be polished in follow-up issues).

The Hyperframes producer needs Node ≥22 + Chromium + FFmpeg and must not run in the Next.js route, so it lives in its own container. Rendering a 10-minute video takes ~10 minutes, so the flow is async: **submit → poll → download**, never a single long-lived connection.

## Related Issues

Part of #866 (does **not** close it — the isolated smoke/golden-render CI acceptance criterion remains a follow-up). Related to #854, depends on #865.

## Changes

- **`render-service/`** — standalone Node 22 + Chromium + FFmpeg container wrapping `@hyperframes/producer`'s **library** API (`createRenderJob`/`executeRenderJob`), not its bundled HTTP server (which only accepts a server-local `projectDir`, not our multi-file asset ZIP). Routes: `POST /render` (multipart ZIP) → `202 { jobId }`, `GET /render/:id` (poll), `GET /render/:id/download`, `DELETE /render/:id` (cancel), `GET /health`.
- **Scalability seams** — `JobStore` / `ArtifactStore` are interfaces. Part A ships `InMemoryJobStore` + `LocalDiskArtifactStore`; a demo-scale deployment drops in Redis + S3 (the download route already passes a `302` presigned URL straight through) **without touching the HTTP contract or the app**. Global concurrency + per-user single-job guards are config knobs.
- **App integration** — thin Next proxy routes under `app/api/export-video/*` (forward only, no rendering; `RENDER_SERVICE_URL` stays server-side) + a `capability` probe. New `use-render-video.ts` uploads the ZIP, polls via `runPolledTask`, downloads the MP4; shares the compile→emit→collect→zip prefix with the existing ZIP path via `build-export-zip.ts`.
- **Export menu** — resolution / fps / quality selectors + a progress bar; degrades to "Download project ZIP" when the service is unconfigured. i18n across all 8 locales.
- **docker-compose** — `render-service` under an opt-in `video-export` profile; `RENDER_SERVICE_URL` preset on the app service. README + `.env.example` documented.
- Entry is `main.ts` (not `server.ts`) because the producer auto-starts its own server on `:9847` when the process entry path ends with `/src/server.ts`.

## Type of Change

- [x] CI/CD or build changes (new Dockerfile + docker-compose profile)

### Steps to reproduce / test

1. `docker compose --profile video-export up --build`
2. In the export menu, pick resolution/fps/quality → **Render MP4**; watch the progress bar advance via polling; the MP4 downloads on completion.
3. Unset `RENDER_SERVICE_URL` → menu offers only **Download project ZIP** (unchanged #865 behavior).

- **Full render, real content**: rendered a real **640.6s (10.7 min)** classroom ZIP (44 audio segments + slide frames) end-to-end in the container → valid **H.264 720p + AAC MP4, duration 640.68s** (matches source), 26 MB, in **~9.6 min (~0.9× realtime, 4-worker frame capture)**. Bottleneck is frame capture (~85%), not encode (~55s).
- **Async progress**: status polled smoothly 20%→100% over the whole 9.6-min job — confirms submit/poll/download survives long renders without connection timeouts.
- **Degrade path**: with `RENDER_SERVICE_URL` unset, POST returns 501 and the client falls back to saving the ZIP.
- **Admission control**: queued-job cancel transitions to `cancelled` and reclaims the temp dir; per-user guard returns `429` for a second concurrent job by the same user, `202` for a different user.
- **Isolation**: main app `tsc`/lint/prettier/i18n pass without the service's Node-only deps (workspace + tsconfig + eslint exclusions verified); 80 existing video-export unit tests pass.
- **Not verified**: `standard`/`high` quality tiers and 1080p/4K timings (only `draft` 720p was benchmarked); no automated smoke-render CI yet (deferred by design).

### Evidence

- [x] Manually tested locally (containerized render of a real 10.7-min classroom → playable MP4, ffprobe-verified)

## Checklist

- [x] I have added/updated documentation as needed (README, render-service/README.md, .env.example)


wyuc (changes requested):
Thanks for putting the end-to-end service-backed render path together. I ran this at `d40366a02a406d5595fe6fcb12fa49582135b2b8`, including a real browser import and MP4 export. The functional path succeeds locally, but I think the deployment/resource boundaries need another pass before merge.

## Blocking findings

1. **The documented Compose path advertises MP4 but cannot render successfully in production mode.** `docker-compose.yml` always gives the app `RENDER_SERVICE_URL=http://render-service:9000`. Without the profile, capability reports enabled although the service is absent, and submit returns 502. With `--profile video-export`, the production SSRF guard rejects the private Compose hostname unless `ALLOW_LOCAL_NETWORKS=true`, which the documented command does not set. Submit then returns 500, while the client only falls back to ZIP on 501. Please make capability health/config aware and make the documented one-command deployment work without globally weakening SSRF protection.

2. **Upload and queue admission are currently unbounded.** The Next route materializes the complete multipart ZIP, the render service buffers it again, and extraction happens before admission with no compressed-size, expanded-size, entry-count, compression-ratio, queued-job, or queued-byte limits. A large archive/ZIP bomb can therefore exhaust the main app or renderer despite process isolation. Please stream/bound the proxy path and enforce strict archive and queue limits before extraction.

3. **`RENDER_MAX_JOBS_PER_USER` is ineffective in the in-app path.** The app never supplies `userId`; the service trusts an optional caller-provided multipart value, and direct clients can rotate it. Normal app submissions therefore skip the default per-user guard and enter an unbounded FIFO. Identity/admission should be derived server-side and reserved atomically before accepting/extracting the archive.

4. **Issue #866's isolated smoke/golden render CI AC is still missing, while this PR closes #866.** The PR explicitly defers that item and `render-service/package.json` has no smoke script/fixture/workflow. Please add the required Node/Chrome/FFmpeg smoke render, or keep/split the issue open rather than closing an unmet AC.

## Additional risks to address or explicitly follow up

- Uploaded HTML is executed by Chromium on the shared Compose network with outbound/internal-service reachability. The renderer should block non-loopback page requests and run on an isolated/authenticated network with bounded writable scratch space.
- Status, cancellation, and download are authorized only by opaque `jobId`; jobs are not bound to a server-derived principal.
- Running jobs have no service-side deadline, and client timeout/error paths do not cancel them, so a hung render can hold a slot and scratch data indefinitely.
- The download proxy's 120-second abort signal remains attached while streaming the response, so a large 1080p/4K file over a slow connection can be truncated. The client also buffers the whole MP4 via `blob()`.

## Verification performed

- Imported a real 8-scene `.maic.zip` through the browser.
- Rendered and downloaded a 512.875-second MP4 at 720p / 24 fps / draft through the app and isolated render service.
- Verified H.264 video, AAC 48 kHz stereo audio, full-file decode, and representative frames at the beginning/middle/end.
- `render-service` typecheck passed; ESLint reported no errors for changed TS/TSX files.

The actual rendering result looks correct; this review is blocking on deployment correctness and production resource/isolation controls rather than output fidelity.

cosarah:
Thanks for the thorough review — all four blocking findings are addressed in `1963085e`, plus the feasible additional risks. Summary:

**#1 — Compose advertises MP4 but can't render in prod**
- Capability now probes the service's `/health` (`checkRenderServiceHealth`), so a configured-but-absent service reports **disabled** and the UI degrades to ZIP rather than 502-ing.
- `RENDER_SERVICE_URL` is operator-supplied trusted config, so the proxy no longer runs it through the SSRF guard. The one-command `docker compose --profile video-export up --build` now works **without** globally weakening SSRF via `ALLOW_LOCAL_NETWORKS`.
- The client degrades to ZIP on **any** failed submit (not only 501).

**#2 — Unbounded upload + queue (ZIP-bomb / DoS)**
- Extraction is bounded **before decompression** via fflate's per-entry `filter`: entry count, per-entry and total expanded size, and compression ratio (env-tunable knobs).
- The proxy rejects oversized uploads (`413`) by `Content-Length` before forwarding.
- A global queue-depth cap (`RENDER_MAX_QUEUE`) rejects with `429`.

**#3 — Per-user guard ineffective + admission after extraction**
- Identity is derived **server-side** (client IP), forwarded as `x-openmaic-client`; the service ignores any client-supplied `userId` and the proxy strips it.
- Admission is split into `reserve()`/`submit()`/`release()` — the slot is reserved **before** extraction, so a rejected caller never triggers a decompression.

**Additional risks**
- Per-job wall-clock **watchdog** (`RENDER_JOB_DEADLINE_MS`) aborts + fails a hung render and frees the slot.
- Download proxy now bounds only time-to-headers, not the body stream, so large MP4s over slow links no longer truncate; the client also `DELETE`s the job when a started render fails/times out.
- Compose puts `render-service` on an `internal: true` network (no host/internet route), sandboxing the Chromium that runs the uploaded HTML. The export ZIP is self-contained, so no outbound is needed; the standalone caveat is documented in `render-service/README.md`.

**Verified in-container:** legal render 202; ZIP-bomb (entry-count + compression-ratio) rejected `400` before any decompression; per-identity guard `429` with a spoofed multipart `userId` ignored; reserve-before-extract leaves no scratch dir on rejection; watchdog aborts an overrunning job and frees the slot.

**Explicit follow-ups (not in this PR):**
- **#866 stays open** — the isolated smoke/golden-render CI acceptance criterion is deferred; PR retitled "Part of #866".
- Distributed/atomic admission across replicas (arrives with the Redis `JobStore` — the documented demo-layer swap).
- Binding status/cancel/download authz to a server principal beyond the opaque `jobId` (the app has no per-user login today; needs real auth).

wyuc:
Follow-up from the real-browser E2E: the MP4 is playable, but narration timing is currently not export-fidelity safe.

I heard sentences being cut off in the rendered output and checked the attached project. The manifest marks every present narration clip as `source: "estimated"`. Comparing the 47 emitted MP3s with `ffprobe`:

- 47/47 actual clips are longer than the scheduled duration
- average underestimate: 4.353s
- maximum underestimate: 10.230s (11.85s scheduled vs. 22.08s actual)

Because each `<audio>` clip's `data-duration` and the next clip's start come from those estimates, the renderer stops clips mid-sentence and advances the timeline early. Please resolve the real media duration after collection and recompute narration starts, scene duration, total duration, and subtitle cues from it. Estimates should only be a fallback when no audio asset exists.

The project ZIP already includes SRT/VTT, but those cues use the same estimated boundaries and the final MP4 has only H.264 + AAC streams. Please add basic subtitles to the final MP4 (burn-in is acceptable for v1; an embedded selectable track is also fine) using corrected timings.

E2E artifacts from `d40366a`:

- [Hyperframes project ZIP](https://github.com/wyuc/OpenMAIC/releases/download/pr-937-e2e-d40366a/pr-937-video-project.zip)
- [Rendered 720p/24fps draft MP4](https://github.com/wyuc/OpenMAIC/releases/download/pr-937-e2e-d40366a/pr-937-rendered-output.mp4)

These artifacts are hosted in a clearly labeled prerelease on my fork solely because GitHub's controlled comment uploader would not accept the local files directly.

cosarah:
Thanks for the deeper E2E — both fidelity issues are fixed in `6a585c29`.

**Narration timing (estimated → real):** you're right that the root cause is missing `AudioFileRecord.duration` — it's only been recorded since #861, so existing classrooms fall back to text-length estimates. Rather than a post-collection recompute, I made it symmetric with the existing **video** probe: `timeline-deps` now reads each narration clip's real duration from its blob via an off-document `<audio>` (`probeAudioDurationMs`), and `audioDurationMs()` prefers that, falling back to the stored duration, then the estimate only when no audio asset exists. Because the pure compiler derives narration starts, scene/total duration, and subtitle cues from that value, all of them correct together with no compiler change.

**Subtitles in the MP4:** the producer has no subtitle track of its own, so the emitter now **burns them in** — a caption overlay with one hidden `<div>` per cue, revealed/hidden by the same paused GSAP timeline at each cue's start/end (using the corrected timings above), captured into frames by Chromium.

**Verified:**
- Emitter unit tests + snapshot updated (overlay + `tl.set` visible/hidden toggles, escaped text, hidden-by-default); all 82 `tests/video-export` pass including the determinism red-line proxy.
- Rendered a synthetic subtitle project through the container and confirmed by pixel analysis that the caption appears only within its cue window: **2429 near-white pixels** in the caption band at t=1.5s vs **0** at t=0.05s (before the cue).

This is on top of the deployment/resource fixes from the earlier review (`1963085e`); the branch is also merged up to latest `main`. Ready for another look when you have a chance.

wyuc (changes requested):
Cross-review + real-browser E2E follow-up at `6a585c2921972385da0ed12286916eb00de0ae34`.

I ran two independent report-only reviewers (remote Claude and Codex) against the same pinned PR diff, then exercised the complete UI flow against this head: generated a real 2-scene classroom with 12 TTS clips, selected 720p / 24fps / draft, and completed submit → poll → download through the isolated render service.

## Blocking findings

1. **[P1] Multi-cue subtitles retain layout space and move the active cue out of position**  
   `lib/video-export/emit-hyperframes/index.ts:150`

   Every inactive cue is still `display:inline-block` with `visibility:hidden`, so it remains in normal flow. With 12 real cues, the overlay becomes multiple rows tall and the active cue keeps its row among all hidden cues.

   This reproduced clearly in the exported MP4:
   - at 20s, cue 2 should be active (11.856–23.736s) but no subtitle is visible;
   - at 60s, cue 6 is at the very top and overlaps the slide title;
   - later cues drift through the title area and middle of the slide.

   The current single-cue snapshot/pixel test cannot catch this. Please position all cues in the same absolute slot or remove inactive cues from layout with `display:none`.

2. **[P1] The compressed upload limit is still bypassable before both processes buffer the body**  
   `app/api/export-video/render/route.ts:39-50`  
   `render-service/src/main.ts:64-97`  
   `render-service/src/config.ts:36`

   The app trusts `Content-Length`; an absent, malformed, or chunked request passes the check and is fully buffered by `req.formData()`. The render service buffers it again through `formData()` / `file.arrayBuffer()` before admission. `config.maxUploadBytes` is documented but never read. Concurrent chunked uploads can still OOM both Node processes before archive/queue guards run.

3. **[P1] `makeProjectDir()` failure permanently leaks the reservation**  
   `render-service/src/main.ts:87-106`

   `reserve()` increments pending and per-identity counts, but `makeProjectDir()` is outside the catch that calls `release()`. The standalone instructions do not create the default `/tmp/openmaic-renders`, so the documented non-Docker path can fail with `ENOENT` on every submit and eventually wedge the service at “queue full” until restart. A transient `ENOSPC` has the same leak in Docker.

## Additional actionable findings

- **[P2] Untrusted forwarding headers feed the per-identity guard.** The documented Compose topology exposes the app directly, so clients can rotate `x-forwarded-for` to fill the queue; without forwarding headers, all callers collapse to `anonymous`. Establish an explicit trusted-proxy boundary or use a signed/session identity.  
  `app/api/export-video/render/route.ts:22-26`

- **[P2] Audio metadata probes are serialized and have no timeout.** One blob that never fires `loadedmetadata` or `error` leaves ZIP/MP4 export stuck in `compiling` forever. Use bounded concurrency and a per-probe timeout that falls back to stored duration/estimate.  
  `lib/video-export-app/timeline-deps.ts:88-103,137-142`

- **[P2] `RENDER_MAX_JOBS_PER_USER=0` cannot disable the guard as documented.** `intEnv` rejects zero and silently falls back to 1.  
  `render-service/src/config.ts:7`

- **[P2] The new service is outside repository CI quality gates and has no security-boundary tests.** Root eslint/tsc exclude it, its own typecheck is not wired into CI, and there are no tests for unzip limits/path traversal, reservation arithmetic, cancellation, or the API routes.  
  `eslint.config.mjs:32-34`, `tsconfig.json:33-42`, `render-service/package.json:10-13`

- **[P2] `unzipSync` can block the service event loop while expanding up to the 1GB default ceiling**, stalling health and poll responses. Prefer async extraction/worker isolation and tighter evidence-based defaults.  
  `render-service/src/unzip.ts:31-52`

- **[P3] UX/docs cleanup:** compiling shows a static 0% ring while the tooltip says Export PPTX; capability is re-probed on every menu mount and causes layout shift; the producer progress-range comment contradicts the normalization code.

## E2E result

The core render path does work: the output was a playable H.264 1280×720 MP4 with AAC 48kHz stereo, duration 159.445s, 3,826 frames. All 12 MP3 durations exactly matched their 12 SRT cue lengths, so the real-media duration fix successfully resolves the earlier narration truncation/timeline issue.

Please keep this PR in changes-requested state until the three P1 items are fixed and covered by regression tests.

wyuc (changes requested):
Round-2 cross-review + real-browser E2E follow-up at `cdc020f497e0b2f137178b3be7eab973090cbdbe`.

I reran two independent report-only reviewers against the complete pinned PR diff, then started a fresh real UI flow on this head: generated a 2-scene classroom with 13 TTS clips, opened video export, and attempted to build/run the exact render-service image.

## Blocking findings

1. **[P1] The committed render-service lockfile makes CI and the production Dockerfile unbuildable**  
   `render-service/package-lock.json`

   Both the current GitHub Render Service job and the exact PR Dockerfile fail during `npm ci` / `npm ci --omit=dev`:

   ```
   EBADPLATFORM: Unsupported platform for @esbuild/aix-ppc64@0.28.1
   wanted aix/ppc64, current linux/x64
   ```

   The nested `node_modules/vitest/node_modules/@esbuild/aix-ppc64` lock entry is missing `optional: true`. Because the production image cannot build, this round could not honestly complete MP4 submit → poll → download. I did not use an old image or modify the lockfile to make the E2E pass.

2. **[P1] Admission still happens after multipart is fully buffered**  
   `render-service/src/main.ts:80-116`

   The raw byte cap prevents an individual body from exceeding 300 MB, but `bounded.formData()` materializes the complete request before `manager.reserve()`. Concurrent near-cap requests therefore bypass queue/per-identity admission while each buffers hundreds of MB; `file.arrayBuffer()` and `unzipSync()` add more copies. Admission needs to precede buffering, and multipart files should stream to bounded disk/storage.

3. **[P1] Untrusted Chromium can still reach the main application**  
   `docker-compose.yml:11-13,46-47`

   The render network is `internal: true`, but both `render-service` and `openmaic` join it. Uploaded HTML cannot reach the internet directly, but it can initiate requests to `http://openmaic:3000`, potentially using the app as an internal/egress bridge. The renderer needs a boundary where Chromium cannot initiate connections back to the app.

4. **[P1] Default Compose effectively allows one render for the entire deployment**  
   `app/api/export-video/render/route.ts:27-32`  
   `render-service/src/config.ts:33`

   Unless `TRUST_PROXY_HEADERS=true`, every caller is assigned the literal identity `direct`; the default `RENDER_MAX_JOBS_PER_USER=1` therefore makes the second user receive 429 for the full duration of the first user's render. `RENDER_MAX_CONCURRENCY=2` and the FIFO queue are unreachable under the documented defaults.

5. **[P1] Whole-archive synchronous extraction still has a large global RAM multiplier**  
   `render-service/src/unzip.ts:31-55`  
   `render-service/src/main.ts:129`

   `unzipSync` materializes every expanded entry in RAM and permits up to 1 GB per archive. Multiple admitted extractions multiply that ceiling; there is no extraction concurrency cap or container memory limit, and the event loop is blocked during expansion. Stream entries to disk/worker isolation and bind extraction concurrency to the actual resource budget.

## Additional actionable findings

- **[P2] 60-second route/fetch budget contradicts the 300 MB body limit.** A legal upload needs roughly 40 Mbps sustained bandwidth; slower valid uploads abort with 502.  
  `app/api/export-video/render/route.ts:12,67-77`

- **[P2] 429/413/502/501 are all reported as “service unavailable” and force an unsolicited ZIP download.** Preserve the real status/detail and make ZIP fallback explicit for temporary/retryable failures.  
  `lib/store/video-render.ts:228-238`

- **[P2] ZIP-export duplicate-submit guard regressed.** `useExportVideo` now lives in the menu that immediately unmounts after clicking ZIP; reopening creates a fresh `exportingRef` and allows a second concurrent snapshot/ZIP pipeline.  
  `components/stage/header-controls.tsx:82-90,254-282`

- **[P2] `.env.example` still tells operators to set `ALLOW_LOCAL_NETWORKS=true` for a render target, although `RENDER_SERVICE_URL` deliberately bypasses that SSRF guard.**  
  `.env.example:275`

- **[P3]** Deadline overruns are marked `cancelled` instead of `failed`; a future fallible `JobStore.create` can leak an identity slot after `reservation.consumed=true`; CI should set `PUPPETEER_SKIP_DOWNLOAD=true`; the 210 MB unzip fixture risks Vitest's default timeout.

## What passed in the real flow

- Fresh classroom generation completed through the UI: 2 content scenes, 13 TTS clips.
- With the render service unavailable, the export menu correctly degraded to Download Project ZIP and explained that MP4 needs the service.
- The project ZIP downloaded successfully.
- The subtitle layout fix is structurally correct in the real emitted project: all 13 cues share `grid-area:1/1`, inactive cues use `display:none`, and only the active cue becomes `inline-block`.
- All 13 MP3 durations match the 13 SRT cue lengths to millisecond rounding.
- The focused subtitle unit test passes 13/13.

Please keep the PR in changes-requested state until the production image builds and the P1 resource/isolation issues are addressed.

wyuc (changes requested):
Re-review at `88613c78ead1c25f09cb6b0ce624d31a4025ce76`. The previous lockfile/build, admission ordering, default-concurrency, async extraction, timeout/status, ZIP dedupe, and subtitle/audio-timing findings are addressed. Focused tests pass, and I completed a fresh real service smoke: multipart submit → poll → download produced a nonblank 3.000s H.264 640×360/24fps MP4. Two production-boundary blockers remain.

## Blocking findings

1. **[P1] The extraction semaphore is entered after every admitted multipart body has already been buffered**  
   `render-service/src/main.ts:101-143`

   `bounded.formData()` completes before `extractionGate.run()`. In Node this materializes the uploaded file into the `File`; the gate only bounds the subsequent `file.arrayBuffer()` copy and unzip. With the defaults, up to `RENDER_MAX_QUEUE=20` admitted requests can therefore each buffer nearly 300 MB before the 2-permit gate, i.e. roughly 6 GB of compressed bodies against the Compose `mem_limit: 4g`, before expanded bytes/Chromium are counted. A concurrent near-cap burst can still OOM the service. Please put multipart buffering itself behind a bounded upload/extraction permit, or stream the file part directly to bounded scratch storage before parsing/extraction. The queue reservation should remain before that permit. Add an integration test proving only the configured number of bodies can enter the buffering section concurrently.

2. **[P1] The Chromium egress security boundary fails open while the service still reports healthy**  
   `render-service/docker-entrypoint.sh:32-37`

   Uploaded HTML is untrusted and the renderer shares the `render` network with `openmaic`. If iptables setup fails (missing capability, backend mismatch, partial rule failure), the entrypoint only warns and starts normally; `/health` still returns `{ok:true}`, so the app advertises MP4 rendering while Chromium can initiate connections to the app. This does not guarantee the boundary required by the previous review. When `RENDER_EGRESS_LOCKDOWN=true`, failure should stop startup (or make health/capability fail closed). Operators who intentionally accept an unisolated standalone setup can explicitly set the toggle false. Please also add a container smoke/assertion that a new connection from the render container to the app/network is blocked while loopback rendering and inbound-response traffic still work.

## Verification

- `PUPPETEER_SKIP_DOWNLOAD=true npm ci`: pass
- render service: 22/22 tests + `tsc --noEmit`: pass
- video-export/choreography/SSRF focused suite: 164 passed, 1 skipped
- real standalone producer smoke: submit/poll/download passed; ffprobe reports H.264, 640×360, 24fps, 3.000s; representative frame is nonblank
- GitHub CI is green, including Render Service and E2E

Docker Desktop on this machine currently returns a daemon API 500, so I could not independently rebuild the Compose image in this round; the committed lockfile now installs locally and the GitHub Render Service job is green.

wyuc (approved):
Re-review at `596ab077536ac198f257fa05b9ee78e4a7803e92`.

The two remaining round-3 blockers are addressed:

- Queue/per-identity reservation still happens before body consumption, while multipart buffering, `file.arrayBuffer()`, and extraction now all run inside the bounded `extractionGate`. Excess requests remain backpressured with unconsumed bodies instead of stacking buffered archives in memory.
- IPv4 egress lockdown now fails closed when root, iptables, CAP_NET_ADMIN, or rule installation is unavailable. The container smoke covers startup, loopback health, and blocked new outbound connections.

Focused verification:
- `npm ci --ignore-scripts`: pass
- render service tests: 23/23 pass
- render service typecheck: pass
- GitHub CI is green, including Render Service and E2E

Non-blocking follow-up: the IPv6 branch still ignores every `ip6tables` failure with `|| true`. The documented Compose network is IPv4-only by default, so this does not block the supported deployment, but a future IPv6-enabled network would reopen egress while logging “lockdown active”. Please either fail closed when IPv6 is enabled or explicitly disable IPv6 for the render network/container.

Approved. The branch is currently behind `main`, so please sync it and let CI rerun before merge.

PR #938, opened 2026-07-16
fix: render whiteboard math via KaTeX instead of raw LaTeX text

### LINKED ISSUE #424: bug: whiteboard math equations render as raw LaTeX text instead of formatted equations
## Description

When a teacher agent draws mathematical equations on the whiteboard during a live session, some equations render correctly while others display as raw LaTeX strings in red (e.g., `\omega_{out} = \omega_{in}\cdot\frac{T_{in}}{T_{out}}`).

## Steps to Reproduce

1. Start a classroom session on a math-heavy topic (e.g., gear ratios, physics equations)
2. Let the teacher agent use the whiteboard to write equations
3. Observe that some equations render as broken red text instead of formatted math

## Expected Behavior

All mathematical equations on the whiteboard should be rendered using KaTeX (via `wb_draw_latex`), displaying properly formatted math notation.

## Actual Behavior

Some equations are written to the whiteboard using `wb_draw_text` with raw LaTeX content. Since `wb_draw_text` does no KaTeX processing, the LaTeX syntax is displayed as plain red text (e.g., `\omega_{out} = \omega_{in}\cdot\frac{T_{in}}{T_{out}}`).

## Root Cause Analysis

Three compounding issues:

1. **Ambiguous tool description** (`lib/orchestration/tool-schemas.ts`): The `wb_draw_text` description says "Use for writing **formulas**, steps, or key points." The word "formulas" causes the LLM to occasionally use `wb_draw_text` for math expressions that require KaTeX rendering instead of `wb_draw_latex`.

2. **Missing negative example** (`lib/orchestration/prompt-builder.ts`): The "Bad Examples" section in the agent prompt has no example showing that raw LaTeX in `wb_draw_text` is incorrect, so the LLM has no explicit signal to avoid this pattern.

3. **No defensive fallback** (`lib/action/engine.ts`): When the LLM incorrectly passes raw LaTeX to `wb_draw_text`, the engine wraps it in a plain `<p>` HTML tag with no KaTeX processing, rendering the LaTeX syntax as-is to the user.

## Environment

- Topic: Antikythera mechanism / gear ratios (math-heavy)
- Affected actions: `wb_draw_text` calls containing LaTeX expressions

Image

mvanhorn:
## Summary

Fix all three layers so the failure is both discouraged and self-healing. In `tool-schemas.ts`, reword the `wb_draw_text` description to drop "formulas" and explicitly steer math/equations to `wb_draw_latex` (mirror the `wb_draw_latex` wording so the boundary is unambiguous).

## Related Issues

Fixes #424

## Changes

Fix all three layers so the failure is both discouraged and self-healing. In `tool-schemas.ts`, reword the `wb_draw_text` description to drop "formulas" and explicitly steer math/equations to `wb_draw_latex` (mirror the `wb_draw_latex` wording so the boundary is unambiguous). In `prompt-builder.ts`, add a concise Bad Example demonstrating that raw LaTeX passed to `wb_draw_text` is wrong and that `wb_draw_latex` must be used for equations. In `engine.ts`, add a defensive fallback in `executeWbDrawText`: when the incoming `content` looks like a LaTeX math expression (heuristic on common markers such as `\frac`, `\cdot`, `\omega`, `_{`, `^{`, `$...$`, backslash-command density), render it through the same KaTeX path `executeWbDrawLatex` already uses instead of wrapping it in a bare `<p>`, so a mis-routed action still displays formatted math. Keep the heuristic conservative to avoid KaTeX-processing ordinary prose, and rely on `throwOnError: false` (already used by the latex path) so non-math strings degrade gracefully. This is a bug fix, which CONTRIBUTING.md says to submit as a PR directly; the PR must link the issue (`Closes #424`), fill the PR template, and be marked AI-assisted per repo policy.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Render whiteboard content containing LaTeX math via `executeWbDrawText`.
2. Confirm math renders through KaTeX instead of raw source.

- Added vitest tests covering the KaTeX render path; reviewed the diff before submission.

### Evidence

- [x] Manually tested locally

## Checklist

AI was used for assistance.


wyuc (changes requested):
Thanks for addressing this across the prompt, schema, runtime fallback, and tests. The overall direction is good, and the targeted test suite passes, but there is one functional gap to fix before merge.

The fallback currently misses valid single-command LaTeX expressions whose command is not in `COMMON_LATEX_COMMAND`. For example, `\vec{v}`, `\overline{x}`, and `\hat{x}` each contain only one command, so `commands.length < 2` returns `null` and they are still rendered as raw `wb_draw_text`. Please recognize a single LaTeX command when it is accompanied by sufficiently strong math syntax such as a braced argument, while keeping path-like prose such as `C:\temp` as text, and add regression tests for both cases.

This branch also conflicts with `main`. Please rebase/merge `main` yourself and preserve the newer `ActionExecutionOptions` / `silent` behavior in `ActionEngine`; when the fallback delegates to `executeWbDrawLatex`, it should forward those execution options so playback does not regain the drawing delay in silent mode.

Once those are addressed, this should be ready for another review.

mvanhorn:
Both addressed. Merged main and kept the newer ActionExecutionOptions/silent path in executeWbDrawText, forwarding options through to executeWbDrawLatex so the fallback no longer regains the drawing delay in silent mode. For the functional gap, getLikelyLatexMath now treats a single command as math when it carries a braced argument, so \vec{v}, \hat{x}, and \overline{x} render via KaTeX, while path-like prose such as C:\temp (a command token with no braced argument) still stays text. Added regression tests for both directions; the focused suite passes locally (10 cases, including the \vec{v} -> latex and C:\temp / D:\path\to\file -> text cases).

wyuc (changes requested):
Thanks for addressing the previous review points. The `silent` option forwarding and single-command cases now look correct, but the path guard is still incomplete.

`COMMON_LATEX_COMMAND.test(trimmed)` runs before the path-like check, so valid paths such as `C:\alpha`, `C:\theta`, or `C:\pi` are still classified as LaTeX and rendered through KaTeX. Please exclude Windows/path-like text before accepting common commands, and add regression coverage for a path whose segment matches `COMMON_LATEX_COMMAND`.

The branch is also behind `main`; please update it and rerun the checks after this fix.

mvanhorn:
Good catch, the ordering was the bug. The drive-letter check now runs before `COMMON_LATEX_COMMAND`, so `C:\alpha`, `C:\theta` and `C:\pi` fall through to text.

Added `C:\alpha` to the Windows-path cases, which is the segment-matches-a-common-command case you asked for. It fails without the reorder and passes with it, 11/11 in `engine-latex-fallback`.

Branch is updated with main as well.

wyuc (approved):
The requested path guard and regression coverage are now in place. The previous review findings are resolved.

mvanhorn:
Thank you @wyuc for seeing the KaTeX rendering change through review. Appreciate the thorough passes on it.

Issue #939, opened 2026-07-16
[Task] @openmaic/storage server-backed runtime — backend seam, HTTP contract + client, Postgres reference (#779)


wyuc:
Tracking: #779 (RFC), #869 (runtime layer). Follows the browser delivery — #880 (RuntimeStore, Part B) — and the app cutover series #885 / #893 / #902 / #922 / #926.

## Motivation

Two standing gaps converge here:

1. **Self-hosted deployments have no server persistence** (#53, #596). Runtime data — what a learner produces — currently lives only in the browser's IndexedDB: clearing site data loses progress, and no deployment can offer cross-device continuity.
2. **The contract has only one real backend.** A contract validated by a single implementation tends to bake that implementation's shape in (the same reason renderer v2 requires a second consumer before freezing its API, #851). The invariants a browser backend gets for free — per-session `seq` atomicity under concurrent appends, `mergeLearner` idempotence, cascade transaction boundaries — are exactly the ones a SQL implementation must earn, so a second backend is what proves the contract is real.

This task delivers the pluggable seam #779 promised ("the pluggable seam is HTTP, not the database driver"), scoped to the runtime component first.

## Part A — backend seam in the app

`lib/runtime/store.ts` constructs the browser store directly. Make the store (and the learner-key provider) injectable at app bootstrap, so a deployment supplies a different `RuntimeStore` implementation without forking app code. Default remains the browser backend; zero behavior change.

## Part B — HTTP contract + client backend

A documented HTTP contract for `RuntimeStore` (sessions CRUD, atomic append, list-by-anchor, `mergeLearner`, cascades) and an `HttpRuntimeStore` client that implements the store interface over it. The contract suite (`runRuntimeStoreContract`) runs against the client backed by a conformance server.

## Part C — Postgres reference implementation

`PgRuntimeStore` over plain SQL (`pg` driver, deliberately no ORM so hosts can port the queries into their own stack): two tables (`runtime_sessions`, `runtime_records`), store-assigned per-session monotonic `seq` safe under concurrent appends, idempotent `mergeLearner`, document-cascade deletes. Contract suite green in CI against a real Postgres (service container).

## Part D — reference server

[...]

PR #946, opened 2026-07-17
feat(storage): server-backed runtime — pluggable backend seam, HTTP contract, Postgres backend, reference server (#939)

wyuc:
Delivers all four parts of #939 as one reviewed unit, per the integration-branch flow announced there. Individual parts landed on `runtime-server-backend` through per-part PRs (#940 Part B, #941 Part C, #943 interface conformance, #944 Part A, #945 Part D), each taken through multi-round cross-review to zero findings plus CI including a real-Postgres contract lane.

## What this enables

Runtime data (chat, quiz attempts, playback — what a learner produces) currently lives only in browser IndexedDB: clearing site data loses progress and no deployment can offer cross-device continuity (#53, #596). After this, a deployment can keep the app code unchanged and plug in a server-backed store:

```ts
configureRuntimeStorage({
  store: new HttpRuntimeStore({ baseUrl: '/api' }),
  learnerKey: () => session.userId,
});
```

## The four parts

- **Part A — app seam** (`lib/runtime/config.ts`): injectable `RuntimeStore` + learner-key provider at bootstrap. Unconfigured behavior is byte-identical to today (full suite green with zero behavior change). Single-shot, sealed after first use, fail-loud guards.
- **Part B — HTTP contract + client** (`docs/runtime-http-contract.md`, `./runtime/http`): documented JSON endpoints for every store operation; server-assigned `seq`; machine-readable error codes; client re-validates and re-migrates everything it reads.
- **Part C — Postgres backend** (`./runtime/pg`): `PgRuntimeStore` over an injected `Queryable` (node-postgres and PGlite both fit; zero new runtime deps). Session-row locking makes per-session `seq` atomic under concurrency; full contract suite runs against real Postgres 16 in CI.
- **Part D — reference server** (`./server`): wires the contract onto any store with the security model the contract mandates: `learnerKey` derives from authentication and is never trusted from the request; merge and admin surfaces are default-deny capability hooks; unowned sessions read as 404 to avoid existence oracles. Threat model in `docs/reference-server.md`.

Shared across the JSON backends: a lossless-JSON payload gate (`json-value.ts`) that fails loud on anything `JSON.stringify` would silently alter (structured-clone-only values, lone surrogates, NUL, accessors, prototype `toJSON`), keeping the payload domain identical across backends instead of letting data quietly change between write and read.

## Review trail

Every part went through cross-vendor review loops (Claude + GPT reviewers, adversarial prompts) with fixes to zero findings — ~50 findings addressed across the branch, from a P1 stage-deletion bypass on injected backends to JSON-fidelity corners. CI on every part PR: lint/typecheck/unit, E2E, and the real-Postgres contract lane.

Requesting a full review pass on the complete picture before this lands on `main`.


cosarah:
Reviewed the 22-file diff in a worktree — typecheck and the full test suite pass (289 tests; real-Postgres lane is CI-only). Overall this is solid, library-grade work: the JSON-fidelity gate, the untrusted-`learnerKey` security model, and the `appendRecord` locking/retry path are all carefully done. LGTM.

One minor point worth a look: `GET /runtime/sessions/{id}/records` slightly contradicts the docs' existence-concealment goal. `docs/reference-server.md` (~L48) says session routes return the same `404` whether a session is another learner's or absent, but in `src/server/index.ts:503-516` another learner's session → `404` while an absent session → `200 []`. So a `404` here leaks "this id belongs to someone else." Impact is tiny (unguessable UUIDs) and it's a real trade-off against the "listRecords returns empty for absent" contract rule — just consider returning `404` for absent sessions on this route too, or documenting the exception. Not blocking.

wyuc:
Good catch — fixed rather than documented, since the clean resolution exists: the records route now answers absent and foreign sessions identically (`404 SESSION_NOT_FOUND`), and `HttpRuntimeStore.listRecords` maps that code back to `[]`, so the store contract's "absent lists as empty" holds unchanged at the client boundary. The contract doc spells out the server MAY/SHOULD (authorization-aware servers SHOULD conceal) and the client MUST. Added a test asserting the two cases are byte-identical at the HTTP layer plus one for the client mapping; full suite green (291 passed).

cosarah (approved):
LGTM

PR #948, opened 2026-07-17
Add Atlas Cloud LLM provider

binyangzhu000-sudo:
## Summary
- Add Atlas Cloud as a built-in OpenAI-compatible LLM provider using `ATLASCLOUD_API_KEY` and `https://api.atlascloud.ai/v1`.
- Register live-verified Atlas Cloud models (`qwen/qwen3.5-flash` and `deepseek-ai/deepseek-v4-pro`) and reuse the existing DeepSeek reasoning metadata for the Atlas DeepSeek model.
- Wire Atlas Cloud into server env config parsing and add focused tests for provider registration, model creation, env mapping, and reasoning metadata.

## Validation
- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm exec vitest run tests/ai/atlascloud-provider.test.ts tests/server/provider-config.test.ts` (59 passed)
- `pnpm exec prettier lib/types/provider.ts lib/server/provider-config.ts lib/ai/providers.ts lib/ai/model-metadata.ts tests/server/provider-config.test.ts tests/ai/atlascloud-provider.test.ts --check`
- `git diff --check`
- Atlas Cloud live `/v1/models` returned 118 models and confirmed both `qwen/qwen3.5-flash` and `deepseek-ai/deepseek-v4-pro`.

Note: `pnpm exec tsc --noEmit --pretty false` currently fails on existing repository-wide type errors/missing workspace declarations such as `@openmaic/dsl`, `@openmaic/storage`, and several pre-existing implicit `any` issues outside these changed files.

## README / promotion compliance
No README changes. This only adds provider code, focused tests, and `.env.example` configuration placeholders; no sponsor, logo, credits, or partner promotion content is added.


YizukiAme:
Hi @binyangzhu000-sudo, thanks for the contribution! The current CI failure is caused by missing thinking metadata for `atlascloud:qwen/qwen3.5-flash`:

```
expected [] but received ["atlascloud:qwen/qwen3.5-flash"]
```

Every built-in model needs either an entry in `THINKING_CAPABILITIES` or an explicit entry in the non-configurable-thinking allowlist (`MODELS_WITHOUT_CONFIGURABLE_THINKING` in `tests/ai/model-metadata.test.ts`).

Please verify how Atlas Cloud handles thinking parameters for both Qwen and DeepSeek through its OpenAI-compatible endpoint:

* If Qwen supports `enable_thinking` / `thinking_budget`, add the corresponding Atlas Cloud metadata mapping.
* Otherwise, explicitly add it to the non-configurable-thinking allowlist.
* Please also confirm that Atlas Cloud accepts the DeepSeek reasoning parameters before reusing `deepseekEffort`; verifying `/v1/models` alone does not establish protocol compatibility.

After updating this, please run:

```
pnpm exec vitest run tests/ai/model-metadata.test.ts tests/ai/atlascloud-provider.test.ts
```

binyangzhu000-sudo:
Thanks for the pointer. I pushed a follow-up commit that explicitly allowlists `atlascloud:qwen/qwen3.5-flash` as a non-configurable-thinking model.\n\nI also verified the Atlas Cloud OpenAI-compatible endpoint with a real request using a browser User-Agent:\n\n- `qwen/qwen3.5-flash` returned a normal chat completion without thinking params.\n- `deepseek-ai/deepseek-v4-pro` accepted `reasoning_effort: low` as a top-level request field and also accepted the same value through the current extra-body style wrapper.\n\nLocal validation now passes:\n\n`pnpm exec vitest run tests/ai/model-metadata.test.ts tests/ai/atlascloud-provider.test.ts tests/server/provider-config.test.ts`\n\nResult: 3 files / 63 tests passed.

wyuc (changes requested):
The current validation does not match the payload produced by the existing `deepseek` reasoning adapter. The adapter sends `thinking: { type: "enabled" }` together with `reasoning_effort: "high"` or `"max"`, while the reported live test only covers `reasoning_effort: "low"`. It also sends `thinking: { type: "disabled" }` when reasoning is disabled.

Please verify these exact enabled and disabled payloads against Atlas Cloud, and verify one real tool call before declaring `tools: true`. If Atlas Cloud does not support the existing DeepSeek payload contract, please add provider-specific metadata or conservatively disable the unverified capabilities.

binyangzhu000-sudo:
Follow-up pushed in b478b36. I verified the exact adapter contract against Atlas Cloud before keeping these capabilities enabled:

- thinking enabled with reasoning_effort=high: HTTP 200
- thinking enabled with reasoning_effort=max: HTTP 200 with reasoning_content
- thinking disabled: HTTP 200
- forced OpenAI-compatible get_weather function call: HTTP 200, finish_reason=tool_calls, one valid tool call

I also added regression coverage for the enabled/high, enabled/max, and disabled request payloads. Local verification: 3 test files / 66 tests passed, target ESLint passed, and git diff --check passed. No credentials or response content are included.

wyuc (approved):

Issue #867, opened 2026-07-07
[Task] Video export polish + higher-fidelity fidelity track (incremental)


cosarah:
## Context

Sub-issue of #854 (Hyperframes-based classroom video export architecture). Follows the L1 emitter #865 and service render (P2). This is the ongoing polish + higher-fidelity track, delivered as small incremental PRs — none of it blocks the first usable export.

## Goal

Improve fidelity and finish beyond the L1 deterministic slice: caption styling, placeholder-card visuals, resolution tiers, incremental descriptor adoption, and eventually richer scene-family treatment.

## Scope (each item a small standalone PR)

1. **Placeholder-card visuals** for non-deterministic / unsupported families (quiz shows question count/score, PBL hero visual, interactive first-frame or icon card) — represented intentionally per #854's AC.
2. **Burn-in subtitle styling** (optional, off by default) + SRT/VTT polish; multi-language stays single-language driven by the speech text for now.
3. **Resolution tiers** (recover #746's UX; e.g. 720p/1080p/2K/4K).
4. **Overlay components read animation descriptors** — migrate app-side spotlight/laser to consume the #863 descriptors, closing the last gap toward one source of truth.
5. **Spotlight geometry unification** — make spotlight's geometry injectable so app (DOM measurement) and exporter (pure `findElementGeometry`) share one render component; add a diff test quantifying the outer-box vs content-box delta.
6. Diagnostics surfaced in export UI ("N interactive sections shown as cards").

## Research track (evaluate, not necessarily commit)

- **L2** — live HTML slide base (SlideCanvas SSR / DOM serialization) enabling text-level animation.
- **L3** — real widget/interactive rendering (scripted postMessage replay + virtual-time capture).
- Full classroom playback recording — evaluate only if compiler-based export cannot cover a runtime behavior cleanly.

## Non-goals

- Anything that blocks the first mergeable export slice.

## Acceptance criteria

- Each polish item lands as an independent PR with its own tests, without regressing the L1 path.
- Research items produce an experiment report before any commitment to implement.

Follows #865 and the service-render phase.

PR #952, opened 2026-07-20
feat(video-export): fidelity polish — spotlight geometry, video clips, formulas, subtitles

cosarah:
## Context

Fidelity polish for the Hyperframes classroom video export (issue #867, follows #865/#866). Fixes four observed problems and reworks subtitles/UX. Each concern is a self-contained commit.

## Changes

1. **Lost video clips** (`0b4918ec`) — `timeline-deps` looked generated videos up by slide element `.id`, but media records are keyed by media ref (`gen_vid_…`); every generated video missed → no `<video>` emitted, no bytes collected. Added an elementId→mediaRef bridge (mirrors the live engine's `resolveMediaPlaceholderId`).

2. **Formula distortion** (`a781ff25`) — KaTeX sizes large delimiters (`\left\{`, `\begin{cases}`) from its `KaTeX_Size` fonts, which load async. The fit-scale was measured once on mount against the fallback font and never recomputed, so a cold snapshot baked a stale scale and the brace desynced from the piecewise body. Re-measure on `document.fonts.ready` / `loadingdone`; snapshot waits two frames for the re-measure to commit.

3. **Spotlight/laser/video offset** (`67f21aea`, #867 item 5) — the pure geometry helper uses the authored outer box against a fixed 1000×562.5 base, but the live overlays (and the frame PNG) measure the `.element-content` box (auto-height text + 10px padding). Added an optional `GeometryProbe` to the compiler DI: when supplied, effects use the measured content-box geometry, degrading to the pure calc on a miss. New `measureSlideElementGeometry` (in `@openmaic/renderer/snapshot`) measures the same box the live `SpotlightOverlay` does. Includes a diff test quantifying the outer-box vs content-box delta.

4. **Posterless videos → blank** (`87bde1bf`) — a `play_video` clip only shows during its play window; the rest of the time the static base-frame PNG backs that spot, but html2canvas can't draw a `<video>` and generated videos usually have no poster. Decode the video's first frame as a poster before the snapshot so that spot shows the first frame instead of blank.

5. **Subtitles + export UX** (`7101c566`, #867 item 2) — long narration was one cue → a 4–6 line block covering the slide, always burned in.
   - Pure `splitCue` (compiler layer): split on sentence → clause → hard-wrap, distribute the time window by character weight (last piece pinned to parent end, sub-1.2s slivers merged). Burned-in overlay and SRT/VTT share one split track.
   - Burn-in now **optional, default off** → clean video + sidecar SRT/VTT; caption clamped to 2 lines.
   - "Download subtitles (.srt)" path that runs only the compiler (no assets/zip/service).
   - Video export moved from a cramped dropdown to a **centered modal** (Output / Subtitles / Actions / Progress), stable across multi-minute renders. i18n across all 8 locales.

## Testing

- `vitest run tests/video-export/` — 107 passed (new `split-cue.test.ts`, `timeline-deps.test.ts`; updated collect/geometry/emit + snapshot).
- `@openmaic/renderer` package tests — 226 passed (new `BaseLatexElement.test.tsx`).
- `tsc --noEmit` clean; eslint purity boundary on `lib/video-export/**` clean; i18n key alignment passes.
- Manually driven end-to-end against a local render service (spotlight alignment, video clips, formulas, subtitle split + download all verified in the app).

## Notes

- Subtitle sync is by character-weight proportion — the TTS pipeline stores only a whole-clip duration (no word-level timestamps). Same cues drive burn-in and sidecar, so they never drift from each other. Word-level alignment is a later track.

Refs #867.


lib/video-export-app/timeline-deps.ts
```diff
@@ -181,14 +206,34 @@ export async function createVideoTimelineDeps(input: {
...
+  // generated video misses and is silently dropped from the export (compile →
+  // present:false → no `<video>` emitted, no bytes collected). Mirrors the live
+  // engine's `resolveMediaPlaceholderId`.
+  const mediaRefByElementId = new Map<string, string>();
+  for (const scene of scenes) {
+    if (scene.type !== 'slide') continue;
+    const elements = (scene.content as SlideContent)?.canvas?.elements ?? [];
+    for (const el of elements) {
+      const ref = elementMediaRef(el as { mediaRef?: unknown; src?: unknown });
+      if (ref) mediaRefByElementId.set((el as { id: string }).id, ref);
```
wyuc: This deck-wide map is last-writer-wins for repeated slide-local element IDs such as `video_001`. In a multi-slide classroom, both duration and asset lookup can resolve an earlier scene's `play_video` action to the last scene's media ref. Please scope this bridge by scene, using the `SceneCore` argument already available to `assets.media`, and add a cross-scene collision test.

lib/video-export/split-cue.ts
```diff
@@ -0,0 +1,226 @@
...
+ * a Latin/ASCII char ~0.5). ~40 units ≈ two comfortable lines of a bottom
+ * caption band at the default font size.
+ */
+const MAX_UNITS = 40;
+
+/** A piece shorter than this (ms) is merged into a neighbour so it doesn't flash. */
+const MIN_CUE_MS = 1200;
+
+/** Sentence-ending punctuation (primary split points), kept with the sentence. */
+const PRIMARY = /([。！？；…\n]|[.!?;])/;
```
wyuc: A bare ASCII period is not always a sentence boundary. This splits decimals and versions such as `3.14` and `v1.2` into separately timed cues, affecting both burned-in captions and SRT/VTT sidecars. Please keep periods between digits inside the same token and cover decimals, versions, abbreviations, and repeated ellipses in the splitter tests.

wyuc (changes requested):
Thanks for the video export fidelity work. The cross-review found two correctness issues that should be addressed before merge:

1. The element-to-media bridge is deck-wide and last-writer-wins. Repeated element IDs across slides can resolve to the wrong video and duration. The mapping should be scoped by scene.
2. Subtitle sentence splitting treats every ASCII period as a boundary, splitting decimals and versions such as `3.14` and `v1.2` into separate timed cues.

Non-blocking follow-ups: the subtitles-only path performs unused geometry and video probes, geometry measurement has weaker font-settling guarantees than snapshot rendering, and subtitle download is excluded from the dialog's concurrency guard.

CI is green, but the affected multi-scene and punctuation cases are not covered by the current tests.

wyuc:
Suggestion for the subtitle splitter follow-up: avoid growing the custom punctuation heuristic. Sentence-boundary detection and subtitle cue shaping should be separate stages.

1. Use `Intl.Segmenter(locale, { granularity: 'sentence' })` as the multilingual/CJK baseline.
2. For English and Latin-script narration, add a mature SBD implementation or a small abbreviation-aware post-merge layer. [`sbd`](https://github.com/Tessmore/sbd) already handles common abbreviations, numbers, URLs/emails, ellipses, and custom abbreviation lists.
3. Apply the existing subtitle-specific constraints only after semantic segmentation: visual-unit budget, clause/word wrapping, minimum cue duration, reading-speed limits, and exact parent-window tiling.

A local Node 24 / ICU 78.3 check shows why `Intl.Segmenter` alone is not enough: it correctly preserves `3.14`, `v1.2`, `e.g.`, `U.S.`, and CJK ellipses, but still splits `Dr. Smith` after `Dr.`. The current implementation has the same remaining edge case because it only suppresses an abbreviation boundary before a lowercase continuation.

Longer term, if the TTS provider exposes word timestamps, cue timing should be derived from word timings and pauses instead of distributing a whole narration window by character weight. That would improve synchronization more than any sentence splitter alone.

cosarah:
@wyuc Thanks for the detailed follow-up on the subtitle splitter. Addressed in f9efecc6 (formatting in 3d0e3913).

I restructured `lib/video-export/split-cue.ts` so sentence-boundary detection and cue shaping are separate stages, as you suggested:

1. **Multilingual/CJK baseline via `Intl.Segmenter`** — `splitSentences` now delegates boundary detection to `new Intl.Segmenter(undefined, { granularity: 'sentence' })` instead of the hand-rolled punctuation scan. ICU's rules already preserve `3.14`, `v1.2`, `e.g.`, `U.S.`, ellipses, and the CJK terminators `。！？…` for free, so the custom heuristic is gone rather than grown.

2. **Abbreviation-aware post-merge** — instead of adding `sbd`: `lib/video-export/**` sits behind a machine-enforced eslint purity boundary (imports restricted to `@openmaic/dsl`, `zod`, `../choreography`, and relative siblings), so a new npm dependency is rejected at lint time. `Intl.Segmenter` is a pure Node ≥20.9 builtin, so I took your "small abbreviation-aware post-merge layer" alternative: after segmentation, a segment is re-joined onto its predecessor when the predecessor ends in a known abbreviation (`ABBREVIATIONS` set — `Dr.`, `Prof.`, `Fig.`, `Mr.`, `etc.` …). This fixes the exact `Dr. Smith` case you flagged, which ICU alone still splits.

3. **Cue-shaping constraints unchanged** — the visual-unit budget, clause/word wrapping, `MIN_CUE_MS` merge, and exact parent-window tiling all run *after* semantic segmentation, untouched. Semicolons (`；;`) are no longer treated as sentence ends (they join clauses) and were moved into the clause-level splitter, so an over-budget semicolon-joined sentence still breaks there.

**Test coverage** for the previously-uncovered cases: `Dr. Smith` / `Fig. 3` / `U.S.`, the semicolon over-budget path, plus the existing decimal/version/`e.g.` cases. One intentional behavior change: `Wait... what happened?` now stays a single under-budget cue (ICU treats it as one sentence) rather than splitting on the dots — I think that reads better; test updated with a note. 22/22 pass.

On the longer-term point: agreed that deriving cue timing from TTS word timestamps + pauses would beat any character-weight distribution. Our pipeline only stores a whole-clip `AudioFileRecord.duration` today, so that's a larger change I've left out of this PR.

Could you take another look when you have a chance? 🙏

lib/video-export/split-cue.ts
```diff
@@ -0,0 +1,304 @@
...
+
+/** Split on a punctuation regex, re-attaching each delimiter to its preceding chunk. */
+function splitKeepingDelimiters(text: string, re: RegExp): string[] {
+  const parts = text.split(re);
+  const out: string[] = [];
+  for (let i = 0; i < parts.length; i += 2) {
+    const chunk = (parts[i] ?? '') + (parts[i + 1] ?? '');
+    if (chunk) out.push(chunk);
+  }
+  return out;
```
wyuc, line 136: [P2] Preserve genuine sentence boundaries after abbreviations

This unconditionally joins the next ICU segment whenever the previous segment ends in a listed abbreviation, so it also removes real sentence boundaries. For example, `splitCueText('It was made by Acme Inc. Next point.')` is grouped as one sentence, and even the explicit boundary in `Acme Inc.\nNext topic` is rejoined. This is common for entries such as `inc`, `ltd`, `corp`, `etc`, and `st`. Please preserve explicit newline boundaries and restrict the post-merge to contexts that indicate the abbreviation continues the same sentence, with regression tests for both cases.

wyuc (changes requested):
The move to `Intl.Segmenter` is a good direction, but the abbreviation post-merge still introduces a blocking subtitle-splitting regression by removing genuine sentence boundaries. Please address the inline case and add regression coverage. This review covers the latest subtitle-fix delta only.

cosarah:
Good catch — fixed in 164d5cb0. The post-merge was too broad and was deleting genuine boundaries.

**Root cause:** it rejoined the next ICU segment whenever the previous one ended in *any* listed abbreviation, so `It was made by Acme Inc. Next point.` collapsed into one cue and even `Acme Inc.\nNext topic` was rejoined.

**Fix — the merge now targets only the case ICU actually mis-splits**, a title before a capitalized proper noun (`Dr. Smith`, `Prof. Wang`), and fires only when *all* hold:
- the previous segment ends in a **continuing title** (`Mr/Mrs/Ms/Dr/Prof/Sr/Jr`) — sentence-trailing abbreviations (`Inc./Ltd./etc./vs./No.`) are dropped from the set entirely, since they legitimately end sentences;
- the previous segment does **not** end at a newline (explicit hard boundary);
- the next segment **starts with a capital** (the proper noun ICU mis-split).

The last two conditions are what make it safe: a title before a *number* (`No. 5`, `Fig. 3`) or a *lowercase* continuation (`etc. and…`) is already kept whole by ICU, so it never reaches the merge.

**Regression coverage added:**
- `It was made by Acme Inc. Next point.` → two cues ✓
- `Bring pens, etc. Then leave.` → two cues ✓
- `Acme Inc.\nNext topic here.` and `Dr.\nSmith arrived later.` → newline boundary preserved ✓
- `Bring pens, etc. and then leave.` → one cue (lowercase continuation) ✓
- retained: `Dr. Smith`, `Prof. Wang`, `Fig. 3`, `No. 5`, `U.S.`, decimals/versions/`e.g.`

26 split-cue tests pass; full `tests/video-export/` suite 123 pass. Mind taking another look? 🙏

lib/video-export/split-cue.ts
wyuc: The reported `Inc.` / `etc.` and explicit-newline regressions are fixed in `164d5cb` with regression coverage. The remaining `Sr.` / `Jr.` suffix ambiguity is low-frequency and accepted as a non-blocking follow-up for this PR.

wyuc (approved):
The reported subtitle-boundary regressions are fixed in `164d5cb` with regression coverage, and CI is green. The remaining `Sr.` / `Jr.` ambiguity is non-blocking.

PR #955, opened 2026-07-21
feat(runtime): complete the #869 learner-data cutover — quiz + playback onto RuntimeStore

wyuc:
## Summary

Completes the #869 data-cutover line. All four learner-runtime families are now dispositioned: chat (#926), PBL (#922), and **quiz** persist through the runtime layer; **playback** is ruled volatile-by-design (see the ruling on #869) with only its resume cursor persisted. Assembled on the integration trunk `integration/runtime-data-cutover` via #956 → #957 → #959.

### Quiz (from #934 / #935, integrated against post-#946/#914 main)
- Attempt lifecycles (draft → submitted → reviewed → retry rollover) persist as learner-scoped RuntimeStore records; localStorage remains only as a synchronous crash-recovery journal
- RuntimeStore is the authoritative read source; legacy state migrates with compare-and-delete safety

### Integration fixes (cross-review round 1 findings)
- `ChatRequestTemplate.storeState` retired; Pi single requests build their snapshot via the shared async RuntimeStore quiz read (Pi bypasses `runAgentLoop`'s per-iteration `getStoreState` and `/api/chat/pi` requires `storeState`)
- HTTP client / reference server / PgRuntimeStore now implement `expectedLastSeq` + `sessionTransition` + 409 `RUNTIME_APPEND_CONFLICT` round-trip (previously browser-only — server-backed deployments would accept conflicting quiz appends)

### Playback (#956 + #957 + #959)
- **Resume cursor**: device-scoped last-write-wins KV (`playback-cursor:<stageId>`), per #779/#869 amended open question 2; sessionStorage keeps same-tab priority, KV covers fresh tabs/reloads; the engine publishes a progress snapshot at every discussion-consumption path so the cursor stays fresh
- **Consumed-discussion state: volatile by decision** — it is UX gating, not learner data (a re-shown card auto-skips; joined discussions' content already lives in chat records; no replay/analytics consumer exists). The `'playback'` kind name stays reserved in the DSL
- Dead Dexie `playbackState` machinery retired: cursor half lazily migrates to KV, row deleted, legacy consumed ids dropped

## Verification
- CI green across the line: Lint/Typecheck/Unit, E2E, Storage Contract (PostgreSQL 16), Render Service
- `pnpm --filter @openmaic/storage test`: 304 passed (contract suites exercise the new append semantics on browser/HTTP/PG)
- 32 playback vitest cases; quiz suites carried from #934/#935
- Targeted live e2e (`e2e/tests/playback-resume-cutover.spec.ts`): cursor persists during playback and survives a fresh browsing context — real Chromium, passing locally and in CI
- Two Claude⇄Codex cross-review rounds + final full-diff review (1 P0 + 6 P1 total, all fixed)

Closes #934. Closes #935. Related to #869.


wyuc:
Closing for now — per-item PRs will merge into the integration trunk `integration/runtime-data-cutover` first; a unified PR to main will be opened once the data-cutover line is complete.

wyuc:
Final verification appended:

- **Full-diff final review** (Codex, adversarial): 1 P0 + 3 P1 found in the playback wiring, all fixed in #957 — the engine now publishes a progress snapshot at every discussion-consumption path (previously the fact never reached persistence), reads fold all post-merge playback sessions, interrupted legacy migrations resume without dropping the tail, and failed appends retry on later ticks. Quiz/Pi/storage decisions re-verified sound.
- **Targeted live e2e** (`e2e/tests/playback-resume-cutover.spec.ts`, real Chromium against `next dev`): seeds a deterministic stage, starts the lecture, and asserts the full chain — discussion auto-skip appends a `discussionConsumed` record to `maic-runtime`, the device cursor lands in KV, and both survive a fresh browsing context. Passing locally (21s) and in CI.
- Trunk CI green across all jobs: Lint/Typecheck/Unit, E2E, Storage Contract (PostgreSQL 16), Render Service, Analyze.

cosarah:
## Cross-review findings (Claude-side)

Reviewed at live head `9b53a0c9` (post-#959). Two P3 findings survive; both are conservative-failure (degradation, not data corruption or cross-learner leakage). Earlier playback findings were mooted by #959 removing the consumed-discussion persistence.

> ⚠️ **codex reviewer was absent** for this PR — four `codex exec review` runs (2×high, 2×medium) each died mid-exploration on the ~5.5k-line diff without emitting a verdict. So these findings have **no second-vendor corroboration**; this diff has effectively had one reviewer.

---

### P3 — `lib/chat/quiz-results-for-store-state.ts:24` — quiz context dropped on legitimate agent turns via reference-identity compare

`didActiveSceneRemainUnchanged` compares the active scene by **object identity**:

```ts
!!sceneBefore && scenesAfter.find((scene) => scene.id === currentSceneIdAfter) === sceneBefore
```

It has already confirmed `currentSceneId` is unchanged, so the extra identity check is stricter than the stated intent ("don't leak a stale scene's results into the next scene's request"). During the async RuntimeStore quiz read (≤1500ms) in `buildFreshAgentLoopStoreState` (`components/chat/use-chat-sessions.ts:181`), any Zustand update that reallocates the scene object — while `sceneId` stays the same — flips this to `false`, and `quizResults` is set to `undefined`.

**Failure scenario:** learner is still on the same graded quiz scene and sends a chat message; a scene patch / playback tick reallocates the scene object within the read window → the outgoing agent request omits the learner's graded answers + feedback → the AI reply that turn can't reference the student's actual answers. Results are safely persisted in RuntimeStore (no data loss); it's a single-turn, self-healing degradation. Suggest comparing on `sceneId` only (or `scene.updatedAt`), not object reference.

---

### P3 — `packages/@openmaic/storage/src/server/index.ts:495-531` — invalid `sessionTransition` surfaces as 500 instead of 400

The records route validates the record envelope (`validateRuntimeRecord`) and the payload-for-kind, but destructures `sessionTransition` and forwards it to `store.appendRecord` **unvalidated**:

```ts
const { expectedLastSeq, sessionTransition, ...init } = body;
// ... validateRuntimeRecord(init) / validatePayloadForKind(...) — but not sessionTransition
await store.appendRecord(init, { ...(sessionTransition === undefined ? {} : { sessionTransition }) });
```

An invalid `sessionTransition.status` is caught only by the store's generic `validateRuntimeSession` throw, which maps to `INTERNAL_ERROR (500)` rather than a classified 4xx.

**Failure scenario:** a non-conforming client posts `{ sessionTransition: { status: "bogus", updatedAt } }` → 500 instead of 400. Only reachable by a client bypassing the typed app path (the app sends `'completed'`); pg/browser still reject atomically, so no integrity risk — purely an error-classification issue. Suggest validating the `sessionTransition` shape in the route alongside the record envelope.

---

*Review-only — whether to fix, accept, or re-review is the author's call. Given #959 rewrote ~671 lines of playback that this Claude line hasn't independently re-reviewed and codex was absent, a fresh review pass on the live head may be worth it.*

wyuc:
Review response — all findings addressed in #966 (merged to the trunk):

**Both P3s fixed:**
- `didActiveSceneRemainUnchanged` now compares by scene id, not object reference — a store update reallocating the scene object during the ≤1.5s async quiz read no longer drops the learner's graded answers from that turn's request. The test asserting identity semantics was updated to pin the id-boundary semantics (same-id reallocation retains results; scene-left-deck rejects).
- The records route validates `sessionTransition` (status enum + `updatedAt` string) and returns a classified 400 instead of surfacing the store's throw as 500.

**Second-vendor gap on #959 closed:** ran a targeted adversarial Codex review over the shrink commit (five verification points). It confirmed three clean, and found two real issues — both also fixed in #966:
- **High**: an engine orphaned by a scene switch during async lecture resume could pass the idle-only recheck, be resurrected, and publish its old scene's progress over the new scene's debounced KV cursor. The resume continuation now requires identity with the installed engine, and `onProgress` drops snapshots from superseded engines.
- **Medium**: the legacy cursor migration's check-then-write could overwrite a newer cursor saved by a concurrent tab (unrecoverable after the legacy-row delete). It now rechecks KV immediately before the write.

CI green on the trunk head (Lint/Unit incl. the updated suites, E2E, Storage Contract PG16, Render, Analyze).

cosarah (dismissed):
## Approving ✅

Fresh cross-review at head `7edfef63`: **0×P1, 0×P2**. The core cutover is sound — quiz attempt lifecycle (append-only + idempotent + tail-guarded CAS), storage-backend conformance (browser/http/pg/server apply `sessionTransition` atomically and honor `expectedLastSeq`, covered by the shared contract suite), the #959 consumed-discussion removal is clean, and my earlier P3 pair (scene-id boundary + `sessionTransition` 400) is properly fixed. Approving.

Three **P3** nits remain — none blocking, worth a quick follow-up:

1. **`packages/@openmaic/storage/src/server/index.ts:504-515`** — the new `sessionTransition` guard checks `typeof updatedAt === 'string'` but not `isIsoTimestamp`, so a non-ISO `updatedAt` slips past the 400 gate and 500s deeper in `validateRuntimeSession` (inconsistent with the sibling PATCH `/status` route). Reuse `isIsoTimestamp` in the guard.
2. **`tests/runtime/stage-storage-chat-failure.test.ts:39` & `tests/runtime/stage-delete-wiring.test.ts:32`** — both still `vi.mock('@/lib/utils/playback-storage', …)`, a module deleted in this PR. Dead mocks; may break depending on Vitest's strictness about factory-mocking a non-existent path. Drop them.
3. **`lib/playback/cursor.ts:124`** — `new Date(legacy.updatedAt).toISOString()` throws on a corrupt legacy timestamp *before* the legacy-row delete, so `loadCursor` re-throws forever and resume is silently disabled for that stage. Edge case (legacy rows were numeric `Date.now()`); guard the conversion.

> Note: the codex cross-vendor reviewer was **absent** this round (repeated tooling failures), so these findings are Claude-side only — no second-vendor corroboration.

wyuc:
Approval follow-up: all three P3 nits fixed in #967 (merged) — ISO gate on `sessionTransition.updatedAt`, dead `playback-storage` mock factories removed, corrupt legacy timestamp no longer wedges cursor migration. Trunk is also updated from main (was BEHIND). CI green. Ready to merge at your convenience.

cosarah (approved):
LGTM

PR #958, opened 2026-07-21
feat(renderer): add optional playback canvas

xuyuanwei678:
## Changes
- add an environment-gated `@openmaic/renderer` playback canvas while keeping legacy `ScreenCanvas` as the default
- preserve generated image/video status UI, controlled video playback, highlights, spotlight, laser, zoom, and canvas scale synchronization
- extend renderer media slots and effects APIs, including explicit video interactivity and prepared default image content
- share generated-media resolution with slide thumbnails and bump `@openmaic/renderer` to `0.0.3`
- add regression coverage for media states, spotlight measurement, scale callbacks, and pointer interaction

## Configuration
- set `NEXT_PUBLIC_MAIC_PLAYBACK_RENDERER_ENABLED=true` to enable the renderer playback canvas
- unset or set to `false` to keep the legacy renderer

## Verification
- `pnpm --dir packages/@openmaic/renderer test` (23 files, 234 tests)
- `pnpm --dir packages/@openmaic/renderer build`
- `pnpm --dir packages/@openmaic/renderer typecheck`
- `pnpm exec vitest run tests/config/feature-flags.test.ts tests/slide-renderer/screen-canvas-renderer.test.ts tests/slide-renderer/use-resolved-slide.test.ts` (3 files, 28 tests)
- `pnpm exec tsc --noEmit`
- `git diff --check`


packages/@openmaic/renderer/src/elements/text/BaseTextElement.tsx
```diff
@@ -13,7 +13,7 @@ export interface BaseTextElementProps {
 export function BaseTextElement({ elementInfo, target }: BaseTextElementProps) {
   const { shadowStyle } = useElementShadow(elementInfo.shadow);
 
-  const vAlign = elementInfo.vAlign ?? 'top';
+  const vAlign = elementInfo.vAlign ?? 'middle';
```
wyuc: [P2] Changing an omitted `vAlign` from top to middle is not gated by the playback feature flag. Thumbnails and `slideToPng` always use this renderer, while the legacy playback path remains top-aligned. Generated slides commonly omit `vAlign`, so with the flag off the same slide is top-aligned live but centered in thumbnails and exported video. Please gate this change or update the legacy path in the same change.

components/slide-renderer/Editor/RendererScreenCanvas.tsx
```diff
@@ -0,0 +1,370 @@
...
+    laserElementId,
+    laserOptions,
+    spotlightElementId,
+    spotlightOptions,
+    zoomTarget,
+  ]);
+
+  return (
+    <SlideCanvas
+      slide={resolvedSlide}
```
wyuc, line 359: [P2] This wires slides with their own `viewportRatio` into `SlideCanvas` together with laser and zoom effects, but `packages/@openmaic/renderer/src/utils/geometry.ts` still divides vertical coordinates by `viewportSize * 0.5625`. Imported 4:3 or 16:10 slides therefore render at their correct ratio while effect Y positions remain calculated as 16:9. Please pass the slide ratio into the geometry helper so rendering and effect coordinates use the same space.

packages/@openmaic/renderer/src/elements/video/BaseVideoElement.tsx
```diff
@@ -25,9 +31,10 @@ export function BaseVideoElement({ elementInfo, renderVideo }: BaseVideoElementP
         left: `${elementInfo.left}px`,
         width: `${elementInfo.width}px`,
         height: `${elementInfo.height}px`,
+        pointerEvents: interactive ? 'auto' : 'none',
```
wyuc, line 34: [P3] Defaulting `interactive` to false now forces `pointer-events: none`, which changes behavior for existing hosts that provide `onElementClick`: the hit target is interactive and the video child previously received native control events and stopped propagation. After this patch those controls are inert unless the host discovers and opts into the new prop. Please preserve the previous default or document and version this as a compatibility break.

wyuc (changes requested):
Thanks for the renderer integration. I found two playback correctness regressions and one public API compatibility issue. Please address the P2 items before merging; the P3 can be fixed here or documented as an explicitly accepted compatibility change.

xuyuanwei678:
Hi @wyuc, I’ve carefully reviewed and addressed the three issues you pointed out:

1. Restored the default `vAlign` behavior to `top`, avoiding inconsistent behavior for slides where `vAlign` is omitted.
2. Updated the renderer geometry helpers to accept `viewportRatio`, and passed the slide’s actual ratio when calculating laser and zoom effect coordinates. The existing `0.5625` default is preserved for backward compatibility.
3. Restored video interaction as the default behavior for existing renderer consumers, while explicitly disabling it for thumbnail rendering to prevent videos from intercepting thumbnail/card clicks.

I’ve also added or updated the corresponding regression tests. The renderer test suite currently passes with 239 tests, and the repository formatting check passes as well.

When you have time, I would appreciate it if you could take another look. Please let me know if I’ve misunderstood any part of the expected behavior or if you’d prefer a different approach.

wyuc (commented):
The runtime fixes address the three requested issues. I found two remaining documentation mismatches. Both are non-runtime P3s, but since they describe public renderer and DSL behavior, please align them before merge.

packages/@openmaic/dsl/src/slides.ts
```diff
@@ -225,8 +225,8 @@ export interface PPTTextElement extends PPTBaseElement {
   /**
    * @since-merge renderer + importer
    * Vertical anchor of the text within the box, parsed from `<a:bodyPr anchor="...">`.
-   * `top` / undefined keeps the legacy top-anchored behavior. `middle` and `bottom`
-   * vertically center / bottom-align the content inside the box.
+   * `middle` / undefined vertically centers the content. `top` and `bottom`
```
wyuc: [P3] This comment still says an omitted `vAlign` centers text, but `BaseTextElement` and its regression test now default it to `top`. Please restore the documented default to `top` so the public DSL contract matches the implementation.

packages/@openmaic/renderer/README.md
```diff
@@ -77,11 +77,15 @@ The main read-only entry. Reads everything from props; zero global state.
...
   slide?: Slide;                       // required unless via <SlideRendererProvider>
   scale?: number;                      // omit = auto-fit container
+  canvasPercentage?: number;           // percent of the parent used by auto-fit
+  onScaleChange?: (scale) => void;     // computed auto-fit scale
   background?: SlideBackground;        // overrides slide.background
   effects?: SlideEffects;              // laser / spotlight / highlight / zoom, all default off
-  renderImage?: (el, src) => ReactNode;
+  renderImage?: (el, src, defaultContent) => ReactNode;
   renderVideo?: (el) => ReactNode;
+  videoInteractive?: boolean;           // opt in to video controls/custom video UI
```
wyuc: [P3] `videoInteractive` is now enabled by default to preserve compatibility, so `opt in` describes the opposite behavior. Please document that callers can set it to `false` to disable pointer interaction.

xuyuanwei678: @wyuc Fixed in 0bd6165. The renderer README now documents that `videoInteractive` defaults to true and callers can set it to false to disable video pointer interaction.

packages/@openmaic/dsl/src/slides.ts
xuyuanwei678: @wyuc Fixed in 0bd6165. The DSL `PPTTextElement.vAlign` comment now documents `top` / undefined as the legacy top-anchored behavior, matching `BaseTextElement` and its regression test.

wyuc (approved):

PR #965, opened 2026-07-21
feat(document): complete the document-persistence cutover — stage/scene/outline onto DocumentStore

wyuc:
## Summary

Completes the document-layer half of the persistence standardization: the app's stage/scene/outline persistence moves from direct Dexie access onto `@openmaic/storage`'s DocumentStore contract, eliminating the largest remaining downstream coupling surface (openmaic-live's forked PG stages/scenes + `/api/stages` adapter can now target one upstream contract).

Stacked on the runtime-data-cutover trunk (#955); will retarget `main` once #955 merges.

### PR1 — Shape contracts (#961)
- `TStage` generic through MaicDocument/DocumentStore/StageRow; injectable `validateStage`; version-guarded `putStage`
- `lib/document-store/` app seam: four-kind scene validator, device-state rejection, pure legacy canonicalizers

### PR2 — Migration + primary cutover (#962)
- Per-stage Web-Lock lazy migration: destination authoritative, coherent snapshot, save→verify→marker, legacy retained read-only one release
- stage-storage save/load/delete/list/rename through the seam; outline single authority (`generationComplete` atomic with final scenes); `currentSceneId` → device KV
- No-lock classification (new/destination-backed writable, legacy-only read-only); per-stage → global lock order

### PR3 — Secondary consumers (#963)
- ZIP import/export, backup (versioned aggregates incl. outlines, old format accepted), clear-cache, stats; import `whiteboards` alias bug fixed
- e2e seeders target `maic-documents`; classroom-interaction exercises live lazy migration

### Final hardening (#964)
- Backup rollback pre-images; reconciled migration markers; clearDatabase race exclusion via shared-epoch enrollment; ghost-card filter

## Verification
- CI green on every PR + trunk (Lint/Typecheck/Unit incl. the chat-cutover acceptance suite, E2E, Storage Contract PG16, Render)
- 149 seam/migration/consumer tests; storage package 308+; live playback e2e green on the reseeded store
- Recon-first workflow: read-only design report → 3 implementation PRs → adversarial final review (3 P1 + 2 P2, all fixed with regression tests)


lib/utils/database.ts
```diff
@@ -540,12 +562,21 @@ function toChatSessionRecord(stageId: string, session: ChatSession): ChatSession
...
   playbackState: PlaybackStateRecord[];
 }> {
-  const stages = await db.stages.toArray();
+  const { accessDocument, getDocumentStore, getLegacyDocumentStore } =
+    await import('@/lib/document-store');
+  const documentStore = getDocumentStore();
+  // Backups must not strand courses that have not yet been opened since cutover.
+  // Route them through the normal lazy-migration seam before enumerating aggregates.
+  const legacyStages = await getLegacyDocumentStore().listStages();
+  await Promise.all(legacyStages.map((stage) => accessDocument(stage.id)));
```
cosarah, line 584: **[P1] Legacy-only courses are silently omitted from backups when Web Locks are unavailable.**

This line runs every legacy stage through `accessDocument()` purely for its migration side-effect, then the backup is assembled from `documentStore.listDocuments()` (line 576). That relies on `accessDocument()` writing the legacy snapshot into the store.

But in `lib/document-store/migration.ts:265-278`, the `DocumentLockUnavailableError` fallback path (taken whenever `navigator.locks` is unavailable — non-secure contexts, some webviews, older Safari) returns a read-only in-memory canonical snapshot for a legacy-only document and **never calls `store.saveDocument`**. `mutateDocument` likewise refuses to migrate legacy-only docs without locks (`migration.ts:305`).

So in a no-locks environment there is no path that persists these stages into the store: `listDocuments()` won't include them, the backup assembly has no legacy-table fallback, and they drop out of the export — even though `loadStageData` still surfaces them to the user via the read-only path. Net result: courses visible in the app are silently missing from the backup, causing data loss on restore/device migration.

Suggested fix: in the export path (or in `accessDocument`'s no-lock branch), persist the canonicalized legacy snapshot into the store before enumerating, or fall back to reading legacy tables directly when locks are unavailable.

lib/utils/database.ts
```diff
@@ -513,11 +515,31 @@ export async function clearDatabase(runtimeStore?: RuntimeStore): Promise<void>
   // must fail loud: reporting success while runtime data remains is misleading.
   await withRuntimeStorageExclusiveLock(async () => {
     await (runtimeStore ?? getRuntimeStore()).deleteAllRuntime();
+    await deleteAllDocuments();
```
cosarah, line 521: **[P1] A concurrent `saveStageData` can repopulate the store after `clearDatabase()` reports success.**

`clearDatabase` takes only the global **exclusive** runtime lock. `saveStageData` (`lib/utils/stage-storage.ts:65-116`, via `mutateDocument`) first takes the **per-stage document lock** `openmaic:document:<id>`, and only then acquires the **shared** runtime lock around the actual `store.saveDocument`.

The shared and exclusive locks share the same lock name, so `saveDocument` and `deleteAllDocuments` never run concurrently — but their ordering is not constrained, and `clearDatabase` holds **no per-stage lock**. That admits this interleaving:

1. `saveStageData` holds the per-stage lock; `migrateLocked` acquires+releases the shared lock (`migration.ts:220-255`), then returns.
2. `work()` goes to re-acquire the shared lock (`stage-storage.ts:72`) but parks behind `clearDatabase`'s pending exclusive request.
3. `clearDatabase` runs `deleteAllDocuments()` + `db.delete()`, releases, and logs `"Database cleared"`.
4. The parked `work()` resumes, acquires the shared lock, and `saveDocument`s the stage back into the just-cleared store.

This violates this function's own contract ("must fail loud: reporting success while runtime data remains is misleading", lines 513-515): the wipe reports success yet a document survives. The per-stage → global lock order documented in `migration.ts:217-219` only protects the migration commit, not this second independent save write.

Suggested fix: have `clearDatabase` also exclude per-stage document mutations (e.g. acquire the per-stage document locks, or gate saves on an epoch/generation that `clearDatabase` bumps so a save that started before the clear is rejected on resume).

cosarah (commented):
Two P1 correctness findings from an independent cross-vendor review (Claude + codex) of the document-persistence cutover. Both concern data durability/consistency on paths this PR reworks. Inline comments below.

lib/utils/database.ts
wyuc: Fixed in #968: backup assembly now appends read-only canonicalized aggregates for legacy-only stages the no-lock path could not persist — the migration discipline (no persistence without cross-realm exclusion) is unchanged; export reads legacy directly. Regression test covers the lockManager-unavailable export.

lib/utils/database.ts
wyuc: Fixed in #968 with your epoch suggestion: a device-scoped storage generation is bumped inside clearDatabase's exclusive section; mutateDocument records it at entry and every guarded write (including the migration commit) re-reads it inside the shared epoch immediately before writing, failing loud on mismatch. Regression test simulates exactly the parked-save interleaving.

lib/utils/database.ts
```diff
@@ -557,7 +617,7 @@ export async function exportDatabase(chatOptions: ChatStorageOptions = {}): Prom
   const { loadChatSessions } = await import('./chat-storage');
   const runtimeChats = (
     await Promise.all(
-      stages.map(async (stage) =>
+      storedDocuments.map(async ({ stage }) =>
```
cosarah: **[P1] Legacy-only courses lose their RuntimeStore chat history in the backup.**

The fix now appends legacy-only *documents* to the backup (`legacyOnlyDocuments`), but `runtimeChats` still iterates only `storedDocuments`. Chat sessions live on the learner RuntimeStore, which is a **separate seam** from the document store — chat migration and document migration are independent cutovers.

So a legacy-only course (document never migrated — the guaranteed state in no-Web-Locks environments, the exact scenario the sibling backup fix targets) whose chats already live in RuntimeStore will have its document exported but its chats silently dropped: `runtimeChats` never calls `loadChatSessions` for it, and the `legacyChats` path only covers the Dexie legacy tables (`db.chatSessions` + `db.chatRestoreStaging`), not RuntimeStore.

This is the same no-locks data-loss class as the original backup-omission P1, just for chats instead of the course itself.

Suggested fix: build `runtimeChats` from `documents` (stored + legacy-only) rather than `storedDocuments`, or add a `loadChatSessions` pass over `legacyOnlyDocuments`.

lib/document-store/migration.ts
```diff
@@ -0,0 +1,359 @@
...
+
+function generationGuardedStore(
+  stageId: string,
+  expectedGeneration: number,
+  deps: DocumentMigrationDeps,
+): DocumentStore<AppScene, AppStage> {
+  const store = resolveStore(deps);
+  return new Proxy(store, {
+    get(target, property) {
+      if (property === 'saveDocument') {
```
cosarah: **[P3] The generation fence only guards `saveDocument`.**

`generationGuardedStore` intercepts `saveDocument` but forwards every other method bound to the real store, so `putStage` / `putScene` / `deleteScene` / `deleteDocument` bypass the storage-generation check. This isn't currently exploitable — after a clear, `putStage`/`putScene` throw "cannot put into missing document" and deletes are idempotent no-ops, so there's no resurrection vector today. Also note `renameStage` mutates via `putStage` with neither the fence nor a shared lock.

Raising it as latent fragility: any future mutation callback that persists through one of the unguarded methods would silently escape the fence. Worth either guarding all mutating methods or documenting that only `saveDocument` is fenced.

cosarah (commented):
Follow-up cross-vendor review (Claude + codex) of the P1 fixes in #968. Both original P1s (no-lock backup omission, clear-then-save resurrection) are verified fixed. Two findings on the fix itself below: one P1 introduced by the wider backup coverage, and one P3 fragility note on the generation fence.

lib/utils/database.ts
wyuc: Fixed in #972: the runtime-chat enumeration now iterates every exported document (stored + legacy-only). The read injects an empty legacy chat store — legacy rows are already collected by direct table reads and deduped — which also keeps the export functional without Web Locks. Regression test: legacy-only stage with a RuntimeStore chat, exported from a no-locks environment, asserts the chat is present.

lib/document-store/migration.ts
wyuc: Fixed in #972: the fence now guards every mutating store method (saveDocument/putStage/putScene/deleteScene/deleteDocument). The exclusive-epoch delete cascades (deleteStageData, deleteStageWithRelatedData) pass storageSharedLockHeld so the widened guard cannot self-deadlock against their own exclusive hold — that interaction is what the fence's earlier saveDocument-only scope was accidentally avoiding.

cosarah (approved):
LGTM

PR #979, opened 2026-07-23
feat(storage): server-backed documents — HTTP contract, Postgres backend, one reference server for both stores

wyuc:
## Summary

Completes 1B of the persistence standardization: the DocumentStore contract (#965) gains its server side, mirroring the runtime line's #939 delivery. A deployment can now serve **all persisted app data** (learner runtime + documents) from PostgreSQL behind one reference server and two documented HTTP contracts, with zero app-code changes (sealed bootstrap injection on both seams).

### PR1 — HTTP contract + client + server routes (#975)
- `docs/document-http-contract.md`; `HttpDocumentStore` with browser-parity validation and migrate-on-read; document routes on the shared reference server (one server, both stores). Documents are author assets — authorization stays a deployment concern via injected hooks (any-authenticated default)

### PR2 — PgDocumentStore (#976)
- Full contract over `document_stages` / `document_scenes` / `document_outlines` (FK cascades); single-transaction aggregate replace under the stage row's FOR UPDATE lock; version-gated incremental writes; FOR SHARE consistent reads; **contract suite green on real PostgreSQL 16 in CI**

### PR3 — app bootstrap seam (#977)
- `configureDocumentStorage` mirrors the runtime seam: sealed after first resolution, factory receives the app validators, reset-for-tests. README roadmap: document HTTP/PG items done

### Final hardening (#978, adversarial review: 0 P0/P1, 3 P2 + 2 P3, all fixed)
- Legacy unversioned writes → 400 (was 500); fail-loud JSON-safety on reads (NOT_JSON_SAFE); bounded request bodies (413, default 32 MiB, configurable everywhere); putStage version guards + corrupt-stamp tolerance promoted into the shared contract suite for all four backends; structured error details retained

## Verification
- Storage package: 395 tests green; document contract suite passes on browser, HTTP, PGlite, and real PostgreSQL 16 (CI job)
- Every PR + trunk CI green across all five jobs
- Recon → 3 implementation PRs → adversarial final review, each cross-reviewed at implementation time

## Downstream
openmaic-live can retire its forked PG stages/scenes + `/api/stages` adapter by pointing `configureDocumentStorage` at `HttpDocumentStore` against these contracts — same integration shape it already uses for runtime (#1051).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main, feat/maic-editor-v0, feat/maic-editor-v1, runtime-server-backend, integration/runtime-data-cutover, integration/document-cutover]
+    branches: [main, feat/maic-editor-v0, feat/maic-editor-v1, runtime-server-backend, integration/document-server-backend]
 
 concurrency:
   group: ci-${{ github.ref }}
--- a/.github/workflows/storage-pg-contract.yml
+++ b/.github/workflows/storage-pg-contract.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main, runtime-server-backend, feat/maic-editor-v0, feat/maic-editor-v1]
+    branches: [main, runtime-server-backend, feat/maic-editor-v0, feat/maic-editor-v1, integration/document-server-backend]
 
 concurrency:
   group: storage-pg-contract-${{ github.ref }}
--- a/lib/document-store/config.ts
+++ b/lib/document-store/config.ts
@@ -1,0 +1,103 @@
+import type { DocumentStore, SceneValidator, StageValidator } from '@openmaic/storage';
+
+import type { AppScene } from '@/lib/types/stage';
+
+import type { AppStage } from './persistence-types';
+import { validateAppScene, validateAppStage } from './validators';
+
+export interface DocumentStorageValidators {
+  validateScene: SceneValidator;
+  validateStage: StageValidator;
+}
+
+export type DocumentStoreFactory = (
+  validators: DocumentStorageValidators,
+) => DocumentStore<AppScene, AppStage>;
+
+export interface DocumentStorageOptions {
+  /**
+   * A DocumentStore instance, or a factory evaluated lazily until it first
+   * succeeds. Factories receive the app validators and must inject them into
+   * server-backed stores so every backend preserves the app write boundary.
+   */
+  store?: DocumentStore<AppScene, AppStage> | DocumentStoreFactory;
+}
+
+let options: DocumentStorageOptions | undefined;
+let resolutionStarted = false;
+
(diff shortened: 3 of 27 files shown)


cosarah:
## Code review — Standards axis (3 findings)

All three are judgement calls, not hard violations. No documented-standard breaches; no correctness bugs found.

### 1. (Most substantive) HTTP status derived by regex-parsing error message prose

`src/server/document.ts` `classifyStoreError` (~236-271) recovers the HTTP status by pattern-matching the *text* of a plain `Error`:

```ts
const versionMatch = /at DSL version (undefined|"(?:[^"\\]|\\.)*")/.exec(message);
```

The established runtime sibling does the opposite: `PgRuntimeStore` throws a typed `RuntimeAppendConflictError` and the server matches `error instanceof RuntimeAppendConflictError` (`server/index.ts:273,341`). Deriving control flow from prose means any wording change in `pg.ts` silently breaks the status mapping — with no compile or test failure to catch it. This is Primitive Obsession; a typed error class (mirroring `RuntimeAppendConflictError`) would match the neighbours. Worth fixing.

### 2. Duplicated code — acceptable, matches in-package convention

`migrateDocument`, `assertStorableScene`, and `assertValid` are repeated near-verbatim across `document/http.ts`, `document/pg.ts`, and `server/document.ts`. This is a Duplicated Code smell in the abstract, **but it is overridden by established in-package precedent**: the runtime siblings (`runtime/http.ts` / `runtime/pg.ts`) already duplicate the same small helpers rather than sharing them. The new code is consistent with its neighbours — noted, not required.

### 3. Inconsistent validator field naming (minor)

The two backends name the same concept differently: `HttpDocumentStore` uses `sceneValidator` / `stageValidator` (`http.ts:150-151`), while `PgDocumentStore` uses `validateScene` / `validateStage`. `http.ts` matches the runtime sibling convention, so `pg.ts` is the outlier. Minor.

wyuc:
Standards review response — all three addressed in #980 (merged to the trunk):

1. **Fixed as suggested**: `DocumentVersionError` (`kind: 'future' | 'not-current'`, `storedVersion`) and `DocumentNotFoundError` are now thrown by the browser/pg stores and matched by `instanceof` in the server — mirroring `RuntimeAppendConflictError`. The prose regex survives only *after* the typed checks, explicitly documented as a fallback for third-party `DocumentStore` implementations; in-package stores can no longer be broken by a wording change. The HTTP client also rematerializes 409s as `DocumentVersionError` for cross-backend symmetry.
2. Left as-is per your own conclusion (matches the runtime siblings' precedent).
3. **Clarification**: the public options were already uniform — all three backends take `validateScene`/`validateStage`; what you saw in `http.ts:150-151` were its *private* fields. Renamed them anyway so the file can't be misread as a public divergence.

398 package tests green incl. new instanceof assertions on every classified path; message texts unchanged. CI green across all five jobs at the new head.

cosarah (approved):
LGTM

PR #982, opened 2026-07-23
feat(persistence): one-command server-backed stack (compose profile + embedded API + docs)

wyuc:
DX follow-up to #979: make server-backed persistence a three-command experience.

- **Two containers only**: the persistence server is embedded in the app at `/api/persistence` (reusing the package's composed handlers + pg transaction helper); no standalone service. `DATABASE_URL` unset = browser-only behavior unchanged.
- **compose profile `server-persistence`**: postgres:16 (volume + healthcheck) + build-arg wiring for the `NEXT_PUBLIC_PERSISTENCE` toggle, following the video-export profile precedent.
- **Client bootstrap**: same-origin seam configuration via the established module-level init point; anonymous self-host works through the device learner key; existing browser data lazy-migrates to the server per-course.
- **Dev auth in one overridable module** (`lib/persistence/server-auth.ts`), documented as development-grade with production guidance.
- README quickstart + reference-server doc note; 27 unit tests.
- **Verified live**: full stack built and run locally — document PUT/GET and runtime session-create land in Postgres (psql-confirmed), wrong token → 401.


wyuc:
Internal CR loop complete before requesting review:

- Adversarial review round found 3 P1 + 3 P2 + 2 P3 — all fixed in f1e741b: structural bootstrap ordering (seam-entry side-effect import — no client-entry luck; configure is all-or-nothing), retryable server init (a transient first-connect failure no longer poisons the route until restart), listStages rethrows so a backend outage surfaces as an error instead of an empty dashboard, token-boundary wording states plainly that any visitor can extract the NEXT_PUBLIC token and impersonate any learner (localhost/trusted-network only), corrected postgres password-rotation docs, timing-safe token compare, HMR-proof pool singleton.
- Re-verified live on the real container stack including outage injection: `put=204 → stop postgres → 500 → start postgres → 200` with no app restart.
- 69 unit tests green; five CI jobs green.

cosarah:
## Review findings

Overall this is clean — it reuses the `@openmaic/storage/server` handler contract, keeps `DATABASE_URL`-unset behavior identical, and the docs are honest about the auth being dev-grade. A few issues, most-severe first.

### 🔴 High: a failed handler init poisons the cache → persistence returns 500 permanently until restart

`app/api/persistence/[...path]/route.ts:54`

```ts
return (handlerPromise ??= createPersistenceHandler(connectionString));
```

`createPersistenceHandler` awaits `ensureSchema()` / `ensureDocumentSchema()`, which reject if Postgres isn't ready yet. The problem:

1. `??=` only reassigns when the value is nullish. A **rejected promise is still a non-null object**, so it's never retried.
2. The moment the first request hits "DB not up yet", `handlerPromise` is pinned to a rejected state, and every subsequent request falls into `handle`'s catch and returns `PERSISTENCE_INIT_FAILED` (500) — **until the process restarts**.
3. In `docker-compose.yml` the `openmaic` service has **no `depends_on` + `condition: service_healthy`** on `postgres` (postgres defines a healthcheck, but nothing consumes it), so the app can absolutely receive its first request before Postgres is ready.
4. `restart: unless-stopped` doesn't help: the Next process doesn't crash, it just serves 500s, so nothing triggers a restart.

Notably, this same PR gets it right in `lib/persistence/bootstrap.ts:12` — it resets `learnerKeyPromise = undefined` in the catch so it can retry. The route is missing the symmetric handling.

**Suggestion**: reset the cache on rejection (`handlerPromise = undefined` after it fails) and add `depends_on: { postgres: { condition: service_healthy } }` to the app service.

### 🟡 Medium: `NEXT_PUBLIC_PERSISTENCE_TOKEN` is compiled into the browser bundle, so the "shared secret" is public

`NEXT_PUBLIC_*` values are baked into the client bundle (the Dockerfile wires them in exactly this way), so this "shared secret" is **plaintext-visible to anyone who loads the page**. Combined with:
- `server-auth.ts` trusting a fully client-asserted `x-learner-key` → anyone can impersonate any learner;
- `authorizeDocuments: async () => true` in the route → any authenticated request can read/write **all** documents, not partitioned by learner.

So this auth stops no real attacker. The README and `server-auth.ts` comments **do** flag this as development-grade and "must be replaced before production", so it's not a hidden defect. But since putting the token in the public bundle drops its protective value to zero, I'd suggest strengthening the wording from "dev-grade shared secret" to something like "fully public to the browser, provides no confidentiality" so nobody assumes it offers even token resistance.

### 🟢 Low: the trickiest part — the Fetch↔Node adapter — has no happy-path test

The hand-written `nodeRequest` / `runNodeHandler` in `route.ts` (faking a Web `Request` as an `IncomingMessage`, bridging `writeHead`/`end`/`destroy` back to a `Response`) is the most bug-prone code in the PR, but the two new tests in `route.test.ts` only cover the `DATABASE_URL`-unset (404) and missing-dev-token (503) **early-return error paths** — both of which return *before* reaching the adapter. The adapter itself (status/header/body round-trip, 204 with no body, multi-value headers) has no round-trip coverage.

I checked the downstream handlers (`server/index.ts`, `document.ts`) and they only use `writeHead`/`end`/`destroy`, all of which the adapter implements, so it should be functionally correct today. But this logic deserves an integration test that runs a GET/PUT through it against a fake store — otherwise a future handler change (e.g. starting to stream via `res.write()`) would silently go uncovered.

wyuc:
Review response:

- **High (init poisoning)**: already fixed before your review landed — head `f1e741b` clears the cached promise on rejection (route.ts:79-81, the exact symmetry with bootstrap.ts you pointed at) and the retry has a regression test plus a live outage-injection verification on the real container stack (`PUT 204 → stop postgres → 500 → start postgres → 200`, no restart). On `depends_on`: compose can't scope it to a profile on the shared app service — an unconditional `condition: service_healthy` would break the default browser-only `docker compose up` (no postgres in that profile). The retry semantics are the mechanism that covers the cold-start race; happy to add a compose override file instead if you'd prefer structure over docs.
- **Medium (public token)**: wording upgraded to your framing in README + server-auth.ts — 'compiled into the public bundle, provides no confidentiality and no user isolation; its only purpose is keeping unrelated network scanners off a trusted-network endpoint'.
- **Low (adapter coverage)**: added a round-trip integration test through the real adapter against a fake handler: PUT body streaming, 201 + multi-value headers, 204 empty body, and encoded path segments arriving un-decoded.

10 persistence tests green; CI re-running at the new head.

cosarah (approved):
LGTM

PR #983, opened 2026-07-23
feat(document): dirty-set incremental saves + operation-level flush

wyuc:
Consumption-layer follow-up to #965/#979: the DocumentStore contract has had version-gated incremental APIs since #961 with zero consumers — every mutation still triggered a debounced full-aggregate rewrite.

- Dirty categories per mutation (scene:<id> / structure / stage / outline / currentScene / chats); scene edits → `putScene`, stage metadata → `putStage`; structure/outline keep the full path; currentScene & chats no longer touch the document
- Revisioned dirt (mutations landing mid-write never lost); failed writes retain dirt and retry
- `flushStageSave()`: single-flight drain; agent apply paths await it — each AI edit is durably persisted before its tool result returns
- visibilitychange/beforeunload best-effort immediate flush
- `PendingChange` stays change-unit-neutral (one seam to swap in an operation log for future collaboration)
- 540 tests green across store/document-store/agent/persistence/chat-cutover suites

Draft for the internal CR loop; will ready + request review after.


wyuc:
Internal CR loop complete: adversarial review found 4 High + 2 Medium — all fixed in the second commit (structural mutations carry their cursor + membership-validated resolution; Stage API guard diffs pre/post state so whiteboard/element/canvas/action-engine writes can't bypass dirty tracking, with an inventory test; flushStageSave is a bounded true drain; setStage departing-flush instead of dirt discard; homogeneous-only incremental fast path — mixed batches take the atomic full save; chat failures keep their dirt). 582 tests green, five CI jobs green.

cosarah:
## AI-assisted code review (Claude)

Reviewed `feat/incremental-doc-persistence` against `origin/main`. Two correctness findings around the revision-based dirt-clearing, one behavior mismatch, plus minor cleanups.

### 🔴 1. Agent durability guarantee can break under a concurrent failing write (spec claim 3)

> "agent apply paths await it — each AI edit is durably persisted **before its tool result returns**"

In `lib/store/stage.ts` (`flushStageSave` / `startFlushRound`, ~685–775):

- `startFlushRound()` returns the **existing** `flushInFlight` promise if a round is already running. That round captured its snapshot *before* the agent's `updateScene`, so it does not include the agent's edit.
- `flushStageSave()` has `if (failedKeys.size > 0) return;`.

When a debounced flush overlaps an agent apply **and** that in-flight round contains a failing write (e.g. a chat-save failure), `flushStageSave`'s first iteration awaits the stale shared promise, gets `failedKeys={chats}`, and returns early — without ever running a round that captures the agent's edit. The tool result then returns claiming durability while the scene edit is still pending, only persisted on the next debounce. The guarantee is stated unconditionally, so this is a real (if narrow) durability gap.

### 🔴 2. Full-save path swallows chat-save failure but still clears the dirt (spec claim 2 — lost update)

> "failed writes retain dirt and retry"

`saveStageData` (`lib/utils/stage-storage.ts:128–160`) calls `saveStageChats(...)` but **ignores its return value**; `saveStageChats` returns `false` on failure after only a `log.warn` (:120–126). So `saveStageData` resolves, `saveToStorage` returns `true` (`stage.ts:520–565`), and the `chats` dirt entry is deleted by revision — even though chats never persisted. The chat update is lost with no retry.

The **incremental** path is correct here: `stage-storage.ts:264` pushes `{ kind: 'chats' }` onto `failedChanges` on failure. So the two paths are inconsistent — the full-save path violates retain-and-retry.

### 🟠 3. Departing-stage flush is best-effort, not retain-and-retry (spec claim 2)

`setStage` (`stage.ts:279–317`) snapshots `departingDirty`, fires `persistDirtySnapshot` fire-and-forget, then `resetPendingChanges(stage.id)` replaces the map for the incoming stage. If the departing write fails it's only `log.warn`'d (:296–302) — the dirt is gone, no retry. Reasonable as best-effort on navigation, but it contradicts the unconditional retain-and-retry wording. Please either add a retry path or clarify in the description that departing-stage flush is best-effort.

### 🟡 4. Minor / cleanup

- **Duplicated code — `lib/utils/stage-storage.ts`.** The stage-stamping object (`{ ...data.stage, id, name: … || 'Untitled Stage', createdAt, updatedAt }`) is written verbatim in both `documentSnapshot` and the `has('stage')` branch of `saveStageDataIncremental`. Extract a `stampStage(stageId, stage, now)` helper so the two paths can't drift. Note the per-scene stamping has **already** diverged — the `putScene` branch omits the `order ?? index` that `documentSnapshot` applies; shared extraction would prevent exactly this class of bug.
- **`setGenerationComplete`** both `markPendingChanges({ kind: 'outline' })` and `void saveToStorage()`; the marked-dirty outline entry is immediately superseded by the aggregate save, so the mark is dead work. Pick one.
- Scope creep (harmless): `loadStageData` now sanitizes the stored cursor against existing scene ids and falls back to `scenes[0]` (`stage-storage.ts:298–305`) — a load-side change not in the save-side spec.

### ✅ Confirmed correct
- Incremental revision bookkeeping (`stage.ts:704–709`) — post-write deletion gated on matching `revision`, so a mutation landing mid-write is retained and retried.
- `DocumentVersionError` `not-current` → full-save fallback (`stage-storage.ts:248–252`).
- Conventional Commits, i18n (only user-facing string is pre-existing), focused scope — no hard standards violations.

**Suggested pre-merge:** fix #1 and #2 (both are real lost-update / durability gaps in the revision-clear logic); clarify or fix #3; the `stampStage` extraction in #4 also removes the per-scene `order` divergence.

wyuc:
Review response — all findings addressed in 8af0a18:

1. **Red (drain gap)**: `flushStageSave`'s early-return now fires only when the failing round's snapshot covered the caller's dirt — an agent edit landing beside an in-flight failing round is durably written before the promise resolves. Regression test simulates exactly the stale-round-failure interleaving.
2. **Red (full-path chat dirt)**: the full-save path reports chat failures through the same failedChanges mechanism as the incremental path — chats dirt survives and reschedules; the tolerated-split outer semantics are unchanged. Regression test included.
3. **Orange (departing flush)**: one bounded retry added; the doc-comment now states the best-effort semantics honestly (navigation is not a durability barrier).
4. **Cleanups**: `stampStage` + shared scene stamping extracted — this also fixed the real divergence you spotted (incremental putScene was dropping the `order ?? index` normalization); dead outline mark in setGenerationComplete removed; load-side cursor sanitization now carries a deliberate defense-in-depth comment.

551 tests green; CI re-running at the new head.

cosarah (approved):
LGTM

PR #984, opened 2026-07-23
fix(storage): bind default fetch to globalThis in both HTTP store clients

wyuc:
Found by a downstream end-to-end test: with server persistence enabled, the page runtime never reached PostgreSQL. The bootstrap was configuring the HTTP stores correctly — but the very first request threw `TypeError: Illegal invocation`, because both clients stored `globalThis.fetch` as an instance field and invoked it as `this.fetchImpl(...)`. Browsers require `this === globalThis` for fetch; node's undici does not — which is exactly why every node-side contract suite stayed green while every real browser failed on request one.

- `(options.fetch ?? globalThis.fetch)?.bind(globalThis)` in HttpDocumentStore + HttpRuntimeStore
- Strict-receiver regression test (Proxy enforcing the browser's receiver rule) guarding the gap node cannot catch natively
- Verified in a real browser against a live stack: homepage now issues `GET /api/persistence/documents`, zero console errors, PG schema auto-provisioned

### DIFF
--- a/packages/@openmaic/storage/src/document/http.ts
+++ b/packages/@openmaic/storage/src/document/http.ts
@@ -130,7 +130,11 @@ export class HttpDocumentStore<
     if (options.baseUrl === '') {
       throw new Error('@openmaic/storage: HttpDocumentStore baseUrl must be non-empty');
     }
-    const fetchImpl = options.fetch ?? globalThis.fetch;
+    // Bind explicitly: browsers require fetch to be invoked with `this === globalThis`
+    // (calling a stored reference as `this.fetchImpl(...)` throws "Illegal
+    // invocation"), while node's undici does not care — which is exactly why
+    // node-only test suites cannot catch the unbound form.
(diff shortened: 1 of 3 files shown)


cosarah (commented):
One issue remains:

**[P2] Validate `fetch` before binding it** (`packages/@openmaic/storage/src/document/http.ts:137`, `packages/@openmaic/storage/src/runtime/http.ts:154`)

If a caller passes a non-function `options.fetch` such as `{}`, `.bind(globalThis)` throws a native `TypeError` before the existing validation runs. Validate the selected value first, then bind it, so both constructors preserve the documented `requires a fetch implementation` error.

wyuc:
P2 fixed in d6d6c63: both constructors validate the selected fetch implementation before binding, so a non-function option surfaces the documented 'requires a fetch implementation' error instead of a native bind TypeError. Regression test added; 401 package tests green.

cosarah (approved):
LGTM

PR #992, opened 2026-07-26
fix(storage): omit undefined object fields at persistence boundaries

### LINKED ISSUE #991: PG persistence rejects PBL v2 state and events containing optional undefined fields
## Summary

When OpenMAIC runs with the PostgreSQL-backed HTTP document/runtime stores, normal PBL v2 state transitions can fail to persist because app-owned payloads contain optional object properties whose value is `undefined`.

The storage package correctly enforces a strict plain-JSON contract, but the PBL/document write paths currently pass these values through without canonicalizing them first.

## Observed failures

Document persistence can fail while saving a scene:

```text
@openmaic/storage: scene "<scene-id>" is not a plain JSON value at
'/content/projectV2/milestones/0/microtasks/0/internalAssessment':
undefined member (dropped by JSON)
```

Runtime persistence can fail while appending a PBL event:

```text
@openmaic/storage: runtime record "<record-id>" payload is not a plain JSON value at
'/event/actorRoleId': undefined member (dropped by JSON)
```

Both surface as a failed stage flush, leaving changes pending in the client.

## Reproduction

1. Start OpenMAIC with the PostgreSQL-backed document and runtime HTTP stores.
2. Open or create a PBL v2 project.
3. Enter a microtask and trigger normal progress/state initialization.
4. Send a learner message or otherwise emit a runtime event.
5. Observe `putScene()` or `appendRecord()` rejecting the payload because an optional field is present with value `undefined`.

Observed with Next.js 16.1.2 using the Webpack dev server and PostgreSQL storage.

## Expected behavior

Normal application state and PBL runtime events should satisfy the storage contract before reaching `DocumentStore.putScene()` / `putStage()` / `saveDocument()` and `RuntimeStore.appendRecord()`.

Optional object members that are absent should be omitted rather than represented as `undefined`. The strict storage validation should remain in place so genuinely non-JSON values such as `Date`, `Map`, or class instances still fail loudly.

## Suggested direction

- Canonicalize app-owned document values at the document-store boundary by recursively omitting `undefined` object members.
- Build PBL runtime record payloads through an equivalent plain-JSON normalization step.
- Keep arrays and non-plain objects subject to the existing strict validation rather than broadly using a lossy `JSON.stringify` round trip.
- Add regression coverage for:
  - `internalAssessment: undefined` in nested PBL project state.
  - `actorRoleId: undefined` in a runtime event.
  - preservation/rejection behavior for other JSON and non-JSON values.

## Impact

This affects both canonical course-document persistence and learner runtime history. In a PostgreSQL deployment, users may see successful in-memory UI updates while the corresponding scene/progress/message data is not durably stored.

wyuc:
## Summary

- Omit `undefined` object properties before document and PBL runtime persistence writes.
- Preserve array structure and strict rejection of non-JSON values such as dates, maps, class instances, accessors, and symbol properties.
- Preserve custom `DocumentStore` method receivers, `instanceof` behavior, test spies, and compatibility with immutable stores.

## Root cause

Application payloads may contain optional properties explicitly assigned to `undefined`. JSON serialization normally omits these properties, but the PostgreSQL-backed HTTP stores perform strict JSON validation before serialization and therefore rejected otherwise valid PBL documents and runtime events.

## Validation

- `pnpm exec tsc --noEmit`
- ESLint on changed files
- `pnpm check`
- 105 targeted persistence, document-store, PBL runtime, hydration, and cutover tests
- Full-suite comparison confirmed that the four remaining `quiz/runtime` cross-tab timeouts also occur on `origin/main`

Fixes #991


cosarah:
I verified the remaining P2 finding and it is reproducible.

`DocumentMigrationDeps.store` bypasses the new plain-JSON write facade. `resolveStore` in `lib/document-store/migration.ts` returns `deps.store` directly, while `getDocumentStore` wraps every injected or configured store with `withPlainJsonDocumentWrites`. During lazy migration, `migrateLocked` then calls `store.saveDocument(expected)` on that raw store.

I reproduced the failure at the exported `accessDocument` seam using the actual `HttpDocumentStore` and a valid legacy PBL scene containing `projectV2.milestones[0].microtasks[0].internalAssessment = undefined`. The migrated document reached `HttpDocumentStore.saveDocument`, where `assertJsonValue` rejected it with `undefined member (dropped by JSON)` before any PUT request was issued. A top-level optional `stage.description = undefined` fails in the same way.

The normal production path is currently protected because app call sites use the default store resolution and do not pass `DocumentMigrationDeps.store`. However, the exported override path has a different persistence contract, and any integration or future caller injecting an HTTP-backed store can hit this failure. The migration regression test currently masks the issue by manually passing `withPlainJsonDocumentWrites(store())`.

Please route migration-injected stores through the same facade, for example by resolving `deps.store` through `getDocumentStore({ store: deps.store })` or `withPlainJsonDocumentWrites`, so `accessDocument` and `mutateDocument` preserve the same JSON-write contract as all other document entry points.

wyuc:
Cross-review is complete.

The review loop resolved:
- persistence verification mismatches after omitting optional `undefined` members;
- repeated PBL runtime snapshots caused by normalized-vs-raw learner-state comparisons;
- document-store facade receiver, frozen-store, and test-reset compatibility issues.

Final independent full-diff reviews from Codex and Claude reported zero actionable findings.

Validation:
- TypeScript, ESLint, and Prettier passed
- 120 targeted persistence and runtime tests passed
- all five required GitHub checks passed

The four local `quiz/runtime` cross-tab timeouts were reproduced unchanged on `origin/main`.

wyuc:
Thanks @cosarah — fixed in `6d13fc1`.

`DocumentMigrationDeps.store` now routes through `getDocumentStore({ store })`, so injected migration and mutation stores use the same plain-JSON write facade as every other document entry point.

I also replaced the manually wrapped browser-store regression with an actual `HttpDocumentStore` test covering both top-level `stage.description: undefined` and nested PBL `internalAssessment: undefined`. The test fails on the parent commit before issuing the PUT.

Validation:
- 120 targeted tests passed
- TypeScript, ESLint, and Prettier passed
- all five required GitHub checks passed
- fresh Codex and Claude delta/full-diff reviews reported zero actionable findings

Please take another look when convenient.

cosarah (approved):
LGTM

PR #994, opened 2026-07-26
feat(roster): single-source the generated agent roster on the stage document

wyuc:
# feat(roster): single-source the generated agent roster on the stage document

## Problem

The generated agent roster currently has a split-brain persistence model:

- The stage document carries `stage.generatedAgentConfigs`, but the actual **read authority** on classroom load is a device-local IndexedDB mirror (`db.generatedAgents`).
- Each agent's **voice binding** (`voiceDesign` / `voiceConfig`) exists *only* on the mirror. It never enters the document, so voices are silently lost whenever the mirror doesn't travel: opening a classroom on another device, server-side fallback hydration, export/import, and shared classrooms all come back voiceless.
- Roster edits persist through a bespoke `debouncedSaveAgents` bypass in the stage store — a second debounce next to the document persistence scheduler, with **no beforeunload / visibilitychange flush**, so an edit made right before closing the tab could be dropped even though the document scheduler already solves exactly this.
- Load-time hydration *wrote* the mirror from the document (and vice versa), so the two sources continuously overwrote each other depending on which path ran last.

## Change

Make the stage document the single source of truth for the roster; demote the mirror to a read-only migration source.

**Contract (`@openmaic/dsl`)**
- `GeneratedAgentConfig` gains optional `voiceConfig` / `voiceDesign`; `VoiceDesign` moves into the contract (the app re-exports it from `lib/audio/voice-design.ts`).
- `AgentVoiceConfig` is deliberately minimal: `{ providerId, voiceId }`. A speculative `modelId` field was dropped after auditing every roster producer (current and historical) and finding none that ever emitted it — contract fields are added only once a producer exists.
- Additive optional fields only — no change to the meaning of any existing field, so per the version policy in `version.ts` this is not a serialized-shape break for this codebase's tolerant structural validators: `DSL_VERSION` stays put and no migration step is added. The generated JSON Schema picks the fields up automatically (covered by a schema test); see Known limitations for what this means for consumers that pin the published schema artifact.

**Writes converge on the document**
- `debouncedSaveAgents` and its private `debounce` helper are deleted. `setStageAgents` now marks `{ kind: 'stage' }` on the shared persistence scheduler (the roster is simply part of the stage document) and performs only synchronous in-memory side effects: registry + selection mirrors.
- The selection mirror honors provenance (matching `restoreAgentSelection` semantics): a stage-derived selection tracks the roster wholesale; a user-set preset selection is not touched at all. A user-set **auto** selection is handled by intent: if it equaled the pre-edit full roster (the AgentBar auto toggle snapshots the whole roster, meaning "everyone"), it keeps tracking the roster wholesale so newly added agents join; a genuine subset is only narrowed — departed agents are dropped, newcomers are never auto-selected. If narrowing would leave an empty selection, the mirror falls back to the full roster and clears `agentSelectionIsUserSet`, mirroring `restoreAgentSelection`'s `length > 0` gate instead of running the classroom with zero agents.
- The registry's zustand `persist` now `partialize`s generated agents out of the localStorage snapshot, so the stage document is the roster's only durable home (the rehydration `merge` filter stays as defense in depth for pre-existing snapshots).
- `saveGeneratedAgents` (IndexedDB + registry) is replaced by a pure in-memory `applyGeneratedAgentsToRegistry(stageId, configs)`; voice warm-up stays as a fire-and-forget.
- Generation embeds the roster (voice included) on the stage before `saveToStorage`; import maps manifest agents straight into `document.stage.generatedAgentConfigs` inside the single atomic document commit; the classroom-load hydration paths no longer write the mirror at all.

**Reads follow the document**
- Classroom load hydrates the registry from the loaded stage's `generatedAgentConfigs`; export reads the stage roster directly (the previous DB-first/stage-fallback dual read is gone).

**Lazy migration (mirror kept, read-only)**
- On load, if the document's roster field is absent entirely, or an agent is missing its voice fields, the mirror is consulted: a missing roster is lifted whole, missing voice fields are backfilled by agent id (document fields always win). The merged result is committed onto the in-memory stage and marked dirty, so the next flush makes it durable — idempotent on subsequent loads. Absent (`undefined`) and explicitly empty (`[]`) rosters are deliberately not collapsed: only an absent field can mean "predates roster persistence" and trigger the full lift; a persisted `[]` is an authoritative empty roster and is never resurrected from the stale mirror (no writer produces `[]` today, but the read side no longer depends on that invariant).
- A fruitless probe is remembered per stage for the session, so classrooms whose producer never binds a voice (e.g. server-generated rosters) don't re-query the mirror on every load forever. Only a probe that **succeeded** and confirmed there is nothing to merge is memoized: a failed mirror read returns a distinct signal (`null`) and is retried on the next load, so a transient IndexedDB error cannot suppress migration for the rest of the session. A successful merge is not memoized either: if its flush fails, the next load retries until the document carries the voices. No persistent marker is introduced.
- Every landing point after an `await` re-checks the current load token and stage id, so switching classrooms mid-load can neither commit a stale roster nor leak it into the global agent registry.

**Deletion safety (deletion-generation model)**
- Stage deletion is fenced by a session-scoped, per-stage **deletion generation (epoch)** plus a deleted flag (`lib/utils/deleted-stages.ts`). `deleteStageData` marks the deletion — bumping the epoch — before removing anything, then discards the stage's queued persistence work. Every persistence path captures the epoch at the moment it captures the data it intends to write (flush-round snapshot, departing-stage snapshot, aggregate save entry) and re-checks `captured epoch == current && !deleted` immediately before each individual write (saveDocument / putScene / putStage / the currentScene and chats tail writes — the two tail writes re-check independently, since no lock spans the tail). Where Web Locks are unavailable the mutation callback is lock-free best-effort LWW, and the per-write re-checks shrink that window.
- A deleted id can legitimately come back, because deletion only removes client-side data: explicit document (re)creation points lift the deleted flag — restoring the server copy on a classroom revisit (`applyClassroomStageAndScenes`) and restoring a backup (`importDatabase`) — so edits after a same-session restore persist again. The epoch is deliberately **never rewound** by a restore: a pre-delete flush still in flight stays permanently stale and cannot overwrite the restored document, while post-restore edits capture the current epoch and persist normally. This is the distinction a boolean tombstone cannot express — after lifting one, pre-delete in-flight writes and post-restore edits were indistinguishable. In-flight flushes never lift the flag.
- Deleting a classroom evicts it from the warm in-memory store, and `loadFromStorage` treats a warm-but-deleted stage as not loaded (discarding the ghost), so navigating Back to a deleted classroom reaches the server-restore path instead of rendering a fully editable ghost from memory whose every edit is silently dropped. The warm-ghost handling is **settlement-driven**: the deletion cascade records an unsettled-cascade count (begun by `deleteStageData`, settled in its `finally`) plus a per-stage settlement promise, and a load that finds the warm stage mid-deletion **parks on that promise** instead of completing — completing early would expose a classroom whose edits are refused without being in the deletion's restore snapshot (silently lost on a failed delete), and whose route the success-path eviction would blank with nothing left to reload. Once settled, the parked load branches on the outcome: a failed delete lifted the flag and restored the pending dirt, so the warm state is kept as the live classroom; a successful delete falls through to a full cold load so the server-restore path recovers the route (the eviction deliberately does not claim a new load token, keeping the parked load current; a real navigation during the park is still caught by the usual token guard). `deleteStageData` is single-flight per stage — a concurrent second call joins the first cascade — and the in-flight state is a counter rather than a bit, so overlapping begin/settle pairs can never expose an undecided cascade as settled. The park cannot deadlock: a parked load holds no document lock, and the cascade never waits on a load. Eviction itself re-checks `isStageDeleted` at eviction time, so a same-id restore completing inside the cascade-tail window is never wiped, and the load path re-checks the deleted flag before landing hydrated data (the read-side mirror of the per-write re-checks), so a lock-free mid-cascade read cannot re-materialize the ghost.
- An epoch-fenced drop is **reported, not disguised**: `saveStageData` / `saveStageDataIncremental` return a distinct `'stale-dropped'` status instead of a success shape, and both store-side consumers react by skipping all success bookkeeping — `saveToStorage` performs no `chatSnapshot` rebind to a snapshot that never landed and no pending-dirt clearing, returning `false` to keep its "true means verified write" contract honest, and the debounced flush path propagates the status out of `persistDirtySnapshot` so the flush round skips its `chatSnapshot` rebind too (a falsely rebound baseline would let later chat saves no-op-skip chats that were never persisted and would corrupt cross-tab conflict detection; pending-map clearing keeps treating a drop as "nothing to retry", which is sound because restored descriptors mint fresh revisions). `capturedEpoch` is a required parameter on both storage entry points, so the type system enforces that the epoch is captured at the same instant as the data snapshot.
- A deletion that fails before the document is removed lifts the deleted flag — and restores the pending changes it discarded (snapshotted before the deletion was marked, including the in-flight flush round's dirt), so pre-delete edits still reach durability. The restore re-queues change descriptors; their flush re-captures the current store state under the current (bumped) epoch, so nothing replays a stale capture. Edits attempted during the deletion window itself are refused and are not part of the snapshot (documented; deletion runs from the home page, where the stage is not being edited).
- A failed backup import rolls back the deletion state it lifted: when the rolled-back document was deleted pre-import, the deletion is re-marked so an outstanding flush cannot recreate the absent document.
- Deleting a stage also clears its rows from the legacy roster mirror (best-effort; a deleted stage needs no migration source). The deletion state is per-tab by design; a sibling tab's scheduler is outside this fence (documented).

**Portability & import hardening**
- Classroom ZIP manifests now carry `voiceConfig` / `voiceDesign` per agent, so export → import preserves agent voices (up to id renaming).
- Manifest voice fields are structurally validated on import (manifests are parsed JSON from user-supplied ZIPs): a malformed `voiceConfig`/`voiceDesign` is dropped per field while the agent survives. On the registry side, `applyGeneratedAgentsToRegistry` validates `providerId` against the known TTS provider registry (built-ins + the `custom-tts-` namespace) instead of blindly casting; unknown providers are treated as "no bound voice". The provider lookups use own-property semantics (`Object.hasOwn`) rather than `in`/bare indexing, so prototype-chain keys (`toString`, `constructor`, …) can neither pass the whitelist nor resolve to `Object.prototype` members.

## Tests

- `tests/store/stage-agents.test.ts` rewritten for the scheduler-based model: document flush carries the roster (voice included), synchronous registry/selection mirroring, the selection-provenance matrix (stage-derived vs user-set × auto/preset, empty-intersection fallback, full-roster intent tracking newcomers, genuine subsets excluding them), scene advances never touch the registry, pending-discard on deletion, deletion fencing (queued-dirt fencing, refusal of new dirt, the departing-stage retry window — including a delete → restore straddle inside the retry window), failed-deletion dirt restore (snapshot includes the in-flight round; restore no-ops after a stage switch; the restored dirt flushes under the current epoch), and post-restore edits persisting under the new epoch.
- New `tests/utils/stage-storage-deletion.test.ts`: storage-layer deletion fencing — the missing-destination full-save is pinned for live stages and blocked for deleted ones (including the lock-race re-check and the lock-free per-write re-checks on both save paths), the incremental currentScene/chats tail is dropped after a mid-mutation delete (and the chat write re-checks independently after the currentScene write), plus the epoch invariants: a round captured pre-delete is dropped even after a same-id restore lifts the deleted flag (incremental, aggregate, and tail), an explicitly stale captured epoch is dropped while the stage is live, post-restore captures persist, mark → unmark → mark keeps the epoch strictly increasing, dropped writes report the distinct `'stale-dropped'` status, and the cascade in-flight state is tracked separately from the deleted flag (re-marks preserve it, overlapping cascades stay in flight until the last settle, and the settlement promise resolves only then). `deleteStageData` wiring: pre-deletion dirt snapshot, mark-before-discard ordering, warm-store eviction on success (and not on failure), eviction only after settlement (in flight during the cascade, settled — with the deleted flag intact — at eviction time and on both failure shapes), settlement-promise resolution on success and on failure, single-flight join of concurrent same-stage deletions (shared cascade and shared failure), mirror-row cleanup (and its failure tolerance), and deleted-flag lift + dirt restore (epoch kept bumped) vs. keep on failed deletions.
- `tests/classroom/load-classroom.test.ts`: existing supersede matrix retained; new coverage for document-first hydration, voice backfill + commit, full-roster lift (absent-field only; an explicitly persisted `[]` is authoritative and never resurrects stale mirror rows), idempotence, the stale-load / stage-id race gates, the fruitless-probe memo (skip on later loads, per-classroom scoping, no memo on successful merges, no memo on failed reads with retry on the next load), and the deletion lifecycle (explicit server-copy restore lifts it and edits persist again; unrestored deletions keep dropping writes; a deleted classroom still warm in the store is restored through the REAL `loadFromStorage` — the back-button path — and edits persist end-to-end afterwards, including a warm ghost with ZERO scenes (the ghost handling keys on stage identity, not scene count); a Back landing mid-cascade parks until the deletion settles, after which a failed delete keeps the warm classroom editable with the pending edits restored end-to-end, a successful delete completes through a cold load into the server restore, and a navigation that moved on during the park is never overwritten by the released load; the eviction refuses to wipe a classroom restored-and-unmarked during the cascade tail). `tests/store/stage-document-persistence.test.ts`: a fence-dropped save returns `false`, does not rebind `chatSnapshot`, and keeps the pending dirt. `tests/store/stage-incremental-persistence.test.ts`: a fenced in-flight flush round skips the `chatSnapshot` rebind, and restored chats flush afterwards against the honest pre-fence baseline. `tests/store/stage-generation-complete.test.ts`: a load whose stage is deleted during hydration does not land (read-side re-check).
- `tests/runtime/database-chat-cutover.test.ts`: a backup import lifts the deletion state of a recreated stage (epoch preserved), and a failed import that rolls the document back reinstates the deletion state it lifted.
- New `tests/lib/orchestration/apply-generated-agents.test.ts` for the registry apply: voice mapping, unknown/custom provider handling, prototype-chain provider keys dropped, cross-stage replacement — plus persistence exclusion (localStorage snapshot carries no generated agents; rehydration merge drops them as backstop).
- New `tests/audio/tts-provider-guard.test.ts`: own-property semantics of `isKnownTTSProviderId` / `getTTSProvider` / `getASRProvider` / `findVoiceDisplayName` against prototype-chain keys.
- `tests/export/classroom-zip.test.ts`: voice round-trip through the manifest, and structural sanitization of malformed manifest voice fields (per-field drop).
- `@openmaic/dsl` schema test: stage rosters with and without voice fields validate.

`pnpm check`, `pnpm lint`, `tsc --noEmit`, and the full vitest suites (root + `@openmaic/dsl`) are green.

## Known limitations / follow-ups

- The new voice fields are non-breaking for this codebase's tolerant TS validators, but they are **not** transparent to consumers that validate documents against a pinned copy of the published `stage.schema.json`: the generated schema sets `additionalProperties: false`, so an older schema artifact rejects any document carrying a voice field. Such cross-language consumers must upgrade their schema artifact in lockstep (a `DSL_VERSION` bump would not help them — an old schema rejects the new documents either way). Whether to drop `additionalProperties: false` from the generated schema for forward compatibility is a separate policy decision, left out of this PR.
- `validateStage` (structural-subset validator) has no `generatedAgentConfigs` branch yet; roster shape is enforced at the ingress points (import sanitization, registry provider whitelist) rather than at the document validator. Extending `validateStage` to cover the roster is left as a follow-up.
- The legacy mirror probe memo is session-scoped by design: a classroom with nothing to migrate costs at most one IndexedDB query per session.
- The deletion-generation state is per-tab: a sibling tab holding pending edits for the deleted stage still flushes through its own scheduler. Deletion vs. live editing in another tab is a cross-tab invalidation problem the per-tab scheduler never handled; documented at the deleted-stages module.

### DIFF
--- a/app/classroom/[id]/page.tsx
+++ b/app/classroom/[id]/page.tsx
@@ -19,7 +19,6 @@ import {
   applyClassroomStageAndScenes,
   defaultClassroomLoadDeps,
   runClassroomLoad,
-  saveGeneratedAgentsForCurrentLoad,
 } from '@/lib/classroom/load-classroom';
 
 const log = createLogger('Classroom');
@@ -59,13 +58,12 @@ export default function ClassroomDetailPage() {
             isCurrent,
             applyStageAndScenes: applyClassroomStageAndScenes,
           }),
-        saveGeneratedAgents: (stageId, agents) =>
-          saveGeneratedAgentsForCurrentLoad(stageId, agents, isCurrent),
         loadRestoredMediaTasks: defaultClassroomLoadDeps.loadRestoredMediaTasks,
         applyRestoredMediaTasks: defaultClassroomLoadDeps.applyRestoredMediaTasks,
         discardRestoredMediaTasks: defaultClassroomLoadDeps.discardRestoredMediaTasks,
-        loadGeneratedAgentRecords: defaultClassroomLoadDeps.loadGeneratedAgentRecords,
-        applyGeneratedAgentRecords: defaultClassroomLoadDeps.applyGeneratedAgentRecords,
+        loadLegacyAgentFallbacks: defaultClassroomLoadDeps.loadLegacyAgentFallbacks,
+        commitMigratedAgentConfigs: defaultClassroomLoadDeps.commitMigratedAgentConfigs,
+        applyGeneratedAgents: defaultClassroomLoadDeps.applyGeneratedAgents,
         getSettings: () => useSettingsStore.getState(),
         getAgent: (id) => useAgentRegistry.getState().getAgent(id),
         restoreAgentSelection: defaultClassroomLoadDeps.restoreAgentSelection,
--- a/app/generation-preview/page.tsx
+++ b/app/generation-preview/page.tsx
@@ -39,7 +39,7 @@ import {
 } from '@/lib/document/bundle';
 import { buildVideoManifestFromOutlines } from '@/lib/media/video-manifest';
 import { nanoid } from 'nanoid';
-import type { Stage } from '@/lib/types/stage';
+import type { GeneratedAgentConfig, Stage } from '@/lib/types/stage';
 import type {
   SceneOutline,
   PdfImage,
@@ -866,11 +866,17 @@ function GenerationPreviewContent() {
           const agentData = await agentResp.json();
           if (!agentData.success) throw new Error(agentData.error || 'Agent generation failed');
 
-          // Save to IndexedDB and registry. The agent-profile LLM has already
-          // bound each agent's voice (from availableVoices); the fallback for an
-          // invalid/unavailable voice is applied later at the live TTS call.
-          const { saveGeneratedAgents } = await import('@/lib/orchestration/registry/store');
-          const savedIds = await saveGeneratedAgents(stage.id, agentData.agents);
+          // Embed the roster (including its voice binding) on the stage — it
+          // persists with the stage document via saveToStorage below — and
+          // mirror it into the in-memory registry. The agent-profile LLM has
+          // already bound each agent's voice (from availableVoices); the
+          // fallback for an invalid/unavailable voice is applied later at the
+          // live TTS call.
+          const generatedConfigs = agentData.agents as GeneratedAgentConfig[];
+          stage.generatedAgentConfigs = generatedConfigs;
+          const { applyGeneratedAgentsToRegistry } =
+            await import('@/lib/orchestration/registry/store');
+          const savedIds = applyGeneratedAgentsToRegistry(stage.id, generatedConfigs);
           settings.setSelectedAgentIds(savedIds);
           // Stage-derived, not a user choice — must not carry across classrooms.
           settings.setAgentSelectionIsUserSet(false);
--- a/components/edit/AgentsView/useAgentRoster.ts
+++ b/components/edit/AgentsView/useAgentRoster.ts
@@ -80,15 +80,18 @@ export function useAgentRoster(): AgentRosterController {
     () => histState.present[0]?.id ?? null,
   );
 
-  // Guard: only persist after a real user edit, not on initial materialization.
+  // Guard: only commit after a real user edit, not on initial materialization.
   // Set to true inside every mutation path (applyOp, add, undo, redo) so the
   // effect below is a no-op until the user actually touches the roster.
   const isDirtyRef = useRef(false);
 
-  // Persist roster edits to the store. Depends only on `histState.present`.
-  // `stage` is intentionally excluded: setStageAgents mutates `stage`, so
-  // depending on it would re-trigger this effect in an infinite loop (React #185).
-  // setStageAgents already no-ops when there is no stage (lib/store/stage.ts:287).
+  // Commit roster edits to the stage store. setStageAgents updates the stage
+  // document (persisted through the shared pending-change scheduler) and
+  // synchronously mirrors the roster into the agent registry + selection.
+  // Depends only on `histState.present`; `stage` is intentionally excluded:
+  // setStageAgents replaces `stage`, so depending on it would re-trigger this
+  // effect in an infinite loop (React #185). setStageAgents already no-ops
+  // when there is no stage.
   useEffect(() => {
     if (!isDirtyRef.current) return;
     setStageAgents(histState.present);
--- a/lib/audio/constants.ts
+++ b/lib/audio/constants.ts
@@ -36,6 +36,7 @@ import type {
   ASRProviderId,
   ASRProviderConfig,
 } from './types';
+import { isCustomTTSProvider } from './types';
 import {
   VOXCPM_AUTO_VOICE,
   VOXCPM_AUTO_VOICE_ID,
@@ -1308,6 +1309,16 @@ export function getAllTTSProviders(
   const builtIn = Object.values(TTS_PROVIDERS);
   const custom = customProviders ? Object.values(customProviders) : [];
   return [...builtIn, ...custom];
+}
+
+/**
+ * Narrow an arbitrary string (e.g. from a persisted document or an imported
+ * manifest, where the contract keeps providerId an open string) to a known
+ * TTS provider id: a registered built-in provider or the user-defined
+ * custom-provider namespace. Anything else must be treated as "no voice".
+ */
+export function isKnownTTSProviderId(id: string): id is TTSProviderId {
+  return id in TTS_PROVIDERS || isCustomTTSProvider(id);
 }
 
 /**
--- a/lib/audio/constants.ts
+++ b/lib/audio/constants.ts
@@ -1318,7 +1318,9 @@ export function getAllTTSProviders(
  * custom-provider namespace. Anything else must be treated as "no voice".
  */
 export function isKnownTTSProviderId(id: string): id is TTSProviderId {
-  return id in TTS_PROVIDERS || isCustomTTSProvider(id);
+  // Object.hasOwn, not `in`: prototype-chain keys ('toString', 'constructor',
+  // …) must not pass a whitelist built from an object literal.
+  return Object.hasOwn(TTS_PROVIDERS, id) || isCustomTTSProvider(id);
 }
 
 /**
@@ -1328,7 +1330,7 @@ export function getTTSProvider(
   providerId: TTSProviderId,
   customProviders?: Record<string, TTSProviderConfig>,
 ): TTSProviderConfig | undefined {
-  if (providerId in TTS_PROVIDERS) {
+  if (Object.hasOwn(TTS_PROVIDERS, providerId)) {
     return TTS_PROVIDERS[providerId as BuiltInTTSProviderId];
   }
   return customProviders?.[providerId];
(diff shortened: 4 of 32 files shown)


cosarah (commented):
The failed-deletion recovery gap is real. Direct aggregate saves (`saveToStorage`, e.g. generation completion or server-restore hydration) are not tracked in `pendingChanges`/`flushInFlight`. If deletion bumps the epoch, the in-flight aggregate save is dropped; if deletion then fails before removing the document, recovery only restores scheduler descriptors, so the aggregate-only change remains memory-only with no retry. A subsequent reload can lose it. Please add coverage for an in-flight aggregate save plus pre-delete failure, and re-queue aggregate persistence on recovery.

wyuc:
Confirmed — the gap was real. Direct aggregate saves run outside the pending map and the flush round, so the deletion path's dirt snapshot could not describe them: a fenced in-flight `saveToStorage` followed by a pre-removal deletion failure left the aggregate-only content memory-only with no retry, and a reload could lose it.

Fixed by merging a full aggregate re-mark (structure / stage / outline / currentScene / chats plus a descriptor per current scene) into the failed-deletion restore, still guarded on the store holding the stage. Since the next flush recaptures the current store state under the current epoch, it necessarily carries whatever the fenced save held — and, by the same recapture, edits refused during the deletion window, so the scope docstrings no longer declare those excluded.

Coverage added in `tests/store/stage-deletion-aggregate-recovery.test.ts`, driving the real store scheduler and real storage layer end to end: a genuinely fenced in-flight `saveToStorage` (generation completion) plus a deletion that fails before `deleteDocument` ends with the flag durable on the next flush; a deletion-window edit survives the same way; a successful deletion re-marks nothing (control). Both recovery tests fail against the previous restore logic.

cosarah (approved):
LGTM

Issue #986, opened 2026-07-24
[Task] Video export: scripted Quiz question-list scrolling


cosarah:
## Context

Follow-up to #985 and the incremental fidelity track in #867.

Quiz scenes use one continuous scrollable question list rather than pagination. After the cover-card increment, the next step is to represent the classroom flow in video as a deterministic cover-to-content transition followed by an automatic vertical scroll.

## Goal

Show the Quiz start card, transition into a read-only question-list view, and scroll from the top to the bottom at a deterministic visual speed.

## Scope

1. Extend the export visual-segment model introduced by #985 so one scene can carry multiple timed visual states.
2. Emit a read-only Quiz question-list surface:
   - single-choice and multiple-choice options;
   - short-answer input-box treatment;
   - no selected answers and no answer feedback;
   - formula rendering consistent with the existing Quiz math path where practical.
3. Script the visual sequence:
   - show the cover during the introductory narration;
   - transition to the question list after the introduction;
   - hold briefly at the top;
   - scroll downward at a deterministic pixel-per-second rate;
   - hold briefly at the bottom.
4. Measure the rendered content height after fonts and formula assets settle. Skip scrolling when the content fits in the viewport.
5. Allow export visual duration to extend the Quiz scene beyond the authored speech-action duration. Later scenes and subtitles must remain aligned with the resulting global timeline.
6. Define bounded timing rules for very short and very long quizzes so exports remain readable without becoming unreasonably long.
7. Keep the implementation independent of Quiz runtime attempts, persistence, submission, grading, and reviewing states.

## Non-goals

- User interaction or clickable controls in the MP4.
- Capturing saved learner answers or results.
- Automatic submission, AI grading, or review-state playback.
- Pagination: Quiz remains one continuous scrolling list.
- PBL workspace playback or interactive iframe replay.

## Acceptance criteria

- Exported Quiz video visibly transitions from the start card to the question list.
- A long fixture scrolls smoothly from its true top position to its true bottom position without clipped final content.
- A short fixture that fits in the viewport remains static and does not perform a meaningless scroll.
- Scene duration includes the transition, top/bottom holds, and computed scroll duration; the next scene starts after the visual sequence completes.
- Single, multiple, short-answer, long-text, CJK/Latin, and formula fixtures render without exposing answers.
- The output remains deterministic across repeated renders at the same resolution.
- Tests cover visual-segment timing, content-height edge cases, emitter output, and one isolated smoke render.

Depends on #985.
[...]

PR #995, opened 2026-07-26
feat(video-export): deterministic Quiz/PBL cover cards (#985)

### LINKED ISSUE #985: [Task] Video export: deterministic Quiz and PBL cover cards
## Context

Follow-up to #867 item 1 and the L1 export work in #854, #864, and #865.

The current video exporter renders slide scenes as frame assets and represents quiz, interactive, and PBL scenes with generic placeholder cards. For the first non-slide increment, quiz and PBL should be represented by deterministic home-page cards only.

## Goal

Render a stable, self-contained cover card for quiz and PBL scenes in exported Hyperframes projects, without entering learner runtime state.

## Scope

1. Add a generic export visual segment/card representation that can coexist with the existing slide frame path and later support multiple visual states.
2. Quiz cover card:
   - quiz title or scene title;
   - question count;
   - total points;
   - static start-button treatment.
3. PBL cover card:
   - PBL v2 Hero-style title and description;
   - learning gains when available;
   - stage/task counts;
   - Instructor and scenario-character summary when available;
   - a deterministic fallback card for legacy PBL payloads.
4. Keep the card deterministic and independent of:
   - quiz attempt hydration, answers, grading, and review state;
   - PBL learner progress, workspace, chat, submission, and completion state.
5. Prefer an export-specific, escaped HTML card path for this increment. Do not require mounting QuizView or PBLRenderer in the exporter.
6. Replace the generic unsupported-scene diagnostic for these two card-rendered scene types with an informational export diagnostic such as "shown as a cover card".

## Non-goals

- Quiz question-list scrolling or answer interaction.
- PBL workspace or role-play playback.
- Interactive iframe replay.
- Pixel-identical reuse of the live classroom components.

## Acceptance criteria

- A mixed lesson containing slide, quiz, and PBL scenes exports without generic unsupported placeholders for quiz/PBL.
- Quiz scenes show only the deterministic start card for their whole scene duration.
- PBL scenes show only the deterministic start/Hero card for their whole scene duration.
- Export output is self-contained and renders offline in Hyperframes.
- Long titles and generated text are escaped and do not break the emitted HTML.
- Existing slide export snapshots and timeline behavior remain unchanged.
- Unit and emitter tests cover both scene types, including legacy/missing-field fallbacks.

Parent: #867

YizukiAme:
Follow-up to #867 item 1: Quiz and PBL scenes in video export render as deterministic cover cards built only from authored design data — never from attempt hydration, learner progress, chat threads, or submissions — so a mixed lesson exports without placeholder cards.

- Pure `visuals` pass (`timeline → visuals → geometry`) emits `quiz-cover` / `pbl-cover` IR for Quiz/PBL (v2, legacy `projectConfig`, and missing-field fallbacks). `scene.visuals` is a discriminated union so #986 can add a question-list visual without retiming scenes (manifest 1 → 2).
- Offline Hyperframes HTML cards with an embedded Inter Latin subset + OFL license (no CDN at render time). Cover CSS scales from the render width the same way the subtitle band does — no `vw`/`clamp`.
- Quiz/PBL `unsupported-scene` warnings become informational `cover-card` diagnostics.
- Optional destination-configurable CTA via `NEXT_PUBLIC_VIDEO_EXPORT_CTA_DESTINATION` (default `open.maic.chat`, `off` to disable), with Unicode/bidi display guards.
- Dense covers drop sections in reverse priority until the fixed panel fits (720p / 480p), keeping scene type, title, stage/task counts, and the short CTA. Chromium layout evidence in `cover-card-layout.browser.test.ts`.
- Locale for card chrome is an explicit `buildExportZip` input resolved before the first await (UI locale, not course-content language).

Verified:
- `npx vitest run tests/video-export tests/workflows/ci-video-export-contract.test.ts` → 319 passed, 1 skipped
- `COVER_LAYOUT_BROWSER=1 npx vitest run tests/video-export/cover-card-layout.browser.test.ts` → 38 passed
- `npx --yes hyperframes@0.7.60 lint` on materialized samples (`quiz` / `pbl-v2` / `pbl-legacy` / `pbl-dense` / `arabic` / `mixed`) → 0 errors, 0 warnings
- `npx --yes hyperframes@0.7.60 render -q draft -f 24` on `mixed` → 1280×720 @ 24fps H.264; frames show Quiz/PBL covers and the CTA

Closes #985


lib/video-export/passes/visuals.ts
```diff
@@ -0,0 +1,170 @@
...
+ * the board, else the first non-learner agent. `currentIssueId` only breaks ties
+ * between issues; it never makes the cover depend on learner progress, because
+ * every candidate resolves to the same authored agent name.
+ */
+function legacyInstructorName(
+  agents: UnknownRecord[],
+  issues: UnknownRecord[],
+  currentIssueId: unknown,
+): string | undefined {
+  const activeIssue =
```
wyuc: [P1] Keep the legacy PBL cover independent of learner progress

This selection still reads `is_active`, `is_done`, and `current_issue_id`. If different issues use different `question_agent_name` values, changing the learner's current or completed issue changes the exported instructor, even though the authored project is unchanged.

That contradicts the main invariant of this PR: cover cards should be derived only from authored design data and remain independent of learner progress. The current test does not catch this because both issues use the same agent name.

Could we select the instructor only from stable authored agent metadata, or omit the instructor when no unambiguous authored role exists? Please also add a regression test with two different question agents and flipped progress state.

wyuc (changes requested):
## Summary

This PR replaces the generic unsupported placeholders for Quiz and PBL scenes in video exports with deterministic static cover cards.

The implementation adds:

- A new `visuals` compiler pass and first-class `quiz-cover` / `pbl-cover` IR segments.
- Support for Quiz, PBL v2, and legacy `projectConfig` data.
- Offline-ready Hyperframes output with an embedded Inter font subset.
- Render-resolution-aware sizing and localized card labels.
- Informational `cover-card` diagnostics instead of unsupported-scene warnings.
- Coverage for escaping, localization, resolution scaling, fallbacks, and exclusion of sensitive runtime fields.

The overall direction is good. It makes mixed courses exportable without pretending that the interactive runtime itself has been serialized.

## Validation

I ran a real UI E2E against commit `a532a51e0e17ac98479b64887097dca5cc0b12c6`, without fixtures, mocks, or direct generation API shortcuts.

The flow covered:

1. Entering a course prompt on the real homepage.
2. Generating a course containing Slide, Quiz, and PBL scenes.
3. Opening the generated classroom and inspecting both interactive scene types.
4. Starting video export from the real export UI.
5. Completing the server-side render and downloading the MP4.
6. Verifying the result with `ffprobe` and extracted video frames.

Actual result:

- Slide introduction
- Quiz with 2 questions and 20 points
- PBL with 3 stages and 6 tasks
- Course completion scene
- 1280×720, 24 fps, Draft
- 128.448 seconds
- H.264 video with AAC audio
- Quiz and PBL covers both present in the exported video
- Chinese subtitles visibly burned into the exported video

This confirms that the primary Quiz and PBL export path works end to end.

## Blocking finding

The legacy PBL instructor selection is not yet deterministic with respect to learner progress.

`legacyInstructorName()` reads `is_active`, `is_done`, and `current_issue_id`. When different issues are assigned to different question agents, the exported instructor can change according to the learner's current progress.

This conflicts with the PR's stated contract that the cover is derived only from authored design data and never from learner progress. The existing regression test uses the same agent for every issue, so it cannot detect this case.

Please derive the legacy instructor exclusively from stable authored agent metadata, or omit the instructor when no stable choice exists, and add a test with different agents across issues.

## Open product questions

### 1. What should a non-interactive exported card ask the viewer to do?

The exported Quiz card currently contains a “Start Quiz” style button, but a video viewer cannot interact with it. The same semantic gap applies to the PBL scene.

Would it be clearer to replace the inactive control with an explicit CTA, for example:

> Want to complete this quiz or project yourself?  
> Visit open.maic.chat or install OpenMAIC.

The destination should probably be configurable so self-hosted deployments can provide their own domain or installation link. A short form such as “Try the interactive course at open.maic.chat” could be used when space is limited.

This CTA could appear at the end of Quiz and PBL covers for a few seconds, or replace the inactive action button without consuming additional layout height.

### 2. What information must survive at constrained resolutions?

The real 720p PBL export is already visually dense. With longer localized titles, gains, instructor information, or role-play content, the current `max-height: 90%; overflow: hidden` panel may hide an entire lower section rather than only clamp individual text fields.

We should define a clear priority order for constrained layouts. My suggestion is:

1. Scene type and title
2. Stage and task counts
3. Short interaction CTA
4. Learning gains
5. Instructor and role-play details

Lower-priority sections could collapse or disappear at constrained heights instead of allowing the panel to clip unpredictably. It would also be useful to add visual coverage at 1280×720 and a smaller target such as 854×480, using long CJK content and the maximum supported PBL metadata.

## Merge readiness

All current CI checks are green, and the primary path passes real E2E validation. However, the PR is currently reported as conflicting with `main`, and the E2E above was run against the current PR head before resolving that conflict.

My recommendation is:

1. Fix the legacy PBL progress dependency.
2. Resolve the conflict with `main`.
3. Rerun the focused compiler tests and one mixed Quiz/PBL video export.
4. Record the CTA and constrained-layout decisions, even if their implementation is intentionally deferred.

Requesting changes for the deterministic legacy PBL issue. The CTA and constrained-resolution behavior are open product questions rather than blockers for this patch.

YizukiAme:
Thanks for running a real end-to-end pass instead of reading the diff!!
The ffprobe/frame evidence on the mixed course is more than I could have asked for, and it caught a contract violation I had written into my own PR description.

**P1 — legacy PBL instructor (fixed in 3ea7bc4)**

The root cause was exactly as you described: the pass picked the question agent of whichever issue happened to be active, so a board whose issues each spawn their own `Question Agent - <issue>` exported a different instructor as the learner advanced. It now reads the first roster agent that is neither the learner nor an issueboard system agent, and when the roster has no such agent it omits the instructor entirely so the card falls back to its localized tutor label. Added a regression test with two different question agents and flipped progress state, since the old fixture reused one agent name and could never have failed.

**Q1 — inert control**

Initially I thought a button-ish affordance would improve the sense of participation. Yet now I totally agree that a button a viewer cannot press is worse than no button, so the quiz cover is now just eyebrow, title and the question/points counts. 

**Q2 — constrained layouts**

Implemented with the priority order you proposed~ The emitter estimates the panel height in design space and drops sections in reverse priority — role-play/instructor, then gains, then the description — until the card fits, and also drops detail rows whose caption type would render below ~7px at the target size. Scene type, title and the stage/task counts are never dropped. Covered at 1280×720, 854×480 and a squat 1280×480, plus `pbl-dense-720p` / `pbl-dense-480p` E2E samples with long CJK content for visual review.

**Status**

Going back to draft: carefully re-reading the whole patch after your review I found a few more problems of my own that I'd rather fix before asking you to look again. Expect one to three days. I'll also bring the branch up to date with main and rerun the compiler tests plus one mixed Quiz/PBL export before marking it ready :-D

lib/video-export/passes/visuals.ts
YizukiAme: Resolved on the current tip:

- Legacy covers name **no** instructor (no progress fields, no roster promotion that could label a student role as Tutor).
- Migrated/legacy covers also filter out synthetic `role-compat-instructor`.
- Regression coverage: two different question agents + flipped progress → identical IR, and no instructor / Question Agent leakage in the emitted cover

YizukiAme:
Thanks for waiting — back from draft.

Quick status against the review:

**P1 — legacy progress dependency**  
Still resolved on tip. Legacy covers name no instructor at all (a v1 roster has student roles + issueboard bots, never a tutor), and the regression with two different question agents + flipped progress remains. The cover also skips `role-compat-instructor` on the v2 path so a migrated project cannot leak a progress-derived name.

**Q1 — inert control / CTA**  
Fake “Start Quiz” is gone. I previously deferred the configurable CTA; that follow-up is now in this PR instead of a separate issue: `NEXT_PUBLIC_VIDEO_EXPORT_CTA_DESTINATION` (default `open.maic.chat`, `off` to disable), with Unicode/bidi display guards, and the short CTA is kept in the dense-layout budget.

**Q2 — constrained layouts**  
Same reverse-priority drop as discussed (people → gains → description; scene type / title / stage·task counts / short CTA never drop). Chromium guardrail covers 1280×720 and 854×480 with long CJK / wide Latin / emoji / Arabic.

Verified locally: `tests/video-export` + CI contract, `COVER_LAYOUT_BROWSER=1` layout suite (38), `hyperframes lint` on the materialize samples, and a mixed `hyperframes render` with Quiz/PBL covers + CTA visible in frames  

01-quiz-cta-720p

02-pbl-dense-480p

cosarah (approved):
Approved. The previous blocker is resolved, the configurable CTA and default destination are acceptable, and the export-specific static card design fits the scope of #985. CI is green on the latest head.

wyuc (approved):

PR #996, opened 2026-07-26
docs: refresh localized documentation and Docker flags

sophietao20-star:
Refresh the localized OpenMAIC documentation and align Docker/client feature-flag behavior with the current implementation.

- Synchronize configuration, deployment, getting-started, supported-models, and VoxCPM2 guidance across English, Simplified Chinese, Traditional Chinese, Japanese, Russian, and Arabic.
- Document Docker host-service access, ComfyUI SSRF requirements, provider configuration boundaries, parsing formats, persistence, and video export.
- Make NEXT_PUBLIC_ENABLE_PPTX_IMPORT consistently accept true/1 and add focused coverage.


wyuc (commented):
Thanks for refreshing the Chinese documentation. I compared it against the actual registries, configuration paths, and deployment behavior. The built-in provider and model catalogs are accurate, but several gaps should be addressed before approval:

1. The Docker `video-export` profile starts the render service but does not compile `NEXT_PUBLIC_ENABLE_VIDEO_EXPORT` into the app, so the export entry remains hidden. The other documented `NEXT_PUBLIC_*` flags are also not exposed as Docker build arguments.
2. Custom OpenAI-compatible providers can currently be added through Settings, but not through arbitrary environment variables or `server-providers.yml`. Unknown server-side provider IDs lack `providerType` and fail during model resolution.
3. Docker users need explicit guidance for mounting `server-providers.yml` and connecting to host-side Ollama, Lemonade, VoxCPM, or ComfyUI services. `localhost` inside the container points to the OpenMAIC container.
4. ComfyUI setup, custom TTS/ASR providers, AliDocMind-only audio/video parsing and its supported formats, and `NEXT_PUBLIC_ENABLE_PPTX_IMPORT` are still missing from the Chinese docs. The unused `BRAVE_API_KEY` example should also be removed because Brave is keyless.

Please update the other localized documentation at the same time. It is fine to have the agent translate the finalized content consistently across languages.

wyuc (commented):
Thanks, most of the previously reported gaps are now addressed. Three items remain before approval:

1. The corresponding English, Traditional Chinese, Japanese, Russian, and Arabic pages are still unchanged. Please have the agent propagate the finalized content across all localized documentation.
2. The documented `host.docker.internal:8188` ComfyUI setup is still rejected by the production SSRF guard because `comfyui-image` is not server-managed. Please document `ALLOW_LOCAL_NETWORKS=true` for this topology, or add a server-managed ComfyUI configuration path.
3. The documentation says feature flags accept both `true` and `1`, but `NEXT_PUBLIC_ENABLE_PPTX_IMPORT` currently checks only for the literal value `true`. Please align the implementation or narrow the documentation.

wyuc (commented):
Thanks, the PPTX flag and ComfyUI SSRF guidance are now aligned with the implementation. One localization gap still remains: the latest commit translated only the newly added review-related snippets, rather than the full finalized documentation refresh.

For example, the non-Chinese `supported-models` pages still list older entries such as GPT-5.4 Mini and Qwen 3.6 while omitting the updated GPT-5.6, Qwen 3.7, Gemini 3.5, and other provider/model entries. The localized `configuration` pages are also still missing the Lemonade, video generation, default routing, feature flag, and server-option sections, while the localized `deployment` pages omit PostgreSQL persistence and MP4 export.

Please have an agent translate the complete finalized `zh-cn` refresh into English, Traditional Chinese, Japanese, Russian, and Arabic, while preserving environment variables, provider IDs, and model IDs exactly.

wyuc (commented):
The documentation and localization gaps are now fully addressed. I verified that all five localized guide sets match the finalized Chinese structure, and that environment variables, provider/model IDs, commands, and code blocks remain consistent across languages.

The remaining blocker is the merge conflict with `main` in `tests/config/feature-flags.test.ts`. Please resolve it by retaining both the current `main` tests and the new PPTX feature-flag tests, then let the full PR checks run again.

wyuc (approved):
All previously identified documentation, localization, configuration, and deployment gaps are resolved. The merge conflict was handled correctly and the full CI suite passes.

PR #998, opened 2026-07-28
chore(packages): publish storage and enforce version bumps

wyuc:
## Summary

- publish `@openmaic/storage`, which had a complete public API but was excluded from every release stage
- release `@openmaic/dsl` 0.5.0 and `@openmaic/storage` 0.1.0
- require a version increase whenever a publishable package file changes
- make a version bump merged to `main` the only thing that triggers a release
- pin the two PostgreSQL schemas the storage package exports
- document the rule for contributors

## Why

Registry versions had stopped representing the source tree. `@openmaic/storage` was never published, and `@openmaic/dsl` gained public runtime and storage contracts after 0.4.0 with no version change, so "the version on npm" was not a reliable name for any particular source state.

## How releases work after this

A package's `version` in its manifest is the release intent, and merging that bump is what releases it. Which packages go out is decided by comparing the manifests against the registry, never by how the run was triggered.

The tag trigger is removed. It never participated in the decision (its own documentation noted that any `@openmaic/*` tag republishes the whole family and that the manifests decide what actually ships), it bypassed review because a tag is not a diff and can be created by anyone with write access on any commit, and it could not be guarded: environment deployment rules match `GITHUB_REF`, so `refs/tags/*` and a main-only rule are mutually exclusive. `@openmaic/<name>@<version>` tags are now an output, written after a package reaches the registry. Nothing is lost, because a manual dispatch from `main` still republishes without a new commit.

## Two gates, in the places where each can actually prove something

**Merge time (`ci.yml`).** A change to a publishable file must carry a version increase. This runs on pull requests and on pushes to `main`, where a real push range always exists, so it sees every commit before it can be released. This is where drift is prevented.

**Release time (`publish-packages.yml`).** A pre-publish gate judges each package against the registry: a version not yet published is a release, an already-published version is reported and left alone, and any registry answer that is not a definitive 404 stops the run.

An earlier revision of this branch also tried to prove at release time that an already-published version still matched its source. That was dropped deliberately. There is no trustworthy record to prove it against: git tags are mutable and can be created or moved by hand, pnpm records no `gitHead`, and a push range describes one push rather than the origin of a release. Treating a forgeable anchor as authoritative is worse than not checking, because it turns "unproven" into "proven".

## Publishing is restricted

- the workflow is split so `validate` never touches `NPM_TOKEN` and runs from any ref, keeping dry runs useful; `publish` is the only job declaring the `release` environment
- `NPM_TOKEN` lives in that environment behind a `main`-only deployment branch rule, not in a repository secret, so a branch-local edit of the workflow cannot reach it
- the commit must be on `main`'s first-parent history, which plain reachability would not give: a pull request that adds a bad tree and reverts it in the next commit leaves that tree an ancestor of `main` forever
- CI must have concluded successfully for that exact commit, identified by workflow file and triggering event rather than by a check-run display name
- packages are published one at a time and marked as they land, so a partial failure stays retryable
- `--ignore-scripts`, because every package's `prepublishOnly` reruns the build and would replace the dist that was just verified
- the build must not rewrite tracked package files, which would publish content absent from the released commit
- storage's PostgreSQL contract suites run against a real PostgreSQL 16 before storage is published; they skip themselves without a database

## Schema pins

`DOCUMENT_PG_SCHEMA` and `RUNTIME_PG_SCHEMA` are public API. A deployment that provisions these tables with its own migration tooling has to reproduce the DDL exactly for `ensureDocumentSchema()` / `ensureSchema()` to stay the intended no-op. Nothing guarded that, and every statement is `CREATE ... IF NOT EXISTS`, so PostgreSQL silently accepts whatever table already exists under the name: a column type, a nullability, an index or a FK action can drift apart from a downstream migration without an error.

Both constants are pinned verbatim, every statement is asserted to stay `IF NOT EXISTS` guarded, and both ensure functions are executed against a recording queryable so the statements they actually issue cannot drift from the constants.

## Known limitations

- **Package inputs are approximated by the package directory.** Renderer and importer inline their dependency graph through Rollup, and dsl and storage emit whatever the lockfile's TypeScript produces, so a toolchain bump can change any of the four tarballs with no diff under the package directory. Closing this means treating the lockfile as an input of every package, which makes every dependency bump demand four version bumps, or externalising bundled dependencies. Both are decisions about how the packages are built.
- **Versions removed with `npm unpublish` are absent from the registry listing**, so the gate can plan a version npm will then reject. It fails at publish rather than unsafely.
- **The repository-wide concurrency group can drop a superseded queued run.** That is preferred over two runs each concluding independently that the same version is unpublished.

## Validation

- `pnpm check`, `pnpm lint` (0 errors), `pnpm exec tsc --noEmit`
- `VITEST_MAX_WORKERS=4` across the published packages: dsl 190, storage 409, renderer 242 passing
- three rounds of independent cross-vendor review on the full branch diff, adversarially framed as "enumerate how to defeat this gate" rather than "confirm it works"
- fault injection: a committed change under `renderer/src` without a bump is rejected; a tree rolled back to an older published version is rejected; altering a column type in `DOCUMENT_PG_SCHEMA` fails the pin; making `ensureDocumentSchema()` skip its index statements fails the recorded-statement assertion while the constant is untouched
- two release-blocking defects found and fixed this way: `git fetch --depth=0` is not valid git, and `git rev-list | grep -qx` under `pipefail` returns 141 once history outgrows the pipe buffer, which would have reported every valid commit as "not on main"


cosarah:
The package-version check does not currently gate publishing. It runs only in `ci.yml`, while `publish-packages.yml` is a separate workflow and can reach `pnpm publish` on main pushes, matching tags, or a manual run with `dry_run=false` even if CI fails or never runs. Please run the version validation inside the publish job, before the publish step, using an appropriate base for each trigger.

wyuc:
Good catch — fixed in 2219f07.

The validation now runs inside the publish job, before the publish step. One finding while implementing it: a per-trigger git base does not actually work here.

- there are no `@openmaic/*@*` tags in the repository, so a tag run has nothing to diff against;
- `github.event.before` is all zeroes for a tag push and absent for `workflow_dispatch`;
- the published versions carry no `gitHead` (they were published through pnpm), so the registry cannot supply a commit either.

So a base-derived check would have been toothless on exactly the two triggers you named. Release mode judges each package against the registry instead, which holds for every trigger: a version that is not published yet is a release, and a version that is already published is accepted only when the package source has not moved since that release. The second half is what actually matters — `pnpm publish` skips an already-published version, so a drifted package would otherwise release as a silent no-op.

To make "has not moved since that release" answerable, the job now writes an `@openmaic/<name>@<version>` tag after every successful publish. That tag is the anchor for subsequent runs, independent of trigger. The push range is kept as a fallback for the packages released before those tags existed (renderer 0.0.3, importer 0.1.0), and with neither anchor available the release fails closed rather than guessing. The tag pushes use the default `GITHUB_TOKEN`, which does not start new workflow runs, so they do not re-trigger this workflow.

I also considered comparing the packed tarball against the published one instead of using a git anchor, and measured it before discarding it: `@openmaic/renderer@0.0.3` reproduces exactly (212/212 files, only the `workspace:*` rewrite in `package.json` differs), but `@openmaic/importer@0.1.0` does not — it bundles third-party code, so its `dist` shifts with transitive dependency resolution rather than with its own source. Content comparison would have produced false failures that block releases.

Behaviour verified for each trigger shape, including fault injection: a committed change under `renderer/src` without a version bump is rejected through both the tag anchor and the push-range fallback.

cosarah (commented):
I found two remaining issues worth addressing before approval:

- The repository now has an `NPM_TOKEN` repository secret in addition to the `release` environment secret. This defeats the workflow's main-only environment boundary: a workflow running from another in-repository ref can access the repository secret without declaring the protected environment. Please remove the repository-level `NPM_TOKEN` and keep it only in the `release` environment.
- The `publish` job grants `contents: write`, and `actions/checkout` persists that credential before `pnpm install` and the package builds run. Dependency or build code can therefore push repository refs. Please use `persist-credentials: false` for jobs that execute package code and move release-marker writes into a narrowly scoped job that does not install or build packages.

Non-blocking: a failed or incorrect release-marker tag is never repaired. The workflow only warns, while a retry omits the already-published package from the release plan and skips its marker handling. This does not affect npm publication safety because tags are non-authoritative, but either the marker should be reconciled on retry or the PR should document that repair is manual.

xuyuanwei678 (approved):

wyuc:
Both blocking points are addressed, and the non-blocking one is fixed rather than documented.

**Repository-level `NPM_TOKEN`** — removed. `NPM_TOKEN` now exists only in the `release` environment, whose deployment branch rule allows `main` alone.

**Git credential reachable from package code** — fixed in `3610512`. The two capabilities are now in different jobs, and no job holds both:

- `validate` and `publish` are the jobs that run `pnpm install` and the package builds. Both check out with `persist-credentials: false`. `validate` is pinned to `contents: read`, and `publish` drops write scope entirely, keeping only `actions: read` for the CI conclusion and `id-token: write` for provenance.
- a new `mark` job is the sole holder of `contents: write` and a git credential. It installs nothing and builds nothing: it reads the manifests, queries the registry, and writes tags.

A side effect worth noting: `publish` no longer runs any test step, so a test that mutated a built artifact after the tracked-file check could no longer affect the bytes that reach npm.

**Release markers never repaired** — `mark` now reconciles instead of only recording the current run. It marks any registry version whose tag is missing, which the release plan could never do, because a version already on the registry is excluded from that plan and the marker logic sat behind it. An existing tag is reported and left alone rather than moved.

---

On the review process for this branch, since the diff grew well past its original scope: it went through three rounds of independent cross-vendor review of the full branch diff, framed adversarially as "enumerate how to defeat this gate" rather than "confirm it works".

Round one rejected the original design outright. It tried to prove, at publish time, that an already-published version still matched its source, anchored on git tags. Tags are mutable and creatable by hand, pnpm records no `gitHead`, and packages released before the scheme had no anchor at all, so the anchor was both forgeable and absent. That whole mechanism was removed; drift prevention lives at merge time, where a real push range always exists.

Rounds two and three each surfaced a defect that would have broken every real release: `git fetch --depth=0` is not valid git, and `git rev-list --first-parent | grep -qx` under `pipefail` returns 141 once main's history outgrows the pipe buffer, which would have reported every valid commit as "not on main". Neither reproduces in a shallow clone.

Three limitations are documented rather than fixed, in `scripts/check-package-version-bumps.mjs` and the PR description: package inputs are approximated by the package directory, so a lockfile or toolchain bump can change a tarball with no diff there; `npm unpublish`ed versions are absent from the registry listing, so the gate can plan a version npm then rejects; and the repository-wide concurrency group can drop a superseded queued run, which is preferred over two runs each concluding the same version is unpublished.

Follow-up filed as #1012: adopt changesets, which supersedes the hand-maintained version gate and adds the changelogs and internal dependency-range updates this cannot do.

Issue #1000, opened 2026-07-28
[Task] @openmaic/storage: server backends for KVStore and AssetProvider (#779)


wyuc:
Tracking: #779. Completes the component table in that RFC — two of the four
components still have a browser backend only.

## Where we are

#779's component table promises every component a browser backend and an HTTP backend:

| Component | Browser backend | HTTP backend |
|---|---|---|
| `DocumentStore` | shipped (#860) | shipped (#979) |
| `RuntimeStore` | shipped (#880) | shipped (#946) |
| `AssetProvider` | shipped (#858) | **missing** |
| `KVStore` | shipped (#858) | **missing** |

So a server-backed deployment (#982) today persists a course's document rows and its
learner runtime, and nothing else. Two consequences show up immediately:

1. **Generated media does not travel.** The bytes behind generated images, audio and
   video live in the browser (`audioFiles` / `imageFiles` / `mediaFiles` in IndexedDB;
   the app's own `lib/storage` seam ships only a no-op provider). A course opened on a
   second device against the same server gets its structure and text but cannot resolve
   its generated media. This is the single largest gap between "server-backed" as
   currently shipped and what users asking for server persistence expect (#582, #585, #596).
2. **`account`-scoped KV has nowhere to go.** `KVStore` already distinguishes `device`
   from `account` scope, and the RFC defines `account` as the scope a server-backed
   deployment syncs across devices. With no server backend, `account` behaves exactly
   like `device` — the distinction is currently unenforceable rather than merely unused.

### Part A — `AssetProvider` server backend

**Moved to #1007.** The `AssetProvider` server backend is being redesigned as part of
the global asset pool (#1007) — allocated asset ids over a content-addressed blob store,
which changes the `AssetRef` model itself. It is no longer tracked here; this issue is
now effectively the KV server backend (Part B/C below). The original sketch is kept
below for reference only.

- HTTP contract for `put(blob, meta?) → ref` / `resolve(ref) → url` / `remove(ref)`,
  documented alongside the existing two contracts in
  `packages/@openmaic/storage/docs/`.
- `HttpAssetProvider` client implementing the DSL `StorageProvider` over it.
- A reference server-side implementation storing bytes on a filesystem path or an
  S3-compatible endpoint, with metadata rows in Postgres — bytes do not go into the
  database (#779 non-goal).
- Content addressing is preserved end to end: the same bytes must produce the same
  `sha256-<hex>` ref as the browser backend, so refs stay portable between backends
[...]

PR #1001, opened 2026-07-28
feat(storage): route settings/user-profile persist through KVStore (no legacy migration)

wyuc:
Part of #857.

Routes the `settings` and `user-profile` zustand `persist` stores through the
`@openmaic/storage` `KVStore` instead of zustand's default `localStorage`, so
one contract owns their persisted state and a server-backed deployment can serve
the `account` scope without the stores knowing. `kvPersistStorage` shipped in the
package (#858) with no importer; this wires it up.

## No automatic migration of legacy data

**This PR does not migrate pre-cutover `localStorage` data.** Earlier revisions
tried to adopt the old raw `settings-storage` / `user-profile-storage` blobs (and
the even older `llmModel` / `providersConfig` keys) into the KV scope. Cross-vendor
and red-team review found that legacy-adoption path to have a state space too
large to keep correct — successive rounds each surfaced a new data-loss P1 (delete
races, resurrection after an external KV wipe, reader-device resurrection). The
maintainer's terminal decision was to remove the whole adoption subsystem rather
than keep patching it.

The result: the adapter reads and writes the KV scope and **touches no legacy
key**. An empty KV scope hydrates the store to its defaults (ordinary zustand
behaviour). An upgrading user reconfigures their settings once — the data is
reconstructable and the cost is one-time.

Removed with the adoption subsystem: `adopt()`, the raw-key read, the
`__migrated:` resurrection marker, the confirm-re-read, the `provisional` value
concept, the `prePersistFallback` + `__pre-persist-consulted:` sentinel, and
`settings.ts`'s `readPrePersistSettings` (the `llmModel`/`providersConfig` reader).
`lib/ai/providers.ts` still reads `providersConfig` independently, so those source
keys are left untouched. Net: the adapter drops ~460 lines.

**Product follow-up (not in this PR):** an upgrading user's settings reset to
defaults on first load. Worth a one-time "storage upgraded — please reconfigure"
notice; this PR ships no such UI.

The stale raw blobs are best-effort purged (`purgeLegacyPersistKey`, fire-and-
forget, no correctness dependency) — they are never read, and the old settings
blob held plaintext provider API keys, so clearing it is a small security win.

## What is kept: async-storage steady state

A `KVStore` can be remote and can fail, in ways synchronous `localStorage` never
did. The machinery that survives is steady-state, not migration scaffolding:

- **Hydration + write gate.** A failed read is not an empty store: the key stays
  unsettled and the write gate refuses writes, so an always-on initializer's
  `set()` cannot land defaults over a value that could not be read. The gate is
  read when `setItem` is *called*, not when its queued turn arrives.
- **`Outcome` discipline.** Every backend call returns a sealed result whose only
  reader is the state machine, so no call site can quietly read a failure as
  "nothing there".
- **Per-key serialization**, so concurrent writes cannot land out of order.
- **Recovery + replay.** A write refused while the backend was down is replayed
  on recovery when it was taken against the authoritative value, and reported as
  lost otherwise; recovery is bounded (backoff + cap) so a read-ok/write-fail
  backend cannot loop.
- **`persist-health`** raises a durable, de-duplicated toast when changes are not
  being saved, and retracts it on recovery.
- **Clearing** enters a synchronous `clearing` phase so a `set()` racing a clear
  cannot write the just-deleted value back.

## Scope choice

The scope is a contract, not a tuning knob — `device` values never leave the
machine under any backend, `account` values are user data a server-backed
deployment may sync across devices.

| Store | Key | Scope |
| --- | --- | --- |
| `lib/store/settings.ts` | `settings-storage` | `account` |
| `lib/store/user-profile.ts` | `user-profile-storage` | `account` |

Both are `account`: provider/model configuration and the user's own profile are
exactly what a second device should not have to be re-entered. The `KVStore`
docstring names "provider/model config" and "profile" as its two `account`
exemplars.

`settings-storage` is not purely `account` — playback volume/speed, sidebar/rail
layout, and per-agent voice picks are device-ish by the #779 taxonomy but ride in
the same blob, and plaintext provider API keys would sync under `account` once the
HTTP KV backend lands (#1000). Splitting the store is a store-logic change out of
scope here; both are follow-ups, called out for a maintainer's opinion.

Neither store uses `device`; it already has real consumers in
`lib/runtime/learner-key.ts` and `lib/document-store/current-scene.ts`.

## Behaviour change: async hydration

Hydration resolves from a promise now (a `KVStore` may be remote). The browser
backend resolves within a few microtasks of module evaluation, so nothing
user-visible changes; tests that assert on persisted state await
`persist.rehydrate()`.

**Follow-up prerequisite for a remote `account` backend (#1000):** the app has no
consumer of hydration state, so one-shot decisions taken against a not-yet-
hydrated store would decide wrongly under a network round trip
(`lib/utils/load-classroom.ts:202-221`, the server-provider reconcile, media
orchestration, scene-generator retry). A hydration gate the app consumes belongs
with the backend that makes it necessary.

## Verification

- `pnpm check`, `pnpm lint`, `tsc --noEmit`, `pnpm check:i18n-keys` clean.
- `pnpm test` — 3342 passed. `pnpm --filter @openmaic/storage test` — 401 passed.
- Full Playwright suite green (27/27) against a production build.

`tests/store/kv-persist.test.ts` (40 cases) covers the steady state: round trip,
scope isolation, empty-KV → defaults, failed-read-never-persists-defaults, writes
issued during hydration, write ordering, clearing (synchronous gate, no replay
across a clear), recovery (rehydrate, replay owed until it lands, changes-lost
when it can't, bounded + re-armable), unreachable browser storage, cold-start
writes, SSR no-op. Faults are injected at the `KVStore` backend, not mocked at the
adapter.

`tests/store/persisted-store-scopes.test.ts` pins each store's declared scope and
proves the "no migration" semantics directly: a seeded raw `settings-storage` /
`user-profile-storage` blob is ignored (the store hydrates to defaults) and
purged, and the ancient `llmModel` / `providersConfig` keys are neither read into
state nor (for `providersConfig`, which has a live reader) removed.

The e2e seed fixtures move from the raw persist key to the KV-scope key.


wyuc:
Resolving the open scope question this PR flagged (BYOK API keys ride under the
`account`-scoped `settings` store, so they will sync once the KV server backend lands).

Decision: **keep `settings` `account`-scoped as a whole; do not split credentials into a
`device`-only field.** Checked the app's real request path first — model calls all go
`browser → the app's own /api/* routes → provider`, with the key attached per request;
there is no browser-direct path. So the key already reaches the server on every request,
and a `device` split would harden only the at-rest surface at the cost of re-entering
keys on every device.

The exposure that a synced backend *does* add (at-rest, cross-device) is closed instead
by two prerequisites now recorded on #1000, required before any deployment enables
settings sync: real per-user auth + isolation, and encryption at rest as a host
obligation. This keeps the scope call here (`account` for both stores) as-is.

cosarah:
There is still a real data-loss race in legacy adoption. After the confirming `getItem()` returns `null`, another tab/device can write a newer account value before `setItem()` runs; the unconditional adoption write then overwrites it with the stale legacy envelope. I reproduced this by pausing the second read after it captured `null`, publishing a newer value, and resuming migration—the final KV value was the stale legacy copy.

The existing concurrency test only writes before the confirming read, so it misses this window. The migration needs a backend-level conditional put/CAS (or equivalent cross-writer serialization); the adapter-local queue does not coordinate tabs or devices.

wyuc:
Thanks for the careful repro — the race is real and your diagnosis is right: the per-key serialization here is per-adapter-instance and can't coordinate across tabs or devices, so closing this window needs a backend-level conditional put / CAS, not anything the adapter can do alone.

I'm not adding a mitigation in this PR, though. The only backend it wires up is the synchronous localStorage `BrowserKVStore`, where the read-null → write gap is sub-millisecond and not reachable across writers — the window only opens against a networked `account` backend, which isn't part of this PR. An adapter-local half-measure here wouldn't actually close it (cross-writer ordering still needs the server) and would only give readers a false sense of safety.

So I'd rather carry this as a prerequisite on the KV server backend: that backend must expose a conditional put, and legacy adoption must write through it, before it ships — that's the layer where the guarantee can actually hold. This PR only ever runs against the browser backend, where the race can't occur. Does that work as the resolution here?

cosarah:
Deferring server-side CAS to #1000 is reasonable, but the claim that this race is unreachable with `BrowserKVStore` is not. I verified it in real Chrome with two same-origin tabs: tab A read `null`, tab B successfully wrote the newer value, then tab A wrote the stale legacy value; the final `localStorage` value was stale. Synchronous `localStorage` calls are atomic individually, not as a cross-tab read-check-write sequence.

The current risk may be narrow, but it is real. Browser adoption needs cross-tab serialization (for example Web Locks), or this PR should explicitly retain the residual risk rather than resolve it as impossible.

wyuc:
You're right, and thanks for the repro — I was wrong to call it unreachable on the browser backend. The adoption "read → check → write" is an async sequence, so two independent tabs interleave across it even though each `localStorage` call is individually atomic; the window is the whole adoption flow, not a single call.

Given that, we're taking your second option for this PR: explicitly retaining the residual risk rather than resolving it. I've added a "Known limitation" section to the PR description recording the cross-tab adoption race, its blast radius, and the fix path. The reasoning for accepting rather than fixing it here: the affected data is user-reconstructable (settings / profile, including keys the user still holds), the trigger requires two tabs racing the *first* migration of the same key, and the real fix is two orthogonal pieces at two layers — Web Locks for cross-tab serialization in the browser, and the conditional put / CAS already tracked on #1000 for the cross-device case. Both are follow-ups; neither belongs in this browser-backend wiring PR.

So this merges with a documented known limitation, not a claim that the race can't happen. Does that resolution work for you?

cosarah:
That resolution works for the first race as an explicitly accepted limitation. One independent data-loss window still remains, though: `dropShadowedLegacy` reads a raw value and confirms it equals the KV value, then awaits the adoption marker before deleting by key. An old tab can write a newer raw value during that await, and the subsequent `removeItem(name)` deletes the new bytes. I reproduced this sequence deterministically; the newer raw value ended as `null`.

The same risk exists in adoption cleanup before the raw delete. This is not covered by the current Known limitation, and Web Locks in the new bundle would not coordinate an old bundle that never acquires the lock. The raw entry should not be deleted without an atomic byte match, or this separate residual risk also needs to be explicitly addressed.

wyuc:
Update, and thanks — your run of repros is what drove this.

Rather than fix another branch of the legacy-adoption logic, we've removed the automatic migration entirely. Each of your findings — the adoption write race, the `dropShadowedLegacy` delete window, and the cross-tab reachability I wrongly called impossible — plus a resurrection chain two independent reviews then found on top (a reader-device shadow, and an in-memory rebound after an external wipe) were all branches of the same state space, one that kept producing same-class data-loss no matter which branch we closed. That is the signal to question the feature, not patch it again.

So this PR now routes settings / user-profile persistence through `KVStore` for **new** data only; it does not migrate old `localStorage` data. Upgrading users' settings reset to defaults and are reconfigured once — the data is user-reconstructable (provider config, profile), and the raw persist blobs are purged best-effort (a small win: the settings blob held plaintext keys). Net −580 lines: the entire adopt / marker / shadow / resurrection-guard subsystem is gone, and with it every race you and the reviews surfaced — there is no adoption path left to race. The async-storage steady state (hydration/write gates, recovery, health channel) is unchanged.

Both an independent cross-vendor review and a red-team pass came back clean on the removal. A one-time "storage upgraded, please reconfigure" notice is noted in the description as a product follow-up. Ready for another look.

cosarah (commented):
Confirmed as a real defect at `c0010128`: importing the settings/profile stores deletes the pre-cutover raw `localStorage` blobs, while hydration reads only the new namespaced KV entries and never copies the old data first. A legacy-upgrade repro resets `playbackSpeed` from 2 to 1, `modelId` from `gpt-4o` to empty, and nickname from `Ada` to empty. The current staged adoption changes fix this focused repro, but the reviewed commit loses existing users' persisted settings and profile.

wyuc:
You're right about the mechanics — at `c0010128`, upgrading users' settings and profile reset to defaults; the old raw blobs are not copied forward. That is intentional, not an oversight, and I'd rather lay out the tradeoff than treat it as a defect to patch back.

What's affected is bounded and user-reconstructable: the settings store (provider/model config incl. API keys, TTS, playback, layout) and the profile store (nickname, avatar). Course content, learner data (chat/quiz/PBL), generated media, and custom agents live in other stores and are untouched — nothing a user spent real time on is lost.

The reason we chose not to migrate is this thread itself. Every version that carried migration produced same-class data-loss: the delete race and cross-tab reachability you found, plus a resurrection chain the reviews then found on top (a reader-device shadow, an in-memory rebound after an external wipe) — six same-class P1s, each fix surfacing another branch. The state space (device role × path × wipe timing, and especially once a server-backed account backend ships) is large enough that patching one branch just moves the leak. So the tradeoff we landed on: a one-time, reconstructable reconfigure is a smaller and far more predictable cost than migration's recurring, hard-to-bound data-loss.

Put bluntly — the data-loss you're guarding against is one migration is *more* likely to actually cause (its races delete a just-written value or resurrect an ancient one), whereas not migrating is deterministic, predictable, and one-time. A one-time "storage upgraded, please reconfigure" notice (a follow-up in the description) gives users the heads-up.

So I'd ask to record this as an accepted, documented product tradeoff rather than a defect. If you still consider losing existing settings unacceptable, let's talk it through — but after this thread's history, spending a one-time reconfigure to remove an entire class of recurring data-loss looked like the right call.

cosarah (dismissed):
The implementation looks good to merge. One follow-up: please align the README and remaining adoption-related comments with the final no-migration behavior.

cosarah (dismissed):
LGTM

wyuc:
CI is green. The only change since your approval is a patch version bump (@openmaic/storage 0.1.0 -> 0.1.1) that the version-bump guard requires for the README doc-alignment edit — no logic change. GitHub dismissed the approval on that commit; could you re-approve when you get a chance so this can merge?

cosarah (approved):
LGTM

PR #1002, opened 2026-07-28
feat(storage): KV HTTP contract and clients (#1000 Part A)

### LINKED ISSUE #1017: [Task] @openmaic/storage: KV HTTP contract + clients (Part A of #1000)
Part A of #1000 (KV server backend).

Delivers the `KVStore` HTTP contract and clients:

- `docs/kv-http-contract.md` — the wire contract (get/set/remove/keys, `no-store` reads, error table, scope-channel and dot-segment guards, principal derivation server-side).
- `HttpKVStore` / `HttpAccountKV` clients, the `DeviceSafeKVStore` capability brand (a `device` write stays local; composites whose device scope routes to a `LocalKVStore` are accepted, pure account transports are not), and `credentials` passthrough.
- The backend-agnostic KV conformance suite.

Closed by #1002.

Remaining under #1000: Part B (Postgres reference backend) and Part C (mount on the reference server). The `AssetProvider` server backend has moved out of #1000 and is being redesigned under #1007 (global asset pool).

wyuc:
> 🤖 **AI-assisted PR.** This change was written with AI assistance and has been AI code-reviewed (cross-vendor) with findings addressed before requesting maintainer review, per CONTRIBUTING.md.

Closes #1017.

Part A of #1000, scoped to **KVStore only**. Delivers the contract and client
for the `account`-scope HTTP backend of `KVStore`; the browser backend is #858.

## Scope change

The asset half of this PR has been withdrawn. Per the maintainer decision, the
`AssetProvider` server backend is being redesigned around a global resource
pool in #1007 — where `AssetRef` moves from content-addressed to an allocated
global id and the contract model itself changes — so the content-addressed
asset client that lived here would have been work on soon-to-be-replaced code.
Everything asset-only that #1002 introduced is removed; the `BrowserAssetProvider`
is reverted to its opaque #858 behavior (which also resolves the opaque-ref
finding: a ref like `bucket/path` is no longer rejected by a client-side guard).

Targets the `integration/kv-asset-server-backend` integration branch.

## KV HTTP contract

`get` / `set` / `remove` / `keys(prefix?)`, **account scope only**, documented in
[`docs/kv-http-contract.md`](https://github.com/THU-MAIC/OpenMAIC/blob/feat/kv-asset-http-contract/packages/%40openmaic/storage/docs/kv-http-contract.md).

The device-never-leaves-the-device rule is part of the primitive, so the backend
is structurally incapable of shipping a device value rather than configured not
to:

- **The contract has no scope.** No scope path segment, query parameter, header,
  or body field exists, so there is nothing to set to `device`. Servers reject a
  request that invents a scope in any channel with `400 VALIDATION_FAILED`.
- **The transport type has no device scope.** `HttpAccountKV`'s scope parameter
  admits `'account'` alone (type error otherwise) and refuses anything else at
  runtime — necessary because bivariant method parameters let the transport
  stand in for the wider `KVStore`.
- **The composite tells the scopes apart.** `HttpKVStore` routes `device` to a
  local backend it requires at construction (`DeviceSafeKVStore`) and `account`
  to the network; an unknown scope fails closed.

The `kvPersistStorage` device overload requires a `DeviceSafeKVStore` — a store
whose device scope stays local, satisfied by a fully-local backend and by the
`HttpKVStore` composite, but not by the pure `HttpAccountKV` transport, which is
double-rejected (type + runtime).

## Reads bypass caches; not-found requires a 404

KV `GET`s carry `cache: 'no-store'` and the contract requires the server to send
it: `account` is the scope another device may have just written, and neither
side can invalidate a cached copy. The `KEY_NOT_FOUND → null` mapping requires
`status === 404` as well as the code, so a proxy answering 401/403/500 with the
same code is not read as "no such entry".

## Tests

`HttpKVStore` / `HttpAccountKV` run the shared `runKVStoreContract` suite against
an in-memory conformance server (`kv-conformance-server.ts`), plus the
device-scope invariant, key-domain, error-table, cache, and base-URL cases. Both
backends of `KVStore` are proven equivalent by the one shared suite, and the
`@ts-expect-error` device-scope probes run under a package-level
`tsconfig.test.json` in CI so the guards can't rot.

## Slicing

| Part | Scope | State |
| --- | --- | --- |
| A (this PR) | KV HTTP contract + `HttpKVStore` / `HttpAccountKV` clients | here |
| B | KV Postgres reference backend, keyed by `(principal, key)` | next |
| C | Mount the KV contract on the reference server | after B |

`AssetProvider` server work is tracked separately in #1007.

## Open questions

1. **`keys()` is unpaginated**, following `GET /documents`. The primitive is
   sized for small configuration values.


cosarah:
Confirmed as a real defect: `HttpKVStore` is the full two-scope store and safely routes `device` operations to its required `LocalKVStore`, but `kvPersistStorage(httpKVStore, "device")` rejects it at both the overload and runtime guard because the wrapper itself is not branded fully local. This conflates “entirely local” with “can safely serve device scope” and blocks the intended application-wide composite store from persisting device-scoped zustand state. Please allow the safe composite case without marking `HttpKVStore` itself local, and add a regression test asserting the write reaches the injected device store with no HTTP request.

wyuc:
Fixed — you were right, and the fix is exactly the distinction you named. The guard tested "is the whole store local," which turned away `HttpKVStore` even though its `device` scope is safe *because* it routes to the `LocalKVStore` it's constructed with.

A new capability brand `DeviceSafeKVStore` (`servesDeviceScopeLocally: true`) separates "a device write stays on this machine" from "the whole store is local." The persist `device` overload and its runtime guard now check that, not `isLocalKVStore`. So `kvPersistStorage(httpKVStore, 'device')` is now accepted — `device` routes to its injected `LocalKVStore` — while `HttpAccountKV` (no local device backend, `device` would hit the wire) stays doubly refused: at the type layer (`false as const`, unassignable to the required `true`, proven by `@ts-expect-error`) and at the runtime guard.

Regression test as you asked: `kvPersistStorage(httpKVStore, 'device')` — setItem/getItem/removeItem reach the injected `LocalKVStore`, land device-scoped there, and make zero HTTP requests. The composite's injected device backend is still required to be a fully-local `LocalKVStore`, so a router loop with no local floor can't be constructed. Both a cross-vendor review and a red-team pass (with compile-level escape probes) confirmed the widening does not reopen the device-out boundary. Ready for another look.

cosarah:
One contract regression remains: `BrowserAssetProvider.resolve` and `remove` now call `assertAssetRef` before lookup, so an opaque ref such as `bucket/path` throws. This conflicts with the DSL-owned `StorageProvider` contract, where `AssetRef` is an opaque string and missing refs resolve to `null` or remove as a no-op. I reproduced this by seeding IndexedDB under `bucket/path`; both operations passed when only the new guards were removed. Either narrow the DSL contract with migration semantics, or keep read/remove refs opaque and safely encode/derive HTTP storage keys.

wyuc:
Good catch — and rather than adjust the guard, we've taken the asset half out of this PR entirely.

The asset story is being redesigned from scratch in #1007 (a global asset pool: allocated ids over a content-addressed blob store), which changes the `AssetRef` model itself — so hardening the HTTP asset backend here, on code that redesign will replace, wasn't worth it. This PR is now KV-only ("KV HTTP contract and clients"). The `HttpAssetProvider`, the asset HTTP contract, and its conformance/tests are removed, and `BrowserAssetProvider` is reverted to its #858 baseline (0-diff from `main`) — so `resolve`/`remove` treat the ref as opaque again, missing refs resolve to null / no-op, and the `AssetRef`-opaque contract you cited is restored. Your `bucket/path` repro no longer throws.

The asset HTTP backend will come back under #1007's model. KV-only here, CI green.

cosarah (commented):
I found three substantive issues before merge:

1. [P1] The required conditional write is missing. Issue #1000 requires put-if-absent/version-matched writes for legacy adoption, but KVStore exposes only set(), HttpAccountKV sends an unconditional PUT, and the contract explicitly specifies last-writer-wins with no CAS. Please add the wire operation, client API, conflict response, and a two-client stale-writer test.

2. [P2] The new BrowserKVStore key validation breaks existing callers. Stage, scene, and learner IDs remain unconstrained strings and are interpolated directly into current-scene, migration, playback-cursor, and PBL watermark keys. A valid stage ID such as stage/one now throws before localStorage. Preserve the old key domain by encoding dynamic components, or narrow all owning ID contracts with migration coverage.

3. [P2] The account-settings security prerequisites from #1000 are not documented next to account scope. Please state that real per-user isolation and encryption at rest are required before enabling settings sync.

Also, the contract still says kvPersistStorage(device) requires LocalKVStore, while committed HEAD accepts DeviceSafeKVStore. Please align the documentation. The PR also lacks a closing issue link required by CONTRIBUTING.md; since this is only Part A, a scoped sub-issue would avoid closing #1000 prematurely.

wyuc:
Thanks — all five addressed.

**1. Conditional write (CAS) — withdrawn, not implemented.** The requirement was for legacy adoption, and #1001 removed automatic migration entirely (settings/profile persist new data only, no adoption), so nothing performs an empty-check → write against the account backend anymore. I've updated #1000 to mark the CAS prerequisite withdrawn; `account` keeps last-writer-wins by design. A conditional write would only return if a concurrent-edit-resolution need arises, which is out of scope here.

**2. Key validation regression — fixed by making keys opaque again.** You're right: stage/scene/learner IDs are contract-legal bare strings, so `stage/one` is valid and composes real keys. I removed the `/` and `\` rejection from `assertKVKey`/prefix on both backends; traversal defense is now by encoding, not rejection — the HTTP client percent-encodes the key into one path segment (`stage/one` → `stage%2Fone`), the server stores the decoded string as an opaque map key / bound value, never a path, and a raw unencoded `../` yields extra segments that match no route (404). The shared suite round-trips keys with `/ \ % :` on both backends, `editor-current-scene:stage/one` round-trips through the reference server, and the mutation-check reddens on re-adding the rejection.

**3. Account-sync security prerequisites — documented.** Added next to the account scope in the contract: before enabling settings sync, a deployment MUST provide real per-user isolation (the `(principal, key)` partition enforced end to end, not a droppable filter) and encryption at rest; the reference server demonstrates principal derivation, not that posture.

**4. Doc/impl alignment — fixed.** The contract now says `kvPersistStorage(device)` accepts a `DeviceSafeKVStore` (matching HEAD), and the "a key is an identifier, never structure" section is rewritten to "a key is an opaque string."

**5. Closing link / scoped issue.** This PR is Part A of #1000, so I opened #1017 (KV HTTP contract + clients) for it to close — that keeps #1000 open for Part B/C. #1000 itself is updated: the asset backend moved to #1007 (the global asset pool redesign), so #1000 is now effectively the KV server backend.

Ready for another look.

cosarah (commented):
Three P2 issues remain:

1. GET routes do not reject a scope smuggled in the request body. The contract requires scope to be rejected in every channel, but routeKv returns from both GET routes without reading the body. A raw GET /kv/keys with {"scope":"device"} currently returns 200 []. Please reject GET bodies (or inspect and reject scope) and add direct conformance cases.

2. The opaque-key fix still rejects valid keys composed from unconstrained IDs. Stage.id remains an unrestricted string, while the primitive-wide 512-byte limit makes BrowserKVStore reject a previously valid key such as editor-current-scene: plus a 500-character stage ID (521 UTF-8 bytes). Preserve the existing opaque key domain and use a derived fixed-size server-side identifier, or formally constrain every owning ID contract with migration coverage.

3. The prefix contract still contradicts opaque-key semantics: it says a legal key can never contain backslash, while the preceding section and tests explicitly allow it. The shared literal-prefix test covers percent and underscore but not backslash, the named PostgreSQL LIKE escape edge. Correct the sentence and add a backslash prefix case.

wyuc:
All three addressed, plus a parity hardening the review of the fix surfaced.

**1. Scope via a GET body.** The two bodyless methods (GET, DELETE) now reject any request body up front (`assertNoRequestBody`); only `set` reads a body, and it's checked for a `scope` field. `GET /kv/keys` with `{"scope":"device"}` is rejected. Direct conformance cases cover all `{get, set, remove, keys} × {query, header, body}` cells (the body cells use a raw `node:http` client, since `fetch` refuses to put a body on a GET — the exact non-conforming shape).

**2. 512-byte limit rejecting valid keys.** Confirmed stage/scene/learner ids are unconstrained strings, so keys are opaque with no length contract. I raised the shared ceiling to a documented DoS guard (a resource limit, not a key-domain constraint) — and, after review of that change caught that a UTF-8-byte bound diverges from the wire, the bound is now on **encoded request-target size** (`encodeURIComponent(key).length`, cap 4096), applied identically on both backends and the conformance server. That's well above any real id (a 500-char stage id under `editor-current-scene:` encodes to ~523) and far below Node's 431 threshold (16 KiB), so no key that validates can 431, and accept/reject is identical on both backends (verified against a real `node:http` server). Not a hashed/derived key — that would break the literal `keys(prefix)` semantics.

**3. Prefix contract contradiction.** Reworded — a prefix is opaque like a key — and the shared literal-prefix matrix now covers `\` (PostgreSQL's default LIKE escape) alongside `%`, `_`, and combinations.

All three are now CI-executable coverage matrices — scope × channel × op, key forms × both backends, prefix charset — so a missed cell can't slip through again. `@openmaic/storage` is bumped to 0.1.3; CI is green. Ready for another look.

cosarah (commented):
Three substantive issues remain:

1. [P1] The shared 4096 encoded-byte ceiling is a breaking BrowserKVStore domain change and does not guarantee transport parity. Stage.id is still an unrestricted string, so existing longer local keys become unreadable/undeletable. Also, with a 15 KB Authorization header, a short key succeeds while an accepted 4096-character key receives Node 431 because the calculation excludes headers and the base path. Keep HTTP transport limits out of the browser primitive and preserve the existing string key domain.

2. [P2] The scope coverage matrix still omits path scopes and the literal scope header. The contract requires path/query/header/body scope attempts to return 400, but /kv/device/keys returns 404 and GET /kv/keys with scope: device returns 200; only x-scope is checked. Define the prohibited header spelling(s), reject scope path segments, and extend the matrix.

3. [P2] The NUL exclusion relies on the false claim that U+0000 cannot be represented. It round-trips as %00, and Map/Web Storage accept it, so this also makes previously valid BrowserKVStore keys inaccessible.

Non-blocking cleanup: mark the PR as AI-assisted per CONTRIBUTING.md, deduplicate the repeated ceiling/validation and listener lifecycle code, and update the package description that still says HTTP is coming later.

wyuc:
All three addressed — structurally this time. The recurring key-limit findings shared one root cause (transport rules leaking into the key domain), so I stopped tuning the ceiling and removed it.

**1. [P1] Ceiling breaking the browser primitive.** Removed the whole key-domain validator (`assertKVKey` / `MAX_KV_KEY_*`) — the browser primitive is a Map/Storage and validates nothing, so any existing key (long, NUL, separators) reads and deletes again; the key domain is opaque per the contract. HTTP transport limits stay HTTP-side and never constrain the primitive — three of them, each failing loud client-side and each storable on the browser:

- an unpaired UTF-16 surrogate (`encodeURIComponent` throws → `KEY_NOT_ENCODABLE`),
- a whole-key `.`/`..` (URL path normalization would alias it to the empty key / parent → `KEY_NOT_ENCODABLE`),
- an over-long key (deployment request-target ceiling → `431`).

The contract lists all three as HTTP-deployment limits, explicitly not key rules. (The `.`/`..` aliasing was itself caught reviewing this change and fixed before this reply — a whole-key `.` was collapsing to the empty key; a key that merely contains a dot, and a `.` prefix in `keys()`, are unaffected.)

**2. [P2] Scope completeness.** A scope path segment (`/kv/device/keys`) now returns `400`, not a 404 route-miss, and the server rejects the whole prohibited header set (`scope`, `x-scope`, `kv-scope`, `x-kv-scope`), not just `x-scope`. The matrix is now `{get, set, remove, keys} × {path, query, each header spelling, body}`.

**3. [P2] NUL.** Removed — `%00` round-trips and Map/Storage accept it, so the exclusion rested on a false "cannot represent" claim that made valid browser keys inaccessible.

Cleanup (all three): PR marked AI-assisted per CONTRIBUTING.md; the package description updated (KV HTTP is delivered, not "coming later"); the listener bind/skip/close lifecycle deduped into one helper. Removing the validator deleted more than the transport checks added — a net code reduction. `@openmaic/storage` is at 0.1.5, CI green. Ready for another look.

cosarah (approved):
Approved. No blocking issues remain; the latest TypeScript checks, storage tests, and CI checks are green.

PR #1005, opened 2026-07-29
feat(courses): add folder grouping

### LINKED ISSUE #1004: [Feature]: Group courses into user-created folders
### Problem or Motivation

As users create more courses, the home page becomes a single flat list with no way to organize related courses together. There is no mechanism to group courses by topic, project, or any user-defined category. This was part of #34 ("课程重命名与文件夹分类"), but only course renaming shipped — the folder/classification half was never implemented.

### Proposed Solution

Add user-created **folders** to group courses on the home page:

- **Create / rename / delete folders** directly in the course list (same visual weight as course cards).
- **Move a course into or out of a folder** via a hover menu on each course card.
- **Navigate** into a folder to see only its courses, with a breadcrumb back to the full list.
- **Search** collapses the hierarchy into a flat list, annotating each course with its owning folder.

Folders are **device-local organization metadata** — they group the courses already on this device and do not sync across devices or touch the course document itself.

### Acceptance Criteria

- [ ] A user can create a folder from the home page
- [ ] A user can rename a folder inline on its card
- [ ] A user can delete a folder, choosing to either keep its courses (move them out) or delete them along with it
- [ ] A user can move a course into / out of a folder via the course card hover menu
- [ ] Navigating into a folder shows only its courses, with a breadcrumb back to all courses
- [ ] Searching shows all matching courses flat, each annotated with its folder
- [ ] Deleting a course cleans up its folder membership
- [ ] The feature is gated behind a feature flag (default OFF) so existing flows are undisturbed
- [ ] All UI strings are internationalized

### Alternatives Considered

- **Tags/labels instead of folders**: more flexible but heavier UX (filter UI, multi-tag assignment). Single-level folders match the simpler mental model most users expect for "group my courses".
- **Nested folders**: adds navigation complexity for little gain at the current scale. Single-level keeps it simple.

### Area

UI / UX

### Additional Context

This completes the folder-classification half of #34. The implementation keeps folder metadata in the existing IndexedDB database (separate `folders` + course→folder mapping tables), entirely separate from the `@openmaic/storage` document aggregate, so course documents and their version semantics are unaffected.

CXuPercy:
## Summary

Adds user-created **folders** to group courses on the home page, completing the folder-classification half of #34 (course renaming already shipped). Folders are device-local organization metadata — they group the courses on this device and do not sync across devices or touch the course document itself.

Closes #1004.

## Changes

- **New Dexie tables** (`folders` + a course→folder mapping table, schema v17) hold folder metadata and membership. The course document aggregate owned by `@openmaic/storage`'s DocumentStore is **untouched** — folder grouping never enters document/version semantics.
- **Folder CRUD**: create, rename (inline on the card), delete with a choice of keeping the member courses (unfiled) or deleting them along with the folder. Deletion uses progressive disclosure — empty folders get a lightweight inline confirmation; non-empty folders get a compact dropdown with a destructive secondary confirmation.
- **Move a course** between folders (or to unfiled) via drag-and-drop (primary) or a hover/touch-accessible menu (fallback); the menu also supports creating a new folder and moving into it in one step.
- **Navigation**: clicking a folder shows only its courses, with a breadcrumb path in the centered header. The library action bar is always present after hydration (one stable surface across root, folder, and empty states).
- **Search** collapses the hierarchy into a flat list, annotating each matching course with its owning folder.
- **Deletion cascade**: deleting a course cleans up its folder membership row.
- **Import context**: importing a course from inside a folder assigns it to that folder (the destination is captured when the import starts, not when it resolves).
- **i18n**: all UI strings internationalized across 10 locales.

## Type of Change

- [x] New feature

## Verification

**Local checks (all pass):**
- [x] `pnpm check` (Prettier)
- [x] `pnpm lint` (ESLint — 0 errors)
- [x] `npx tsc --noEmit` (0 errors in PR files)
- [x] `pnpm check:i18n-keys` (10 locales aligned)
- [x] `pnpm dev` — home page compiles and renders (HTTP 200)

**Manually verified locally:**
- Create / rename (inline) / delete (both modes) / move (drag + menu) / navigate into a folder / search (flat + folder annotation).
- Stable page layout across create/delete transitions (hero does not expand after first visit).
- Import from inside a folder assigns to that folder.

## Checklist

- [x] Code follows the project style (Prettier + ESLint pass)
- [x] Self-reviewed the diff
- [x] All UI strings are internationalized (`pnpm check:i18n-keys` passes)
- [x] No new warnings introduced


wyuc:
Could you post screenshots of the UI?

CXuPercy:
> Could you post screenshots of the UI?

I have added screenshots to the PR body.

wyuc (commented):
The folder workflow is useful, but I think two changes would make it feel complete:

- Make dragging a course card onto a folder the primary move interaction, with clear drop-target feedback. Keep the current menu as the accessible fallback for keyboard, touch, and precise moves.
- Replace the placeholder-style folder tile with a stable stack of up to three course covers. Use the folder icon only for empty folders, and keep the name and course count visible.

The page already has the course thumbnails and folder membership data, so the cover stack should not require a storage change. Could you include both improvements in this PR?

wyuc (changes requested):
Thanks — the folder workflow and the latest drag-to-file / cover-stack refinements look good in manual testing. I also ran a full-diff cross-vendor review against the current head (`5a7feed9`). Please address the following before merge:

1. **The New Folder action is dead while Recent is collapsed.** The toolbar button remains mounted, but `NewFolderDialog` is inside the `recentOpen` subtree. Clicking the button does nothing, and expanding the section later opens the stale dialog. Mount the dialog outside the collapsible content or explicitly expand before opening it.

2. **Folder management disappears when the global course list is empty.** Empty folders are manageable while other courses still exist; the failing case is specifically `classrooms.length === 0`, because that guard removes the entire Recent/folder surface. This prevents creating a first folder and makes persisted folders unreachable after the last course is deleted.

3. **Rename bypasses the folder-name invariants.** The rename path does not reuse `validateFolderName` or duplicate checks, and `renameFolder` persists unchecked input. It currently accepts empty and over-width names. Please enforce the same rules on create and rename, preferably at the storage boundary as well as in the UI.

4. **A partial “remove folder and courses” failure leaves stale UI.** Member courses are deleted sequentially. If a later deletion fails, earlier deletions are already durable, but the catch path skips `loadFolders()` / `loadClassrooms()`, so deleted course cards and stale counts remain visible. Refresh authoritative state on both success and failure, and cover the partial-failure case.

5. **The current folder path is rendered twice for non-empty folders.** The top section header shows `Recent > Folder`, and the content area renders the same breadcrumb again. Empty folders happen not to duplicate it because they enter the empty-state branch earlier. Please keep one consistent breadcrumb for both empty and non-empty folders; my preference is to keep the top bar as the section title/count/collapse control and use the content breadcrumb for navigation.

Please add focused tests for the new storage/UI contract, especially rename validation, membership joins, both folder-deletion modes, and partial failure. The existing changes mostly update mocks and the database-version assertion, so these regressions are currently uncovered.

wyuc (changes requested):
Thanks for the follow-up fixes. I re-reviewed the latest head (`9f77686f`) and ran a real-browser smoke test. The root state with zero courses and two empty folders renders both folder cards correctly, and the tested validation paths do not persist invalid names. However, the folder navigation and several correctness paths still need changes before this is ready.

1. **[P1] Make the folder view the primary content mode and keep the path in the centered header.** When a folder is opened, the full course-creation hero remains above it, so the Recent bar and empty-folder state are pushed close to the bottom of the viewport and footer. The centered bar still says only `Recent 0`; a non-empty folder gets a separate left-aligned breadcrumb, while an empty folder has no breadcrumb at all. Opening a folder should bring the library content up immediately, show a single centered path such as `Recent > Folder name`, and place the empty state directly below that bar. Clicking `Recent` should return to the root folder grid.

2. **[P2] Clear folder drop state when a drag leaves the card.** `components/discovery/folder-card.tsx:99-102` only clears the highlight when `currentTarget === target`. Because `dragleave` bubbles from the card's children and overlay, dragging away or cancelling can leave one or more folder cards permanently highlighted. Use a drag-enter counter or check whether `relatedTarget` is still contained by the card. Also gate the target on the expected `text/stage-id` payload.

3. **[P2] Use the same folder-name normalization in the dialog and storage layer.** `components/discovery/folder-dialogs.tsx:59-72` checks duplicates case-sensitively, while storage rejects them case-insensitively. Creating `math` when `Math` exists therefore reports the generic creation failure instead of the duplicate-name message. Map `FolderNameError` in the create path or share the same normalization helper.

4. **[P2] Report the actual width for rejected renames.** `app/page.tsx:331-334` hardcodes `{ width: 0, max: 40 }`. A 41-character rename is correctly rejected but visibly reports `Width 0/40`. Compute the submitted width and use `FOLDER_NAME_MAX_WIDTH`; the browser repro should read `Width 41/40`.

5. **[P2] Keep name uniqueness atomic across tabs.** `createFolder` and `renameFolder` perform a read/check/write sequence that can race across tabs, allowing two normalized duplicate names to pass the check before either write completes. Put the check and write in one read-write transaction or enforce uniqueness with a normalized indexed key.

6. **[P2] Keep the move fallback reachable without hover.** The move menu trigger uses `opacity-0 group-hover:opacity-100`; HTML drag-and-drop is unavailable on typical touch devices, and keyboard focus can land on an invisible trigger. Add visible focus treatment and a coarse-pointer/touch affordance.

Additional issues worth addressing in this pass:

- With folders enabled and no folders/courses, the page says "No courses match your search" even when no search is active.
- Several newly added strings remain English in Arabic, Japanese, Korean, Portuguese, and Russian resources, while Traditional Chinese uses Simplified Chinese text.
- A folder with courses but no cached cover can display the empty-folder icon next to a non-zero course count.

Verification performed on this head:

- Real browser: zero courses plus two empty folders, collapse/create flow, blank/over-width/duplicate renames, entering and deleting empty folders, and console inspection.
- Relevant unit tests: 54 passed.
- TypeScript, selected ESLint, Prettier, i18n-key checks, and `git diff --check` passed.
- GitHub CI, E2E, and PostgreSQL contract checks are green.

Requesting changes primarily for the folder-view layout/path blocker and the confirmed correctness/accessibility issues above.

wyuc (changes requested):
Follow-up findings from manual validation of the latest head (`19729038`):

1. **[P1] Do not treat "zero courses with existing folders" as the full-screen empty landing state.** The hero layout currently uses `classrooms.length === 0 && !effectiveCurrentFolderId`, so a root containing persisted empty folders still receives `min-h-[calc(100dvh-8rem)]`. This pushes the Recent bar and folder grid into the next viewport. The full-height hero should apply only when there are zero courses **and** zero folders. Because both collections hydrate asynchronously from IndexedDB, please also wait for storage hydration before selecting the layout, or use a stable shell, so loading the folders does not cause another visible layout shift.

2. **[P1] Remove the course-folder config gate and ship the feature unconditionally.** This is now intended to be a standard course-library capability, not an experimental deployment variant. Please remove `NEXT_PUBLIC_ENABLE_COURSE_FOLDERS`, `COURSE_FOLDERS_ENABLED`, the flag helper/config documentation, and the flag-on/flag-off UI branches. Always load folder metadata and expose create, move, drag, and folder navigation. The existing IndexedDB v16 data must remain intact.

3. **[P2] Keep the centered bar geometrically stable when entering and leaving a folder.** In the root view, the 28 px New Folder action is mounted; inside a folder it is removed, while the folder path is inserted. The control row therefore changes height and its contents visibly shift on both axes. Give the bar a fixed height, reserve a fixed action slot even when the New Folder action is hidden, and separate the centered path from the right-side action cluster so changing path length cannot move the search/import controls. Folder navigation should change the label and available action, not the bar's position.

These are in addition to the previous review. The most visible remaining issue is that the empty-folder root still places real library content below a full viewport of landing-page whitespace.

wyuc (changes requested):
One more empty-state regression found during manual validation of `860a1170`:

1. **[P2] Remove the duplicate floating import controls from the zero-course layout.** `app/page.tsx:910-932` renders a second Import Classroom/PPTX action whenever `classrooms.length === 0`, even though the stable Recent bar already exposes the same actions at `app/page.tsx:1060-1081`. With persisted folders, and especially inside an empty folder, this legacy CTA floats above the navigation bar and changes position as the hero switches layout modes. Keep import in one stable action area. The folder view should not show a global empty-library CTA above its path.

2. **[P1] Keep folder creation reachable for a truly empty library after removing the gate.** The Recent section is now guarded by `(classrooms.length > 0 || folders.length > 0)` at `app/page.tsx:939`, but the New Folder action lives inside that section. A new user with zero courses and zero folders therefore cannot create the first folder, which contradicts the new always-on behavior. Render the library action bar after hydration even when both collections are empty, or provide the same create/import actions in a single stable empty-state action area. Do not retain two independent import surfaces.

The expected invariant is one stable library action bar across root, folder, and empty states. State changes may alter the path and enabled actions, but must not duplicate, hide, or vertically relocate the import control.

CXuPercy:
@wyuc The latest commits (`6142ce5`, `37be471`) address all the items from your last review — single stable library action bar across root/folder/empty, duplicate import CTA removed, empty-library create/folder reachability, hydration-guarded hero layout, and the fixed-height bar.

CI did not trigger on these pushes (the PR went through a draft → ready cycle). Could you approve the workflow runs on the **Checks** tab so the suite picks back up? Thanks!

wyuc (changes requested):
One more folder-context bug found during manual validation of `37be4716`:

**[P1] Courses imported from inside a folder are silently placed at the root.** `useImportClassroom` creates and retains the new `stageId`, but its success callback is invoked without that ID at `lib/import/use-import-classroom.ts:236-239`. The page callback at `app/page.tsx:249-253` can therefore only reload the list; it cannot call `setStageFolder` for the active destination. From the user's perspective, Import Classroom is launched from `Recent > Folder`, succeeds, and then the course disappears from the current view because it was filed as ungrouped.

Please make folder context part of the import contract:

- Import launched from the root should remain ungrouped.
- Import launched from a folder should assign the new course to that folder before the authoritative list refresh, so the card appears immediately and the folder count increments.
- Capture the destination folder when the file picker/import action starts. Do not read whichever folder happens to be active after the asynchronous import finishes.
- If folder assignment fails, surface an explicit error or recovery state. Do not silently fall back to the root.
- Add focused tests for root import, folder import, and navigation changing while an import is in flight.

The broader UX invariant should be that any action which creates a persisted course from within a folder inherits that folder context unless the UI explicitly asks for a different destination.

wyuc (changes requested):
Two additional UX findings from manual validation of `5e12984f`:

1. **[P1] Deleting the last folder must not switch the page back to a full-screen landing layout.** `app/page.tsx:730-732` derives the hero height directly from the live `classrooms.length === 0 && folders.length === 0` condition. When the user deletes the second of two empty folders, that condition flips and the hero expands to `min-h-[calc(100dvh-8rem)]`, pushing the stable library bar and empty hint close to the bottom of the viewport. Create/delete operations must not change the bar's Y position. Keep the post-hydration library layout compact across item-count transitions. If a first-visit landing treatment is still desired, model it as an initial session state rather than recomputing it from mutable collection counts. Please cover the `2 empty folders -> 1 -> 0` transition and assert stable bar geometry.

2. **[P2] Replace the heavy two-card delete dialog with an anchored action menu.** The current dialog at `components/discovery/folder-dialogs.tsx:191-258` uses two large bordered radio cards for a simple choice, and the native radio plus manual top margin does not align reliably with the option title. Clicking the folder delete icon should open a compact menu beside the icon with:
   - `Delete folder only`, which moves courses to Ungrouped.
   - `Remove folder and N courses`, styled as destructive.

   The safe action can execute directly. Only the irreversible action should open a lightweight confirmation explaining that the courses and chat history will be removed. For an empty folder there is no choice to make, so delete directly with a short undo/recovery affordance if practical. The common path should not require a full modal.

The intended result is a stable page layout and progressive disclosure: lightweight controls for routine organization, with additional confirmation only when course data will be destroyed.

CXuPercy:
@wyuc The latest head (`bd9917e3`) addresses the two items from your last review:

1. **[P1] Hero no longer expands on folder delete.** The full-screen landing layout is gated by a session-scoped `librarySeen` flag that latches `true` once the library bar renders, so deleting the last folder (or any create/delete transition) cannot flip the hero back to full-screen.

2. **[P2] Lightweight delete flow.** The heavy two-card modal is gone. Empty folders get an inline confirmation overlay on the card tile (matching the course-delete pattern). Non-empty folders get a compact dropdown beside the delete icon — "Delete folder only" (courses move to unfiled) runs directly; "Delete folder and N courses" opens a small destructive confirmation. No full modal for routine organization.

Also fixed from QA and a post-commit cross-review:
- Breadcrumb count is now contextual (total at root, in-folder count inside a folder).
- Renaming to empty/whitespace shows an error + shake instead of silently exiting.
- Removed 6 dead i18n keys from the old dialog; translated all remaining English folder strings in ja/ko/pt/ru/ar; fixed zh-TW simplified→traditional.
- Delete-confirm overlay closes before the async delete so a failure doesn't leave the card stuck.

TypeScript, ESLint, Prettier, i18n-key alignment, and unit tests all pass locally. Could you approve the workflow runs on the Checks tab if CI hasn't picked up yet?

wyuc (changes requested):
The latest UX now looks good in manual validation. I also completed a final full-diff cross-review of `bd9917e3`. Before approval, please address the remaining integration and correctness issues below.

1. **[P1] Resolve the current `main` conflicts without dropping either migration contract.** `main` already uses Dexie v16 to add the `audioFiles.stageId` index, while this branch independently uses v16 for `folders` and `stageFolders`. Preserve the existing v16 audio migration and introduce the folder tables in a new v17 schema, including the version constant and affected test expectations. The import hook also conflicts with a newer `main` fix: keep the success callback outside the rollback region, while extending it to receive the imported stage ID. A consumer callback failure must not roll back an already committed course and its owned assets.

2. **[P2] Validate folder destinations atomically before writing membership.** `setStageFolder` currently writes any `folderId` unconditionally. If an import starts inside a folder and that folder is deleted before the ZIP finishes, the import writes an orphan membership. This produces inconsistent menu selection and blank folder annotations in search. Validate the destination and coordinate the check/write with folder deletion in one transaction so assignment cannot race deletion. Add a focused race test.

3. **[P2] Disable course dragging while inline rename is active.** The ClassroomCard root remains `draggable` while its descendant name input is being edited. In Chromium/WebKit, click-dragging inside that input can start the card drag instead of selecting text and may move the course if released over a folder. Gate dragging with `!editing` as well as `!confirmingDelete`.

4. **[P2] Do not render the empty-folder visual for a non-empty folder with no thumbnail.** `isEmpty` is derived from `covers.length`, so a folder containing a zero-scene course, or courses whose first-slide thumbnails are unavailable, displays the same icon as a truly empty folder next to a non-zero count. Derive emptiness from `courseCount` and use a distinct no-cover fallback for non-empty folders.

5. **[P2] Make the landing-layout state intentional.** `hydrated` and `librarySeen` are set in the same batched update, so `hydrated && !librarySeen` is never rendered. Either retain the compact layout unconditionally and remove the dead state/comment, or latch `librarySeen` after the initial full-screen state has actually rendered.

6. **[P3] Finish the small validation and localization cleanup.** Submitting whitespace with Enter in New Folder currently reports `Width 0/40` instead of the existing empty-name message. `deleteFolderUngroupDesc` also remains English in ar/ja/ko/pt/ru and uses Simplified Chinese in zh-TW. Please enforce `FOLDER_COUNT_LIMIT` at the storage boundary as well, not only from the dialog prop.

After resolving conflicts, please update the stale PR title/body that still describe a feature flag, then run the full CI suite. The folder-focused tests, TypeScript, Prettier, and i18n key alignment pass locally; GitHub currently shows no checks on this head.

CXuPercy:
@wyuc The latest head (`470f5fea`) addresses all six items from your latest review:

1. **[P1] Dexie v17.** Main's v16 audio migration is preserved; the folder tables (`folders` + `stageFolders`) are now in v17. The import hook is merged: `onSuccess` receives the `importedStageId` and is called outside the rollback region, gated by `importCommitted`.

2. **[P2] setStageFolder atomic validation.** Assignment to a non-undefined folder now checks the folder exists and writes the membership inside one `rw` transaction over `[folders, stageFolders]`, so a concurrent folder deletion cannot leave an orphan membership. Unfiling always succeeds.

3. **[P2] Drag disabled during rename.** `draggable={!confirmingDelete && !editing}`.

4. **[P2] Non-empty folder without covers.** Emptiness is now derived from `courseCount`, not `covers.length`. A non-empty folder with no cached thumbnails shows the folder icon alongside its course-count badge instead of the empty-folder visual.

5. **[P2] librarySeen dead code.** The hero condition is `!librarySeen` (not `hydrated && !librarySeen`). `librarySeen` is initialized `false` (SSR-safe) and read from `sessionStorage` in `useEffect` — no hydration mismatch, no dead state.

6. **[P3] Cleanup.** Whitespace-only submit now shows the empty-name message. `FOLDER_COUNT_LIMIT` is enforced at the storage boundary inside the create transaction. `deleteFolderUngroupDesc` is translated in all locales. Added all folder keys to the new `fr-FR` locale.

Also fixed from a post-commit cross-review: Escape-cancelled drags now clear folder highlights (via a `course-drag-end` event); the `FolderNameError('limit')` race path maps to the specific count-limit message.

The branch is merged with latest `main`. TypeScript, ESLint, Prettier, i18n-key alignment, and unit tests all pass locally. Could you approve the workflow runs on the Checks tab?

wyuc (changes requested):
Thanks for the follow-up. Dexie v17, the import callback boundary, drag-during-rename, validation, localization, and CI are now in good shape. Two blocking issues remain on `470f5fea`, plus one visual follow-up from the previous review.

1. **[P2] The orphan-membership race still exists in `remove` mode.** `setStageFolder` now correctly uses one transaction over `folders` and `stageFolders`, and this serializes with the `ungroup` deletion transaction. However, `deleteFolder(id, 'remove')` reads the member list, deletes courses sequentially, and only then calls `db.folders.delete(id)` outside a transaction that also clears memberships. During that potentially long deletion loop, `setStageFolder` can still observe the folder, write a new membership, and complete before the final folder delete. The folder is then removed while that newly assigned course was not in the original member snapshot, leaving exactly the orphan row this fix intends to prevent.

Please make the folder unavailable for new assignments atomically with capturing and clearing its memberships before the long course-deletion cascade, or introduce a deleting/tombstone state that `setStageFolder` rejects. Add an interleaving test for `remove starts -> assignment commits during member deletion -> remove finishes`.

2. **[P2] Reading `librarySeen` in an effect introduces a refresh layout jump.** The SSR and first client render always use `librarySeen === false`, so an already-seen session first renders the full-screen hero. The effect then reads `sessionStorage` and switches to the compact layout, visibly moving the library bar after hydration. This avoids the React hydration warning but not the geometry regression. Since the accepted UX is the stable compact layout after the library has been introduced, the simplest solution is to remove the first-visit full-screen branch and keep the hero compact unconditionally. Otherwise, defer rendering the layout until the session value is known so the first visible frame is already final.

3. **[P3] The non-empty no-cover fallback is still the empty-folder visual.** The implementation renamed the condition to `hasCovers`, but both zero-course and non-zero-course folders without covers still render the same folder icon; the course-count badge was already always present. The previous review requested a distinct no-cover fallback for a non-empty folder, not just deriving a comment from `courseCount`. Please either render a neutral course-stack placeholder for `courseCount > 0 && !hasCovers`, or explicitly treat the shared icon as an accepted design limitation.

The four GitHub checks are green, and the PR is mergeable but currently behind `main`. Please update the branch after these fixes and let CI rerun.

wyuc (approved):
The latest changes address the remaining folder deletion race, stable layout, and no-cover fallback. I verified the focused storage tests and the updated dev build, and all required checks are green. Approved.

PR #1006, opened 2026-07-29
refactor(pbl-v2): route runtime agents through the shared LLM entry point

### LINKED ISSUE #1003: refactor(pbl-v2): route runtime agents through the shared LLM entry point
## Problem

The PBL v2 runtime is the only part of the tree that reaches the model without going through the shared entry point in `lib/ai/llm.ts`. Five call sites invoke the AI SDK directly:

- `lib/pbl/v2/agents/instructor.ts:1501`, `:1774` (`streamText`)
- `lib/pbl/v2/agents/evaluator.ts:158` (`streamText`)
- `lib/pbl/v2/agents/simulator.ts:281` (`generateText`), `:503` (`streamText`)

Everything else — 21 call sites across 15 files — already goes through `callLLM` / `streamLLM`, and these five are the only exceptions in `lib/` and `app/`. That is why this reads as an oversight rather than a deliberate split: `planner.ts` landed in the same batch, by the same author, on the same day, and it already used `callLLM`.

## What the five sites therefore lose

**1. Usage accounting.** `recordUsageSafe` is called only from `callLLM` (`llm.ts:355`) and `streamLLM` (`llm.ts:399`); `git grep recordUsage -- lib/pbl` returns nothing. So no PBL v2 runtime call is accounted for — not instructor turns, not evaluator passes, not any simulator character line or narrator pass. The comment at `llm.ts:275` —

> Every server-side LLM call funnels through callLLM/streamLLM, so usage is recorded here in one place.

— does not hold today. This is the most consequential gap of the three, because PBL v2 runtime is exactly the high-frequency, long-context, multi-step tool traffic.

**2. The global thinking kill switch.** `getGlobalThinkingConfig()` (`LLM_THINKING_DISABLED`) is read inside the wrappers only, so it has no effect on these five sites.

**3. A single thinking resolution path.** `runtime-thinking.ts` exists precisely because the agents bypass the wrapper, and it is wired into instructor (×2) and evaluator — but into neither simulator site. One config therefore takes three different meanings:

| call site | native adapters (openai / anthropic / google) | OpenAI-compatible providers |
|---|---|---|
| instructor ×2, evaluator | `providerOptions` built from the request's `thinkingConfig` — honored, possibly enabled | `withThinkingDisabled` seeds `thinkingContext` — force-disabled, the config is overridden |
| simulator `streamText` | honored | nothing seeds `thinkingContext`, so the `providers.ts` fetch wrapper injects nothing — the config is silently dropped |
| simulator `generateText` (narrator) | no `providerOptions` at all — dropped | dropped |

This is worth more than an aesthetic complaint because `thinkingConfig` is not only client-supplied: for routed requests it comes from the stage route (`lib/server/resolve-model.ts:114`), i.e. server-side deployment config for the `pbl-v2-runtime:*` stages. Today that configuration is honored or silently discarded depending on which provider family the stage happens to point at, with no log either way.

## There is no capability gap

`callLLM` / `streamLLM` are generic pass-throughs typed straight off the SDK signatures:

```ts
type GenerateTextParams = Parameters<typeof generateText>[0]
type StreamTextParams = Parameters<typeof streamText>[0]
```

Every option the five sites use — `tools` + `stopWhen`, `messages` (including multimodal), `system` / `prompt`, `abortSignal`, `providerOptions`, a caller-supplied `onFinish` — is already supported by pass-through.

## Proposed change

One PR, three parts:

1. `callLLM` records `result.usage`, which on a multi-step tool run (`stopWhen`) reports the final step only; `result.totalUsage` is the aggregate. (`streamLLM` already prefers `event.totalUsage`.) Instructor is exactly such a multi-step run, so this has to land with the migration or the new accounting under-reports.
2. Move the five sites onto `callLLM` / `streamLLM`. This can be done with **no behavioural change**: pass the disabled config through the wrapper's own `thinking` parameter at the sites that force-disable today, and keep their existing `providerOptions` spread — `injectProviderOptions` yields to a caller-set `providerOptions` (`llm.ts:243`). `withThinkingDisabled` then has no callers and goes away.
3. Add a lint guard so `streamText` / `generateText` can only be imported by `lib/ai/llm.ts`. Without it the split simply grows back.

**Deliberately out of scope:** unifying the thinking semantics in the table above — should a stage-route `thinking` be honored uniformly, or is PBL v2 runtime meant to force it off everywhere? That is a product decision, not a refactor, so this PR preserves today's behaviour per call site.

Related question for whoever owns that decision: the stated reason for force-disabling no longer matches the tree. `runtime-thinking.ts` says

> The instructing turn forces `begin_turn` via `tool_choice`, which several providers reject when thinking is on — DeepSeek returns 400 "Thinking mode does not support this tool_choice".

but `begin_turn` now appears only inside that comment, and the only `toolChoice` left in `lib/` or `app/` is `'auto'` in `lib/agent/runtime/stream-fn.ts:253`. The instructing turn passes `tools` + `stopWhen(stepCountIs(...))` and forces nothing. If that incompatibility is genuinely gone, the force-disable can be dropped outright instead of being carried through the wrapper — but I would rather not assume it, in case other providers still hit it for reasons that never made it into the comment.

wyuc:
Closes #1003.

## What this changes

The five PBL v2 runtime call sites invoked the AI SDK directly, while the other 21 call sites in `lib/` and `app/` go through `callLLM` / `streamLLM`. They were the only exceptions in the tree, and they silently opted out of everything the shared entry point applies.

| | before | after |
|---|---|---|
| `instructor.ts` (teaching turn) | `withThinkingDisabled(() => streamText(…))` | `streamLLM(…, 'pbl-v2-instructor', thinkingConfig)` |
| `instructor.ts` (setup follow-up) | same | `streamLLM(…, 'pbl-v2-instructor-setup-followup', thinkingConfig)` |
| `evaluator.ts` | same | `streamLLM(…, \`pbl-v2-evaluator-${kind}\`, thinkingConfig)` |
| `simulator.ts` (character line) | `streamText(…)` | `streamLLM(…, 'pbl-v2-simulator', thinkingConfig)` |
| `simulator.ts` (narrator) | `generateText(…)` | `callLLM(…, 'pbl-v2-simulator-narrator')` |

## Why it matters

**Usage accounting was missing entirely.** `recordUsage` is reached only from the wrappers, so no instructor turn, evaluator pass, simulator line or narrator pass was ever accounted for — the highest-frequency, longest-context traffic in the product. The comment at `llm.ts:275`, that every server-side call funnels through the wrappers, did not hold.

**The `LLM_THINKING_DISABLED` kill switch had no effect** on those five sites, since `getGlobalThinkingConfig()` is read inside the wrappers only.

**One thinking config meant three different things.** `runtime-thinking.ts` existed precisely because the agents bypassed the wrapper, and it was wired into instructor (×2) and evaluator but neither simulator site — so the same `thinkingConfig` was force-disabled, silently dropped, or ignored outright depending on which of the five sites and which provider family you landed on. This is not only a client-supplied value: for routed requests it comes from the stage route (`resolve-model.ts:114`), i.e. deployment config for the `pbl-v2-runtime:*` stages.

## Also in here

`callLLM` recorded `result.usage`, which on a multi-step tool run (`stopWhen`) is the final step alone; `totalUsage` is the aggregate, and the instructor turn is exactly such a run. Without this the new accounting would under-report on the busiest path. `streamLLM` already preferred the aggregate. It also recorded only after validation passed, so with `retries > 0` a rejected attempt — and a result handed back once the retries were exhausted — went unaccounted although both were billed; usage is now recorded per completed attempt. Consequence worth knowing downstream: a retried `callLLM` site now emits one row per attempt rather than one per logical call, and the rows are near-identical in shape, so a naive dedup by (source, model, approximate tokens) would wrongly collapse them. Observed live — a planner retry produced two rows differing only by ~30 tokens.

A lint guard restricts reaching `generateText` / `streamText` to `lib/ai/llm.ts`, so the split cannot grow back. Every import form that reaches the SDK is covered, and each one was checked by feeding it to eslint rather than assumed:

| bypass attempt | blocked by |
|---|---|
| `import { streamText } from 'ai'` | the new `@typescript-eslint/no-restricted-imports` block |
| `import * as ai from 'ai'` | same block — ESLint reports a namespace import when `importNames` is set |
| `await import('ai')` | the new `no-restricted-syntax` block |
| `` await import(`ai`) `` | same, template-literal selector |
| `require('ai')` | the repo-wide `@typescript-eslint/no-require-imports`, already inherited from `eslint-config-next/typescript` |

Two blocks rather than one because of a flat-config trap: rule options are *replaced* per key, not merged. The named/namespace ban uses the `@typescript-eslint` variant so it cannot collide with the base `no-restricted-imports` that `lib/choreography` and `lib/video-export` already configure. The dynamic-import ban has no such escape — `no-restricted-syntax` is the only rule that sees an `ImportExpression`, and five earlier blocks configure that key — so it `ignores` those directories rather than silently replacing their boundaries.

Every ignored directory is still covered, and covered by a rule rather than by an argument:

- `packages/@openmaic/renderer` and `packages/@openmaic/storage` — the same selectors are spread into their own blocks as a shared `AI_SDK_DYNAMIC_IMPORT_BAN`. An earlier revision of this PR excluded them on the reasoning that they are built in isolation against `@openmaic/dsl` with the SDK nowhere in their dependency graph; review showed `void import('ai')` under the renderer source path passing lint, which is why that reasoning was replaced with enforcement.
- `lib/choreography` and `lib/video-export` — their blocks ban **every** `ImportExpression` outright, which subsumes this one.

Both blocks match `**/*.{ts,tsx,js,jsx,mjs,cjs}`, the same set the module-boundary blocks use — an earlier revision matched only `ts,tsx`, which left `app/api/**/route.js` and `scripts/*.mjs` free to import the SDK. `eval/**` and `tests/**` stay exempt (offline harnesses, plus an integration test whose subject *is* the SDK).

`tests/lint-llm-entry-guard.test.ts` pins all of this: every import form × every source extension × nine production paths, plus the exemptions and the shared-rule-key boundaries, run against the real config through the ESLint Node API.

An `eslint-disable` comment still defeats any of it, which is the point — the bypass then has to be written down where a reviewer sees it.

## Behaviour

All five sites now resolve thinking through exactly one path: whatever the request / stage route asked for goes to `streamLLM` / `callLLM`, which resolves provider options and seeds the thinking context. The runtime holds no opinion of its own — `runtime-thinking.ts` and the hand-rolled `resolveThinkingProviderOptions` calls are deleted.

That is a change from before, and it is the point rather than a side effect. An earlier revision of this PR kept the `PBL_V2_TEACHING_THINKING` override and disclosed its edges; keeping it meant the teaching turns still went through a second, private thinking path, which is precisely the thing this PR exists to remove. The edges went with it: an override that silently lost to a caller-set `providerOptions` on native adapters, and a "disabled" that resolved to `effort: medium` on `gpt-5.4-pro` or `thinkingBudget: -1` — *dynamic* thinking — on `gemini-2.5-pro`, whose catalogued capability has no off value.

**What changes at runtime:**

1. **On OpenAI-compatible providers the teaching turns no longer force thinking off.** A thinking-by-default model now thinks on instructor and evaluator turns unless the route says otherwise: slower first token, reasoning tokens on the bill. Measured on DeepSeek V4 Pro, first token goes from ~1.4 s to ~3.0 s. A deployment that wants the old behaviour pins `thinking` off on the `pbl-v2-runtime` stage route — a config line, where a deployment preference belongs, instead of a constant inside the agent that overrides the route.
2. **All five sites now obey `LLM_THINKING_DISABLED`.** The simulator and narrator bypassed the kill switch entirely before.
3. **A stage-route thinking config now reaches OpenAI-compatible providers at every site.** The hand-rolled resolution returned `undefined` for those adapters, so the config was silently dropped; the wrapper seeds the context their fetch wrapper reads.

That includes the director-narrator pass, which was the last holdout: its two callers passed a `thinkingConfig` in and the helper never destructured or forwarded it, so the provider silently fell back to the model default. Caught in review, fixed in 9bbb720 — so *all five* sites now behave identically.

**Why the policy was safe to delete, and what replaces it.** Two things argued for keeping it; neither survived being made explicit.

The DeepSeek 400 is real. Probed against the live API with the instructing turn's exact shape:

| request shape | result |
|---|---|
| thinking on + `tools` + `stopWhen` | streams fine, 66 reasoning tokens |
| thinking on + forced `toolChoice` | 400 — *"Thinking mode does not support this tool_choice"* |

The incompatibility is bound to a **forced** tool choice, and nothing in `lib/` or `app/` forces one (the only `toolChoice` left is `'auto'` in `lib/agent/runtime/stream-fn.ts`). So the override was guarding a scenario that does not exist, at the cost of the unification — and it would have been an unreliable guard anyway, since it only ever reached OpenAI-compatible providers. The hazard is now a comment at the instructor's tools block, where someone about to force a tool choice will actually read it.

The latency cost is real too, and is recorded at the same call site with the numbers — but it is a deployment preference, so it now lives in the stage route rather than in the agent.

## Known limitation this PR does not close

Streamed turns still record nothing when the stream does not run to completion, which is the common case for a learner who navigates away or hits stop: `sse.ts` breaks out of the agent generator on `req.signal`, which cancels the SDK stream.

Measured on ai@6.0.168 with a mock provider:

| path | `onFinish` | `onAbort` | `onStepFinish` |
|---|---|---|---|
| consumer `break`s out of `fullStream` | — | — | — |
| `abortSignal` fires mid multi-step run | — | — | fires, with the completed step's full usage |

So no terminal callback is reliable here, and `streamLLM` records from `onFinish` only. For a single-step abort the tokens genuinely aren't recoverable (the provider stream was cut before the usage chunk), but a multi-step instructor turn aborted after step 1 does drop usage the SDK already knew.

Closing it means recording incrementally from `onStepFinish`, which changes accounting granularity from one row per call to one row per step for all 22 call sites — a shared-infra decision that belongs in its own PR, not smuggled into this refactor. Flagged on #1003 as the follow-up. The evaluator's `case 'error': return` cancels the stream the same way and loses usage for the same reason.

## Verification

Re-run after every commit on the branch.

- `npx tsc --noEmit` — clean.
- `npx eslint` — 0 errors (17 pre-existing warnings in unrelated files).
- `pnpm check` (`prettier . --check`) — clean.
- `pnpm test` — 3303 tests passed, 4 skipped (includes `origin/main` merged in at 964ce79c).
- Usage-log pollution fix checked empirically: ran `minimax-provider`, `anthropic-serialization` and `usage-storage` tests and confirmed the live `data/usage/` line count is unchanged (64 → 64), with the storage tests still green.
- The lint guard was mutation-checked form by form: a probe file using all five bypass shapes produces five errors (see the table above), `lib/ai/llm.ts` stays clean, and injecting `import 'react'` + `import { streamText } from 'ai'` into `lib/choreography` still fails on both pre-existing boundary rules (`no-restricted-imports` *and* `no-restricted-syntax`), confirming neither new block clobbers them.
- `tests/pbl/v2/runtime-llm-entry.test.ts` (was `runtime-thinking.test.ts`) asserts the new invariant at the provider boundary, driving a real evaluator turn through a `MockLanguageModelV3`: an incoming config arrives intact, **no config means no config** — a re-added policy constant fails this one — the global kill switch still applies, nothing leaks past the turn, and the turn is accounted for under its own usage source. That also confirms empirically that the AsyncLocalStorage still propagates into the lazily-issued provider call through `streamLLM` — the property the retired helper's docstring warned about.

## Verified end-to-end against the real DeepSeek API

Driven through the live UI in a real browser (Chrome for Testing over CDP), `deepseek:deepseek-v4-pro` on `api.deepseek.com`, with `MODEL_ROUTES` pinning `pbl-v2-runtime` to that model as a bare string — **no route-level thinking config**, so the only thing that can suppress DeepSeek's default-on thinking is the context seeding the wrappers now do. `DEFAULT_MODEL` was the same model and it was the only configured provider, so nothing could fall back elsewhere.

Streaming, per turn (SSE frames tee'd in the page):

| turn | endpoint | HTTP | SSE frames | first → last token | `error` frames |
|---|---|---|---|---|---|
| instructor greeting | `/api/pbl/v2/open-task` | 200 | 186 | 1 ms → 7 424 ms | 0 |
| instructor reply | `/api/pbl/v2/instructor` | 200 | 215 | 2 335 ms → 10 735 ms | 0 |
| task evaluation | `/api/pbl/v2/evaluate` | 200 | 2 | n/a — JSON-only by design | 0 |
| scenario prep greeting | `/api/pbl/v2/open-task` | 200 | 356 | 0 ms → 6 475 ms | 0 |
| simulator ×3 | `/api/pbl/v2/simulator` | 200 | 66 / 46 / 100 | 5 784 ms → 14 766 ms | 0 |
| wrapup setup turn | `/api/pbl/v2/open-task` | 200 | 102 | → 2 310 ms | 0 |

Both projects ran to completion (task evaluated 85/100; scenario 6/6 with a final evaluation). No `LLM_ERROR`, no `STREAM_ERROR`, no `Thinking mode` / `tool_choice` string, no provider 4xx — the historical DeepSeek 400 did not reappear. No browser console errors, no failed requests.

Accounting: 16 `pbl-v2*` rows in `data/usage/`, all naming the DeepSeek model with non-zero tokens — `pbl-v2-instructor` ×4, `pbl-v2-evaluator-task`, `pbl-v2-evaluator-final`, `pbl-v2-simulator` ×3, `pbl-v2-simulator-narrator` ×4. Before this PR that file would have had none of them.

**Scope note on this run.** The browser pass above predates the deletion of the hardcoded thinking policy, so what it observed about *thinking* (teaching turns at `reasoningTokens: 0`, simulator turns thinking) described the old override and no longer describes this PR. What it establishes still holds, because the policy deletion did not touch either mechanism: the SSE path streams end to end through the wrappers on a real OpenAI-compatible provider, and the five sites produce usage rows where they previously produced none.

The final behaviour — the deployment decides, the runtime holds no opinion — was re-probed against the live DeepSeek API on the current tree, with the instructing turn's shape:

| `pbl-v2-runtime` route | reasoning tokens | first token |
|---|---|---|
| no `thinking` set (the default a deployment gets) | 60 | 2 554 ms |
| `thinking` pinned off | 0 | 493 ms |

That is the whole behavioural story in two rows: nothing in the runtime substitutes a policy, and an operator who wants the old teaching-turn behaviour sets one line on the stage route and gets `reasoningTokens: 0` back.

Not covered: `pbl-v2-instructor-setup-followup` and `pbl-v2-evaluator-milestone`. The follow-up path needs a setup turn that advances *and* a next microtask in the same milestone (`instructor.ts:1675-1712`); the planner gave the wrapup milestone a single microtask, so `runSetupFollowup` never ran. Its `streamLLM` call is structurally identical to the instructing one.

## Review

Cross-reviewed independently by two vendors before this description was written (`codex exec review` on the base diff, a packet-driven adversarial pass, and a read-only Claude reviewer). Four consensus findings came back and are fixed in 2d7913c: the `callLLM` retry accounting hole, the simulator's dropped stage-route config on OpenAI-compatible providers, the undisclosed kill-switch delta, and the "disabled isn't always off" catalog nuance. The abort/early-break accounting gap was also consensus and is documented above rather than half-fixed. Every number quoted in this description was re-measured locally, not taken from the review text.


cosarah (commented):
The simulator narrator still drops the resolved thinking config. Both callers pass thinkingConfig into runDirectorNarratorPass, but the helper omits it when calling callLLM, so the provider sees the global/model default instead of the pbl-v2-runtime:simulator route policy. I reproduced this at the provider boundary: an incoming { mode: "enabled", effort: "low" } arrived as undefined. Please pass it as callLLM’s fourth argument (after undefined retryOptions) and add simulator coverage.

wyuc:
Good catch, and you were right that it belonged in this PR rather than in a follow-up. Fixed in 9bbb720.

`runDirectorNarratorPass` never destructured `thinkingConfig` at all, so the argument its two callers passed was dead on arrival — it now goes to `callLLM` as the fourth argument, with `undefined` for `retryOptions` as you suggested.

Earlier revisions of this PR documented the narrator as pre-existing and deliberately untouched. That defence stopped holding once the hardcoded thinking policy was deleted a couple of commits ago: the invariant became "every PBL v2 site resolves thinking through the shared entry point, and the deployment decides", and the narrator was the one site contradicting it.

Coverage added as asked, driving a real simulator turn against a scenario fixture so both provider calls are exercised in one turn:

- an incoming `{ mode: 'enabled', effort: 'low' }` reaches **both** the streamed character line and the narrator pass;
- neither substitutes a config when the request carries none (a re-added policy constant would fail this);
- both are accounted for under `pbl-v2-simulator` and `pbl-v2-simulator-narrator`.

The assertion that the two calls are exactly `['generate', 'stream']` is there so the test can't silently degrade into covering only the streamed one. Fault-injected to confirm it bites: dropping the new fourth argument again turns that test red.

Full suite 3285 passed, `tsc` / `eslint` / `prettier` clean, CI green on 9bbb720.

cosarah (commented):
Two follow-ups remain:

1. The PR description is stale: it still says the narrator ignores thinkingConfig, but 9bbb720f now forwards it. Please update the description so it matches the final all-five-sites behavior.
2. The lint guard can still be bypassed in production via namespace, dynamic, or require imports (for example, import * as ai from "ai"). Please close those forms too so generateText/streamText truly remain restricted to lib/ai/llm.ts.

The narrator fix itself and its provider-boundary coverage look correct.

wyuc:
Both addressed in ab049c53. Thanks — the second one was a real hole.

**1. Stale description.** Fixed. It now says all five sites behave identically, and names the narrator as the last holdout that 9bbb720 closed rather than describing it as untouched.

**2. The guard.** I checked each form against eslint rather than reasoning about it, and the picture is narrower than your list — two of the three were already closed:

| bypass attempt | before | now |
|---|---|---|
| `import { streamText } from 'ai'` | blocked | blocked |
| `import * as ai from 'ai'` | **blocked** — ESLint reports a namespace import when `importNames` is set (`* import is invalid because 'generateText,streamText' from 'ai' is restricted`) | blocked |
| `require('ai')` | **blocked** — by the repo-wide `@typescript-eslint/no-require-imports` inherited from `eslint-config-next/typescript` | blocked |
| `await import('ai')` | **open** | blocked |
| `` await import(`ai`) `` | **open** | blocked |

So the dynamic form was the actual gap, in both literal and template-literal shapes. It needs `no-restricted-syntax` (the only rule that sees an `ImportExpression`), and five earlier blocks already configure that key — flat config replaces rule options per key rather than merging, so a repo-wide block would have silently dropped those module boundaries. The new block therefore `ignores` those four directories explicitly. That costs nothing in coverage: `lib/choreography` and `lib/video-export` already allowlist their import sources — importing `'ai'` there is an error today, and `video-export` bans `ImportExpression` and `require` outright — and the two `@openmaic` packages are built in isolation against `@openmaic/dsl` with the SDK nowhere in their dependency graph.

Also corrected the comment above the guard, which claimed namespace imports were out of scope because `importNames` cannot see them. That was simply wrong.

Verified with a probe file containing all five shapes: five errors, one per line, `lib/ai/llm.ts` still clean. And injecting `import 'react'` plus an `'ai'` import into `lib/choreography` still produces errors from **both** pre-existing boundary rules, confirming neither new block clobbers them.

One thing I left deliberately: an `eslint-disable` comment still defeats all of it. I think that is the right place to stop — at that point the bypass is written down where a reviewer sees it, which is what the guard is for.

`eslint` 0 errors, `tsc` / `prettier` clean, 3303 tests passing.

cosarah (commented):
Two issues remain:

1. The dynamic-import guard still excludes packages/@openmaic/renderer/** and packages/@openmaic/storage/**. I verified that `void import("ai")` under the renderer source path passes ESLint, while the same probe under lib/ fails. Please extend the guard to these production package paths too.
2. The PR E2E section still describes the deleted thinking policy: zero reasoning tokens for instructor/evaluator, simulator as “deliberately unpoliced,” and the force-disable reaching the provider. Please update that section to match the final deployment-decides behavior.

The runtime changes and focused tests otherwise look correct.

wyuc:
Both fixed in bcdbff05. Reproduced your probe first — `void import('ai')` under `packages/@openmaic/renderer/src/` passed lint, exactly as you said.

**1. The two `@openmaic` packages.** My previous justification for excluding them was that they are built in isolation against `@openmaic/dsl` with the SDK nowhere in their dependency graph. That is an argument for why a bypass would not matter, not a rule that stops one — the same class of reasoning this PR is trying to replace with enforcement, so thanks for not letting it stand.

The two selectors are now a shared `AI_SDK_DYNAMIC_IMPORT_BAN` spread into both package blocks. I did that rather than dropping the directories from the repo-wide block's `ignores`, because those blocks configure `no-restricted-syntax` for their own `@/…` host-path boundary and flat config replaces rule options per key — matching their files would have silently deleted that boundary.

`lib/choreography` and `lib/video-export` stay ignored and stay covered: their blocks ban **every** `ImportExpression` outright, which subsumes this one. That is now a checked claim rather than an asserted one.

Verified with a `void import('ai')` probe in six production paths — both `@openmaic` packages, `choreography`, `video-export`, `lib/pbl`, `app/api` — one error each. `lib/ai/llm.ts` and `tests/` stay clean, and an `@/lib/foo` import under the renderer still fails its own boundary rule, so nothing was clobbered.

**2. The e2e section.** Rewritten. You are right that it described the policy that no longer exists — that browser run predates the deletion. I split what survives from what does not: the streaming and accounting evidence stands, since the policy deletion did not touch either mechanism (same wrapper path, same SSE path), while its thinking observations describe the old override and are gone.

In their place, the final behaviour re-probed against the live DeepSeek API on the current tree, with the instructing turn's shape:

| `pbl-v2-runtime` route | reasoning tokens | first token |
|---|---|---|
| no `thinking` set (deployment default) | 60 | 2 554 ms |
| `thinking` pinned off | 0 | 493 ms |

Two rows, the whole story: the runtime substitutes nothing, and an operator who wants the old teaching-turn behaviour sets one line on the stage route and gets `reasoningTokens: 0` back.

`eslint` 0 errors, `tsc` / `prettier` clean, 3303 tests passing.

cosarah (commented):
One blocking gap remains: both AI SDK lint guards only match **/*.{ts,tsx}. I verified that a valid app/api/*.js file containing both `import { streamText } from "ai"` and `import("ai")` passes ESLint with exit code 0. Please extend both production scopes to js/jsx/mjs/cjs as well as ts/tsx, while keeping the intentional tests, eval, and lib/ai/llm.ts exemptions.

wyuc:
Confirmed and fixed in 118f480d. Your probe reproduced exactly as described — `app/api/**/route.js` with both a named and a dynamic `'ai'` import, exit code 0 — and the repo has real `.mjs` under `scripts/`, so it was not hypothetical. Both blocks now match `**/*.{ts,tsx,js,jsx,mjs,cjs}`, the same set every module-boundary block in that file already used. That `.js` probe now reports 2 errors.

Beyond the fix: this is the third scope hole you have found in this one guard — the two `@openmaic` package directories, now the extension list — and each time I closed it and verified by hand, which is precisely the process that produced the next hole. Hand-verification does not compose, so the matrix is now an executable contract in `tests/lint-llm-entry-guard.test.ts`:

- every import form (named, namespace, dynamic, dynamic-template) × every source extension × nine production paths, including both `@openmaic` packages, `choreography`, `video-export`, `app/api`, `components`, and `scripts`;
- the three deliberate exemptions (`lib/ai/llm.ts`, `eval/**`, `tests/**`) asserted to still pass, so the guard cannot be tightened into uselessness either;
- the module boundaries that share the `no-restricted-syntax` key asserted to still fire, since flat config replaces rule options per key and a careless edit to the guard would silently delete them;
- an eslint-ignored path counts as *not covered*, so a path that quietly stops being linted cannot masquerade as a pass.

It runs the real eslint against the real `eslint.config.mjs` over in-memory text — no fixtures to drift, nothing written to the tree, ~1.4 s.

Fault-injected against both of your earlier findings to prove it bites: narrowing the scope back to `ts,tsx` turns the js/jsx/mjs/cjs rows red, and re-adding the two `@openmaic` packages to `ignores` turns the dynamic-import rows red.

`eslint` 0 errors, `tsc` / `prettier` clean, 3329 tests passing (26 of them the new matrix).

cosarah (approved):
The blocking issues are resolved and the implementation now matches #1003. Two non-blocking follow-ups remain: add `require("ai")` to the lint coverage matrix so the inherited rule is regression-tested, and refresh the stale package-isolation paragraph in the PR description. The render-service lint scope predates this PR and can be handled separately.

wyuc:
Thanks for the review — four rounds and every one of them landed on something real.

Both non-blocking items are handled:

1. **`require('ai')` in the matrix** — #1018 (test-only). It is already rejected in all six source extensions, but by the inherited `@typescript-eslint/no-require-imports` rather than by the guard's own rules, so it was the one covered form with nothing pinning it. Guarded rows assert the property rather than a rule id, so the inherited rule stays a legitimate implementation; exempt rows had to stop asserting "zero errors" since that require ban legitimately fires in `tests/` and `eval/` too, and now assert that neither guard rule fires instead. Fault-injected: turning that rule off turns every require row red.
2. **Stale package-isolation paragraph** — this description is updated. The "built in isolation against @openmaic/dsl" reasoning is gone, replaced by what actually covers each ignored directory: the shared `AI_SDK_DYNAMIC_IMPORT_BAN` spread into both `@openmaic` package blocks, and `choreography` / `video-export` banning every `ImportExpression` outright. It also records the `ts,tsx`-only scope mistake, since that is the kind of thing worth leaving in the record.

`render-service/**` I have left alone as agreed — it is in `globalIgnores` with its own config, and predates this work.

PR #1009, opened 2026-07-29
fix(interactive): load micropip before generated Pyodide imports

### LINKED ISSUE #1008: [Bug]: Generated Pyodide widget fails to initialize when importing micropip
## Bug Description

A classroom generated by OpenMAIC contained an interactive Python widget that installs `requests` with `micropip`.

The generated bootstrap code executes `import micropip` immediately after `loadPyodide()`, without first calling `pyodide.loadPackage('micropip')`. Because `micropip` is included in the Pyodide distribution but is not loaded by default, widget initialization fails with `ModuleNotFoundError`. The Run button remains disabled and the classroom demonstration cannot continue.

The classroom content was generated by OpenMAIC and was not manually edited.

## Steps to Reproduce

1. Deploy OpenMAIC with Docker.
2. Generate a classroom containing an interactive Python code playground that uses the `requests` package.
3. Open the generated interactive scene.
4. Wait for the Pyodide environment to initialize.
5. Observe that the environment fails to load and the Run button remains disabled.

The generated HTML contained this initialization pattern:

```javascript
pyodide = await loadPyodide({
  indexURL: 'https://cdn.jsdelivr.net/pyodide/v0.26.4/full/'
});

await pyodide.runPythonAsync(`
  import micropip
  await micropip.install('requests')
`);
```

This minimal pattern reproduces the failure because `micropip` must be loaded before Python imports it.

## Expected Behavior

Generated Python widgets should load `micropip` before importing it. The Python environment should initialize successfully and enable the Run button.

## Actual Behavior

Initialization fails, the environment displays an error, and the Run button remains disabled.

## Deployment Method

Docker

## OpenMAIC Version

Commit `0cf2a33`

## Affected Area

Playback / presentation (generated interactive classroom content)

## Browser

Chrome 151

## Operating System

Windows

## Relevant Logs

```text
ModuleNotFoundError: No module named 'micropip'

The module 'micropip' is included in the Pyodide distribution, but it is not installed.
You can install it by calling pyodide.loadPackage("micropip").
```

## Additional Context

A prompt-only fix would prevent newly generated widgets from producing this code, but existing saved classrooms would still fail. A narrow runtime fallback in `patchHtmlForIframe` can also recover already-generated classroom content.

This issue was diagnosed and prepared with AI assistance. The reproduction, root cause, and proposed fix were manually verified against a real generated classroom.

likeyeee:
## Summary

Fix generated interactive Python widgets that fail during initialization when their bootstrap code imports `micropip` before loading the package.

The generator prompt now emits the documented package-loading order, while a narrow render-time fallback also recovers already-generated classrooms. This PR was developed with AI assistance, then manually reviewed, refined, and tested by the contributor.

## Related Issues

Fixes #1008

## Changes

- Require `pyodide.loadPackage('micropip')` before generated Python executes `import micropip`
- Repair existing generated HTML only when it imports `micropip` without already loading it
- Leave unrelated HTML, existing package arrays, and chained `loadPyodide()` expressions unchanged
- Add regression coverage for missing, existing, unrelated, and unsupported bootstrap patterns

## Type of Change

- [x] Documentation update

### Steps to reproduce / test

1. Render generated interactive HTML that assigns the result of `await loadPyodide(...)`.
2. Have the generated bootstrap immediately execute `import micropip`.
3. Confirm the iframe patch inserts `await <runtime>.loadPackage('micropip')` before the import.
4. Confirm already-correct and unsupported bootstrap patterns are not modified.

- Reproduced `ModuleNotFoundError: No module named 'micropip'` in a real OpenMAIC-generated classroom
- Confirmed the same saved classroom reports `Python 环境就绪` and enables the Run button after the fix
- AI self-review identified and added guards for package arrays and chained loader expressions
- `pnpm exec vitest run tests/lib/utils/iframe.test.ts` — 11 tests passed
- Targeted Prettier check — passed
- `pnpm lint` — 0 errors (17 pre-existing warnings outside this change)
- `pnpm build` — passed, including the Next.js production build and TypeScript phase

The repository-wide Prettier check reports the existing Windows CRLF checkout as formatting differences across the repository. The three changed files pass a targeted Prettier check; Linux CI completed the authoritative full-repository check successfully.

### Evidence

- [x] Manually tested locally and against the affected generated classroom
- [ ] Screenshots / recordings attached (no visual design change)

## Checklist

### DIFF
--- a/lib/prompts/templates/code-content/system.md
+++ b/lib/prompts/templates/code-content/system.md
@@ -59,6 +59,20 @@ If user code needs packages like numpy, load them during initialization:
 await pyodide.loadPackage(['numpy']);
 ```
 
+`micropip` is included in the Pyodide distribution but is **not loaded by
+default**. Before Python executes `import micropip`, JavaScript MUST load it:
+
+```javascript
+await pyodide.loadPackage('micropip');
+await pyodide.runPythonAsync(`
+    import micropip
+    await micropip.install('package-name')
+`);
+```
+
+Never call `import micropip` first; that aborts initialization with
+`ModuleNotFoundError`.
+
 ### 4. Wait for Pyodide Initialization
 
 - Disable the run button until Pyodide is fully loaded
--- a/lib/utils/iframe.ts
+++ b/lib/utils/iframe.ts
@@ -99,6 +99,118 @@ const ERROR_CAPTURE_SHIM = `<script data-iframe-error-shim>
 </script>`;
 
 /**
+ * Repair a common LLM-authored Pyodide bootstrap bug.
+ *
+ * `micropip` ships with the Pyodide distribution but is not loaded into the
+ * runtime by `loadPyodide()` itself. Generated widgets sometimes immediately
+ * execute `import micropip`, which aborts their whole initialization with
+ * `ModuleNotFoundError`. When that exact mismatch is present, insert the
+ * documented `loadPackage('micropip')` call after the assigned loader promise.
+ *
+ * This is intentionally narrow: unrelated HTML is untouched, already-correct
+ * widgets stay byte-for-byte stable, and pages whose loader result is not
+ * assigned are left for the runtime-error reporter instead of being guessed at.
+ */
+function patchMissingMicropipLoad(html: string): string {
+  if (!/\b(?:import\s+micropip|from\s+micropip\s+import)\b/.test(html)) return html;
+  if (/\.loadPackage\s*\(\s*(?:['"]micropip['"]|\[[^\]]*['"]micropip['"][^\]]*\])/.test(html)) {
+    return html;
+  }
+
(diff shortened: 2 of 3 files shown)


wyuc (changes requested):
Thanks for tracking this down. The prompt change is sufficient for this issue, but I don't think we should add a 127-line source rewriter to the shared iframe rendering path for a narrow generation bug.

The runtime patch scans and rewrites arbitrary generated HTML, which adds disproportionate complexity and still has correctness gaps. For example, the global `loadPackage('micropip')` check does not verify ordering or the associated Pyodide runtime, so a page that imports `micropip` before loading it is left broken as long as a later load call exists.

Please remove the `patchMissingMicropipLoad` runtime fallback and its tests, and keep the prompt guidance that emits the correct package-loading order. Existing affected classrooms can be regenerated or repaired directly rather than permanently expanding the generic renderer.

PR #1016, opened 2026-07-30
chore(release): compute versions with changesets, keep the hardened publish path

### LINKED ISSUE #1012: Adopt changesets for versioning and publishing the @openmaic packages
## Problem

Version numbers on npm are maintained by hand. Nothing derives them, so keeping `@openmaic/dsl`, `@openmaic/storage`, `@openmaic/renderer` and `@openmaic/importer` truthful depends on every contributor remembering to edit a `package.json` in the same change that touches the package source.

That memory failed. `@openmaic/dsl` gained public runtime and storage contracts after 0.4.0 with no version change, and `@openmaic/storage` was never published at all. #998 fixes the resulting state and adds a CI check that refuses a package change without a version increase, but the check only enforces *that* a number moved. It cannot say whether the number moved by the right amount, it cannot update the internal dependency ranges between these packages, and it produces no changelog.

## Proposal

Adopt [changesets](https://github.com/changesets/changesets).

Contributors stop editing version numbers. A change that affects a published package ships with a small file declaring which packages moved and by how much:

```markdown
---
'@openmaic/dsl': minor
'@openmaic/storage': patch
---

Add runtime and storage contracts to the DSL entry points.
```

On merge, a bot opens a "Version Packages" pull request that applies the bumps, updates the dependency ranges between these packages, and writes `CHANGELOG.md`. Merging that pull request releases. The release input stays exactly what it is today, a version bump that landed on `main`, so the publishing restrictions added in #998 continue to apply unchanged.

## Why this shape rather than deriving versions from commit messages

Tools such as release-please infer the bump from Conventional Commits, which this repository already uses, so the contributor friction would be lower. The reason to prefer an explicit declaration is `@openmaic/dsl`: it is the contract other packages and downstream deployments validate against, and whether a change is breaking is a judgement about the contract, not about the commit type. `fix(dsl):` can narrow what an existing document may contain, and `feat(dsl):` can be purely additive. A commit prefix cannot express that; a reviewed line saying `major` can. Changesets also lets one change declare different levels for different packages, which a per-commit scheme cannot.

## Keeping contributors from forgetting

Two mechanisms, both from the changesets project:

- `changeset status --since=origin/main` in CI, which fails a pull request that touches a published package without a changeset. This replaces the version-bump check added in #998.
- [changeset-bot](https://github.com/apps/changeset-bot), which comments on every pull request with the current status and a one-click link to add a changeset through the GitHub web editor. This matters most for drive-by contributors, for whom "clone the repo and run a CLI" is a real drop-off point.

Changes that intentionally release nothing (docs, CI, tests, internal refactors) use `pnpm changeset --empty`, which satisfies the check and records the intent in the diff. Without that escape hatch a strict check becomes a source of false failures, and a gate that blocks legitimate work is a gate that gets disabled.

## Migration notes

- The workspace contains vendored forks (`mathml2omml`, `pptxgenjs`) whose names we do not own. They must be excluded in `.changeset/config.json`, or changesets will try to version and publish them.
- Manual version edits become incorrect once this lands, because the Version Packages pull request owns them. `CONTRIBUTING.md` needs updating in the same change.
- The release-mode gate and the per-package publish loop in `publish-packages.yml` are superseded by `changeset publish`. The publishing restrictions around them are not, and stay: `main`-only environment holding `NPM_TOKEN`, first-parent history check, required green CI for the exact commit, `--ignore-scripts`, the tracked-file check, and the PostgreSQL contract run.
- Best done after #998 merges, so the migration starts from a registry state that already matches the source.

## Open question

Whether to also generate GitHub Releases from the changelogs, or keep the `@openmaic/<name>@<version>` tags as the only release record.

wyuc:
Closes #1012

Version numbers were maintained by hand, and the check added in #998 only enforced that *a* number moved. It could not say whether the number moved by the right amount, could not update the dependency ranges between these packages, and produced no changelog.

changesets is adopted for **version calculation and changelogs only**. Publishing stays on the path #998 landed.

## Why not `changeset publish`

An earlier revision of this pull request used it. It is a blunter instrument than the loop it would replace, and it costs three enforcement properties:

- **the four-package allowlist.** `changeset publish` publishes every non-private workspace package, so the validated set and the published set stop being the same set. A new public `@openmaic/foo` would reach the registry having passed through none of the build, tests, tarball inspection or registry preflight.
- **the per-package assertion.** `changeset status` fails only when a range contains zero changesets. Changing the renderer while declaring only storage passes. Changing the renderer with `--empty` passes. Proven below.
- **a sound registry preflight.** Its own preflight only *warns* when the tree is behind the registry, and a shell replacement built on `npm view <name> version` compares the mutable `latest` tag, orders with `sort -V` rather than semver, and reads any timeout or 5xx as "never published".

So `scripts/check-package-version-bumps.mjs` is back, minus its merge-time half, and the per-package publish loop is back. The only problem changesets was adopted to solve is "humans forget to bump versions", and that is solved at merge time.

### Merge-time gate — `scripts/check-changesets.mjs`

Replaces the version-bump check. It fails unless **every** publishable package changed in the range is named by a changeset added in that same range. Only changesets added in the range count, so one left pending by an earlier pull request cannot stand in.

"Publishable input" keeps the reviewed per-package definition from the deleted diff mode — `test/`, `docs/`, `vitest.config.ts`, `CHANGELOG.md`, and the importer's `src1/`, `SKILL.md`, `index.html` and friends — rather than the repository-wide glob list the earlier revision used. The `files` whitelists differ between these packages, so one shared list would have to be either wrong for someone or too short to be useful.

**The release itself** consumes every pending changeset, so it can never carry one — including on the pull request that carries it, which is where branch protection looks first. It needs an exemption, and that exemption is the most dangerous rule in the file: it is the one path by which a package change can reach `main` with nothing declaring it. Two rounds of review were spent on it, and the rule it ended at is:

- **nothing the range deletes may lose a merged release intent.** Every package a deleted changeset named must either be released here, or be named again by a changeset this range *adds*. Checked first and on its own, so a range that legitimately declares its own change cannot quietly drop somebody else's pending release on the way past.
- **every file the range touches is a `CHANGELOG.md`, a deleted changeset, or a manifest edited the way `changeset version` edits one** — the `version` field, and the range of a dependency on a sibling in this family. Nothing else. A hand-written version bump next to a real source edit does not qualify; neither does a manifest whose `files`, `exports`, `scripts` or third-party dependencies moved, because each of those decides what the tarball contains.

The `dropped` half has a deliberate way out, because "the change it described was reverted" is a real situation: name the package in a changeset of your own, or let the release go out.

### Release path — `scripts/openmaic-packages.mjs`

The allowlist is now an executed assertion, not two `private` lines a future edit can drop. `--assert-set` fails if the publishable workspace packages are not exactly those four names, if an allowlisted package is missing or private, or if an ignored package is not private. It also refuses to guess when `pnpm-workspace.yaml` no longer matches the layout it knows how to enumerate. It reads no dependency, so it runs *before* `pnpm install` — a bad release is rejected without executing third-party code.

Every loop in `publish-packages.yml` derives its package set from `--list`, six sites in all: build, test/typecheck and pack in `validate`, build and the publish loop in `publish`, and the tag loop in `mark`. The set that is built, tested, packed and published is one set rather than six hand-kept copies, and a failing `--list` aborts the step, because a plain assignment takes the exit status of its command substitution.

Publishing stays **sequential in allowlist order, dsl first**, and stops at the first failure. `changeset publish` uses `Promise.all`; with dependents resolving their dsl range against this tree, a concurrent publish could put a dependent on the registry pointing at a dsl version whose own publish had failed.

### Two bugs fixed on the way

**`git diff --quiet -- packages/@openmaic` compared the worktree to the index**, in both jobs, so a build that rewrote a tracked file *and* staged it compared clean. Now `git diff --quiet HEAD --`. This bug is on `main` from #998.

**`mark` required the whole publish job to succeed**, so a publish that got two packages out and then died stranded those two unmarked. It now also runs on `needs.publish.result == 'failure'`; `skipped` is still excluded so a dry run cannot write tags, and the run still fails, because `publish` did. Review round 2 changed *how* it decides what to mark — see below.

### `workspace:*` → `workspace:^`

`pnpm publish` resolves `workspace:*` to an exact pin. The registry shows what that costs:

```
@openmaic/storage@0.1.1  -> "@openmaic/dsl": 0.5.0
@openmaic/renderer@0.0.3 -> "@openmaic/dsl": 0.4.0
@openmaic/importer@0.1.1 -> "@openmaic/dsl": 0.4.0
```

Installing storage alongside renderer today resolves **two copies of the dsl**. For a package that is a schema and a set of validators, that means a document produced against one instance is validated by the other instance's schema.

`storage`, `renderer` and `importer` now declare `workspace:^`, which packs as a caret range — verified by packing storage and reading the tarball's manifest: `"@openmaic/dsl": "^0.5.0"`. `updateInternalDependencies` moves to `minor` to match: a dsl patch stays in range and releases nothing extra, a dsl minor leaves the range and releases the dependents with an updated one.

This necessarily produces one patch release each for renderer and importer, which is what actually ends the split — their *published* ranges only change when they are republished. This pull request carries that changeset.

## What the two review rounds changed

Every item below was found by review or by fault injection, and each is verified in the table further down.

**The enforcement layer is itself repository state.** The post-build tracked-file check only covered `packages/@openmaic`, so build-time code could rewrite the scripts that enforce everything — including the package list the publish loop reads — before the token-bearing steps used them. It is now the whole tree, after `pnpm install` *and* after the build, in both jobs.

**`mark` no longer runs repository code.** It checks out with `persist-credentials: false`, installs nothing, builds nothing, executes no `node`/`npm`/`pnpm`, and takes its package list from `publish`'s receipt artifact rather than from a repository script — while holding the only `contents: write` in the file. Every receipt line is re-validated against an exact pattern before it reaches a git command, and the token reaches exactly one command through an `http.extraheader`.

**`mark` marks only what this run positively published.** The earlier version asked the registry which versions existed and marked those. "The registry has this version" is not "this run published it": another publisher landing that version between the preflight and the push would have been enough to tag this commit for bytes it never produced. The receipt distinguishes `published` from `skipped`; a `skipped` line without a marker is now a warning for a human to place by hand, which is the honest failure.

**The PostgreSQL contract suites are asserted to have run.** The required-variable guard is a `throw` inside each `*.pg.test.ts` module, so it only fires if the module is loaded — and what decides that is `vitest.config.ts`, which never reaches a tarball and which no publishable-input rule covers. `scripts/assert-pg-contract-suites.mjs` reads vitest's own JSON results and requires both files, with passing assertions and none pending.

**`none`, and edited-rather-added changesets, are refused.** `none` is legal changesets input that releases nothing, so it cannot discharge the obligation to release a changed package. A pending changeset that a range *edits* is somebody else's release intent being borrowed.

**E404 is read from a structured npm error, never from stderr text.** `npm view --json` writes `{"error":{"code":"E404"}}` on stdout for a genuinely missing package. A stderr match could only ever catch something else — a proxy or CDN page whose text contains "404 Not Found" — and would turn it into "this package was never published", the most dangerous wrong answer available here.

**`pnpm-workspace.yaml` is no longer parsed at all.** The finding was that a hand-rolled line scanner can accept a file whose YAML meaning is different. `getWorkspacePackages` walks two fixed directories rather than expanding globs, so the only question the file has to answer is whether the layout still matches that walk — and that question is now answered by comparing the file to a pinned constant. It cannot misread the file because it does not read it, and it needs no parser dependency, which is what lets these checks keep running *before* `pnpm install`. Line endings and trailing whitespace are normalised; nothing else is interpreted. A symlinked package directory is now resolved with `statSync` rather than `Dirent.isDirectory()`, which reports `false` for a symlink and would have hidden such a package from every check.

**`publish` declares `contents: read`.** Specifying any permission sets the rest to `none`, and the job checks the repository out. It is still not `write`.

**The documentation matches what CI enforces.** `CONTRIBUTING.md` and `.changeset/README.md` no longer point contributors at `pnpm changeset --empty`: an empty changeset releases no package, so it can never satisfy the per-package gate. Both now say so, and say what to do instead.

## Repository setup this cannot do from a pull request

**[changeset-bot](https://github.com/apps/changeset-bot) should be installed on the repository.** It comments on every pull request with the current changeset status and a one-click link to add one through the GitHub web editor, which matters most for drive-by contributors. Note that it reports changesets' own existential status, not this repository's per-package rule, so it is a discoverability aid rather than the gate.

## Verification

Node 22.22.0, pnpm 10.28.0, against `origin/main` at `f378c25a`. PostgreSQL 16 in a container for the contract suites.

| Check | Result |
| --- | --- |
| `pnpm check` (prettier, whole repo) | pass |
| `pnpm lint` | 15 problems, **0 errors**, 15 warnings (all pre-existing) |
| `pnpm exec tsc --noEmit` | exit 0 |
| `pnpm install --frozen-lockfile` | pass |
| `pnpm test:package-tarballs` | pass |
| YAML parse, both workflows | pass |
| `bash -n` on all 35 `run:` blocks (both workflows) | pass |

Unit tests, `VITEST_MAX_WORKERS=4`, with `PG_CONTRACT_URL` pointing at a real PostgreSQL 16 and `STORAGE_PG_CONTRACT_REQUIRED=1`: dsl **6 files / 190 tests**, storage **16 / 468**, renderer **25 / 242**, importer **11 / 33** — **933 tests, 0 failures**. Without a database, storage runs 14 files / 409 tests and the two contract suites skip themselves, which is why `validate` provides the service and then asserts they ran.

### The keep-list, re-verified as enforced rather than present

A script parses `publish-packages.yml` with a YAML parser — not grep — and asserts each property against the parsed tree. **24 of 24 pass**, including: three jobs with `validate` at `contents: read` only and no environment; `publish` holding `environment: release`, `id-token: write` and `contents: read`; `persist-credentials: false` on all three checkouts, `mark` included; `mark` the sole `contents: write` holder, installing and building nothing, and the only site in the file that pushes a tag; no tag trigger anywhere in `on:`; repository-wide `concurrency` with `cancel-in-progress: false`; the first-parent check in its `grep -cx` form with the SIGPIPE comment explaining why; the green-CI wait keyed on `workflows/ci.yml/runs?head_sha=$GITHUB_SHA&event=push`; `--ignore-scripts` with its `prepublishOnly` reason; the tracked-file check in the `HEAD` form in both jobs and the old form nowhere; and no `changeset publish` in any job.

### The per-package gate, both ways

Injected a change to `packages/@openmaic/dsl/src/index.ts` on top of this branch, whose changeset names storage, renderer and importer but not dsl.

```
$ node scripts/check-changesets.mjs origin/main
Publishable inputs changed: @openmaic/dsl, @openmaic/storage, @openmaic/renderer, @openmaic/importer.
Every @openmaic publishable package change must ship with a changeset that names it:
- @openmaic/dsl: publishable package inputs changed but no changeset added in this range releases it. ...
EXIT=1
```

Adding an **empty** changeset does not buy a pass, and this is exactly what `changeset status` would have missed:

```
$ pnpm exec changeset status --since=origin/main   # the existential check
EXIT=0                                             # it would have passed

$ node scripts/check-changesets.mjs origin/main
- @openmaic/dsl: publishable package inputs changed but no changeset added in this range releases it. ...
EXIT=1
```

Naming dsl at `patch` passes:

```
@openmaic/dsl: declared patch in .changeset/scratch-dsl.md.
@openmaic/storage: declared patch in .changeset/quiet-pandas-agree.md.
@openmaic/renderer: declared patch in .changeset/quiet-pandas-agree.md.
@openmaic/importer: declared patch in .changeset/quiet-pandas-agree.md.
Changeset check passed.
EXIT=0
```

`test/`- and `docs/`-only changes still pass with no changeset at all.

### The Version Packages pull request passes its own CI

Ran `changeset version` on a branch cut from a base holding the changeset — the real shape, since a version branch is always cut from a main that already carries them. It produced storage 0.1.2, renderer 0.0.4, importer 0.1.2, all with `workspace:^` preserved, plus three fresh `CHANGELOG.md` files.

```
$ node scripts/check-changesets.mjs <base holding the changesets>
This range applies 1 changeset(s) and touches nothing but package manifests and changelogs,
so it is a release and needs no changeset of its own.
  released: @openmaic/storage
  released: @openmaic/renderer
  released: @openmaic/importer
EXIT=0
```

This is the `pull_request` path, not only `push`: branch protection blocks the pull request first, so a `push`-only exception would have been too late.

### The release exemption, fault-injected in thirteen shapes

Each row builds a real commit on a throwaway clone and runs the gate against the merge base. **13 of 13 behave as intended.**

| Shape | Expected | Result |
| --- | --- | --- |
| genuine Version Packages PR (deletes the changeset, bumps three versions, writes three changelogs) | pass | exit 0, three `released:` lines |
| release shape that also rewrites the internal dsl range | pass | exit 0 |
| **laundering 1** — real `renderer/src` change + deleted changeset + correct bumps | fail | exit 1, "it also changes `packages/@openmaic/renderer/src/context.tsx`", then the per-package failure for all three |
| **laundering 2** — hand-increased renderer version + deleted changeset, no bumps for the rest | fail | exit 1, "drops the pending release of storage, importer" |
| **laundering 3** — release shape that also adds `src1` to renderer's `files` | fail | exit 1, "it also changes `packages/@openmaic/renderer/package.json`" |
| **laundering 4** — release shape that also adds an internal dependency to dsl | fail | exit 1, same |
| deletes pending changesets and nothing else | fail | exit 1, "its pending release intent would be lost" ×3 |
| deletes somebody else's pending changeset while declaring its own change | fail | exit 1, names the two packages it would have dropped |
| withdrawal re-declared by an added changeset (the documented way out) | pass | exit 0 |
| changed package declared at `none` | fail | exit 1, "declares @openmaic/renderer at `none`, which releases nothing" |
| changed package covered by *editing* a pending changeset | fail | exit 1, "no changeset **added** in this range releases it" |
| ordinary change with a proper changeset | pass | exit 0 |
| `test/`-only change | pass | exit 0, "No publishable @openmaic package input changed" |

The last two laundering rows are the round-3 additions: before them, a range could delete somebody else's pending changeset while declaring its own change and pass, and the exemption accepted *any* edit to a published manifest.

### The PostgreSQL suite assertion

Narrowed `storage/vitest.config.ts` to `exclude: ['test/**/*.pg.test.ts']` — the shape the guard inside the tests cannot catch, because the guard never loads.

```
vitest:  14 files, 409 tests, 409 passed, 0 failed, success=true      <- fully green
assert:  pg-document-store.pg.test.ts did not run. The PostgreSQL contract is unverified…
         pg-runtime-store.pg.test.ts did not run…
EXIT=1
```

On the honest run the same assertion reports `32 assertions passed` and `27 assertions passed`, against a real database.

### The pinned workspace layout

| Injection | Result |
| --- | --- |
| a fourth glob added to `pnpm-workspace.yaml` | **exit 2**, found vs expected printed |
| the same three globs rewritten as an equivalent YAML flow sequence | **exit 2** — a parser would have accepted this while `getWorkspacePackages` kept walking two directories |
| CRLF line endings and trailing whitespace | exit 0 — normalised, so formatting churn is not a false alarm |

### The allowlist assertion

`--list` on the clean tree emits exactly four lines — the four packages and their directories, nothing else.

| Injection | Result |
| --- | --- |
| `"private": true` removed from the vendored `pptxgenjs` | **exit 1** from both `--assert-set` and `--list`, naming it twice: once because the workspace would publish a package the release path does not know about, once because `.changeset/config.json` ignores it while its manifest is no longer private |
| a new public `@openmaic/scratchpkg` in the workspace | **exit 1** — same first message, naming it |

The failing `--list` also aborts the workflow step: `list="$(… --list)"` takes the exit status of its command substitution, so `set -e` fires before any loop body runs.

### The registry preflight, four directions

| Situation | Result |
| --- | --- |
| a name that has never been published (definitive E404) | exit 0 — "not on the registry yet, 1.0.0 is the first release" |
| registry unreachable (ECONNREFUSED) | **exit 2** — "Registry error for @openmaic/dsl: ECONNREFUSED", *not* treated as unpublished |
| storage at 0.0.9 against published 0.1.1 | **exit 1** — "not greater than the published 0.1.1; refusing to release from a stale tree" |
| storage at 0.2.0 against published 0.1.1 | exit 0 — "releasing 0.2.0 (published: 0.1.1)" |

### The integrity check, widened

Two separate bugs, both fixed. Appending to the tracked, build-generated `packages/@openmaic/renderer/fonts.css` and staging it, which is what a build would do:

```
git diff --quiet --        => CLEAN  (the rewrite is invisible)
git diff --quiet HEAD --   => DIRTY  (caught)
```

And the scope. A build-time write to `scripts/openmaic-packages.mjs` — the file that holds the allowlist the publish loop reads:

```
git diff --quiet HEAD                        => DIRTY  (caught)
git diff --quiet HEAD -- packages/@openmaic  => CLEAN  (this was the gap)
```

## Accepted, with reasoning

**`publish` rebuilds instead of publishing the bytes `validate` packed.** Handing tarballs between jobs would move the question rather than remove it: the job holding `NPM_TOKEN` would publish bytes produced by a job that does not hold it, and the artifact store becomes something to trust. What is guaranteed instead is that both jobs build the same commit with the same frozen lockfile, that `validate` proved that build packs and imports cleanly, and that the tracked-file check proves the build rewrote nothing tracked. The residual risk is a non-reproducible build, which is a property of the packages rather than of this workflow. Noted in the file.

Four gaps are known and accepted. Each is a decision about how the packages are built or released rather than something the merge-time check can settle, and each is written down in `scripts/check-changesets.mjs` and `.changeset/README.md` as well as here.

**1. Root-level build inputs change tarballs with no changed package directory.** A lockfile or toolchain bump rewrites renderer's and importer's Rollup bundles, and dsl's `dist` and shipped JSON Schema are whatever the lockfile's TypeScript and `ts-json-schema-generator` emit. Closing this means treating the lockfile and toolchain as an input of every package — every dependency bump would then demand four changesets — or externalising the bundled dependencies. Both are decisions about how the packages are built.

**2. A later `files` whitelist change can turn a currently-ignored directory into published content.** The per-package ignore lists record what each package does not ship *today*: `test/`, `docs/`, `vitest.config.ts`, the importer's `src1/`. Adding one of those to a package's `files` would start shipping it while the check still treats it as free. Nothing detects that on its own, so the lists have to be revisited whenever a `files` whitelist grows.

**3. `workspace:^` deduplicates within a version line only.** Publishing the family with caret ranges means a consumer on one dsl line installs one copy. A consumer that mixes an older dependent needing `^0.5.0` with a newer one needing `^0.6.0` still resolves two dsl copies — a smaller version of exactly the problem `workspace:*` caused. Only a coordinated major across the family, or making dsl a peer dependency, would prevent it.

**4. Whether a breaking dsl change is declared `minor` rather than `patch` stays a human judgement.** The gate enforces that a changed package is declared and that the level releases something. It cannot read a diff and decide what it does to consumers. `@openmaic/dsl` is where this matters, and it is a review question.

## Observed, not changed: `@openmaic/importer` inlines `@openmaic/dsl`

Reported because it interacts with the `workspace:*` → `workspace:^` fix above, not because this pull request touches it.

`packages/@openmaic/importer/rollup.config.js` has no `external` list, and one shared config produces all three outputs: `dist/index.umd.js` (umd, global `pptxtojsonPro`), `dist/index.cjs` and `dist/index.js`. So dsl is bundled into every one of them. Verified on the built artifacts: the only `@openmaic/dsl` occurrences in any of the three files are dsl's own error-message string literals, minified in place, and there is no `import` or `require` edge to dsl left anywhere — the remaining `require(`/`import(` hits are pdf.js internals.

The package's `exports` map exposes only `import` → `dist/index.js` and `require` → `dist/index.cjs`. `dist/index.umd.js` is not in the map, and an `exports` field without a `./dist/*` subpath blocks deep imports, so that file is reachable only by a CDN (unpkg, jsdelivr) that serves raw paths, or by a bundler configured to ignore `exports`. It is ~1.7 MB of a ~5 MB `dist`.

The cost of the current state is not size, it is identity. `@openmaic/dsl` is declared as a real runtime dependency (`workspace:^`), so a consumer of the importer downloads it anyway; a consumer using both packages therefore runs two copies of `normalizeSlideWith` and the schema, one of them minified inside the importer bundle. That is the same class of problem as the two dsl copies the caret-range fix addresses, one layer down.

**Is a per-output `external` feasible — dsl external for ESM and CJS, still inlined for UMD?** Partly, and the CJS half is the problem.

- *Feasible mechanically.* `external` is a per-config Rollup option and `createConfig` already takes the output per call, so it is a parameter, not a restructure. Keeping UMD self-contained is right: a `<script>` consumer has no module resolution.
- *The ESM output is straightforward.* dsl is already a declared dependency and already appears in `dist/index.d.ts`, so the type surface does not change and no new install-graph edge appears.
- *The CJS output is blocked today.* `@openmaic/dsl` is `"type": "module"` and its `exports` map has only an `"import"` condition — no `"require"`. A CJS bundle emitting `require('@openmaic/dsl')` gets `ERR_PACKAGE_PATH_NOT_EXPORTED`, and Node 22's `require(esm)` support does not help, because the failure is in `exports` resolution rather than in module format. Externalising dsl from `dist/index.cjs` therefore requires giving dsl a `require` condition — a dual build, or at least a CJS entry — which is a change to dsl's published shape and needs its own release and its own review.
- *It is also a behavioural change, not a packaging one.* Today the importer's behaviour is frozen to the dsl it was built against. Externalised, the caret range becomes load-bearing at runtime: a consumer resolving a different dsl minor within `^` changes what the importer does. That is the intended semantics of the range, but it would want a `minor` changeset and a deliberate decision.

So: ESM alone is cheap and closes most of the duplication; ESM + CJS needs a dsl dual build first. Not implemented here.

## Decisions for a maintainer

**1. Who opens the "Version Packages" pull request.** This deliberately does not add [`changesets/action`](https://github.com/changesets/action). That action needs a job holding `contents: write` and `pull-requests: write` that also runs `pnpm install`, and this repository's postinstall builds every workspace package — a write-scoped job executing repository and third-party code is exactly what the `validate`/`publish`/`mark` split was built to avoid, and not a change to make as a side effect of adopting changesets. Until it is added, a maintainer runs `pnpm changeset version` and opens the pull request; the release itself is still automatic once it merges. I can add it as a follow-up, hardened with `pnpm install --frozen-lockfile --ignore-scripts` so no package code runs in the write-scoped job.

**2. GitHub Releases.** The open question in #1012 is untouched: `@openmaic/<name>@<version>` tags remain the only release record. `@changesets/changelog-github` would add pull request links to the generated changelogs, at the cost of a dependency and a `GITHUB_TOKEN` requirement when running `changeset version`. Not adopted.


wyuc:
Closing this without merging. The evaluation is the useful output, so here it is.

**What was being solved.** Registry versions had stopped representing the source tree. That is fixed and merged: the gate in `scripts/check-package-version-bumps.mjs` refuses a change to a publishable file that does not carry a version increase, on every pull request and every push to `main`. Changesets was proposed on top of that, to compute the versions rather than have contributors edit them by hand.

**Why it is not worth it here.** Adopting it required a new enforcement layer of roughly 870 lines across three scripts, and three rounds of independent cross-vendor review each found ways through it.

The root cause is structural rather than a series of bugs. `changeset version` consumes the very files that declare intent, which creates a legitimate commit that changes package manifests while carrying no declaration. That commit has to be exempted, and the exemption is the most intricate logic in the whole change. It was shown to be buyable three separate times, in three different shapes: deleting someone else's pending changeset while declaring your own; pairing a real source change with a deleted changeset and a hand-written version; and rewriting `files`, `exports` or dependencies through the one path that needs no changeset at all. Each fix closed a shape and exposed another.

The existing gate needs no exemption, because "did the version increase" is monotone: there is no legitimate commit that changes a package without bumping it. Adopting the tool broke a clean invariant and then spent 470 lines trying to reconstruct it.

**What the three benefits are actually worth here:**

- *Version calculation.* Already enforced by the merged gate. Hand-editing plus a machine check is working.
- *Changelogs.* This repository already maintains a curated `CHANGELOG.md` in Keep a Changelog form, grouped by theme, with pull request links and contributor credit, alongside GitHub Releases. Generated per-package changelogs would be a second, mechanical set sitting beside a better one, and the internal dependency cascade would fill three of them with `Updated dependencies` entries.
- *Internal dependency ranges.* A one-time correction, not something that needs a tool. It is being sent separately.

**What is being kept.** Three defects surfaced during this work that exist on `main` today and have nothing to do with changesets. They are going out as a small separate pull request: internal dependencies pin exactly, so installing two of these packages yields two copies of `@openmaic/dsl`; the build-integrity check compares against the index rather than `HEAD`, so a staged rewrite passes it; and the PostgreSQL contract suites can be silenced by excluding them from the vitest config, since the guard lives inside the test files.

Two further findings about release retries and a registry race are being filed as their own issue.

PR #1019, opened 2026-07-30
fix(release): dedupe the published dsl, close two release-path blind spots

wyuc:
Three defects on the release path, plus the compensating guards two rounds of cross-vendor review showed each of them needed. Everything here exists on `main` today.

## Merging this publishes four packages

**`@openmaic/dsl@0.5.1`, `@openmaic/storage@0.1.2`, `@openmaic/renderer@0.0.4`, `@openmaic/importer@0.1.2`.**

For the three dependents that republish *is the fix*, not a side effect. The corrected dependency range only exists inside a tarball, and no tarball carrying it has been cut yet, so every consumer install keeps getting two copies of the dsl until those three versions are on the registry.

The dsl patch is smaller: documenting the new format-version release rule next to the constants is an edit under `packages/@openmaic/dsl`, which `check-package-version-bumps.mjs` correctly treats as a publishable change.

---

## 1. Internal dependency ranges pin exactly, so consumers get two copies of the DSL

`@openmaic/storage`, `@openmaic/renderer` and `@openmaic/importer` all declared `"@openmaic/dsl": "workspace:*"`. pnpm publishes that form as an **exact pin**. What is on the registry right now:

| package | published dsl range |
| --- | --- |
| `@openmaic/storage@0.1.1` | `0.5.0` |
| `@openmaic/renderer@0.0.3` | `0.4.0` |
| `@openmaic/importer@0.1.1` | `0.4.0` |

```
$ npm view @openmaic/renderer dependencies.@openmaic/dsl
0.4.0
```

Installing storage together with renderer therefore resolves two separate copies of `@openmaic/dsl`. The dsl carries the schema, the validators and the version constants, so two copies mean a document produced against one instance can be validated by the other instance's schema revision, and neither side can notice.

All three now declare `workspace:^`, which publishes as `^0.5.1`. Under 0.x that admits `0.5.x`, so a dsl patch reaches dependents without republishing them while a format move requires a deliberate release (see 5).

Verified by packing:

```
openmaic-storage-0.1.2.tgz:  @openmaic/storage@0.1.2  -> "@openmaic/dsl": "^0.5.1"
openmaic-renderer-0.0.4.tgz: @openmaic/renderer@0.0.4 -> "@openmaic/dsl": "^0.5.1"
openmaic-importer-0.1.2.tgz: @openmaic/importer@0.1.2 -> "@openmaic/dsl": "^0.5.1"
```

**A caret in `dependencies` is only half of the rule.** `peerDependencies` and `optionalDependencies` are published constraints too, so an exact entry in either reintroduces the duplicate while `dependencies` still reads correctly. An owned package must therefore be declared **exactly once, in `dependencies`**; entries in the peer and optional fields are rejected outright. Both checks enforce this: `scripts/check-internal-dependency-ranges.mjs` on the source at review time, `scripts/smoke-test-package-tarballs.mjs` on the packed manifest at release time.

The checks also name the three dependents that must be present rather than counting declarations, so a dependent quietly dropping its dsl dependency fails instead of passing on the survivors.

### Known limitation, deliberately left open

`^` deduplicates **within one version line only**. A consumer that mixes an older dependent requiring `^0.5.0` with a newer one requiring `^0.6.0` still installs two dsl lines. Removing that entirely would mean making the dsl a peer dependency of all three dependents, which changes their public installation contract — every consumer would then have to install the dsl themselves. Out of scope here. What keeps the caret meaningful in the meantime is rule 5.

## 2. The build-integrity check compared against the index, not HEAD

The publish workflow guards against the build rewriting tracked, publishable files (renderer's generated `fonts.css` and its KaTeX font snapshot). It ran `git diff --quiet -- packages/@openmaic`, which compares the worktree to the **index**. The steps before it run `pnpm install` and the package builds — third-party code, which needs no credential to rewrite a tracked file and `git add` it. After that the guard reported clean and the rewritten bytes were packed and published.

Both copies now use `git diff --quiet HEAD --`. The same check also runs in `ci.yml` after its install step, so a stale generated file is a pull request failure rather than a release failure.

```
old form:  git diff --quiet -- packages/@openmaic       -> exit 0, reports CLEAN
new form:  git diff --quiet HEAD -- packages/@openmaic  -> exit 1, reports DIRTY
restored:                                               -> exit 0
```

## 3. The PostgreSQL contract suites could silently stop running

`@openmaic/storage`'s `*.pg.test.ts` suites skip themselves without `PG_CONTRACT_URL` and throw when `STORAGE_PG_CONTRACT_REQUIRED=1` demands them. Both workflows supply those variables, so the suites do run today.

The weakness is *where that guard lives*. It is a `throw` inside each **test module**, so it can only fire if the module is loaded — and whether it is loaded is decided by `packages/@openmaic/storage/vitest.config.ts`. That file **and the whole `test/` directory** are on the ignore list of publishable inputs, so the entire surface those suites live on can be rewritten with no version bump and no release gate noticing.

An assertion over vitest's JSON results alone does not close this. `assertionResults` are test **cases**, not `expect()` calls, so a file named `pg-document-store.pg.test.ts` containing nothing but `test('x', () => {})` satisfies it. And `test/setup.ts` is already wired as `setupFiles`, so one `vi.mock('pg', ...)` there redirects both suites at an in-memory fake.

`scripts/assert-pg-contract-suites.mjs` therefore runs in two phases:

- **Phase 1** — the two suite files appear in vitest's results, passed, with at least one passing case and none pending.
- **Phase 2** — connect to `PG_CONTRACT_URL` from **outside** the vitest process and require all five tables the two backends own (`document_stages`, `document_scenes`, `document_outlines`, `runtime_sessions`, `runtime_records`) to have **gained inserts during the run**, measured as a delta against a baseline captured before the vitest step. An absolute `> 0` would be satisfied forever by an earlier run against a non-ephemeral database; both workflows use a fresh per-job service, but the check should not depend on that staying true.

Insert counters rather than surviving rows, because the suites clean up after themselves.

### Threat model, stated plainly

**What this proves:** during this run, rows were inserted into those five tables in a real PostgreSQL, and two files with the contract suites' names ran and passed.

**What it does not prove:** that the built `PgDocumentStore` and `PgRuntimeStore` were the code that inserted them. `test/` is on the publishable-input ignore list, so test code could create the schema and insert directly, and this audit would read the same either way. Closing that needs a harness outside the ignored `test/` surface — separate work, not attempted here.

That limit is acceptable because of who each threat is. This guard exists to catch **accidental** silencing: a vitest `include`/`exclude` change, a missing environment variable, a renamed suite file, a dropped workflow step. Those are how this coverage actually disappears, and they are all caught. It is not a defence against someone deliberately faking coverage from inside `test/` — and it does not need to be, since that person can merge changes to the production sources just as easily. The success message says exactly this rather than claiming attribution.

### Fault injections

**(a) `vi.mock('pg', ...)` in `test/setup.ts`, backed by the in-memory PGlite that is already a devDependency**, against a pristine database:

```
vitest:  Test Files  1 failed | 15 passed (16)
              Tests  2 failed | 466 passed (468)
         pg-document-store.pg.test.ts (32 tests)  all passed against the fake
```

466 of 468 pass against a fake database, the document contract entirely. Only the two genuinely-concurrent runtime cases fail, because the shim shares one connection.

**(b) Both suite bodies replaced by `test('placeholder', () => { expect(1).toBe(1) })`** — everything green:

```
vitest:                        Test Files  16 passed (16)
                                    Tests  411 passed (411)   exit 0
check-package-version-bumps:   Package version check passed.  exit 0   (test/ is ignored)
assert-pg-contract-suites:
  phase 1  pg-document-store.pg.test.ts: 1 test cases ran and passed.   <- would have passed
           pg-runtime-store.pg.test.ts:  1 test cases ran and passed.   <- would have passed
  phase 2  document_stages does not exist in the contract database ...
           (all five tables)                                   exit 1   <- caught
```

**(c) Phase 2 in isolation** — the genuinely all-green JSON from a real run, pointed at an untouched database: phase 1 reports 32 and 27 cases passed, phase 2 exits 1 on all five tables.

**(d) Pre-seeded database, no new inserts** — the delta check, against a database that already carries a previous run's counters:

```
baseline: document_stages=24 document_scenes=51 document_outlines=24
          runtime_sessions=36 runtime_records=30
audit:    document_stages exists but gained no inserts during this run
          (before 24, after 24) ...                            exit 1

the old absolute rule would have seen: inserts=24 > 0 -> PASS  (all five)
```

Restored, against a real PostgreSQL 16:

```
pg-document-store.pg.test.ts: 32 test cases ran and passed.
pg-runtime-store.pg.test.ts:  27 test cases ran and passed.
document_stages:   24 inserts during this run (0 -> 24).
document_scenes:   51 inserts during this run (0 -> 51).
document_outlines: 24 inserts during this run (0 -> 24).
runtime_sessions:  36 inserts during this run (0 -> 36).
runtime_records:   30 inserts during this run (0 -> 30).
```

## 4. Validation packed with lifecycle scripts; publishing does not

`pnpm publish --ignore-scripts` is what the publish step runs, but the tarball smoke test and the workflow's dry-run pack both packed **with** lifecycle scripts. So they inspected a different tarball than the registry receives: committing `workspace:*` alongside a `prepack` that rewrites it to `workspace:^` satisfied the range assertion and then published the exact pin, because `prepack` never runs at publish time.

Both now pass `--config.ignore-scripts=true`.

```
pack WITH lifecycle scripts (old):    packed range "^0.5.1"  <- assertion would PASS
pack --config.ignore-scripts (new):   packed range "0.5.1"   <- what npm actually receives

check-internal-dependency-ranges.mjs  exit 1   (caught at pull-request time)
smoke-test-package-tarballs.mjs       exit 1   (caught at release time)
```

## 5. `workspace:^` unpins the serialized-format version that the exact pin was silently holding

`DSL_VERSION` and `RUNTIME_DSL_VERSION` are **serialized-format** versions, explicitly decoupled from the npm version — `version.ts` says a format break bumps them "regardless of where the package version happens to be". Storage compares them by value across the package boundary and throws.

Under the old exact pin, a published `storage@0.1.x` was welded to one dsl build and therefore to one `DSL_VERSION`. Under a caret it floats. So dsl ships `0.5.1` moving `DSL_VERSION` `'0.1.0'` to `'0.2.0'` with a migration, which its own docs call legitimate. Installation A resolves 0.5.1 and `storage@0.1.2` stamps `dslVersion: '0.2.0'` into `document_stages`. Installation B, pinned to dsl 0.5.0, reads that row and hard-fails as "newer than this client's 0.1.0". Two installs of the *same* published storage version, silently data-incompatible.

**The rule, enforced in the merge-time gate:** changing either constant requires a dsl version increase that **the dependents' caret range will not admit**, computed from the pre-change version:

- while dsl is 0.x, `^0.5.1` admits `0.5.x`, so a **MINOR**;
- once dsl reaches 1.0.0, `^1.4.2` admits minors as well, so a **MAJOR**.

Stated as "escapes the caret" rather than as a fixed level on purpose: a rule that said MINOR would silently stop being sufficient at the 1.0 boundary, exactly when the package matured. The error message and the docs in `version.ts` use the same wording.

**The caret to escape comes from the highest dsl version on the base branch, not from the merge base.** An ordinary minor landing on main while a branch is open moves the reference: with merge base 0.5.1, base tip 0.6.0 and HEAD 0.6.1, escaping `^0.5.1` needs only 0.6.0 — but dependents released against 0.6.0 publish `^0.6.0`, which admits 0.6.1, and the format reaches them anyway. 0.6.1 is also the minimum the ordinary version check allows, so that was the default outcome rather than an unlucky one. Whether the format moved *at all* still comes from the merge base, since that is a question about the branch.

**Fails closed.** The gate resolves the constants against an ordered list of candidate paths, and reports an error if none of them declares the constants at either revision while dsl has any publishable change. A list rather than a single path so a rename stays landable: prepend the new path in the same change and the comparison still happens across the move. It resolves *every* candidate rather than the first that exists, because during a half-finished rename both files can be present with the still-exported one carrying the new values — two declaring candidates is an error. The regex is anchored to the start of a line, so a commented-out declaration cannot be read as the value, and a duplicate declaration within one file is an error rather than a coin flip.

Neither reference is applied to the ordinary version comparison, where the base tip alone is correct — the merge base there would let a branch reuse a version another branch published meanwhile. Note that `mergeBaseWithHead` is currently a no-op in both CI invocations: `actions/checkout@v4` checks out the merge ref on `pull_request`, and the before-SHA is already an ancestor on `push`. It is used anyway because it is the right reference for the question, and because hand-run local invocations are not on a merge ref. That is recorded in the code rather than left as an unexercised claim.

Fault injections:

```
merge base dsl 0.5.1 / base tip 0.6.0 / HEAD 0.6.1 with a format change
  escape from merge base 0.5.1 = 0.6.0 -> 0.6.1 >= it? YES, would PASS (the bug)
  escape from base tip   0.6.0 = 0.7.0 -> 0.6.1 >= it? NO, correctly FAILS   exit 1
same shape bumped to 0.7.0                                              exit 0

two candidate sources both declaring the constants
  "... both declare the serialized-format constants, so this check cannot
   tell which one is authoritative"                                     exit 1
rename finished, old file dropped, minor bump                           exit 0

rename version.ts + DSL_VERSION change + patch bump, gate not updated
  "none of packages/@openmaic/dsl/src/version.ts exists at HEAD ...
   This check must not pass by default."                                exit 1
same, candidate list updated, still a patch bump
  "needs an increase that the dependents caret range will NOT admit:
   here at least 0.6.0 (a MINOR) ... ^0.5.1 admits everything below"    exit 1
same, candidate list updated, MINOR bump                                exit 0

dsl already at 1.4.2, format change shipped as a MINOR (1.5.0)
  "here at least 2.0.0 (a MAJOR) ... ^1.4.2 admits everything below"    exit 1
same, shipped as a MAJOR (2.0.0)                                        exit 0

commented-out decoy `// export const DSL_VERSION = '9.9.9'`
  old unanchored regex reads: "9.9.9"   <- the decoy
  anchored regex reads:       ["0.1.0"] <- the live value only          exit 0
duplicate declaration
  "declares DSL_VERSION 2 times; this check cannot tell which one
   is authoritative"                                                    exit 1

un-rebased branch whose base tip carries someone else's format change
  vs base tip  : DSL_VERSION 0.4.0 -> 0.1.0, pkg 0.7.0 -> 0.5.1  (backwards)
  vs merge base: DSL_VERSION 0.1.0 -> 0.1.0, rule correctly silent      exit 0
```

## 6. The range and build-integrity guards gave no pull-request feedback

Both ran only in the publish workflow, so a pull request restoring `workspace:*` or leaving a stale generated file passed CI and failed at release time, after a version number had been spent. `ci.yml` now runs the cheap source-level range check and the `git diff --quiet HEAD` integrity check; the tarball assertion stays where it is as the release-time proof of actual published behaviour.

## One package list

Six hardcoded copies of the owned-package list meant a package missing from one of them was silently exempt from that check. `scripts/openmaic-packages.mjs` is now the single source, consumed by all three checking scripts, and validated in both diff and release mode of the version gate — release mode being the one that decides what gets published.

The workflow's own YAML lists cannot import JavaScript, so they are cross-checked from the outside. Textually, but not naively: full-line comments are stripped first (a commented-out trigger path is exactly how a package's release goes quiet while a substring search stays green), the `on.push.paths` list and every `for pkg in ...` loop are compared as exact **sets** so unexpected entries fail too, and `--filter` usage is checked loosely because individual steps legitimately filter subsets. What that cannot distinguish — a deliberate subset from an accidental one — is stated where the check lives.

`INTERNAL_DEPENDENTS` is cross-checked in both directions, since deleting an entry would otherwise exempt a package from the source check *and* from the packed-manifest assertion that iterates the same map. An owned package in another owned package's `devDependencies` is rejected outright: it satisfies the workspace link while the tarball declares no dependency at all.

```
commented-out trigger path `# - "packages/@openmaic/importer/package.json"`
  new: "on.push.paths omits importer"                                   exit 1
  old substring form: includes(...) = true                           -> PASS
publish loop dropping a package                                         exit 1
publish loop naming an unknown package                                  exit 1
new package on disk, missing from the list                              exit 1
package in the list, missing from publish-packages.yml                  exit 1
INTERNAL_DEPENDENTS entry deleted, dsl still in dependencies
  "declares the owned dependency @openmaic/dsl but is absent from
   INTERNAL_DEPENDENTS ... so nothing checks how it is published"       exit 1
same, with the dsl declaration moved to devDependencies
  "belongs in `dependencies` and nowhere else"                          exit 1
a dependent silently drops its dsl dependency                           exit 1
exact peerDependencies entry alongside the caret                        exit 1 (both checks)
exact optionalDependencies entry alongside the caret                    exit 1 (both checks)
```

### Threat model for these gates

Stated in `openmaic-packages.mjs`: this list and the checks built on it are configuration living beside the code they check, so anyone who can merge a change to `scripts/` can shorten the list or relax a rule. No amount of further checking closes that, since every new check has its own editable configuration. It is out of scope because it is not the threat: these gates catch **mistakes** — a package added in one place and forgotten in another, a range that publishes differently from how it reads, a suite that stops running — which are silent by default and expensive to find later. Someone willing to edit these scripts to hide a change can edit the production sources directly; code review, not tooling, is what stands in the way of that.

## Verification

Node 22.22.0, `VITEST_MAX_WORKERS=4`.

| check | result |
| --- | --- |
| `pnpm check` (prettier, whole repo) | all matched files pass |
| `pnpm lint` | 0 errors, 15 pre-existing warnings |
| `pnpm exec tsc --noEmit` | exit 0 |
| `pnpm install --frozen-lockfile` | exit 0 |
| `@openmaic/dsl` tests | 6 files, 190 tests passed |
| `@openmaic/storage` tests (real PostgreSQL 16) | 16 files, 468 tests passed, 0 skipped |
| `@openmaic/renderer` tests | 25 files, 242 tests passed |
| `@openmaic/importer` tests | 11 files, 33 tests passed |
| `pnpm test:package-tarballs` | exit 0, packed ranges and fields asserted |
| `check-internal-dependency-ranges.mjs` | exit 0, 3 expected dependents |
| `check-package-version-bumps.mjs origin/main` | exit 0 |
| YAML parse of all three workflows | all parse |

The only lockfile change is the three `specifier:` lines moving from `workspace:*` to `workspace:^`.

## Known and separately tracked, not fixed here

**Validation and publication operate on different tarballs.** The publish job reinstalls and rebuilds, and `dist` is gitignored so the tracked-file check cannot see it — the bytes `validate` inspected are not literally the bytes published. This is a pre-existing property of the release path on `main`, not something this pull request introduces. The real fix is architectural: pack once, hand the `.tgz` files to the publishing job as artifacts, publish those exact files. Filed as #1021 along with two other release-path findings (retry and registry-race behaviour).

**Attribution of the PostgreSQL writes**, as described under 3.

**Making the dsl a peer dependency**, as described under 1.

## Out of scope

No changesets adoption. No restructuring of the publish jobs. `@openmaic/importer`'s rollup configuration is untouched — it inlines the dsl, which is a separate interface decision.


wyuc:
## Review summary

This went through three rounds of independent cross-vendor review of the full branch diff, framed adversarially as "enumerate how to defeat these guards" rather than "confirm they work". The final round returned zero P1 findings from one reviewer, and the other's remaining P1s all reduced to a single class addressed below.

**Round 1** found that the tarball assertion and the publish step ran different lifecycle semantics (validation packed with scripts enabled while publishing uses `--ignore-scripts`, so a `prepack` hook could make validation see a corrected range that publication never produces), and that the PostgreSQL guard treated vitest's `assertionResults` as assertions when they are test cases, so a `test('placeholder', () => {})` in each required filename satisfied it.

Round 1 also surfaced the finding that changed this pull request's shape: moving to `workspace:^` unpins the *serialized-format* version, which the exact pin had been holding silently. `DSL_VERSION` and `RUNTIME_DSL_VERSION` are deliberately decoupled from the npm version, and storage compares them by value across the package boundary. Without a compensating rule, dsl could ship a patch that moves the format, and the same published `storage` version would stamp a newer format in one installation and reject it in another. That is why this change also adds a rule requiring a format-version change to carry a dsl bump the dependents' caret range will not admit.

**Round 2** showed that rule was still defeatable: the constants were located at one hard-coded path, so renaming the file switched the rule off silently, and the regex would read a commented-out declaration as the value. It also found that an owned package declared in `peerDependencies` or `optionalDependencies` bypassed both range checks while still publishing a second, exact constraint. Making the rule fail closed initially made a rename unlandable, which is the shape of false failure that gets a guard disabled, so the path became an ordered candidate list resolved per revision.

**Round 3** found the caret escape was computed from the merge base rather than from the highest version reachable on the base branch. With merge base at dsl 0.5.1, main at 0.6.0 and the branch at 0.6.1 plus a format change, the rule accepted the minimum bump the version check allows, while dependents released against 0.6.0 publish `^0.6.0` and admit 0.6.1. That was the default outcome, not an unlucky one. It is currently masked because `actions/checkout` gives a merge ref on pull requests, so the merge base equals the base SHA in both CI invocations; the comment now says so rather than claiming a protection that is not exercised.

Every fix was fault-injected in both directions before and after, and the numbers are in the commit messages.

## Limits, stated rather than papered over

**The PostgreSQL audit proves inserts happened during the run, not that the production backends performed them.** The entire `test/` directory is on the ignore list of publishable inputs, so test code can create the schema and insert directly. Proving attribution needs a harness outside that surface, which is separate work. The guard does catch what it was written for: a vitest config exclusion, a missing environment variable, or a renamed file, each of which silences the suites while the run stays green.

**These guards catch mistakes, not a determined editor of the guards.** Every in-repo check has configuration that whoever can merge a change to it can also edit, and adding another check only moves the question. Someone able to merge edits to `scripts/` can edit production code directly.

**`^` deduplicates within one version line.** A consumer mixing a dependent that requires `^0.5.0` with one that requires `^0.6.0` can still resolve two dsl copies. Closing that entirely means making dsl a peer dependency, which changes the install contract for every consumer and is deliberately out of scope.

**Validation and publication build separately**, and `dist` is gitignored, so the published tarball is not the one that was smoke-tested. This is pre-existing on `main` rather than introduced here, and the fix is architectural: pack once and publish those exact artifacts. Filed separately with two related release-path findings.

## On merging

This publishes `@openmaic/dsl@0.5.1`, `@openmaic/storage@0.1.2`, `@openmaic/renderer@0.0.4` and `@openmaic/importer@0.1.2`. The republish is the fix rather than a side effect: the corrected dependency range only reaches consumers inside a tarball that has not been cut yet. The dsl release is documentation-only, and the version gate required it.

xuyuanwei678 (dismissed):

xuyuanwei678:
Thanks a lot for the thorough release-path hardening here. I took another look at the current state, and the remaining blocker appears to be the merge conflicts from the package metadata/version bumps on the updated main branch. Once those conflicts are resolved and CI is green again, this looks good to merge from my side. Thanks again for the careful write-up and the extra safeguards.

Issue #1007, opened 2026-07-29
[RFC] @openmaic/storage: a global asset pool — allocated asset ids over a content-addressed blob store


wyuc:
## Goal

Generated media (images, audio, video) is referenced today by *context-relative
placeholder strings*, and its bytes are stored raw in the browser with no storage
abstraction. A DSL document is therefore not self-contained: a reference only
resolves in the presence of the stage it was authored under, the bytes cannot be
persisted anywhere but IndexedDB, and every export path re-implements "which media
does this course use?" on its own.

This RFC promotes generated media into a first-class subsystem of
`@openmaic/storage`: a **global asset pool** with a stable, allocated `asset_id`
layered over a content-addressed blob store. In the browser the blob table is embedded
in the registry's own database, because reference counting, byte writes and reclamation
have to share one IndexedDB transaction with the registry write that authorizes them.
The primitives of the `AssetProvider` shipped for #779 (#858) — `sha256` hashing and
object-URL caching — are absorbed as internal parts of the registry, and the standalone
provider is removed rather than kept as a seam nothing can be substituted through. A
content-addressed HTTP asset contract was drafted alongside it in #1002 but was
withdrawn from that PR during review and never reached `main`; the asset server backend
is now tracked here, to return under this RFC's model.

## Motivation — where we are today

**References are logical placeholder strings, not asset identities.** DSL documents
carry per-medium placeholder strings, not self-contained references:

- `PPTImageElement.src` (`packages/@openmaic/dsl/src/slides.ts:334`) and
  `PPTVideoElement.src` / `PPTVideoElement.mediaRef`
  (`packages/@openmaic/dsl/src/slides.ts:731-732`) hold values shaped like
  `gen_img_1` / `gen_vid_1`, recognised by the placeholder predicate
  `isMediaPlaceholder` (`lib/store/media-generation.ts:67`).
- Audio is referenced by `SpeechAction.audioId`
  (`packages/@openmaic/dsl/src/action.ts:48`) — a separate, action-local id space.

**Locating the bytes depends on ambient context.** A placeholder alone is not enough
to find bytes; the store key is the *current* `stageId` composed with the placeholder:
the Dexie compound primary key is `` `${stageId}:${elementId}` `` (built by
`mediaFileKey`, `lib/utils/database.ts:237-239`; declared at
`lib/utils/database.ts:168`). The compound key was itself introduced to fix
cross-course collisions, which is the tell: because the reference is
[...]

PR #1024, opened 2026-07-31
feat(storage): asset registry — allocated ids over the content-addressed blob layer (#1007 part 1)

wyuc:
Implements part 1 of the delivery plan in #1007: the asset registry — allocated opaque ids layered over the content-addressed blob store.

## What lands

- **`AssetId`** (`ast_` + 128 bits, Crockford-style base32): allocated, opaque, fixed-width ids with **zero validation on the id domain** — `toAssetId` brands without checking, and there is deliberately no `isAssetId`/`assertAssetId` (per #1007 OQ1).
- **`BrowserAssetStore`** — the registry over the blob layer, one IndexedDB database with `assets` (`asset_id → {contentHash, mime, meta}`) and `blobs` (`contentHash → bytes`) stores:
  - `put` always allocates a new id and, on the success path, discloses nothing about whether the bytes already existed; identical bytes de-duplicate to one stored blob internally.
  - `resolve` verifies the registry entry's content identity (hash + MIME) on every call, so cross-instance removals and replacements converge without any side channel; unknown/foreign ids are misses, never errors.
  - `replace` regenerates the bytes behind a stable id atomically (entry update + old-hash refcount + conditional reclamation in one transaction) — the operation the allocated-id model exists to support.
  - `remove` releases the entry and reclaims bytes only when the last referencing entry is gone, with the refcount read inside the deleting transaction.
  - **Snapshot semantics** for object URLs: a resolved URL is an immutable snapshot, valid until `release(ref)` or `close()`; mutations affect future resolves only. (An earlier design revoked URLs proactively across instances; it produced a family of revoke-vs-delivery races and was removed in favor of this model — see the commit history.)
- **`BrowserAssetProvider` demoted to the internal blob backend** (removed from the package entry point; content-addressed refs are no longer an outward contract), with a branded `ContentHash` and a property-typed `BlobStore` interface so the blob layer is not assignable to the DSL `StorageProvider` slot under `strictFunctionTypes`.
- **DSL contract text** (`@openmaic/dsl` `storage.ts`, prose only — no signature changes): `put` allocates and does not disclose pre-existence; the content-addressed-refs recommendation is gone; metadata values must be structured-cloneable.

## Test coverage highlights

- Backend-agnostic contract suite (`runAssetStoreContract`) that a hash-embedding backend cannot pass: ids are checked against every common encoding of the payload's digest, id width is content-independent, and the existence-non-disclosure matrix compares a dedupe hit against a first store facet by facet with pinned expectations.
- Lifecycle pins: mint-after-commit, single-flight resolution, cross-instance remove/replace convergence, snapshot liveness across macrotask boundaries, `close()` reclaiming late-settling mints, connection close, retention bounds.
- Key invariants were mutation-checked during review (dropping the refcount guard, returning an existing id on dedupe, minting inside the transaction, disabling the liveness check, and others each fail specific tests).

## Notes for review

- The browser registry stores **no principal column** — with no trusted session to derive one from, a client-supplied principal would record a claim, not a boundary. The column arrives with the server backend (#1007 delivery plan part 4).
- Quota/storage-accounting channels can still disclose existence on the error path; the docs scope the non-disclosure guarantee to the success path, matching the RFC's amended wording.
- `@openmaic/dsl` takes a minor bump (behavioural contract change for implementers), `@openmaic/storage` a minor bump (entry-point export removed).

Part 2 (DSL reference unification: `src` / `mediaRef` / `audioId` → asset ids, `audioUrl` removal) builds on this.


cosarah:
This is a real upgrade break. `BrowserAssetStore` now opens a different DB by default (`maic-asset-pool`) and still only creates the v1 `assets` + `blobs` schema, so existing `BrowserAssetProvider` data in `maic-assets` has no migration/read-through path. After upgrading, previously stored `AssetRef`s can no longer be resolved.

wyuc:
Good flag — worth forcing explicit. It is deliberate, and it's now documented in the diff: 15e8d0fc adds an "Upgrading from 0.1.x" section to the storage README. The evidence that no migration path is warranted:

- **In-repo, nothing ever wrote to `maic-assets`.** `BrowserAssetProvider` has zero consumers outside the package's own tests — the app's storage seam is still the no-op factory with zero call sites. That's the greenfield premise #1007 states ("the blob layer already exists but has zero consumers"), re-verified by grep on this branch during review.
- **Published surface is ~48 hours old.** `@openmaic/storage` 0.1.0 first hit npm on 2026-07-30; the npm downloads API doesn't yet report the package at all. There is no realistic population of persisted 0.1.x data to strand.
- **The 0.2.0 minor is the deliberate signal** for the entry-point removal (0.x semver).
- **A silent read-through would be a design regression, not a kindness**: it would accept `sha256-`-shaped strings as outward references — exactly the model this release removes — which is why the contract suite pins sha256-shaped refs as misses.

If a real 0.1.x consumer materializes, the supported shape is an explicit one-time import helper (enumerate the old rows, allocate an id per blob, return an old-ref → new-id mapping for the caller to apply to its documents) — that's in the README note too, and I'll build it the moment there's a consumer to build it for.

cosarah:
I agree the default `maic-assets` -> `maic-asset-pool` break can be intentional and documented. That explanation does not cover the custom `dbName` case, though.

If a 0.1.x caller used `new BrowserAssetProvider({ dbName: "custom-assets" })` and upgrades to `new BrowserAssetStore({ dbName: "custom-assets" })`, the new store opens the existing database at version 1. Because the old provider also created version 1, `onupgradeneeded` does not run. The database still has only the old `assets` object store, with no `blobs` store and no `by-content-hash` index.

I reproduced this with `fake-indexeddb`:

```text
legacy db version/stores: 1 [ "assets" ] legacyRef: sha256-...
put failed: NotFoundError: No objectStore named blobs in this database
resolve legacy ref failed: NotFoundError: No objectStore named blobs in this database
```

So this is not asking for silent read-through of `sha256-` refs. The issue is that reusing a legacy custom DB name makes even new `BrowserAssetStore.put(...)` fail hard, rather than treating old refs as misses or failing with an explicit legacy-schema error. A fix could explicitly detect this old schema and either reject with an actionable upgrade error, delete/recreate only under an explicit reset path, or later route users to the one-time import helper.

wyuc:
You're right — the custom-dbName collision was a hard, unactionable crash, and that's a real defect regardless of the migration stance. Fixed in b83a7a0f, taking your first option (explicit detection, actionable error; no destructive reset, no auto-migration):

- `BrowserAssetStore.openDb` now validates the schema after opening (both object stores + the `by-content-hash` index). On mismatch it closes the connection and rejects every operation with: *IndexedDB database "{dbName}" has an incompatible asset schema. It may have been created by the 0.1.x content-addressed BrowserAssetProvider or another incompatible version. Use a fresh dbName, or wait for or request the explicit one-time import helper.* The failed open is not memoized, so a corrected environment can retry.
- The guard is mirrored in the internal blob provider for the reverse collision — opening a registry-pool database there would have silently written hash-keyed rows into the registry's `assets` store, which is worse than crashing.
- Tests cover your exact repro (legacy v1 db with only `assets` → all five ops reject with the actionable message, not `NotFoundError`), the reverse direction, fresh-name operation, and retry-after-correction. README upgrade note extended.

CI is re-running on the push.

packages/@openmaic/storage/test/asset-contract.ts
```diff
@@ -14,61 +19,304 @@ const bytes = (s: string): Uint8Array => new TextEncoder().encode(s);
...
+      id = await store.put(blob(content));
+    });
+    const ref = id ?? '';
+    const resolved = await bytesAt(store, ref);
+    const observation: PutObservation = {
+      putOutcome,
+      idType: typeof id,
+      idPrefix: ref.slice(0, 4),
+      idLength: ref.length,
+      idCollidesWithEarlierId: seen.has(ref),
```
cosarah, line 148: The non-disclosure matrix does not observe the returned ID body, so it cannot enforce the contract it claims to police. I verified this with a deliberately non-conforming `StorageProvider` that returned `ast_0…` when the payload already existed and `ast_1…` when it was new, while preserving the prefix, width, uniqueness, resolve/remove behavior, and avoiding digest substrings. The complete `runAssetStoreContract` suite still passed: 28/28 tests.

That is a real conformance hole because the DSL contract says every value returned by a successful `put` must be independent of prior existence. The current `BrowserAssetStore` appears safe—`newAssetId()` uses independent random bytes and `put()` does not query blob existence—but a future backend can expose an existence oracle in the opaque ID body without this central suite noticing it.

Please make this red-capable using deterministic allocator/backend instrumentation that verifies ID generation is independent of the storage lookup outcome; comparing only prefix, length, type, and collision status is insufficient.

wyuc: Confirmed and fixed in 04fbd1f2, using the deterministic-instrumentation shape you asked for:

- `BrowserAssetStoreOptions` gains an `idAllocator` seam (default `newAssetId`); the store draws ids from it and from nothing else, and its docstring states the independence requirement.
- `runAssetStoreContract` now REQUIRES an allocator-instrumented factory (`{makeStore, makeStoreWithAllocator}`) and adds an allocation-independence section: a scripted deterministic allocator drives an interleaved fresh/dedupe schedule — put(fresh A), put(dup A), put(fresh B), put(dup B), put(dup A), put(fresh C) — and the suite asserts the returned ids equal the allocator's outputs byte-for-byte in call order, with exactly one draw verified after each put (not merely in aggregate), and that `replace` consumes no draw.
- Red-capability verified by mutation: your first-character existence oracle (`ast_0…`/`ast_1…`) now fails the section, as does a backend that draws twice on the fresh path only (schedule skew).

Since the id body must now be byte-identical to an allocator output that never sees the storage lookup, an existence signal in the id has no channel left inside a conforming backend. CI re-running on the push.

cosarah: Thanks — the deterministic path catches the original mutant, but the contract still accepts two unrelated construction callbacks, so it does not prove the path users actually instantiate.

I verified this with one backend class whose default path (used by `makeStore`) returns fixed-width unique `ast_0…` IDs for dedupe hits and `ast_1…` IDs for fresh payloads, while its injected path (used by `makeStoreWithAllocator`) returns the supplied allocator output byte-for-byte and consumes exactly one output per `put`. The complete updated `runAssetStoreContract` still passes: 29/29 tests. The ordinary disclosure test continues to discard the ID body, and the exact-output assertion exercises only the separate instrumented factory.

So the suite currently proves that `makeStoreWithAllocator` is well behaved, not that the production/default construction path delegates to it. Please make the instrumented and production construction structurally the same path and add a mutation against the default path itself.

There is also a production invariant issue with the chosen seam: `BrowserAssetStoreOptions` is publicly exported, so a caller can pass `idAllocator: () => toAssetId('fixed')`; `put()` then persists and returns a non-random, non-`ast_`, non-128-bit ID. The instrumentation hook should remain internal/test-only rather than making the RFC's required ID scheme optional in the public backend.

wyuc: Both points taken — reworked in 0eb52d23:

**Single construction path.** `idAllocator` is gone from the constructor entirely; `BrowserAssetStore.put()` unconditionally draws from the id module's `newAssetId()`. The contract factory is now `{makeStore, withAllocator}`: `withAllocator(script, run)` installs the scripted allocator into the backend's *production* allocation source, and the allocation-independence section builds its store with the same `makeStore` every other test uses — there is no separate instrumented constructor left to be separately well-behaved.

**Internal/test-only seam.** The hook is `__setAssetIdFactoryForTesting` on the id module: consulted by `newAssetId()`, not re-exported from the package entry (the export-surface pin test now asserts its absence), and the exports map blocks deep imports — so the RFC id scheme is no longer optional through any public surface.

**Mutations against the default path itself** (both verified, then restored green): (a) a default-path `put` that derives the id's first body char from blob existence after hashing → fails (the dedupe branch returns its oracle id instead of the second scripted id); (b) your two-path shape — oracle default + faithful injected path — can no longer be adapted honestly: `withAllocator` wiring cannot reach a default path that doesn't flow through the instrumented source, and the first scripted id mismatches → fails.

CI re-running on the push.

packages/@openmaic/storage/test/asset-contract.ts
```diff
@@ -1,74 +1,387 @@
...
+    // success path facet by facet so a newly discovered observable has to be
+    // added to `PutObservation` to count as covered.
+    test('successful put paths do not disclose whether bytes already existed', async () => {
+      const s = makeStore();
+      const seen = new Set<string>();
+      // Seed so the next put of these bytes is a de-duplication hit...
+      seen.add(await s.put(blob('pre-existing bytes')));
+      const dedupeHit = await observePut(s, 'pre-existing bytes', seen);
+      // ...compared against a put of bytes the store has never seen, at the
+      // same call ordinal on the same store instance.
```
cosarah, line 242: [P2] These observations are not at the same call ordinal: the dedupe hit is call 2, while the fresh write is call 3. The deterministic allocation schedule has the same coverage gap—its second call is always a duplicate and its next fresh write is call 3.

I verified a concrete mutant that calls the instrumented production allocator exactly once per `put` and returns its output unchanged except when **call 2 is fresh**, where it changes one fixed-width ID-body character. The complete updated contract still passes: 29/29 tests. A caller can distinguish the histories `[candidate, candidate]` and `[unrelated, candidate]`, so this remains an existence oracle that the central suite accepts.

Please compare fresh and duplicate outcomes at identical ordinals—for example, run `[A, A]` and `[B, A]` with the same scripted allocator sequence and compare the second returned ID byte-for-byte.

packages/@openmaic/storage/src/asset/browser-store.ts
```diff
@@ -0,0 +1,465 @@
...
+    const dataType = data.type ?? '';
+    const { contentHash, bytes } = await contentHashOf(data);
+    const entry: AssetEntry = {
+      contentHash,
+      mime: storedMeta.contentType ?? dataType,
+      meta: storedMeta,
+    };
+    const id = newAssetId();
+    await this.tx('readwrite', async ({ assets, blobs }) => {
+      await reqP(blobs.put(bytes, contentHash));
```
cosarah, line 292: [P2] The registry still bypasses the blob-layer abstraction required by #1007. The RFC says `BrowserAssetProvider` becomes the internal blob backend beneath the registry and that S3/CDN should be replaceable through the same interface. Here `BrowserAssetStore` computes the hash and directly performs `blobs.put/get/delete`; it never receives or calls `BlobStore`, while `BrowserAssetProvider implements BlobStore` remains an unused parallel implementation.

This is more than an implementation detail: the registry's reference counting, byte reads, and reclamation are coupled to IndexedDB object stores and indexes. Replacing the content layer with S3/CDN therefore requires rewriting the registry. The current `BlobStore` contract is also URL/MIME-oriented, whereas this registry needs hash/byte operations with MIME retained per registry entry, so it is not presently a usable seam beneath the registry.

Please either layer the registry over a blob interface suitable for these operations, or explicitly revise the RFC/delivery-plan claim and remove the unused abstraction rather than describing it as the backend below this registry.

packages/@openmaic/storage/test/asset-contract.ts
wyuc: Fixed in 21975764 with your suggested shape, plus the ordinal-3 pair: fresh stores run paired histories under identical scripted allocator sequences — `[A, A]` vs `[A, B]` (ordinal 2 dup vs fresh) and `[A, B, A]` vs `[A, B, C]` (ordinal 3 dup vs fresh) — asserting every returned id equals the script's output for its ordinal byte-for-byte AND that the id at the differing ordinal is byte-identical across the two histories of the pair. The interleaved schedule stays. Your exact mutant (faithful single-draw echo except a one-character perturbation when call 2 is fresh) now fails both paired tests; verified and restored green.

packages/@openmaic/storage/src/asset/browser-store.ts
wyuc: Taken — and you're right that it was more than an implementation detail, so I took your second option and made it structural rather than building pluggability the browser cannot honor: 2a734706 deletes the standalone `BrowserAssetProvider` and the `BlobStore` interface outright (net −487 lines). Rationale, now stated in code, README, and the amended RFC text: browser-side reference counting, byte writes, and reclamation must share one IndexedDB transaction with the registry write that authorizes them; a transaction cannot span two databases, let alone an object store, so a browser-side blob seam could only ever be substituted with something in the same database — which is not substitutability. The provider's real contributions (sha256 hashing, object-URL caching, the branded `ContentHash`) survive as registry primitives in `blob.ts`. Pluggability (local object storage / S3 / CDN behind one interface) is the *server* registry's property — delivery plan part 4 — where authorization and blob consistency can genuinely be separated. #1007's four spots that claimed the provider 'becomes the internal blob backend' have been revised to match. CI re-running.

cosarah (approved):
Approved. The implementation now matches the amended #1007 design, the reported Spec issues are resolved, and CI is green.

Three non-blocking Standards follow-ups remain:

1. The @openmaic/dsl 0.5.0 -> 0.6.0 bump carries a breaking StorageProvider contract change, while CONTRIBUTING.md defines changes consumers must react to as major.
2. The @openmaic/storage 0.1.6 -> 0.2.0 bump removes BrowserAssetProvider and changes the persisted reference model, creating the same documented version-policy mismatch. Please either align the versions or document the intended 0.x exception.
3. The paired-history and interleaved allocation tests duplicate their scripted IDs and allocator/call-count machinery; consolidating that helper would reduce future test divergence.

PR #1028, opened 2026-07-31
fix(i18n): translate ASR provider names in the generation toolbar

niitbeo:
## The bug

The ASR provider dropdown in the media popover renders `p.name` straight from the `ASR_PROVIDERS` registry:

```ts
for (const p of Object.values(ASR_PROVIDERS)) {
  groups.push({ groupId: p.id, groupName: p.name, ... });
}
```

`name` there is a registry label, not a UI string, and several entries are written in Chinese:

| id | registry `name` |
| --- | --- |
| `browser-native` | `浏览器原生 ASR (Web Speech API)` |
| `qwen-asr` | `Qwen ASR (阿里云百炼)` |

So a user on any non-Chinese locale gets Chinese in the toolbar. The settings dialog shows the same providers correctly, because it resolves through `t('settings.providerBrowserNative')` — the translations already exist in every locale file; the toolbar simply never used them.

Screenshot from a Vietnamese UI, before this change — everything is Vietnamese except the provider itself:

> **Nhận dạng giọng nói** — `浏览器原生 ASR (Web Speech API)`

## The change

The id→i18n-key tables move to `lib/audio/provider-display.ts`, and both the settings dialog and the toolbar resolve names through it, so the two surfaces cannot drift apart again.

- The settings dialog keeps its own handling of custom providers, which carry a user-supplied `customName` rather than an i18n key.
- TTS is included in the shared module for symmetry — `doubao-tts` (`豆包 TTS 2.0（火山引擎）`) and `qwen-tts` have the same problem — though the popover's TTS tab is currently a toggle with no provider list.
- Image and video groups in the same popover also render `p.name`, but those registry labels are already English, so no user-visible string changes. Left alone to keep this focused; happy to fold them in if you would rather have it uniform.

## Verification

- `npx tsc --noEmit` reports nothing in `components/` or `lib/audio/`.
- `npx vitest run tests/store tests/prompts` — 342 passed.
- Prettier passes on the touched files.

### DIFF
--- a/components/generation/media-popover.tsx
+++ b/components/generation/media-popover.tsx
@@ -24,6 +24,7 @@ import {
 import { Switch } from '@/components/ui/switch';
 import { cn } from '@/lib/utils';
 import { useI18n } from '@/lib/hooks/use-i18n';
+import { resolveASRProviderName } from '@/lib/audio/provider-display';
 import { useSettingsStore } from '@/lib/store/settings';
 import { IMAGE_PROVIDERS } from '@/lib/media/image-providers';
 import { VIDEO_PROVIDERS } from '@/lib/media/video-providers';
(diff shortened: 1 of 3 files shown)


wyuc (approved):
Verified that all referenced ASR translation keys exist across the current locales, custom provider names remain intact, and language changes recompute the displayed labels. The current `main` merges cleanly.

PR #1029, opened 2026-07-31
refactor(dsl): name the default slide font instead of repeating it

niitbeo:
## Problem

The typeface applied to slide text when a document names none is written as the string literal `'Microsoft YaHei'` in every producer independently:

| Package | Sites |
| --- | --- |
| `@openmaic/dsl` | `normalize.ts` (text + shapeText defaults) |
| `@openmaic/renderer` | `SlideElement.tsx`, `elements/shape/BaseShapeElement.tsx` |
| app | `lib/action/engine.ts`, `lib/agent/client/apply-regenerate.ts`, `lib/api/stage-api-defaults.ts`, `lib/edit/slide-defaults.ts`, `lib/export/use-export-pptx.ts`, `lib/generation/scene-builder.ts`, `lib/generation/scene-generator.ts`, `components/slide-renderer/Editor/ScreenElement.tsx`, `components/slide-renderer/components/element/ShapeElement/BaseShapeElement.tsx`, `app/eval/whiteboard/page.tsx` |

Fourteen copies across three packages and the app, each separately answering the same question.

## Change

Add `DEFAULT_SLIDE_FONT` to `@openmaic/dsl` and import it at every site.

The DSL is the right home: it already owns `ELEMENT_DEFAULTS`, it has no runtime dependencies, and both `@openmaic/renderer` and `@openmaic/importer` already depend on it — so no new edge is added to the dependency graph.

**No behaviour change.** The value stays `'Microsoft YaHei'`.

### Deliberately left as literals

- The `@default` JSDoc on `defaultFontName` in `slides.ts`. It is compiled into the published JSON Schema and cannot reference a constant. The lockstep test in `normalize.test.ts` already fails loudly if it drifts from `ELEMENT_DEFAULTS`, so this stays enforced rather than merely documented.
- Test assertions — a test should pin a concrete value, not restate the constant it is testing.
- Comments citing the name as an example of a legacy or CJK face (`text-format-bar.tsx`, `textSerializer.ts`).
- The CSS font stack in `eval/pbl-v2-planner/compare-html.ts`, which styles the eval report page rather than slide content.

## Why this is worth doing

The face is chosen for Simplified Chinese and its coverage stops there. Checked against `msyh.ttc` with fontTools `getBestCmap`, seven of eight sampled Vietnamese letters have no glyph:

| Code point | Letter | Present |
| --- | --- | --- |
| U+1EBF | ế | no |
| U+1ED9 | ộ | no |
| U+1EEF | ữ | no |
| U+1EA1 | ạ | no |
| U+01B0 | ư | no |
| U+01A1 | ơ | no |
| U+1EB1 | ằ | no |
| U+0111 | đ | yes |

A missing glyph does not fail loudly — the browser substitutes another installed face for that single character, so a Vietnamese slide renders in two typefaces interleaved and shifts weight and spacing mid-word. The PPTX exporter writes the face name into the OOXML and PowerPoint substitutes the same way. Every non-CJK script is affected, not just Vietnamese.

This PR does not change the default. It makes retargeting a deployment a single edit instead of an audit, and the constant's doc comment records the two constraints on doing so:

1. Update the matching `@default` annotations in `slides.ts` — the schema lockstep test enforces this.
2. Use a single family name, not a CSS fallback list: the renderer would accept a list, but the PPTX exporter passes the value through as an OOXML `fontFace` where only one family name is valid.

## Verification

- `tsc --noEmit` clean.
- `@openmaic/dsl` 190/190, `@openmaic/renderer` 242/242, `@openmaic/importer` 33/33 pass — including the schema-lockstep test that pins `ELEMENT_DEFAULTS` to the JSON Schema `default` annotations.
- Full app suite: 3373 passed / 28 failed. The same 28 fail on unmodified `main` in a full run (`tests/quiz/runtime.test.ts`, `tests/runtime/chat-storage.test.ts`, `tests/store/kv-persist.test.ts` — they pass in isolation, so they look like shared-IndexedDB/Web-Locks interference rather than anything this touches). Zero regressions introduced.

### DIFF
--- a/app/eval/whiteboard/page.tsx
+++ b/app/eval/whiteboard/page.tsx
@@ -4,7 +4,7 @@ import { useEffect, useState } from 'react';
 import { ScreenElement } from '@/components/slide-renderer/Editor/ScreenElement';
 import { SceneProvider } from '@/lib/contexts/scene-context';
 import { useStageStore } from '@/lib/store/stage';
-import type { PPTElement } from '@openmaic/dsl';
+import { DEFAULT_SLIDE_FONT, type PPTElement } from '@openmaic/dsl';
 
 const EVAL_STAGE_ID = '__eval_stage__';
 const EVAL_SCENE_ID = '__eval_scene__';
@@ -41,7 +41,7 @@ function WhiteboardCanvas() {
               backgroundColor: '#ffffff',
               themeColors: ['#5b9bd5'],
               fontColor: '#333333',
-              fontName: 'Microsoft YaHei',
+              fontName: DEFAULT_SLIDE_FONT,
             },
             elements: [],
           },
@@ -66,7 +66,7 @@ function WhiteboardCanvas() {
               backgroundColor: '#ffffff',
               themeColors: ['#5b9bd5'],
               fontColor: '#333333',
-              fontName: 'Microsoft YaHei',
+              fontName: DEFAULT_SLIDE_FONT,
             },
             elements: incoming,
           },
--- a/components/slide-renderer/Editor/ScreenElement.tsx
+++ b/components/slide-renderer/Editor/ScreenElement.tsx
@@ -14,6 +14,7 @@ import { BaseVideoElement } from '../components/element/VideoElement/BaseVideoEl
 import { BaseCodeElement } from '../components/element/CodeElement/BaseCodeElement';
 import { useSceneSelector } from '@/lib/contexts/scene-context';
 import type { SceneContent } from '@/lib/types/stage';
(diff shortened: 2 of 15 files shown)


wyuc:
Thanks for this, @niitbeo, and sorry it sat so long. A single named default font is the right idea, but `main` has moved about 200 commits since this was opened: the literal now appears in 20 files, and this branch conflicts in six of them, so a rebase would be close to a rewrite. I'm closing this to keep the queue honest; if you'd like to pick it up again, a fresh PR against current `main` is very welcome.

PR #1025, opened 2026-07-31
feat(i18n): add Vietnamese (vi-VN) locale

niitbeo:
## Summary

Adds a complete Vietnamese translation, bringing the interface to 10 languages.

Vietnamese is among the largest user bases in Southeast Asia for education tooling, and the locale registry already had room for it — this fills that gap. The translation was produced and used in a real deployment (a university lecturer generating coursework from Vietnamese textbooks), so the wording is checked against actual classroom use rather than translated in the abstract.

## Changes

| File | What |
| --- | --- |
| `lib/i18n/locales/vi-VN.json` | New locale, all 1514 keys |
| `lib/i18n/locales.ts` | Register `vi-VN` so it appears in the language picker |
| `README.md`, `README-zh.md` | Correct the language list |

## Translation decisions

- **Language names under `settings.lang_*` keep their endonyms** (`日本語`, `Bahasa Indonesia`, `English (United States)`), matching how the existing locales handle them.
- **Product and provider names stay untranslated** (`OpenAI TTS`, `Nano Banana`, `Qwen Image`), as in the other locales.
- Everything else is translated, including the `settings.qwenVoice*` voice-personality descriptions.

## Drive-by README fix

The feature list advertised *"7 languages"* and the project-tree comment in `README-zh.md` listed 7 locale codes, but `locales.ts` already shipped 9 — `ko-KR` and `es-MX` had been added without updating either list. Both now reflect the real set (10 with this change). Happy to split this into its own PR if you would rather keep the change focused.

## Verification

- `pnpm run check:i18n-keys` passes — `vi-VN.json` is key-for-key identical to `en-US.json`, no extra or missing keys.
- `pnpm run check` (Prettier) passes on the touched files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1030, opened 2026-08-01
i18n(agents): translate the built-in agent names and personas

niitbeo:
## Problem

Five of the six built-in agents are named only in Chinese, as string literals in `lib/orchestration/registry/store.ts`:

| id | name | role |
| --- | --- | --- |
| default-1 | `AI teacher` | teacher |
| default-2 | `AI助教` | assistant |
| default-3 | `显眼包` | student |
| default-4 | `好奇宝宝` | student |
| default-5 | `笔记员` | student |
| default-6 | `思考者` | student |

Every deployment renders those names verbatim, so a Vietnamese, Korean or Spanish classroom lists most of its roster in a script its users may not read. The personas are English-only for all six.

## Change

Move both fields to `agents.defaults.<id>.{name,persona}` and resolve them at read time.

The Chinese names are not deleted — they are filed where they belong. `zh-CN` keeps the originals; `zh-TW` gets the traditional forms (`顯眼包`, `好奇寶寶`, `筆記員`). The remaining locales carry the English text until translated, which is what the existing behaviour already produced for personas.

### Implementation notes

- **Resolved on read**, not at module load, so switching language re-reads it.
- **`localizeAgent` returns early on the server.** This module is imported by API routes, and `@/lib/i18n` pulls in the react-i18next bindings; evaluating them during the server build fails with `TypeError: (0 , J.createContext) is not a function` while collecting page data. Requiring it behind a `typeof window` guard keeps it off the server path. Nothing is lost: server-side the persona is only ever an LLM prompt, and the teaching language is set by the course's language directive, not by the language the prompt happens to be written in.
- **A missing key falls back to the literal**, so a locale that has not translated these yet shows English rather than a raw key path.
- **The literals are now English**, so that fallback is meaningful rather than Chinese.
- **Persist version 11 → 12.** Existing snapshots hold the old hardcoded text. The migration replaces entries flagged `isDefault` and leaves anything the user added or edited untouched.

## Verification

- `pnpm check:i18n-keys` — passes, 9 locale files aligned.
- `tsc --noEmit` clean, `eslint` clean on the changed file.
- `vitest run tests/i18n tests/orchestration tests/agent` — 39/39 pass.
- `pnpm build` succeeds (this is what surfaced the server-side i18n import problem above).

## Note for #1025

#1025 adds `vi-VN.json`. Whichever of the two lands second will need the other's keys, or `check:i18n-keys` will fail on the merge. I will update #1025 to carry `agents.defaults` as soon as this one is reviewed, so the ordering does not matter.

### DIFF
--- a/lib/i18n/locales/ar-SA.json
+++ b/lib/i18n/locales/ar-SA.json
@@ -1735,5 +1735,33 @@
     "title": "أدخل رمز الوصول",
     "placeholder": "رمز الوصول",
     "error": "رمز الوصول غير صالح. يرجى المحاولة مرة أخرى."
+  },
+  "agents": {
+    "defaults": {
+      "default-1": {
+        "name": "AI Teacher",
+        "persona": "You are the lead teacher of this classroom. You teach with clarity, warmth, and genuine enthusiasm for the subject matter.\r\n\r\nYour teaching style:\r\n- Explain concepts step by step, building from what students already know\r\n- Use vivid analogies, real-world examples, and visual aids to make abstract ideas concrete\r\n- Pause to check understanding — ask questions, not just lecture\r\n- Adapt your pace: slow down for difficult parts, move briskly through familiar ground\r\n- Encourage students by name when they contribute, and gently correct mistakes without embarrassment\r\n\r\nYou can spotlight or laser-point at slide elements, and use the whiteboard for hand-drawn explanations. Use these actions naturally as part of your teaching flow. Never announce your actions; just teach.\r\n\r\nTone: Professional yet approachable. Patient. Encouraging. You genuinely care about whether students understand."
+      },
+      "default-2": {
+        "name": "AI Teaching Assistant",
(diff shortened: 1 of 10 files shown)


niitbeo:
Correction to the note at the end of the description — I had the ordering backwards.

I said I would update #1025 "as soon as this one is reviewed". I tried it, and it is wrong: adding `agents.defaults` to `vi-VN.json` while this PR is unmerged leaves that locale holding twelve keys the other nine do not have, and `check:i18n-keys` fails on #1025 itself.

The dependency runs the other way. Whichever of the two merges **second** is the one that needs updating:

- **This PR merges first** → #1025 rebases and adds `agents.defaults` to `vi-VN.json`.
- **#1025 merges first** → this PR rebases and picks up `vi-VN.json` as a tenth locale.

Either is fine and neither blocks the other from being reviewed now. I will do whichever rebase applies once the first one lands, so the second never merges red.

wyuc:
Thanks, @niitbeo, and sorry for the long wait. This is still a real gap on `main`: five of the six built-in agents are named only in Chinese in `lib/orchestration/registry/store.ts`, so non-Chinese classrooms show a mostly Chinese roster. The conflicts are only in the locale JSON files, since `main` has added strings there since August. Could you rebase onto current `main`? Once it's green we'll review it promptly.

PR #1023, opened 2026-07-31
fix(editor): localize Pro mode right rail labels

cosarah:
## Summary

- Localize Pro mode right-rail tab labels through `edit.agent.title` and `edit.roster.title`
- Localize the classroom roster panel copy, placeholders, aria labels, and action labels
- Add the new keys across all supported app locales
- Localize the Korean roster subsection label found during agent code review

Fixes #1022

## Verification

- Agent code review passed after the Korean locale follow-up
- GitHub CI checks passed
- `node scripts/check-i18n-keys.mjs`
- `/home/ys/OpenMAIC/node_modules/.bin/prettier --check components/edit/RightRailTabs.tsx components/edit/AgentPanel/AgentPanel.tsx components/edit/AgentsView/AgentRosterPanel.tsx lib/i18n/locales/en-US.json lib/i18n/locales/zh-CN.json lib/i18n/locales/zh-TW.json lib/i18n/locales/ja-JP.json lib/i18n/locales/ko-KR.json lib/i18n/locales/es-MX.json lib/i18n/locales/pt-BR.json lib/i18n/locales/ru-RU.json lib/i18n/locales/ar-SA.json`
- `/home/ys/OpenMAIC/node_modules/.bin/eslint components/edit/RightRailTabs.tsx components/edit/AgentPanel/AgentPanel.tsx components/edit/AgentsView/AgentRosterPanel.tsx`
- `/home/ys/OpenMAIC/node_modules/.bin/vitest run tests/i18n/edit-pr2-locales.test.ts tests/i18n/edit-elements-locales.test.ts`

## Notes

Full-repo `tsc --noEmit --pretty false` still fails on current `main` due to existing package export, missing dependency, and unrelated type errors; none of the failures point to this change.

PR #1033, opened 2026-08-01
i18n(editor): localize the classroom roster panel

niitbeo:
## Problem

`components/edit/AgentsView/AgentRosterPanel.tsx` was never wired to i18n — no `useI18n` import, zero `t()` calls. Its fifteen UI strings are Chinese literals, so every deployment renders them verbatim:

| String | Meaning |
| --- | --- |
| `课堂阵容` | Classroom roster (panel title) |
| `点击一位展开编辑` | Click one to edit |
| `主讲` | Lead (teacher badge) |
| `AI 同学` / `位` | AI classmates / members |
| `人设描述` | Persona (field label) |
| `描述角色的性格、教学风格与任务…` | textarea placeholder |
| `暂无人设` / `未命名` | No persona yet / Untitled |
| `移出课堂` / `添加角色` | Remove from classroom / Add character |
| `上移` / `下移` | Move up / Move down (aria-labels) |
| `撤销` / `重做` | Undo / Redo |

The roster tab label in `RightRailTabs.tsx` is hardcoded the same way — sitting next to eleven strings that already go through `t()`, which is what makes the gap easy to miss.

This is the whole panel, not a stray label. A Vietnamese, Korean or Spanish classroom manages its roster entirely in a script its users may not read.

## Change

Move the strings to `agents.roster.*` and resolve them through `useI18n` in the four components that render them (`PersonaEditor`, `TeacherCard`, `ClassmateCard`, `AgentRosterPanel`), plus the tab label.

The Chinese is not deleted — it is filed where it belongs. `zh-CN` keeps the originals; `zh-TW` gets the traditional forms (`課堂陣容`, `復原`, `新增角色`). The remaining locales carry the English text until translated.

Comments that referenced the panel by its Chinese name were changed to English so the file reads consistently.

## Verification

- `pnpm check:i18n-keys` — passes, 9 locale files aligned.
- `tsc --noEmit` clean; `eslint` clean on both changed components.
- No behaviour change beyond the rendered strings.

## Related

Same class of gap as #1030 (built-in agent names/personas). Both were found while running this project as a Vietnamese deployment. If you would like, I can sweep for the remaining cases — the pattern is files carrying CJK literals with no `useI18n` import, and there are a few more.


LING-6150:
Thanks for the contribution. I rechecked this against the PR’s base and current main. The roster panel had already been localized in #1023 before this PR was opened: AgentRosterPanel was using useI18n for the roster labels and actions, and the tab label was already wired to edit.roster.title.
This PR therefore duplicates the existing localization while introducing a second agents.roster.* key namespace and another RightRailTabs implementation. The current code has since moved further on, so rebasing would not leave a necessary incremental change.
Closing as already addressed by #1023. Thank you for identifying and working on the localization gap.

PR #1034, opened 2026-08-01
feat(llm): fall back to another model when one is out of quota

niitbeo:
## Problem

A free-tier key meters **each model separately, per day**. The model a stage is routed to runs dry while its siblings still have headroom.

The existing retry does not help — it re-sends to the *same* exhausted model, spending its attempts on an error that cannot clear within the call:

```
[Scene Content API] Scene content generation failed
  [scene="…", model=google:gemini-3-flash-preview]
  AI_APICallError: You exceeded your current quota
…
[Outlines Stream] Outline generation failed after 3 attempts: LLM returned empty response
```

Outlines are the first step of a lecture, so one exhausted model there fails the whole generation.

## Change

An opt-in chain to walk when, and only when, the failure is a capacity rejection:

```env
MODEL_FALLBACKS=google:gemini-3.6-flash,google:gemini-3.5-flash-lite,google:gemini-3.1-flash-lite
```

Unset means the previous behaviour, unchanged. Wired into `callLLM` (which covers scene content, scene actions, agent profiles, quiz grading, web search) and into the outlines route.

Only 429/503 triggers it. Any other failure still fails where it happened — retrying a bad prompt or a malformed schema elsewhere would hide the real cause and spend a second model's quota to reproduce the same error.

### Two details worth flagging

**`verify-model` is excluded.** It probes the exact model a user typed in and reports whether it works. Answering from a different model would report a dead or mis-keyed model as healthy — worse than the failure it was asked to detect. There is a test pinning this.

**The outlines route needed `onError`, not just a catch.** `streamText` does not reject its text stream on an upstream failure; it ends the stream and reports through the callback. A 429 therefore arrives as a zero-length response and the catch never sees it. My first attempt type-checked, passed review by eye, and did nothing at runtime — it took testing against a real exhausted key to find.

Usage is attributed to the model that actually served the call, not the one originally selected.

## Verification

- 19 unit tests, including: an ordinary error does **not** switch models; auth and bad-request failures do not; a self-referential `cause` chain terminates; the chain never hands out the same model twice; an unresolvable entry is skipped rather than abandoning the rest.
- `tsc --noEmit` clean.
- **End to end against a genuinely exhausted key.** Pointed the primary at a model that was actually out of quota; the call fell through two models and produced outlines on the third:

```
Outlines: model out of capacity; falling back to google:gemini-3.6-flash.
Outlines: model out of capacity; falling back to google:gemini-3.5-flash-lite.
→ HTTP 200, 7440 bytes
```

## Scope note

`streamLLM` itself is untouched — it returns the stream synchronously and has no retry loop, so failover there would need a different shape. The outlines route is handled because it owns its own attempt loop. Happy to extend if you would like it broader.

### DIFF
--- a/app/api/generate/scene-outlines-stream/route.ts
+++ b/app/api/generate/scene-outlines-stream/route.ts
@@ -36,6 +36,7 @@ import type {
 import { apiError } from '@/lib/server/api-response';
 import { createLogger } from '@/lib/logger';
 import { resolveModelFromRequest } from '@/lib/server/resolve-model';
+import { isCapacityError, nextFallbackModel } from '@/lib/server/model-fallback';
 import { sortDocumentImagesForVision } from '@/lib/document/bundle';
 import { resolveVocationalActive } from '@/lib/config/feature-flags';
 const log = createLogger('Outlines Stream');
@@ -311,6 +312,13 @@ export async function POST(req: NextRequest) {
     };
     requirementSnippet = requirements?.requirement?.substring(0, 60);
 
+    // The UI locale is the default teaching-language signal. Without it a
+    // language-neutral requirement ("ok", "go", a bare topic) leaves the model
+    // to infer the language from the reference document, so a Vietnamese
+    // teacher uploading an English PDF gets an English course. An explicit
+    // request in the requirement text still wins — see the prompt's decision
+    // rules — this only fills the gap when the requirement carries no signal.
+
     // Build user profile string for language inference context
     const userProfileText =
       requirements.userNickname || requirements.userBio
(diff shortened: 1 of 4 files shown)


wyuc (changes requested):
Thanks, @niitbeo, and sorry for the long wait. Opt-in failover on quota errors is useful, and keeping the unset case identical to today's behavior is the right default. It still merges cleanly. Two issues (both raised by the automated review) need addressing before merge:

1. **Model-specific limits are not recomputed after a fallback.** The swap replaces the model but keeps the original call's `maxOutputTokens`. Several callers set that to the original model's full output window, so falling back to a model with a smaller window turns a quota error into a request the fallback rejects. Please re-derive the output limit (and any other per-model parameters) for the fallback model, or clamp it to the fallback's window.

2. **Evaluation calls should not fall back.** The eligibility check is allow-by-default, so the `eval-*` sources would silently run on a different model while their reports still carry the original model's name, which corrupts the comparison. Please exclude evaluation sources (or switch to an explicit allow-list of sources that may fail over).

A test for each would be great: a fallback to a smaller-window model, and an evaluation source that stays on its model.

PR #1037, opened 2026-08-02
chore(storage): address asset registry review follow-ups

wyuc:
The three non-blocking follow-ups from the #1024 approval:

1. **Pre-1.0 version policy documented** (CONTRIBUTING.md): for packages below 1.0.0, a minor bump signals a breaking change and a patch bump a compatible one, per common 0.x semver practice; the major rule applies from 1.0.0. This records the intended exception behind the `@openmaic/dsl` 0.6.0 and `@openmaic/storage` 0.2.0 bumps.
2. **Allocation-test machinery consolidated** (`test/asset-contract.ts`): one shared scripted-allocator/history-runner helper now backs the paired-history and interleaved-schedule tests; every assertion survives in meaning (byte-for-byte ordinal equality, one draw per put, replace consumes no draw, cross-history same-ordinal identity), and the call-2-fresh perturbation mutant still fails the ordinal-2 test (re-verified).
3. Docs and tests only — no behavior or version changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -136,6 +136,8 @@ Choosing the number is a [semver](https://semver.org/) judgement, and it is your
 - **minor** — new behaviour that existing consumers can ignore
 - **major** — anything an existing consumer must react to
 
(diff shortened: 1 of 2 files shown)


cosarah (approved):
LGTM

PR #1038, opened 2026-08-02
feat(dsl): unify media asset reference types (#1007 part 2, step a)

wyuc:
Step (a) of #1007's delivery plan part 2 — type unification with **zero behavior change**, per the RFC's sequenced-removal amendment.

## What lands

- `PPTImageElement.src`, `PPTVideoElement.src`/`mediaRef`, and `SpeechAction.audioId` take the `AssetRef` type (a string alias — no runtime or downstream type impact; root and per-package `tsc` verified clean). Each field's docstring states uniformly what it holds today: placeholder/derived ids and concrete URLs are foreign values to the asset pool, addressed by later delivery-plan steps.
- `SpeechAction.audioUrl` **stays**, deprecated in prose (replacement: `audioId`). Per the amended RFC, its removal ships together with the byte-ingestion step (part 2 step c) — a pure migration can neither ingest bytes nor decide the co-presence case, where a server-issued `audioId` may be a dangling reference whose only live handle is the URL beside it. The `@deprecated` tag form is deliberately avoided: `ts-json-schema-generator` emits it as a string-typed `deprecated` key, which is invalid against JSON Schema 2019-09+ meta-schemas; a schema test now asserts no property in any generated root carries a non-boolean `deprecated`.
- Legacy migration verification hardening (`lib/document-store/migration.ts`): the comparison baseline now runs through the DSL migration ladder with the app-owned `outline` stripped and re-attached, mirroring the document store's own read path — so the first real ladder step (part 2 step c) cannot break legacy Dexie verification. The DSL migrate is an injectable seam in `DocumentMigrationDeps`, and both call sites are pinned by injection tests (mutation-verified: reverting the wiring fails a test).
- A pin that the agent edit gate still resolves the new `$ref` indirection for the three fields to `string`, so a future schema-generation change degrades loudly.

`@openmaic/dsl` 0.6.0 → 0.6.1 (docstrings and a type alias; no value domain narrows — patch per the pre-1.0 policy). `DSL_VERSION` is untouched.

Steps (b) write paths and (c) reference conversion + `audioUrl` removal + `DSL_VERSION` bump follow as separate PRs.

### DIFF
--- a/lib/agent/tools/edit-elements-gate.ts
+++ b/lib/agent/tools/edit-elements-gate.ts
@@ -223,7 +223,7 @@ function decodeRefName(ref: string): string | null {
 
 assertGateSchemaSource(schemaDefinitions);
 
-export function resolveSchema(schema: JsonSchema, seen = new Set<string>()): JsonSchema {
+function resolveSchema(schema: JsonSchema, seen = new Set<string>()): JsonSchema {
   if (!schema.$ref) return schema;
   const name = decodeRefName(schema.$ref);
   if (!name || seen.has(name)) return schema;
--- a/lib/agent/tools/edit-elements-gate.ts
+++ b/lib/agent/tools/edit-elements-gate.ts
@@ -223,7 +223,7 @@ function decodeRefName(ref: string): string | null {
 
 assertGateSchemaSource(schemaDefinitions);
 
-function resolveSchema(schema: JsonSchema, seen = new Set<string>()): JsonSchema {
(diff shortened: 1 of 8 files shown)


cosarah (approved):
The implementation looks good to merge. I verified the migration verification path against the actual document-store behavior, including DSL migration and outline preservation, and the targeted migration, schema, and edit-gate tests pass. I found no blocking defects.

PR #1039, opened 2026-08-02
feat(media): allocate generated assets through the registry (#1007 part 2, step b)

wyuc:
Step (b) of #1007's delivery plan part 2 — write paths. Newly generated media and TTS audio are
allocated through the Part-1 asset registry (`BrowserAssetStore`), and documents reference the
allocated ids; existing documents with `gen_*` placeholders keep working unchanged.

## What lands

- **Allocation on generation**: image/video/poster and TTS bytes go `pool.put → Dexie
  compatibility row (same id) → document rewrite` — in that order, so any step's failure leaves
  the document pointing at the old value. Regeneration of a solely-owned allocated ref uses
  `pool.replace` (the stable-identity operation); a ref shared by duplicated slides forks a fresh
  id for the targeted element only. Generation completing before its scene exists reconciles at
  scene insertion (both orders converge; resume consults the document, so reopening a classroom
  never re-generates or re-bills).
- **Sealed resolution state machine** (`lib/media/resolve-media-ref.ts`): every consumer surface
  (renderers, thumbnails, hydration, PPTX and video export) maps a ref through one decision table
  — url / pending / failed(retryable) / placeholder / disabled / raw — with an executed
  state-space × surface matrix. An opaque ref can never reach a DOM `src`; unknown concrete
  addresses fall back to the browser exactly as before.
- **Lease-based URL ownership** (`lib/media/use-asset-url.ts`): the only module that talks to the
  pool for URLs; per-ref counted leases with serialized release/reacquire, batch hooks that
  republish on same-id replacement, and a scoped one-shot lease for export paths. A boundary
  guard test keeps direct `pool.resolve/release` out of the rest of the app.
- **Reclamation confined to stage deletion**: the one deletion-triggered reclaim entry (plus
  in-transaction rollbacks for generation/TTS/import). The stage cascade enumerates the stage's
  own document plus stage-matched rows and never attributes stage-less legacy audio rows.
  Inline reclamation on element/cue/scene deletion was deliberately removed after review showed
  every variant races an undo window; orphaned pool entries are bounded transitional garbage
  until part 3's document-truth sweep.
- **Import re-mints**: imported zips allocate fresh ids for media/audio/posters (bytes into the
  pool + Dexie), rewrite refs by membership, preserve concrete addresses verbatim, and roll back
  all allocations on any failure — imported courses can never alias or mutate another course's
  assets.
- **Dexie double-write stays**: export paths, thumbnails, and import/export continue reading the
  compatibility rows (keyed by the allocated id); convergence to the pool alone is part 3.
- Clear-cache wipes the asset pool database too (deferred loudly when another tab blocks it,
  with localized guidance); `@openmaic/dsl` 0.6.2 (one additive optional field).

## Notes for review

- The predicate `isMediaPlaceholder` now means "needs resolution" (anything that is not a
  concrete address); surfaces never blank an unknown string — they fall back to it.
- Key invariants are mutation-verified by the suite: crash-safety ordering, single-flight leases,
  fork isolation (source task untouched), legacy-row survival in the deletion cascade,
  allocation-independence of ids, and the no-opaque-ref-in-src matrix.

Step (c) — reference conversion, `audioUrl` removal, and the `DSL_VERSION` bump in one unit —
follows.


components/slide-renderer/components/element/VideoElement/BaseVideoElement.tsx
```diff
@@ -4,68 +4,57 @@ import { useRef, useEffect } from 'react';
...
-  const showSkeleton =
-    isPlaceholder &&
-    !concreteSrc &&
-    !showDisabled &&
-    (!task || task.status === 'pending' || task.status === 'generating');
-  const showError = isPlaceholder && !concreteSrc && task?.status === 'failed';
+  const mediaGenerationDisabled = useSettingsStore((state) => !state.videoGenerationEnabled);
+  const mediaRef = videoMediaRefForResolution(elementInfo);
+  const task = useMediaGenerationStore((state) =>
+    selectVideoMediaTaskForElement(state.tasks, elementInfo, stageId),
```
cosarah: **P2 — Centralize legacy video recovery across all consumers**

This fallback is only used by `BaseVideoElement`. `VideoElement` and `resolveSlideMediaState` still require an exact or `placeholderRef` match. For a restored legacy classroom whose persisted video row uses a mismatched key, read-only playback recovers the clip while edit mode and resolved-slide consumers keep the placeholder. The same document therefore renders different media depending on the surface. Please move this recovery into the shared task-resolution seam, or remove the surface-specific fallback.

components/slide-renderer/components/element/VideoElement/useResolvedVideoMedia.ts
```diff
@@ -0,0 +1,80 @@
...
+  const reconciled = Object.values(tasks).find(
+    (candidate) => candidate.stageId === stageId && candidate.placeholderRef === mediaRef,
+  );
+  if (reconciled) return reconciled;
+
+  const storedVideos = Object.values(tasks).filter(
+    (candidate) =>
+      candidate.stageId === stageId && candidate.type === 'video' && candidate.status === 'done',
+  );
+  return isLegacySequentialVideoRef(mediaRef) && storedVideos.length === 1
```
cosarah: **P2 — Do not apply the singleton fallback independently per element**

`storedVideos.length === 1` only proves that the stage has one completed video task; it does not prove that the document has one unmatched video element. With two elements (`gen_vid_1` and `gen_vid_2`) and one completed task whose key matches neither, each `BaseVideoElement` calls this selector independently and both receive `storedVideos[0]`. I reproduced the downstream result as `{ first: "blob:only-stored-video", second: "blob:only-stored-video" }`. The thumbnail hydration loop repeats the same heuristic. Please resolve this at document scope and only fall back when both the stored candidate and unmatched element are unambiguous.

wyuc: Both points taken — fixed in 4da416c7 by centralizing rather than patching per surface:

- **Document-scoped decision**: the recovery now lives in a shared seam (`lib/media/media-task-resolution.ts`) that receives every video element of the document (slides + whiteboards). The mapping applies only when there is exactly ONE unmatched legacy element AND exactly ONE completed stored row (your two-elements repro now keeps both as placeholders — pinned); the exact-ref failure guard is unchanged. Thumbnail hydration consumes the same decision instead of repeating the heuristic.
- **All surfaces converge**: BaseVideoElement's surface-specific fallback is removed; playback, edit-mode `VideoElement`, `resolveSlideMediaState`, and thumbnail hydration all use the shared lookup, so a restored legacy classroom renders identically everywhere (surface-matrix test extended).

CI re-running on the push.

lib/media/media-task-resolution.ts
```diff
@@ -0,0 +1,100 @@
...
+  tasks: Readonly<Record<string, T>>,
+  documentElements: readonly PPTElement[],
+  stageId: string,
+): Record<string, T> {
+  const unmatchedLegacyVideos = documentElements.filter(
+    (element) =>
+      isLegacySequentialVideoElement(element) &&
+      !resolveMediaTaskForElement(tasks, element, stageId),
+  );
+  const completedVideos = Object.entries(tasks).filter(
```
cosarah: **[P2] Exclude completed video tasks already owned by another element**

This candidate set includes every completed video task, even when another document element already resolves to that task by exact ref. For example, with one element referencing `gen_vid_unique_legacy`, one unmatched legacy element referencing `gen_vid_1`, and a single completed task keyed by `gen_vid_unique_legacy`, both counts are one. The recovery then adds `placeholderRef: "gen_vid_1"` to the already-owned task, so both elements resolve to the same video blob.

I reproduced this deterministically with a focused test: the legacy element unexpectedly received the task `{ elementId: "gen_vid_unique_legacy", placeholderRef: "gen_vid_1" }` instead of remaining unresolved. Please exclude tasks consumed by any exact/targeted document element from the recovery candidates, and add this mixed matched/unmatched case as a regression test.

wyuc: Confirmed and fixed in 716e0523. Since this is the third finding of the same shape (cardinality heuristics), I rewrote the recovery as an explicit matching problem rather than adding another condition:

- **Pass 1** resolves every document video element (slides + whiteboards) against the task table by exact ref / targeted key / `placeholderRef`, recording `matchedElements` and `consumedTasks`.
- **Pass 2** takes `candidates = completed video tasks − consumedTasks` and `unmatched = document video elements − matchedElements` (an element whose exact ref carries a failure row stays excluded — that guard remains authoritative).
- Recovery applies **only** when `|candidates| === 1 && |unmatched| === 1`, then pairs them. Every other cardinality recovers nothing, so "both counts happen to be one" is no longer expressible.

Your mixed case is now a verbatim regression test (the `gen_vid_1` element stays unresolved while `gen_vid_unique_legacy` keeps its own task), alongside ownership pins for targeted-key and `placeholderRef` claims and a 12-cell matrix over {matched exactly / unmatched / failure row} × {consumed / free} × {one / two candidates} — only the unambiguous cell recovers. I also audited the module for the same hazard elsewhere: images use the shared lookup directly, and posters/audio have no cardinality recovery, so there was nothing else to convert.

CI re-running on the push.

components/slide-renderer/components/element/VideoElement/useResolvedVideoMedia.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/** Shared direct-video binding used by both the read-only and editor elements. */
+export function useResolvedVideoMedia(
+  element: PPTVideoElement,
+  task: MediaTask | undefined,
+  mediaGenerationDisabled: boolean,
+): ResolvedVideoMedia {
+  const mediaRef = videoMediaRefForResolution(element);
+  const resolution = useResolvedMediaRef(mediaRef ?? element.src, task, mediaGenerationDisabled);
```
cosarah: [P1] Prefer a concrete video `src` over an opaque `mediaRef` here. The server intentionally preserves elements that contain both a usable direct URL and a generated-media reference. If the browser has no task or pool entry for that opaque reference, this call resolves to `placeholder`, so both direct video components render a skeleton even though `element.src` is playable. `resolveSlideMediaState` already gives a concrete `src` precedence; this binding should follow the same rule. I reproduced this with `{ src: "https://cdn.example/direct.mp4", mediaRef: "ast_opaque" }`: the direct binding returned an empty source with `kind: "placeholder"`.

lib/media/media-orchestrator.ts
```diff
@@ -82,47 +123,492 @@ export async function retryMediaTask(elementId: string): Promise<void> {
...
+      ? await replaceablePosterRefs(task.stageId, currentPosterRefs(task.stageId, replaceAssetId))
+      : [];
+
+  // Only legacy/failure placeholders are disposable. Keep them durable while
+  // the retry is in flight, then remove them after a successful allocation.
+  const failureRefs = !replaceAssetId
+    ? new Set([elementId, target?.elementId].filter(Boolean) as string[])
+    : undefined;
+
+  const progressKey = shared ? scopedTarget!.elementId : elementId;
```
cosarah: [P2] Make `play_video` observe this element-scoped retry task. For shared media, retry progress is stored under `scopedTarget.elementId`, but `ActionEngine.executePlayVideo` still resolves the element back to the source media reference and watches `tasks[placeholderId]`. During a targeted retry it can therefore see the stale source task rather than this pending fork, causing playback to be skipped or started before the selected video is renderable. The action path should use the same element-aware task lookup as the renderers.

components/slide-renderer/components/element/VideoElement/useResolvedVideoMedia.ts
wyuc: Fixed in 7825afde (+ 8f3b71f3). Your repro is now a verbatim test: `{ src: 'https://cdn.example/direct.mp4', mediaRef: 'ast_opaque' }` yields the playable concrete source instead of an empty placeholder.

Rather than teach this one binding the precedence rule, I converged the decision — *for this video element, which reference do we use and which task governs it* — into a single exported resolver in `lib/media/media-task-resolution.ts`. It now owns concrete-src precedence, mediaRef vs legacy-src selection, poster binding, the exact/targeted/placeholderRef task lookup, and the document-scoped legacy recovery. Every consumer was rewired to it: both direct video components, `resolveSlideMediaState`/`use-resolved-slide`, thumbnail hydration, and the `play_video` action path — each site's own src/mediaRef choosing and `tasks[...]` indexing is gone.

This is the fourth finding of the same shape on this branch (a consumer missing a shared rule), so there is also a boundary guard test: outside the resolver module, nothing under `components/`, `lib/action/`, `lib/playback/`, or `stage-storage.ts` may index the task table for video resolution or pick between `src` and `mediaRef` itself. The test comment states plainly what the regex cannot catch (aliases, computed access, helper indirection).

lib/media/media-orchestrator.ts
wyuc: Fixed in 7825afde. `ActionEngine.executePlayVideo` no longer resolves the element back to the source reference and watches `tasks[placeholderId]`; it goes through the same element-aware resolver the renderers use, so during a targeted retry it observes the pending fork rather than the stale source task.

`tests/action/play-video-media-resolution.test.ts` pins it: playback waits until the fork is renderable instead of being skipped or started early, and the consumer matrix in `media-resolution-surface-matrix.test.ts` now carries the action path as its own column alongside the renderer/thumbnail/export surfaces — so the action seam is covered by the same state-space assertions rather than only by its own test.

CI re-running on the push.

lib/utils/database.ts
```diff
@@ -529,6 +542,7 @@ export async function clearDatabase(runtimeStore?: RuntimeStore): Promise<void>
     await deleteAllDocuments();
     await clearDocumentStoreKeys();
     await db.delete();
+    await clearAssetPool();
```
cosarah, line 545: **[P2] Handle blocked asset-pool deletion as partial completion**

`clearAssetPool()` can reject here after runtime data, documents, document keys, and the main Dexie database have already been deleted. I reproduced this with another IndexedDB connection held open: `clearDatabase()` rejected with `AssetPoolDeletionDeferredError`, the preceding stores were already empty, and the current asset pool refused new writes until retry. `GeneralSettings` consequently skips local/session/KV cleanup and the reload. Closing the other tab and retrying does recover, but the first attempt leaves the live page in a partially cleared state while reporting failure. Please either explicitly finish/reload after this partial completion, or persist the clear phase and resume it automatically on retry or reload.

lib/export/use-export-pptx.ts
```diff
@@ -962,39 +1047,31 @@ export async function buildPptxBlob(
...
         if (mediaLookupKey) {
-          const task = useMediaGenerationStore.getState().tasks[mediaLookupKey];
-          if (task?.status === 'done' && task.objectUrl) {
-            resolvedSrc = task.objectUrl;
-          } else if (!resolvedSrc || isMediaPlaceholder(resolvedSrc)) {
-            continue; // Media not ready, skip
-          }
+          const stored = await resolveStoredMediaBlob(mediaLookupKey, stageId);
+          if (stored) resolvedSrc = await blobToDataUrl(stored);
+          else resolvedSrc = renderableMediaUrl(exportMediaResolution(mediaLookupKey, stageId));
```
cosarah: **[P1] Preserve a concrete video `src` when the `mediaRef` lookup misses**

This branch first resolves `el.src`, but a video with any `mediaRef` always enters this block. When the opaque lookup misses, this assignment overwrites an already-playable concrete source with the empty result for the opaque ref. For example, `{ src: "https://cdn.example/video.mp4", mediaRef: "ast_missing" }` is silently omitted from the PPTX even though the direct URL is valid. This contradicts the shared video-binding rule that concrete `src` takes precedence. Please retain the existing concrete source when the stored/ref fallback cannot produce a URL, and cover the full `buildPptxBlob` element path rather than only the resolver helper.

lib/export/use-export-pptx.ts
```diff
@@ -360,6 +368,84 @@ function buildSpeakerNotes(scene: Scene): string {
...
+  const tasks = useMediaGenerationStore.getState().tasks;
+  const task =
+    tasks[ref] ??
+    Object.values(tasks).find(
+      (candidate) =>
+        candidate.placeholderRef === ref && (!stageId || candidate.stageId === stageId),
+    );
+  const effectiveTask = task && (!stageId || task.stageId === stageId) ? task : undefined;
+  if (stageId) {
+    const record = await db.mediaFiles.get(mediaFileKey(stageId, ref)).catch(() => undefined);
```
cosarah, line 438: **[P2] Prefer replacement pool bytes over a stale compatibility row**

A same-ID retry deliberately keeps the newly replaced pool bytes when the Dexie compatibility write fails, recording `MEDIA_COMPATIBILITY_STORE_LAGGED`. This helper nevertheless reads Dexie first and immediately returns that row, so PPTX export can use the old blob while the renderer resolves the new pool bytes. The video collector and thumbnail hydration appear to use the same compatibility-first ordering. Please make the asset pool authoritative for allocated refs in this supported lagged state, falling back to the compatibility row only when the pool misses.

lib/import/use-import-classroom.ts
```diff
@@ -15,9 +15,263 @@ import {
...
 
+export interface ImportedMediaMappings {
+  refToNewId: Record<string, string>;
+  posterRefToNewId: Record<string, string>;
+  posterByMediaRef: Record<string, string>;
+}
+
+function rewriteImportedMediaRef(value: string, mapped: string | undefined): string | undefined {
+  if (mapped) return mapped;
+  if (isConcreteMediaAddress(value) || isGeneratedMediaPlaceholder(value)) return value;
```
cosarah, line 35: **[P2] Preserve valid extensionless relative media addresses on import**

`isConcreteMediaAddress()` only accepts relative paths that look extension-bearing, so browser-valid values such as `media/123` or `api/media?id=1` reach the `undefined` return below. `rewriteImportedSlideMediaRefs()` then blanks image/background sources and can remove corresponding video metadata. Imports previously preserved these addresses verbatim. Please recognize valid relative URL forms independently of a filename extension and add extensionless path/query cases to the import tests.

cosarah (commented):
Four additional integration findings from the latest review. The cache-clear finding is marked P2 after reproducing both the partial-clear behavior and the supported retry recovery path.

lib/export/use-export-pptx.ts
wyuc: Fixed in c82dd98e. You were right that patching the helper was not enough — the PPTX element path now goes through the same unified video resolver (`resolveVideoMediaForElement`) the renderers, thumbnails and action path use, so concrete-src precedence is inherited rather than re-derived: a concrete `src` wins over an unresolved `mediaRef`, and the stored-media fallback only runs for an opaque selected source, never overwriting a playable URL. Poster selection uses the same binding.

Your example is pinned at the `buildPptxBlob` level (not the helper): `{ src: 'https://cdn.example/video.mp4', mediaRef: 'ast_missing' }` is embedded via its direct URL. The boundary guard test now also covers this call site, so a future local re-derivation here fails loudly.

lib/utils/database.ts
wyuc: Fixed in c82dd98e. A deferred asset-pool deletion is now treated as **partial completion** rather than total failure: the clear workflow continues with local/session/persisted-store cleanup and still reloads, while returning a distinct deferred result (hard failures still reject and abort). The UI surfaces the deferred warning after the cleanup finishes, so the page no longer sits in a half-cleared state reporting failure.

The pool database is removed once the blocking tab closes, and this page's closed-store behavior stays loud in the meantime — that part is deliberate. `tests/settings/general-settings.test.ts` pins that deferred cleanup continues and stays distinguishable from a hard failure.

lib/import/use-import-classroom.ts
wyuc: Fixed in 609d3c96. `isConcreteMediaAddress()` no longer requires a filename extension: unambiguous relative URL forms (path segments, query strings) are recognized, so `media/123`, `api/media?id=1`, `/path/to/resource` and `path?query=val` are preserved verbatim on import instead of falling through to the blanking branch. Extensionless path and query cases are in the import tests.

lib/export/use-export-pptx.ts
wyuc: Fixed in 609d3c96. The asset pool is now authoritative for allocated refs: the PPTX helper, the video collector and thumbnail hydration all resolve from the pool first and fall back to the Dexie compatibility row only on a pool miss — so the supported `MEDIA_COMPATIBILITY_STORE_LAGGED` state serves the newly replaced bytes everywhere instead of exporting the stale blob while the renderer shows the new one. A lagged-state test covers all three call sites.

cosarah:
Confirmed against the latest PR head (`c82dd98ef34582322903f81201433a59979f1ad5`): this is a real cross-document mutation defect.

`retryMediaTask` determines sharing from `collectStageAssetRefs` over only the active stage (`lib/media/media-orchestrator.ts:126-150`). If an allocated asset ID is referenced once in classroom A and also by classroom B, classroom A's local count is still 1, so `shared` is false. Even when the UI supplies an exact element/scene/slide target, the code chooses `replaceAssetId = elementId` and does not pass the target into the fork/rewrite path.

`BrowserAssetStore.replace` then updates the global registry row for that ID (`packages/@openmaic/storage/src/asset/browser-store.ts:370-395`). Classroom B's document is not rewritten, but its unchanged reference now resolves to classroom A's regenerated bytes.

I reproduced this with a temporary focused Vitest case using two classroom documents that referenced the same allocated ID. After a targeted retry in classroom A, the assertion for classroom B failed deterministically:

```text
Expected: "shared-original"
Received: "retried-target"
```

The existing same-stage duplicated-reference test passed alongside this repro, which isolates the gap: copy-on-write protection works for multiple references within the active stage, but not across persisted documents. Cross-document aliases are valid under the global portable-reference contract and are not rejected by persistence; for example, backup restore saves document references verbatim in `importDatabase`.

Before calling `replaceAsset`, the code must prove exclusivity across persisted documents. If global ownership is unavailable or unknown, a targeted retry should allocate a fresh ID and rewrite only the selected element. A regression test should retain a second surviving document and assert that both its reference and resolved bytes remain unchanged.

wyuc:
Confirmed and fixed in a6d459a6 — thank you for the repro, the cross-document case was a genuine gap.

Ownership is now proven **repository-wide** before any in-place replacement. `retryMediaTask` enumerates every persisted document (via `listDocuments()`, then the same deletion-path enumerator that `collectStageAssetRefs` uses — one enumeration implementation, not a second heuristic) and only calls `pool.replace` when the allocated ref has exactly one logical owner across the whole repository, exactly one owner inside the retry task's own document, and no explicitly supplied target. Every other case — shared across documents, shared within the stage, an explicit element/scene/slide target, or ownership unknown because enumeration failed — allocates a fresh id and rewrites only the selected element. The explicit target is now threaded through the fork/rewrite path, and the same repository-wide check gates poster replacement.

Regression coverage follows your scenario: two persisted classroom documents referencing the same allocated id, targeted retry in A, asserting B's reference string **and** its resolved bytes are unchanged while A's element receives a fresh id; plus an enumeration-failure case that forks rather than replaces. The same-stage duplicate pin and the genuinely-exclusive pin (replace is still taken when the ref is exclusive repository-wide) both stay green.

Cost note: the check is one repository-wide reference enumeration per allocated retry — O(total persisted references), only on that path.

CI re-running on the push.

cosarah:
Confirmed against the current PR head `a6d459a653e0b019d866d79cffc8558b1baef743`. This is a real cross-document reclamation defect.

`loadStageAssetInventory()` reads only rows indexed to the stage being deleted. `buildStageAssetReclamationPlan()` then adds every candidate present in that stage-local `poolOwned` set to `poolRefs`, and `executeStageAssetReclamation()` unconditionally calls `removeAsset()` for those globally keyed refs. The production deletion path does not check surviving documents between deleting the document and executing the plan.

I verified this with a temporary regression test at the exact PR SHA: two persisted documents reference the same allocated ref, and only the deleted stage supplies the compatibility row used to authorize reclamation. `collectPersistedDocumentAssetRefs()` reports a repository reference count of 2, but deleting one stage still calls `removeAsset(sharedRef)`. The focused test fails with:

```
AssertionError: expected vi.fn() to not be called with arguments: [ 'ast_cross_document_alias' ]
```

Command: `vitest run tests/media/reclamation-matrix.test.ts -t "preserves a globally shared pool ref when one owning stage is deleted"`

Fresh `BrowserAssetStore.put()` allocations use random IDs and modern ZIP import re-mints assets, so the affected state is an already-aliased persisted state (for example legacy, copied/restored, or externally authored data), rather than normal new allocation. However, this PR already treats cross-document aliases as valid in retry handling by scanning all persisted documents before allowing same-ID replacement. Stage reclamation should use the same repository-wide liveness rule.

Suggested fix: after the authoritative document deletion, determine refs still referenced by surviving persisted documents and exclude them from global pool removal, failing closed if liveness cannot be established. The deleted stage's compatibility rows can still be removed independently.

wyuc:
Confirmed and fixed in e994daf2 — and you are right that it is the same rule, so it is now literally the same code.

Both paths consume one shared guard, `isAllocatedAssetRefReferencedBySurvivingDocument(ref, excludedDocumentId?)` (in `lib/media/collect-stage-asset-refs.ts`, built on the repository-wide enumeration added for the retry fix). It returns `true` when any persisted document other than the optionally excluded one logically owns the ref, and it **fails closed**: a listing failure, a load failure, or a listed-but-missing document also returns `true`, so a global mutation never proceeds on unknown liveness.

- **Reclamation**: after the authoritative document deletion, `poolRefs` is filtered through the guard — refs still referenced by surviving documents are excluded from global pool removal, while the deleted stage's own compatibility rows are still removed independently.
- **Retry / poster replacement**: the exclusivity decision from a6d459a6 now routes through the same helper, so there is one repository-wide liveness implementation rather than two parallel ones.

Regression coverage follows your scenario verbatim: two persisted documents share an allocated ref with only the deleted stage supplying the compatibility row — `removeAsset` is not called for that ref, the surviving document still resolves its bytes, and the deleted stage's rows are still cleaned up. There is also an enumeration-failure case (bytes preserved, rows still deleted) and the exclusive-owner case still reclaims. All existing reclamation-matrix pins stay green.

CI re-running on the push.

lib/media/collect-stage-asset-refs.ts
```diff
@@ -0,0 +1,269 @@
...
+    const documents = await Promise.all(
+      summaries
+        .filter(({ id }) => id !== excludedDocumentId)
+        .map(async ({ id }) => {
+          const document = await store.loadDocument(id);
+          if (!document) throw new Error(`Listed document ${id} could not be loaded`);
+          return document;
+        }),
+    );
+    return (collectPersistedDocumentAssetRefs(documents).referenceCounts.get(ref) ?? 0) > 0;
```
cosarah: **P2 — Preserve manifest-only owners during reclamation**

`referenceCounts` intentionally excludes manifest metadata, but this liveness check uses only that map. A surviving document can legitimately contain an allocated video ID only as a `stage.videoManifest` key when media finishes before its scene is inserted. I reproduced deletion of another stage calling `removeAsset()` for that shared ID even though the surviving manifest still owns it. Please determine liveness from each surviving document's complete `refs.document` set; keep `referenceCounts` only for operations that need logical-owner multiplicity.

lib/video-export-app/collect.ts
```diff
@@ -161,42 +227,73 @@ function snapshotMediaRef(element: SnapshotMediaElement): string | undefined {
 async function resolveGeneratedMedia(
   source: Slide,
   mediaByElementId: Map<string, MediaFileRecord>,
+  stageId?: string,
 ): Promise<{ slide: Slide; revoke: () => void }> {
   const slide = structuredClone(source);
   const objectUrls: string[] = [];
 
   for (const element of slide.elements as SnapshotMediaElement[]) {
```
cosarah: **P2 — Resolve allocated backgrounds before video snapshots**

This resolver walks only `slide.elements`, so an allocated image background such as `ast_background` reaches `slideToPng` unchanged. I reproduced the snapshot receiving the opaque ID instead of a blob URL. The renderer has no app asset-pool resolver and treats the value as a CSS URL, so the exported base frame loses the background. Please resolve `slide.background.image.src` through the same pool/Dexie path and include its object URL in the revoke lifecycle.

lib/utils/stage-storage.ts
```diff
@@ -777,49 +841,112 @@ export async function getFirstSlideByStages(
         if (firstSlide && firstSlide.content.type === 'slide') {
           const slide = structuredClone(firstSlide.content.canvas);
 
-          const mediaElements = slide.elements.filter((el) =>
-            getThumbnailMediaRef(el as ThumbnailMediaElement),
+          const mediaElements = slide.elements.filter(
```
cosarah: **P2 — Hydrate allocated backgrounds for home thumbnails**

The hydration candidates come only from `slide.elements`, so a persisted/imported slide whose image background was rewritten to an allocated ID never loads the matching media row. The home page renders `SlideThumbnail` without a `MediaStageProvider`; I reproduced the unresolved `ast_background` becoming `src: ''`, which removes the recent-course background. Please hydrate the background alongside element media and revoke its blob URL in `revokeThumbnailSlideMediaUrls()`.

lib/media/reclaim-stage-assets.ts
```diff
@@ -0,0 +1,109 @@
...
+ * Edits and partial deletions must never call this function.
+ */
+export async function executeStageAssetReclamation(
+  plan: StageAssetReclamationPlan,
+  deletedDocument: null,
+): Promise<StageAssetReclamationPlan> {
+  void deletedDocument;
+
+  for (const ref of plan.poolRefs) {
+    if (await isAllocatedAssetRefReferencedBySurvivingDocument(ref)) continue;
```
cosarah: **P2 — Batch surviving-document checks during stage deletion**

This loop invokes a repository-wide `listDocuments()` plus every surviving `loadDocument()` once per pool ref. A focused reproduction with 3 refs and 2 documents called `listDocuments()` 3 times and loaded documents 6 times, confirming O(A×D) document reads. These are real IndexedDB transactions, HTTP requests, or database queries and run while the deletion cascade holds its exclusive runtime lock. Please enumerate surviving documents once, build a complete live-ref set, and test every planned ref against that snapshot.

cosarah (commented):
Four additional P2 findings verified against the current PR head with focused deterministic reproductions.

lib/media/collect-stage-asset-refs.ts
wyuc: Fixed in 23d6fc8c. Liveness is now computed from each surviving document's complete `refs.document` set — every category including `stage.videoManifest` keys — so a document that owns an allocated id only through its manifest keeps the bytes alive. `referenceCounts` is back to serving only the operation that genuinely needs logical-owner multiplicity (the retry exclusivity count). Your scenario is pinned: a surviving document owning the id solely via `videoManifest` is not reclaimed when another stage is deleted.

lib/video-export-app/collect.ts
wyuc: Fixed in 23d6fc8c. `slide.background.image.src` now resolves through the same pool/Dexie path as element media before the snapshot, and its object URL joins the revoke lifecycle.

Since this and the thumbnail-hydration finding share a root cause — the background category keeps being missed by code that iterates `slide.elements` — I added a single exported slot helper (`lib/media/slide-media-slots.ts`) that yields every media-bearing reference slot of a slide (element `src`/`mediaRef`/`poster`, background image, whiteboard entries). The export collector, thumbnail hydration, canonical reference collection, document rewriting and poster lookup all consume it, and the boundary guard test now fails any registered media resolver that walks `slide.elements` directly (the comment states what the regex cannot catch: aliases, destructuring, computed access, helper indirection).

lib/utils/stage-storage.ts
wyuc: Fixed in 23d6fc8c. Thumbnail hydration now loads the background's media row alongside element media and revokes its blob URL in `revokeThumbnailSlideMediaUrls()`, so a persisted/imported slide whose background was rewritten to an allocated id keeps its recent-course background instead of collapsing to `src: ''`. It goes through the shared slot helper described in the export-collector thread, so the two paths cannot drift apart again.

lib/media/reclaim-stage-assets.ts
wyuc: Fixed in 23d6fc8c. Reclamation now enumerates surviving documents once, builds a complete live-ref snapshot (with the manifest-inclusive semantics from the other thread), and tests every planned ref against that snapshot. Fail-closed behavior is unchanged: an enumeration failure treats all refs as live.

Measured on your shape (3 refs / 2 documents): before 3 × `listDocuments()` + 6 × `loadDocument()`; after 1 × `listDocuments()` + 2 × `loadDocument()`, independent of ref count. A test asserts those call counts stay O(1)/O(D) as refs grow, so the per-ref pattern cannot come back under the exclusive runtime lock.

cosarah (commented):
Confirmed by tracing every production retry caller and with a focused differential test.

lib/media/media-orchestrator.ts
```diff
@@ -82,47 +133,505 @@ export async function retryMediaTask(elementId: string): Promise<void> {
...
+    } catch (error) {
+      log.warn(`Could not prove exclusive ownership of media ${elementId}:`, error);
+    }
+  }
+  const referencedByAnotherDocument = allocated
+    ? await isAllocatedAssetRefReferencedBySurvivingDocument(elementId, task.stageId)
+    : false;
+  const activeDocumentCount = activePersistedRefs?.referenceCounts.get(elementId) ?? 0;
+  const exclusive = activeDocumentCount === 1 && !referencedByAnotherDocument;
+  const replaceAssetId = allocated && !target && exclusive ? elementId : undefined;
```
cosarah: [P2] Preserve stable IDs for exclusive targeted retries

Every production retry surface passes a target, so `!target` is false throughout the actual UI path. Even when `exclusive` is true, this forces `targetedFork`, allocates a fresh asset ID, and rewrites the selected element instead of replacing the bytes under the existing ID. The superseded pool entry and its usable compatibility row then remain unreferenced because the cleanup below only removes empty/error rows. I reproduced this with a production-shaped `{ elementId, sceneId, slideId }` retry: `pool.replace` had zero calls. Please allow a globally exclusive allocated ref to use replacement regardless of whether the caller supplied a target, while retaining the cross-document ownership proof.

tests/media/media-orchestrator.test.ts
```diff
@@ -0,0 +1,1475 @@
...
+  it('replaces allocated media bytes without changing the document reference', async () => {
+    const assetId = await pool.put(new Blob(['image-old'], { type: 'image/png' }));
+    mocks.document = documentWithMedia(assetId, 'image');
+    const before = structuredClone(mocks.document);
+    const replace = vi.spyOn(pool, 'replace');
+    const release = vi.spyOn(pool, 'release');
+    serveImage('image-replaced');
+    useMediaGenerationStore.setState({ tasks: { [assetId]: failedTask(assetId) } });
+
+    await retryMediaTask(assetId);
```
cosarah: Please exercise the invocation shape used by production here. This calls `retryMediaTask(assetId)` without a target, but every UI caller passes either an explicit target object or `mediaRetryTarget(...)`. As written, the test covers the only shape that reaches the replacement branch and therefore cannot catch the stable-identity regression. Passing `{ elementId: 'image-1', sceneId: 'scene-1', slideId: 'slide-1' }`, with stage state containing exactly one reference, should still assert `pool.replace`, an unchanged document ref, and no fresh allocation.

lib/media/media-orchestrator.ts
wyuc: You are right, and this one was self-inflicted: I added "no explicitly supplied target" as a precondition for replacement while fixing the cross-document case. Since every production surface passes a target, that condition made the real UI path always fork — disabling stable identity exactly where it matters and stranding the superseded entry. Fixed in 1432d3d6.

A target is no longer a reason to fork; it only selects which element gets rewritten when a fork does happen. Final decision table:

| Ownership state | Result | Target role |
|---|---|---|
| Exactly one owner in the task document and repository-wide | Replace bytes under the existing id | Does not cause a fork |
| Multiple owners in the task document | Fork to a fresh id | Selects the element to rewrite |
| Referenced by another document | Fork to a fresh id | Selects the element to rewrite |
| Ownership/liveness unprovable | Fail-closed fork | Complete scene/slide target required |
| Legacy/unallocated placeholder | Fresh allocation | Scopes the rewrite |

The compatibility row is refreshed under the same id on the replace path, so nothing is left unreferenced there; the fork paths only apply when the old id still has other owners.

tests/media/media-orchestrator.test.ts
wyuc: Agreed — the target-less shape is what hid the regression. Converted in 1432d3d6: the exclusive-owner retry now calls the production shape with `{ elementId: 'image-1', sceneId: 'scene-1', slideId: 'slide-1' }` over stage state holding exactly one reference, and asserts `pool.replace` is called, the id and document reference are unchanged, no fresh allocation happens, and the compatibility row is refreshed.

I also added a differential pin at the same shape where the ref is shared repository-wide — that one forks to a fresh id while leaving the original bytes and compatibility row intact — and swept the rest of the suite: the exclusive video, legacy placeholder upgrade and compatibility-store recovery tests were converted too. No target-less retry calls remain.

lib/media/media-orchestrator.ts
```diff
@@ -82,47 +133,505 @@ export async function retryMediaTask(elementId: string): Promise<void> {
...
+    } catch (error) {
+      log.warn(`Could not prove exclusive ownership of media ${elementId}:`, error);
+    }
+  }
+  const referencedByAnotherDocument = allocated
+    ? await isAllocatedAssetRefReferencedBySurvivingDocument(elementId, task.stageId)
+    : false;
+  const activeDocumentCount = activePersistedRefs?.referenceCounts.get(elementId) ?? 0;
+  const exclusive = activeDocumentCount === 1 && !referencedByAnotherDocument;
+  const replaceAssetId = allocated && exclusive ? elementId : undefined;
```
cosarah, line 146: **[P1] Revalidate exclusivity before replacing a stable asset**

The ownership proof is computed here before the asynchronous image/video request and then trusted when replaceAsset runs. If the slide is duplicated while generation is pending, duplicateSlideScene preserves the media ref, so the asset gains another owner but completion still replaces the bytes behind the shared ID. The selected retry consequently changes both slides instead of forking only the target.

Please re-count ownership immediately before replacement. If the ref is no longer exclusive, allocate a fresh asset and rewrite only the scoped target. Add a regression test for exclusive at start → duplicated while pending → shared at completion.

lib/video-export-app/collect.ts
```diff
@@ -161,42 +228,105 @@ function snapshotMediaRef(element: SnapshotMediaElement): string | undefined {
...
+          blobWithType(posterBytes, posterRecord?.mimeType || posterBytes.type || 'image/jpeg'),
+        );
+        objectUrls.push(poster);
+        element.poster = poster;
+        resolvedPoster = true;
+      } else if (!renderableMediaUrl(resolveMediaRef(element.poster, undefined))) {
+        element.poster = undefined;
+      }
+    }
     const ref = snapshotMediaRef(element);
```
cosarah: **[P2] Resolve video exports from the element-level binding**

snapshotMediaRef(element) can select an opaque stale mediaRef even when element.src is a playable concrete URL. The fallback can then resolve or clear media using the wrong key, while timeline dependency collection also plans play_video against that stale ref. Video export can therefore diverge from live playback and PPTX export.

Please route this boundary through resolveVideoMediaForElement before selecting the reference and bytes. The existing surface-matrix coverage passes an already-selected ref, so it does not exercise this selection boundary.

lib/media/media-task-resolution.ts
```diff
@@ -0,0 +1,194 @@
...
+    stageId && documentElements
+      ? withDocumentLegacyVideoRecovery(tasks, documentElements, stageId)
+      : tasks;
+  const mediaRef = mediaTaskRefForElement(element);
+  const concreteSrc = element.src && isConcreteMediaAddress(element.src) ? element.src : undefined;
+  const sourceRef = concreteSrc ?? mediaRef ?? element.src;
+  const task = concreteSrc
+    ? undefined
+    : resolveMediaTaskForElement(effectiveTasks, element, stageId);
+  const posterRef = element.poster ?? task?.poster;
```
cosarah, line 111: **[P2] Keep explicit posters independent of task fallback**

When element.poster is concrete and task.poster exists, this constructs a task-backed binding for the explicit URL. Consumers then apply the source task state to it: a completed task selects task.poster, while a pending task hides the explicit poster entirely. This reverses the documented element.poster precedence.

Only attach posterTask when element.poster is opaque and actually needs task-backed resolution. A concrete explicit poster should resolve independently. Please add completed and pending cases combining an explicit poster with a task poster.

lib/media/media-orchestrator.ts
wyuc: Confirmed and fixed in b3f33bad. You found the time dimension of a mistake I had been making in space: every earlier fix widened the *scope* of the exclusivity proof (element -> stage -> repository) but kept trusting a proof computed before an asynchronous request. Ownership is now re-proven with the same fail-closed shared guard after the bytes come back and immediately before the first pool write; if exclusivity was lost in the meantime, the completion switches to the scoped fork path, allocating a fresh id and rewriting only the target.

Regression test follows your scenario exactly: exclusive at dispatch, slide duplicated while the request is pending, shared at completion — asserts a fresh id, the target element rewritten, and the other slide's reference and bytes untouched. Videos also revalidate replaceable poster refs.

Cost on the completion path for same-id replacement: one active-document load, one listing, parallel loads of surviving documents, and a linear scan of their references. Allocating and already-forked paths add no enumeration.

lib/video-export-app/collect.ts
wyuc: Fixed in b3f33bad. Snapshot selection now goes through `resolveVideoMediaForElement`, and `timeline-deps` uses the same unified binding, so video export selects the same reference the renderer and PPTX export do instead of an opaque stale `mediaRef`.

Your point about the existing coverage was the important part: the surface matrix passes an already-selected ref, so the selection boundary was untested. There are now two tests at that boundary — `{ src: concrete playable URL, mediaRef: stale opaque }` exports via the concrete src, and the timeline dependency targets the same reference the renderer resolves.

lib/media/media-task-resolution.ts
wyuc: Fixed in b3f33bad. `posterTask` is attached only when `element.poster` is opaque and actually needs task-backed resolution; a concrete explicit poster now resolves independently of task state, restoring the documented precedence. Both cases you asked for are pinned: explicit concrete poster + completed task poster (explicit wins), and explicit concrete poster + pending task (explicit stays visible instead of being hidden).

lib/media/media-orchestrator.ts
cosarah: **[P1] The completion revalidation still misses unflushed editor owners**

This remains reproducible on b3f33bad. The new completion-time proof reloads only the persisted document. Slide duplication updates the Zustand aggregate synchronously, while insertSceneAfter schedules persistence behind the 500 ms debounce. If a slide is duplicated during generation and the response completes inside that window, the persisted document still reports one owner, so proveExclusiveAssetOwnership returns exclusive and replaceAsset overwrites the shared ID.

I reproduced the exact state split with a focused test: one owner in the persisted mock document and two owners in the active stage snapshot. The retry retained the original ID, called replace once, and changed both the pool bytes and compatibility row from image-original to image-retried. The existing pending-generation test does not cover this because it pushes the duplicate into both mocks.document and mocks.stageState before completion.

Please make the final proof require exclusivity in the active stage snapshot when it represents this stage, in addition to the persisted and cross-document checks, and pin the unflushed write-behind case with a regression test.

lib/media/asset-replacement-events.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const observers = new Set<AssetReplacementObserver>();
+
+export function observeAssetReplacements(observer: AssetReplacementObserver): () => void {
+  observers.add(observer);
+  return () => observers.delete(observer);
+}
+
+export async function notifyAssetReplaced(ref: string, pool: AssetReplacementPool): Promise<void> {
+  await Promise.all([...observers].map((observer) => observer(ref, pool)));
```
cosarah: **[P2] Broadcast same-ID replacement invalidations across tabs**

notifyAssetReplaced only iterates this module-local observer set. When the same classroom is mounted in another tab, that tab has a separate JavaScript realm and never receives this notification. Its active asset URL lease therefore remains pinned to the old immutable blob URL after a same-ID replacement; BrowserAssetStore can observe the new bytes only after another resolve, which a continuously mounted consumer does not perform without invalidation.

Please propagate replacement events across tabs, for example with a BroadcastChannel keyed by the asset ref, and have each receiving tab invalidate and re-resolve its active leases. Add a multi-realm test showing that a mounted consumer in tab B refreshes after tab A replaces the same asset ID.

lib/media/media-orchestrator.ts
wyuc: Confirmed and fixed in eb214215 — you are right that reloading the persisted document was still the wrong source of truth for this window.

The completion-time proof now also counts owners in the live stage snapshot, and only when that snapshot represents the stage being retried (`stage.id === stageId`); otherwise it abstains. It reuses `collectStageAssetRefs` over the store's `{ stage, scenes }` rather than adding a second traversal, so the manifest/background/whiteboard categories are counted identically. Exclusivity now requires: exactly one owner in the persisted document, exactly one in the live snapshot when it applies, and no other document referencing the ref.

The regression test models the split state you described — the duplicate is pushed into the stage snapshot only, with the persisted document explicitly asserted to still hold one scene — and checks that the retry forks to a fresh id, the unflushed copy keeps the original id, and the original bytes and compatibility row are untouched. I also mutation-checked it: removing the live-snapshot term makes exactly that test fail.

lib/media/asset-replacement-events.ts
wyuc: Fixed in eb214215. `notifyAssetReplaced` now posts the ref on a `BroadcastChannel` (`maic-asset-replacements`) in addition to running local observers. A receiving realm runs its own observers against **its own** pool, so it invalidates and re-resolves its active leases; the message carries only the ref, which means a spoofed message can at most force a re-resolve and cannot introduce foreign bytes.

Failure handling is deliberate: an environment without `BroadcastChannel`, a channel that fails to construct, or a `postMessage` that throws never fails the replacement itself — those realms simply refresh on their next resolve. Tests cover a peer realm receiving the invalidation for a same-id replacement and the no-channel fallback still notifying local observers.

lib/media/asset-replacement-events.ts
```diff
@@ -0,0 +1,64 @@
...
+async function notifyLocalObservers(ref: string, pool: AssetReplacementPool): Promise<void> {
+  await Promise.all([...observers].map((observer) => observer(ref, pool)));
+}
+
+export function observeAssetReplacements(observer: AssetReplacementObserver): () => void {
+  observers.add(observer);
+  return () => observers.delete(observer);
+}
+
+export async function notifyAssetReplaced(ref: string, pool: AssetReplacementPool): Promise<void> {
```
cosarah, line 75: [P2] Initialize replacement listeners in passive tabs

`ensureChannel` is only called from `notifyAssetReplaced`, which is the sender path. A tab that only renders a classroom can register local observers without ever creating its `BroadcastChannel`, so replacements broadcast by another tab are never received. The test manually creates a peer channel and therefore proves transport delivery, but not the production receiver lifecycle. Please initialize and bind the channel when a passive realm starts observing or opens its asset pool.

lib/media/asset-replacement-events.ts
```diff
@@ -0,0 +1,64 @@
...
+  } catch {
+    // Cross-tab invalidation is an enhancement; a realm without a usable channel
+    // still refreshes on its next resolve.
+    channel = undefined;
+  }
+}
+
+async function notifyLocalObservers(ref: string, pool: AssetReplacementPool): Promise<void> {
+  await Promise.all([...observers].map((observer) => observer(ref, pool)));
+}
```
cosarah, line 68: [P2] Keep observer failures outside the committed write path

`Promise.all` propagates any observer rejection. At the call site, `BrowserAssetStore.replace` has already committed before `notifyAssetReplaced` runs, so a failure such as a stale/offline dynamic import in the cache invalidator makes `replaceAsset` report failure after the durable write succeeded. The orchestrator can then skip its compatibility write while the pool already contains the new bytes. Notifications should be best-effort (for example, per-observer handling or `allSettled`) or explicitly separated from the durable commit result.

lib/media/reclaim-stage-assets.ts
```diff
@@ -0,0 +1,111 @@
...
+    } catch (error) {
+      log.warn(`Failed to reclaim asset ${ref} for stage ${plan.stageId}:`, error);
+    }
+  }
+  if (plan.mediaRowIds.length > 0) {
+    await db.mediaFiles.bulkDelete([...plan.mediaRowIds]);
+  }
+  // Audio deletion is unconditional: a legacy/imported row may have no pool entry.
+  if (plan.audioRowIds.length > 0) {
+    await db.audioFiles.bulkDelete([...plan.audioRowIds]);
```
cosarah: [P2] Preserve compatibility rows for shared audio references

The surviving-document liveness check protects the shared pool entry, but this unconditional deletion still removes the sole globally keyed `audioFiles` row when another persisted document references the same `audioId`. Playback, classroom export, and video export read `db.audioFiles` directly and do not recover from the preserved pool blob. I reproduced this with two documents sharing one allocated audio ID: after deleting one document, the survivor's classroom export collected zero audio files. Please apply the same surviving-reference guard to audio row deletion, including fail-closed behavior when liveness enumeration fails.

cosarah (commented):
Three correctness issues found during the architecture and blast-radius review:

lib/media/asset-replacement-events.ts
wyuc: Fixed in c888e5ec. Channel creation moved out of the sender path into an explicit `bindAssetReplacementChannel(poolFactory)`, called where the observer is registered — when the `asset-pool` module loads, behind a `typeof window` guard so SSR is unaffected. A tab that only renders therefore starts listening as soon as it touches the pool module, without ever sending anything. The receiving realm resolves its own pool lazily and skips the notification if the pool is unavailable (clearing, or no IndexedDB) rather than throwing.

You were right that my earlier test only proved transport: it created the peer channel by hand. There is now a test in `tests/media/asset-pool.test.ts` that pins the production lifecycle — it imports the module, then has a peer post a message, and asserts local observers ran. I mutation-checked it: removing the bind makes exactly that test fail, whereas my previous suite did not catch the removal at all.

lib/media/asset-replacement-events.ts
wyuc: Fixed in c888e5ec. Notifications are best-effort now: observers are settled individually and failures are logged rather than propagated, so a replacement that has already committed in `BrowserAssetStore.replace` can no longer be reported as a failure and cannot make the orchestrator skip its compatibility write.

One detail worth flagging because it bit me while testing: an observer that throws **synchronously** escapes `Promise.allSettled` — the throw happens inside the `.map` callback, before the array exists. The observer invocation is wrapped so the synchronous case is captured too, and the regression test uses a synchronously-throwing observer (the shape your stale dynamic import example produces) alongside a healthy one, asserting the healthy observer still runs and the call resolves.

lib/media/reclaim-stage-assets.ts
wyuc: Confirmed and fixed in c888e5ec. `audioFiles` rows are keyed globally by `audioId`, so the unconditional `bulkDelete` removed the only row for an id a surviving document still referenced — and as you noted, playback, classroom export and video export read that table directly, so the preserved pool blob does not rescue them.

Rows are now filtered against the same surviving-reference set that guards pool removal: a row still referenced by another document stays, one nothing references is removed, including legacy/imported rows that never had a pool entry. The fail-closed path keeps its existing contract — a failed enumeration withholds only the irreversible pool removal, while the deleted stage's own rows are still cleaned up.

Two regression tests: two documents sharing an allocated audio id (the survivor keeps both the pool entry and its row), and an exclusive plus a legacy row (both removed). Mutation-checked — restoring the unconditional delete fails the shared-row test.

lib/media/reclaim-stage-assets.ts
```diff
@@ -0,0 +1,123 @@
...
+  // surviving document must stay: playback, classroom export and video export
+  // read this table directly and cannot fall back to the preserved pool blob.
+  // Rows no survivor references are still removed, including legacy/imported
+  // ones that never had a pool entry.
+  // Failing to enumerate survivors withholds the irreversible pool removal, but
+  // the deleted stage's own rows are still cleaned up, as they always were.
+  if (plan.audioRowIds.length > 0) {
+    const removableAudioRowIds =
+      liveRefs === null
+        ? [...plan.audioRowIds]
```
cosarah: **[P2] Preserve shared audio rows when liveness enumeration fails**

`null` explicitly means that survivor liveness is unknown, but this branch treats every planned audio row as removable. If `listDocuments()` or any `loadDocument()` fails after the target document has been deleted, pool removal correctly fails closed while `bulkDelete` still removes the sole globally keyed `audioFiles` row. A surviving document sharing that `audioId` then loses playback, classroom export, and video export because those consumers read Dexie directly and cannot recover from the retained pool blob.

I reproduced this deterministically with two documents sharing one allocated audio ID and a rejected repository listing: the pool entry survived, but `audioFiles` became empty. Unknown liveness must preserve these rows too (for example, make the removable set empty when `liveRefs === null`), accepting bounded garbage until a later cleanup can prove exclusivity.

lib/media/media-orchestrator.ts
```diff
@@ -131,62 +692,259 @@ async function generateSingleMedia(
...
+              createdAt,
+            }),
+          ),
+        );
+      }
+    } catch (error) {
+      if (activeReplaceAssetId && poolReplacementCommitted) {
+        throw new MediaApiError(
+          `Asset pool replacement succeeded but the compatibility media store lagged: ${error instanceof Error ? error.message : String(error)}`,
+          'MEDIA_COMPATIBILITY_STORE_LAGGED',
```
cosarah, line 770: **[P2] Keep classroom ZIP exports consistent after compatibility-store lag**

When a same-ID replacement has committed new pool bytes but this compatibility write fails, the document intentionally remains on the same ID and the task records `MEDIA_COMPATIBILITY_STORE_LAGGED`. Rendering, PPTX export, and video export then resolve the new pool bytes, but classroom ZIP export still calls `collectMediaFiles(stageId)` and serializes the stale Dexie row. The export succeeds while containing different media from what the classroom currently renders.

The earlier pool-first fix covered PPTX, video collection, and thumbnail hydration, but not classroom ZIP collection. This state needs either a repair of the compatibility row before ZIP export or pool-first resolution for the document references included in the ZIP.

lib/media/reclaim-stage-assets.ts
wyuc: Confirmed and fixed in a9887590. My justification for the previous behaviour was wrong: I treated the row deletion as recoverable because the pool entry survived, but for the consumers that matter here — playback, classroom export, video export — the row **is** the byte source, so deleting it is exactly as irreversible as removing the pool entry. Unknown liveness (`liveRefs === null`) now preserves every planned audio row, leaving bounded garbage for a later pass that can prove exclusivity, which matches the pool's fail-closed behaviour.

Two regression tests: a shared audio id with a rejected repository listing (pool entry and row both survive), and the pre-existing fail-closed case updated to assert row preservation instead of removal. Mutation-checked: restoring the unknown-liveness deletion fails both.

lib/media/media-orchestrator.ts
wyuc: Fixed in a9887590 — you are right that the earlier pool-first pass covered PPTX, video collection and thumbnail hydration but not classroom ZIP. `collectMediaFiles` now resolves each allocated ref through the pool and only falls back to the compatibility row on a miss, so a lagged same-id replacement exports what the classroom actually renders. Legacy placeholder rows keep their stored bytes unchanged.

Following your finding I audited every direct reader of `db.mediaFiles`/`db.audioFiles` for the same gap, and found one more that you had not flagged: **playback**. Speech regeneration also replaces bytes under a stable id (`generateAndStoreTTS` calls `replaceAsset` and deliberately does not roll the pool back when the Dexie write fails), while `AudioPlayer` read the row directly — so a learner would keep hearing the superseded narration. That path is now pool-first with the same stored-row fallback for legacy and imported audio.

Tests for both, each mutation-checked. Remaining direct readers were checked individually: the write paths, the import/export mirrors and the reclamation enumerator do not need pool precedence.

lib/audio/regenerate-speech-tts.ts
```diff
@@ -56,37 +65,19 @@ export async function audioObjectUrl(audioId: string): Promise<string | null> {
...
 ): Promise<string | null> {
   if (!isManagedTtsActive()) return null;
   const text = action.text?.trim();
   if (!text || !action.id) return null;
-  const audioId = speechAudioId(sceneOrder, action.id);
-  await generateAndStoreTTS(audioId, text, language, signal);
-  return audioId;
+  const requestId = `tts_request_s${sceneOrder}_${action.id}`;
+  const stageId = useStageStore.getState().stage?.id;
+  return generateAndStoreTTS(requestId, text, language, signal, undefined, undefined, stageId);
```
cosarah: The existing `action.audioId` is accepted here but never used: `undefined` is always supplied as `replaceAssetId`. Both the single-line and batch callers pass the previous ID, so regenerating an exclusively owned, pool-backed clip always allocates a new asset, rewrites the action, and leaves the previous pool entry and `audioFiles` compatibility row orphaned until stage reclamation. This also contradicts the stable-ID replacement path already implemented and tested in `generateAndStoreTTS`.

Please establish ownership before synthesis: when the current `audioId` exists in the pool and is provably exclusively referenced (including persisted, unflushed, and cross-document references), pass it as `replaceAssetId`; when it is shared or ownership cannot be proved, keep the fresh-allocation/fork behavior. The regeneration tests should cover both branches instead of pinning fresh allocation for every existing ID.

wyuc: Confirmed and fixed in 21d4d3e7. Ownership is now established before synthesis: when the action's `audioId` exists in the pool and `proveExclusiveAssetOwnership` clears it, that id is passed as `replaceAssetId` and the bytes are replaced in place; a shared clip, a legacy id with no pool entry, or unprovable ownership keeps the fresh-allocation path so other holders keep their audio.

Rather than restate the rule in the audio path, I moved `proveExclusiveAssetOwnership` (with its persisted, unflushed-snapshot and cross-document terms) out of `media-orchestrator` into the shared reference module, so media retry, poster replacement and speech regeneration now consume one implementation.

Tests cover both branches as you asked — exclusive (asserts the id is passed through and reused), shared (asserts a fresh allocation and that the shared clip is untouched), and a legacy id with no pool entry (asserts ownership is not even consulted). Mutation-checked: forcing `replaceAssetId` back to `undefined` fails the exclusive case.

I also swept every in-place replacement entry point: `replaceAsset` has exactly three call sites (media, poster, TTS), all behind this proof, and `generateAndStoreTTS`'s `replaceAssetId` has a single caller. No other path mutates a globally keyed asset without it.

lib/hooks/use-scene-generator.ts
```diff
@@ -318,20 +338,66 @@ export async function generateAndStoreTTS(
...
-    createdAt: Date.now(),
-  });
+    language,
+    provider: {
+      id: settings.ttsProviderId,
+      model: ttsProviderConfig?.modelId,
+    },
+  } as const;
+  const assetId = replaceAssetId ?? (await putAsset(blob, assetMeta));
+  if (replaceAssetId) await replaceAsset(replaceAssetId, blob, assetMeta);
```
cosarah, line 357: **[P2] Keep every audio consumer on the replaced pool bytes**

When stable-ID TTS regeneration reaches `replaceAsset` but the following `audioFiles.put` fails (for example, quota pressure or a transient IndexedDB error), the pool has already committed the new narration while the compatibility row remains stale. The existing action deliberately keeps the same ID, and `AudioPlayer` resolves pool-first, but `audioObjectUrl`, `collectAudioFiles`, and `createVideoTimelineDeps` still read `db.audioFiles` directly. I reproduced this exact failure path: editor preview, classroom ZIP export, and video-export dependencies all returned `old-narration` while the pool contained `new-narration`.

Please route all allocated-audio readers through one pool-first resolver, with Dexie retained only as the legacy/import fallback, and add a regression test where replacement succeeds but the compatibility overwrite rejects.

lib/media/collect-stage-asset-refs.ts
```diff
@@ -0,0 +1,316 @@
...
+  } catch (error) {
+    log.warn(`Could not prove exclusive ownership of asset ${assetId}:`, error);
+  }
+  const referencedByAnotherDocument = await isAllocatedAssetRefReferencedBySurvivingDocument(
+    assetId,
+    stageId,
+  );
+  // The live snapshot only participates when it represents this stage; when it
+  // does, an owner it knows about counts even though persistence is pending.
+  const liveOwners = unflushedStageOwnerCount(assetId, stageId);
```
cosarah, line 282: **[P2] Include unflushed owners from other tabs in the exclusivity decision**

`proveExclusiveAssetOwnership` combines the current tab\x27s live Zustand snapshot with persisted-document liveness, but `unflushedStageOwnerCount(assetId, stageId)` only models the active realm. If another tab has duplicated or rebound the same classroom asset during its persistence debounce, this tab can still observe `liveOwners === 1` and replace the globally shared ID in place before the other tab flushes. The other tab can then persist a second owner of bytes that have already been mutated. The existing unflushed-owner test covers one realm only; please add a cross-tab/repository pending-state case or make the ownership proof fail closed when other realms cannot be observed.

components/settings/general-settings.tsx
```diff
@@ -89,7 +93,7 @@ export function GeneralSettings() {
       window.location.reload();
```
cosarah, line 99: **[P2] Do not reload away the deferred-clear retry state**

When `clearAssetPool()` reports a blocked/deferred deletion because another tab still has `maic-asset-pool` open, this branch correctly shows actionable guidance to close the other tab and retry. The unconditional reload immediately afterward discards that state, while the asset-pool database is still undeleted. The user can therefore lose the retry context and land back in the settings page with the original cache still present. Please return without reloading for the deferred result, or persist an explicit retry-pending state across the reload and restore the retry flow on startup. The current tests stop at `runClearCache` and do not exercise this component-level branch.

lib/hooks/use-scene-generator.ts
wyuc: Fixed in e1af38ca. Since this is the fifth finding of the same shape, I routed all allocated-audio readers through one resolver (`lib/media/resolve-audio-bytes.ts`) instead of patching each: `audioObjectUrl`, `collectAudioFiles`, `createVideoTimelineDeps` and `AudioPlayer` now share it, with Dexie kept as the fallback for legacy and imported rows that were never pool-backed.

Regression coverage sits on the resolver itself (pool wins over a lagging row; stored row serves legacy ids; concrete addresses never consult the pool; null when neither store has bytes) plus the existing player test, and is mutation-checked — dropping the pool branch fails both the resolver and player cases.

lib/media/collect-stage-asset-refs.ts
wyuc: Fixed in e1af38ca, taking your second option. A peer's unflushed state genuinely cannot be observed — it lives in another realm and leaves no persisted trace during its debounce — so rather than trying to read it, the proof now asks whether any other realm currently has the stage open and treats that as disqualifying.

Presence is a small BroadcastChannel handshake (`lib/media/stage-realm-presence.ts`, bound where the pool binds its replacement channel, so passive tabs answer too), and `proveExclusiveAssetOwnership` gains a `!peerRealmEditing` term. The effect is that a peer editing the same classroom forces the fork path instead of a global mutation decided from state we cannot see.

Tests: presence returns true for a peer holding the same stage, false for a peer on a different stage and when nobody answers; and a unit test on the proof itself asserts a peer realm makes it non-exclusive, alongside the persisted, cross-document and unflushed-owner cases. Mutation-checked — removing the peer term fails that case.

components/settings/general-settings.tsx
wyuc: Fixed in e1af38ca. The deferred branch now returns without reloading, so the guidance to close the other tab and retry stays on screen while the pool database is still on disk; only a completed clear reloads.

I moved the decision into `shouldReloadAfterClear` in the workflow module rather than leaving it inline — the first version of this fix was inline and survived a mutation because the settings tests only reach the pure layer. With the rule extracted, forcing it to always reload now fails a test.

wyuc:
## Proposal: land this and move the remaining polish to follow-ups

All review findings to date are addressed and CI is green. I would like to propose a landing criterion rather than continuing open-ended, and I want to be explicit about why.

**Where this stands.** Every data-integrity invariant this PR introduces is now pinned by a mutation-verified regression test:

- bytes are never lost — reclamation preserves anything a surviving document references, and fails closed when liveness cannot be established;
- a globally keyed asset is never mutated on behalf of one owner — exclusivity is proved across persisted, unflushed, cross-document and cross-realm state, and any doubt forks to a fresh id;
- generation is never silently repeated — resume consults document truth, so reopening a classroom makes no new API calls;
- no consumer serves superseded bytes — media and audio readers resolve pool-first through one shared resolver each, with the compatibility rows kept as the legacy fallback;
- no opaque reference reaches a DOM `src`, and no consumer blanks a concrete address it cannot resolve.

The last three review rounds have produced no P1: the findings have moved from data loss to consistency polish (a background image missing from an export frame, a reload that discards retry guidance). Those matter, but they are recoverable and visible, unlike the class this PR started with.

**What I propose.** Unless the next round surfaces another data-integrity defect, land this as is and track the remaining polish as follow-up issues. Two reasons beyond the state of the code:

1. Part 2 step (c) — the reference converter, `audioUrl` removal and the `DSL_VERSION` bump — is blocked behind this, and it is the step that actually migrates existing documents. Holding (b) open delays the work that carries user data.
2. At 103 files this PR has outgrown what a single review pass can reliably cover, which is itself a source of the churn: several findings were introduced by my own fixes to earlier findings. Smaller units will review better.

**If you would rather not.** I am equally happy to split the cross-realm consistency work (presence probing, replacement broadcast, lease refresh) into its own PR and land the core write-path change first — that is the part where the remaining edges live, and without it the fallback is simply "a peer tab refreshes on its next resolve", not corruption. Say which you prefer and I will do it.

Either way, thank you for the depth here — the cross-document replacement and reclamation defects in particular were silent data corruption that would have been extremely hard to trace in the field.

lib/media/stage-realm-presence.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+/**
+ * True when another realm answers that it has this stage open, and also true
+ * when presence cannot be established at all — an unanswerable question must
+ * not be read as "nobody else is editing".
+ */
+export async function isStageOpenInAnotherRealm(stageId: string): Promise<boolean> {
+  if (typeof BroadcastChannel !== 'function') return false;
```
cosarah: [P2] Fail closed when the presence transport is unavailable

The contract above says unknown presence must force allocation of a fresh ID, but these early returns map both an unsupported API and an unbound channel to `false` ("no peer"). `bindStageRealmPresence` also catches constructor failures and leaves `channel` undefined, and the fire-and-forget binding in `asset-pool.ts` creates a real not-yet-bound interval. The `postMessage` catch below has the same fail-open behavior.

I verified all three initialization cases against this PR head: `BroadcastChannel` unavailable, binding not yet completed, and a constructor that throws. In every case, `proveExclusiveAssetOwnership()` returned `exclusive: true` for a single persisted/local owner. A production-shaped media retry then called `BrowserAssetStore.replace(existingId, ...)` instead of allocating and rewriting a target-scoped ID, so another tab's unflushed owner can silently receive the new bytes.

Please represent unavailable/failed probing as unknown and fail closed in the ownership decision (or conservatively return `true` here), including the send-failure path, and add coverage that reaches the actual ownership/write boundary.

wyuc: Confirmed and fixed in 9f079017. My own docstring said an unanswerable probe must count as a peer while the implementation mapped every unavailable path to `false` — you verified all three initialization states, and they did reach `replace`.

The probe now returns `'present' | 'absent' | 'unknown'`, so the two cases cannot be collapsed by accident: only a probe that was actually sent and went unanswered is `absent`; a missing `BroadcastChannel`, a constructor that threw, a send failure, and the not-yet-bound window are all `unknown`. The ownership decision reads `!== 'absent'`, treating unknown exactly like present.

For the binding race you flagged: the pool now declares its intent synchronously (`expectStageRealmPresenceBinding`) before the dynamic import, so a probe in that window awaits the bind rather than concluding presence is unavailable, and the gate is released if the import fails so probes fall through to `unknown`.

Coverage reaches the write boundary as you asked: the three initialization states are pinned on the probe, and an end-to-end orchestrator test asserts that with presence `unknown` a production-shaped targeted retry forks to a fresh id, never calls `replace`, and leaves the original bytes intact. Mutation-checked — collapsing unknown into absent fails both the unit and the write-boundary test.

cosarah (commented):
## Additional P2 findings

### 1. Preserve extensionless relative media addresses — confirmed

The current classifier in `lib/media/resolve-media-ref.ts:30` only recognizes browser-relative addresses when they contain a file extension. This regresses valid extensionless relative references such as `media/123`, `api/media?id=1`, and `path?query=val`.

I reproduced the issue against the current PR worktree:

- `isConcreteMediaAddress('media/123')` returns `false`
- `resolveMediaRef('media/123', undefined, { status: 'missing' })` resolves to a placeholder instead of a raw URL
- `rewriteImportedSlideMediaRefs` rewrites `media/123` and `path?query=1` to empty strings

The import path then clears image/background `src` values and removes video `mediaRef`/poster values. The base revision had explicit regex branches and a parameterized test for these cases; both were removed by this patch.

Please restore the extensionless-relative classification and its regression coverage.

### 2. Resolve audio from the pool at collection time — likely P2, needs a focused repro

In `lib/video-export-app/collect.ts:401-404`, audio bytes are collected from records loaded during timeline planning. If stable-ID TTS regeneration replaces the pool bytes after planning but before collection, this path can embed the superseded narration while playback and other export paths use the current pool bytes. The precomputed duration can also disagree with the collected audio.

The collector should resolve the current audio reference through the shared pool at the byte boundary, or otherwise revalidate the reference/bytes immediately before embedding them. A concurrency regression test should replace the pool entry between planning and collection and assert that the exported bytes and duration come from the replacement.

cosarah (approved):
Approved for the latest remote PR head `9f079017`.

I reviewed the committed diff against `origin/main` along both standards and spec axes. I found no new blocking correctness issues: the extensionless relative-media compatibility and pool-first audio-resolution fixes are present, the asset allocation/reclamation/import invariants match the PR requirements, and the latest CI checks are green.

There is one low-priority maintainability observation: retry/media-state UI logic is duplicated across several renderer surfaces, but the surfaces have intentionally different layouts and this is not a merge blocker.

Note: this approval applies to the committed remote PR head `9f079017`. The local worktree contains a separate uncommitted staged overlay that is not part of this PR head and should not be included without a separate review.

cosarah (dismissed):
LGTM

cosarah (approved):
LGTM

Issue #1021, opened 2026-07-31
Release path: publish the validated artifacts, and make retries and races safe


wyuc:
Three findings from adversarial review of the release path. None is introduced by a recent change; all describe `main` as it stands. They share one root: the publish job reconstructs what it publishes instead of publishing what was validated.

**1. The published tarball is not the validated one.** `validate` builds, packs and smoke-tests, then `publish` performs a fresh install and build and asks `pnpm publish` to create another tarball from the directory. The integrity check between them covers tracked files only, and every package gitignores `dist`. A build that becomes environment- or time-dependent emits validated bytes in one job and different bytes in the other, with every check still green.

Fix: pack once in `validate`, hand the `.tgz` files to `publish` as artifacts with their digests, and publish those exact files. This also removes the need for the publish job to install or build at all, which shrinks what executes next to the registry token.

**2. A retry from newer `main` can absorb later work into an older version.** Release planning compares current manifests against the registry. If a release publishes dsl and then fails before renderer, and a later pull request changes renderer source while its manifest still holds the version that failed to publish, a new run from current `main` publishes a tarball containing the later change under the older version and changelog. Retries need to run from the original release commit rather than from whatever `main` has become.

**3. The registry snapshot is taken well before publication.** The monotonicity check reads the registry while planning, and publication happens after install, build, tests and packing. Another publisher releasing a higher version inside that window is not detected; npm rejects only a name and version that already exist, and an ordinary publish moves `latest`, so consumers can be moved backwards. Re-check immediately before each publish, or publish under a temporary dist-tag and promote after a final check.

Items 2 and 3 are narrower than 1 and could be handled independently, but 1 changes the shape of the job in a way that makes both easier to reason about.

yipwingtim:
hi,I'd like to work on this issue.
My plan is to address finding 1 first by packing the tarballs once in the validation job, smoke-testing those exact files, uploading them with SHA-256 digests, and having the publish job verify and publish only those artifacts. This should also remove dependency installation and builds from the token-bearing publish job.
I can then handle findings 2 and 3 on top of that release flow. Please assign this issue to me if that approach sounds good.

wyuc:
Part 1 is implemented in #1040: the publish job no longer installs or builds, and publishes the exact tarballs that `validate` packed and uploaded.

[...]

PR #1040, opened 2026-08-03
Publish validated package tarballs

wyuc:
## Summary

- Pack each of the four allowlisted `@openmaic/*` packages exactly once in `validate` with lifecycle scripts disabled, record SHA-256 digests, smoke-test those exact tarballs, verify them again, and upload them as one workflow artifact.
- Download the artifact in `publish`, fail closed unless it contains exactly the four expected name/version tarballs plus `SHA256SUMS`, re-verify every digest, and publish those exact `.tgz` files sequentially with `npm publish <tarball>`.
- Remove pnpm setup, dependency installation, and package builds from `publish`. Keep the registry preflight in the token-bearing job immediately before the sequential publish loop.
- Anchor post-install/build whole-tree integrity checks to immutable `GITHUB_SHA`, reject a locally moved `HEAD`, and disable Git replacement objects during commit resolution and comparison. The equivalent package-scoped CI install check uses the same protection.
- Keep provenance enabled. npm documents tarballs as accepted `npm publish` package specs and `--provenance` as applying to the publish; npm 11.16 also derives the provenance subject digest from the supplied tarball bytes. A local dry run accepted the prebuilt tarball. See [npm publish](https://docs.npmjs.com/cli/publish/) and [Generating provenance statements](https://docs.npmjs.com/generating-provenance-statements/).
- Disable persisted checkout credentials in all three jobs. The marker job remains the sole `contents: write` holder and creates tags through the GitHub API without installing or building anything.

## Security impact

This structurally removes three findings that depended on the token-bearing publish job running install and build code:

- build code tampering with the git index;
- validated bytes differing from published bytes; and
- build code rewriting enforcement scripts before the token-bearing step uses them.

The publish job now runs no install or build command and sends only digest-verified artifacts from `validate` to npm.

The integrity checks no longer trust local `HEAD`. These checks exist because repository and build code run before them; that code can create a commit and move `HEAD` without any GitHub credential. The runner-provided `GITHUB_SHA` is the immutable release reference, and `--no-replace-objects` prevents a local replacement ref from redirecting the comparison.

## Known limitation

The smoke test reads the writable local tarball directory, not the uploaded snapshot. Because upload happens first, a test process can replace a poisoned tarball locally with a valid one after upload; the smoke test can then pass against the replacement while `publish` downloads and publishes the poisoned uploaded snapshot, whose own `SHA256SUMS` remains internally consistent.

This cannot be closed while packing, uploading, and testing all run in one job with a writable filesystem. Closing it requires a separate packing-only job that runs no tests. That structural change is not attempted here.

## Scope

This implements item 1 of #1021. It does not attempt the retry semantics in item 2 or the broader registry-race changes in item 3. The existing registry preflight is only placed immediately before publication, as requested, to reduce its current race window.

Part of #1021

## Verification

- Node 22.22.0 and pnpm 10.28.0.
- `pnpm install --frozen-lockfile`: passed across 7 workspace projects; the lockfile was current, dependencies were already up to date, and postinstall completed.
- `pnpm check`: 0 formatting mismatches.
- `pnpm lint`: 0 errors, 15 existing warnings.
- `pnpm exec tsc --noEmit`: passed.
- Four package suites with `VITEST_MAX_WORKERS=4`: 58 files passed, 2 skipped; 1,125 tests passed, 59 skipped.
  - dsl: 6 files, 196 tests passed.
  - storage without PostgreSQL requirement: 16 files passed, 2 skipped; 654 tests passed, 59 skipped.
  - renderer: 25 files, 242 tests passed.
  - importer: 11 files, 33 tests passed.
- PostgreSQL-required storage rerun: 18 files and 713 tests passed. Both contract files ran (32 and 27 cases), and the baseline audit observed positive insert deltas in all 5 owned tables: 24, 51, 24, 36, and 30.
- Gate scripts passed explicitly: vendored importer assertion, i18n alignment (9 locales), internal dependency ranges (3 dependents), package allowlist (4 packages and every workflow enumeration), package version diff plus serialized-format rule, release registry preflight (0 packages currently planned), PostgreSQL contract audit, exact-tarball smoke test, and artifact digest verification (4 tarballs).
- Workflow audit: 4 YAML files parsed; all 38 `run` blocks passed `bash -n`.
- Release invariant audit: 3 jobs with explicit permissions; 3 checkouts with `persist-credentials: false`; 1 `contents: write` holder and 1 tag writer, both `mark`; 9 action uses pinned to full 40-hex SHAs; `publish` has 0 install commands, 0 build commands, and 0 contents scope; upload precedes both test steps; both whole-tree checks use `GITHUB_SHA`, assert `HEAD` equality, and disable replacement objects.
- The remaining release invariants were rechecked: release environment, digest anchoring before repository scripts, per-package digest verification immediately before each sequential publish, registry preflight immediately before the loop, `--provenance`, first-parent membership with the `grep -cx`/SIGPIPE safeguard, exact-SHA CI lookup keyed by `ci.yml` and `event=push`, PostgreSQL baseline/delta audit, repository-wide concurrency, and no tag trigger.
- Fault injection at commit `34d271c1e53154ad4f4d56034ebc5b3de720fb50`: the ordinary clean tree passed with exit 0. A local commit moved `HEAD` to `475e556e289273e60561f887021722a865c0cf5a`; the guard printed `HEAD moved from ... to ... during the build` and failed with exit 1.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,15 +80,23 @@ jobs:
       # rewrites one; catching it here means a stale generated file is a pull
       # request failure instead of a release failure.
       #
-      # Against HEAD, not the index: a bare `git diff` compares the worktree to
-      # the index, so anything the install staged would be invisible to it.
+      # Against GITHUB_SHA, not the index or HEAD: a bare `git diff` misses
+      # staged changes, while install code can create a commit and move HEAD.
+      # This check exists precisely because code ran before it, so its reference
+      # point must be something that code cannot move. Disabling replacement
+      # objects also prevents a local replace ref from redirecting the commit.
       - name: Verify the install did not rewrite tracked package files
         run: |
           set -euo pipefail
-          if ! git diff --quiet HEAD -- packages/@openmaic; then
+          actual_head="$(git --no-replace-objects rev-parse HEAD)"
+          if [ "$actual_head" != "$GITHUB_SHA" ]; then
+            echo "::error::HEAD moved from $GITHUB_SHA to $actual_head during installation."
+            exit 1
+          fi
+          if ! git --no-replace-objects diff --quiet "$GITHUB_SHA" -- packages/@openmaic; then
             echo "::error::Installing rewrote tracked files under packages/@openmaic."
             echo "Regenerate them and commit the result."
-            git --no-pager diff --stat HEAD -- packages/@openmaic
+            git --no-replace-objects --no-pager diff --stat "$GITHUB_SHA" -- packages/@openmaic
             exit 1
           fi
           echo "Tracked package files match the commit under test."
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,8 +65,8 @@ jobs:
           BEFORE_SHA: ${{ github.event.before }}
 
       # The tarball smoke test proves this about the bytes that would actually
-      # be published, but it packs and installs four packages, so it runs only
-      # on the release path. This is the cheap source-level form, here so that a
+      # be published, but it installs four package tarballs, so it runs only on
+      # the release path. This is the cheap source-level form, here so that a
       # pull request restoring `workspace:*` fails at review time rather than at
       # release time, after a version number has already been spent.
       - name: Internal dependency ranges
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -334,12 +334,9 @@ jobs:
           set -euo pipefail
           # Anchor the validated digests in this shell before any repository
           # script runs with the npm token. A child process can alter files,
-          # but it cannot rewrite its parent shell's associative array.
+          # but it cannot rewrite its parent shell's scalar.
           node scripts/verify-package-artifacts.mjs "$RELEASE_ARTIFACTS"
-          declare -A trusted_digests
-          while read -r digest filename; do
-            trusted_digests["$filename"]="$digest"
-          done < "$RELEASE_ARTIFACTS/SHA256SUMS"
+          trusted_digests="$(< "$RELEASE_ARTIFACTS/SHA256SUMS")"
 
           node scripts/check-package-version-bumps.mjs --release
           for pkg in dsl storage renderer importer; do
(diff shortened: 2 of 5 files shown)


wyuc:
## Review summary

Three rounds of adversarial review, framed as "enumerate how to defeat this" rather than "confirm it works".

**Round 1** found that tests ran between the integrity check and packing, so a passing suite could rewrite an included file and have it packed, digested and smoke-tested as if it were clean. The previous flow contained that accidentally, because the publish job did its own fresh checkout and build, so removing that redundancy also removed a protection nobody had written down. It also found the downloaded tarballs were verified once and then left unverified while other steps ran, and that the new artifact boundary depended on movable action tags.

**Round 2** found that sealing the tarballs early was not enough, because `SHA256SUMS` sat on the same writable filesystem: replacing a tarball and rewriting its digest row is self-consistent and passes. The fix was structural rather than another check. Artifacts are now uploaded immediately after packing, before any repository test code runs, and `needs: validate` still prevents publication if a later step fails. Nothing can alter what has already been uploaded, so the question of whether the digest file can be trusted stops needing an answer.

**Round 3** found that the integrity check compared the worktree against local `HEAD`, which build code can move by making a local commit, no credential required. It now compares against `$GITHUB_SHA` and additionally asserts that `HEAD` equals it, so a moved `HEAD` is itself an error. The check exists because build code runs before it, so its reference point must be something that code cannot move.

Each fix was fault-injected in both directions.

## What this structurally removes

The publish job no longer runs `pnpm install` and no longer builds. Three findings that earlier rounds kept re-deriving depended on it doing so, and they are now gone rather than guarded: build code tampering with the git index so the integrity check goes blind, the published bytes differing from the validated ones, and build code rewriting the enforcement scripts before the token-bearing step uses them.

## Known limitation

The smoke test reads the local tarball directory rather than the uploaded snapshot. Since upload now happens before the tests, a test process could replace a poisoned tarball locally with a valid one after upload; the smoke test would pass against the replacement while the publish job publishes the uploaded snapshot, whose own digests remain internally consistent.

This cannot be closed while packing, uploading and testing share one job and a writable filesystem. Closing it means packing in a job of its own that runs no tests, which is a structural change tracked as follow-up on #1021 rather than stretched into this pull request.

xuyuanwei678 (approved):

PR #1042, opened 2026-08-03
fix(render-service): make parallel capture effective

### LINKED ISSUE #1041: [Bug]: Render service defaults defeat parallel capture and fresh containers fail font-cache initialization
## Summary

The Docker render-service defaults currently prevent the documented fast-capture path from working reliably, and a fresh container may fail before capture begins because the font cache resolves under `/root` after privilege dropping.

This issue tracks the related fixes as one render-pipeline hardening effort:

1. Fix the font-cache `HOME`/permissions failure in fresh containers.
2. Make the Compose memory profile consistent with `PRODUCER_BROWSER_GPU_MODE=hardware`.
3. Make producer frame workers actually parallel instead of sharing a bottlenecked Chromium browser pool.
4. Avoid long-render CDP timeout/fallback behavior and remove the fixed cost of static-dedup verification when it cannot complete.

## Environment

- OpenMAIC: `main` at `297e405d`
- `@hyperframes/producer`: `0.7.60`
- Docker image: current `render-service/Dockerfile`
- Node: 22.22.2
- Chromium: 150.0.7871.124
- Reference project: OpenMAIC classroom export, 1280x720, 30 fps, 640.65 s, 19,220 frames, 44 audio tracks, 5 images, no video tracks
- Bench containers were limited to 4 vCPU for worker comparisons. The host had another long render in progress, so absolute times are noisy; the large relative differences below were repeatable across sequential runs.

## Problem 1: fresh containers can fail creating the font cache

The entrypoint starts as root and drops privileges with:

```sh
setpriv --reuid=render --regid=render --init-groups ...
```

However, the inherited environment still contains `HOME=/root`. Producer then tries to initialize its font cache under `/root/.cache`, after the process has become the unprivileged `render` user.

Observed error:

```text
EACCES: permission denied, mkdir '/root/.cache/hyperframes/fonts/inter'
```

Adding `HOME=/app` to the container environment makes the same render proceed.

### Expected fix

- Set a writable and stable home/cache directory in the image or entrypoint, for example `HOME=/app` and/or `XDG_CACHE_HOME=/app/.cache`.
- Ensure the directory is created and owned by `render` during image build.
- Add a fresh-image smoke test that submits a minimal render without any runtime environment workaround.

## Problem 2: the 4 GB Compose limit silently defeats hardware/beginFrame capture

Compose documents and sets:

```yaml
PRODUCER_BROWSER_GPU_MODE=hardware
mem_limit: 4g
```

Producer treats cgroup memory at or below 8 GB as low-memory mode. With the current 4 GB limit it logs:

```text
[SystemMemory] cgroup memory limit detected: 4096 MiB
Low-memory render profile active — screenshot capture, auto-worker calibration skipped, pinned to 1 worker
```

The effective strategy is therefore:

```text
workerCount=1
forceScreenshot=true
captureMode=screenshot
captureOperation=captureScreenshot
```

This overrides the documented `hardware`/`HeadlessExperimental.beginFrame` optimization.

### Expected fix

Choose and document one internally consistent default/profile:

- Performance profile: enough memory to stay outside producer low-memory mode, `PRODUCER_LOW_MEMORY_MODE=false`, hardware/beginFrame capture, and concurrency sized to the memory budget; or
- Constrained-memory profile: explicitly document that it uses screenshot capture and one worker, and reduce service concurrency accordingly.

The service should log its resolved capture mode, worker count, memory profile, and relevant overrides at startup or job start so operators do not have to infer the effective configuration.

## Problem 3: pooled multi-worker capture is slower than one worker

A 30-second slice of the reference project produced these service-side timings:

| Configuration | Total | Capture | Avg CPU | Peak memory |
| --- | ---: | ---: | ---: | ---: |
| 1 worker, beginFrame | 88.2 s | 60.6 s | 85% | 0.90 GiB |
| 4 workers, beginFrame, default browser pool | 134.4 s | 104.6 s | 81% | 0.83 GiB |
| 4 workers, beginFrame, browser pool disabled | 58.6 s | 29.5 s | 189% | 1.61 GiB |

With the default browser pool, four capture sessions share Chromium resources and do not consume additional CPU effectively. Disabling `PRODUCER_ENABLE_BROWSER_POOL` launched four independent Chromium instances and made the same four-worker capture approximately 3.5x faster than pooled four-worker capture.

There is also a pipeline-path effect for short renders: one worker can use streaming encode, while multiple workers currently use disk capture followed by a separate encode pass. This makes pooled multi-worker rendering even less competitive.

### Expected fix

- Do not pool Chromium sessions for parallel frame-range workers when that causes them to share the compositor/GPU-process bottleneck.
- Either default to independent browsers for `workerCount > 1`, or make the service select this mode explicitly for its latency-optimized profile.
- Keep the memory trade-off visible and bounded. The independent four-browser run peaked around 1.6 GiB on this 720p fixture.
- Add a performance regression harness that compares 1 worker with 2/4 workers on a deterministic 30-second fixture.

## Problem 4: long jobs time out and fall back from 4 workers to 2

On long renders, parallel workers make initial progress but eventually hit the default 300-second Puppeteer/CDP protocol timeout:

```text
Parallel capture timed out; retrying missing frames.
fromWorkers: 4
toWorkers: 2
Runtime.callFunctionOn timed out
```

For the 640.65-second reference project, each of four workers receives roughly 4,805 frames. Even after disabling the browser pool, a range can be close to or above 300 seconds, so the timeout remains a likely failure boundary.

On the full fixture, the first 60 seconds of capture showed only a modest pooled-worker gain:

- 1 worker: 921 frames
- 4 pooled workers: 1,043 frames

The later timeout and adaptive retry can erase that initial gain.

### Expected fix

- Split long renders into bounded frame chunks whose worst-case capture time stays below the protocol timeout, rather than assigning one very long range to each worker.
- Prefer the existing distributed/chunked producer path for long single-job latency where practical.
- If a larger protocol timeout is retained as a fallback, derive it from calibrated per-frame cost and assigned range size instead of using a fixed value.
- Preserve already completed frames without forcing large ranges to restart.

## Problem 5: static-dedup verification adds ~16 seconds and then disables itself

The fixture is mostly static slide imagery, but each capture session spends about 16 seconds verifying predicted static frames before logging:

```text
static-frame dedup: disabled (verification budget exhausted ...)
```

The verifier has a 15-second wall-clock budget. The render pays the verification cost but still captures every frame.

On the 30-second fixture, using independent browsers and disabling the dedup attempt changed the result from:

| Configuration | Total | Capture |
| --- | ---: | ---: |
| 4 independent browsers, default dedup verification | 58.6 s | 29.5 s |
| Same, `HF_STATIC_DEDUP=false` | 27.3 s | 14.4 s |

Disabling dedup is only a workaround; this project shape should eventually benefit from safe static-frame reuse rather than permanently opting out.

### Expected fix

- Make verification sampling scale with the available budget and predicted-static runs, without attempting a verification plan that cannot finish.
- Cache/reuse a verified static-frame plan across capture workers for the same composition hash.
- For known OpenMAIC-generated compositions, investigate a compiler-provided static/dynamic frame map that does not require every worker to repeat browser-side verification.
- Until that is available, allow the render service to skip known-doomed verification without changing output correctness.

## Additional configuration trap

With low-memory mode disabled, setting only `PRODUCER_MAX_WORKERS=1` can still resolve to two workers for jobs above `PRODUCER_MIN_PARALLEL_FRAMES`. The worker resolver's minimum-parallel branch overrides the effective max-worker value. A real one-worker control currently also needs a high `PRODUCER_MIN_PARALLEL_FRAMES` or a programmatic `workers: 1` option.

The max-worker setting should be a real upper bound.

## Proposed delivery shape

It may be useful to expose two explicit service presets:

### Latency-optimized single-job profile

- Independent Chromium instances for producer workers
- 2–4 workers depending on CPU and memory
- `RENDER_MAX_CONCURRENCY=1`
- Chunked long renders and calibrated protocol deadlines

### Throughput-optimized multi-job profile

- One producer worker per job
- Hardware/beginFrame capture
- `RENDER_MAX_CONCURRENCY` sized to CPU/RAM

In a two-job test on 4 vCPU, two one-worker renders completed in 114.7 seconds versus approximately 185.6 seconds sequentially: about 1.62x aggregate throughput, with slightly higher per-job latency.

## Acceptance criteria

- [ ] A fresh Docker image renders a minimal project without setting `HOME` manually and without any font-cache permission error.
- [ ] The default Compose configuration's documented capture mode matches the resolved runtime mode.
- [ ] `PRODUCER_MAX_WORKERS=1` never resolves above one worker.
- [ ] On the deterministic 30-second fixture, four configured workers use independent capture resources and are measurably faster than one worker, or the service deliberately selects one worker when parallelism is predicted to lose.
- [ ] A 10+ minute render does not fall back because a single worker range exceeds a fixed CDP timeout.
- [ ] Static-dedup initialization does not repeatedly spend the full verification budget and then disable itself for every worker.
- [ ] Render logs expose resolved memory profile, capture API, browser-pool mode, worker count, streaming/disk strategy, and retry history.
- [ ] Resource limits remain explicit and tested so the performance fixes do not reintroduce OOM or isolation regressions.

cosarah:
## Summary

Fix the render-service container startup failure and make its documented multi-worker capture path actually use independent Chromium resources.

The default Docker Compose configuration is now a bounded single-job adaptive safety profile: producer observes the 4 GiB cgroup, keeps workers unset, and selects its low-memory one-worker screenshot path. The measured four-worker beginFrame setup remains documented as an explicit 1280x720-only opt-in profile.

## Related Issues

Closes #1041

## Changes

- Reset `HOME`/`XDG_CACHE_HOME` to `/app` before dropping privileges, create the cache directory in the image, and keep it owned by the `render` user.
- Keep `PRODUCER_MAX_WORKERS` unset by default and only pass `job.config.workers` when the operator explicitly configures an override, preserving producer CPU/memory/low-memory/small-job/capture-cost guards.
- Keep service defaults at one active render and one active extraction; Compose leaves workers and low-memory mode unset so producer retains its cgroup-aware guards.
- Install Debian `chromium-headless-shell`, configure both `PUPPETEER_EXECUTABLE_PATH` and producer-consumed `PRODUCER_HEADLESS_SHELL_PATH`, and fail jobs whose worker-reported capture mode is not exactly `beginframe` when `RENDER_REQUIRE_BEGINFRAME=true`.
- Configure Compose for the adaptive safety path:
  - beginFrame-capable headless shell remains available when memory permits;
  - no explicit worker or low-memory override;
  - browser pool disabled for operators who later opt into measured parallel capture;
  - 900-second Puppeteer/CDP protocol timeout;
  - skip the static-dedup verification that currently exhausts its budget and disables itself on OpenMAIC slide exports;
  - one admitted render/extraction, a 2 GiB `/dev/shm` ceiling, and the existing 4 GiB container memory limit.
- Log the resolved render concurrency, worker count or auto-sizing, capture mode inputs, headless-shell path, browser-pool mode, low-memory mode, and dedup setting at service startup.
- Document latency-oriented and throughput-oriented deployment profiles.
- Add regression coverage for adaptive worker defaults, explicit worker overrides, the Compose container contract, and required beginFrame mode.

## Type of Change

- [x] Documentation update
- [x] CI/CD or build changes

### Steps to reproduce / test

1. Run `cd render-service && npm test && npm run typecheck`.
2. Run Prettier and ESLint against the changed files.
3. Run `docker compose --profile video-export config` with a local environment file.
4. Build/start the render-service image and submit an exported OpenMAIC project ZIP.
5. Confirm the cache path, requested worker count, worker-reported capture mode, and number of Chromium launches in the service logs.

- 36 render-service tests pass.
- TypeScript type checking passes.
- Prettier and `git diff --check` pass.
- ESLint reports no errors for the changed TypeScript files; only the existing missing-pages warning is emitted.
- The Compose `video-export` profile parses successfully.
- A 4 GiB cgroup smoke test with the default Compose-equivalent environment reported `requestedWorkers: auto`, `lowMemoryMode: true`, then explicitly pinned capture to `workerCount: 1` with screenshot streaming; no four-browser multiplier was created.
- A fresh container runs as the unprivileged `render` user with writable `HOME=/app` and `XDG_CACHE_HOME=/app/.cache`.
- Debian `chromium-headless-shell` 151.0.7922.71 was installed in the smoke container using the same packages selected by the Dockerfile.
- A complete render using the 640.65-second reference `video.zip`:
  - cached Inter fonts under `/app/.cache/hyperframes/fonts/inter` without EACCES;
  - reported `requestedWorkers: 4`, `workerCount: 4`, `forceScreenshot: false`, and `protocolTimeout: 900000`;
  - launched four independent `HeadlessChrome/151.0.7922.71` instances with `beginframe` and `headlessShell=true`;
  - emitted no beginFrame fallback warning;
  - captured all 19,220 frames and completed successfully in approximately 201 seconds, including audio processing and encoding;
  - passed the service assertion against producer worker performance data (`drawElement.mode === beginframe`).
- Prior controlled 30-second comparison on the same export path:
  - four pooled beginFrame workers: 104.6-second capture;
  - four independent beginFrame workers: 29.5-second capture;
  - independent workers with the known-doomed dedup verification disabled: 14.4-second capture.

### Evidence

- [x] Manually tested locally
- [ ] Screenshots / recordings attached (not applicable: service/container change)

## Checklist


render-service/src/config.ts
```diff
@@ -28,13 +28,15 @@ const MB = 1024 * 1024;
...
+  maxConcurrency: intEnv('RENDER_MAX_CONCURRENCY', 1),
   /**
    * Archives extracted simultaneously. Extraction holds the expanded archive in
    * memory, so this bounds the RAM multiplier (≈ this × maxExpandedBytes) even
    * when many jobs are admitted at once. Defaults to the render concurrency.
    */
-  maxConcurrentExtractions: intEnv('RENDER_MAX_CONCURRENT_EXTRACTIONS', 2),
+  maxConcurrentExtractions: intEnv('RENDER_MAX_CONCURRENT_EXTRACTIONS', 1),
+  /** Explicit per-job producer workers; passing this avoids producer auto-minimum surprises. */
+  producerWorkers: intEnv('PRODUCER_MAX_WORKERS', 4),
```
wyuc: [P1] This default is always forwarded as an explicit `job.config.workers`. In producer 0.7.60, a defined worker count short-circuits the low-memory, CPU, memory, small-job, and capture-cost guards. Non-Compose low-memory deployments, as well as supported 1080p and 4K renders under the 4 GiB Compose limit, can therefore start four Chromium instances without producer's resource protection. Keep workers unset unless the operator explicitly configures them, or restore equivalent resource-aware bounds.

docker-compose.yml
```diff
@@ -93,8 +93,28 @@ services:
       # API, not hardware rasterization — beginFrame falls back gracefully if the
       # host lacks a GPU.
       - PRODUCER_BROWSER_GPU_MODE=hardware
```
wyuc, line 90: [P1] `PRODUCER_BROWSER_GPU_MODE=hardware` does not make beginFrame reachable in the current image. Producer 0.7.60 selects beginFrame only when `resolveHeadlessShellPath()` finds `PRODUCER_HEADLESS_SHELL_PATH` or a programmatic `chromePath`, while this image only sets `PUPPETEER_EXECUTABLE_PATH`. The effective capture mode remains screenshot, so the profile disables the low-memory guard without receiving the stated beginFrame benefit. Wire the path producer consumes and assert the resolved capture mode.

wyuc (changes requested):
Two blocking issues remain in the default performance profile. Please preserve producer's memory-aware worker selection when no explicit override is supplied, and make sure the container configures the browser path that producer actually uses for beginFrame. Details inline.

cosarah:
Addressed both blocking review comments in commit 597ae8b5. PRODUCER_MAX_WORKERS is now optional: the service omits job.config.workers when no operator override is set, preserving producer 0.7.60 CPU, memory, low-memory, small-job, and capture-cost guards; Compose still opts into four workers explicitly. The image now installs Debian chromium-headless-shell and configures both PUPPETEER_EXECUTABLE_PATH and producer-consumed PRODUCER_HEADLESS_SHELL_PATH. PRODUCER_BROWSER_GPU_MODE=hardware is enabled, and RENDER_REQUIRE_BEGINFRAME=true rejects jobs whose worker-reported drawElement.mode is not exactly beginframe. Verification: 36 render-service tests passed, typecheck passed, ESLint passed with only the existing missing-pages warning, Compose video-export config parsed successfully, and the full ~/video.zip smoke test completed successfully with 19,220 frames and four workers. Logs showed HeadlessChrome/151.0.7922.71, beginframe, headlessShell=true for all four browsers with no fallback warning.

docker-compose.yml
```diff
@@ -93,8 +93,30 @@ services:
...
+      # and one extraction active so the 4 GiB boundary remains meaningful.
+      - PRODUCER_LOW_MEMORY_MODE=false
+      - RENDER_MAX_CONCURRENCY=1
+      - RENDER_MAX_CONCURRENT_EXTRACTIONS=1
+      # Producer's browser pool shares a Chromium compositor across capture
+      # sessions. Independent browsers made four-worker capture ~3.5x faster on
+      # the reference fixture, at the cost of additional bounded memory.
+      - PRODUCER_ENABLE_BROWSER_POOL=false
+      # RenderManager also passes this value explicitly as job.config.workers, so
+      # producer's minimum-parallel threshold cannot silently change 1 -> 2.
```
wyuc: [P1] This explicit override still bypasses producer's memory-aware worker sizing in the default Compose profile. The completed smoke test covers the 1280x720 reference fixture, but the product defaults to 1920x1080 and also offers 4K. Four independent Chromium processes, Node, extraction, FFmpeg, frame buffers, and `/dev/shm` all share the same 4 GiB cgroup, so a supported higher-resolution export can OOM-kill the whole service and discard its in-memory queue. Use adaptive or resolution-aware workers, or provide peak-memory validation for the supported resolutions and size the limit from that evidence.

wyuc (changes requested):
The beginFrame path and non-Compose adaptive worker behavior are fixed. One blocking issue remains: the default Compose profile still forces four independent browsers under a 4 GiB limit, while the supported default 1080p and selectable 4K paths have no validated peak-memory bound. Please restore adaptive sizing in Compose, make the worker count resolution-aware, or validate the supported resolutions and adjust the memory limit accordingly.

docker-compose.yml
cosarah: Fixed in 8d028912. The default Compose profile no longer sets PRODUCER_MAX_WORKERS, disables producer low-memory mode, or requires beginFrame. Producer now observes the 4 GiB cgroup and retains its memory/CPU/small-job/capture-cost guards. A Compose-equivalent 4 GiB smoke test reported requestedWorkers=auto and lowMemoryMode=true, then pinned workerCount=1 with screenshot streaming. The four-worker beginFrame configuration is now documented only as an explicit 1280x720 opt-in profile, with a warning not to apply it to 1080p/4K until peak cgroup memory is measured.

wyuc (approved):

PR #1048, opened 2026-08-05
feat(classroom): route generate-classroom per stage instead of one model

CXuPercy:
## Problem

The one-shot `generateClassroom` path (skill API / `/api/generate-classroom`) resolved the model once with `resolveModel({ stage: 'generate-classroom' })` and bound that single `languageModel` into two closures: `aiCall` (shared by outline generation + agent profiles) and `sceneAiCall` (shared by every scene-content and scene-actions call).

As a result, none of the `MODEL_ROUTES` entries for these stages had any effect on this path:

```jsonc
MODEL_ROUTES='{"scene-content":"openai:gpt-5.4"}'        // ignored
MODEL_ROUTES='{"scene-content:quiz":"deepseek:..."}'     // ignored
MODEL_ROUTES='{"scene-actions":"openai:gpt-5.4"}'        // ignored
MODEL_ROUTES='{"agent-profiles":"openai:gpt-5.4-mini"}'  // ignored
```

The browser UI already routes each stage independently via the `/api/generate/*` endpoints (each calls `resolveModelFromRequest` with its own stage), but the one-shot API did not — so operators could not mix models (e.g. strong model for outlines, cheap model for scene content / agent profiles) on this path.

A separate latent bug on this same path: PBL scene content generation needs the resolved `LanguageModel` object (because `generatePBLSceneContent` drives its own LLM calls through it, not the `aiCall` closure), but the call site never passed `languageModel` in options. So PBL scenes silently returned `null` and were skipped. The browser UI already passes it (`/api/generate/scene-content`); the one-shot path did not.

## Fix

Resolve the model per stage instead, mirroring the browser UI's `/api/generate/*` routing exactly:

| Stage on this path | Routes via | Browser UI counterpart |
|---|---|---|
| scene-content | `scene-content:<type>` (slide/quiz/interactive/pbl) → `scene-content` | `/api/generate/scene-content` |
| scene-actions | `scene-actions` | `/api/generate/scene-actions` |
| agent-profiles | `agent-profiles` | `/api/generate/agent-profiles` |
| outline | `generate-classroom` (unchanged) | `/api/generate/scene-outlines-stream`* |

\* The one-shot path's outline call keeps using the classroom model (the `generate-classroom` closure), matching prior behavior; the browser UI uses a separate `scene-outlines-stream` endpoint. Outline routing on this path is intentionally out of scope.

Interactive widget sub-types are not split — only the four core scene types are routable via composite keys.

Each stage is resolved lazily and only when a route is actually configured (`getStageModel` returns `undefined`), so unrouted deployments reuse the classroom model at zero extra cost. A route that fails to resolve (e.g. an unknown provider) degrades to the classroom model with a warning, never aborting generation — the same handling already used for `web-search-query-rewrite`.

The PBL fix passes the resolved `LanguageModel` **and** the route's `thinkingConfig` into `generateSceneContent`'s options when `outline.type === 'pbl'`, so PBL scenes are no longer silently skipped and a `scene-content:pbl` route with a `thinking` block is fully honored (not just the model).

The `generate-classroom-scene` telemetry/billing label is preserved on all scene calls (scene-content + scene-actions); agent-profiles keeps the `generate-classroom` label. No existing logging or accounting changes.

## Verification

- `tsc --noEmit` — clean
- `eslint lib/server/classroom-generation.ts` — clean
- `vitest run tests/server/model-routes.test.ts` — 19/19
- `vitest run tests/server/classroom-generation-retry.test.ts` — 4/4 (exercises the full pipeline incl. scene-content/actions retry)
- `vitest run tests/server/classroom-agent-mode.test.ts` — 3/3
- `vitest run tests/server/resolve-model.test.ts` — 19/19

## Behavior

| `MODEL_ROUTES` config | Result on this path |
|---|---|
| (none) | All stages reuse the classroom model — identical to before (PBL scenes now generate instead of being skipped) |
| `{"scene-content":"X"}` | All scene content uses X |
| `{"scene-content:quiz":"A","scene-content:slide":"B"}` | quiz→A, slide→B, other types fall back to `scene-content` then the classroom model |
| `{"scene-actions":"Y"}` | Scene actions use Y |
| `{"agent-profiles":"Z"}` | Agent profile generation uses Z (when `agentMode: 'generate'`) |
| `{"scene-content:pbl":{"model":"P","thinking":{"enabled":true}}}` | PBL scenes use P with thinking enabled |
| `{"scene-content":"nope:x"}` (bad provider) | warn + fall back to classroom model; generation completes |

## Scope

Single file changed: `lib/server/classroom-generation.ts`. No changes to `scene-generator.ts`, `model-routes.ts`, `resolve-model.ts`, or any `/api/generate/*` route.

## Review

Cross-reviewed (two independent reviewers on the same diff). Both converged on the same finding (PBL route's `thinking` config being dropped alongside the model), which is fixed in this revision. Agent-profiles routing was added to fully align with the browser UI per review feedback.

### DIFF
--- a/lib/server/classroom-generation.ts
+++ b/lib/server/classroom-generation.ts
@@ -18,7 +18,9 @@ import { createLogger } from '@/lib/logger';
 import { isProviderKeyRequired } from '@/lib/ai/providers';
 import { resolveClassroomWebSearchConfig } from '@/lib/server/web-search-config';
 import { resolveModel } from '@/lib/server/resolve-model';
-import { getStageModel } from '@/lib/server/model-routes';
+import { getStageModel, type LlmStage } from '@/lib/server/model-routes';
+import type { LanguageModel } from 'ai';
+import type { ThinkingConfig } from '@/lib/types/provider';
 import { resolveVocationalActive } from '@/lib/config/feature-flags';
 import { buildSearchQuery } from '@/lib/server/search-query-builder';
 import { formatSearchResultsAsContext, searchWeb } from '@/lib/web-search';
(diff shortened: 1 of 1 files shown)


wyuc (approved):
Validated the one-shot generation path end to end: per-stage routing for agent-profiles, scene-content:pbl, and scene-actions; PBL classroom persistence; and invalid scene-actions route fallback. CI is green.

PR #1050, opened 2026-08-05
fix(chat): harden runtime sync, legacy migration, and record validation

wyuc:
## Summary

Hardens the chat runtime-sync engine and the `@openmaic/storage` HTTP client against malformed legacy data and retry storms. Three commits:

1. **`fix(chat): harden runtime sync and legacy migration`**
   - Extracts the pure sync core into `lib/utils/chat-storage-core.ts` (`planChatSync`, `buildChatRecordInit`, folding/diffing helpers) with type-only imports, so the engine is importable outside the browser (e.g. Node-side tooling).
   - Legacy row normalization at the single entry point: `sceneId` kept only when it is a string, `lastActionIndex` only when it is an integer (nullable/dirty rows previously leaked `null` into record envelopes and were rejected server-side), and timestamps outside the representable ISO range are repaired deterministically from the row's other timestamp — idempotent across repeated migrations.
   - Retry classification in `syncOne`: deterministic failures (400/401/403/413, non-race `VALIDATION_FAILED`, `FUTURE_VERSION`) fail fast instead of burning 12 immediate retries; genuinely recoverable interleavings (inactive-session append race, vanished-session 404, append conflicts, network errors) keep retrying with capped exponential backoff (≤500ms per sleep, ≤2.7s per session).
   - Data-safety guards: malformed legacy rows are skipped and reported, and the legacy source is never cleared while any row was skipped (Dexie `clear` deletes rows permanently); backup restore fails loud and rolls back instead of silently dropping rows; historical `soft-closing` state records are readable again and the status is normalized on migration, so those sessions are no longer invisible or vulnerable to state-less-generation cleanup.
2. **`fix(storage): validate HTTP runtime appends client-side`** — `HttpRuntimeStore.appendRecord` now runs the same `validateRuntimeRecord` pre-flight as the browser and PG stores, so invalid envelopes fail loud client-side instead of surfacing as server 400s.
3. **`fix(storage): harden runtime validation and legacy guards`** — the pre-flight throws `HttpRuntimeStoreError(400, VALIDATION_FAILED)` (indistinguishable from the server verdict, so callers classify it as deterministic); unobserved loads no longer wipe the malformed-row guard; adds a shared record-envelope contract test across all three backends; bumps `@openmaic/storage` to 0.2.2.

## Testing

- `tests/runtime/` chat suites and the full `@openmaic/storage` suite green, including the two PostgreSQL contract suites against a real PostgreSQL 17 container (719/719).
- `pnpm check`, `pnpm lint`, `tsc --noEmit`, and `scripts/check-package-version-bumps.mjs` all pass.


xuyuanwei678:
Found one blocking issue:

`isLegacyRecord` only verifies that `messages` is an array; it does not validate the message entries themselves. A malformed legacy row such as `{ messages: [null] }` passes this check, then later crashes when `statePayload()` reads `message.id`. As a result, the entire migration fails instead of reporting and skipping the malformed row, and valid rows in the same legacy payload cannot be migrated.

Please validate message items at the `fromLegacyRecord` boundary, or ensure payload-construction failures are caught by the per-row skip path in `fromLegacyRecords`.

wyuc:
Fixed in 380bef80 — thanks for catching this. Two layers:

1. **Boundary validation**: legacy message entries are now validated per-element (non-null object, string `id`, and a safe `parts` shape — derived from exactly what payload construction reads), so `{ messages: [null] }` is rejected at the `fromLegacyRecord` boundary.
2. **Backstop**: `fromLegacyRecords` wraps per-row normalization in a try/catch, so any unexpected construction failure routes that row into the per-row skip path (reported once, legacy source retained) instead of aborting the whole batch.

Regression tests cover your exact scenario — a `{ messages: [null] }` row alongside a valid row: the malformed row is skipped and reported, the valid row migrates, and the legacy clear guard stays engaged — plus a missing-string-id entry.

xuyuanwei678 (approved):

Issue #1058, opened 2026-08-05
Retire PBL v1: converge generation, rendering, and storage on v2


wyuc:
## Where v1 stands today

PBL has had two generations coexisting since v2 landed. An audit of `main` (`d895b895`) shows v1 is now three very different things:

**1. A generation fallback that is rare by construction.** The pbl branch of `generateSceneContent` routes to v2 — single-call first, then the loop planner — and only falls back to v1's `generatePBLContent` when *both* v2 attempts fail for an ordinary project (`lib/generation/scene-generator.ts:989-1011`). Scenario projects already refuse to degrade (`:982-987`). The only switch, `PBL_V2_DISABLED` (`:915`), is set in no environment file in the repo.

**2. Dead rendering code.** The renderer resolves every scene to v2 — `content.projectV2 ?? upgradeLegacyPBLConfigToProjectV2(projectConfig)` (`components/scene-renderers/pbl-renderer.tsx:50-54`) — and enters `PBLV2Container` (`:113`). The v1 branches below it are unreachable: `isEmptyLegacyPBLConfig` (`lib/pbl/v2/compat.ts:150`) guarantees any non-empty legacy config upgrades and returns early, and empty configs are caught by the placeholder branches (`:132`, `:141`) before reaching `PBLRoleSelection` (`:150`) or `PBLWorkspace` (`:161`). That makes the six v1 components under `components/scene-renderers/pbl/` (~830 lines), `usePBLChat`, and the `/api/pbl/chat` route — whose only callers are those components — dead code.

**3. A storage write requirement.** `validateAppScene` requires `projectConfig` on every pbl scene (`lib/document-store/validators.ts:49-53`). This is the reason the v2 generator and the v2 runtime still write a down-converted v1 mirror next to every `projectV2` (`lib/generation/scene-generator.ts:968-971`, `lib/pbl/v2/runtime/document-persistence.ts:30` via `projectV2ToLegacyProjectConfig`). Every v2 payload carries a legacy shadow forever.

Meanwhile, scenes generated before v2 became the default are stored in v1-native shape and already play correctly through the read-time upgrade.

## Proposal: retire everything except the read path — one PR

Stored v1 data is **not** migrated. The read-time upgrade already renders it; leaving the original data in place keeps information the upgraded projection drops (per-issue agents, chat history), and any future improvement to the upgrade function benefits old scenes for free. What retires is v1's *write side* and its dead UI.

[...]

PR #1060, opened 2026-08-05
refactor(pbl): retire v1 write path and dead UI, converge on v2

wyuc:
Implements #1058.

## What this does

Retires PBL v1 everywhere except the read path, in the five coupled pieces the issue lays out:

- **Dead v1 UI deleted**: the six `components/scene-renderers/pbl/*` components, the unreachable v1 branches of `pbl-renderer.tsx`, `usePBLChat`, and `/api/pbl/chat`.
- **`projectConfig` is now optional**: `validateAppScene` accepts pbl content with `projectV2` only; legacy scenes with only `projectConfig` still validate and round-trip.
- **v1 mirror writes removed**: v2 generation and runtime persistence write `projectV2` only; `projectV2ToLegacyProjectConfig` is deleted. Shape assumptions keyed on `projectConfig` (scene-builder discriminant, scene-generator discriminant, store merge logic, `createDefaultPBLContent`) now key on the pbl `type` / `projectV2`.
- **v1 generation fallback deleted**: `lib/pbl/generate-pbl.ts`, `lib/pbl/mcp/*`, `pbl-system-prompt.ts`, the `pbl-design` template, and `PBL_V2_DISABLED` are gone. An ordinary PBL project whose v2 single-call and loop-planner attempts both fail now surfaces an error, matching the existing scenario-project policy.
- **Legacy read path contained**: `upgradeLegacyPBLConfigToProjectV2` plus the minimal v1 types now live in `lib/pbl/legacy/read.ts`, and the video-export legacy cover in `lib/video-export/legacy/read.ts`. Both carry doc headers stating the contract: read-only support for pre-v2 stored scenes, kept indefinitely, no writers.

Net: 46 files, +487 / −2,890.

### Why two legacy modules instead of one

The ESLint architecture rules forbid video-export code from importing host-app PBL modules, so the video-export cover keeps its own read-only adapter instead of importing `lib/pbl/legacy/read.ts`.

## Behavior changes (the only two)

1. A planner double-failure now surfaces a typed `PBLGenerationError` (carrying the upstream provider status) instead of a silently degraded v1 project; ordinary and scenario projects share this policy, and classroom generation contains the failure to the affected scene (retry, then skip).
2. Newly written pbl scenes no longer carry a `projectConfig` mirror. Stored legacy scenes keep theirs and still render through the v2 container via the read-time upgrade.

## Tests

- New synthetic v1-native fixture (`tests/fixtures/pbl-v1-scene.ts`) exercised end to end: validation round-trip, upgrade through the legacy-read module, rendering through the v2 container, and document persistence.
- `tests/pbl/legacy/read.test.ts` covers the legacy-read contract; compat tests were moved/rewritten accordingly.
- Generation tests assert the new fail-loud behavior for ordinary projects and unchanged scenario behavior.
- Full suite green locally (tsc, eslint, prettier, vitest).


lib/generation/scene-generator.ts
```diff
@@ -1615,7 +1577,7 @@ export async function generateSceneActions(
     return generateDefaultInteractiveActions(outline);
   }
 
-  if (outline.type === 'pbl' && 'projectConfig' in content) {
+  if (outline.type === 'pbl' && 'projectV2' in content) {
```
cosarah: **[P2] Update the live PBL action prompt for the v2 workflow**

This switches the successful action-generation path to `projectV2`, but `lib/prompts/templates/pbl-actions/system.md` still tells the model to explain the available roles and encourage students to select one. Role selection was removed and v2 exposes a single Instructor, so successful AI generation can still narrate an impossible workflow even though the fallback text was updated. Please update the primary prompt and its assertions to use the v2 workflow and terminology.

lib/document-store/validators.ts
```diff
@@ -46,11 +46,18 @@ export const validateAppScene: SceneValidator = (scene) => {
...
+    content.projectConfig !== undefined &&
+    (!objectValue(content.projectConfig) || Array.isArray(content.projectConfig))
+  ) {
+    errors.push({ path: '/content/projectConfig', message: '`projectConfig` must be an object' });
+  } else if (
+    value.type === 'pbl' &&
+    content.projectV2 !== undefined &&
+    (!objectValue(content.projectV2) || Array.isArray(content.projectV2))
+  ) {
+    errors.push({ path: '/content/projectV2', message: '`projectV2` must be an object' });
```
cosarah: **[P2] Validate the complete v2 payload before persisting it**

This accepts any non-array object as projectV2. I reproduced the end-to-end path with `{ projectV2: { title: "V2 project" } }`: `validateAppScene` returns valid, `BrowserDocumentStore` saves and reloads it unchanged, and `PBLRenderer` then throws `TypeError: Cannot read properties of undefined (reading "filter")` in `normalizeScenario` because `milestones` is missing (`roles.find` would fail next). Please reject structurally incomplete v2 payloads at the write boundary and preferably keep a defensive guard in the renderer as well. Note that the existing `isPBLProjectV2` check is shallow, so malformed nested array entries can still reach unsafe consumers unless the persistence validation is stronger.

wyuc:
Cross-review (Claude + Codex independently, three rounds) converged; summary of what it changed:

**Round 1** found the significant issues:
- P1: the new fail-loud throw for ordinary-project planner double failure propagated past the retry wrapper and failed the entire classroom generation job, discarding every already-generated scene. Fixed by typing the error (`PBLGenerationError`) and containing it at the classroom layer, which routes it into the same retry-then-skip machinery scenario projects already use, so the blast radius is one scene on both paths.
- `regenerate_scene_actions` no longer worked on stored legacy scenes after the discriminant moved to `projectV2`; the tool now upgrades legacy content through the legacy-read module before entering the generation layer.
- The eslint import boundary the video-export legacy module relies on did not actually cover its directory (`lib/video-export/*` does not cross `/`); the boundary glob now includes `lib/video-export/legacy/**`.
- Empty new scenes (`{ type: 'pbl' }`) rendered the "old format" message instead of the empty-project placeholder; assertions dropped with the old compat tests (language detection, `current_issue_id` fallback) were restored.

**Round 2** fixed a regression round 1 introduced: the hardened legacy-config guard was stricter than the tolerant upgrade code, so sloppy-but-renderable stored v1 scenes (e.g. an issue missing `notes`) would have degraded to the empty placeholder. The guard now checks only the structural containers the upgrade path dereferences, and unparseable chat timestamps fall back instead of throwing.

**Round 3** (fresh-context final audit) closed the tail: legacy chat records without a string `message` are dropped during upgrade instead of crashing the v2 chat UI, the validator rejects array-valued `projectConfig`/`projectV2`, `PBLGenerationError` is exported from the generation barrel, and a loop-level test pins that one failing PBL scene skips while the rest of the classroom persists.

**Accepted behavior (intentional, documented and pinned by tests):** a legacy scene that gains `projectV2` at runtime keeps its original `projectConfig` verbatim through persistence and export. The old mirror write replaced it with a lossy down-conversion on every save; preserving the original v1 record (chat history, per-issue agents) is the point of this issue's read-path design.

Final state: both reviewers report zero actionable findings; CI green.

lib/generation/scene-generator.ts
wyuc: Good catch. The primary `pbl-actions` template still narrated role selection; updated it (and the wording assertions) to the v2 workflow: the intro now previews the project's driving goal and key milestones and invites students into the workspace instead of explaining selectable roles. The packaged copy in `@openmaic/generation` (which landed on main mid-review and pins byte-identity) is synced as well.

lib/document-store/validators.ts
wyuc: Fixed at both boundaries. `validateAppScene` now requires the structural containers the render path unconditionally dereferences (milestones/roles/threads/submissions/evaluations/engagementEvents arrays with object items, plus per-milestone `microtasks` and per-thread `messages`), via a `hasPBLProjectV2Containers` predicate exported next to `isPBLProjectV2`; `PBLRenderer` applies the same predicate defensively so already-stored malformed payloads render the empty-project placeholder instead of crashing. Your exact repro is pinned as a validator rejection test plus a renderer no-throw test. Leaf scalars stay unvalidated on purpose, mirroring the tolerant legacy-side guard: containers are what the normalize path hard-requires, and the renderer guard is the last line for anything deeper. An explicit `projectV2: null` is treated like absent (valid, legacy fallback) since documents stored before this validation existed could carry it.

cosarah:
KEEP — Normalize legacy PBL content in the actions API (P2)

`POST /api/generate/scene-actions` forwards a v1 `{ type: "pbl", projectConfig }` payload directly. `generateSceneActions` only handles PBL content containing `projectV2`, so it returns an empty action list; `buildCompleteScene` also requires `projectV2` and returns `null`, causing `GENERATION_FAILED`. The agent-tool adapter covers a separate path, so historical v1 action regeneration through this API remains broken.

KEEP — Preserve provider status through PBL planner fallback (P2)

Both PBL planner attempts catch and discard their original errors, then throw a new `PBLGenerationError` without a `cause` or HTTP status. Consequently, `llmApiError` cannot recover upstream 401/403/429/5xx statuses and reports an internal 500 instead. This is reachable whenever both planner attempts fail and at least one failure is an upstream provider HTTP error.

wyuc:
Both findings fixed, thanks for the thorough passes.

**Actions API + legacy content**: the upgrade normalization is now a single shared helper (`normalizeLegacyPBLContent` in `lib/pbl/legacy/read.ts`) used by both the agent-tool adapter and `POST /api/generate/scene-actions`, so a stored v1 scene generates actions and builds a scene through the API instead of returning GENERATION_FAILED. Covered by a route-level test using the v1 fixture. I checked the sibling generate routes: scene-content only produces new content and cannot receive stored legacy payloads, so no other route needed the normalization.

**Provider status through planner failures**: `PBLGenerationError` now carries the last attempt's error as `cause` and exposes `statusCode` (last attempt's status, falling back to the first attempt's; numeric strings parsed the same way `llm-error-response.ts` does), so `llmApiError` maps upstream 401/403/429/5xx instead of a generic 500. Pinned by planner tests (including the string-status case) and a route-level 429 mapping test.

Also merged main to resolve the conflict with the planner-core split (#1069): the branch now uses the new `kernel/` module paths and injected LLM call seams while keeping the retirement semantics; full suite green on the merge.

cosarah (commented):
Triage: both discovery findings are actionable and are included as inline comments.

lib/generation/scene-generator.ts
```diff
@@ -914,114 +937,81 @@ async function generatePBLSceneContent(
...
+  const lastError = attemptErrors[attemptErrors.length - 1];
 
-    return { projectConfig };
-  } catch (error) {
-    log.error(`PBL v1 generation also failed:`, error);
+  if (scenarioRoleplay) {
+    log.error(
+      `PBL v2 scenario generation failed for "${outline.title}"; refusing to generate an ordinary PBL project.`,
+    );
     return null;
```
cosarah: KEEP: When both planner attempts fail for a `scenarioRoleplay` outline, this branch returns `null`. `/api/generate/scene-content` treats `null` as `GENERATION_FAILED` with HTTP 500, so provider status codes (401/429/503) from the caught errors never reach `llmApiError`. Ordinary PBL wraps the final error in `PBLGenerationError` with `statusCode`, so scenario failures lose authentication, rate-limit, and provider-unavailable semantics and can be retried as generic failures. Please preserve or propagate the upstream error status here.

lib/generation/scene-generator.ts
```diff
@@ -1624,7 +1614,7 @@ export async function generateSceneActions(
     return generateDefaultInteractiveActions(outline);
   }
 
-  if (outline.type === 'pbl' && 'projectConfig' in content) {
+  if (outline.type === 'pbl' && 'projectV2' in content) {
```
cosarah: KEEP: The revised `pbl-actions` prompt asks the model to preview key milestones, but this branch only interpolates outline/`pblConfig`/course context. `content.projectV2` (including its generated milestones, learning objective, and gains) is available yet never passed to `buildPrompt`, so the narrator cannot reliably describe the plan actually generated. Please add a compact project summary/milestone payload to the prompt variables and cover it with a behavioral test.

lib/generation/scene-generator.ts
wyuc: Unified with the ordinary path: a scenario double failure now throws the same `PBLGenerationError` carrying `cause` and the recovered upstream `statusCode`, so `/api/generate/scene-content` maps 401/429/503 through `llmApiError` instead of a generic 500. The classroom layer's containment converts the typed error back into the existing retry-then-skip flow, so refuse-to-degrade behavior there is unchanged. Pinned by a scenario rejection test with a status-carrying planner error.

lib/generation/scene-generator.ts
wyuc: Agreed, the narrator was flying blind. `generateSceneActions` now builds a compact plain-text project summary from the generated `projectV2` (title, description/goal, gains, ordered milestone titles) and interpolates it through a new `{{projectSummary}}` variable in a "## Project Plan" template section that instructs the narrator to ground the intro in the actual plan (and not to invent or rename milestones). Legacy/empty content falls back to the outline topic. Covered by behavioral tests asserting the milestone titles reach the prompt, plus the template-wording tests; the packaged byte-identical copy is synced.

cosarah:
Triage of the discovery findings:\n\n1. [P2] Preserve non-retryable PBL failures in batch generation — KEEP\n   When both PBL planner attempts fail with a non-retryable provider status (for example 401/403/404), PBLGenerationError preserves that status, but lib/server/classroom-generation.ts catches it in the per-attempt callback and returns null. withGenerationRetry therefore sees an empty result and retries five more times. Each attempt runs both planners, so one failed scene can cause up to 12 planner calls before being silently skipped or ending in a generic no-scenes failure.\n\n2. [P2] Keep packaged PBL prompt inputs backward-compatible — KEEP\n   The published @openmaic/generation PBL system template adds the required {{projectSummary}} input, while the public buildPrompt API leaves omitted variables as literal placeholders and exposes no PBL-specific contract or fallback. Existing consumers using the previous valid variable set can therefore send {{projectSummary}} directly to the model.\n\nThe malformed legacy scalar-field finding was reviewed and dismissed as a long-tail case limited to corrupted or hand-edited historical v1 JSON; normal v1 generation enforced string fields, and new writes use v2.

cosarah:
Triaged against the reviewed PR revision:

1. KEEP — A non-null malformed projectV2 makes PBLRenderer return null instead of falling back to projectConfig; video export also gives any record-shaped v2 payload precedence. A hybrid scene with valid legacy data and damaged v2 data therefore renders an empty or incorrect result. Trigger: rare historical or manually edited hybrid records, but this breaks an explicit compatibility path.

2. KEEP — The persisted-project guard validates container shapes but not scalar leaves, while buildPBLProjectSummary calls .trim() on gains, project fields, and milestone titles. Inputs such as gains: [42] or a numeric milestone title can make /api/generate/scene-actions fail with a 500. Normal planner output is safe, but accepted legacy or hand-edited data can reach this recovery path.

3. KEEP — The routing test only checks that generateSceneContent returns a v2 object containing milestones and roles. It does not exercise scene construction or the document validator requirements for submissions, evaluations, threads, engagementEvents, and nested containers, so a future planner regression can pass the test and fail at persistence.

KEEP set: 1, 2, 3.

wyuc:
All five findings from the last two triage rounds are fixed:

**Non-retryable batch failures**: PBL containment moved outside `withGenerationRetry`, so the typed error is classified by the retry layer's existing status rules first — 429/5xx planner failures keep the backoff retries, while 401/403/404 and status-less validation failures skip the scene after a single `generateSceneContent` call (the planner ladder itself is the retry). The behavior matrix (429 retried-then-skipped / 401 exactly one attempt / status-less one attempt / non-PBL errors unchanged) is pinned by loop-level tests.

**Packaged prompt backward compatibility**: `projectSummary` now has a registered default in both prompt loaders (app and `@openmaic/generation`), so consumers using the previous variable set get a neutral fallback line instead of a literal `{{projectSummary}}` reaching the model. Covered by a loader test on the package side.

**Hybrid scenes with damaged v2**: the renderer now falls through to the upgraded legacy config when `projectV2` fails the container check, and video export applies the same decision — the cover-side predicates mirror `hasPBLProjectV2Containers` and `isEmptyLegacyPBLConfig` clause for clause (the eslint boundary forbids importing them, so the mirrors carry doc pointers). For v2-only scenes the cover stays permissive: a sparse cover beats an empty one and there is no legacy data to diverge from.

**Summary building**: `buildPBLProjectSummary` treats non-string leaves as absent (string-filtered gains, positional `Task N` fallbacks, index-based ordering when `order` is not a number) — no container-valid payload can 500 the actions route.

**Contract pinning**: the routing tests now run the planner output through `hasPBLProjectV2Containers` and `validateAppScene`, so a planner regression that drops any persisted container fails the test instead of failing at persistence.

Also merged main twice more along the way (contract widget/interactive validator, transform pipeline foundation) and bumped `@openmaic/generation` to 0.1.2 after the version collision with main's own 0.1.1. Full suite and CI green on the merge.

cosarah:
**[P2] Apply the legacy fallback before runtime hydration**

This is an actionable damaged-hybrid load failure. `hydratePBLScenesFromRuntime` runs before React renders the scene and only checks `content.projectV2` for truthiness. A stored scene with a usable `projectConfig` and malformed v2 data such as `{ title: 'broken' }` reaches `stripToDesignTemplate` (which dereferences `threads.map` / `milestones.map`) and throws. The catch then calls `documentContainsLearnerState`, which repeats the same unsafe path, so the error propagates and classroom loading aborts before `PBLRenderer` can fall back to the valid legacy config. The shallow validator accepts any non-array object as `projectV2`, so this state is not excluded at the persistence boundary.

cosarah:
**[P2] Normalize malformed hybrid PBL content before action regeneration**

This is an actionable data-loss/inconsistency path. The `/api/generate/scene-actions` route does not structurally validate `content`; `generateSceneActions` and `buildCompleteScene` branch on the presence of the `projectV2` property, not on valid v2 containers. Given a usable legacy `projectConfig` plus truthy malformed `projectV2`, the request proceeds with the damaged payload and `buildCompleteScene` emits a scene containing only that malformed `projectV2`, silently discarding the recoverable legacy config. The `regenerate_scene_actions` tool has the same malformed-v2 pass-through (though its action-only response does not itself rewrite content).

wyuc:
Both findings fixed, and since this was the third call site of the same damaged-hybrid class, the fix is structural this time:

**One resolver.** `resolvePBLContent` (lib/pbl/legacy/read.ts) is now the single owner of the v2/legacy/empty decision: v2 only when the payload passes `hasPBLProjectV2Containers`, legacy when a structurally sound non-empty `projectConfig` exists, empty otherwise. The renderer, runtime hydration, document persistence, the actions route, the agent tool, and the generation discriminants all consume it (video export keeps its boundary-mirrored predicates). Every `projectV2` truthiness check in the repo was swept and converted or justified.

**Finding 1 (hydration)**: `hydratePBLScenesFromRuntime` resolves content first — a damaged-hybrid scene is skipped untouched so the renderer's legacy fallback gets it intact, and `documentContainsLearnerState` guards its own input instead of throwing. Pinned by hydration tests with the damaged-hybrid fixture.

**Finding 2 (actions route/tool)**: `normalizeLegacyPBLContent` is resolver-based, so a truthy-but-invalid `projectV2` with a usable legacy config now upgrades from the legacy side before `generateSceneActions` / `buildCompleteScene`; the built scene carries the legacy-derived project instead of silently dropping it. Route- and tool-level tests assert the upgraded content wins.

**One deliberate design point**: document persistence gates on the same full container predicate as the validator and resolver, rather than a looser learner-state-only check. A payload the renderer will not treat as v2 is inert bytes everywhere — persistence neither synchronizes nor rewrites it, so no runtime side effects occur for a document the write barrier would reject, and stored damaged bytes round-trip untouched. The residual corner (a damaged v2 payload that embeds old learner state is exported as-is, since it cannot be interpreted safely enough to strip) is accepted and covered by a pass-through test.

Full suite and CI green.

cosarah:
Triage of the discovery findings — KEEP set

1. [P2] Malformed v2 payloads can block persistence
`preparePBLScenesForDocumentPersistence` leaves container-invalid `projectV2` payloads unchanged, but the document-save path validates the same scene and rejects it. A damaged hybrid can therefore make a later full save fail, even when the user edits an unrelated part of the stage.

2. [P2] Empty-but-structural v2 can incorrectly override usable legacy data
The v2 acceptance check allows empty `milestones` / `roles` / `threads` containers. A hybrid scene can therefore select v2 over valid legacy content, render the v2 Hero, and enter a workspace with no active microtask.

3. [P2] Leaf fields are not validated before runtime use
The structural guard accepts invalid leaves such as `gains: [42]` or `proficiencyAssessment: {}`. These values later reach code that calls `.trim()` or indexes `transitions[0]`, so malformed projects can fail during rendering or persistence.

4. [P2] Provider failures still trigger the fallback planner
The generation loop falls back from the single-call planner to the loop planner for every exception, including authentication, rate-limit, server, and cancellation errors. When combined with the outer retry wrapper, one provider failure can cause unnecessary duplicate planner calls and amplify latency, cost, and rate-limit pressure.

5. [P3] Legacy shells with metadata but no runnable milestones are treated as valid
A legacy config containing only a title or description is considered non-empty and upgraded to a v2 project with zero milestones. The learner can then enter a workspace with no active task and no useful Instructor flow.

Recommended fixes should make the v2/legacy resolver require a runnable v2 shape, validate runtime-used leaf fields, restrict planner fallback to validation failures, and treat malformed or non-runnable legacy shells as recoverable/empty rather than launching an unusable workspace.

wyuc:
All five findings addressed:

**1. Damaged v2 no longer blocks saving stored hybrids.** `validateAppScene` tolerates a container-invalid `projectV2` as inert bytes when the scene carries structurally sound, non-empty legacy data — the entire pre-cutover stored population is hybrid (the old write barrier required `projectConfig`), so unrelated edits keep saving. V2-only content keeps the strict container check (that cohort is exclusively new planner writes), and an empty `projectConfig` stub does not disable v2 validation, so the exception cannot be used as a bypass.

**2. Empty-but-structural v2 no longer outranks usable legacy.** `resolvePBLContent`'s v2 arm now requires a runnable project (at least one milestone); the video-export precedence applies the same rule. A hybrid with rich legacy and a milestone-less v2 payload renders (and exports) from the legacy upgrade.

**3. Leaf hardening went into the readers, not the validator.** Every `.trim()`/index on project-derived leaves reachable from an accepted payload was audited (~20 sites across renderer components, agents, runtime ops, generation and export) and guarded; `transitions[0]` handling and a malformed `proficiencyAssessment` are treated as absent on both the snapshot and the apply side, so runtime state can neither capture nor resurface them. Containers-only validation stays, with defensive readers as the last line.

**4. Planner fallback is validation-only.** The loop planner runs only when the single-call attempt failed schema validation (`PlannerV2Error`); provider/HTTP-status failures skip the second planner run and surface immediately with their status, so a 401/429 costs one planner call instead of up to twelve.

**5. Non-runnable legacy shells resolve as empty.** The resolver's legacy arm requires at least one issue; title-only shells render the empty-project placeholder instead of launching a zero-milestone workspace.

Persistence now consumes the same resolver as the renderer (an inert payload is never synchronized or rewritten), so render, export, persistence and validation stay in agreement: the resolver decides what is *renderable*, the validator decides what is *storable* — a stored shell that renders as the placeholder still saves, by design.

Full suite and CI green.

cosarah:
Triage: KEEP for the following findings:

1. **Require a runnable project before selecting v2 (P2).** `PBLRenderer` selects any truthy `content.projectV2`; `normalizeProjectRuntime` cannot invent a missing instructor/task, and `runInstructorTurn` returns `NO_ACTIVE_MICROTASK`, so malformed hybrid v2 data can shadow usable legacy content and produce a non-runnable workspace.

2. **Align validation with the resolver's runnable predicate (P2).** `validateAppScene` only requires `projectV2` to be a non-array object, while the renderer still prioritizes it over legacy data; a malformed v2 object can therefore be persisted and later crash or render empty instead of falling back.

3. **Exercise the real document-persistence boundary (P2).** The helper tests mock `synchronizePBLProjectRuntime`, and the store/document-store tests mock the preparation helper, so no test currently performs an actual full and incremental save followed by reload/hydration; wiring regressions could pass unnoticed.

wyuc:
Findings 1 and 2 describe an earlier revision of this branch — both requested behaviors are already on HEAD:

1. `PBLRenderer` no longer selects a truthy `projectV2`. It resolves through `resolvePBLContent` (lib/pbl/legacy/read.ts), whose v2 arm requires the full container set AND at least one milestone — exactly the runnable predicate this finding asks for. Malformed or milestone-less hybrid v2 falls back to the upgraded legacy config, pinned by renderer, hydration, actions-route, and video-export tests.
2. `validateAppScene` has required `hasPBLProjectV2Containers` for v2-only scenes since several rounds back. The only tolerance is for hybrids carrying structurally sound, non-empty legacy data (the stored pre-cutover cohort), where damaged v2 is inert bytes the renderer never prioritizes over legacy; an empty `projectConfig` stub does not disable the check.

Finding 3 was a fair gap and is now covered: a no-mock integration suite (tests/pbl/v2/persistence-hydration-roundtrip.test.ts) drives the real chain — `preparePBLScenesForDocumentPersistence` → real `BrowserDocumentStore` save over fake-indexeddb → reload → real `hydratePBLScenesFromRuntime` over a real `BrowserRuntimeStore`/KV — for the full-save path (lib/store/stage.ts) and the incremental dirty-scene path (lib/utils/stage-storage.ts), plus byte-identical pass-through for legacy-only and damaged-hybrid scenes. No wiring defect surfaced.

cosarah:
Triaged discovery findings against PR HEAD: two issues remain actionable.



1. **[P2] Require a runnable v2 project before taking precedence**

`resolvePBLContent` currently treats any container-valid `projectV2` with at least one milestone as authoritative. A persisted hybrid can therefore select a v2 project whose active milestone has no usable microtask (or whose roles lack an Instructor), shadowing runnable legacy data. `normalizeProjectRuntime` cannot create a missing task, and `runInstructorTurn` exits with `NO_ACTIVE_MICROTASK`. The precedence check should require a runnable Instructor/thread and usable microtask before choosing v2.



2. **[P3] Assert runnable planner outputs in the routing tests**

The routing fixtures use milestones with empty `microtasks` arrays and roles without an Instructor type, yet they pass `hasPBLProjectV2Containers` and `validateAppScene`. The real planner gates currently reject these shapes, but the routing tests would miss a regression that bypasses or weakens that gate. The fixtures/helper should assert the runnable planner contract, not only container presence.

wyuc:
Both addressed:

**Runnable precedence**: the resolver's v2 arm now uses `isRunnablePBLProjectV2` — containers plus what the workspace runtime actually cannot do without: at least one instructor-type role (normalization can create the instructor's thread but never the role, progress.ts) and at least one milestone containing a microtask object (normalization only activates existing tasks; without one `runInstructorTurn` exits `NO_ACTIVE_MICROTASK`). A hybrid whose v2 payload misses either falls back to the upgraded legacy config, in the renderer and in the video-export precedence alike; persistence follows the same verdict and passes the hybrid through untouched.

**Planner-contract fixtures**: the routing fixtures now carry an instructor role and microtask-bearing milestones, and the success-path assertions run the generated content through `isRunnablePBLProjectV2` in addition to the container predicate and `validateAppScene`, so a regression that weakens the planner gate fails the routing tests directly.

Full suite and CI green.

wyuc:
Thanks for the thorough passes. Nine rounds in, this review has closed out roughly 30 findings, and the recent ones are trending toward shapes that only hand-edited or partially corrupted historical data can produce, plus a couple that no longer reproduced on HEAD.

For further rounds, please include an impact assessment with each finding: what realistically stored data or user flow triggers it, and what the consequence is for someone actually using the product. Findings whose trigger requires artificially constructed payloads are better filed as accepted limitations than as review blockers at this point.

If nothing meets that bar against the current HEAD, please approve.

cosarah:
Triage of the discovery findings: all three are KEEP.\n\n1. Hybrid validation does not match resolver behavior — lib/document-store/validators.ts:92-97 (P2)\nA title-only legacy config makes isEmptyLegacyPBLConfig return false, so a damaged projectV2 is accepted at the write barrier. On reload, resolvePBLContent rejects the v2 arm and also rejects the legacy arm because it has no issues, so the scene falls through to the empty-project state. This is a silent, user-visible failure for affected hybrid documents.\n\n2. Single-call planner errors can skip the loop fallback — lib/generation/scene-generator.ts:988 (P2)\nThe break runs for every single-call failure that is not a PlannerV2Error, including parsing, runtime, and provider errors. That prevents the explicitly intended loop planner fallback from running even when it could succeed, reducing ordinary PBL generation reliability.\n\n3. Runnable v2 validation does not verify role/task identities — lib/pbl/v2/types.ts:994-995 (P2, lower-frequency edge case)\nisRunnablePBLProjectV2 checks only for an instructor role type and a non-empty microtask array. Payloads such as {type: "instructor"} and {} therefore become authoritative v2 projects despite missing IDs and titles, while runtime lookups, task-update APIs, and UI labels depend on those fields. This is reachable with malformed imported, hand-edited, or regressed persisted payloads.

wyuc:
Applying the impact bar to the three:

**2 — fixed.** You were right that the fallback gate was inverted: it should be a denylist, not an allowlist. The loop planner now runs after ANY single-call failure except a provider/HTTP-status failure (the loop would hit the same provider the same way) or a user cancellation — so schema/parse failures (already wrapped in `PlannerV2Error`) and unexpected runtime errors both keep the intended fallback. This is the one of the three that affects every ordinary generation flow. Pinned by tests for the runtime-error fallback, the provider skip, the string-status skip, and cancellation.

**1 — accepted limitation.** The trigger needs a doubly degenerate stored scene: a legacy config with no issues (title-only shell) AND a damaged v2 payload in the same scene. On main that cohort got a zero-milestone workspace or a renderer crash; on this branch it gets the recoverable empty-project placeholder, and its document keeps saving. Rejecting it at the write barrier instead would brick saves for the same stored documents, which is the worse trade. This is the storable-vs-renderable split already documented on the persistence module and in this thread.

**3 — accepted limitation.** Role/task identity fields (`id`, `title`) are leaf data; the runnable predicate deliberately stops at what the runtime cannot proceed without (an instructor role, a microtask to activate), and the leaf-reader audit from an earlier round already guards label/lookup paths against missing scalars. Extending structural validation into leaf identity for hand-edited payloads is the depth we agreed to file as accepted rather than block on.

CI green on HEAD. If nothing else meets the impact bar, please approve.

cosarah:
KEEP — Validate v2 task and role identities before taking precedence

`isRunnablePBLProjectV2` currently checks only container shapes, `role.type === "instructor"`, and that at least one microtask object exists. It does not require the stable role/task IDs or learner-facing fields defined by `PBLRole`/`PBLMicrotask`.

A persisted or hand-edited payload such as `roles: [{ type: "instructor" }]` or `microtasks: [{}]` therefore wins over a usable legacy fallback. Runtime normalization can repair status, but it cannot invent IDs or titles: task lookup/update then receives an undefined ID, while the sidebar and Instructor prompt render undefined labels. Because the scene validator accepts these leaf-incomplete objects, this can persist across save/reload and leave the project unusable. The shared acceptance predicate should validate the identity fields (or reject the v2 arm and fall back).

wyuc:
Fixed. `isRunnablePBLProjectV2` (and its video-export mirror) now also require the identity leaves the runtime genuinely cannot synthesize: the instructor role's non-empty string `id` (thread binding) and string `name`, and a non-empty string `id` plus string `title` on every microtask (any of them becomes the active task as earlier ones complete, and lookup/update paths key on the id). Identity-less payloads now fall back to usable legacy data in the renderer, export, and persistence alike; other scalar leaves deliberately stay outside the structural predicate, guarded by the defensive readers.

That closes every kept finding. CI is green on HEAD — please approve.

cosarah:
## Triage of discovery-review findings

All four findings are **KEEP**:

1. **[P2] Complete and unique milestone/task identities — KEEP**
   `isRunnablePBLProjectV2` checks Instructor identity and microtask IDs/titles, but not milestone IDs/titles or uniqueness. A persisted or hand-edited payload with missing or duplicate identities can therefore take v2 precedence while runtime `.find` lookups, learner-state maps, React keys, and event references address the wrong object.

2. **[P2] Align write validation with runtime eligibility — KEEP**
   `validateAppScene` accepts any v2 payload passing `hasPBLProjectV2Containers`, while `resolvePBLContent` requires `isRunnablePBLProjectV2`. A v2-only imported or hand-edited payload with no Instructor or usable tasks can save successfully, then resolve to the empty-project state after reload.

3. **[P3] Preserve non-empty legacy projects without issue rows — KEEP**
   `resolvePBLContent` explicitly requires `issueboard.issues.length > 0`, even though `isEmptyLegacyPBLConfig` treats a title, agents, or chat as non-empty and the pre-cutover renderer upgraded/rendered such content. Those partial historical scenes are now silently shown as `pbl.emptyProject`.

4. **[P3] Preserve supported legacy locales — KEEP**
   `detectLegacyLanguage` recognizes only Japanese, Korean, Arabic, Cyrillic, and Han scripts, mapping Han to `zh-CN` and all other text to `en-US`. Supported Spanish, French, Portuguese, and Traditional Chinese legacy scenes therefore receive compatibility criteria/debrief text in the wrong language.

**KEEP set: 1, 2, 3, 4.**

wyuc:
Closing out the last four as maintainer:

1 and 2 are filed as accepted limitations under the impact bar stated earlier: both require imported or hand-edited payloads (planner output is gated and contract-tested for identities and instructor presence), and the storable-vs-renderable split is deliberate, documented design.
3 asks to reverse a behavior this review itself requested two rounds ago (metadata-only legacy shells resolving to the recoverable empty state instead of launching a zero-task workspace). Keeping the current behavior; if we want a richer treatment for partial legacy shells it deserves its own issue rather than another round here.
4 is behavior moved verbatim from the pre-existing v1 compat code, not introduced by this PR; happy to take a follow-up issue for extending legacy language detection.

Twelve rounds and ~38 findings in, every reproducible product-impacting finding has been fixed and pinned by tests, internal cross-review converged, and CI is green. Merging as maintainer; thanks for the depth of the review — it materially hardened the legacy read path.

Issue #1057, opened 2026-08-05
[Epic] @openmaic/generation — extract the course generation pipeline into a package (#720 next phase)


wyuc:
## Goal

Promote the course generation pipeline — outline → per-scene content → per-scene actions — from `lib/generation/**` into a published `@openmaic/generation` package, so that generating a classroom is a capability any plain-Node consumer can use (server workers, CLI tooling, eval harnesses, CI), not something only reachable from inside the Next.js bundler.

This is the natural next phase of #720: "every contract, renderer, importer and exporter lives in a package with a published boundary; the app keeps only business glue." Generation is the largest remaining in-app capability with an already-clean contract surface.

## Why now

**The pipeline is not loadable outside the Next bundler today.** The repo-root `package.json` has no `"type": "module"`, so a plain Node process loads `lib/**` as CommonJS — while `@openmaic/dsl` and `@openmaic/storage` are ESM-only (their `exports` maps declare an `import` condition and no `require`). A bare-Node `require` of the generators therefore fails with `ERR_PACKAGE_PATH_NOT_EXPORTED`. The app's own server path only works because Next bundles it. Any out-of-app consumer — a worker process, a CLI, an eval runner — is currently impossible without a private build step.

**The hard part is already done: the model seam is clean.** The generators take an injected `AICallFn` (`lib/generation/pipeline-types.ts:60`) and do no provider resolution, no fetch, no env reads of their own. Extraction is mostly moving code across a boundary the code already respects.

**Two orchestrators duplicate step logic.** The browser flow and `lib/server/classroom-generation.ts` each re-implement the step sequence around the same primitives. A package of primitives is the precondition to deduplicating them (#697 lives in the same area).

## Where the code is today

| Piece | Location | Notes |
|---|---|---|
| Outline generation | `lib/generation/outline-generator.ts` (213 lines) | `generateSceneOutlinesFromRequirements`, `applyOutlineFallbacks`, `sanitizeProceduralSkillOutline` |
| Scene content + actions | `lib/generation/scene-generator.ts` (1916 lines) | `generateSceneContent`, `generateWidgetContent`, `generateSceneActions`, `createSceneWithActions` |
| Pipeline types | `lib/generation/pipeline-types.ts` | `AICallFn`, `GenerationResult` — the injection seam |
| Prompt system | `lib/prompts/` | file-based templates + `{{snippet:…}}` composition + conditional blocks; loader reads from repo-relative paths |
[...]

PR #1065, opened 2026-08-06
feat(generation): move outline generation into the package (#1057 Part B)

wyuc:
Part B of #1057, per the [Part 0 RFC](https://github.com/THU-MAIC/OpenMAIC/issues/1057#issuecomment-5193492850): the outline layer moves into `@openmaic/generation`. Shim-first — no app-side file changes; `lib/generation/**` originals stay untouched until Part D repoints consumers.

## What moves

- `generateSceneOutlinesFromRequirements`, `applyOutlineFallbacks`, `sanitizeProceduralSkillOutline`, `DEFAULT_LANGUAGE_DIRECTIVE` (from `lib/generation/outline-generator.ts`), `changeOutlineType`, `uniquifyMediaElementIds` (outline-scoped; the rest of scene-builder is Part C), `parseJsonResponse`/`tryParseJson`, and the outline-facing pure types. Adaptations limited to import paths, inlined constants, and logger injection — zero behavior or prompt change.
- **`buildOutlinePrompt`** (RFC §3.5): the prompt-construction half extracted as an exported pure function; the generator calls it internally, and golden snapshots pin its output across the conditional matrix.
- **Logger injection** (RFC §3.6): `GenerationLogger` structural interface, no-op default, optional per-call. The package contains zero `process.env` reads.
- Runtime deps added as actually imported: `jsonrepair`, `nanoid`. The boundary lint allowlist now names the RFC's approved leaf dependencies.

## The second consumer (RFC §7, outline half)

`scripts/generation-node-smoke.mjs`: a bare-Node ESM script — flags only (`--requirement/--endpoint/--model[/--api-key]`), an `AICallFn` implemented with plain `fetch` against any OpenAI-compatible endpoint, no app imports, no env reads. CI runs it against a local canned-response server (`generation-node-smoke-server.mjs`); the script cannot tell the difference. Part C extends it with scene generation.

## Tests

84 package tests: ported app-side assertions for the moved code, mocked-`AICallFn` behavior paths (happy/repairable-JSON/fallbacks/type-conversion/id-collision), and `buildOutlinePrompt` golden snapshots.

## Verification (Node 22)

Package build/typecheck (src+test)/tests · `pnpm check` · `pnpm lint` (0 errors) · root `tsc --noEmit` · smoke script real run against the local server · temp-dir and dist import smokes · five-package tarball smoke · version bump `0.0.1 → 0.1.0`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,35 @@ jobs:
       - name: Unit Tests (generation)
         run: pnpm --filter @openmaic/generation test
 
+      - name: Node consumer smoke (generation outlines)
+        shell: bash
+        run: |
+          set -euo pipefail
+          port=43127
+          node scripts/generation-node-smoke-server.mjs --port "$port" >generation-smoke-server.log 2>&1 &
+          server_pid=$!
+          cleanup() {
+            kill "$server_pid" 2>/dev/null || true
+            wait "$server_pid" 2>/dev/null || true
+          }
+          trap cleanup EXIT
+          for attempt in {1..50}; do
+            if node -e "fetch('http://127.0.0.1:$port/health').then(r => { if (!r.ok) process.exit(1) }).catch(() => process.exit(1))"; then
+              break
+            fi
+            if [ "$attempt" -eq 50 ]; then
+              cat generation-smoke-server.log
+              exit 1
+            fi
+            sleep 0.1
+          done
+          output="$(node scripts/generation-node-smoke.mjs \
+            --requirement "Explain dependency injection" \
+            --endpoint "http://127.0.0.1:$port" \
+            --model smoke-model)"
+          printf '%s\n' "$output"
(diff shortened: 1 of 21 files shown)


xuyuanwei678 (dismissed):

wyuc:
@xuyuanwei678 sorry for the churn — your review got auto-dismissed by a push that raced it. Delta since the commit you looked at: a post-review hardening round from an independent code audit (drops three no-op `logger` options that overpromised the RFC signatures, narrows the boundary-lint allowlist to the declared dependencies, adds a CI parity guard for the duplicated prompt assets until Part D removes the app copy, two mixed-conditional prompt snapshots, minor smoke-script/log/export tweaks), one code-scanning nit (a provably-constant escape subcondition in json-repair), and a merge of main. No behavior changes to the moved generation code itself — 86 package tests green. Would appreciate a re-approve when you get a chance.

xuyuanwei678 (approved):

Issue #1049, opened 2026-08-05
[Epic] Inclass Native Tool Harness — Runtime-backed whiteboard and incremental rollout


LING-6150:
## Status

COMPLETE — the planned Native Tool Harness and RuntimeStore-backed whiteboard rollout has landed.

- PR1 RuntimeStore foundation: #1075
- Shared Pi transport prerequisite: #1108
- PR2 Native Child + scoped Spotlight: #1111
- PR3 Native Web Search: #1133
- Native ownership/completion cleanup: #1148
- PR4 minimal Native whiteboard slice: #1152
- PR5 additive whiteboard tools: #1156
- PR5 destructive and code-edit operations: #1173

The delivered whiteboard inventory now includes RuntimeStore-backed reads,
visibility controls, seven additive element tools, targeted deletion, clear,
and in-place code editing. The merged path preserves same-Child continuation,
CAS/replay/reconciliation semantics, confirmed-commit-only projection, and
Legacy isolation.

#1011 remains a historical architecture spike and is superseded by the
incremental merged PRs above. It is not an implementation base and will be
closed without merge.

A future Legacy cutover or broader cleanup, if needed, should be handled
separately rather than extending this completed Epic.

## Goal

Deliver the Inclass Native Tool Harness through small, independently reviewable
PRs while moving learner-produced live whiteboard state onto OpenMAIC's existing
RuntimeStore abstraction, following the latest architecture review on #1011.

This Epic reuses the existing persistence layer; it does not introduce or
redesign RuntimeStore.

### 1. Server tools

Examples: Web Search.

- Executed on the server.
- Results return directly to the same Native Child.
- No browser delivery or browser acknowledgement is involved.

### 2. Transient UI effects

Examples: spotlight, laser, scrolling, focus, and proposed native
`wb_open` / `wb_close`.

- Best-effort SSE dispatch.
- Do not write RuntimeStore.
- Do not wait for browser rendering or terminal browser receipts.
- Tool success means the server accepted and dispatched the effect; it does not
  prove that the browser visibly completed it.

### 3. RuntimeStore-backed whiteboard state

Learner-produced live whiteboard state is durable runtime data.

- Reuse the RuntimeStore already present in OpenMAIC; #869 and #939 are
  background references for the existing abstraction and backend seam.
- Records are append-only and ordered by store-assigned per-session `seq`.
- `wb_read` reads the persisted and folded RuntimeStore state, not browser state
  or a request-start snapshot.
[...]

PR #1011, opened 2026-07-29
feat(pi): add native child tool runtime foundation

LING-6150:
## Summary

This PR builds the first Inclass Native Tool Runtime vertical on top of the evidence-aware Pi Director runtime merged in #971.

It is intentionally broader than replacing one JSON action with one tool call. The goal is to prove the bounded runtime contract:

```text
Child native tool_call
→ Runtime authorization and execution
→ server/client result or ACK
→ tool_result returned to the same Child
→ Child continues, corrects, or completes
→ structured result returned to the Director
```

The current whiteboard scope is deliberately limited to one Teacher `wb_draw_text` tool. Legacy JSON actions remain available behind default-off feature flags.

### Phase 0 — bounded Native Child runtime

- Typed Child run, tool execution, identity, depth, budget, cancellation, timeout, and trace contracts.
- Same-Child `tool_call → tool_result → continuation`.
- Finite settlement when a tool or provider ignores abort, with late-event isolation.
- No recursive Agent tree; Inclass remains `Director → Teacher / Assistant / Student`.

### Phase 0T — task state and terminal control

- Request-scoped `TaskContract` with monotonic revision and trusted completion events.
- Runtime preflight for `cue_user` and `close_session`.
- `TASK_INCOMPLETE` is returned as a tool result so the same Director can continue within its budget.
- Explicit clarification, explicit real-user turns, explicit session end, rejection limits, tool-attempt limits, and Runtime exhaustion are distinguished.
- Requests without an explicit contract remain `observe_only`; the current UI does not yet derive trusted outcomes from natural language.

### Phase 1 — Native Child Web Search

- Web Search can be selected and executed by the Child rather than prepared by the Director.
- Success and explicit failure are returned to the same Child context.
- Timeout, abort, missing source URLs, stale evidence, and final Director result provenance are covered.
- Existing Director/Legacy Web Search behavior remains available when the Native Child flag is off.

### Phase 2 — first client-effect loop

- One Teacher `wb_draw_text` native tool is wired end to end.
- Ordered presentation barrier: preceding visible text is committed before browser execution; TTS completion is not required.
- `accepted` does not resume the Child; only authoritative `effect_committed` returns success.
- Capability-bound ACK endpoint, target binding, idempotency key, args digest, stable element ID, content postcondition, duplicate delivery/ACK handling, response-loss replay, hard deadline, pause/resume, cancellation, late ACK, and bounded tombstones.
[...]

PR #1075, opened 2026-08-07
feat(runtime): add learner whiteboard RuntimeStore foundation

LING-6150:
## Context

PR 1 of epic issue #1049.

This PR introduces the internal RuntimeStore foundation for learner-produced
live whiteboard state. It follows the existing RuntimeStore identity and
ordering model rather than introducing a whiteboard-specific persistence,
generation, or incarnation mechanism.

This branch was created from current `main`. It does not inherit the browser
authority or revisioned-mutation architecture from Draft PR #1011; that PR
remains a reference spike.

## What this PR does

- Defines a versioned RuntimeStore payload for learner whiteboard state.
- Adds strict payload and whiteboard validation shared by Browser, PostgreSQL,
  and HTTP RuntimeStore paths.
- Adds deterministic whiteboard RuntimeSession selection scoped by
  `(stageId, learnerKey)`.
- Enforces the Whiteboard application invariant of one active session per
  learner and Stage.
- Adds append/fold services using RuntimeStore `expectedLastSeq` CAS.
- Adds stable operation IDs, duplicate replay handling, and read-only operation
  reconciliation for append-response-loss cases.
- Rejects stale or invalid domain transitions before persistence.
- Uses the existing RuntimeStore maintenance-lock family for whiteboard append
  and Stage runtime cleanup, preserving its existing shared/exclusive and
  fallback behavior.
- Pins the existing same-ID RuntimeStore lineage semantics:
  - successful cleanup followed by same-ID restore starts with empty whiteboard
    Runtime state;
  - cleanup failure must not report an empty Runtime state;
  - whiteboard append cannot cross a successfully settled Stage cleanup;
  - once RuntimeStore whiteboard state exists, same-ID restore does not
    re-trigger Legacy import.
- Adds a provenance-gated Legacy `Stage.whiteboard` import seam.
- Preserves canonical JSON keys as ordinary own data properties so exact-key
  validation cannot be bypassed by special keys such as `__proto__`.

## RuntimeStore boundary

Whiteboard Runtime state follows the existing RuntimeStore identity and ordering
rules:

```text
(stageId, learnerKey)
→ one active whiteboard RuntimeSession
→ append-only RuntimeRecord sequence
→ expectedLastSeq CAS
→ deterministic fold
```

The production service resolves learner identity from trusted application
context. The caller or model cannot select a different `learnerKey`.

A stale `expectedLastSeq` commits no mutation and returns the existing
RuntimeStore conflict semantics. Exact operation replay is idempotent; a
conflicting replay cannot reuse the previous result.

This PR does not introduce a whiteboard-only generation, incarnation, registry,
distributed lock, or recovery protocol.

## Validation and fold safety

Whiteboard Runtime payloads are validated before persistence:

- exact payload, operation, source, and Whiteboard envelope keys;
- version and operation-kind checks;
- lossless JSON validation;
- canonical JSON cloning and digest generation;
- generated `PPTElement` schema validation;
- unique and safe board/element identifiers;
- finite geometry;
- deterministic RuntimeRecord/session/sequence validation during fold.

Canonical cloning preserves every enumerable own JSON key without invoking the
legacy `__proto__` setter. Out-of-contract root or nested keys remain visible to
the exact-key validator and are rejected before learner/session resolution or
persistence.

Malformed records and invalid state transitions fail closed rather than being
silently skipped.

## Replay and reconciliation

The append path distinguishes:

- exact committed replay;
- conflicting reuse of an operation ID;
- stale `expectedLastSeq`;
- append response loss;
- post-commit reconciliation failure;
- invalid domain transitions.

Response-loss recovery is read-only. A request is treated as replayed success
only when the committed record proves the same operation ID and canonical
payload digest. The service does not perform blind append retries.

Invalid anticipated tails and attempts to replace existing RuntimeStore state
with a second distinct Legacy import are rejected before persistence.

## Same-ID cleanup semantics

Whiteboard follows the current shared RuntimeStore lineage rule for
`(stageId, learnerKey)`.

- Successful Stage runtime cleanup followed by same-ID restore yields an empty
  whiteboard Runtime state.
- Cleanup failure retains the existing Runtime state and the established
  non-blocking diagnostic/fail-soft behavior.
- Whiteboard append and Stage runtime cleanup use the existing maintenance-lock
  family.
- Existing RuntimeStore whiteboard state prevents the Legacy import path from
  running again.

This is the current shared RuntimeStore behavior, not a general guarantee of
isolation between possible future classroom incarnations. Any future
incarnation key must be introduced consistently across RuntimeStore kinds
rather than locally for Whiteboard.

## Legacy compatibility boundary

The Legacy import seam is intentionally dormant and has no production caller.

Its policy is:

- a local BrowserDocumentStore and BrowserRuntimeStore may support a one-time
  import only under matching device-local learner provenance;
- server-backed, shared, custom, or unknown document provenance is never
  automatically attributed to the current learner;
- malformed or multiple Legacy boards are skipped as ambiguous rather than
  selecting the first or last board;
- RuntimeStore state always wins once a Runtime record exists;
- exact import replay is idempotent and cannot append a second record;
- conflicting committed Runtime state is reported as `runtime_won`;
- unreadable or unprovable settlement remains uncertain rather than being
  reported as successful;
- `Scene.whiteboards` remains opaque round-trip compatibility data;
- existing document-authored `wb_*` playback behavior is deferred to the later
  parity PR.

A static production-import Gate verifies that the dormant importer has no
caller under `app/`, `components/`, or `lib/`.

Activating this seam is deferred to a separately reviewed compatibility
cutover.

## Reviewer focus

Please focus on:

1. whether RuntimeSession identity and the one-active-session policy correctly
   reuse existing RuntimeStore semantics;
2. whether append, mandatory `expectedLastSeq` CAS, operation replay, and
   reconciliation preserve atomicity and idempotency;
3. whether same-ID cleanup semantics match the existing shared RuntimeStore
   lineage rule;
4. whether Whiteboard append and Stage cleanup use the existing
   maintenance-lock family without introducing a whiteboard-specific lock;
5. whether payload validation, canonical cloning, persistence validation, and
   deterministic fold fail closed across Browser, PostgreSQL, and HTTP paths;
6. whether the dormant Legacy import boundary is sufficiently isolated from
   production behavior;
7. whether the PR remains limited to the RuntimeStore foundation defined for
   PR 1 of #1049.

## Explicit non-goals

This PR does not:

- expose `wb_read` or any Native whiteboard mutation tool;
- add browser projection or SSE delivery;
- add browser acknowledgement or receipt machinery;
- cut over existing `Stage.whiteboard` readers or writers;
- activate Legacy import in production;
- introduce generation/incarnation IDs;
- introduce session convergence or lifecycle recovery;
- add a whiteboard-specific or distributed lock;
- change Legacy whiteboard actions or current product behavior;
- include the Browser Authority, observation ledger, target registry,
  accepted/terminal ACK, tombstone, or revisioned mutation architecture from
  Draft PR #1011.

## Cross-review

Before requesting maintainer review, this PR went through independent full-diff
reviews by Codex and Claude Opus 5 against the #1049 architecture decisions and
the frozen PR 1 implementation scope. Subsequent fixes received targeted
closure reviews.

Review outcome:

- Both full-diff reviews agreed that RuntimeStore ownership, session identity,
  mandatory CAS, replay/reconciliation, cleanup, Legacy dormancy, and scope
  isolation were correctly implemented.
- One single-sided P2 finding identified that canonical cloning could silently
  drop an own `__proto__` key before exact-key validation.
- The finding was manually adjudicated as valid and fixed in `4ff9b96`.
- Targeted closure review confirmed that root and nested `__proto__` data keys
  are preserved through canonical cloning, rejected by validation, and never
  persisted.
- The subsequent test-only Code Quality suggestion was applied in `ba3cc9f`
  without weakening the regression.
- No unresolved P0, P1, or P2 findings remain.

This section records agent-assisted review evidence and does not replace
maintainer review.

## Verification

Local verification:

- Runtime/Persistence/Document/Whiteboard regression Gate:
  20 files, 255 tests passed
- TypeScript: PASS
- targeted ESLint: PASS
- targeted Prettier: PASS
- `git diff --check`: PASS

GitHub CI for the current PR HEAD:

- Lint, Typecheck & Unit Tests: PASS
- E2E Tests: PASS
- Storage Contract (PostgreSQL 16): PASS
- CodeQL: PASS
- Render Service: PASS


wyuc (commented):
Thanks for putting this together. The Legacy import seam is careful and conservative, but I think there is still a scope checkpoint to resolve before we review this as the RuntimeStore foundation promised by PR 1 of #1049.

PR 1 was intended to define the minimal learner-whiteboard operation schema, its deterministic fold, and the domain-level append/CAS contract. In the current diff, however, the V1 operation union contains only `legacy_snapshot_imported`, and the fold can only install that whole Legacy snapshot. There is no normal learner-produced whiteboard mutation yet. As a result, this PR proves that Legacy data can be imported safely, but it does not yet prove that the payload shape, fold, post-state contract, or service abstraction work for the live mutations that the foundation is meant to support. Those decisions are effectively deferred to PR 4, after this fairly large long-lived contract has already merged.

I do not think we should review this purely as a migration PR, because it also establishes the production `whiteboard` RuntimeSession, payload V1, record validation, fold, replay, and application-service APIs that later PRs will inherit. Could we resolve this in one of two ways?

1. Add at least one minimal, tool-agnostic learner mutation and its fold/integration tests here. This does not require exposing a Native tool or adding browser delivery.
2. Explicitly rescope and rename this as Legacy import plus RuntimeStore plumbing, and leave the learner-whiteboard operation/fold portion of PR 1 visibly incomplete for a follow-up review.

My preference is the first option, because it validates the core abstraction before we commit to the surrounding migration and validation machinery.

LING-6150:
@wyuc Thanks for your precious opinions.  I agree with the scope gap. I've reviewed the current RuntimeStore contract and would like to confirm the minimal learner mutation shape with you.

I propose adding one internal, tool-agnostic operation:

```ts
{
  kind: 'element_added';
  whiteboardId: string;
  element: PPTElement;
}
```

I chose `element_added` rather than `wb_draw_text` because PR 1 should validate
the whiteboard domain contract without coupling it to a Native tool or browser
delivery path. One incremental element operation is sufficient to exercise:

```text
canonical learner operation
→ expectedLastSeq CAS
→ deterministic fold
→ committedSeq + folded post-state
→ exact replay / stale conflict
```

For an empty learner whiteboard, the first committed mutation would materialize
a canonical board with the existing canvas defaults:

```text
viewportSize: 1000
viewportRatio: 0.5625
white background
empty animations
```

`whiteboardId` would remain explicit in the persisted operation so the fold is
fully record-driven, but it would be supplied by a trusted application-side
construction rather than accepted from model output. Once the first mutation
commits, later mutations must target the ID in the folded state.

The implementation would use the same pure transition for pre-commit validation
and deterministic fold, so invalid target or duplicate-element transitions are
rejected before persistence.

This remains limited to the RuntimeStore domain foundation. It would not add
Native tools, `wb_read`, SSE/browser projection, browser acknowledgement, or any
edit/delete/clear operation.

Could you please check this matches the minimal learned mutaion for our goal PR1, particularly the tool-agnostic `element_added` shape, first-mutation materialization, and
trusted ownership of `whiteboardId`?

wyuc:
Yes, the proposed `element_added` operation is the right kind of minimal learner mutation for PR 1. It exercises the real RuntimeStore contract without coupling the domain model to `wb_draw_text`, Native Child execution, or browser delivery.

I would make one simplification: omit `whiteboardId` from each learner mutation. The RuntimeSession already identifies exactly one whiteboard for `(stageId, learnerKey)`, so the mutation target is implicit. Repeating a board ID in every record creates a second identity surface, adds mismatch handling, and suggests a multi-board model that we explicitly do not have.

The minimal operation can therefore be:

```ts
{
  kind: 'element_added';
  element: PPTElement;
}
```

When folding the first learner mutation, materialize the canonical board with a deterministic application-owned ID derived from the RuntimeSession identity and the shared canonical defaults. When state already exists, including state created by Legacy import, preserve the folded board ID and apply the element to that single board. This remains fully deterministic and record-driven because the session envelope and prior folded state are part of the fold input.

Please keep the proposed pre-commit transition validation, duplicate-element rejection, exact replay, stale CAS conflict, and folded post-state tests. With that simplification, this matches the PR 1 scope I had in mind.

LING-6150:
@wyuc Thanks for your updated suggestions, I agree with it. So I’ll remove `whiteboardId` from `element_added` and derive the initial application-owned board ID deterministically from the selected RuntimeSession identity. Existing state, including Legacy-imported state, will preserve its folded board ID.

I’ll keep the change limited to the operation schema, shared preflight/fold
transition, CAS/replay behavior, and integration tests, rather than adding Native
tools or browser delivery.

LING-6150:
@wyuc  I’ve updated PR 1 following your latest feedback.

The RuntimeStore V1 contract now includes the minimal tool-agnostic
`element_added` learner mutation. It omits `whiteboardId`: the first learner
mutation materializes the canonical board from the selected RuntimeSession
identity, while later mutations preserve the existing or Legacy-imported board
identity and metadata.

The implementation also covers pre-commit transition validation, stale CAS,
exact and conflicting replay, duplicate-element rejection, and folded
post-state verification.

The final cross-review and local gates pass at `8f34f24` (8 files / 86 tests).

Could you please review whether this now satisfies the intended RuntimeStore
foundation scope for PR 1 and whether it is ready to merge?

wyuc (approved):
The PR 1 RuntimeStore foundation scope is now satisfied, including the tool-agnostic `element_added` mutation and deterministic fold. The dormant same-ID restore Legacy re-import scenario is accepted as a non-blocking limitation for this PR and should be revisited before the Legacy importer is activated.

PR #684, opened 2026-06-08
feat(rag): add local knowledge workflow

jackefn:
## Summary

This PR adds a local knowledge base and RAG workflow for grounding classroom generation in uploaded reference materials.

Users can upload and index materials into a PostgreSQL/pgvector-backed knowledge base, enable **Local Knowledge** during generation, review retrieved excerpts before continuing, and inspect the saved evidence snapshot from the generated classroom. Retrieved excerpts are passed into outline generation, scene content prompts, widget prompts, and teacher action prompts so generated courses can follow source-specific terminology, procedures, and constraints.

## Related Issues

N/A

## Changes

- Add server-side knowledge base APIs for document listing, deletion, retrieval, and RAG snapshot confirmation.
- Add PostgreSQL/pgvector-backed knowledge repository helpers, text chunking, BGE embedding support, retrieval configuration normalization, and snapshot persistence.
- Add a `/knowledge-base` UI and a `/knowledge-base/settings` page for managing indexed materials and retrieval hyperparameters.
- Add a **Local Knowledge** generation toolbar toggle and persist the user preference locally.
- Add a retrieval review step in the generation preview flow so users can confirm the specific excerpts used for generation.
- Pass confirmed retrieved context into outline generation, scene content generation, widget generation, and teacher actions.
- Add classroom evidence panels/docks so generated classrooms can expose the saved RAG evidence snapshot.
- Add i18n coverage for the Local Knowledge UI and generation step across supported locales.
- Add focused tests for RAG config normalization, chunking/embedding helpers, retrieval selection, and prompt grounding.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [x] CI/CD or build changes

### Steps to reproduce / test

1. Configure PostgreSQL/pgvector and BGE embedding environment variables, then start the app.
2. Open `/knowledge-base`, upload or index reference material, and confirm that the material appears in the knowledge base.
3. On the home page, enable **Local Knowledge**, enter a course requirement, and start generation.
4. Review the retrieved excerpts in the generation preview, select the excerpts to keep, and continue generation.
5. Open the generated classroom and verify that the RAG evidence panel shows the saved sources and excerpts.

### What you personally verified

- Ran `pnpm check:i18n-keys`.
[...]

PR #1078, opened 2026-08-09
feat(rag): add document lexical retrieval foundation

jackefn:
## Summary

This PR delivers the first RAG foundation for Milestone 5. It connects the existing document Transform pipeline to a provider-neutral, text-first knowledge ingestion path and adds deterministic lexical retrieval for document and HTML blocks.

The implemented contract makes the retrieval boundary explicit: `workspaceId` is always required, resources and replacements use an exact optional `courseId`, and query/delete operations require an explicit `courseScope: 'exact' | 'all'`. A resource version represents one complete indexed chunk set and is atomically replaced as a unit, including smaller and empty replacements, without exposing partial state on failure.

The implementation intentionally stops at the indexing/retrieval foundation. It does not expose a user-facing knowledge-base workflow, evidence snapshot contract, vector provider, or generation grounding path. This keeps the PR independently testable while preserving deterministic document lineage and locators for later multimodal document, video, and virtual-experiment RAG work.

## Related Issues

Related to [Issue #621](https://github.com/THU-MAIC/OpenMAIC/issues/621).

This is intentionally a non-closing reference: #621 is the still-open MAIC ETL RFC, and later Milestone 5 PRs remain outside this foundation change.

The design was informed by [PR #684](https://github.com/THU-MAIC/OpenMAIC/pull/684), but this PR does not copy its PostgreSQL/pgvector, embedding, UI, snapshot, or generation-integration scope.

## Changes

- Add a provider-neutral RAG contract under `lib/rag` for document/HTML resources and chunks, document locators, versioned lineage, metadata filters, index capabilities, and scored hits. Resource, chunk, and replacement identities use `KnowledgeExactScope`; query and delete requests use a discriminated `KnowledgeIndexScope` so exact-unscoped and workspace-wide operations cannot be confused.
- Add `ingestDocumentForRag`, which runs the existing default document Transform registry with `purpose: 'rag'`, preserves transformed artifacts, diagnostics, metrics, and Transform lineage, removes provider-specific `providerRaw` from the returned RAG artifact, and marks resources with no searchable chunks as `partial`.
- Define `resourceVersionId` from an explicitly ordered fingerprint containing resource scope, source hash, extractor, ordered Transforms, effective chunk policy, and normalized indexed resource metadata. Scope-shaped metadata keys are excluded, metadata key order does not affect the ID, and metadata changes that affect filtering produce a new resource version.
- Add block-aware document and HTML chunking with empty/image-only block filtering, sanitized HTML projection, `<br>` line-break preservation, valid block-end tags with HTML ASCII whitespace, paragraph/whitespace-first splitting, collision-safe stable chunk IDs, SHA-256 content hashes, and page/block/heading locators.
- Define `maxChars` in grapheme clusters and use deterministic Unicode-safe boundaries that do not split surrogate pairs, combining sequences, or extended grapheme clusters. Text that exactly fits `maxChars` remains in one chunk, while true overflow still splits; the changed behavior is recorded in chunk-policy version `1.1.2` and effective lineage.
- Add `InMemoryLexicalIndex` with consistent lowercase + NFC normalization for indexed text and queries, Unicode/CJK token matching, deterministic scoring and tie-breaking, explicit exact/all course scope, metadata filtering, top-K limiting, scoped deletion, and defensive chunk snapshots.
- Replace chunk-level upsert semantics with atomic `replaceResourceVersion`: the complete candidate set is validated and cloned into a private map before one active-map swap. Smaller and empty replacements remove stale chunks, and validation or later-clone failure leaves the previous complete set intact.
- Keep the public surface limited to behavior implemented here: document/HTML modalities and document locators only. Grounding snapshots, persistence/vector providers, media/video/experiment locators, retrieval confirmation, and generation integration remain follow-up work.
- Add focused regression coverage for Transform-first ingestion, partial/empty resources, deterministic rebuilds, canonical version identity, indexed-metadata identity, duplicate/collision-safe block IDs, HTML sanitization and line-break projection, Unicode-safe and exact-fit chunking, lexical ranking/filtering, complete-set replacement, later-clone failure atomicity, exact-unscoped query/delete, explicit workspace-wide operations, and workspace/course isolation.

## Type of Change

- [x] Bug fix (non-breaking correctness fixes within this unreleased PR surface)
- [ ] Breaking change (fix or feature that would cause existing released functionality to change)
- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Install the repository dependencies with `pnpm install --frozen-lockfile`.
2. Run the focused RAG tests:

   ```bash
   pnpm exec vitest run tests/rag
   ```

3. Run the combined RAG, document-transform, and generation-boundary regression set:

   ```bash
   pnpm exec vitest run tests/rag \
     tests/document/transform-pipeline.test.ts \
     tests/document/transform-registry.test.ts \
     tests/document/text-extractor.test.ts \
     tests/generation/outline-type.test.ts \
     tests/generation/scene-api-retry-boundary.test.ts
   ```

4. Run the changed-scope static checks:

   ```bash
   pnpm exec tsc --noEmit --pretty false
   pnpm exec eslint lib/rag tests/rag
   pnpm exec prettier --check lib/rag tests/rag PR_M5_RAG_FOUNDATION.md
   pnpm check:i18n-keys
   git diff --check
   ```

### What was verified locally

- `tests/rag` passed: 6 files, 43 tests.
- The combined RAG/document/generation regression set passed: 11 files, 73 tests.
- The existing document-transform regression set passed: 3 files, 10 tests.
- The existing generation-boundary regression set passed: 2 files, 20 tests; no generation code was changed.
- Scoped ESLint, Prettier, the i18n key check, and `git diff --check` passed.
- The full merged-branch TypeScript command still reports unrelated PBL v2 type-migration errors outside this PR. Its output contains no diagnostics from `lib/rag` or `tests/rag`.
- Repeated ingestion of the same artifact and source descriptor reproduces the same resource version, chunk IDs, and hashes; input artifacts remain unchanged.
- Reordering `KnowledgeVersion` object properties or normalized metadata keys does not change `resourceVersionId`; changing indexed metadata does change it.
- Exact-unscoped query/delete affects only resources whose `courseId` is absent. Workspace-wide query/delete occurs only with explicit `courseScope: 'all'`, and exact course operations remain isolated across workspaces and courses.
- Indexed and queried Unicode text is normalized consistently: composed `café` and decomposed `cafe\u0301` retrieve and score identically. CJK character-level matching and locale-independent case handling remain covered.
- HTML projection preserves `<br>` variants and valid closing tags such as `</p >` and `</li\t>` as text boundaries while removing script content and unsupported markup.
- Grapheme chunking preserves emoji and combining clusters, keeps `abc d` in one chunk at `maxChars=5`, splits true overflow, and handles the 100,000-character regression input without repeated rescans.
- Complete-resource replacement removes stale chunks for smaller and empty sets. A forced failure while cloning a later replacement chunk preserves the previous complete set and exposes none of the candidate chunks.
- A direct driver through the public `lib/rag` barrel also exercised exact/all scope, NFC-equivalent retrieval, whitespace-bearing HTML end tags, exact-fit chunking, canonical lineage/metadata identity, and later-clone failure atomicity.
- Repository-wide CI/full-suite success is not claimed by this PR description. Known unrelated PBL migration and local legacy WebGL-dataset issues remain outside the changed RAG scope.
- Not verified in this PR: PostgreSQL/pgvector, BGE or other embedding services, browser UI/API flows, immutable evidence snapshot persistence and authorization, generation prompt grounding, video transcription, or virtual-experiment/WebGL indexing. These are subsequent Milestone 5 PRs by design.

### Evidence

- [x] Manually tested locally through the public library barrel; no user-facing UI or API is included in this foundation PR
- [ ] Screenshots / recordings attached (not applicable; no UI changes)
- [x] Focused automated tests pass (`tests/rag`: 6 files, 43 tests; combined RAG/document/generation set: 11 files, 73 tests)
- [x] Existing document-transform regression tests pass (3 files, 10 tests)
- [x] Existing generation-boundary regression tests pass (2 files, 20 tests)
- [x] RAG-scope TypeScript diagnostics are clear; unrelated merged-branch PBL diagnostics are documented above
- [x] Scoped lint and formatting checks pass
- [x] Local i18n key check passes (`pnpm check:i18n-keys`)
- [x] Git whitespace/error check passes (`git diff --check`)

## Checklist

- [x] I have performed a self-review of the code and test boundaries
- [x] I have added/updated documentation as needed (this PR description)
- [x] My code introduces no warnings in the changed-scope lint checks
- [x] The public contract exposes only document/HTML behavior implemented and tested in this PR
- [x] Workspace/course access scope is explicit and separate from auxiliary metadata filters
- [x] Complete-resource replacement, deterministic version identity, and failure atomicity are covered by regressions
- [x] This PR is limited to the RAG foundation and does not include vocational/task-engine/procedural-skill changes
- [x] RAG implementation commits contain only RAG implementation/tests; the separate upstream merge commit only synchronizes the PR branch with remote main
- [x] This Markdown remains a local PR-body file and is not included in the implementation commits

## Non-goals and follow-up

This PR does not include PostgreSQL/pgvector persistence, a vector provider or embedding-model integration, knowledge-base APIs or UI, retrieval confirmation, immutable evidence snapshots, generation prompt integration, classroom evidence panels, media/video/experiment locators, video transcript/keyframe retrieval, or virtual-experiment/WebGL resource graphs.

Follow-up PRs should introduce those capabilities together with their own storage, ownership, authorization, lifecycle, and regression-test semantics rather than expanding this foundation contract ahead of implementation.


wyuc:
Thanks for moving Milestone 5 forward. The overall pipeline follows #621 well: RAG builds on the document Transform layer, stays provider-neutral, and lands as a focused foundation before persistence, UI, or generation integration.

I am not ready to merge this yet because the PR also defines several domain semantics that #621 did not specify. I would like us to make those semantics explicit before treating these types as public contracts:

1. What is the replacement and versioning unit? If a new version of a resource produces fewer chunks, does indexing atomically replace the previous chunk set? The current `upsert` contract only updates incoming chunk IDs, so removed content can remain searchable.
2. Which fields define retrieval scope? `workspaceId` and `courseId` are first-class resource fields, but filtering is implemented through chunk metadata. We should distinguish access or tenant scope from optional metadata.
3. What exactly does a grounding snapshot represent? Is it an immutable set of retrieved evidence, assembled generation context, or only a retrieval result ID? Its ownership and lifecycle should be clear before `GroundingContextRef` becomes a shared contract.
4. Which locator types are supported by current behavior? The document locator is exercised by this PR, while the media and experiment locators appear to be future-facing contracts.

My preference is to define one invariant clearly: a resource version produces a complete chunk set, the index replaces that set atomically within an explicit scope, retrieval results can be frozen into an immutable evidence snapshot, and generation references that snapshot by ID.

We can either document and enforce those invariants in this PR, or narrow this PR to the document chunking and lexical reference implementation and leave the broader contracts for a follow-up design discussion.

A few smaller correctness issues also need follow-up, including HTML `<br>` projection and Unicode-safe chunk boundaries, but I think the domain semantics should be settled first.

jackefn:
I checked each point against the current implementation. The public `lib/rag` barrel currently exposes more domain surface than this PR implements, and the existing `upsert(chunks)` behavior can leave stale chunks searchable when a resource is re-indexed with a smaller chunk set.

I will take the narrowing option for this PR while making the semantics required by the current document/HTML lexical path explicit and enforceable.

## Changes I will make in this PR

1. **Define the resource replacement and versioning unit.**

   A logical resource will remain stable within its retrieval scope, while a resource version will represent the complete chunk set produced from one source version and its extractor, Transform, and chunk-policy lineage. Indexing that version will atomically replace the previously active chunk set for the same scoped resource. A successful replacement with fewer chunks, including an empty set, will remove chunks that are no longer present; a failed replacement must leave the previous set intact.

   I will replace the current chunk-level `upsert(chunks)` contract with a resource-scoped complete-set replacement operation and add regression coverage for stale-chunk removal, empty replacements, and failure atomicity.

2. **Separate retrieval scope from optional metadata filters.**

   `workspaceId` will be the required tenant boundary, and `courseId` will be an optional first-class retrieval scope. Both will be carried explicitly through resource/chunk identity and index replace, query, and delete operations. Metadata filters will remain auxiliary search facets such as chapter, block type, role, or tags; they will not define or override access scope.

   This also removes the current ambiguity where `courseId` is first-class on a resource but the lexical implementation only sees it when callers duplicate it into chunk metadata.

3. **Remove the premature grounding snapshot contract from this PR.**

   `GroundingContextRef` and its helper will be removed from the current public API because this PR does not yet implement snapshot creation, storage, ownership checks, retention, or generation consumption.

   The follow-up snapshot PR will introduce the full invariant together with its implementation: a snapshot is a server-owned, immutable set of selected retrieval evidence, scoped to its workspace and optional course. It is not assembled prompt text and not merely a retrieval-result identifier. Generation will reference the snapshot by ID, while the server resolves and authorizes the immutable evidence record.

4. **Limit locators and modalities to behavior implemented here.**

   The public contract in this PR will expose only the document locator used by document and HTML ingestion. Media/video and experiment locators and their modalities will move back to the Milestone 5 design/follow-up work, where their temporal and resource-graph semantics can be introduced with working providers and tests.

5. **Fix the two current chunking correctness issues.**

   HTML projection will preserve `<br>`, `<br/>`, and equivalent case variants as line breaks before sanitization. Long-text splitting will use Unicode-safe boundaries so it cannot split surrogate pairs or grapheme clusters. I will define the `maxChars` unit explicitly, update the chunk-policy version for the changed boundary behavior, and add focused HTML and Unicode regression tests.

## Resulting PR boundary

After these changes, this PR will provide only the behavior it exercises end to end: Transform-first document/HTML ingestion, deterministic Unicode-safe chunking with document locators and lineage, explicit workspace/course retrieval scope, and an in-memory lexical index with atomic complete-resource replacement.

Persistence and vector providers, immutable evidence snapshots, retrieval confirmation, generation integration, and media/experiment retrieval will remain separate follow-up PRs with their own lifecycle and storage semantics.

jackefn:
Thanks, @wyuc ! I've updated the PR branch to align the implementation with the review comments and the narrowed document/HTML lexical scope. The latest version now enforces atomic resource replacement, explicit workspace/course scope, removes premature future-facing contracts, and fixes HTML `<br>` projection and Unicode-safe chunking. All focused tests and current CI checks pass.

wyuc (changes requested):
Thanks for narrowing the PR and implementing the resource-level replacement contract. I reviewed the current head `2ae1026` and am requesting changes for the following correctness and contract issues:

1. **[P1] Make omitted course scope unambiguous.** `KnowledgeScope.courseId` is optional and reused by replace, query, and delete, but absence has different meanings. `replaceResourceVersion` treats it as the exact unscoped course value, while query and delete treat it as a workspace-wide wildcard. As a result, deleting an unscoped resource with its natural scope can delete same-ID resources from every course, and the API cannot express an exact-unscoped query or delete. Please use distinct exact-scope and wildcard request shapes, or require an explicit wildcard mode. See `lib/rag/types.ts:11-14,69-83` and `lib/rag/providers/in-memory-lexical-index.ts:32-40,105-140`.

2. **[P2] Normalize Unicode before lexical tokenization.** `tokenize` lowercases without NFC or NFKC normalization, so canonically equivalent text such as indexed `café` and query `cafe\u0301` produces different tokens and fails retrieval. Please normalize indexed text and queries consistently and add a regression test. See `lib/rag/providers/in-memory-lexical-index.ts:15-16`.

3. **[P2] Preserve valid HTML end tags with whitespace.** The block-end regex matches `</p>` but not valid forms such as `</p >` or `</li\t>`. Sanitization then removes the tag without inserting a boundary, which can concatenate adjacent words. Please accept whitespace before `>` and add a projection test. See `lib/rag/chunking/document.ts:32-33`.

4. **[P2] Do not split text that exactly fits `maxChars`.** The grapheme loop fills the pending buffer to the limit without probing iterator completion. For example, with `maxChars=5`, `abc d` is split into `abc` and `d`, although all five graphemes fit. Please detect completion before applying the whitespace split and cover this boundary case. See `lib/rag/chunking/grapheme.ts:28-47`.

5. **[P2] Canonicalize `resourceVersionId` inputs.** The ID hashes `JSON.stringify(lineage)`, so semantically identical `KnowledgeVersion` objects with different property insertion order produce different resource version IDs. Please hash explicitly ordered scalar fields or canonical JSON. See `lib/rag/ingest/document.ts:68-89`.

6. **[P2] Decide whether indexed metadata is part of a resource version.** Metadata is copied into chunks and changes filtering behavior, but it is not included in `resourceVersionId`. That allows two different complete indexed chunk sets to share the same version ID. Either include normalized indexed metadata in the version identity or document and test that metadata is intentionally mutable within a version. See `lib/rag/ingest/document.ts:81-98,117-135` and `lib/rag/chunking/document.ts:122-136`.

The complete-set replacement itself looks sound by inspection: validation occurs before mutation, chunks are cloned into a local map, and the active map is swapped only after all clones succeed. Please add a failure-atomicity test where cloning a later chunk fails, plus exact-unscoped query/delete coverage, while addressing the items above.

jackefn:
Thanks, @wyuc! I've updated the PR branch to address the latest review comments and revised the original PR description to match the current implementation and narrowed document/HTML lexical scope. The latest version now uses explicit exact/all course scope, NFC-normalized lexical matching, correct HTML end-tag and exact-fit grapheme handling, canonical resource-version identity including indexed metadata, and regression coverage for later-clone failure atomicity. The focused RAG/document/generation suite passes (73/73).

wyuc (changes requested):
Thanks for addressing the previous review. I re-reviewed the updated branch at `ec1e6b7a` after synchronizing it with current `main`. The earlier six findings are fixed, but the fresh full-diff review found three remaining correctness issues:

1. **[P2] Preserve boundaries for HTML optional end tags.** `blockText` only inserts line breaks for explicit block end tags. Valid HTML may omit end tags for elements such as `p` and `li`; for example, `<p>First<p>Second` and `<ul><li>First<li>Second</ul>` are currently projected as concatenated `FirstSecond`, changing searchable tokens. Please project text using parsed block semantics, or otherwise handle the valid optional-end-tag forms, and add regression coverage. See `lib/rag/chunking/document.ts:61-74`.

2. **[P2] Detach array-valued metadata from caller-owned input.** `normalizeResourceMetadata` copies metadata entries but retains each `string[]` reference. Mutating the caller array after ingestion therefore changes returned resource/chunk metadata after `resourceVersionId` was computed, breaking the version identity invariant. Please clone array values during normalization and test post-ingestion input mutation. See `lib/rag/ingest/document.ts:102-113`.

3. **[P2] Do not report best-effort transform failures as ready.** With `failurePolicy: "best-effort"`, a transform failure is recorded and processing continues, but any non-empty chunk result is still returned with resource status `ready`. This advertises partially transformed content as complete. Please derive `partial` from failed transform records/error diagnostics as well as an empty chunk set, and add a focused regression test. See `lib/rag/ingest/document.ts:152-166`.

The synchronized branch otherwise preserves the explicit exact/all course scope, NFC matching, canonical metadata-aware version identity, exact-fit grapheme behavior, HTML end-tag whitespace handling, and failure-atomic complete-set replacement from the previous correction.

jackefn:
Thanks, @wyuc! I’ve addressed all three remaining correctness issues from the latest review:

1. **HTML optional end tags** — block-level HTML start tags now preserve boundaries for valid omitted `p`/`li` end tags and related block transitions before sanitization. Regression coverage was added for `<p>First<p>Second` and `<ul><li>First<li>Second</ul>`.
2. **Caller-owned metadata arrays** — array-valued metadata is cloned during normalization, so later caller mutations cannot change the resource or chunk metadata after `resourceVersionId` is computed. A post-ingestion mutation regression test covers this behavior.
3. **Best-effort Transform failures** — resources with failed Transform records or error diagnostics now remain `partial` even when searchable chunks are produced. A focused best-effort failure regression test verifies the status and diagnostics.

The focused RAG/Transform regression suite passes, while the previously fixed exact/all scope, NFC matching, canonical version identity, exact-fit grapheme splitting, HTML end-tag whitespace, and atomic replacement behavior remain unchanged.

wyuc (changes requested):
Thanks for addressing the three RAG correctness findings. I verified the fixes and regression coverage for optional HTML end tags, caller-owned metadata arrays, and best-effort Transform failure status. The earlier scope, Unicode, version-identity, grapheme, HTML-whitespace, and atomic-replacement fixes also remain intact, and the current checks are green.

There is one remaining scope blocker: commit `b1270753` (`fix(render-service): publish terminal state after cleanup`) adds unrelated changes to `render-service/src/render-coordinator.ts` and `render-service/test/render-coordinator.test.ts`. This is a focused RAG foundation PR, and the render-service lifecycle change should be moved to its own PR or otherwise removed from this branch. Once that unrelated commit is removed and CI remains green, the RAG portion is ready for final approval.

jackefn:
Thanks, @wyuc! I’ve removed the unrelated `b1270753` Render Service change from this RAG branch. The current CI checks remain green, and the PR is now limited to the focused RAG foundation scope.

wyuc (approved):
Final re-review completed against `b3eb2969`: no actionable findings. The earlier scope, Unicode, HTML boundary, grapheme, version identity, metadata aliasing, transform-status, and replacement-atomicity findings are resolved with regression coverage; the unrelated Render Service change is no longer in the PR; and the current checks are green.

Non-blocking documentation nit: the PR description still mentions chunk-policy version `1.1.2`, while the implementation and tests now use `1.1.3`.

PR #1070, opened 2026-08-06
feat(dsl): promote the interactive and pbl content kinds into the contract (#1061)

wyuc:
First of three PRs for #1061 — the contract side only. Zero changes outside `packages/@openmaic/dsl`; app consumption follows separately (interactive/widget first, pbl after #1060/#1069 land).

## What the contract gains

- **`interactive`**: `InteractiveContent<TWidgetConfig extends WidgetConfigBase = WidgetConfigBase>` with `WidgetType` (closed six-member union + frozen `WIDGET_TYPES` + guard) and `WidgetConfigBase = { type: WidgetType }`. The rich per-widget payload types deliberately stay app-side — no consumer in the repo reads any field of them beyond `type`, so the contract promises exactly what is consumed and apps regain full precision through the generic parameter (the `Scene<TAction, TContent>` pattern this package already prescribes). `url` is optional and documented as the iframe-src fallback used only when `html` (a complete document rendered via srcDoc) is absent.
- **`pbl`**: `PBLContent = { type: 'pbl'; projectV2?: PBLProject; projectConfig?: Record<string, unknown> }`. `PBLProject` is the design-time definition **plus the planner-seeded initial skeleton** — statuses, assignees, thread seats (`{ agentId, messages: [] }`), empty ledgers — because that is the exact shape stored documents carry; skeleton fields are required and documented as canonically seeded. Learner runtime state (assessments, runtime events, pending handovers, message contents) stays app-side. `projectConfig` is an opaque deprecated blob: pre-v2 stored scenes carry it, current code never writes it, and under `additionalProperties: false` the contract must acknowledge it or declare existing documents invalid.
- **Schema**: `SerializedScene` widens to the four-member discriminated union; `scene.schema.json` now covers every kind `SceneType` names. Validated against both a legacy-shaped pbl fixture and a seeded v2 fixture.
- **Companions**: `validateInteractiveContent` / `validatePBLContent` (report-style, wired into `validateScene`, which previously rejected non-slide/quiz kinds), `normalizePBLProject` (idempotent; fills missing seeds with canonical values, fails loud on wrong-typed input), guards `isInteractiveContent` / `isPBLContent` / `isWidgetType`.

## Versioning

[...]

PR #1079, opened 2026-08-09
refactor(pbl): consume the contract's PBL project types (#1061)

### LINKED ISSUE #1061: Promote the interactive and pbl content kinds into @openmaic/dsl
## Why now

`@openmaic/dsl` owns the lesson skeleton and the two universal content kinds (`slide`, `quiz`); `interactive` and `pbl` are valid `SceneType` values whose *content shapes* deliberately stayed app-side, injected through `Scene`'s generic parameters (`packages/@openmaic/dsl/src/stage.ts:19-28`, `schema-roots.ts:17-20`). That boundary was drawn when every producer and consumer of those kinds lived inside the app.

That premise is about to break. #1057 moves the generation pipeline — which *produces* interactive and pbl content — into a published package. A published generator needs a published home for the types it emits, and per #720 that home is the contract package, not the generator. Two more forces point the same way:

- The app's `lib/types/stage.ts` describes both kinds as "kept here rather than in `@openmaic/dsl` because it couples to …" — but the couplings named (`WidgetType`/`WidgetConfig`, the PBL project config) are themselves pure type surfaces with no runtime dependency. Nothing structural keeps them out.
- #1058 retires PBL v1, after which the pbl content shape converges on a single payload (`projectV2`) — the cleanest state it will ever be in to standardize.

## What moves

- **`InteractiveContent`** (`{ type: 'interactive'; url; html?; widgetType?; widgetConfig? }`) plus the `WidgetType` / `WidgetConfig` type surface it references.
- **`PBLContent`** plus the design-time `PBLProjectV2` type family (project info, roles, milestones, microtasks, scenario config, proficiency tiers).
- **Schema coverage**: `schema-roots.ts`'s `SerializedScene` widens from the two-member spelled-out union to four, so `scene.schema.json` emits a discriminated `anyOf` covering every kind the contract names in `SceneType`.
- **Pure companions**: `validate*` / type guards / `normalize*` defaulters for the two new kinds, same discipline as the existing ones (`validate.ts`, `normalize.ts` — report vs. repair, fail loud on malformed input).
- The app's `lib/types/stage.ts` widening shims shrink accordingly; `isInteractiveContent` / `isPBLContent` guards join their siblings in the contract.

## What does not move

- PBL planning/generation code (that is #1057's business), the runtime agents, API routes, UI.
- Runtime learner state and its operations — the contract gains the *design-time* project definition, not session state.
- dsl stays zero-dependency: types, schema, pure functions only.

## Design questions the implementing PR must answer first

1. **Legacy `projectConfig`.** Today `PBLContent.projectConfig` is the required v1 field and `projectV2` optional; after #1058 the polarity flips. Does the contract model the legacy field as an optional deprecated member (documented read-only), or does it stay an app-side widening during the transition? The answer decides whether pre-v2 stored scenes validate against the contract schema as-is.
2. **Design-time vs. runtime-skeleton boundary inside `PBLProjectV2`.** The type as it exists today includes deterministically seeded runtime bootstrap (statuses, thread anchors, proficiency seats). Decide which fields are contract (what a stored project definition *is*) and which are producer artifacts, and document the line — this is the highest-taste decision in the move.
3. **Widget surface.** `WidgetConfig` carries per-widget-type payloads; audit which members are contract-worthy and which are app feature detail that should stay behind an extension point.
4. **Versioning.** Adding content kinds widens the scene schema. Decide whether this is a compatible widening under the existing version model or requires a version bump, and align the storage-side validators in the same change.

## Sequencing

Fork-and-land ahead of #1057's extraction parts: once this is in, the generation package depends on `@openmaic/dsl` for *every* content type it emits and owns none of them — the dependency rule (`generation → dsl` only) gets its full meaning. #1058 landing first simplifies question 1 materially.

## Acceptance

- [ ] `InteractiveContent`, `PBLContent`, `PBLProjectV2` family, and the widget type surface exported from `@openmaic/dsl`
- [ ] `scene.schema.json` covers all four kinds as a discriminated union
- [ ] validators / guards / normalizers for both kinds, with tests, including malformed-input fail-loud cases
- [ ] app consumes the contract types; the widening shims in `lib/types/stage.ts` reduce to re-exports or disappear
- [ ] versioning decision documented in the dsl package (and reflected in storage validators if a bump is taken)
- [ ] no runtime dependency added to dsl

Tracks #720. Unblocks #1057 (published home for emitted types). Interacts with #1058 (legacy pbl field polarity). Companion refactor: #1062 (planner-internal layering; deliberately decides no type homes — that is this issue's job).

wyuc:
Third and final PR for #1061 (after #1070/#1073, sequenced behind #1060/#1069). `lib/pbl/v2/types.ts` becomes a barrel over `@openmaic/dsl` plus an app-side runtime overlay; no importer changes anywhere.

- **Composition**: `PBLProjectV2 = RuntimeOverlay<contract PBLProject, app overlay>` applied recursively (microtasks, milestones, thread seats), so the app's hydrated shape keeps its exact structure while the contract owns the design-time definition. Type assertions pin app → contract assignability and a canonical stored design template satisfies both.
- **`PBLContent`** widens the contract's with the app-hydrated `projectV2`.
- **Write validator untouched by construction**: `lib/document-store/validators.ts` is byte-identical to base (same SHA-256) — no new field enforcement over stored shapes; the contract validator stays strict for external consumers only.
- **Storage**: comment-only correction of the stale "owns the universal slide/quiz kinds" docs; patch bump `0.2.2 → 0.2.3`, no code changes.

Verification (Node 22): dsl 220 / generation 87 / storage 659 package tests · PBL+classroom+document-store 857 · **full app suite 4,173 passing** · root tsc (the 72+175 importer surfaces compile unchanged) · check/lint · version-bump gate.

Closes #1061.

### DIFF
--- a/lib/pbl/legacy/read.ts
+++ b/lib/pbl/legacy/read.ts
@@ -56,6 +56,7 @@ interface LegacyPBLChatMessage {
 }
 
 export interface PBLProjectConfig {
+  [key: string]: unknown;
   projectInfo: LegacyPBLProjectInfo;
   agents: LegacyPBLAgent[];
   issueboard: {
--- a/lib/pbl/v2/types.ts
+++ b/lib/pbl/v2/types.ts
@@ -40,21 +40,7 @@ export type PBLEvaluationKind = 'task' | 'milestone' | 'final';
  *  `record_closing_check` tool before `advance_micro_task` is allowed. */
 export type PBLClosingQuality = 'weak' | 'ok' | 'strong';
 
-// Contract PBL tree nodes deliberately tolerate unknown stored fields via an
-// index signature. Remove only that signature before replacing known nested
-// properties, so the app surface remains as precise as it was before the
-// contract extraction.
-type WithoutIndexSignature<T> = {
-  [K in keyof T as string extends K
-    ? never
-    : number extends K
-      ? never
-      : symbol extends K
-        ? never
-        : K]: T[K];
-};
-
-type RuntimeOverlay<Base, Overlay> = Omit<WithoutIndexSignature<Base>, keyof Overlay> & Overlay;
+type RuntimeOverlay<Base, Overlay> = Omit<Base, keyof Overlay> & Overlay;
 
 /** Set by Instructor when advancing a task via `advance_micro_task`.
  *  Never shown to the learner — internal teaching record. */
--- a/lib/pbl/v2/types.ts
+++ b/lib/pbl/v2/types.ts
@@ -1,93 +1,60 @@
 /**
- * PBL v2 — Project Schema
+ * PBL v2 — contract design types plus the app runtime overlay.
  *
- * Replaces v1's role-selection + read-only issueboard + @question/@judge
- * model with a single-Instructor guided flow (Hero → Workspace → Completion).
- *
- * Legacy v1 project data is supported only through the read-only adapter in
- * `../legacy/read.ts`; new generation and persistence use this v2 model.
- *
- * The product ships a single Instructor. The multi-agent chat-thread
- * structure is kept generic so additional roles can be introduced later
- * with their own design, but no other role type exists today.
+ * The persisted design/skeleton belongs to `@openmaic/dsl`. This barrel keeps
+ * the established app-facing names while adding learner/runtime state to the
+ * nested contract shapes that carry it.
  */
+import type {
+  PBLMicrotask as ContractPBLMicrotask,
+  PBLMilestone as ContractPBLMilestone,
+  PBLProject as ContractPBLProject,
+  PBLProficiency,
+  PBLRoleType,
+  PBLThreadSeat as ContractPBLThreadSeat,
+} from '@openmaic/dsl';
 import type { SceneOutline } from '@/lib/types/generation';
 
-// ---------------------------------------------------------------------------
-// Enums (string-literal unions; no runtime enum cost)
-// ---------------------------------------------------------------------------
-
-/** Project lifecycle status. */
-export type PBLProjectStatus = 'designing' | 'review' | 'active' | 'completed' | 'archived';
-
-/** Milestone lifecycle. The LOCKED→ACTIVE transition is gated by the
- *  learner clicking "Continue" on a milestone handover card (see
- *  `PBLHandover`). */
-export type PBLMilestoneStatus = 'locked' | 'active' | 'completed';
-
-export type PBLMicrotaskStatus = 'todo' | 'in_progress' | 'completed' | 'skipped';
-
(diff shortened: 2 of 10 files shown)


xuyuanwei678 (changes requested):
**P1: `RuntimeOverlay` drops all declared contract fields**

`WithoutIndexSignature<T>` does not preserve the explicit properties of contract types that extend `Record<string, unknown>`. For those types, `keyof T` is widened to `string | number`, so this mapped type filters out every key. As a result, `PBLProjectV2`, `PBLMilestone`, and `PBLMicrotask` retain only their runtime overlay fields and lose their contract fields from the app-facing type.

For example, `PBLProjectV2` no longer requires or exposes fields such as `title`, `status`, `roles`, or `createdAt`; a microtask similarly loses `id`, `title`, `status`, etc. The new assignability tests do not catch the inverse case where these required fields are omitted.

Please avoid removing the index signature this way. Keeping the base contract type and overriding the nested runtime fields directly, or separating the permissive persisted-input type from the precise declared contract type, would preserve the existing app-side type guarantees. Add a negative type test showing that a project without `title` (and a microtask without `id`) fails to type-check.

wyuc:
@xuyuanwei678 thank you — the analysis was exactly right, and the root cause sat one level deeper: the dsl contract interfaces carried `extends Record<string, unknown>` (added for historical-extras schema tolerance), which is what forced the key-erasing `WithoutIndexSignature` here. Fixed at the root:

- **dsl**: the four PBL interfaces are precise TypeScript again (no index signature); the schema's tolerance for historical runtime-overlay extras is now applied in the generation script instead (`@additionalProperties true` JSDoc turned out to be ignored by the generator — verified empirically). Required arrays and enum checks unchanged; dsl bumped to 0.8.0 since the narrowing is compile-visible.
- **app**: `WithoutIndexSignature` deleted; the overlay is plain `Omit<Base, keyof Overlay> & Overlay`, which is correct now that `keyof` is the literal union.
- **tests**: your requested negatives are in — a project missing `title` and a microtask missing `id` fail to type-check — plus `keyof PBLProjectV2` sentinels on contract fields (title/status/roles/createdAt) and overlay fields, so key erasure cannot silently return.

Full app suite 4,174 passing; schema fixtures (legacy + seeded + overlay-extras) all green.

xuyuanwei678 (approved):
Re-reviewed the follow-up commit. The index signatures are now removed at the DSL contract boundary, so `Omit<Base, keyof Overlay> & Overlay` preserves the declared contract fields. Schema compatibility for historical runtime overlays is restored only for the four affected PBL definitions, and the new negative/keyof sentinel tests cover the original regression. Approved.

PR #1083, opened 2026-08-09
fix(asr): drop region subtag before calling the transcription API

### LINKED ISSUE #1082: openai-whisper ASR fails with "unsupported language" when the language setting carries a region subtag
### Summary

Speech-to-text through `openai-whisper` fails with `unsupported language: pt-BR` whenever the ASR language setting carries a region subtag. The transcription API takes ISO-639-1, but the app forwards the setting verbatim.

### Root cause

`asrLanguage` is a single global setting in the store, shared across ASR providers. The provider catalogues disagree on its shape:

- `openai-whisper`, `qwen-asr`, `qwen3-asr-flash`, `azure-asr` → `'pt'`
- `browser-native` → `'pt-BR'`, `'pt-PT'`

So a user who had browser-native selected and picked `pt-BR` keeps that value when switching to (or being served by) a server-configured `openai-whisper`, and `lib/audio/asr-providers.ts` passes it straight through:

```ts
providerOptions: {
  openai: {
    language: config.language === 'auto' ? undefined : config.language,
  },
},
```

The upstream API rejects the whole request:

```
[ERROR] [Transcription] Transcription failed [provider=openai-whisper, model=whisper-large-v3-turbo]: AI_APICallError: unsupported language: pt-BR
```

The same applies to any region-tagged value (`zh-CN`, `pt-PT`, ...), and to any OpenAI-compatible transcription endpoint.

### Reproduction

1. Configure `ASR_OPENAI_API_KEY` (any OpenAI-compatible transcription endpoint).
2. Set the ASR language to a region-tagged value — e.g. select `browser-native` and pick `pt-BR`, then switch to `openai-whisper`.
3. Record anything. The request fails; the UI simply produces no transcription.

### Proposed fix

Reduce the value to its ISO-639-1 base subtag at the provider boundary, keeping `auto` as "detect". Happy to send a PR.

thanos-openclaw:
## Summary

`openai-whisper` transcription fails outright when the ASR language setting carries a region subtag (`pt-BR`, `pt-PT`, `zh-CN`, ...). This reduces the value to the ISO-639-1 code the API expects.

## Related Issues

Closes #1082

## Changes

- `lib/audio/asr-providers.ts`: new `toISO6391()` helper; `transcribeOpenAIWhisper` sends the base subtag instead of the raw setting. `auto` still maps to `undefined` (detect).

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Configure `ASR_OPENAI_API_KEY` / `ASR_OPENAI_BASE_URL` against any OpenAI-compatible transcription endpoint.
2. `POST /api/transcription` with `language=pt-BR` and an audio file.

Before: `500`, and the server logs `AI_APICallError: unsupported language: pt-BR`. After: `200` with the transcription.

On a live instance backed by Groq's OpenAI-compatible endpoint (`whisper-large-v3-turbo`), same audio file in all three cases:

- `language=pt` — worked before and after (this is why the bug is invisible when testing the API directly with a bare code).
- `language=pt-BR` — **before:** upstream 400 `unsupported language: pt-BR`; **after:** `{"success":true,"text":" A fotossintese transforma a luz solar em energia química."}`.
- `language=pt-PT` — **after:** same successful transcription.

I also confirmed the source of the mismatch: `browser-native` is the only ASR provider whose `supportedLanguages` are region-tagged (`pt-BR`, `pt-PT`), while `openai-whisper`, `qwen-asr`, `qwen3-asr-flash` and `azure-asr` all list bare `pt` — and `asrLanguage` is a single global setting shared across them.

What I did **not** verify: the Azure and Qwen ASR paths (untouched by this change, and they build their own request), and whether any OpenAI-compatible provider actually wants a region-tagged code (I found none documented; the API reference specifies ISO-639-1).

### Evidence

```
# before
[ERROR] [Transcription] Transcription failed [provider=openai-whisper, model=whisper-large-v3-turbo]: AI_APICallError: unsupported language: pt-BR

# after
$ curl -F 'audio=@fala.mp3' -F 'providerId=openai-whisper' -F 'modelId=whisper-large-v3-turbo' -F 'language=pt-BR' .../api/transcription
{"success":true,"text":" A fotossintese transforma a luz solar em energia química."}
HTTP=200
```

### DIFF
--- a/lib/audio/asr-providers.ts
+++ b/lib/audio/asr-providers.ts
@@ -304,6 +304,16 @@ function getOptionalBearerAuthHeaders(apiKey?: string): Record<string, string> {
 /**
  * OpenAI Whisper implementation (using Vercel AI SDK)
  */
+/**
+ * Reduce a language setting to the ISO-639-1 code the OpenAI-compatible
+ * transcription API expects. `auto` (and anything empty) means "detect".
+ */
+function toISO6391(language: string | undefined): string | undefined {
+  if (!language || language === 'auto') return undefined;
+  const base = language.split('-')[0]?.trim().toLowerCase();
+  return base || undefined;
+}
+
 async function transcribeOpenAIWhisper(
   config: ASRModelConfig,
   audioBuffer: Buffer | Blob,
@@ -330,7 +340,12 @@ async function transcribeOpenAIWhisper(
       audio: audioData,
       providerOptions: {
         openai: {
-          language: config.language === 'auto' ? undefined : config.language,
+          // The transcription API takes ISO-639-1, but `asrLanguage` is a single
+          // global setting shared with browser-native ASR, whose catalogue is
+          // region-tagged (pt-BR, pt-PT, zh-CN...). Carrying such a value over
+          // to this provider fails the whole request with
+          // "unsupported language: pt-BR", so drop the region subtag.
+          language: toISO6391(config.language),
         },
       },
     });


wyuc (changes requested):
Thanks for tracing the failure and providing a concrete reproduction. However, I do not think we should add this normalization at the OpenAI provider boundary.

The normal provider-switch path already resets a language that is unsupported by the target provider. The remaining failure appears to come from provider-selection paths that bypass that invariant. That is a narrow state-consistency problem, and it should be fixed where the provider is selected rather than by teaching every provider boundary how to reinterpret stale settings.

`toISO6391` is also a misleading and incomplete abstraction: taking the first BCP-47 subtag does not necessarily produce the code accepted by OpenAI (`yue-Hant-HK` and `fil-PH` are existing counterexamples), and this path is shared by custom OpenAI-compatible providers whose language requirements are not guaranteed to match OpenAI. Making the conversion robust would require provider-specific mappings, validation, and tests, which is disproportionate to the limited failure path.

Given that maintenance cost and the small practical benefit, I do not think this design has a good ROI or should be merged as-is. Please address the provider/language state invariant instead, if the remaining automatic-selection path is worth fixing.

wyuc:
Thanks @thanos-openclaw for this contribution. I'm closing it to keep the queue tidy, since there hasn't been an update since the changes requested on 2026-08-12. It isn't a rejection of the idea: if you'd like to pick it back up, you're welcome to reopen it or open a fresh PR against the current `main` that addresses the review, and we'll take another look.

PR #1085, opened 2026-08-09
fix(agents): bind generated agents to a voice of the matching gender

### LINKED ISSUE #1084: Generated agents get voices of the wrong gender (voice gender never reaches the model; fallback is index-based)
### Summary

Generated agents regularly get a voice of the wrong gender: a female-named teacher speaks with a male voice. Three separate gaps stack up, and none of them is configuration.

### 1. The voice catalogue's gender is dropped on the way to the model

`TTSVoiceInfo` carries `gender`, but the payload built for the profile generator does not:

```ts
// app/generation-preview/page.tsx
p.voices.map((v) => ({
  providerId: p.providerId,
  voiceId: v.id,
  voiceName: v.name,
  voiceLanguage: v.language,
}))
```

So the model has to infer gender from the id and the display name — and MiniMax names its voices in Chinese (`御姐音色`, `精英青年`, `温润男声`), while the course is generated in another language entirely. The prompt asks it to "pick a voice that suits the agent's personality and role", never to match gender.

### 2. The vocal identity the model writes is discarded for most providers

The prompt asks for `voiceDesign.identity` as "gender + age + role". That field is then only ever read by VoxCPM:

```ts
// lib/audio/agent-voice.ts
export async function resolveAgentVoiceOptions(...) {
  if (opts.providerId !== VOXCPM_TTS_PROVIDER_ID) return undefined;
```

With MiniMax, Qwen, OpenAI or Azure TTS, everything the model designed about the voice is thrown away at the boundary.

### 3. The fallback is gender-blind

When the model returns no `voice`, the pick is pure index round-robin:

```ts
// lib/audio/voice-resolver.ts
voiceId: first.voices[agentIndex % first.voices.length].id
```

Position in the catalogue decides. For MiniMax's list, agent 0 lands on a male voice regardless of who agent 0 is.

### Proposed fix

Pass the registry's `gender` through to the generator and require it to match; have the generator emit an explicit, language-independent `gender` on each agent; and use it to narrow the fallback pool before indexing. Happy to send a PR.

Note on approach: parsing `voiceDesign.identity` for "male"/"female" is not sufficient on its own — that prose follows the course language directive ("professora de meia-idade" for a pt-PT course), so an English matcher silently does nothing for exactly the deployments where the display names are least readable.

thanos-openclaw:
## Summary

Generated agents were regularly given a voice of the wrong gender — a female-named teacher speaking with a male voice. This passes the registry's voice `gender` through to the profile generator, has the generator emit an explicit per-agent `gender`, and uses it to narrow the deterministic fallback pool.

## Related Issues

Closes #1084

## Changes

- `app/generation-preview/page.tsx`: include `voiceGender` in the voice list sent to `/api/generate/agent-profiles` (the registry had it; the payload dropped it).
- `app/api/generate/agent-profiles/route.ts`: expose each voice's `gender` to the model and require it to match the agent; ask for an explicit `gender` per agent in literal English values (`male`/`female`/`neutral`) so it survives the language directive; validate and pass it through.
- `lib/orchestration/registry/types.ts`: `gender` on the agent config.
- `lib/audio/voice-resolver.ts`: `inferAgentGender()`; the fallback now filters to matching (and `neutral`) voices before indexing, keeping the existing rotation.
- `tests/audio/voice-resolver-gender.test.ts`: new.

Existing behaviour is preserved where there is no gender information: an agent without one, or a catalogue whose voices carry none, still gets the plain index pick, and a pool with no match falls back to the full list.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Configure MiniMax TTS (its voice display names are Chinese) and generate a classroom in a non-Chinese language.
2. Inspect the generated agents' assigned voices against their names.

**End to end, on a live instance** (MiniMax `MiniMax-M3` + MiniMax TTS, course in European Portuguese). Called `/api/generate/agent-profiles` with five MiniMax voices whose display names are Chinese and whose genders differ, and checked every returned agent:

```
OK  Professora Lúcia   gender=female   voice=Chinese (Mandarin)_Warm_Girl   (voice is female)
OK  Tiago              gender=male     voice=male-qn-jingying               (voice is male)
OK  Sofia              gender=female   voice=female-shaonv                  (voice is female)
OK  Mateus             gender=male     voice=Chinese (Mandarin)_Gentleman   (voice is male)
```

4/4 consistent, with Portuguese names and Chinese voice names — the case that used to fail.

**Mutation-checked the fallback.** The new fixture is ordered so plain index round-robin returns the wrong gender. Removing the gender filter turns exactly the three tests that assert the new behaviour red, and leaves the three that pin the unchanged behaviour green:

```
× gives a female agent at index 0 a female voice, not the first voice in the list
× gives a male agent at index 1 a male voice, not the second voice in the list
× still rotates across the matching voices as the index advances
✓ keeps the plain index behaviour when the agent has no gender information
✓ falls back to the full list when no voice matches the wanted gender
✓ leaves a gender-less catalogue on the plain index behaviour
Tests  3 failed | 8 passed (11)
```

**Regression.** `tests/audio/` — 26 files, 160 tests, all passing. Full `tests/` suite on this branch matches the baseline on a clean `main` checkout: the only failures are the 16 pre-existing ones in `tests/ci/publish-openmaic-skill.test.ts`, which fail identically without this change. `prettier --check` and `eslint` clean; `tsc --noEmit` reports 0 errors outside `packages/@openmaic/*/test/**` (pre-existing, environment-related).

What I did **not** verify: providers other than MiniMax TTS (the change is provider-agnostic and reads only the registry's `gender`, but I only exercised MiniMax against a live API), and I did not touch the separate issue that `voiceDesign` is ignored for every non-VoxCPM provider — see point 2 of #1084, which this PR deliberately leaves open.

### Evidence

See the agent table and the mutation output above.

### DIFF
--- a/app/api/generate/agent-profiles/route.ts
+++ b/app/api/generate/agent-profiles/route.ts
@@ -29,6 +29,7 @@ interface RequestBody {
     voiceId: string;
     voiceName: string;
     voiceLanguage?: string;
+    voiceGender?: 'male' | 'female' | 'neutral';
   }>;
 }
 
@@ -96,6 +97,7 @@ export async function POST(req: NextRequest) {
               id: `${v.providerId}::${v.voiceId}`,
               name: v.voiceName,
               language: v.voiceLanguage || 'unknown',
+              gender: v.voiceGender || 'unknown',
             })),
           )
         : '';
@@ -103,7 +105,13 @@ export async function POST(req: NextRequest) {
     const voicePrompt = voiceListStr
       ? `- Each agent should be assigned a voice that matches their persona from this list: ${voiceListStr}
   - Prefer a voice whose language matches the course language directive
-  - Pick a voice that suits the agent's personality and role (e.g. authoritative voice for teacher, lively voice for energetic student)
+  - The voice's "gender" MUST match the gender implied by the agent's name and by
+    the "identity" of its voiceDesign. A female-named agent must not get a voice
+    whose gender is "male", and vice versa. Voice display names may be in a
+    language you are not generating in — trust the "gender" field, not the name.
+  - Within the voices of the right gender, pick one that suits the agent's
+    personality and role (e.g. authoritative voice for teacher, lively voice for
+    energetic student)
   - Try to use different voices for each agent`
       : '';
 
@@ -129,6 +137,10 @@ Requirements:
   - Use the "path" value as the avatar field in the output
 - Each agent must be assigned one color from this list: ${JSON.stringify(AGENT_COLOR_PALETTE)}
   - Each agent must have a different color
+- Each agent needs a "gender" field — always one of the literal English values
+  "male", "female" or "neutral" — consistent with the agent's name. This is a
+  machine field: it stays in English even when names and personas follow the
+  language directive, and it is what binds the agent to a matching voice.
 - Each agent needs a "voiceDesign" object describing their VOCAL identity (not personality), written following the language directive and consistent with the persona, as three short comma-free phrases:
   - "identity": gender + age + role (e.g. "middle-aged male teacher")
   - "texture": pitch + vocal quality (e.g. "warm low-pitched slightly husky")
@@ -142,6 +154,7 @@ Return a JSON object with this exact structure:
       "name": "string",
       "role": "teacher" | "assistant" | "student",
       "persona": "string (2-3 sentences)",
+      "gender": "male" | "female" | "neutral",
       "voiceDesign": { "identity": "string", "texture": "string", "delivery": "string" },
       "avatar": "string (from available list)",
       "color": "string (hex color from palette)",
@@ -175,6 +188,7 @@ Return a JSON object with this exact structure:
         avatar: string;
         color: string;
         priority: number;
+        gender?: string;
         voice?: string;
         voiceDesign?: unknown;
       }>;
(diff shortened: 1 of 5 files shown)


wyuc (changes requested):
Thanks for addressing the gender-blind voice assignment. The direction is right, but the current implementation does not yet make the generated agent's voice the single source of truth.

1. **The teacher bypasses the agent binding during discussion.** In `use-discussion-tts.ts`, a teacher always returns `globalTtsVoice` before `resolveAgentVoice()` is called. A generated female teacher can therefore still speak with a male global voice even when this PR stores a matching `voiceConfig` on the agent. The Agent Bar resolves through the agent binding, so the displayed and actually played voices can also diverge. Please define one source of truth and cover the production teacher discussion path with a test. My recommendation is to bind the teacher to the global lecturer voice at generation time, then have subsequent playback consistently consume the agent binding.

2. **A model-selected mismatched voice is never corrected.** The route accepts any returned `providerId::voiceId` and stores it as `voiceConfig`. Since `resolveAgentVoice()` prioritizes that binding, the new gender-aware fallback only helps when the model omits or returns an unavailable voice. Please validate the selected voice against the agent gender and catalogue metadata deterministically, rather than relying only on prompt compliance.

3. **The persisted DSL contract is missing the new field.** These generated objects are stored in `stage.generatedAgentConfigs`, but `@openmaic/dsl`'s `GeneratedAgentConfig` and its generated strict schema do not declare `gender`. Schema-validating consumers will reject documents containing it. Please add the field to the DSL contract/schema coverage and add a persistence/schema test.

There is also a separate server-side `agentMode=generate` path in `lib/server/classroom-generation.ts` that does not produce gender metadata, so please either bring it into the same invariant or explicitly document and test the intended boundary.

wyuc:
Thanks @thanos-openclaw for this contribution. I'm closing it to keep the queue tidy, since there hasn't been an update since the changes requested on 2026-08-12. It isn't a rejection of the idea: if you'd like to pick it back up, you're welcome to reopen it or open a fresh PR against the current `main` that addresses the review, and we'll take another look.

Issue #987, opened 2026-07-24
[Research] Video export: interactive iframe action replay and deterministic capture


cosarah:
## Context

Research track from #867 item L3. This follows the static non-slide card work in #985 and the Quiz visual-segment work in #986.

Interactive scenes currently run generated HTML inside a sandboxed iframe. Classroom playback controls the iframe through postMessage widget actions such as SET_WIDGET_STATE, HIGHLIGHT_ELEMENT, ANNOTATE_ELEMENT, and REVEAL_ELEMENT. The video exporter currently represents interactive scenes with a placeholder and does not attempt to capture runtime state.

## Goal

Build a constrained proof of concept for deterministic interactive-scene replay inside the Hyperframes/Chromium export path, and document whether it is suitable for a later production implementation.

## Research questions

1. Can an exported composition load a self-contained interactive srcDoc or packaged HTML in an iframe and wait for a reliable ready signal?
2. Can the existing widget action protocol be replayed against that iframe at VideoTimeline timestamps?
3. Can Chromium capture the resulting iframe state reliably in the existing render service?
4. What breaks determinism: requestAnimationFrame, setTimeout, Date.now, Math.random, external resources, Canvas/WebGL, audio/video, or page-specific script behavior?
5. Can the iframe remain sandboxed while still supporting the replay bridge?
6. What diagnostics can identify an iframe that failed to load or did not respond to an action?

## Scope

1. Create one or two controlled interactive fixtures that use local HTML/CSS/JS only.
2. Reuse the current widget action vocabulary and postMessage payloads where possible.
3. Add a minimal exporter-side replay bridge:
   - iframe load/ready handshake;
   - timestamped action dispatch;
   - bounded readiness and action timeouts;
   - failure diagnostics.
4. Capture representative key frames or a short MP4 through the existing Hyperframes/Chromium path.
5. Compare repeated renders for deterministic output.
6. Produce an experiment report with:
   - architecture recommendation;
   - supported and unsupported page behaviors;
   - required packaging/network constraints;
   - timing and synchronization findings;
   - go/no-go recommendation;
   - follow-up implementation issues if the POC succeeds.

## Non-goals

- General browser screen recording or learner input capture.
- Production support for arbitrary imported/generated interactive pages.
- Guaranteeing support for WebGL, external CDN dependencies, cross-origin embeds, or nondeterministic scripts.
- Expanding #866 render-service deployment or quota behavior.
- Replacing the existing classroom iframe host.

[...]

PR #1086, opened 2026-08-10
feat(video-export): capture self-contained static interactive HTML

### LINKED ISSUE #1052: [Task] Video export: self-contained static interactive HTML capture
### Problem or Motivation

Interactive scenes are currently exported as generic placeholders. This blocks the first useful video-export release from preserving the authored GenUI surface, even when the page can be packaged and rendered safely without replaying learner or teacher interactions.

This is the highest-priority child of #987 because a deterministic static HTML representation is independently valuable and is also the required fallback for later replay work.

### Proposed Solution

Integrate self-contained interactive HTML into the existing VideoTimeline to Hyperframes to render-service path.

- Add an interactive static-HTML visual representation to the export IR and asset plan.
- Package embedded HTML and inline supported external assets before rendering.
- Render the page in a sandboxed iframe without allow-same-origin.
- Wait for a bounded ready/settle condition, then freeze the page at a defined static state for the scene duration.
- Fall back to the existing cover/placeholder representation when HTML is missing or cannot load.
- Surface packaging, readiness, runtime, and external-resource failures as explicit diagnostics.
- Keep action replay out of scope for this increment.

### Acceptance Criteria

- [ ] A local self-contained interactive fixture renders through the current Hyperframes/Chromium render service instead of the generic placeholder.
- [ ] The exported ZIP remains self-contained and works with render-service network egress disabled.
- [ ] The iframe remains sandboxed without allow-same-origin.
- [ ] The captured page is visually static for the whole scene duration.
- [ ] Missing HTML, load timeout, runtime failure, and unresolved external resources produce explicit diagnostics.
- [ ] A failed static HTML render falls back to the current cover/placeholder path without failing the whole video export.
- [ ] Existing slide, Quiz, and PBL export behavior remains unchanged.
- [ ] Unit, emitter, packaging, and one isolated Chromium smoke test cover the new path.

### Alternatives Considered

- Keep generic placeholders only: rejected because it loses the authored GenUI surface in the first video-export release.
- Pre-render every page to PNG in a separate Chromium pass: stronger isolation, but it requires a two-pass export architecture and is unnecessary for the first static increment.
- Implement replay first: rejected because static HTML is both independently useful and the required replay fallback.

### Area

Import / export

### Additional Context

Parent: #987

This issue is on the first-wave video-export launch path and should land before interactive action replay.

cosarah:
## Summary

Add deterministic, self-contained static HTML capture for interactive/GenUI scenes in the existing video-export pipeline. Interactive pages are packaged into the export ZIP, rendered in a sandboxed iframe, frozen after a bounded settle phase, and fall back to the existing cover when static capture is unavailable.

## Related Issues

Closes #1052
Related to #987

## Changes

- Add an interactive static-HTML visual and asset representation to the video-export IR and compiler passes.
- Prepare interactive documents for offline rendering by inlining supported assets, including `srcset` candidates, sanitizing invalid CSS URLs, applying CSP/runtime guards, and emitting diagnostics for unresolved resources or preparation failures.
- Emit sandboxed Hyperframes iframe scenes without `allow-same-origin`, with bounded readiness/settling and deterministic freeze behavior; static capture disables Worker execution.
- Preserve the existing cover fallback when HTML is absent or cannot be prepared.
- Package prepared HTML in export ZIPs and use probed audio durations when compiling the application timeline.
- Keep runtime diagnostics typed and available through the live DOM report and in-memory manifest bridge; do not ship an immutable sidecar that cannot be updated after packaging.
- Add unit, emitter, packaging, security, materialization, and opt-in Chromium smoke coverage; extend CI to lint and smoke-test the new sample.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes a bug)
- [ ] Documentation update
- [x] CI/CD or build changes

### Steps to reproduce / test

1. Run `vitest run tests/video-export tests/export/inline-assets.test.ts tests/security/iframe-sandbox.test.ts tests/workflows/ci-video-export-contract.test.ts`.
2. Run ESLint and Prettier checks over the changed source and test files.
3. Export a single GenUI page from a real `.maic` classroom archive and render the generated Hyperframes project to MP4.

### What was verified

- Targeted regression suite: 23 test files passed, 1 skipped; 381 tests passed, 41 skipped.
- ESLint completed with 0 errors; changed files pass Prettier.
- Isolated Chromium smoke: 2 tests passed with `INTERACTIVE_STATIC_BROWSER=1`; the smoke covers deterministic freeze, Worker blocking, and runtime fallback diagnostics.
- `srcset` packaging tests cover multi-density candidates and unresolved candidates.
- Full repository TypeScript checking was run; it still reports pre-existing main/dependency errors outside this change.

Manual render artifacts are intentionally kept out of git.

## AI Assistance

This PR is AI-assisted. The implementation was reviewed by independent Spec and Standards subagents, and the reported findings were addressed in this branch.

## Checklist

- [x] I have performed a self-review of the code


wyuc (changes requested):
Requesting changes for four packaging correctness issues found against the current head fe66a7b3:

- [P2] `prepare-interactive-html.ts:276`: HTML with a doctype but no explicit `<head>` receives injected markup before the doctype. Chromium then enters quirks mode, which can change the captured layout. Please create a head or insert after the doctype.
- [P2] `inline-assets.ts:114-117`: external URL-bearing elements such as `iframe[src]`, `object[data]`, and `embed[src]` are not reported as unresolved. CSP blocks them, but readiness can still report a successful frozen frame with missing content. Please reject or explicitly package these resources.
- [P2] `inline-assets-importmap.ts:12-18`: the regex-based module scan treats import-like text in comments or strings as real imports. This can mutate valid source or reject a page because of a nonexistent dependency. Please use a JavaScript lexer/parser for module specifiers.
- [P2] `inline-assets.ts:298-302`: the same stylesheet imported under different media/layer/supports conditions is globally deduplicated, so later valid conditional imports disappear. Cycle detection should track the active recursion stack while allowing repeated imports under distinct conditions.

The existing Chromium smoke and a real generated GenUI sample both render successfully, but neither covers these cases.

cosarah:
Thanks for the detailed review. I addressed all four findings in d97bca0a:

- Preserve standards mode for documents with a doctype but no explicit head element.
- Detect iframe src, object data, and embed src as unsupported residual resources instead of letting them bypass offline-packaging diagnostics.
- Replace regex-based module scanning with es-module-lexer so imports inside comments and strings are ignored while original specifier text is preserved when unchanged.
- Limit CSS import deduplication to the active recursion path, preserving repeated imports that use different media, layer, or supports conditions while still breaking cycles.

Validation completed:
- 68/68 focused tests passed.
- ESLint, Prettier, and git diff --check passed.
- All 5 GitHub checks are green, including E2E and Lint/Typecheck/Unit Tests.

wyuc (changes requested):
Thanks for addressing the previous four findings. Those cases now look fixed, but two remaining packaging correctness issues still need attention:

- [P2] `lib/video-export-app/prepare-interactive-html.ts:29-37`: head detection scans the raw HTML with a regular expression, so a literal `<head>` inside a script string, template, or comment is treated as the document head. For example, `const template = "<head>"` receives the CSP and capture injection inside the authored script, corrupting otherwise valid interactive HTML and causing export fallback or partial output. Please locate/create the actual document head using an HTML-aware parser or equivalent structure-aware logic.

- [P2] `lib/export/inline-assets.ts:122-125,157-164`: the new iframe/object/embed checks still discard `data:` and `blob:` references through the shared `push` filter. The package is therefore marked present for `<iframe src="data:text/html,...">` or `<object data="data:image/svg+xml,...">`, while the emitted CSP blocks them with `frame-src none` / `object-src none`. Chromium then produces a frozen-success frame with missing content. Please reject these schemes specifically for iframe/object/embed, or package and allow them consistently without rejecting valid data-URI images.

I reproduced both cases locally. The focused 68 tests pass, the three Chromium assertions pass, and all five GitHub checks are green; these edge cases are not covered by the current suite.

cosarah:
Thanks for the follow-up review. I addressed both remaining packaging issues in cca0d99d:

- Head injection now uses parse5 source locations instead of raw HTML regular expressions. Both the iframe shims and static-capture runtime are inserted into the parsed document head without rewriting authored HTML, so head-like text inside scripts, templates, or comments is preserved.
- data: and blob: references on iframe[src], object[data], and embed[src] now enter the existing unresolved-resource rejection path, while valid inline image data URIs remain supported.

I also merged the latest main branch at 26d280c9; the resulting merge commit is 5402091d and has been pushed.

Post-merge validation:
- 77/77 focused unit tests passed.
- 3/3 Chromium interactive-static tests passed, including standards mode and execution of an authored script containing "<head>".

wyuc (changes requested):
Thanks for fixing the two previous findings. The parse5-based head insertion preserves authored head-like text, and real data/blob iframe/object/embed references now enter the rejection path. The focused 77 unit tests and 3 Chromium tests pass locally, and all five GitHub checks are green.

A fresh full review found the following remaining correctness issues:

- [P1] `lib/video-export-app/prepare-interactive-html.ts:177-178`: import-map-resolved bare specifiers are still added to residual resources. After an entry such as `dep` is successfully converted to a data-URI import-map mapping, `collectAssetRefs(..., { includeRelative: true })` still returns the bare `dep` imported by the module. Valid pages using `import { x } from "dep"` are therefore always rejected as `unresolved-resource`. Please resolve module specifiers against the final import map before classifying residuals.

- [P2, fix-induced] `lib/export/inline-assets.ts:125-128,164-171`: data/blob embedded-document detection still scans raw HTML with element-shaped regular expressions. Inert authored text such as a JavaScript template containing `<iframe src="data:text/html,..."></iframe>` is classified as a live iframe and rejects the entire scene. This is now reachable because the new branch retains data/blob references. Please collect these attributes from actual parsed elements rather than raw source text.

- [P2] `lib/export/inline-assets.ts:353-378`: valid local CSS fragment references such as `filter: url(#blur)` are resolved to `about:blank#blur`, fetched, and reported as failures. Self-contained SVG filters, masks, and gradients consequently degrade to placeholders. Fragment-only URLs should remain local and must not enter asset fetching.

- [P2] `lib/video-export-app/prepare-interactive-html.ts:168-170`: the readiness failure branch posts a failure without freezing or unloading the child document. The parent only hides the iframe, so authored intervals, animation frames, and media continue running during the render and can consume enough CPU to slow or stall the export. Freeze or unload the child before completing fallback.

- [P2] `lib/export/inline-assets.ts:146-150`: SVG `<image href>` and external `<use href>` resources are neither collected nor rewritten. Preparation marks the page present, then CSP blocks the resource and the runtime falls back despite the asset being bundleable. Please include SVG href/xlink:href resource forms in the structured inventory.

Suggested direction

The recurring false positives and omissions come from using regular expressions as structural HTML/CSS scanners. Since this branch already adds parse5, consider making one structured document analysis the source of truth:

1. Parse HTML with parse5 and traverse real element nodes, including HTML and SVG namespaces. Use `sourceCodeLocation` / attribute ranges to patch the original string in descending offset order, preserving authored bytes without serializing the document.
2. Continue using es-module-lexer for module specifiers, but resolve bare names against the final parsed import map before residual validation.
3. Parse style blocks and style attributes with PostCSS plus postcss-value-parser, or css-tree, so `url()`, `@import`, comments, escapes, and local fragments have correct semantics.
4. Continue using parse-srcset for responsive image candidates.
5. Reuse the same structured resource inventory for inlining and residual validation instead of rescanning the rewritten HTML with separate regular expressions.

This should close the current findings together and prevent the scanner and rewriter from drifting again.

cosarah:
Thanks for the thorough follow-up review. I addressed all five findings in bbdbbb4b.

- Replaced the remaining raw HTML resource scans with one parse5-based inventory of actual HTML/SVG elements. Rewrites use descending source ranges, so authored HTML is preserved without parse5 serialization and inert element-like text inside scripts/comments is ignored.
- Moved strict residual reporting into `inlineHtmlAssets()` and resolve module specifiers against the final offline import map before classifying them as unresolved.
- Replaced CSS URL/import scanning with PostCSS plus postcss-value-parser. Local fragment references such as `url(#blur)` remain local, while real declarations, style attributes, and conditional imports are handled structurally.
- Freeze the child document on readiness failure and remove every failed iframe from the parent, so timers, RAF callbacks, media, and animations cannot continue behind the fallback.
- Added SVG `<image href>` and `<use href>` / `xlink:href` packaging, fetching external SVG resources without the fragment and preserving the fragment on the resulting data URI.

Validation completed:

- ESLint, Prettier, `tsc --noEmit`, and `git diff --check` passed.
- The full relevant Vitest suite passed: 411 passed, 43 skipped.
- Chromium interactive-static tests passed: 4/4, including child-frame unloading on runtime and readiness failures.
- A real GenUI-heavy classroom fixture was packaged as a single interactive scene and rendered to a 2-second H.264 MP4 at 1280x720, 30 fps; the generated UI rendered correctly rather than falling back or appearing blank.
- All five GitHub checks are green.

lib/export/inline-assets.ts
```diff
@@ -147,6 +196,52 @@ function guessMime(url: string): string {
...
+        // Leave unresolvable values untouched for strict residual validation.
+      }
+    }
+  });
+
+  const replacements = new Map<string, string>();
+  await mapWithConcurrency([...refsByUrl.entries()], 8, async ([abs, raws]) => {
+    const got = await fetchAsset(abs);
+    if (got) {
+      for (const raw of raws) replacements.set(raw, toDataUri(got.bytes, got.contentType));
```
wyuc: [P2] Preserve SVG fragments when converting CSS URLs to data URIs

For a real CSS reference such as `filter:url(https://cdn.test/filters.svg#blur)`, `abs` includes `#blur`, but the replacement stores only `toDataUri(...)`. The packaged CSS therefore points at the whole SVG instead of the `blur` element, and filters, masks, or paint servers silently stop working even though residual validation passes. Split the fragment before fetching and append it to the generated data URI, as the SVG `href` path already does.

lib/export/html-asset-inventory.ts
```diff
@@ -0,0 +1,285 @@
...
+    if (element.tagName === 'object') addAttribute('object-data', 'data');
+    if (element.tagName === 'embed') addAttribute('embed-src', 'src');
+    if (element.namespaceURI === SVG_NAMESPACE && element.tagName === 'image') {
+      addAttribute('svg-image', attrs.href !== undefined ? 'href' : 'xlink:href');
+    }
+    if (element.namespaceURI === SVG_NAMESPACE && element.tagName === 'use') {
+      addAttribute('svg-use', attrs.href !== undefined ? 'href' : 'xlink:href');
+    }
+
+    const styleAttribute = element.attrs.find((attribute) => attribute.name === 'style');
```
wyuc, line 234: [P2] Inventory SVG URL-valued presentation attributes

The structured scan only parses the `style` attribute, so SVG presentation attributes such as `fill="url(https://cdn.test/pattern.svg#p)"`, `filter`, `mask`, and `clip-path` are neither inlined nor reported as unresolved. Preparation marks the scene successful, then the export CSP blocks the remaining request and the captured frame loses the effect. Parse these URL-valued SVG attributes structurally, or conservatively retain them as unresolved.

lib/export/html-asset-inventory.ts
```diff
@@ -0,0 +1,285 @@
...
+    if (element.tagName === 'source') {
+      addAttribute('source', 'src');
+      addAttribute('srcset', 'srcset');
+    }
+    if (element.tagName === 'video') {
+      addAttribute('video', 'src');
+      addAttribute('poster', 'poster');
+    }
+    if (element.tagName === 'audio') addAttribute('audio', 'src');
+    if (element.tagName === 'iframe') addAttribute('iframe-src', 'src');
```
wyuc: [P2] Reject or recursively freeze iframe srcdoc documents

An `iframe[srcdoc]` has no `src`, so it bypasses this embedded-document rejection and is accepted as self-contained. The injected `freeze()` only patches the owning window; it does not recurse into nested frames. In Chromium, a packaged fixture with a timer in `srcdoc` remained live after the parent entered `frozen` state (the nested counter advanced from 11 to 23), making captured frames nondeterministic. Either reject `srcdoc` here or package and freeze nested documents recursively before signaling readiness.

lib/export/inline-assets.ts
```diff
@@ -159,140 +254,222 @@ export async function inlineCssUrls(
...
+      if (!report.failed.some((g) => g.url === f.url)) report.failed.push(f);
+    // Merge: start from originals, overlay inlined data: entries.
+    // Merge: original prefix entries are retained as online fallback; inlined explicit
+    // data: entries take precedence for the modules we inlined. Keeping both is safe per
+    // the importmap spec (explicit specifier shadows prefix key) and strictly more correct:
+    // a sub-path not seen during static analysis can still resolve via the prefix online.
+    const merged: Record<string, string> = keepFallbacks ? { ...orig, ...inlined } : inlined;
+    patches.push({
+      range: importmap.contentRange,
+      replacement: JSON.stringify({ imports: merged }),
```
wyuc: [P2] Preserve and resolve import-map scopes

Re-serializing every map as `{ imports: merged }` drops valid top-level fields, especially `scopes`, and the resolver only consults global `imports`. A module under `https://app.test/` whose `dep` is mapped by a scope to `https://cdn.test/dep.js` is instead resolved as `https://app.test/dep` and rejected as unresolved. Preserve the full parsed map and make dependency resolution scope-aware, or reject scoped maps explicitly instead of silently changing their semantics.

wyuc (changes requested):
The five findings from the previous round are addressed, and the structured parsing direction is a substantial improvement. I found four remaining correctness gaps in the offline/static guarantee. Each can produce a scene that is rejected incorrectly or accepted as packaged while still rendering nondeterministically or incorrectly.

cosarah:
Thanks for the detailed follow-up. I addressed all four findings in 7fc963e5.

- CSS asset packaging now separates the fetch URL from its fragment. It fetches `filters.svg` once and appends `#blur` (or the original fragment) to the resulting data URI.
- The parse5 inventory now includes URL-valued SVG presentation attributes (`fill`, `filter`, `mask`, `clip-path`, markers, stroke, and cursor). Their values go through the same structured CSS parser and residual validation as style declarations.
- `iframe[srcdoc]` is now an explicit unsupported embedded-document resource. It enters the existing `unresolved-resource` fallback path, so a nested live document cannot be accepted as frozen.
- Import-map serialization preserves supported top-level fields instead of rebuilding only `{ imports }`. Non-empty or invalid `scopes` are deliberately rejected as `unsupported-importmap-scopes` before any fetch or rewrite, because data-URI module packaging changes referrer URLs and cannot preserve scoped resolution semantics safely yet. Empty `scopes` and other top-level fields such as `integrity` are preserved.

Regression coverage was added for each case. Validation completed:

- Focused packaging tests: 76/76 passed.
- Full export/video-export suite: 415 passed, 43 skipped.
- Chromium interactive-static suite: 4/4 passed.
- ESLint, Prettier, `tsc --noEmit`, and `git diff --check` passed.
- All five GitHub checks are green.

wyuc (approved):
Approved — the core rendering path, focused unit tests, Chromium E2E, and CI are all green.

Non-blocking follow-ups for uncommon input shapes:

- Keep scoped import-map rejection from silently bypassing unrelated asset inlining for classroom ZIP and PPTX/resource-pack callers, or make those callers surface the unresolved result.
- Skip image decoding during readiness when an img element has neither src nor srcset.
- When module source bytes are rewritten, remove or recompute an existing integrity attribute.
- Consider handling statically resolvable template-literal dynamic imports such as import(`https://cdn.example/dep.js`).
- Quote or escape rewritten CSS data URLs when an SVG fragment contains characters that are invalid in an unquoted url() value.

These are low-incidence edge cases and are not blockers for this PR.

PR #1087, opened 2026-08-10
feat(generation): move scene generation and PBL single-call planning into the package (#1057 Part C)

wyuc:
Part C of #1057, per the [Part 0 RFC](https://github.com/THU-MAIC/OpenMAIC/issues/1057#issuecomment-5193492850) — the largest move, and still purely additive: app-side `lib/**` is byte-untouched (repointing and deletion are Part D).

## What moves

- **Scene layer**: `generateSceneContent` (all four kinds), `generateWidgetContent`, `generateSceneActions`, widget-config extraction (with type seeding), `withGenerationRetry`, `parseActionsFromStructuredOutput`, `postProcessInteractiveHtml`, the remaining prompt formatters, and `buildCompleteScene` with the RFC's injectable `sceneId` (default stays random; stable identity makes a re-run an upsert instead of a duplicate append).
- **The two forbidden edges are cut**, per the RFC: no stage-store API (`createSceneWithActions` stays app-side as the host adapter) and no `lib/pbl` imports.
- **PBL single-call planning re-seats on `AICallFn`**: the pure planner core and kernel operations move in (contract types come from `@openmaic/dsl`, never copied), the planner templates ship as package-relative assets, and the pbl branch fails loud without the optional injected `pblLoopFallback`. A re-seat proof test drives identical canned responses through the app path and the package path and asserts deep-equal projects.
- The AI SDK and zod stay out — the loop planner remains app-side, injectable.

## Verification (Node 22)

Generation 160 package tests (behavior per scene kind, re-seat proof, sceneId upsert semantics, type-surface sentinels, golden prompt snapshots) · dsl 220 / storage 659 · **full app suite 4,295 passing** · Node consumer smoke extended to a full slide scene validated by `validateScene` · temp-dir dist import · five-package tarball smoke · CI prompt-parity guard extended to the PBL planner assets · version `0.1.1 → 0.2.0`; dependencies (`katex`, `partial-json`) and the boundary-lint allowlist widened together to the actual import set.


xuyuanwei678 (approved):
Reviewed the package-boundary migration, PBL single-call planner failure/fallback behavior, DSL/runtime compatibility, and published package contents.

No blocking issues found.

Verified locally:
- @openmaic/generation typecheck and 163 tests
- @openmaic/dsl typecheck and 220 tests
- 50 targeted app compatibility tests
- package build, lint, tarball contents, and Node ESM import

The root test invocation also surfaced four timeouts in the untouched cross-tab quiz runtime suite; those tests are outside this PR's changed paths, so they are not treated as PR findings.

PR #1088, opened 2026-08-10
fix(persistence): stop corrupting binary response bodies in the route adapter

wyuc:
The embedded persistence route bridges a Node `RequestListener` to the Fetch API by hand. Its response object is cast with `as unknown as ServerResponse`, so the compiler checks none of that surface, and two parts of it are wrong in ways only a byte-carrying handler would hit.

**`end` decoded bytes as text.** It ran `Buffer.from(chunk).toString()` on any `Uint8Array` chunk, which is UTF-8. `ServerResponse.end` accepts a `Uint8Array`, and those bytes are not necessarily valid UTF-8, so every unpaired byte became U+FFFD — silent corruption with no error anywhere. The adapter now buffers chunks as bytes and builds the response from them.

**`write` was absent.** Any chunked handler was therefore a runtime `TypeError` rather than a compile error. It is part of the surface this object claims to implement, so it is now implemented.

Neither reaches a caller today: the document and runtime handlers only ever end with JSON strings, which took the string branch. Both are traps laid for the first handler that carries bytes — which is why I found them while working on the asset server backend, whose byte routes would have hit both.

Both cases are pinned by tests that fail before this change: the binary one on a byte comparison, the chunked one with a `500` from the `TypeError`. The existing adapter round-trip test, whose comment already called this the most bug-prone code in the route, covered only the string path.


wyuc:
I ran an adversarial pass over this adapter rather than only the two lines it fixes, and it is worth recording what it found — not to expand this PR, but because the reason these defects existed is that `as unknown as ServerResponse` silences the compiler for the whole interface, and that is still true after this change.

**Nothing found affects the handlers wired into this route today.** The call shapes the storage package actually uses — UTF-8 strings, `Uint8Array`, `end()` with no chunk — are correct after this fix. What follows is latent.

Three traps remain, each a place where the adapter diverges from `ServerResponse`:

1. **String encodings and callbacks are not implemented.** `write(s, 'latin1')` produces UTF-8 bytes; `end(callback)` treats the function as a chunk and throws; `end(chunk, callback)` drops the callback. Storage passes neither encodings nor callbacks.
2. **Terminal state is not enforced.** A `write()` after `end()` returns `true` and appends unreachable data instead of `ERR_STREAM_WRITE_AFTER_END`. Storage's handlers return immediately after `end()`.
3. **Bodyless responses do not suppress buffered data.** `writeHead(204); write('x'); end()` builds a Fetch `Response` with a body and throws, surfacing as a 500 where Node would send a bodyless 204. The same mismatch applies to HEAD. Storage's 204 paths call `end()` with no data, and the asset handler's HEAD path passes `undefined` explicitly.

I have deliberately left all three. This PR is about bytes being corrupted; making the adapter a faithful `ServerResponse` is a different change with a different blast radius, and the third one in particular is only reachable by a handler that writes a body to a bodyless status — which would be a bug in that handler too.

Worth flagging for whoever wires the asset store into this route: `createStorageHttpHandler` is not currently given an `assetStore` here, so the byte-carrying paths this fix exists for are not yet reachable in production. When they are, item 3 is the one to re-check first, since asset responses are the only ones that will carry bytes and a HEAD on them is a hot path.

app/api/persistence/[...path]/route.ts
```diff
@@ -140,20 +149,24 @@ function runNodeHandler(handler: RequestListener, request: Request): Promise<Res
...
         return this;
       },
+      write(chunk: string | Uint8Array, encodingOrCallback?: unknown, callback?: unknown) {
+        // `write` is part of the `ServerResponse` surface this object claims to
+        // implement. Omitting it made any chunked handler a runtime TypeError
+        // that the `as unknown as ServerResponse` cast hid from the compiler.
+        headersSent = true;
+        appendChunk(chunk);
+        const done = typeof encodingOrCallback === 'function' ? encodingOrCallback : callback;
+        if (typeof done === 'function') done();
```
Comment (cosarah) [inline app/api/persistence/[...path]/route.ts]: Defer `write` callbacks until after `write` returns. Node invokes these callbacks asynchronously after handing off the chunk, but this implementation calls them inline. A handler that writes its next chunk from each callback therefore recurses synchronously and can eventually overflow the stack; callback exceptions also occur in the wrong execution phase.

app/api/persistence/[...path]/route.ts
```diff
@@ -140,20 +149,24 @@ function runNodeHandler(handler: RequestListener, request: Request): Promise<Res
         if (values) setHeaders(headers, values);
         return this;
       },
+      write(chunk: string | Uint8Array, encodingOrCallback?: unknown, callback?: unknown) {
+        // `write` is part of the `ServerResponse` surface this object claims to
+        // implement. Omitting it made any chunked handler a runtime TypeError
+        // that the `as unknown as ServerResponse` cast hid from the compiler.
+        headersSent = true;
+        appendChunk(chunk);
```
Comment (cosarah) [inline app/api/persistence/[...path]/route.ts]: Honor the encoding passed to `write`. `encodingOrCallback` is currently only inspected for a callback, while `appendChunk` always encodes strings as UTF-8. A valid handler calling `response.write("é", "latin1")` should emit byte `0xe9`, but this adapter emits `0xc3 0xa9`, corrupting the body and potentially invalidating `Content-Length`.

app/api/persistence/[...path]/route.ts
```diff
@@ -140,20 +149,24 @@ function runNodeHandler(handler: RequestListener, request: Request): Promise<Res
...
-              ? undefined
-              : typeof chunk === 'string'
-                ? chunk
-                : Buffer.from(chunk).toString(),
-            {
-              status,
-              headers,
-            },
-          ),
+          new Response(body.length === 0 ? undefined : Buffer.concat(body), {
```
Comment (cosarah) [inline app/api/persistence/[...path]/route.ts]: Suppress response bodies for statuses that cannot carry one. Node discards writes after `writeHead(204)` or `writeHead(304)`, but this passes the buffered body to the Fetch `Response` constructor, which throws for those statuses. The request consequently becomes a 500 instead of returning the intended 204/304.

wyuc:
All three are fixed in `41e384ab`, and you were right to push back on the scoping.

I had found the same three in my own review pass and deliberately left them, on the grounds that no current caller reaches them and that turning a binary-corruption fix into "make the adapter a faithful `ServerResponse`" was a different change. That reasoning does not hold up. `write` being absent was also something no current caller reached — right up until it became the reason this PR exists. Using today's call surface to justify leaving tomorrow's broken is the argument that produced these defects in the first place, in an adapter whose entire purpose is to have handlers plugged into it.

Your framing was also sharper than mine: I had them as shape mismatches, and the synchronous callback is a stack overflow with a concrete trigger — a handler that writes its next chunk from each callback, which is the ordinary streaming idiom.

What changed:

- **Callbacks are deferred.** A handler that writes from its callback now completes without synchronous recursion, and the callback runs after `write` returns.
- **The encoding argument is honoured** on both `write` and `end`, with the full overload set parsed. For an invalid encoding I probed Node 22's real `ServerResponse` rather than guessing: it throws `ERR_UNKNOWN_ENCODING` synchronously, so the adapter validates with `Buffer.isEncoding` and lets `Buffer.from` raise the native error rather than inventing one.
- **Bodyless statuses discard the buffer** — 204, 205 and 304, and `HEAD`.

Each is pinned by a test that fails when its fix is reverted, including the callback one, which fails with the recursion the deferral prevents.

cosarah (approved):
LGTM

wyuc:
Your approval was invalidated by the `main` merge you pushed here, so this needs one more look. The delta since you approved is your merge and nothing else — I have pushed no commits since `41e384ab`.

That merge brought in #1087 and #1090, which delete `lib/generation` and move scene generation into the package: 148 files, roughly +8700/−9700. None of it touches the persistence route adapter, but a change that size is worth verifying rather than assuming, so I ran the merged tree end to end: install, build, typecheck, prettier, lint, and the whole suite at 4307 tests, all clean.

No action needed from me that I can see — re-approving is all this is waiting on. If you would rather I rebase onto current `main` instead of carrying your merge commit, say so and I will.

cosarah (approved):
LGTM

PR #1089, opened 2026-08-10
feat(storage): wire the asset backend into the app (#1007)

wyuc:
Connects the #1007 asset backend to the app. The server registry shipped with #1007 but was reachable from nothing — the persistence route never passed an `assetStore`, and no file under `app/`, `lib/` or `components/` imported the asset server or client modules.

**Additive.** Nothing changes when server persistence is off: the pool is still `BrowserAssetStore`, and the authenticator change is server-only.

> **Depends on #1088**, which is merged into this branch. That PR fixes the Fetch↔Node adapter decoding binary response bodies as UTF-8 — without it every asset `GET` through this route returns corrupted bytes. This diff shrinks once #1088 lands.

> **Breaking for `@openmaic/storage`, hence `0.3.0`.** `S3AssetByteStoreOptions` gained a required `commands`, so an existing `{ client, bucket }` construction no longer type-checks. That is the cost of the S3 module no longer importing the SDK at module load. The realistic blast radius is zero — the path is an optional peer dependency reached only by setting `ASSET_S3_BUCKET`, and it shipped hours ago — but it is a break and the version says so.

## Server side

The route builds a `PgAssetStore` on the same pool and transaction as the document store, with `PgAssetByteStore` by default and `S3AssetByteStore` when `ASSET_S3_BUCKET` opts a deployment in. The AWS SDK stays an optional peer, reached through one lazy loader in one resolution scope; with the bucket unset nothing resolves it, and a test mocks the SDK to throw on resolution and asserts that importing the byte-store module never does.

The offline collector is deliberately not scheduled here, and the route says so where a reader would look for it — leaving it unscheduled grows storage without bound, and scheduling it is the deployment's decision.

The route also states plainly that this development authenticator takes the asset partition key from a client-supplied header, so the cross-principal isolation the asset contract describes is **not** in force under it. That warning previously lived only in the auth module's own header, where nobody mounting the route would see it.

## Browser side

The pool gains a single-shot configuration seam modelled on the document store's, so with server persistence enabled it is an `HttpAssetStore` carrying the same auth headers as the other stores.

Three browser-specific mechanisms needed decisions rather than translation:

**`clearAssetPool`** deletes an IndexedDB database today. In server mode there is nothing local to delete, and a "clear cache" action must not remove server-side assets — that would destroy user data from a button promising to free space. It now revokes local object URLs and closes the client, deletes nothing remote, and always clears the singleton so the next call rebuilds a usable client.

**The cross-tab replacement broadcast** is unchanged in intent but needed a fix to remain effective. Re-resolving on the broadcast coalesced with a GET that started *before* the peer's replacement — both behaviours correct alone, and together they returned the old revision and left the consumer stranded on it. The client gained an invalidation that retires its cached snapshot and advances the per-id generation without revoking any URL already issued. `release` would have been wrong: the contract keeps an issued URL valid until its holder releases it.

**The exclusivity proof** that decides whether an in-place `replace` is safe reasons over this browser only — local documents, unflushed state, a cross-tab presence probe. In server mode those cover a strict subset of the holders, since another device can reference the same id and no probe can see it, and asking the server who else references an id would be an existence oracle over other principals. The proof therefore returns not-exclusive unconditionally in server mode and regeneration takes the existing fork path. Fail-closed, and a graceful degradation rather than a break.

## Verification

Whole repository: 4306 tests. Storage package against PostgreSQL 16: 861, none skipped that infrastructure could have run.

Three of the defects above were found by review rather than by the tests that existed, and the tests were not wrong so much as aimed past them — the route test used JSON text so it never exercised the byte path, and the S3 test injected a loader rather than exercising the imports. Each fix now carries a test that fails when the fix is reverted; the broadcast race in particular fails without its one-line invalidation.


cosarah:
Triage verdicts for the discovery findings:

1. **KEEP — Asset collector is not wired into the persistence route.** `PgAssetStore.remove` and content-changing `replace` only mark `asset_blobs.unreferenced_at`; `AssetCollector.collect` is the sole deletion path, while the route explicitly does not schedule it. The documented Compose deployment runs only the app and PostgreSQL, so normal server-backed asset churn retains PostgreSQL bytes or S3 objects indefinitely.

2. **KEEP — `S3AssetByteStore` constructor compatibility.** `S3AssetByteStoreOptions.commands` is now required and the methods call it directly, so existing `{ client, bucket }` callers either fail TypeScript compilation or hit an undefined-command error at runtime. `./asset/s3-bytes` is a public package export, making this a real downstream API break.

wyuc:
Both KEEPs are addressed on the branch:

1. **Collector scheduling** — c6ef5113 schedules `AssetCollector` from `instrumentation.ts` (run once per server process by Next), on by default with a 15-minute interval and the one-hour grace period, so the shipped Compose stack collects with no operator action. `ASSET_COLLECTION_ENABLED` / `ASSET_COLLECTION_INTERVAL_MS` / `ASSET_COLLECTION_GRACE_MS` change or disable it; without `DATABASE_URL` nothing is scheduled. Byte-store selection moved to `lib/persistence/asset-byte-store` so the collector reclaims through the same layer the request path wrote through (a PostgreSQL collector against S3 writes would orphan objects). 8b1637c1 then bounds each pass (`batchSize`, default 1000, oldest-unreferenced-first) and adds `collectPass()` so a backlog is drained across intervals instead of in one unbounded loop.

2. **S3 constructor compatibility** — 6062ffba makes `S3AssetByteStoreOptions.commands` optional again: omitted, the store resolves the AWS SDK command constructors through a dynamic import on first use and caches them, so the existing `{ client, bucket }` construction type-checks and runs. Module import and construction still never resolve the optional peer, and the isolation tests pin that. With the break gone the package version is back to a patch (0.2.4 → 0.2.5) rather than 0.3.0.

cosarah:
P2 KEEP: Preserve re-openability after clearing a concrete asset pool. In the latest commit 8b1637c1, clearAssetPool() closes and clears the cached singleton, while resolveConfiguredAssetPoolStore() returns the same concrete store object again when configured with { store: existingStore }. After configureAssetPoolStorage({ store: existingStore }) and clearAssetPool(), the next getAssetPool() can therefore reinstall the already-closed instance. Factory-based configuration avoids this, but the public API explicitly accepts an already-constructed store.

cosarah:
P2 KEEP: Isolate asset-store initialization from unrelated persistence requests. In the latest commit 8b1637c1, createPersistenceHandler() resolves ASSET_S3_BUCKET and eagerly awaits createAssetByteStore() before returning the shared persistence handler. An invalid configured bucket or an unavailable optional S3 SDK rejects handler initialization, so handlePersistenceRequest() fails for runtime and document traffic too, even though the asset backend is optional. The failure should be contained to asset storage or initialization should be deferred.

wyuc:
Both P2s are fixed in f7216933:

1. **Re-openability after clear** — `resolveConfiguredAssetPoolStore()` now hands a concrete instance out once. After `clearAssetPool()` closes it, the next resolution throws, naming the factory form (`store: () => new ...`) as the fix, instead of reinstalling the closed store. Factory configuration is unchanged and still rebuilds on every resolution. Both shapes are pinned by tests in `tests/media/asset-pool.test.ts`.

2. **Asset backend initialization** — the route no longer awaits the byte store during handler initialization. `lazyAssetByteStore()` defers bucket validation and SDK resolution to the first asset byte operation, so a malformed `ASSET_S3_BUCKET` or an unresolvable SDK fails asset requests only; document and runtime traffic initializes and serves regardless. A failed construction is not cached, so the next asset request retries, mirroring the route's own no-poisoned-singleton rule. Covered by three tests in `tests/persistence/route.test.ts`, including the retry-after-SDK-failure case.

cosarah (approved):
LGTM

PR #1100, opened 2026-08-11
feat(storage): opt-in indirect asset byte egress (#1007)

wyuc:
Adds the signed object-storage shape the asset HTTP contract has anticipated since it shipped ("Self-hosted proxy paths and signed object-storage URLs remain possible later shapes; they are additive"). With indirect egress enabled, an asset byte `GET` is answered `302 Found` with a short-lived signed URL instead of the materialized bytes; large media downloads move off the application server onto the object store.

**Additive and opt-in.** Off by default and byte-for-byte unchanged when off: a deployment that never sets `ASSET_BYTE_EGRESS=redirect` cannot tell the branch exists, and a byte layer that cannot sign (the PostgreSQL byte column cannot) still answers `200` with bytes even when opted in.

The design keeps every invariant the contract already guarantees:

- **Authorization is still per-read.** The `302` is minted only after the same policy hook and per-principal ownership check the direct read performs, inside the same coordinated read; a miss is still `404 ASSET_NOT_FOUND`. The redirect changes who serves the bytes, never who may ask.
- **The signed URL is short-lived** (default 60s, far below the collector grace period), because its signature is a bearer credential for its whole lifetime.
- **The contract's response headers are pinned into the signature** (`response-content-type` / `response-content-disposition` / `response-cache-control`), so the renderable-allowlist relabeling, the fixed disposition, and the no-store posture survive the redirect.
- **HEAD and writes never redirect**, so `X-Asset-Revision` revalidation works exactly as under direct egress; the `302` itself also carries the revision for clients that do not follow opaquely.

The disclosure tradeoff is stated in the contract text rather than elided: the shipped byte layers key objects by content hash, so a `Location` into one names the hash and reintroduces the byte-equality disclosure the three-layer design exists to prevent. Enabling indirect egress with a hash-keyed layer is a deliberate deployment tradeoff (public generated media is the motivating case); a deployment that needs the no-disclosure property keeps direct egress.

## Implementation

- `AssetByteStore` gains an optional `signReadUrl(hash, headers)`; `S3AssetByteStore` implements it via `@aws-sdk/s3-request-presigner`, a new optional peer resolved with the same lazy discipline as `@aws-sdk/client-s3` (never resolved unless a deployment both configures S3 and opts into redirect egress). The packaged version goes to `0.3.0`.
- `PgAssetStore.resolveIndirect` performs the same ownership-checked entry read and blob-row lock as `resolve`, then signs; the presign is local arithmetic, so no byte read happens on the redirect path.
- The server's asset handler gains `byteEgress: 'direct' | 'redirect'` (default `direct`) and `signedUrlTtlSeconds`; the app route wires them from `ASSET_BYTE_EGRESS`. `lazyAssetByteStore` forwards the signing capability, declining when the resolved layer cannot sign.

## Tests

New coverage pins: off-by-default with a signing store; the 302 shape and its headers; fallback on a non-signing store and on a declined sign; authorization and 404 behavior identical to the direct path; HEAD never redirects; the presigner is never resolved unless opted in; the route's env wiring. The one pre-existing `http-asset-store.test.ts` multipart-message failure on some machines is unchanged and unrelated.

Refs #1007


wyuc:
Codex review loop has converged: 11 rounds, findings through round 10, final round clean. What the rounds changed, by theme:

- **Client survival under redirect.** The first shape had the client follow the 302 and lose the revision; the fix went through a HEAD-probe design before landing on descriptor negotiation: the client asks via `Accept`, the server answers a JSON `{"url", "revision"}`, and the signed URL is fetched with no deployment headers at all — a followed 302 forwards custom credential headers cross-origin, which is exactly the header-authenticated deployment this contract cares about.
- **Negotiation hygiene.** CORS-safelisted `Accept` instead of a custom request header (no new preflights for direct-egress deployments); exact, case-insensitive media-type matching on both sides; `q=0` honored; the descriptor media type is reserved from the renderable allowlist so no payload can parse as a descriptor.
- **Lifetime invariants.** Signed URL TTL capped at fifteen minutes so a URL cannot outlive the reclamation grace; the app refuses `redirect` + a too-short `ASSET_COLLECTION_GRACE_MS` at startup; an empty grace value reads as unset, matching the collector's own parsing.
- **Fallback discipline.** A missing optional presigner declines to direct bytes instead of failing the read; signing moved outside the registry transaction, since credential resolution can wait on the network while otherwise holding the blob-row lock.
- **Deployment.** Both AWS packages are declared runtime dependencies and traced into the standalone image via `serverExternalPackages` plus literal-but-never-called import thunks, verified against a local standalone build (`node_modules/@aws-sdk` now present in the output).

Local suites are green except one pre-existing multipart-message assertion that fails identically on the base commit.

packages/@openmaic/storage/src/asset/s3-bytes.ts
```diff
@@ -222,6 +316,54 @@ export class S3AssetByteStore implements AssetByteStore {
...
+    } catch {
+      // An unresolvable optional presigner means this layer cannot sign after
+      // all, which is a capability answer, not a failed read: decline, and the
+      // caller falls back to serving the bytes directly.
+      return undefined;
+    }
+    try {
+      return await signer.sign(
+        this.client,
+        commands.get({
```
cosarah, line 352: Enabling S3 redirect egress always violates the shared-blob response-header contract here. S3 `GetObject` adds `ETag` and `Last-Modified`, and these response overrides cannot remove either header. Because every deduplicated PUT rewrites the hash-keyed object, a holder of an existing asset can observe `Last-Modified` change when another principal later uploads the same bytes. The documented tradeoff accepts hash-equality disclosure, but not shared-object write timing. Please either use an egress layer that strips these headers or explicitly revise the contract/tradeoff; the shipped S3 signer cannot currently satisfy it.

packages/@openmaic/storage/src/asset/http.ts
```diff
@@ -204,6 +205,14 @@ export class HttpAssetStore implements StorageProvider {
     }
     const headers = await this.headers(method, path, body !== undefined);
     if (body !== undefined) headers['content-type'] = body.type;
+    if (method === 'GET') {
+      // Ask for a descriptor answer on the byte read: a redirect-egress
+      // server then returns the signed URL in a JSON body instead of a 302
+      // the platform fetch would follow with these headers attached, custom
+      // credential headers included. Accept is CORS-safelisted, so the
+      // negotiation adds no preflight.
+      headers['accept'] = ASSET_DESCRIPTOR_MEDIA_TYPE;
```
cosarah: This `Accept` value says the descriptor is the only acceptable representation, even though this client intentionally supports ordinary asset bytes when the server uses direct egress or signing fallback. The bundled handler ignores `Accept` on its direct path, but a strict compatible server or content-negotiating layer can legitimately return `406 Not Acceptable`. Please advertise both representations, for example `application/vnd.openmaic.asset-descriptor+json, */*;q=0.9`, while retaining the descriptor preference.

lib/persistence/asset-byte-store.ts
```diff
@@ -87,5 +99,14 @@ export function lazyAssetByteStore(
     write: async (hash, bytes) => (await resolve()).write(hash, bytes),
     read: async (hash) => (await resolve()).read(hash),
     delete: async (hash) => (await resolve()).delete(hash),
+    // The wrapper always carries the method, so capability is decided by the
+    // answer, not by method presence: an inner layer without a signer answers
+    // `undefined`, and the caller falls back to direct bytes. That keeps the
+    // opt-in indirect egress a no-op for the PostgreSQL byte column without
+    // the wrapper knowing which layer it will resolve to.
+    signReadUrl: async (hash, headers) => {
```
cosarah, line 112: When `ASSET_S3_BUCKET` is unset, this wrapper resolves to `PgAssetByteStore`, but it still advertises `signReadUrl`. `PgAssetStore.resolveIndirect` therefore performs the ownership query and blob-row `FOR SHARE`, this method returns `undefined`, and the handler falls through to `resolve`, repeating the transaction, query, and lock before reading the bytes. This happens on every cold GET in the documented `ASSET_BYTE_EGRESS=redirect` PostgreSQL fallback. Please omit the signing method when the wrapper already knows the configured layer is PostgreSQL, while preserving lazy validation for non-empty bucket values.

cosarah (commented):
Three actionable findings from the indirect-egress triage.

wyuc:
All three findings are addressed in `ea38d982`:

- **Accept advertises both representations** (`http.ts`): the byte GET now sends `application/vnd.openmaic.asset-descriptor+json, */*;q=0.9` — descriptor preferred, bytes always acceptable — so a strict negotiating layer can never answer 406.
- **No signing method on the known-PostgreSQL wrapper** (`asset-byte-store.ts`): with no bucket configured the layer is known at construction, so `signReadUrl` is simply absent and the handler's feature-detect fails fast. The double ownership-query-and-lock you flagged on every cold GET is gone. With a bucket configured, lazy validation is unchanged and the wrapper still declines when the resolved layer cannot sign.
- **The tradeoff now includes `Last-Modified`** (`asset-http-contract.md`): the contract text states plainly that object-store GETs carry `ETag`/`Last-Modified`, that response overrides cannot strip them, and that because deduplicated PUTs rewrite the hash-keyed object, `Last-Modified` moves when another principal re-uploads the same bytes — shared-object write timing, a disclosure beyond byte equality. Deployments for whom that signal is sensitive must keep direct egress; a never-rewrite byte layer (or per-entry keys) would not expose it, and none ships today.

cosarah:
Triage of the discovery findings (all four are actionable):

- **[P1] Isolate invalid grace configuration from shared routes — KEEP.** `assertEgressWithinCollectionGrace()` runs while constructing the shared persistence handler. With redirect egress and an invalid/too-short `ASSET_COLLECTION_GRACE_MS`, initialization rejects and closes the pool, so `/runtime` and `/documents` also return 500 even though they do not use asset egress.

- **[P1] Validate collection grace at the package boundary — KEEP.** `createAssetHttpHandler()` validates the signed-URL TTL (including the 900-second cap) but has no knowledge of `AssetCollector.graceMs`. A package consumer can combine `signedUrlTtlSeconds: 60` with `graceMs: 30_000`, allowing collection to delete an object while its signed URL remains valid.

- **[P2] Preserve nosniff on signed S3 responses — KEEP.** Direct byte responses set `X-Content-Type-Options: nosniff`, but `S3AssetByteStore.signReadUrl()` signs only content type, disposition, and cache-control. The followed S3 response therefore lacks the contract-required `nosniff` header.

- **[P2] Exercise the real application egress boundary — KEEP.** The app route tests mock `createStorageHttpHandler`, `PgAssetStore`, and byte stores, so they never perform an actual `/assets/.../content` request through the composed handler. Wiring, redirect/descriptor behavior, and direct fallback regressions could therefore pass unnoticed.

**KEEP set:** all four findings above.

wyuc:
All four findings are addressed in `6fb7810f`:

- **Grace misconfiguration degrades instead of breaking the shared handler.** `ASSET_BYTE_EGRESS=redirect` with an invalid or too-short `ASSET_COLLECTION_GRACE_MS` now logs a warning and falls back to direct egress; document and runtime traffic is never touched. (This also corrects my earlier choice to fail loudly — the asset backend is optional, so its config errors are contained to it.)
- **TTL/grace invariant at the package boundary.** The package now exports `assertSignedUrlTtlWithinGrace(ttlSeconds, graceMs)` so a deployment wiring the handler and the collector separately validates the invariant once with both numbers in hand; the handler alone could not know the collector's grace.
- **nosniff on signed responses.** You are right that the signed response cannot carry it — response-header overrides cannot set `X-Content-Type-Options`. The contract now states that explicitly and why it is safe for conforming clients: the descriptor path mints a typed `blob:` URL from the fetched bytes, so nothing ever sniffs the signed response's type; the allowlist relabeling has already decided what the bytes claim to be. Direct navigation to a signed URL gets no nosniff protection, which the text now calls out.
- **Real egress boundary test.** The route test now drives an actual byte `GET` through the Fetch↔Node adapter and the real `createStorageHttpHandler` composed with an in-memory asset store — wiring, direct fallback, and response shape are exercised end to end rather than through mocks.

cosarah:
Triage of the discovery findings against the effective worktree code:

- **[P1] Enforce the grace/TTL invariant at the package boundary — KEEP.** `createAssetHttpHandler()` only caps `signedUrlTtlSeconds`; `AssetCollector` accepts an independently configured `graceMs`. A package consumer can therefore enable redirect egress with a 60-second signed URL and a 30-second collection grace, allowing reclamation to delete the object while the URL is still valid. The application-specific guard does not protect other `@openmaic/storage` consumers.

- **[P2] Preserve missing-object semantics for indirect reads — KEEP.** `PgAssetStore.resolveIndirect()` checks the registry row but does not verify that the S3 object still exists before signing. If the object is deleted or orphaned while the row remains, the descriptor is returned and the client maps the signed-URL 404 to `MALFORMED_RESPONSE`, violating the documented `ASSET_NOT_FOUND` miss semantics.

- **[P2] Exercise redirect egress through the application route — KEEP.** The route tests use mocked `createStorageHttpHandler`/`PgAssetStore` wiring and do not perform an actual `/assets/.../content` request through the composed handler. A regression in route option wiring, descriptor negotiation, `resolveIndirect`, or byte-store composition could therefore pass.

- **[P3] Browser cross-origin descriptor flow — DISMISS.** The client already fetches signed URLs with `credentials: omit` and without deployment headers; object-store CORS is explicitly deployment-owned by the contract. Browser coverage would be useful, but this is an environment-specific long-tail test gap rather than a package defect.

**KEEP set: findings 1, 2, and 3.**

wyuc:
All three KEEP findings are addressed in `4ece7de4`. Each one is resolved by relocating a rule rather than by adding a check, which I think is why the first two kept coming back.

**1. Grace/TTL invariant at the package boundary.** You were right twice over, and the exported `assertSignedUrlTtlWithinGrace` was the wrong answer both times: it protects only consumers who remember to call it, which is not a boundary. The invariant now lives in the option shape instead:

```ts
byteEgress?: 'direct' | {
  mode: 'redirect';
  collectionGraceMs: number;        // required
  signedUrlTtlSeconds?: number;     // defaults to 60
}
```

Enabling indirect egress without declaring the reclamation window it lives inside no longer type-checks, and `createAssetHttpHandler` validates both numbers at construction. Your 60s-TTL-with-30s-grace combination is now unrepresentable rather than merely detectable. The flat `signedUrlTtlSeconds` option and its cross-field `requires byteEgress "redirect"` check are gone with it, so this is a net deletion. The 900-second ceiling stays, but alongside the ratio rather than in place of it: the two bind in different deployments, the ratio under a short grace and the ceiling under a long one. There is deliberately no default for the grace, since a default here would be a guess about a component the handler does not own, and the guess is exactly what lets a deployment hold the invariant wrongly while every individual setting looks reasonable. On the app side both the route and the collector now resolve the variable through one parser, so the number the handler checks is the number the collector runs with.

**2. Missing-object semantics for indirect reads.** The asymmetry is real and now fixed, though not where you proposed. The client maps a `404` from the signed-URL fetch to `ASSET_NOT_FOUND`, matching the direct path. Verifying object existence before signing would reopen something this design closed: `signReadUrl`'s contract forbids reading or probing the bytes, because an existence probe makes the mint's cost vary with prior presence, and it would also restore the round trip indirect egress exists to remove. So the requirement moves onto the byte layer instead: an absent object MUST answer `404`. The contract now states that, and states that S3 answers `404 NoSuchKey` only when the signing identity holds `s3:ListBucket` (without it a missing key answers `403`, indistinguishable from a real credential fault, and the miss degrades to `MALFORMED_RESPONSE`). A `403` is deliberately *not* mapped to a miss, so a genuine credential fault still fails loud.

**3. Redirect egress through the application route.** Fair, my previous boundary test only drove the direct case. The real composed handler now covers the full matrix, all through the route's `Fetch<->Node` adapter and the real `createStorageHttpHandler`:

| | expected |
|---|---|
| redirect, no `Accept` | `302` + `Location` + `X-Asset-Revision` + `no-store` |
| redirect, descriptor `Accept` | `200` + descriptor media type + `{url, revision}` |
| redirect, byte layer declines to sign | `200` + bytes |
| redirect configured, grace too short | `200` + bytes (route degraded at wiring time) |
| direct | `200` + bytes |

I verified all three fixes by mutation rather than by the suite passing: dropping the client's 404 mapping, dropping the handler's grace check, and dropping the route's degradation each turn the corresponding test red.

On the P3 you dismissed, agreed, and no change made.

Two process notes. If this matrix is the right closure criterion for the route-boundary finding, could you either take it as such or name the missing row now, so it does not come back as a third round of the same item. And if you are satisfied, could you leave the review as an explicit state: everything so far on this PR has been `COMMENTED`, so it carries no decision and cannot merge on your review alone.

Deferred, and I will file it rather than leave it in a thread: the `Last-Modified` shared-object write-timing signal you found in round 2 is documented as an accepted tradeoff, and the byte layer that would remove it (per-entry object keys, or never rewriting an existing object) is a design change beyond this PR.

cosarah:
P2 — Propagate egress configuration through the reference server\n\n`createStorageHttpHandler` accepts and forwards `byteEgress` and `signedUrlTtlSeconds`, but `createReferenceRuntimeServer` does not expose or forward either option. A consumer using the documented reference-server factory with an S3-backed `assetStore` therefore cannot enable redirect egress; reads silently remain direct unless the consumer bypasses the factory and composes the handler manually. Please add the options to `ReferenceRuntimeServerOptions`, forward them into `handlerOptions`, and cover the path with a test.

cosarah:
P3 — Exercise the browser CORS boundary for signed fetches\n\nThe current `HttpAssetStore` tests use Node fetch doubles, so they do not enforce browser CORS, preflight, credentials, or cross-origin response-reading behavior. With redirect egress enabled and the object store on another origin, a missing `Access-Control-Allow-Origin`/preflight configuration can make the descriptor or signed-byte fetch fail in a real browser while all unit tests remain green. Please add a browser-level or Playwright integration test covering the cross-origin descriptor and signed-byte fetch, or otherwise add an equivalent deployment-boundary check.

wyuc:
The P2 is addressed in `7632cac3`: `ReferenceRuntimeServerOptions` now accepts `byteEgress` — in the current `'direct' | { mode: 'redirect'; collectionGraceMs; signedUrlTtlSeconds? }` shape, not the retired flat field — and forwards it into the storage handler options alongside every other forwarded option. Two tests drive the composed reference server end to end: redirect config yields `302` + `Location` through the factory, and omission keeps direct `200` bytes. Verified by mutation: dropping the forwarding line turns the redirect test red.

The P3 I am not taking into this PR, and I want to be plain about why. Your 08-13 triage dismissed this exact item: *"Browser coverage would be useful, but this is an environment-specific long-tail test gap rather than a package defect."* Nothing on the code side has changed since that dismissal — the client still fetches signed URLs with `credentials: 'omit'` and no deployment headers, and object-store CORS remains explicitly deployment-owned by the contract. A real browser-level cross-origin test is new e2e infrastructure, not a diff-local fix: I attempted one locally, and `page.route` interception fulfills responses without producing CORS preflights, so an honest test requires standing up real separate origins. That belongs in its own change, so I have filed it as #1122 rather than grow this PR's surface again.

Which brings me to the process point, made once more and now directly. This PR has had eleven self-review rounds and four review rounds on your side. Every round since 08-13 has produced only new-dimension P2/P3 items — no previously fixed finding has regressed — and every KEEP item has been addressed within one round. The finding stream needs a closure criterion. If defects remain, please name the complete list now, including anything held back as minor, and I will address all of it in one final round. Otherwise, please leave an explicit review decision: every review so far has been COMMENTED, which carries no decision and cannot merge.

cosarah:
## Triage verdict and priority

I reviewed all five discovery findings against PR `HEAD` (`7632cac`) and their stated trigger conditions. All five remain **KEEP**, but they are not equally urgent.

### P1 — Fix before merge

1. **Preserve service-level 404 failures from signed reads — KEEP / P1**
   At PR HEAD, every 404 from the signed object-store fetch is converted to `null`, while the direct S3 path treats only `NoSuchKey` as an asset miss. A wrong bucket, access point, or endpoint can therefore make a live asset appear absent. This is a concrete correctness and outage-masking bug.

### P2 — Address before enabling or relying on redirect egress

2. **Fail closed when descriptor negotiation returns a redirect — KEEP / P2**
   The authenticated descriptor GET leaves fetch redirect handling at its default. If negotiation is lost or ignored, a cross-origin 302 can forward custom headers such as `x-learner-key` when the destination CORS policy admits them. The trigger is long-tail, but the impact is a credential-disclosure failure at the boundary this protocol is meant to protect.

3. **Document the deployment switch and storage prerequisites — KEEP / P2**
   The bundled route reads `ASSET_BYTE_EGRESS`, but `.env.example` and the documented server-persistence deployment path do not expose the switch or point operators to the required S3 CORS and `s3:ListBucket` configuration. The feature is opt-in, so this is primarily a deployment-readiness and operator-error issue rather than a default-path regression.

### P3 — Low-priority follow-up, not a current runtime blocker

4. **Export the descriptor media type from the package — KEEP / P3**
   `ASSET_DESCRIPTOR_MEDIA_TYPE` is the public descriptor protocol discriminator, but it is not exported from the package root or a supported subpath. External protocol consumers must duplicate the literal. This is API completeness debt with a small, local fix.

5. **Exercise the signed fetch in a real browser — KEEP / P3**
   Redirect egress is a browser cold-read path when enabled, but current tests use Node and mocked `fetch`; they cannot exercise cross-origin CORS, preflights, redirect behavior, or response-header exposure. This is a meaningful test-coverage gap, but no production failure is demonstrated by the finding itself.

**Priority summary:** 1 is the merge-blocking correctness issue; 2 and 3 should be handled before operators enable redirect egress; 4 and 5 are worthwhile follow-ups.

**KEEP set:** 1, 2, 3, 4, 5.

wyuc:
Findings 1–4 are addressed in `0ea552c5`:

1. **Signed-read 404 semantics.** The signed fetch now reads the object store's error body and maps only a declared `NoSuchKey` to a miss (mirroring the direct S3 path's `isNotFound`). Any other 404 — a different XML code like `NoSuchBucket`, a plain-text or empty body, an unreadable body — fails loud as `MALFORMED_RESPONSE`, so a wrong bucket/endpoint can no longer make a live asset read as absent. The contract text now states that only a declared `NoSuchKey` may read as a miss.
2. **Descriptor negotiation fails closed.** Every byte GET goes out with `redirect: 'manual'`, and any 3xx (or a browser's `opaqueredirect`) is rejected before the target is ever fetched — the deployment headers on that GET are never forwarded anywhere. The test pins that the redirect target is not called.
3. **Deployment docs.** `ASSET_BYTE_EGRESS` is in `.env.example` (default `direct`), and the README's server-persistence section now states the two prerequisites for redirect mode: object-store CORS admitting the app origin, and `s3:ListBucket` on the signing identity so a missing key answers `404 NoSuchKey` rather than `403`.
4. **Descriptor media type exported.** `ASSET_DESCRIPTOR_MEDIA_TYPE` is now exported from the package root and re-exported from the `./asset/http` subpath, with an export-surface test.

Finding 5 is the item I am keeping out of this PR, now for the third round of it being raised: it was dismissed in your 08-13 triage as "an environment-specific long-tail test gap rather than a package defect," and it is tracked as #1122. Nothing about the code has changed in between. I would rather ship the browser-level CORS coverage as that focused follow-up than keep it gating this PR.

And the standing request, third time: I have now closed every actionable finding from every round within one round each. If more remain, please post the complete list — I will address all of it in a single final round. Otherwise please leave an explicit review decision; a chain of COMMENTED reviews cannot merge.

cosarah (approved):
LGTM

PR #1101, opened 2026-08-11
feat(media): convert legacy references to allocated asset ids (#1007 part 2c)

wyuc:
Lands epic #1007 part 2(c) as one delivery unit: the app-side reference converter, the `audioUrl` removal, and the `DSL_VERSION` bump — inseparable per the epic, because only a non-pure converter can read local bytes and probe whether a URL is still reachable.

## The converter

`convertDocumentAssetRefs` walks a loaded document (slide canvases, whiteboards, video-manifest keys, speech actions) and rewrites legacy handles to allocated pool ids:

- **Placeholder with local bytes** (`gen_img_*` / `gen_vid_*` with a healthy `mediaFiles` row): ingest, rewrite; one logical ref allocates one asset across every slot and manifest key naming it.
- **`audioId` with local bytes**: ingest and rewrite, keeping 2(b)'s deliberate Dexie double-write.
- **Co-present `audioUrl` + `audioId`** collapses to one asset: from local bytes when they exist (the URL is never fetched), from a fetched URL when the `audioId` dangles.
- **A dead URL** (definitive 4xx) converts to no asset and an emptied reference — a dead URL is never carried into the new format. A transient failure (5xx, network) keeps both handles for a later retry.
- **Bytes unavailable** keeps the legacy references untouched; the document converts on a later open.
- **Idempotent**: a converted document returns by identity with no writes and no allocations.

Conversion runs lazily on document open, inside the existing document migration lock (both the existing-destination and legacy-snapshot branches), and at fetch time for server-generated classrooms. No batch job.

## `audioUrl` removal and the version bump

`SpeechAction.audioUrl` is gone from the DSL; every consumer (playback, actions edit surface, slide defaults, zip manifest) reads the converted shape. The server classroom generator keeps emitting its derived-id + serving-URL pair as an explicitly typed pre-conversion transport, consumed at fetch time before persistence.

`DSL_VERSION` becomes `0.2.0` with the first real ladder entry. The pure transform handles only the one case it can decide without destroying a live handle: a URL with no `audioId` beside it is dropped (the same outcome the converter produces for a dead URL), while a co-present pair is kept verbatim for the converter. `@openmaic/dsl` goes to 0.9.0 per the release rule.

## Tests

16 converter tests (every rule above, plus one-asset-per-ref and non-mutation), 4 convert-on-open wiring tests, 5 ladder-fixture tests. Full suites green except two pre-existing timing-flaky files and one pre-existing multipart assertion, all verified failing on the clean base.

Refs #1007


wyuc:
Codex review loop has converged: 10 rounds, findings through round 9, and round 10's single P1 did not survive verification. What the rounds changed, by theme:

- **Never destroy a possibly live handle.** The ladder entry is a pure stamp: dropping URL-only `audioUrl` fields there would destroy narration before the converter could probe it, since the ladder runs on every read first. The converter is the only component that removes `audioUrl` — ingesting bytes, collapsing pairs, or emptying references only on definitive death (404/410; 408/429/401/403 stay retryable).
- **Conversion is best-effort everywhere.** The legacy URL stays a live fallback: playback (including direct `Audio` src for CORS-less cross-origin URLs), editor status and preview, slide duplication and text edits clearing both handles, and ZIP export fetching the bytes into the archive rather than dropping or leaking the field.
- **Retry idempotence.** Converted media and audio rows mirror under the allocated id (export/import round-trips keep working); retries reuse those mirrors (media via `placeholderRef`, audio via `originAudioId`/`originAudioUrl`) instead of orphaning a fresh allocation per attempt, and a failed compatibility write compensates its allocation.
- **Bounded waits and cancellation.** Speech conversion runs through one cross-scene worker pool (4) with per-probe (15s) and aggregate (60s) budgets; liveness is rechecked at each commit boundary so a superseded classroom load stops producing side effects instead of completing them; a save-back failure no longer fails the document read.

On round 10's P1 (legacy `audioUrl` rejected by strict validation before conversion): the claimed failure does not exist on the document open path. `assertValidDestination` runs the app validators (`validateAppStage`/`validateAppScene` → the DSL's `checkAction`), which check required fields and tolerate unknown ones — verified directly: `validateAction` and `validateScene` both return `{"valid":true}` for a speech action carrying `audioUrl`. The `additionalProperties: false` strictness lives only in the generated `action.schema.json`, which external consumers use and nothing on the load or save path enforces. A document that retains the field as inert data until the converter reaches it may fail that external strict validation in the meantime — that is the tradeoff the ladder comment documents, accepted deliberately over dropping a live handle.

Local suites are green except two timing-flaky files and one multipart-message assertion, all verified failing identically on the base commit.

lib/media/convert-legacy-asset-refs.ts
```diff
@@ -0,0 +1,662 @@
...
+            report.converted += 1;
+            return { kind: 'allocated', assetId };
+          }
+          return fetched.kind === 'dead' ? { kind: 'dead' } : { kind: 'unavailable' };
+        })();
+      urlOutcomeByRef.set(urlKey, pendingUrl);
+      const outcome = await pendingUrl;
+      if (outcome.kind === 'allocated') {
+        const next: LegacySpeechAction = { ...speech, audioId: outcome.assetId };
+        delete next.audioUrl;
```
cosarah, line 806: [P1] Preserve converted narration across asset principals

With server-backed persistence, `putAsset` stores this allocation under the device-derived `x-learner-key`, while documents are not partitioned by that key. Deleting `audioUrl` therefore leaves a document that another browser or a reminted learner key can load but whose narration asset it cannot resolve. Please keep a portable recovery reference or align document and asset ownership before dropping the URL.

lib/media/convert-legacy-asset-refs.ts
```diff
@@ -0,0 +1,662 @@
...
+
+  const convertSpeechAction = async (action: Action): Promise<Action> => {
+    assertContinuing();
+    if (action.type !== 'speech') return action;
+    const speech = action as LegacySpeechAction;
+    const audioId = speech.audioId || undefined;
+    const audioUrl = speech.audioUrl || undefined;
+    if (!audioId && !audioUrl) return action;
+
+    if (audioId && (await resolvedDeps.assetRefExists(audioId))) {
```
cosarah: [P1] Avoid downloading every audio asset during migration checks

`assetRefExists` is implemented through `resolve()`, so the HTTP store performs a full GET, buffers all bytes, and creates a Blob URL rather than issuing a metadata-only existence check. Since conversion runs on every document open, an already-converted narrated classroom re-downloads every referenced audio asset, and the HTTP request has no timeout. Please use the existing HEAD/identity path or another bounded metadata-only probe here.

lib/document-store/migration.ts
```diff
@@ -284,17 +362,40 @@ async function migrateLocked(
...
         }
         await finishMigrationMetadata(snapshot, kv);
       }
-      return { document: existing, readOnlyLegacy: false };
+      // Lazy reference conversion on open: rewrite legacy media handles to
+      // allocated asset ids and save the converted document back under the
+      // same lock. An unconverted document is returned as-is (identity), so
+      // already-converted documents pay no write.
+      const converted = await convertLoadedDocument(existing, deps);
+      await saveConvertedDocument(store, stageId, existing, converted, deps, expectedGeneration);
```
cosarah: [P1] Do not overwrite remote edits with a stale conversion save

This saves the complete aggregate loaded before conversion, which may have spent seconds probing legacy URLs. Web Locks do not coordinate independent browsers, and the HTTP document contract has no revision or conditional-write support, so another client can edit or delete the stage before this PUT and have those changes overwritten or the document recreated. Please add optimistic concurrency or persist only the converted references against a verified revision.

lib/media/convert-legacy-asset-refs.ts
```diff
@@ -0,0 +1,662 @@
...
+            ((keys.audioId !== undefined && row.originAudioId === keys.audioId) ||
+              (keys.audioUrl !== undefined && row.originAudioUrl === keys.audioUrl)),
+        )
+        .first(),
+    removeAsset: (ref) => removeAsset(ref),
+    fetchLegacyUrl: async (url) => {
+      try {
+        // Bounded wait: conversion runs on the document load path, and one
+        // stalled URL must not hold the document lock indefinitely.
+        const response = await fetch(url, { signal: AbortSignal.timeout(15_000) });
```
cosarah: [P2] Fetch legacy URLs through the CORS-safe media proxy

A cross-origin audio URL can be playable by an `<audio>` element while browser `fetch` cannot read it without CORS, so this path leaves such references unconverted indefinitely. Direct fetching also bypasses the existing proxy SSRF checks and 25 MiB response limit. Please route legacy URL ingestion through `/api/proxy-media` (with the same timeout semantics).

lib/export/use-export-classroom.ts
```diff
@@ -78,6 +79,45 @@ export function useExportClassroom() {
...
+          const stampedId = (action as SpeechAction).audioId;
+          if (stampedId && audioIdToPath.has(stampedId)) continue;
+          uniqueLegacyUrls.add(legacyUrl);
+        }
+      }
+      const fetchedLegacy = await mapWithConcurrency([...uniqueLegacyUrls], 4, async (url) => {
+        try {
+          // Bounded wait, like the load-path converter: one stalled legacy
+          // endpoint must not wedge the whole export.
+          const response = await fetch(url, { signal: AbortSignal.timeout(15_000) });
```
cosarah: [P1] Use the media proxy for legacy ZIP narration

This direct browser fetch fails for the same cross-origin URLs that can still play through an audio element. On failure no `audioUrlToPath` entry is created, and `actionsToManifest` strips `audioUrl`, so the exported classroom silently loses otherwise playable narration. Please use `createProxiedFetch` here, as this exporter already does for other CORS-locked assets.

lib/media/convert-legacy-asset-refs.ts
```diff
@@ -0,0 +1,662 @@
...
+        // A URL that no longer resolves converts to NO asset and an emptied
+        // reference: a dead URL is never carried into the new format.
+        const next: LegacySpeechAction = { ...speech };
+        delete next.audioId;
+        delete next.audioUrl;
+        changed = true;
+        report.emptied += 1;
+        return next;
+      }
+      // Transient fetch failure: keep both legacy handles and retry on a
```
cosarah, line 820: [P2] Carry this retained URL fallback into video export

Keeping the legacy handles is correct for playback, but the video-export dependency builder only inspects `audioId` and pool/Dexie records; it never consumes `audioUrl`. When this fallback is reached for a dangling or absent ID, live playback can still work while video export marks the narration missing and omits it. Please teach the video-export dependency/collection path to ingest this retained URL.

wyuc:
Four of the six findings are fixed in `e0a74368`:

- **Metadata-only existence checks** (`convert-legacy-asset-refs.ts:422`): both asset stores gained `exists()` — a bounded (15s) `HEAD` for the HTTP store, an entry lookup for the browser store — and `assetRefExists` prefers it. Migration probes no longer download every referenced asset or mint URLs they never use.
- **Proxy ingestion** (`convert-legacy-asset-refs.ts:232`, `use-export-classroom.ts:104`): legacy URL fetches in the converter and the ZIP exporter now go through `/api/proxy-media` via one shared helper (`lib/media/fetch-media-url.ts`, same 15s discipline), so CORS-locked-but-playable narration converts and exports instead of silently vanishing — and the SSRF guard and 25 MiB limit now apply to these paths.
- **Video export** (`convert-legacy-asset-refs.ts:561` thread): `timeline-deps.ts` resolves a speech action's audio id-first and retained-URL second, ingesting the URL through the same proxy, so unconverted clips no longer render missing while playing fine.

The two remaining P1s are design decisions I don't want to take silently — I'll follow up on both:

1. **Principal alignment** (`convert-legacy-asset-refs.ts:547`): documents are global while converted assets are partitioned by the minting browser's learner key. Options on the table: keep the `audioUrl` as an inert recovery reference on converted documents (other browsers keep playing through the URL fallback), accept the cross-browser gap as a documented limitation of the dev authenticator, or store converted narration under a shared principal (which gives up per-principal isolation semantics).
2. **Optimistic concurrency** (`migration.ts:387`): the narrow mitigation is reload-then-apply-reference-diffs before save, which shrinks the overwrite window from seconds (URL probing) to milliseconds; the full fix is conditional-write support in the HTTP document contract, which is contract-level work beyond this PR.

wyuc:
Both remaining P1s are now settled in `afbb5f32`:

- **Principal alignment** (`convert-legacy-asset-refs.ts:547`): converted actions keep the `audioUrl` as inert recovery data when the pool is server-backed (`retainLegacyUrl`, wired from `isAssetPoolServerBacked()`). Another browser can then still play the narration through the URL fallback — playback, preview, and video export all consume it, and ZIP export strips it after fetching the bytes. A dead URL is still dropped, and with a local pool the field is dropped as before. The principled fix (aligning document and asset ownership) belongs to a later contract change, not this PR.

- **Optimistic concurrency** (`migration.ts:387`): the save-back now reloads the document first; a fresh read that differs from what was converted gets its own conversion pass, and that reconciled document is what is saved and returned. The overwrite window shrinks from the full conversion (seconds of URL probing) to a single load+save, without touching the HTTP contract. A full conditional write remains possible later contract work. Covered by a new test where a concurrent rename lands mid-conversion and survives with the conversion applied on top.

lib/media/convert-legacy-asset-refs.ts
cosarah: [P1] This remains actionable for slide media as well as narration. retainLegacyUrl only protects speech actions; gen_img_* and gen_vid_* references are rewritten to asset IDs with no portable fallback. With server-backed persistence, browser A stores those IDs in its learner partition, while browser B can load the global document but receives ASSET_NOT_FOUND. Please align document and asset ownership, or retain a portable recovery source for converted image and video references, before treating the principal issue as resolved.

lib/document-store/migration.ts
cosarah: [P1] The reconciliation reload narrows this race but does not make the write conditional. A remote edit or deletion can land after loadDocument returns latest and before saveDocument sends its unconditional full-document PUT; that write can still overwrite the edit or recreate the deleted document. Web Locks do not coordinate independent browsers, so this needs a revision or ETag compare-and-swap, or a server-side reference-mutation contract.

packages/@openmaic/storage/src/asset/browser-store.ts
```diff
@@ -366,6 +366,19 @@ export class BrowserAssetStore implements StorageProvider {
     await this.urls.invalidate(ref);
   }
 
+  /**
+   * Metadata-only existence probe: the registry entry lookup alone, with no
+   * blob read and no URL minted. This is what migration-time checks should
+   * use; resolve downloads and pins.
+   */
+  async exists(ref: AssetRef): Promise<boolean> {
```
cosarah, line 374: [P1] Bump the published storage package version. This adds the public BrowserAssetStore.exists method and the corresponding HTTP-store behavior under a publishable package, but packages/@openmaic/storage/package.json remains at 0.2.5. The repository package-version gate compares publishable inputs against the base and rejects an unchanged version, so this PR cannot pass that CI/release check until the package version is increased.

lib/video-export-app/timeline-deps.ts
```diff
@@ -219,6 +223,25 @@ export async function createVideoTimelineDeps(input: {
     if (record) audioById.set(audioId, blob ? { ...record, blob } : record);
     else if (blob) audioById.set(audioId, { id: audioId, blob } as AudioFileRecord);
   }
+  // An unconverted document can carry narration only as a legacy URL: the
+  // playback paths fall back to it, and the export must too, or a playable
+  // clip renders as missing. Fetched bytes are keyed by the URL itself, which
+  // the lookup below prefers exactly when no id resolves.
+  await mapWithConcurrency([...legacyAudioUrls], PROBE_CONCURRENCY, async (url) => {
```
cosarah, line 264: [P2] Add an end-to-end test for the legacy-URL video-export fallback. This path now spans proxy fetching, blob insertion, duration probing, ID-versus-URL lookup, timeline planning, and final asset collection, while the current timeline dependency tests cover only media-ref bridging and geometry. Please exercise a speech action with an absent or dangling audioId plus audioUrl and verify fetched bytes, probed timing, and a present collected narration asset.

lib/classroom/load-classroom.ts
```diff
@@ -245,7 +254,33 @@ export async function fetchClassroomFromApi(classroomId: string): Promise<Classr
...
+  // the payload is applied or persisted, so the classroom is born with
+  // allocated asset ids and the raw URLs are never stored. Conversion failure
+  // degrades to the unconverted payload rather than blocking the load -- the
+  // document load path retries conversion on the next open.
+  try {
+    const { convertDocumentAssetRefs } = await import('@/lib/media/convert-legacy-asset-refs');
+    // The guard travels into the conversion: a load superseded mid-fetch
+    // stops producing side effects as soon as it is known to be unwanted,
+    // and the abort degrades to the unconverted payload like any failure.
+    const converted = await convertDocumentAssetRefs(
```
cosarah: [P2] Tie fallback conversion side effects to acceptance of the classroom payload. shouldConvert prevents work after an already-superseded load and rechecks at converter commit boundaries, but conversion can finish successfully before fallback hydration is later rejected as stale. The allocated assets and Dexie mirrors then remain without a committed document; concurrent cold loads can also allocate distinct IDs before either payload becomes authoritative. Please add rollback or an acceptance/commit boundary that owns these allocations.

lib/document-store/migration.ts
```diff
@@ -284,17 +385,47 @@ async function migrateLocked(
...
           }
         }
         await finishMigrationMetadata(snapshot, kv);
       }
-      return { document: existing, readOnlyLegacy: false };
+      // Lazy reference conversion on open: rewrite legacy media handles to
+      // allocated asset ids and save the converted document back under the
+      // same lock. An unconverted document is returned as-is (identity), so
+      // already-converted documents pay no write.
+      const converted = await convertLoadedDocument(existing, deps);
```
cosarah: [P2] Keep legacy URL probing outside the global runtime shared lock. migrateLocked holds this lock across the complete conversion pass, whose stalled URL probes can consume the 60-second aggregate budget. Whole-store maintenance uses a five-second exclusive acquisition timeout, so clearDatabase and similar operations fail predictably while an unrelated legacy URL is slow. Only the conversion commit and persistence fence need this maintenance exclusion.

tests/export/classroom-zip.test.ts
```diff
@@ -111,22 +109,23 @@ describe('actionsToManifest', () => {
     expect(result[1]).toMatchObject({ type: 'spotlight', elementId: 'el1' });
   });
 
-  test('preserves audioUrl when audioId is absent', () => {
+  test('a speech action without audioId exports no audio reference at all', () => {
```
cosarah, line 112: [P2] Cover the legacy-URL ZIP path through the real exporter. The current test only checks the pure actionsToManifest helper, so it never verifies URL collection, proxy fetching, zip-path assignment, mediaIndex insertion, or the final zip.file write. A regression at any of those boundaries would still produce a green suite while exporting a classroom with silently missing narration.

lib/media/convert-legacy-asset-refs.ts
```diff
@@ -0,0 +1,693 @@
...
+  const results: R[] = new Array(items.length);
+  let next = 0;
+  const workers = Array.from({ length: Math.min(limit, items.length) }, async () => {
+    while (next < items.length) {
+      const index = next;
+      next += 1;
+      results[index] = await fn(items[index], index);
+    }
+  });
+  await Promise.all(workers);
```
cosarah, line 147: [P2] Do not let Promise.all return while sibling conversion workers are still committing effects. When one worker rejects, the other workers continue allocating assets and writing compatibility rows after the caller has already degraded to the original document. Retry mirrors reduce duplicate allocation later, but they do not join, cancel, or compensate the still-running batch. Please use shared cancellation plus compensation, or await all workers and reconcile every committed result before returning failure.

tests/media/convert-legacy-asset-refs.test.ts
```diff
@@ -0,0 +1,666 @@
...
+
+let allocationCounter = 0;
+
+function makeHarness() {
+  const pool = new Map<string, { blob: Blob; meta: AssetMeta }>();
+  const mediaRows = new Map<string, MediaFileRecord>();
+  const audioRows = new Map<string, AudioFileRecord>();
+  const urlFetches = new Map<string, LegacyUrlFetch>();
+
+  const deps: LegacyAssetConversionDeps = {
```
cosarah, line 28: [P2] Exercise the production conversion dependency graph. Every converter test replaces defaultDeps with in-memory maps, so the suite does not cover the real proxy request, asset-store exists calls, authentication headers, Dexie queries, or compensation against the configured stores. The asset stores have their own contract tests, but the converter wiring between these boundaries can regress while this suite stays green.

tests/classroom/load-classroom.test.ts
```diff
@@ -237,7 +237,10 @@ describe('runClassroomLoad', () => {
 
       // The warm ghost did not starve the restore: the server fallback ran,
       // the restored classroom is live, and the deletion was lifted.
-      expect(deps.fetchClassroom).toHaveBeenCalledExactlyOnceWith('stage-warm-ghost');
+      expect(deps.fetchClassroom).toHaveBeenCalledExactlyOnceWith(
```
cosarah, line 247: [P2] Test fetchClassroomFromApi rather than only an injected fetchClassroom mock. This assertion proves that isCurrent is forwarded, but it never executes the API JSON boundary, dynamic converter import, conversion-error degradation, or cancellation after converter side effects. Please add a focused test around the real exported function so the only server-transport conversion path has behavioral coverage.

cosarah:
Triage follow-up: the remaining actionable findings are ordered below by impact.

**Address first**

1. Cross-principal loss of converted slide image/video assets: https://github.com/THU-MAIC/OpenMAIC/pull/1101#discussion_r3763871387
2. Conversion save-back can still overwrite or resurrect remote documents without conditional writes: https://github.com/THU-MAIC/OpenMAIC/pull/1101#discussion_r3763871461
3. The publishable `@openmaic/storage` changes require a package version bump: https://github.com/THU-MAIC/OpenMAIC/pull/1101#discussion_r3763871545
4. Add end-to-end coverage for the legacy-URL video-export fallback: https://github.com/THU-MAIC/OpenMAIC/pull/1101#discussion_r3763871634

**Recommended in this PR**

5. Tie server-fallback conversion side effects to payload acceptance or rollback: https://github.com/THU-MAIC/OpenMAIC/pull/1101#discussion_r3763871712
6. Move slow URL probing outside the global runtime maintenance lock: https://github.com/THU-MAIC/OpenMAIC/pull/1101#discussion_r3763875053
7. Cover legacy-URL ZIP export through the real exporter and archive contents: https://github.com/THU-MAIC/OpenMAIC/pull/1101#discussion_r3763875106

**Hardening / lower priority**

8. Join, cancel, or compensate sibling conversion workers after a batch failure: https://github.com/THU-MAIC/OpenMAIC/pull/1101#discussion_r3763875175
9. Exercise the converter production dependency wiring rather than only in-memory maps: https://github.com/THU-MAIC/OpenMAIC/pull/1101#discussion_r3763875261
10. Test the real `fetchClassroomFromApi` conversion and degradation path: https://github.com/THU-MAIC/OpenMAIC/pull/1101#discussion_r3763875351

The previously reported localhost/private-host proxy rejection is not included because it was dismissed during trigger triage.

wyuc:
All ten triage items are addressed in `ff2b8cf5`, except the two that are design positions rather than defects — those get my reasoning below.

**By your ordering:**

1. **Cross-principal media loss** — this one I read differently, and want to lay out why before you spend more cycles on it. In the legacy shape, slide media references (`gen_img_*` / `gen_vid_*`) resolve from the browser-local `mediaFiles` Dexie table only; there is no portable handle to retain — no URL, no server copy. So for browser B, a legacy classroom's media was *already* unresolvable before conversion (no Dexie row), exactly as it is after conversion (`ASSET_NOT_FOUND` from B's partition). Conversion is cross-browser-neutral for media and strictly improves the minting browser (durable server-side bytes). The asymmetry you rightly called out for narration existed for narration only because the URL made it portable; no equivalent exists for media. The real fix is the ownership alignment you named, and I agree it belongs to the contract work — tracked as the follow-up below rather than this PR.

2. **Conditional writes** — agreed the reload narrows rather than eliminates the race; that was the explicit scope decision for this PR (the residual window is now one load+save, not seconds of URL probing). A revision/ETag compare-and-swap belongs in the HTTP document contract and would balloon this diff past reviewability. I'll file the follow-up issue covering both this and the ownership alignment, since they share the contract layer.

3. **Version bump** — done, `@openmaic/storage` is 0.3.0.
4. **Video-export e2e** — added: dangling pair + legacy URL now exercised through proxy fetch, blob insertion, duration probe, id-vs-URL lookup, and the collected asset comes out present (`tests/video-export/timeline-deps.test.ts`).
5. **Rollback boundary** — done: the payload carries its fresh allocations, and both discard points (superseded mid-conversion, stale hydration) roll them back; the abort error carries the list.
6. **Lock scope** — done: probing now runs outside the global runtime shared lock; only fence-checked commits hold it.
7. **ZIP export coverage** — the legacy-URL collection is extracted into `collectLegacyAudioForExport` and tested through the real helpers: URL collection, proxy fetch, zip-path assignment, manifest mapping (`tests/export/classroom-zip-legacy-audio.test.ts`). The hook plumbing and the final `zip.file` write remain thin and uncovered, noted.
8. **Worker cancellation** — done: the pool cancels scheduling on first rejection and joins all workers before propagating.
9. **Production wiring coverage** — added: the converter runs against real Dexie + the real browser pool + the real proxy request shape, network stubbed only (`tests/media/convert-legacy-asset-refs-wiring.test.ts`).
10. **Real `fetchClassroomFromApi`** — added: JSON boundary, degradation on converter failure, and abort rollback (`tests/classroom/fetch-classroom-from-api.test.ts`).

lib/media/convert-legacy-asset-refs.ts
```diff
@@ -0,0 +1,726 @@
...
+
+  const convertSpeechAction = async (action: Action): Promise<Action> => {
+    assertContinuing();
+    if (action.type !== 'speech') return action;
+    const speech = action as LegacySpeechAction;
+    const audioId = speech.audioId || undefined;
+    const audioUrl = speech.audioUrl || undefined;
+    if (!audioId && !audioUrl) return action;
+
+    if (audioId && (await resolvedDeps.assetRefExists(audioId))) {
```
cosarah: [P2] Stop probing every converted narration on every document access

Conversion runs from both `accessDocument` and `mutateDocument`, and every retained converted speech action reaches this `assetRefExists` call. With `HttpAssetStore`, that is one network `HEAD` per action on every open and mutation, including repeated IDs, while the document lock is held. There is no conversion-complete marker or pass-level cache, so the cost remains O(number of speech actions) indefinitely. Please add a durable completion signal and at least deduplicate existence checks within a pass.

lib/classroom/load-classroom.ts
```diff
@@ -245,7 +275,46 @@ export async function fetchClassroomFromApi(classroomId: string): Promise<Classr
...
+      { ...json.classroom },
+      undefined,
+      shouldConvert,
+    );
+    return {
+      stage: converted.document.stage,
+      scenes: converted.document.scenes,
+      allocatedAssetIds: converted.allocatedIds,
+    };
+  } catch (error) {
```
cosarah: [P2] Compensate allocations when any conversion worker fails

Only `LegacyConversionAbortedError` exposes and rolls back the pass allocation ledger. If one conversion worker fails normally after another worker has allocated and mirrored data, the converter throws an ordinary error and this catch returns the original payload, leaving the successful sibling's side effects unattached. Joining workers prevents late commits but does not compensate commits that already completed. Please roll back the pass ledger for every failed conversion, not only liveness aborts.

lib/video-export-app/timeline-deps.ts
```diff
@@ -203,11 +204,14 @@ export async function createVideoTimelineDeps(input: {
...
   for (const scene of scenes) {
     for (const action of scene.actions ?? []) {
-      if (action.type === 'speech' && (action as SpeechAction).audioId) {
-        audioIds.add((action as SpeechAction).audioId!);
-      }
+      if (action.type !== 'speech') continue;
+      const speech = action as SpeechAction;
+      if (speech.audioId) audioIds.add(speech.audioId);
+      const legacyUrl = (speech as { audioUrl?: string }).audioUrl;
+      if (legacyUrl) legacyAudioUrls.add(legacyUrl);
```
cosarah: [P2] Fetch retained URLs only when the asset ID misses

This collects every retained legacy URL before resolving the corresponding `audioId`. Server-backed conversion deliberately keeps these URLs, so a normally converted export downloads narration from both the asset store and the original URL. A stalled or expired origin can add repeated 15-second waves even when all required bytes already resolved by ID. Please collect/fetch a URL only after its action's `audioId` failed to produce a usable blob.

lib/classroom/load-classroom.ts
```diff
@@ -125,18 +135,35 @@ export async function runClassroomLoad<TMediaTasks = unknown>({
...
+      // payload but never converts it.
+      const classroom = await fetchClassroom(classroomId, isCurrent);
       if (!isCurrent()) return;
 
       if (classroom) {
         const { stage, scenes } = classroom;
+        // A discarded payload must not keep what its conversion allocated:
+        // roll the fresh allocations back whenever the load is superseded or
+        // the hydration rejects the payload as stale.
+        const rollbackPayloadAllocations = async () => {
```
cosarah: [P2] Roll back compatibility rows with a discarded payload

This rollback removes only pool allocations, but conversion has already written full blobs to the stage-owned Dexie `audioFiles` and `mediaFiles` compatibility tables. When hydration rejects or supersedes the payload, those rows remain even though the converted document was never accepted, consuming local quota and leaving stale recovery metadata. The conversion result needs a compatibility-write ledger, or the fallback flow needs a transaction/cleanup routine that rolls back both storage layers.

lib/media/convert-legacy-asset-refs.ts
cosarah: [P2] Avoid cross-principal narration reallocation

With the HTTP asset store, this existence check is scoped to the current learner principal. Browser B therefore sees browser A's converted narration ID as missing, follows the retained URL, allocates a new B-owned asset, and saves that private ID into the global document. Alternating browsers can repeatedly replace the reference and consume quota with abandoned allocations. The shared document should not be rewritten to a reader-private asset ID; asset ownership or the migration representation needs to be portable across authorized document readers.

lib/media/fetch-media-url.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Fetch a media URL through the same-origin media proxy when it is remote.
+ * A plain browser fetch is CORS-blocked for cross-origin media exactly where
+ * a media element would still play, and the proxy carries the SSRF guard and
+ * its response limit. Local schemes (data:, relative) go direct. Always
+ * bounded; the caller maps the response.
+ */
+export function fetchMediaUrl(url: string, timeoutMs: number): Promise<Response> {
+  if (url.startsWith('http://') || url.startsWith('https://')) {
```
cosarah, line 13: [P1] Fetch same-origin classroom media directly

This proxies every absolute HTTP(S) URL, including absolute `/api/classroom-media/...` URLs generated by this application. On localhost and private-network deployments, `/api/proxy-media` rejects that target through the SSRF guard unless `ALLOW_LOCAL_NETWORKS` is enabled, even though a direct same-origin browser request works. This breaks conversion and export for a common self-hosted setup. Please detect same-origin URLs and fetch them directly, using the proxy only for cross-origin URLs.

lib/classroom/load-classroom.ts
```diff
@@ -125,18 +135,35 @@ export async function runClassroomLoad<TMediaTasks = unknown>({
 
     if (!getCurrentStage()) {
       log.info('No IndexedDB data, trying server-side storage for:', classroomId);
-      const classroom = await fetchClassroom(classroomId);
+      // Conversion has pool and Dexie side effects, so it runs only while
+      // this load is still the current one; a superseded load fetches the
+      // payload but never converts it.
+      const classroom = await fetchClassroom(classroomId, isCurrent);
       if (!isCurrent()) return;
```
cosarah, line 137: [P2] Roll back a converted payload rejected immediately after fetch

This currentness check returns before the payload rollback helper is constructed. There is a real, narrow window where conversion passes its final liveness check, awaits its final compatibility write, returns fresh `allocatedAssetIds`, and the load becomes stale before this line executes. The payload is then discarded without compensation. Please establish rollback ownership before this check, or have `fetchClassroomFromApi` perform a final acceptance check and compensate its completed conversion.

wyuc:
Round 3 findings are all addressed in `cf65ed9d`:

- **Per-open probing** (`convert:475`): existence checks are memoized per pass, and allocation-shaped ids are trusted without a probe — an already-converted classroom no longer pays one HEAD per narration on every open. The prefix is used as an allocation-shape heuristic only, not validation (the id domain stays deliberately opaque); a false negative would cost one probe, and legacy ids never take that shape.
- **Failure compensation** (`load-classroom:307`): the pass now shares its allocation ledger with the caller via an out-param, so rollback happens on every failure mode, not just liveness aborts.
- **Fetch on miss only** (`timeline-deps:214`): a retained URL is fetched only when the action's id produced no usable bytes; a normally converted export no longer double-downloads.
- **Dexie rollback** (`load-classroom:149`): rollback covers the compatibility rows too (audio by allocated id, media by the stage-scoped key), centralized in `rollbackConvertedAllocations`.
- **Cross-principal ping-pong** (`convert:475`): in retain-URL mode, a pool-missing allocated id with a retained URL is left exactly as-is — no rewrite to a reader-private id, no reallocation by each reader. The owner converts once; every other browser plays through the URL fallback.
- **Same-origin direct fetch** (`fetch-media-url.ts:9`): same-origin absolute URLs bypass the proxy (its SSRF guard would reject a self-hosted deployment's own media routes); the proxy is used only for genuinely cross-origin URLs.
- **Rollback timing** (`load-classroom:142`): rollback ownership is established before the first discard point, closing the window between the converter's last liveness check and the early return.

cosarah:
CI diagnosis for the failing `Lint, Typecheck & Unit Tests` job:

1. **Media test assertion is stale after the new metadata-only existence path**

   `tests/media/media-orchestrator.test.ts:656` expects `pool.release(assetId)` during an exclusively-owned retry. In the current implementation, `retryMediaTask()` calls `assetRefExists()`, and `lib/media/use-asset-url.ts` now uses the asset store's `exists()` API when available. `BrowserAssetStore.exists()` performs only a registry lookup and does not create an object-URL lease, so there is no corresponding `release()` call. The replacement itself passes; the failing assertion is left over from the previous `withAssetUrl()` path.

2. **Lazy migration races with `clearDatabase()` after reads moved outside the shared lock**

   The two failures in `tests/runtime/database-chat-cutover.test.ts:656` and `:712`, plus the unhandled `DatabaseClosedError`, come from the same lock-boundary change. `migrateLocked()` now performs `store.loadDocument()`, legacy-store reads, and conversion before entering `withRuntimeStorageSharedLock()`; only the commit is shared-lock protected.

   Consequently, a migration can read the old document before `clearDatabase()` deletes it, then later return that stale snapshot after the clear (the test expected `null`). In the save-race case, the migration can continue into the legacy Dexie read while `clearDatabase()` has deleted/closed the database, so `DatabaseClosedError` is thrown before the generation fence can produce the expected “storage was cleared during the mutation” error. That rejection is also what Vitest reports as unhandled.

   This is a real concurrency regression, not a flaky assertion: the conversion work may be outside the lock, but the read snapshot must be invalidated/rechecked before returning or touching the deleted Dexie instance.

wyuc:
CI diagnosis confirmed on both counts — thanks for the precise read. Fixed in `650a4bcf`:

1. **Stale assertion** — `media-orchestrator.test.ts` now spies on `exists()` instead of `release()`. The metadata-only probe takes no lease, so there is nothing to release. (I initially misread this local failure as environmental noise before pushing; it was exactly this assertion.)

2. **Lock boundary race** — real regression, and your framing was the right one: `clearDatabase` bumps the generation *before* it deletes, so a probe read outside the lock can observe content whose epoch already moved and still match the fence. `migrateLocked` is now three phases: reads inside the shared lock (cheap), URL probing outside it (slow), and a commit phase that re-checks the generation and, on any epoch change or a `DatabaseClosedError` from a clear landing mid-read, redoes the reads inside the lock and answers from current state. Both cutover races and the unhandled rejection are covered by the existing tests, which pass again.

lib/media/convert-legacy-asset-refs.ts
```diff
@@ -0,0 +1,757 @@
...
+    return pending;
+  };
+
+  const convertSlide = async <T extends SlideLike>(slide: T): Promise<T> => {
+    assertContinuing();
+    const slots = [...slideMediaReferenceSlots(slide)];
+    const rewrites: Array<{ index: number; assetId: string }> = [];
+    for (let index = 0; index < slots.length; index += 1) {
+      const ref = slots[index].read();
+      if (!isGeneratedMediaPlaceholder(ref)) continue;
```
cosarah: [P2] Ingest server classroom media URLs before persistence

Server classroom generation replaces image and video placeholder src values with /api/classroom-media/... URLs before this converter runs. This guard only accepts gen_* placeholders, so those generated image/video bytes are never ingested and fallback hydration persists deployment-specific raw URLs despite the new reference model. Please recognize the server classroom media transport URLs, fetch their bytes, and allocate asset references before the payload is saved.

lib/document-store/migration.ts
```diff
@@ -76,6 +91,88 @@ const MARKER_PREFIX = 'document-migration:';
...
+  }
+  try {
+    // Conversion may have spent seconds probing legacy URLs, and the document
+    // lock does not coordinate independent browsers: another client can have
+    // written while we converted. Reload and reconcile instead of blindly
+    // overwriting -- a fresh read that differs from what we converted gets
+    // its own conversion pass, and that is what we save.
+    const latest = await store.loadDocument(stageId);
+    if (
+      latest &&
```
cosarah: [P1] Treat a missing latest document as a concurrent deletion

The reconciliation branch only handles a truthy latest value. With the shared HTTP document store, another browser can delete the document while this browser is doing a slow conversion; loadDocument() then returns null, this condition is skipped, and the stale converted snapshot is saved below, recreating the deleted classroom. A missing latest document must be treated as a deletion conflict and must not fall through to saveDocument(converted).

lib/document-store/migration.ts
```diff
@@ -265,45 +366,129 @@ async function migrateLocked(
...
-    if ((await readGeneration(deps.kv)) !== expectedGeneration) {
-      throw new DocumentStorageGenerationChangedError(stageId);
+  if (probe !== null && probe.existing === null && probe.snapshot === null) {
+    return { document: null, readOnlyLegacy: false };
+  }
+
+  // Phase 2, no lock: the expensive part. URL probing can consume the full
+  // aggregate budget, and a shared lock held that long fails clearDatabase's
+  // five-second exclusive acquisition on an unrelated slow URL.
+  const converted = probe?.existing ? await convertLoadedDocument(probe.existing, deps) : null;
```
cosarah: [P1] Fence conversion side effects against database clearing

This phase intentionally runs outside the runtime maintenance lock, but conversion performs durable side effects before phase 3 checks the generation: it allocates pool entries and writes Dexie compatibility rows. If clearDatabase() completes while a legacy URL fetch is pending, the subsequent allocation can recreate the cleared side stores; phase 3 then notices the generation change and returns no document, leaving the media behind after a successful user-requested wipe. Please bind these side effects to the storage generation or keep a ledger and roll them back when phase 3 rejects the converted result.

lib/media/convert-legacy-asset-refs.ts
```diff
@@ -0,0 +1,757 @@
...
+      } catch {
+        return { kind: 'unavailable' };
+      }
+    },
+  };
+}
+
+/** A media row is a usable byte source only when it holds real bytes. */
+function usableMediaRecord(record: MediaFileRecord | undefined): record is MediaFileRecord {
+  return !!record && !record.error && !!record.blob && record.blob.size > 0;
```
cosarah: [P2] Convert CDN-backed legacy rows when local blobs are absent

mediaFiles and audioFiles explicitly support ossKey as the live byte source after local eviction, and existing export code already fetches it. This predicate rejects every zero-length local blob, and the audio branch has the same blob.size > 0 requirement, so valid CDN-backed legacy references remain unconverted forever. Please fetch ossKey when the resident blob is empty and only keep the legacy reference when neither source is available.

lib/document-store/migration.ts
```diff
@@ -76,6 +91,88 @@ const MARKER_PREFIX = 'document-migration:';
...
+    // its own conversion pass, and that is what we save.
+    const latest = await store.loadDocument(stageId);
+    if (
+      latest &&
+      !isEqual(
+        omitUndefinedObjectMembers(stripDocument(latest)),
+        omitUndefinedObjectMembers(stripDocument(existing)),
+      )
+    ) {
+      const reConverted = await convertLoadedDocument(latest, deps);
```
cosarah: [P2] Roll back allocations discarded during reconciliation

The first conversion pass may already have allocated pool entries and written compatibility rows before this reload discovers a newer document. If the newer document removed or replaced those legacy references, reConverted does not use the first pass allocations, but nothing reclaims them. Track the first pass allocation ledger and remove entries not referenced by the reconciled document, otherwise an ordinary concurrent edit leaves stage-owned orphan rows and pool quota behind.

wyuc:
Round 4 findings are all addressed in `27e48c4d`:

- **Missing latest = concurrent deletion** (`migration.ts`): the reload now treats a null `latest` as a deletion and never falls through to `saveDocument` — the wiped classroom stays wiped, and the caller observes null. Covered by a new test where the delete lands mid-conversion and the persisted document stays untouched.
- **Side effects fenced against clearing** (`migration.ts`): every conversion pass writes into a shared ledger, and phase 3 reconciles it against the document actually returned — a null answer (deletion or epoch change) releases every entry, a reconciled or redone document keeps only what it references. A clear completing mid-conversion no longer leaves pool entries or Dexie rows behind.
- **Orphan drop on reconcile** (`migration.ts`): same mechanism — first-pass allocations not referenced by the reconciled document are rolled back, computed via `collectStageAssetRefs`.
- **Classroom-media URL ingestion** (`convert:547`): slide slots and video-manifest keys shaped like `/api/classroom-media/...` are fetched (same-origin direct) and allocated before the payload persists, with a compatibility mirror like the placeholder path, so deployment-specific raw URLs never reach the document store.
- **ossKey byte source** (`convert:168`): media and audio rows with an evicted local blob convert through their `ossKey` (the live byte source export already honors); only rows with neither blob nor ossKey stay legacy.

lib/document-store/migration.ts
```diff
@@ -265,45 +374,168 @@ async function migrateLocked(
...
+        );
+        await cleanup(settled);
+        return { document: settled, readOnlyLegacy: false };
+      }
+      const freshSnapshot = await getLegacyDocumentStore(deps).read(stageId);
+      if (!freshSnapshot) {
+        await cleanup(null);
+        return { document: null, readOnlyLegacy: false };
+      }
+      const freshExpected = await convertLoadedDocument(
```
cosarah, line 556: [P2] Roll back allocations when legacy document creation fails

The legacy-only creation path allocates assets and writes compatibility rows before the first destination save, but saveDocument and verification are not covered by compensation. Quota pressure or a transient backend failure on this initial write leaves pool entries and mirror rows with no destination document referencing them. Please wrap the birth transaction in the same ownership ledger and rollback discipline used by the other conversion exits.

lib/media/convert-legacy-asset-refs.ts
```diff
@@ -0,0 +1,860 @@
...
+        // a mirror-write failure.
+        assertContinuing();
+        // Mirror the row under the allocated id, like the audio path and the
+        // generation write path: collectMediaFiles derives export references
+        // from row keys, so without the copy an exported manifest would name
+        // media the ZIP only knows by the old placeholder. The original ref
+        // stays on placeholderRef for reload reconciliation.
+        await resolvedDeps.putMediaRecord(stageId, assetId, {
+          ...record,
+          placeholderRef: record.placeholderRef ?? ref,
```
cosarah, line 506: [P1] Keep converted asset references portable across learners

With server-backed persistence, documents are shared but asset entries are partitioned by the learner key that minted them. Rewriting local or ID-only media to a newly allocated ID removes the only portable recovery source, so a second browser with another learner key can load the document but cannot resolve its image, video, or narration assets. The compatibility mirror here is local to browser A and does not help browser B. Please retain a portable source or align document and asset ownership before persisting the rewrite.

lib/export/classroom-zip-utils.ts
```diff
@@ -81,20 +83,85 @@ export async function collectMediaFiles(stageId: string): Promise<CollectedMedia
...
+      const legacyUrl = (action as { audioUrl?: string }).audioUrl;
+      if (!legacyUrl) continue;
+      const stampedId = (action as SpeechAction).audioId;
+      if (stampedId && audioIdToPath.has(stampedId)) continue;
+      uniqueLegacyUrls.add(legacyUrl);
+    }
+  }
+  const blobs: LegacyAudioBlob[] = [];
+  const audioUrlToPath = new Map<string, string>();
+  const fetched = await mapWithConcurrency([...uniqueLegacyUrls], 4, async (url) => {
```
cosarah, line 144: [P2] Cap the aggregate legacy URL probe budget

Concurrency is limited to four, but total work is not: each worker can process another 15-second timeout repeatedly. A classroom with many unique stalled legacy narration URLs therefore waits about ceil(urlCount / 4) * 15 seconds, turning best-effort recovery into a multi-minute ZIP export stall; the video timeline path has the same shape. Please add a shared wall-clock deadline or abort the remaining probes after a bounded aggregate budget.

lib/document-store/migration.ts
```diff
@@ -76,6 +92,95 @@ const MARKER_PREFIX = 'document-migration:';
...
+      // effects back.
+      return null;
+    }
+    if (
+      !isEqual(
+        omitUndefinedObjectMembers(stripDocument(latest)),
+        omitUndefinedObjectMembers(stripDocument(existing)),
+      )
+    ) {
+      const reConverted = await convertLoadedDocument(latest, deps);
```
cosarah: [P2] Track reconciliation allocations in the cleanup ledger

The reconciliation conversion is called without passLedger. Any assets allocated while converting latest are therefore invisible to cleanup; if saveDocument(reConverted) then fails, the method returns converted in-memory state while storage remains legacy and the new allocations are never compensated. Please pass the current ledger into this conversion and clean up allocations not retained by a successful persisted result.

lib/classroom/load-classroom.ts
```diff
@@ -245,7 +299,45 @@ export async function fetchClassroomFromApi(classroomId: string): Promise<Classr
...
+  const { convertDocumentAssetRefs } = await import('@/lib/media/convert-legacy-asset-refs');
+  // The pass shares its allocation ledger with us, so every failure mode --
+  // a liveness abort or an ordinary worker error -- rolls back what it
+  // committed, not just the aborts.
+  const allocated: string[] = [];
+  try {
+    // The guard travels into the conversion: a load superseded mid-fetch
+    // stops producing side effects as soon as it is known to be unwanted,
+    // and the abort degrades to the unconverted payload like any failure.
+    const converted = await convertDocumentAssetRefs(
```
cosarah: [P1] Serialize server fallback conversion

This conversion allocates pool assets and writes compatibility rows before fallback persistence enters the document mutation lock. Two cold tabs can therefore fetch and convert the same server classroom concurrently, mint different IDs, and race to persist competing documents; the losing conversion leaves orphaned allocations and the winner depends on last-writer timing. Please move conversion and fallback creation under one cross-realm commit boundary, or otherwise serialize and compensate the losing pass.

lib/classroom/load-classroom.ts
```diff
@@ -125,18 +156,38 @@ export async function runClassroomLoad<TMediaTasks = unknown>({
...
+        if (classroom) {
+          await rollbackConvertedAllocations(classroom.stage.id, classroom.allocatedAssetIds ?? []);
+        }
+      };
+      if (!isCurrent()) {
+        await rollbackPayload();
+        return;
+      }
 
       if (classroom) {
```
cosarah, line 139: [P1] Transfer allocation ownership before rollback

After applyFallbackScenes succeeds, applyClassroomStageAndScenes has already started saveToStorage without awaiting it. If this load becomes stale immediately afterward, the rollback path can delete the freshly allocated assets while that detached save is still able to persist a document referring to them. Please await the persistence handoff or explicitly transfer ownership before stale-load compensation can run.

tests/classroom/fetch-classroom-from-api.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  );
+}
+
+afterEach(() => {
+  vi.clearAllMocks();
+  vi.unstubAllGlobals();
+});
+
+describe('fetchClassroomFromApi', () => {
+  it('returns the converted payload with its fresh allocation list', async () => {
```
cosarah: [P2] Exercise server media conversion through the API boundary

This boundary test replaces convertDocumentAssetRefs with a mock and uses an empty scene list, while the production-wiring test only covers dangling speech audio. Regressions in /api/classroom-media image and video detection, byte mirroring, video-manifest re-keying, or persistence of the converted payload would still pass. Please add an end-to-end boundary case with server image/video URLs and assert the retained persisted payload and manifest use the allocated references.

wyuc:
The recurring cross-principal theme is settled at the root in `c8c27912`, and the birth-path rollback gap with it.

**The root fix** (your `convert-legacy-asset-refs.ts` P1): you were right that converting to learner-partitioned ids removes portability, and the reason earlier fixes kept circling is that the inconsistency was never in the converter — it is in the ownership model itself. Documents are stored globally with no partition; assets were partitioned by the client-supplied learner key. So the development authenticator now issues **one shared asset principal** for every request, matching the documents' shape. Any browser that can load the document resolves its assets, which is exactly the document's existing semantics. Nothing real is given up: the client-supplied header never provided isolation (the module's own docstring says so), and the contract's principal machinery stays in place for production auth, which will derive the principal from server-controlled claims and can partition then. Runtime sessions keep the learner key, since they are genuinely per-learner state.

This also retroactively answers the media variant: with a shared principal, converted media ids resolve for every reader of the document, and the retained-URL machinery stays as a harmless belt-and-suspenders fallback.

**Birth-path rollback** (`migration.ts`): both birth commits now roll the pass ledger back if the first `saveDocument` or the verification fails, same discipline as every other conversion exit.

cosarah:
## Triage: KEEP findings

The following findings remain actionable after tracing the real production paths:

1. **Transfer asset ownership before fallback persistence** (`lib/classroom/load-classroom.ts:180-183`)
   `applyFallbackScenes` applies the converted stage and starts fire-and-forget `saveToStorage()` before the loader's currentness check. If navigation supersedes the load in that window, rollback can delete assets while the pending save still persists references to them.

2. **Serialize server fallback conversion with document commit** (`lib/classroom/load-classroom.ts:159-162`)
   Server payload conversion allocates assets before the document mutation lock/save. Two simultaneous cold loads can allocate different asset IDs and race their saves, leaving the losing allocations and compatibility rows orphaned.

4. **Pass the ledger into reconciliation conversion** (`lib/document-store/migration.ts:163-164`)
   The reconciliation conversion pass does not share the caller's allocation ledger. If its save fails, newly allocated assets/compatibility rows are untracked and cannot be cleaned up.

7. **Do not persist raw server media when conversion fails** (`lib/classroom/load-classroom.ts:335-339`)
   Converter failure falls back to the raw server payload, which is subsequently persisted. This allows deployment-specific `/api/classroom-media` URLs to enter document storage and leaves ZIP export without a compatibility row for those assets.

9. **Do not mark recovered legacy audio as missing** (`lib/export/use-export-classroom.ts:86-92`)
   Legacy URL recovery populates `audioUrlToPath`, but the missing-audio pass checks only `audioIdToPath`. A fetchable legacy URL therefore produces a valid recovered entry plus a false `missing: true` entry.

10. **Keep the `exists` fallback bounded** (`packages/@openmaic/storage/src/asset/http.ts:473-476`)
    An unclassifiable HEAD response falls back to `resolve()`, whose GET has no timeout. A stalled persistence endpoint can therefore hold migration/conversion indefinitely.

12. **Skip eager conversion for wholesale replacements** (`lib/document-store/migration.ts:437-449`)
    `mutateDocument` eagerly converts the existing document before every mutation callback, including backup restore. Assets allocated for content immediately replaced by the restore can remain orphaned.

13. **Exercise the real server-media converter at the API boundary** (`tests/classroom/fetch-classroom-from-api.test.ts:44-49`)
    The boundary test mocks `convertDocumentAssetRefs` and supplies no scenes, so regressions in real server image/video transport-URL conversion or raw-URL persistence can leave the suite green.

14. **Revoke fetched legacy-audio URLs on stop** (`lib/utils/audio-player.ts:98-109`)
    Successful legacy URL playback creates a blob URL that is revoked only on `ended` or rejected `play()`. Stopping or replacing playback before completion drops the audio element without revoking that URL, leaking each fetched narration for the page lifetime.

**KEEP set:** 1, 2, 4, 7, 9, 10, 12, 13, 14.

wyuc:
All nine KEEP findings are addressed in `596cc72c` (on top of the main merge, `71fcefe0`, which was conflict-free).

The headline change is that findings 1, 2, and 7 are closed structurally rather than per call site, because they were the same defect at three addresses: side effects used to be produced ahead of the commit decision. Conversion and the first document save now run **inside** the per-stage document lock (`mutateDocument` with a new `requireLock` option), and the ownership rule is inverted: once the document commits, it owns every allocation, and the loader never rolls anything back — the payload-carried `allocatedAssetIds` and loader-side rollback are deleted, not extended. A losing cold load finds the winner's committed document and reuses it verbatim. Conversion failure now fails the load (`null`) instead of applying or persisting the raw payload, and an incomplete conversion (any residual `/api/classroom-media/...` transport URL, checked by `containsClassroomMediaUrls`) fails the same way — raw deployment addresses can no longer reach document storage.

The rest, by your numbering:

- **4** — the reconciliation re-conversion receives the caller's pass ledger (`saveConvertedDocument` threads it at both `migrateLocked` call sites); a failed reconciled save rolls back both passes.
- **9** — missing-audio detection is now `collectMissingAudioRefs(scenes, audioIdToPath, audioUrlToPath)`: a speech action whose legacy URL was recovered travels under its own zip path and is not also flagged `missing: true`.
- **10** — `HttpAssetStore` gains `probeTimeoutMs` (default 15s); an unclassifiable HEAD falls back to a **bounded** `resolveFresh` instead of the unbounded warm `resolve()`.
- **12** — `mutateDocument` gains `mode: 'replace'`, which skips eager conversion entirely; applied to backup restore (`importDatabase`) and `useImportClassroom`, so no assets are allocated for content the restore immediately replaces.
- **13** — the API boundary test now drives the real converter end to end (real Dexie, real pool, real document store under the real lock) with scenes carrying image/video transport URLs and a dangling audio pair, asserting no raw URLs survive and compatibility rows exist.
- **14** — `AudioPlayer` tracks its current blob URL and revokes it in `stopAudioElement` (stop/replace) as well as the ended / play-reject / token-mismatch paths; pinned by four leak tests.

The closure criterion I would like to agree on is the interleaving matrix, each cell pinned by a named test:

| window | resolution |
|---|---|
| conversion worker fails (any mode) | shared pass ledger rolls back |
| load superseded before/during fetch conversion | final liveness gate inside the lock; abort rolls back; nothing commits |
| two simultaneous cold loads | same per-stage lock; loser reuses winner's document, one allocation set |
| navigation supersedes after commit | document owns allocations; no rollback |
| conversion incomplete / fails | load fails closed; nothing raw persisted |
| reconciliation save fails | caller's ledger covers the re-conversion |
| wholesale replacement (import/restore) | no eager conversion, no allocations to orphan |
| stalled persistence endpoint | every migration-time probe bounded at 15s |

If a cell is missing, name it and I will cover it in one final round. Otherwise, an explicit review state would move this forward — every review so far has been a comment, which carries no decision.

Verification: full app suite (4607 tests) and the `@openmaic/storage` suite green on Node 22; `tsc`, `eslint`, and `prettier --check` clean. The PG/S3-backed contract suites cannot run locally and are covered by the Storage Contract CI job.

cosarah:
## Final acceptance triage

I rechecked the discovery findings against the final acceptance contract and the PR HEAD (`596cc72c`). The earlier comment was too broad: retained server-backed `audioUrl` is an intentional design choice, not a blocker by itself.

### Already covered in the PR

- Core legacy conversion, compatibility mirrors, retry idempotence, and converter-side rollback.
- Per-stage locking for server fallback, allocation ledgers for the main conversion paths, and wholesale-replacement mode.
- Shared asset principal, ordinary URL/existence probe bounds, cancellation checks, and blob-URL cleanup.

### Required before approval

1. **Save-back ownership:** a failed converted-document save must not make cleanup preserve allocations that durable storage still does not reference.
2. **Speech completeness:** the transport-URL completeness check must inspect speech actions; mixed image/video success must not allow an unconverted speech transport URL to be persisted.
3. **ZIP filtering:** export only document-referenced media; legacy compatibility rows and allocated mirrors must not both be materialized on import.
4. **Probe budget:** the 60-second aggregate budget must cover `ossKey` recovery as well as classroom URLs and `audioUrl` fetches.
5. **Speech cache identity:** cache URL outcomes by the full legacy pair (`audioId`, `audioUrl`) so different URLs cannot share one result.
6. **Missing allocated IDs:** probe an `ast_`-looking ID before discarding a co-present recovery URL; imported/cross-pool legacy documents are in scope.
7. **Regression coverage:** the save-back failure test must perform a real allocation and assert cleanup of the pool and compatibility rows.

### Accepted design position

After successful server-backed speech conversion, retaining `audioUrl` as inert recovery metadata is acceptable. The blocker is incomplete server conversion: raw `/api/classroom-media/...` transport URLs must not enter a newly persisted document.

### Decision rule

Approve only when the required items above are fixed and the interleaving matrix is covered: worker failure, superseded load, simultaneous cold loads, post-commit navigation, incomplete conversion, reconciliation save failure, wholesale replacement, and stalled persistence. Otherwise, request changes.

**Blocking KEEP set:** 1, 3, 4, 5, 6, 7, 8. **Finding 2 is dismissed as an accepted design choice.**

wyuc:
All eight findings are addressed in `9de97591`:

1. **Save-back rollback.** A failed save-back now acts as if conversion never happened: the pass ledger is rolled back and the *unconverted* document is returned (the read still never fails), so repeated opens cannot accumulate orphans against durable legacy references.
2. **Retained URLs after the shared principal.** `retainLegacyUrl` is gone entirely — the dev authenticator's shared asset principal made it vestigial. A co-present `audioUrl` is dropped once the allocated id is confirmed resolvable, so stale/deployment-specific URLs no longer persist in documents.
3. **Completeness guard.** `containsClassroomMediaUrls` now also covers speech actions' `audioUrl`; a partially converted payload (audio failed, media succeeded) fails closed instead of persisting a raw transport URL.
4. **ZIP media collection.** Each converted asset ships once: the legacy row superseded by an allocated-id mirror (via `placeholderRef`) is skipped, so import no longer materializes unreferenced duplicates.
5. **Probe budget.** `ossKey` recovery fetches now route through the same aggregate deadline as every other probe; stalled CDN fetches can no longer exceed the 60-second pass budget.
6. **URL outcome cache.** Keyed by the `(audioId, audioUrl)` pair — actions sharing an id but carrying different URLs no longer share an outcome. Identical pairs still share one allocation.
7. **Allocated-shape trust.** Narrowed rather than removed: a co-present URL is dropped only after the memoized per-pass probe confirms the id resolves (imported/cross-browser documents keep their live handle). An allocation-shaped id with *no* co-present URL is still shape-trusted without a probe — nothing recoverable is at stake, and an already-converted classroom still pays zero per-open HEADs. This keeps round 3's per-open-probing fix intact.
8. **Real-allocation rollback test.** The migration suite's converter now performs real allocations (pool entries + audio/media compatibility rows) against a stand-in store, and the save-back-failure test asserts they are actually cleaned up.

Verification: full app suite green (4620 tests) on Node 22, `tsc` clean, `eslint` clean on touched files (one pre-existing warning elsewhere), `prettier --check` clean. PG/S3 contract suites remain CI-covered.

And the standing process request, once more: if further findings remain, please post the complete list — I will address all of it in one final round. Otherwise please leave an explicit review decision; COMMENTED reviews carry no merge decision.

wyuc:
Integrated the now-merged #1100 (`b169c41d`). The one real conflict was `http.ts`'s byte GET — it keeps both sides (bounded probe timeout + fail-closed redirect handling). The integration also surfaced a genuine interaction: the route boundary tests from #1100 stored assets under a per-header learner principal, which this branch's shared asset principal intentionally replaced — the tests now store under the shared principal. Full app suite (4635 tests), storage package suite, typecheck, and prettier all green on the merged tree.

cosarah:
Thanks for documenting the intended `audioUrl` lifecycle and the shared-principal direction. With those design choices taken as given, the remaining issues are edge cases around recovery, concurrency, and trust boundaries.

### P1

- **F-1101-006 — Export can capture references from before lazy conversion.** `use-export-classroom.ts` prepares the Zustand stage/scenes before calling `accessDocument()`, then uses the accessed document only for the latest name. If access converts and saves legacy references, the manifest can still contain the old handles while media collection sees the newly keyed rows, producing an inconsistent ZIP. Please access first and build one export snapshot that combines the post-conversion references with intentional unsaved edits; add a round-trip test covering an unconverted persisted document.

### P2

- **F-1101-001 — Definitively dead classroom media URLs are preserved instead of removed.** `allocateUrlMediaRef()` currently maps both `dead` and `unavailable` to `null`, so callers cannot distinguish a 404/410 from a retryable failure. This can leave dead slide, whiteboard, or video-manifest transport URLs behind and make an otherwise usable classroom fail the completeness guard. Propagate the three-state outcome through every media slot: remove confirmed-dead references, preserve transient ones, and allocate nothing for dead entries.

- **F-1101-002 — Durable narration recovery matches only part of a legacy pair.** When both `audioId` and `audioUrl` identify a narration, `getMirroredAudioRecord()` currently accepts a row matching either origin field. A retry can therefore reuse the mirror for `(shared-id, URL-A)` while converting `(shared-id, URL-B)`. Require both fields to match when both are supplied, while retaining exact id-only and URL-only recovery. Please cover conflicting pairs, exact-pair deduplication, and crash recovery from durable rows.

- **F-1101-003 — A failed reconciliation save returns the stale pre-concurrency snapshot.** If another writer changes the document during conversion, the code reloads and converts that authoritative `latest` document. If saving the reconciled result fails, the common catch rolls back correctly but returns the original `existing` snapshot, temporarily hiding the concurrent edit from the caller. After rollback, return the authoritative reloaded document without the discarded conversion; test the concurrent-edit plus rejected-save race and assert both the returned and durable snapshots.

- **F-1101-004 — Zero-byte blobs are still treated as usable in several paths.** A 200 response with an empty body, an empty compatibility row, or an empty pool asset can be allocated or selected and can suppress a usable co-present URL. Require `Blob.size > 0` at conversion ingestion and in the shared audio/media resolvers. Empty bytes should remain retryable, fall back to the URL when one exists, and never produce an allocated asset.

- **F-1101-008 — Superseded legacy narration fetches cannot be cancelled.** `AudioPlayer.play()` uses a plain `fetch(legacyUrl)` and checks its request token only after the fetch settles. A stalled request can keep `play()` and its callers pending after stop, replacement, navigation, or destruction. Give each play an `AbortController` and finite timeout, abort it on lifecycle supersession, and retain the direct media-element fallback for ordinary fetch/CORS failures.

- **F-1101-009 — Rollback tests do not exercise the production persistence boundary.** Migration tests replace the real rollback helper with in-memory sets, while classroom cleanup tests do not combine successful early allocations with a later conversion or save failure. Add integration coverage using the real pool, real audio/media compatibility tables, and the real rollback helper, then assert that pool assets and both mirror types are absent after discarded migration and server-fetch passes.

- **F-1101-010 — One reviewed commit uses an unsupported Conventional Commit type.** Commit `673af150` is titled `release: v0.3.2`, but `CONTRIBUTING.md` allows `feat`, `fix`, `docs`, `refactor`, `test`, `chore`, `ci`, `perf`, and `style`. Please rewrite the subject with the repository's release-chore convention while preserving the release content, then audit the full reviewed range for unsupported types.

- **F-1101-011 — The existence-probe deadline stops at the descriptor response.** `probeTimeoutMs` bounds the descriptor GET but is not carried into the signed-object fetch or its body/error consumption. An unclassifiable HEAD followed by a prompt descriptor and a stalled signed response can still hold conversion and its document migration lock indefinitely. Use one absolute deadline across HEAD, descriptor, signed fetch, and body parsing, and verify cancellation for both a stalled fetch and a stalled body. Ordinary playback `resolve()` behavior does not need to change.

- **F-1101-012 — Indirect-read URLs are accepted without scheme validation.** Both the server handler and HTTP client currently accept any non-empty string as a signed URL. A custom signer can therefore emit a relative, malformed, `data:`, `file:`, or other non-HTTP URL, and the client may attempt to fetch it. Parse and require an absolute `http:` or `https:` URL before server emission and again before the client's second fetch. Invalid server results should fail internally without a descriptor or `Location`; invalid client descriptors should fail as `MALFORMED_RESPONSE` without a follow-up request.

These are the complete remaining findings from this review pass.

wyuc:
Thank you for the complete list — that is exactly what this needed. All eight actionable findings are addressed across `60e984eb`, `d38be56b`, `31dd09ed`, and `42fc338b`:

- **F-1101-006 (P1)** — export now accesses the document FIRST and builds one snapshot: post-conversion references combined with intentional unsaved edits. Round-trip test covers an unconverted persisted document.
- **F-1101-001** — the three-state outcome (`ok`/`dead`/`unavailable`) now propagates through every media slot (slide, whiteboard, video manifest): confirmed-dead references are removed, transient ones preserved, and dead entries allocate nothing. Removed dead URLs no longer trip the completeness guard — covered by a new boundary test (dead endpoints → load completes, no raw URLs, zero allocations) alongside the transient-failure test (load fails closed, nothing persisted).
- **F-1101-002** — `getMirroredAudioRecord` requires BOTH origin fields to match when both are supplied; id-only and URL-only recovery retained. Conflicting pairs, exact-pair dedup, and crash recovery from durable rows are all pinned.
- **F-1101-003** — a failed reconciliation save now returns the authoritative reloaded document (unconverted) after rollback, not the stale pre-concurrency snapshot. The concurrent-edit + rejected-save race is tested, asserting both the returned and durable snapshots.
- **F-1101-004** — `Blob.size > 0` is required at conversion ingestion and in the shared audio/media resolvers; empty bytes stay retryable, fall back to the URL, and never produce an allocation.
- **F-1101-008** — each `play()` gets an `AbortController` and a finite timeout; stop/replace/destroy abort the in-flight fetch at the network layer; the direct media-element fallback is retained for ordinary fetch/CORS failures (and for zero-byte bodies).
- **F-1101-009** — integration coverage now runs the real pool + real Dexie compatibility tables + the real `rollbackConvertedAllocations`: converter-level rollback, discarded migration save-back, and discarded server-fetch pass each assert pool assets and both mirror types are gone (and legacy source rows survive).
- **F-1101-011** — `exists()` now runs under ONE absolute deadline covering HEAD, descriptor read, the signed fetch, and body parsing; a stalled fetch and a stalled body each cancel within the budget. Playback `resolve()` is unchanged.
- **F-1101-012** — both sides validate: the server refuses to emit a non-absolute-http(s) signed URL (500, no descriptor, no `Location`, and the refused URL never appears in the body), and the client rejects such a descriptor as `MALFORMED_RESPONSE` without issuing the follow-up request.

**F-1101-010** I am not acting on, and I want to explain rather than silently skip: `673af150` (`release: v0.3.2`) is a commit on `main` produced by the release process; it entered this branch by merge and is not part of this PR's diff. Rewriting a published commit on main is not something a feature PR can or should do. If the type taxonomy should accept `release` (or releases should use `chore:`), that is a contribution-policy fix for CONTRIBUTING.md and the release tooling, happy to file it separately. I did audit this branch's own commits: all use allowed types.

Verification on the final tree: full app suite 4663 green, `@openmaic/storage` suite 853 green, `tsc` clean, `eslint` clean, `prettier --check` clean, and the package version gate passes (storage 0.3.0 → 0.4.0, dsl 0.8.0 → 0.9.0).

That is the complete list, closed. If another pass finds something new I will of course address it — but with the full remaining set now fixed, an explicit review decision would let this land.

wyuc:
One more pass from our side before your next look — after closing your complete list, we ran our own full-diff adversarial sweep (two independent reviewers over the cumulative diff: assertion falsification, async interleaving, invariant-to-test reconciliation, consumer consistency). It surfaced 8 issues, all in already-named classes, all fixed in `c71325fe`…`cb0ef90c`:

- **exists() trusted identity-less 200s (P1-class).** A 200 HEAD without a valid revision/media-type identity (auth wall, redirect landing page) reported existence, and the converter could then drop a co-present live URL. Identity-less successes are now unclassifiable and fall through to the bounded byte read — and that fallback no longer mints object URLs, so probes pin nothing.
- **Manifest-only refs survived in the document but not in cleanup.** The orphan computation now uses the full document ref set (including video-manifest keys), so a manifest-only allocation is never rolled back while persisted.
- **Export-path allocations released.** The ephemeral export snapshot's fresh allocations are rolled back once the ZIP has captured their bytes (or on any failure), unless the converted document was committed.
- **Mirror pair-matching closed in the last branch.** The local-bytes branch now matches durable mirrors on both origin keys — the same rule the dangling branch got — and a co-present URL is dropped only when the allocation provably came from that pair.
- **ZIP never ships empty audio.** A row with no usable bytes produces no zip path, so the legacy-URL rescue is no longer suppressed for evicted rows.
- **Every migration exit rolls back.** The generation-fence throw and the birth-path lost-document throw now compensate the pass ledger too.
- **Cold loads without Web Locks.** Locks-absent now degrades to the app's established lock-free route (with ledger+rollback intact) instead of a silent null; lock-capable contention still fails the load. The requireLock knob became dead with that split and was removed.

Each fix carries a mutation-checked test (reverting the fix turns the new test red). Full app suite (4676), storage suite (856), tsc, eslint, prettier, and the package version gate are green.

This self-sweep is now part of our loop rather than a one-off, so future rounds should have materially less for you to find. Your declared-complete list remains fully closed; we believe the branch is ready for a decision.

cosarah (approved):
LGTM

lib/export/classroom-zip-utils.ts
wyuc: Honest status on this one, since it is the one thread here that is not fully closed by the branch.

The aggregate budget exists in the converter: legacy URL probes in a conversion pass share a 60-second wall-clock deadline (`URL_PROBE_BUDGET_MS`) on top of the per-probe 15-second timeout, and `ossKey` recovery fetches were brought under the same deadline. That is the path that runs on every document open.

The two paths you named here — `collectLegacyAudioForExport` and the video timeline fallback — are still bounded only by concurrency (4) and the per-probe timeout, so your `ceil(urlCount / 4) * 15s` figure still holds for them. Both only run for a document whose narration could not convert (bytes unavailable, or a transient probe failure), so it is a tail case rather than the common path, but the bound should be the shared one.

Rather than grow this branch further, I have filed it as #1129 (item 3), together with the two design deferrals from earlier rounds: production principal derivation with document/asset ownership aligned, and conditional writes for the document store. The fix there is to hoist the aggregate deadline into the shared probe helper so conversion, export, and the video timeline all abandon the remainder once the budget is spent.

PR #1109, opened 2026-08-12
feat(scene-gen): expose content/actions/tts phase in sidebar

weinaike:
The classroom scene-sidebar "generating" card only showed a flat "Generating..." shimmer with no indication of which step a scene was on. Replace it with a per-scene 3-dot indicator (content -> actions -> tts) driven by a new transient store field.

- store (stage): add generatingPhases map keyed by outline id + SceneGenerationPhase type + set/clear setters. Per-outline (not single value) so the card stays accurate under parallel content pre-warming.
- hook (use-scene-generator): write the phase to the store at every transition incl. a new 'tts' phase (previously only content/actions were emitted via callback, which the classroom consumer doesn't even wire). Covers both generateRemaining and retrySingleOutline; clears on completion. onPhaseChange type widened to include 'tts'.
- scene-sidebar: render 3 dots + current-phase label on the generating card; collapses to 2 dots when TTS is off; keeps shimmer + paused/failed behavior unchanged.
- i18n: add phaseContent/phaseActions/phaseTts (+ short) under generation across all 10 locales (required by check-i18n-keys alignment gate).

## Changes

-

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1.
2.
3.

-

### Evidence

- [ ] Manually tested locally

## Checklist

- [ ] My code follows the project's coding style
- [ ] I have performed a self-review of my code


wyuc (commented):
The direction looks good, CI is green, and I verified the feature through a real UI flow with server-side TTS enabled. A three-scene course completed successfully; the sidebar showed the three-dot indicator during Actions and TTS, removed the placeholder on completion, and produced no console errors.

Two lifecycle gaps should be addressed:

1. `retrySingleOutline` writes phases directly with `setGeneratingPhase`, so the newly widened `onPhaseChange` callback does not receive `content`, `actions`, or `tts` transitions during retries. Please route retry transitions through the same phase helper used by `generateRemaining`, or explicitly invoke the callback there.
2. `generatingPhases` is only cleared on the normal batch-completion path. The early `pending.length === 0` completion path and final successful retry path leave phase entries behind. A per-outline removal action, called whenever an outline leaves `generatingOutlines`, would make cleanup consistent across success, failure, retry, abort, and completion paths.

Please also add focused tests for phase transitions and cleanup, including retry and TTS-disabled cases. The Content state was observable in the accessibility tree during E2E, but it transitions too quickly for reliable screenshot capture; Actions and TTS visual states were both confirmed.

wyuc:
Thanks @weinaike, and sorry for the slow follow-through. I'm closing this to keep the queue tidy: the branch conflicts with `main` and hasn't moved since the review in August. Generation progress presentation has also changed a fair bit since then (the Pro workbench has its own progress surface), so the sidebar approach here may no longer fit as-is.

If you'd like to pick it up again, please check the current generation flow on `main` first; opening an issue to confirm the direction before a new PR would save churn. Thanks!

PR #1111, opened 2026-08-12
feat(pi): add native child runtime with scoped Spotlight

LING-6150:
## Summary

This PR adds the default-off Native Child execution harness and its first
Native tool, Spotlight.

It is the PR2 delivery in #1049 and builds on the merged shared Pi transport
prerequisite in #1108.

## What changed

- Add an explicit request-scoped `legacy | native` Child runtime selector.
- Keep Legacy as the default runtime.
- Run Native Child through the existing OpenMAIC-resolved model, `streamLLM`,
  shared Pi transport adapter, and Pi Agent loop.
- Add a dedicated Native Child prompt and same-Child tool-result continuation.
- Add Native-only Spotlight with:
  - a non-empty `elementId` schema;
  - request-start current-Scene identity capture;
  - registration only for a current slide with matching consumed Scene evidence;
  - exact element-ID authorization;
  - preflight and effect-boundary authorization checks;
  - existing best-effort action SSE dispatch.
- Preserve Director-provided Web evidence as untrusted prompt data without
  registering a Native `web_search` tool.
- Bound lower-level Native execution resources with 4 admitted tool-call
  attempts, 5 provider transports, a 60-second deadline, duplicate
  termination, and caller-owned cancellation propagation.
- Keep `executionCount` as diagnostic accounting only; there is no generic
  Native tool/action execution budget or coupling to `maxActionsPerAgent`.
- Stream accepted Native speech incrementally through the existing classroom
  `text_delta` path. Native tools remain available through the tool channel;
  this PR does not add a speculative structured-output classifier.
- Support action-only completion after accepted Spotlight dispatch.
- Return `native_empty_response` for an ordinary empty stop with no dispatched
  action.
- Keep the Legacy whiteboard shadow state factory-scoped and lazily memoized so
  Native does not instantiate it.

## Rollout

Both capabilities are server-only and default off:

```env
OPENMAIC_ENABLE_PI_NATIVE_CHILD_RUNTIME=true
OPENMAIC_ENABLE_PI_NATIVE_CHILD_SPOTLIGHT=true
```

The Spotlight flag does not select the runtime. When the Native runtime flag is
absent or false, the existing Legacy Child path remains active.

## Architecture boundaries

This PR does not:

- introduce another provider or model-resolution path;
- invoke Pi providers directly;
- modify the shared `build-agent.ts` or `stream-fn.ts` transport;
- add Native Web Search;
- add `wb_read`, whiteboard mutation, or RuntimeStore integration;
- add Browser ACK/receipt, projection, replay, registry, or telemetry machinery;
- change the Legacy Spotlight schema or classroom action behavior;
- define a generic business-policy budget for Native tools or actions.

Pi remains the Agent loop and tool-execution harness. OpenMAIC remains the sole
provider and model-resolution authority.

## Correctness coverage

Tests cover:

- production route → Director → `read_scene` → `call_agent` → Native Child →
  Spotlight → same-Child continuation;
- default-off route preservation of `pi-chat-child`;
- request-start current-Scene and exact element-ID authorization;
- Scene/Web evidence take-once behavior and failure isolation;
- independent attempt, execution, dispatched-action, and transport accounting;
- four admitted valid executions with continuation and no generic execution cap;
- decoupling Native execution from the Legacy `maxActionsPerAgent` policy;
- fifth-call over-budget precedence and same-batch termination;
- duplicate tool-call prevention;
- incremental, ordered Native classroom `text_delta` delivery;
- rejected, pending, timed-out, and late-settling Spotlight dispatch;
- provider/tool late rejection observation;
- caller cancellation and internal deadline settlement;
- action-only completion and `native_empty_response`;
- Legacy shadow-state lifecycle preservation;
- absence of a Native Child `web_search` tool.

## Validation

- Targeted PR2 regression gate: 8 files / 129 tests passed
- TypeScript: passed
- ESLint: passed
- Prettier: passed
- `git diff --check`: passed
- GitHub required checks: 5 / 5 passed on the latest head

The implementation and review follow-up were independently cross-reviewed.
The latest reviewer clarification is reflected here: incremental streaming is
restored, the speculative buffering/classifier change is absent, and the
generic execution-budget removal remains in place.

## Related

- Part of #1049
- Requires the already merged transport prerequisite #1108


wyuc (changes requested):
Thanks, the overall PR scope looks right: the Native Child stays behind a default-off flag, uses the existing provider path, and limits the first capability to evidence-authorized Spotlight. I have two changes to request before merging.

1. Please remove the generic action/tool-execution budget from the Native runtime.

   I do not think `maxToolExecutions`, especially when wired to `maxActionsPerAgent`, belongs at this layer. The runtime should bound lower-level execution resources and failure modes such as provider turns/transports, total deadline, abort, and protocol anomalies. The number of Spotlight, whiteboard, video, or future tool actions is an upper-layer business policy, and those tools have different semantics and costs.

   The current implementation also exposes the layering problem behaviorally: after the Nth permitted tool execution, `afterToolCall` immediately claims `native_tool_execution_budget` and terminates. That prevents the required tool-result continuation in the same Child and marks an already accepted, in-budget Spotlight as an error. Native Spotlight should validate authorization, dispatch best-effort, return its tool result, and let the same Child continue. If Spotlight-specific throttling becomes necessary later, it should live in the Spotlight tool or policy layer rather than the generic runtime.

2. Please prevent structured fallback output from being streamed as visible speech.

   In the Native path, `onVisibleTextDelta` only removes Markdown before immediately sending `text_delta`. If the model violates the speech-only prompt and emits the legacy JSON action format or another structured fallback, that internal JSON is shown directly in the classroom UI. The Legacy path already suppresses this class of model-contract failure. The Native path needs equivalent classification/buffering before dispatch, rather than relying on the prompt alone.

The rest of the scoped runtime design and evidence authorization path looked sound in this review.

LING-6150:
@wyuc Thanks for the review. Both requested changes are now addressed:
1. The Native generic execution budget and its maxActionsPerAgent coupling have been removed. Successful Spotlight results now continue in the same Child.

2. Native visible output is buffered and classified before classroom delivery, including structured fallback preceded by model prose or fences.

Targeted tests and all five GitHub checks are green. Could you please spare some time to check again? Thanks.

wyuc:
@LING-6150 Sorry, after tracing the actual Native Child inputs again, I realized my second review request was not justified.

The Native Child receives the native tool schema and no Legacy JSON action-format instruction through its prompt or normal history. I treated a speculative model/protocol violation as a concrete bug without first establishing a real production path. The full-turn buffering added in response also regresses streaming, which is an important product behavior. This was my review mistake, not a misunderstanding on your side, and I am sorry for sending you down this path.

Please revert the second change and restore the original `text_delta` streaming. No replacement structured-output classifier is needed in this PR. The first change, removing the generic execution budget and `maxActionsPerAgent` coupling while preserving same-Child continuation, is correct and should stay.

Based on the current review, I see no other blockers after streaming is restored. I will re-check the updated head and CI once that change is pushed.

LING-6150:
@wyuc  Thanks for the clarification. The latest head is `17c967a0`.

I have:

- removed the full-turn buffering and Native structured-output classifier;
- restored incremental classroom `text_delta` streaming;
- kept the generic execution-budget removal and same-Child continuation;
- added a regression test proving two provider deltas produce two ordered classroom events;
- updated the PR body to remove the reverted behavior;
- added a focused timeout only to the full route integration test for CI stability.

All 5 required checks are green. Could you please re-check the latest head and merge it if everything looks good?

wyuc (approved):
Re-checked the latest head after updating the base. The generic execution-budget coupling is removed, same-Child continuation is preserved, and incremental text_delta streaming is restored. All required checks are green.

PR #1115, opened 2026-08-13
feat(video-export): add bounded local chunk executor

### LINKED ISSUE #1094: [Task] Add a bounded local chunk executor and deterministic assembler
Parent Epic: #1091

## Goal

Add a bounded local implementation of `plan → renderChunk → assemble` without introducing a distributed queue.

## Deliverables

- Immutable render plan containing frame boundaries, assets, fonts, runtime, producer, and FFmpeg versions.
- Configurable chunk count and explicit per-chunk internal worker count.
- Local chunk execution with bounded CPU, memory, browser, and temporary-disk usage.
- Deterministic closed-GOP chunk output.
- Ordered, hash-checked assembly.
- Idempotent retry for the same plan hash and chunk index.
- Failure-injection tests for missing, duplicate, stale, and mismatched chunks.

## Acceptance criteria

- Single-process execution renders and assembles the reference classroom.
- Assembled output passes ffprobe and representative frame/audio parity checks.
- A failed chunk can be retried without restarting the entire video.
- Nested parallelism is bounded; chunk fan-out does not silently multiply producer workers.
- The HTTP contract remains unchanged.

cosarah:
## Summary

This PR was developed with AI assistance and reviewed by a subagent before maintainer review.

Adds a bounded single-process implementation of HyperFrames `plan -> renderChunk -> assemble` for video export.

## Related Issues

Fixes #1094

## Changes

- Add immutable render-plan metadata with project, asset, font, runtime, producer, and FFmpeg fingerprints.
- Add configurable chunk count, per-chunk workers, chunk concurrency, fixed chunk size, and target chunk size.
- Execute closed-GOP chunks with bounded fan-out, hash-checked sidecars, reuse, retry, cancellation, and deadline propagation.
- Keep the existing HTTP API unchanged and preserve producer plan hashes by storing local executor metadata and outputs outside the producer plan directory.
- Add failure-injection and validation tests for missing, duplicate, stale, and mismatched chunks.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Run `cd render-service && npm run typecheck`.
2. Run `cd render-service && npm test -- --run`.
3. Enable `RENDER_CHUNK_EXECUTION=true` and configure `RENDER_CHUNK_COUNT`, `RENDER_CHUNK_WORKERS`, `RENDER_MAX_PARALLEL_CHUNKS`, `RENDER_CHUNK_SIZE_FRAMES`, or `RENDER_TARGET_CHUNK_FRAMES`.

- TypeScript typecheck passed.
- 12 test files and 63 tests passed.
- Docker smoke and multi-chunk benchmark runs passed, including 4 chunks with 4-way concurrency on 8 CPUs.
- The benchmark corpus verification passed locally; benchmark scripts, corpus, and result artifacts remain local-only and are not included in this PR.

### Evidence

- [x] Manually tested locally

Benchmark evidence: `static-deck-2m` completed in 36.8s with 8 CPUs, 16 GiB, 4 chunks, and 4 concurrent chunks; `webgl-lab-20m` completed in 461.3s under the same execution settings.

## Checklist


cosarah:
The PR is ready for review. Could @wyuc please review this change? All CI checks are passing, and the requested AI-assisted subagent review findings have been addressed.

render-service/src/chunk-executor.ts
```diff
@@ -0,0 +1,821 @@
...
+    } catch (error) {
+      if (error instanceof ChunkExecutorError && error.code === 'mismatched_chunk') throw error;
+      await removeOutput(chunk.outputPath);
+    }
+  }
+  await removeOutput(chunk.outputPath);
+  let lastError: unknown;
+  for (let attempt = 1; attempt <= 2; attempt += 1) {
+    try {
+      const result = await abortable(
```
wyuc: [P1] Cancellation/deadline only rejects the wrapper promise; `@hyperframes/producer@0.7.60` `renderChunk()` accepts no signal and continues running Chromium/FFmpeg. The retry loop then evaluates a second `renderChunk()` call before `abortable()` observes the already-aborted signal, so one cancellation can leave two renders writing the same output after the executor returns and the coordinator cleans up/releases its slot. I reproduced two calls and a late output write. Do not retry cancellation, and do not report terminal completion/release resources until every started producer operation has actually settled (or add real cancellation support to the producer API).

render-service/src/chunk-executor.ts
```diff
@@ -0,0 +1,821 @@
...
+    while (true) {
+      if (signal?.aborted)
+        throw new ChunkExecutorError('chunk_execution_failed', 'Render cancelled');
+      const index = cursor++;
+      const value = values[index];
+      if (value === undefined) return;
+      await worker(value);
+    }
+  };
+  await Promise.all(Array.from({ length: Math.min(concurrency, values.length) }, () => run()));
```
wyuc: [P1] `Promise.all` rejects as soon as one worker loop fails, while sibling loops keep rendering. `executeRenderChunks()` then restores process-global worker state and returns, allowing coordinator cleanup and the next admitted job to race with those still-active chunks. This defeats the configured concurrency bound and can delete plan/chunk files still in use. Stop scheduling new chunks after the first failure and drain all already-started worker promises before returning the batch error.

render-service/src/render-executor.ts
```diff
@@ -183,6 +192,87 @@ export class InProcessExecutor implements RenderExecutor {
...
+            chromium: this.runtimeVersions.chromium,
+          },
+          ...this.chunkExecution,
+          ...request.chunkExecution,
+        };
+        const result = await this.chunkExecutor(chunkRequest);
+        const chunkMetrics: RenderExecutionMetrics = {
+          resourceProfile: config.resourceProfile.name,
+          requestedCaptureMode: config.resourceProfile.requestedCaptureMode,
+          actualCaptureMode: result.plan.captureMode,
```
wyuc: [P1] This is the capture mode requested/locked at plan time, not the mode actually used by chunk workers. Producer can probe BeginFrame as unsupported and silently relaunch in screenshot mode, while the plan remains `beginframe`; chunk sidecars currently do not expose the actual session mode. Consequently the standard profile can succeed and report `actualCaptureMode=beginframe` after a screenshot fallback, bypassing `requireBeginFrame`. Propagate the observed mode from each chunk and fail if any required worker did not actually use BeginFrame.

render-service/src/chunk-executor.ts
```diff
@@ -0,0 +1,821 @@
...
+        maxParallelChunks: chunkCount,
+        abortSignal: request.signal,
+      },
+      planDir,
+    );
+  } finally {
+    if (previousWorkers === undefined) delete process.env.PRODUCER_MAX_WORKERS;
+    else process.env.PRODUCER_MAX_WORKERS = previousWorkers;
+  }
+  if (result.chunkCount !== chunkCount) {
```
wyuc: [P2] The producer treats `maxParallelChunks` as a cap and legitimately resolves fewer chunks for short videos or when `chunkSizeFrames` / `targetChunkFrames` is set. Requiring exact equality makes documented configurations fail with `mismatched_chunk` even though the producer returned a valid deterministic plan. Accept and persist the producer's resolved chunk count, or reject incompatible configuration combinations before planning with clearly documented exact-count semantics.

render-service/src/chunk-executor.ts
```diff
@@ -0,0 +1,821 @@
...
+    },
+    producer: result,
+    chunks: plan.chunks,
+    assets: plan.assets,
+    fonts: plan.fonts,
+    runtime: plan.runtime,
+    planHash: plan.planHash,
+    localFingerprint: digest({
+      schemaVersion: CHUNK_PLAN_SCHEMA_VERSION,
+      input: {
```
wyuc, line 711: [P2] The persisted envelope input includes `chunkSizeFrames` and `targetChunkFrames`, but this write-time fingerprint input omits both. `readImmutableRenderPlan()` hashes the complete `raw.input`, so setting either documented option makes the executor reject its own freshly written cache as `Local render plan hash mismatch` on the next call. Include the optional sizing fields by hashing the exact same input object used in the envelope; add cached-reread tests for both knobs.

wyuc (changes requested):
The bounded chunking direction is sound, and the current typecheck/tests/CI are green. However, the new path still has lifecycle and contract issues that can exceed the advertised resource bounds, bypass the required BeginFrame safeguard, and break documented sizing/cache behavior. The cancellation behavior was reproduced against a non-cancellable renderChunk-shaped dependency: execution returned after abort, invoked the chunk twice, and the underlying operation wrote output after the executor had settled. Please address the P1 lifecycle/capture issues and add tests using the real dependency shape before merging.

cosarah:
Addressed wyuc review findings in fce6a460: worker failures now drain sibling loops before returning; cancellation no longer retries and waits for producer-shaped operations to settle; producer-resolved chunk counts are accepted; chunk sizing fields are included in cache fingerprints; and sidecars with an observed non-BeginFrame capture mode are rejected. Local render-service typecheck and 63 tests pass.

cosarah:
Added the requested producer-shaped lifecycle regression tests in 79282582. They verify that an already-started non-cancellable chunk is drained after abort without retry, and that sibling chunk workers settle before the executor returns the first batch error. Render-service typecheck and all 65 tests pass locally.

cosarah:
Follow-up fix in b8e1854c: upgraded render-service to @hyperframes/producer 0.7.107, whose EffectiveChunkResult/perf sidecars expose the actual worker captureMode. The executor now requires every chunk worker to report BeginFrame for the standard profile and reports the observed mode; mocks/container contract were updated. Local typecheck and all 65 tests pass.

cosarah:
@wyuc, all requested changes are now in fce6a460, 79282582, and b8e1854c, with all CI checks passing. Please re-review the updated head.

cosarah:
Addressed the latest subagent review findings in a9f6777b: requested fan-out is now part of cache identity (no 2->4 silent reuse), chunk workers/concurrency are hard-bounded by resource profile, and cached plans validate producer plan path, chunk count/sequential boundaries, and chunk output confinement. Added regression tests; local render-service suite now has 67 passing tests.

render-service/src/chunk-executor.ts
```diff
@@ -0,0 +1,853 @@
...
+  if (result > maximum) {
+    throw new TypeError(`${name} must be <= ${maximum} for ${config.resourceProfile.name}`);
+  }
+  return result;
+}
+
+async function settleProducer<T>(operation: Promise<T>, signal?: AbortSignal): Promise<T> {
+  // The producer distributed API has no cancellation parameter. Await it to settle before
+  // releasing the coordinator slot or cleaning the plan directory.
+  const result = await operation;
```
wyuc, line 201: [P1] The deadline cannot interrupt an in-flight chunk here. `InProcessExecutor` aborts its signal when `RENDER_JOB_DEADLINE_MS` expires, but `renderChunk()` accepts no signal and this function awaits the operation before checking `signal.aborted`. If Chromium or FFmpeg hangs, the promise never settles, the documented hard deadline is ineffective, and the service’s single coordinator slot remains occupied indefinitely, blocking every queued render. Run each chunk behind a genuinely terminable process/worker boundary (or another bounded termination mechanism) and cover a never-settling chunk in a deadline regression test.

render-service/src/chunk-executor.ts
```diff
@@ -0,0 +1,853 @@
...
+    );
+  }
+  const sidecar = await defaultReadChunkResult(chunk.outputPath);
+  if (!sidecar || sidecar.sha256 !== result.sha256 || sidecar.planHash !== planHash) {
+    throw new ChunkExecutorError(
+      'stale_chunk',
+      `Chunk ${chunk.index} has no matching plan/hash sidecar for plan ${planHash}`,
+    );
+  }
+  if (sidecar.captureMode !== undefined && sidecar.captureMode !== 'beginframe') {
```
wyuc: [P1] Make capture-mode validation profile-aware. The supported `low-memory` profile explicitly forces screenshot capture, so every real chunk sidecar reports `captureMode: screenshot`; this unconditional BeginFrame check then rejects every chunk whenever `RENDER_RESOURCE_PROFILE=low-memory` and `RENDER_CHUNK_EXECUTION=true`. Enforce BeginFrame only when the selected profile requires it (and test the low-memory chunk path), or explicitly disallow this configuration at startup instead of documenting it as supported.

wyuc (changes requested):
The latest changes address the earlier drain, cache-identity, capture observability, and resource-bound findings, and the current CI/test suite is green. Two blocking runtime-contract issues remain in the opt-in chunk execution path: an in-flight chunk cannot be interrupted at the advertised job deadline, and the documented low-memory screenshot profile is rejected unconditionally. Please address both and add focused regression coverage before merging.

cosarah:
Addressed the latest blocking review in 9d1839a0. Built-in chunk renders now run in detached child process groups; deadline/cancellation sends SIGKILL to the complete Chromium/FFmpeg process group and waits for the worker exit before returning. Capture validation is now profile-aware (standard=beginframe, low-memory=screenshot). Verified with local typecheck, 67 tests, and a real Docker 2-chunk/2-concurrency smoke on producer 0.7.107 (61.1s, successful BeginFrame output).

cosarah:
Resolved the latest main conflict in 3f0422af while preserving main capture policy changes (standard now prefers BeginFrame but allows screenshot fallback; low-memory remains screenshot-only) and the chunk resource limits. Local post-merge render-service typecheck and 68 tests pass; full CI is now running on a mergeable head.

cosarah:
@wyuc, the two latest blocking findings are addressed and all post-merge CI checks pass. Please re-review head 3f0422af.

render-service/src/chunk-executor.ts
```diff
@@ -0,0 +1,920 @@
...
+    stdio: ['ignore', 'inherit', 'inherit', 'ipc'],
+  });
+  return new Promise((resolveResult, rejectResult) => {
+    let settled = false;
+    let aborting = false;
+    const finish = (callback: () => void): void => {
+      if (settled) return;
+      settled = true;
+      signal?.removeEventListener('abort', onAbort);
+      child.disconnect();
```
wyuc: [P1] Guard the IPC disconnect after worker termination. When SIGKILL terminates this detached child, Node closes the IPC channel before emitting exit; the aborting exit handler enters finish(), where child.disconnect() on an already-disconnected channel emits an unhandled ERR_IPC_DISCONNECTED and terminates the parent process. I reproduced exit code 1 with the same detached fork plus process-group SIGKILL sequence. Thus any chunk deadline or cancellation can crash the entire render service instead of returning deadline_exceeded. Check child.connected, or otherwise make disconnect idempotent, and add a regression test with a real child that is killed by deadline while the parent remains alive.

wyuc (changes requested):
Most of the previous lifecycle and capture-profile findings are addressed, and all checks are green. One blocking failure remains in the new hard-deadline path: killing a detached chunk worker closes IPC before exit, and finish then calls child.disconnect() unconditionally. I reproduced Node exiting with an unhandled ERR_IPC_DISCONNECTED under the same fork/SIGKILL sequence. This can turn a single chunk deadline or cancellation into a render-service process crash. Please guard the disconnect and add a real child-process termination regression test that proves the parent survives and returns deadline_exceeded.

cosarah:
Fixed the remaining IPC teardown crash in b368f638. finish() now disconnects only while connected and tolerates the connected/disconnect race after SIGKILL. Added a real detached child-process regression fixture: the worker never replies, deadline kills its process group, the promise returns Render cancelled, and the Vitest parent remains alive. Local typecheck and all 69 tests pass.

wyuc (approved):
Re-reviewed at b368f638. The IPC teardown race is guarded, the real detached-child deadline regression exercises the production helper, and all required checks pass.

cosarah:
@wyuc, the IPC teardown crash from the latest review is fixed in b368f638. The new regression test kills a real detached child process and verifies the parent survives with Render cancelled; all 5 CI checks now pass. Please re-review the updated head.

PR #1116, opened 2026-08-13
refactor(media): share one stored-bytes resolver across export paths (#1007 part 3a)

wyuc:
First PR of the #1007 part 3 (export convergence) line, now that part 2(c) (#1101) is on `main`.

Three export paths (classroom ZIP, PPTX, video export) each re-implemented the same byte-resolution fallback chain — pool → Dexie compatibility row → task/CDN URL — with subtle per-path differences. This replaces them with one shared helper, `lib/media/resolve-stored-bytes.ts`.

The genuine divergences between the three implementations are preserved as explicit options rather than silently unified:

- **Fetch strictness** (ZIP accepts any response status but rejects zero-byte answers, PPTX requires `response.ok`, video requires both) → `fetchPolicy`.
- **Regeneration gating** (PPTX/video suppress stale bytes through `resolveMediaRef` while a regeneration is in flight; ZIP did not) → `resolutionGating`.
- **Fallback levels** (ZIP pool-only; PPTX adds the task URL; video adds the `ossKey` CDN fetch) → `loadCompatRow` / `compatRowCdnFallback` / `taskUrlFallback`.
- **Dexie row source** (read internally by compound key vs. pre-loaded by the caller) → `loadCompatRow` vs `record`.

Replaced: ZIP's `pooledBytesForRef`, PPTX's `resolveStoredMediaBlob`, video export's `resolveMediaBytesWithFallback`, plus two duplicate `mediaRefFromRecordId` helpers. Call sites keep their old local signatures through thin wrappers, so no calling code changed shape.

## Behavior preservation, and the one divergence

The three call sites resolve the same bytes from the same inputs as before. That is not an assertion you have to take on trust: `tests/media/resolve-stored-bytes-equivalence.test.ts` reproduces the three pre-refactor implementations verbatim as they stand at the merge base and drives them side by side with the current helper under each call site's real option set, over a declared matrix (ref shapes × task states × pool outcomes × fetch outcomes × record shapes × stage scoping), comparing returned bytes, thrown error types, blob identity by reference, and the ordered lease / Dexie / fetch sequences including their arguments. Run it with `pnpm exec vitest run tests/media/resolve-stored-bytes-equivalence.test.ts`. One divergence, deliberate:

**A compatibility row with no `blob` no longer throws.** On `main` the PPTX path evaluated `record.blob.size` unguarded, so such a row rejected the whole PPTX export with a `TypeError`; the shared helper guards it and falls through to the next byte source. This restores the guard `main`'s video path already had (`blob && blob.size > 0`) and is what makes the module's documented "never throws" contract true. No writer in the repo can currently produce that row shape — success rows, failed rows (an empty placeholder blob), classroom import, and the conversion mirror all populate `blob` — so the divergence is unreachable in practice, but it is a divergence and not worth hiding.

One further note for reviewers weighing "pure refactor": the extraction introduces async-function boundaries the inline chains did not have — one at the pool level, and a second at the compatibility level for the PPTX shape, whose row check used to be inline. Every operation after them therefore starts one or two microtasks later than it did on `main`. The order, the count, and the identity of the operations themselves are unchanged, and every value whose observation point mattered — the effective ref, the task, and a supplied row's `error` verdict — is taken before the boundaries. What that leaves is a scheduling difference: a microtask that mutated a row in place, or committed a Dexie write, in exactly that window would be observed by one version and not the other. No writer in the repo does that — every `mediaFiles` writer replaces the record rather than mutating a held reference — so the difference has no trigger today, but it is a difference and not a proof of equivalence.

## Rebase across #1101

Two hunks needed a decision rather than a replay, both where #1101's later hardening landed in the same lines:

- The ZIP's pool fetch maps to `{ requireOk: false, requireNonEmpty: true }`. The branch predated #1101's zero-byte rejection, so a verbatim replay would have restored shipping empty media files. Now pinned by a test that reddens if `requireOk` is tightened.
- `collectMediaFiles` keeps #1101's skip of a legacy row superseded by its allocated-id mirror, expressed through the shared `mediaRefFromRecordId` so the local duplicate goes away with the others.

## Review

Six cross-vendor review rounds (Codex and DeepSeek, each reviewing the same diff independently, findings merged by consensus).

- **Round 1** — an option surface that promised more than the code did: `taskUrlFallback` was inert without `resolutionGating`, and an unguarded `record.blob` broke the module's "never throws" contract. Both fixed.
- **Round 2** — no behavior defects; every finding was a test that would still pass with the behavior broken. Six mutation-verified cases close them.
- **Round 3** — no behavior defects. Its findings were the `blob` divergence documented above and an overstated mutation claim in a commit message.
- **Round 4** — the one real defect the earlier level-by-level rounds missed: the extraction moved a supplied row's `error` verdict past the pool await, so an `error` landing mid-lookup changed which bytes shipped. Fixed, with both directions pinned.
- **Round 5** — no new code defect. One reviewer enumerated every read in the helper against its pre-refactor observation point and could not falsify the sweep; the other confirmed the fix and raised only commit-message accuracy.
- **Round 6** — no new code defect and no undocumented divergence from either reviewer; one reviewer independently confirmed the two divergences stated above are the complete set. Remaining findings were documentary and are folded in.

Refs #1007

### DIFF
--- a/lib/export/classroom-zip-utils.ts
+++ b/lib/export/classroom-zip-utils.ts
@@ -3,11 +3,10 @@ import type { ManifestAction } from './classroom-zip-types';
 import { db } from '@/lib/utils/database';
 import type { AudioFileRecord, MediaFileRecord } from '@/lib/utils/database';
 import type { Scene } from '@/lib/types/stage';
-import { isConcreteMediaAddress } from '@/lib/media/resolve-media-ref';
 import { resolveAudioBlob } from '@/lib/media/resolve-audio-bytes';
 import { fetchMediaUrl } from '@/lib/media/fetch-media-url';
 import { mapWithConcurrency } from '@/lib/media/convert-legacy-asset-refs';
-import { withAssetUrl } from '@/lib/media/use-asset-url';
+import { mediaRefFromRecordId, resolveStoredBytes } from '@/lib/media/resolve-stored-bytes';
 
 // ─── Export: Collect Media ─────────────────────────────────────
 
@@ -55,26 +54,15 @@ export async function collectAudioFiles(scenes: Scene[]): Promise<CollectedAudio
  * compatibility write then fails, the task records
  * `MEDIA_COMPATIBILITY_STORE_LAGGED` and the document deliberately keeps the
  * same reference. Rendering and the other export paths resolve the pool, so the
- * ZIP must too — otherwise it ships media the classroom no longer shows.
+ * ZIP must too -- otherwise it ships media the classroom no longer shows. The
+ * ZIP predates the `response.ok` validation the other export paths added and
+ * keeps that laxity, but a zero-byte pool answer is not usable bytes: the
+ * compatibility row stays the fallback rather than shipping an empty file.
  */
 async function pooledBytesForRef(ref: string): Promise<Blob | null> {
-  if (isConcreteMediaAddress(ref)) return null;
-  try {
-    return await withAssetUrl(ref, async (url) => {
-      if (!url) return null;
-      const blob = await fetch(url).then((response) => response.blob());
-      // Zero-byte pool answers are not usable bytes: the compatibility row
-      // stays the fallback rather than shipping an empty media file.
-      return blob.size > 0 ? blob : null;
-    });
-  } catch {
-    // The compatibility row remains the fallback when pool access fails.
-    return null;
-  }
-}
-
-function mediaRefFromRow(record: MediaFileRecord): string {
-  return record.id.includes(':') ? record.id.split(':').slice(1).join(':') : record.id;
+  return resolveStoredBytes(ref, {
+    fetchPolicy: { requireOk: false, requireNonEmpty: true },
+  });
 }
 
 export async function collectMediaFiles(stageId: string): Promise<CollectedMedia[]> {
@@ -88,11 +76,13 @@ export async function collectMediaFiles(stageId: string): Promise<CollectedMedia
   // deriving its rows from the document's speech actions instead of
   // enumerating the table.
   const supersededLegacyRefs = new Set(
-    records.map(mediaRefFromRow).filter((ref) => records.some((row) => row.placeholderRef === ref)),
+    records
+      .map((row) => mediaRefFromRecordId(row.id))
+      .filter((ref) => records.some((row) => row.placeholderRef === ref)),
   );
   const collected: CollectedMedia[] = [];
   for (const record of records) {
-    const elementId = mediaRefFromRow(record);
+    const elementId = mediaRefFromRecordId(record.id);
     if (supersededLegacyRefs.has(elementId)) continue;
     const ext = record.mimeType?.split('/')[1] || 'jpg';
     const pooled = await pooledBytesForRef(elementId);
--- a/lib/export/use-export-pptx.ts
+++ b/lib/export/use-export-pptx.ts
@@ -22,8 +22,8 @@ import { createLogger } from '@/lib/logger';
 import { inlineHtmlAssets, createAssetFetcher } from './inline-assets';
 import type { FetchAsset } from './inline-assets';
 import { createProxiedFetch } from './proxied-fetch';
-import { db, mediaFileKey } from '@/lib/utils/database';
-import { withAssetUrl, type AssetUrlLeaseState } from '@/lib/media/use-asset-url';
+import type { AssetUrlLeaseState } from '@/lib/media/use-asset-url';
+import { resolveStoredBytes } from '@/lib/media/resolve-stored-bytes';
 import {
   MISSING_ASSET_LEASE,
   isConcreteMediaAddress,
@@ -370,15 +370,6 @@ function buildSpeakerNotes(scene: Scene): string {
   return parts.join('\n');
 }
 
-async function fetchBlob(url: string): Promise<Blob | null> {
-  try {
-    const response = await fetch(url);
-    return response.ok ? await response.blob() : null;
-  } catch {
-    return null;
-  }
-}
-
 async function blobToDataUrl(blob: Blob): Promise<string> {
   return new Promise<string>((resolve, reject) => {
     const reader = new FileReader();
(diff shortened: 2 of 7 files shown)


wyuc:
Codex review converged on the first round with zero actionable findings (expected for a pure refactor: the fallback order, validation policies, and resolution gating of each path are preserved, and every genuine divergence is an explicit option). Local validation: root tsc / eslint / prettier clean, 740 tests passing across the export, video-export, and media suites.

Note: this PR is stacked on #1101, so CI does not run until it retargets to main after #1101 merges.

cosarah:
I have no additional code findings beyond the following three retained items:

1. **F-1116-001 — Add the frozen-base differential harness.** Commit the reviewer-runnable harness and its complete snapshot/output, using the verbatim `e1578083c` ZIP/PPTX/video implementations. It must compare results, blob identity, arguments, lease/Dexie/fetch ordering, and the permitted microtask differences across the declared matrix, allowing only the documented blob-less PPTX divergence.

2. **F-1116-002 — Check abort before consuming one-shot evidence.** A pre-aborted request or tool signal must not consume Scene/Web evidence, and a later valid delegation must still receive both packets. Existing invalid-agent, failed-child ownership, abort propagation, and listener-cleanup behavior must remain unchanged.

3. **F-1116-003 — Align the SDK documentation with the package contract.** Update all SDK/cookbook version references to the checked-in package versions and list both required S3 peers, `@aws-sdk/client-s3` and `@aws-sdk/s3-request-presigner`. Verify the documented standalone setup with a clean install and an isolated `signReadUrl` smoke test.

Once these three findings are resolved on the PR head and the follow-up review finds no new regressions, I consider the code ready to merge from this review's perspective.

wyuc:
Thanks — taking these one at a time.

**F-1116-001 — agreed, and it is the right call.** The PR description claims a differential comparison against the verbatim pre-refactor implementations, but that harness was run out-of-tree and never committed, so nothing in the repository lets you reproduce it. A claim a reviewer cannot re-run is not evidence. I am adding the harness and its output to this PR: the three `e1578083` implementations frozen verbatim, driven against the current helper over the declared matrix, asserting returned bytes, blob identity, call arguments, and the ordered lease / Dexie / fetch sequences, with the documented blob-less PPTX case as the only permitted divergence and the microtask offsets stated explicitly. I will report back when it is on the head.

**F-1116-002 and F-1116-003 — I think these belong to other pull requests.** Neither has any surface in this branch, and I would rather flag that than silently implement something unrelated. This PR changes exactly six files:

```
lib/export/classroom-zip-utils.ts
lib/export/use-export-pptx.ts
lib/media/resolve-stored-bytes.ts
lib/video-export-app/collect.ts
tests/export/classroom-zip-pool-first.test.ts
tests/media/resolve-stored-bytes.test.ts
```

Grepping the complete `origin/main...HEAD` diff returns zero occurrences of the subject matter of either finding:

- **F-1116-002** — `evidence`, `delegation`, `packet`, `invalid-agent`, `failed-child`, `abort`, `listener`, `spotlight`, `call-agent`: no matches. The behavior described (one-shot Scene/Web evidence, delegation packets, failed-child ownership, abort propagation, listener cleanup) lives in `lib/agent/runtime/run-native-child.ts`, `lib/agent/runtime/native-child-contract.ts`, and `lib/chat/pi/tools/call-agent.ts` — the native child tool runtime, none of which this branch touches. #1011 looks like the intended home.
- **F-1116-003** — `signReadUrl`, `aws-sdk`, `cookbook`, `peerDependencies`, `s3`: no matches. That surface is `@openmaic/storage` (its README/docs and the two S3 peers already declared in its `package.json`) plus `lib/persistence/asset-byte-store.ts`. This branch modifies no package under `packages/` and adds no dependency.

Could you confirm whether those two were meant for this PR? If there is an indirect path through this diff that I am not seeing, tell me which line and I will take them on. Otherwise I will treat F-1116-001 as the only outstanding item here.

One process note: this arrived as an issue comment rather than a submitted review, so the PR still reports `REVIEW_REQUIRED`. Once the harness lands and you are satisfied, it will need an actual review submission to unblock.

wyuc:
**F-1116-001 is on the head** (`tests/media/resolve-stored-bytes-equivalence.test.ts`).

The three implementations are reproduced verbatim as they stand at `e1578083` — still byte-identical to `main`, since nothing merged between the two touched these functions, so the frozen base and the current base agree. Each is driven side by side with the shared helper under the option set its real call site passes, over the declared matrix: 6 ref shapes (opaque id, `gen_*` placeholder, http, `blob:`, `data:`, path-like) × 9 task states (absent, done with and without a URL, generating, pending, retryable failure, sensitive failure, other-stage, `placeholderRef`-matched) × 3 pool outcomes (URL, miss, throw) × 4 fetch outcomes (ok, non-OK, empty, throw) × 7 record shapes (absent, ok, errored, empty blob, empty blob with `ossKey`, blob-less with `ossKey`, non-compound id) × 2 stage scopes.

Per combination it compares the outcome (bytes, `null`, or the thrown type), **blob identity by reference** — whether the answer is the caller's own row blob or freshly fetched bytes — and the **ordered lease / Dexie / fetch sequence including arguments**. Absolute microtask offsets are not compared, for the reason the description gives: the extraction moves later operations one or two ticks while leaving the sequence, the arguments, and the results alone. Run it with:

```
pnpm exec vitest run tests/media/resolve-stored-bytes-equivalence.test.ts
```

The blob-less PPTX row is the only permitted divergence, and the harness pins *where* it may occur rather than merely tolerating it: every divergent combination must be that row with a stage, and the total must equal the combinations that actually reach the row — all of them except the ones where the frozen path returns pooled bytes before the row read. That count is derived in the file, not hard-coded to whatever the run produced.

I also checked that the harness bites, since a harness that cannot fail is not evidence either. Breaking the current helper:

| change to the helper | result |
|---|---|
| drop the zero-byte rejection in `fetchBytes` | ZIP and video red |
| drop the `response.ok` check in `fetchBytes` | PPTX and video red |
| drop the compound-id `effectiveRef` derivation | video red |
| leak the task into the gate (`gate = task`) | **all green — and correctly so** |

The last row is the harness's boundary, and I documented it in the file rather than papering over it: all three call sites pass `resolutionGating: true`, so `gate === task` for them and that change is unobservable in production. It is covered by a targeted case in `resolve-stored-bytes.test.ts` instead. The same goes for state that mutates mid-await — the supplied row's `error` verdict timing — which a static matrix cannot reach and which has its own two tests.

F-1116-002 and F-1116-003 are unchanged from my previous comment: I could not find any surface for either in this branch's six files, and I would rather ask than implement something that belongs elsewhere. Happy to take them on if you can point me at the line here that they touch.

cosarah:
F-1116-001 remains open for the frozen-base equivalence evidence.

The committed harness covers the declared matrix, but it still lacks a media ref containing an additional `:` (for example, `ast_opaque_1:variant`), so the compound-id preservation claim is not exercised by the differential matrix. The blob-less PPTX branch also asserts only that the new implementation does not throw; it does not assert the exact permitted fallback result, blob identity where observable, or the complete post-divergence lease/Dexie/fetch trace.

Please extend the harness with those cases using the verbatim frozen adapters and attach the complete reproducible output. No other scoped media code finding remains from this review.

wyuc:
**F-1116-001 is addressed on head `45635e22`.** Both gaps are closed in the harness (`tests/media/resolve-stored-bytes-equivalence.test.ts`); the three frozen implementations remain byte-identical to `e1578083`.

**1. Compound-id preservation is now exercised by the differential matrix.** `REFS` gains a ref carrying an additional `:` — `ast_opaque_1:variant` — and the record factories are parameterized by the loop ref, so compound ids are `stage-1:${ref}` instead of hard-coding one base ref. The derivation is now driven on both sides of the record:

- PPTX reads the row through the full compound key `stage-1:ast_opaque_1:variant`;
- video derives its ref from the record id with `split(':').slice(1).join(':')`, and the preserved trailing colon flows through task lookup, gating, and lease acquisition. The colon ref's non-compound-id case is driven too — the derivation sees `ast_opaque_1:variant` and produces `variant`, and both implementations agree there as well.

Matrix sizes: ZIP 648 → 756, PPTX 9,072 → 10,584, video 9,072 → 10,584 combinations.

**2. The blob-less PPTX branch asserts the exact permitted outcome.** Every divergent comparison now deep-equals the current implementation's outcome against a capture of `frozenPptxRepaired` — byte-identical to the verbatim frozen PPTX plus exactly the documented one-line blob-presence guard (the guard `main`'s video path already had). The outcome record carries the returned bytes, blob identity by reference, and the complete ordered lease/Dexie/fetch sequence including arguments, so the single equality pins the exact fallback result, identity where observable, and the full post-divergence trace together. The verbatim frozen adapter throws at these combinations — that is the divergence — so a repaired variant is the only way to state what the permitted result must be; it is labeled as not part of the frozen set and is the only relaxation.

The divergence total is re-derived for the widened ref set: 756 blob-less staged combinations − 42 that return pooled bytes before the row read (3 non-concrete refs × 1 pool-url case × 2 fetch outcomes the frozen `fetchBlob` accepts × 7 task states that still resolve to a URL) = **714**, asserted with the formula documented in the file.

**The extension bites** — mutation checks against the new cases, each reverted after running:

| mutation | result |
|---|---|
| compound-id derivation → `split(':')[1]` (drops the trailing colon) | red on colon-ref video lease/fetch trace; green after revert |
| blob-less row guard removed | red on the repaired-PPTX outcome comparison (result and trace); green after revert |
| divergence count `+1` (715 vs actual 714) | red; green after revert |

And no new divergence surfaced: the widened matrix reproduces the frozen behavior everywhere except the documented blob-less PPTX combinations.

**Reproducible output** (Node 22, `VITEST_MAX_WORKERS=4`, `NODE_OPTIONS=--max-old-space-size=1024`):

```text
$ pnpm exec vitest run tests/media/resolve-stored-bytes-equivalence.test.ts
 Test Files  1 passed (1)
      Tests  3 passed (3)
   Duration  1.08s

$ pnpm exec vitest run tests/media/resolve-stored-bytes.test.ts tests/export tests/video-export tests/media
 Test Files  65 passed | 1 skipped (66)
      Tests  912 passed | 5 skipped (917)
   Duration  7.14s
```

`pnpm check`, `pnpm lint`, and `tsc --noEmit` are clean at this head.

cosarah (approved):
LGTM

PR #1117, opened 2026-08-13
feat(dsl): standardize the asset manifest and converge the export paths (#1007 part 3)

wyuc:
Lands epic #1007 part 3: the standardized asset manifest and the export convergence on top of it. Stacked on #1101 (part 2c) and #1116 (part 3a, the shared byte resolver); rebased onto `main` after both merged.

**The manifest** lives in `@openmaic/dsl` (`asset-manifest.ts`, dsl 0.9.0 → 0.10.0): `enumerateAssetManifest(document, { metadata? })` is an IO-free enumeration of every media reference a document touches — allocated ids, legacy `gen_*` placeholders, and concrete URLs verbatim — one entry per distinct `(ref, kind)` pair in deterministic first-occurrence order, with kind and optional metadata. This settles #779's open question 4: the manifest is the id-based reference enumeration with metadata, explicitly not content hashes and not a resolution result. The `(ref, kind)` pairing matters because `AssetRef` is an unconstrained string alias: one ref can legally occur as image/video media and as narration audio, and consumers that partition entries by kind would otherwise silently drop a role. The optional metadata callback is caller-controlled enrichment; the enumeration itself is pure over the document and the callback must not mutate it. A new `@openmaic/exporter` package was deliberately not created. Media-role classification is single-sourced: the DSL exports a canonical read-only slide-media slot descriptor, and the app's slot walker and ownership/ref collection layer on top of it — adding a role is one edit, not three synchronized walkers.

**The convergence.** All three consumers derive their reference sets from the manifest:

- **ZIP**: reference sets come from the manifest instead of scanning the stage's Dexie tables. Behavior change, stated plainly: archives contain exactly the media the document references. Orphan rows no longer ride along, failed/byteless rows ship nothing, and referenced pool-only assets are now collected. The classroom ZIP intentionally excludes `Stage.whiteboard` refs: the classroom format does not serialize that field and import cannot reconstruct it (scene whiteboards remain portable). Byte resolution is pool-first, and after a pool miss it uses the referenced Dexie compatibility row as the legacy byte fallback (the frozen-base differential harness pins that relocation exactly: 702 helper-boundary null-to-bytes divergences, `ok`-row only, `row-blob` identity, identical call trace). Archive names never interpolate refs — sequential safe paths (`media/asset-<n>.<ext>`, `audio/audio-<n>.<ext>`) with the original ref preserved through an explicit `sourceRef` mapping on every independently indexed media entry: generated media assets, poster assets, primary narration, and legacy URL narration (the legacy URL itself is the entry's sourceRef). Extensions are allowlisted per kind. The one exception is the legacy sibling poster byte copy (`media/asset-<n>.poster.<ext>`, written next to its video when the video record still carries the pre-pool poster bytes): it is not an independently referenced document asset, so it has no mediaIndex entry or sourceRef of its own — its identity is derivable from its parent video entry, and import reconstructs it by sibling-path derivation from that entry, reusing the poster's own indexed allocation when one exists.
- **Video**: reference sets from the manifest; scene-scoped elementId→mediaRef bridge semantics preserved. Audio loading is speech-action-only — the manifest acts as a membership gate while `speechPairs` order is retained — slide audio is never loaded or duration-probed, and evicted speech rows (empty local blob, valid `ossKey`, stored duration) are retained so the CDN fallback still runs.
- **PPTX**: media reference candidates derive from the manifest (element iteration remains for layout/coordinates/z-order/video binding); one resolver-backed helper resolves them; the old placeholder gate that skipped allocated `ast_*` refs is gone.

**Media coherence contract.** The export contract guarantees label coherence: archive paths, filename extensions, and serialized MIME metadata are mutually coherent and are a canonical function of the authoritative kind metadata. It intentionally does not guarantee that those labels match the payload bytes, because exporters neither inspect nor transcode bytes; a byte/kind mismatch is already-corrupt store state outside the export contract. This matches runtime and renderer behavior, which likewise trust authoritative kind metadata. A 42-cell coherence matrix (3 kinds × 6 metadata classes × 7 surfaces) enforces the contract end to end — canonical path + serialized metadata + import classification per cell — with a boundary pin and mutation evidence.

**Ownership and hygiene.** The manifest's `referenceCounts`, the app's exclusive-ownership consumer, and the video-export planner all key owners positionally or by `(id, kind)` rather than user-controlled strings — duplicate ids are structurally valid documents and shared refs used across kinds keep independent ownership. Because `AssetRef` is unconstrained, a ref can legally be `"__proto__"` or `"constructor"`: every ref-keyed lookup this change introduces is prototype-safe (Map / null-prototype containers, explicit membership, string validation; media-task lookup centralized in one own-property-checked `lookupMediaTask`, pinned by a `buildPptxBlob` regression with a re-keyed adversarial task).

**Review.** Nine single-vendor red-team rounds (Codex) with fix rounds to closure, including the three review points above (safe archive paths + sourceRef mappings; canonical slot descriptor; manifest-derived consumer sets) and the subsequent hardening rounds: extension/kind/MIME coherence (the 42-cell matrix), narration order, cross-kind planner ownership, and the coherence-contract boundary. The final convergence gate (round 9) passed with zero findings. The second review pass's two findings — the task-owned poster binding through the PPTX manifest guard, and legacy-narration sourceRef completeness — are fixed with red→green coverage at this head.

**Validation** (Node 22): TypeScript, Prettier, and ESLint clean (changed files: 0 errors, 0 warnings). The root export/video/media/import battery passes 985 tests (5 skipped), `@openmaic/dsl` passes 238, and the package-version gate passes at `@openmaic/dsl` 0.9.0 → 0.10.0. The frozen-base differential harness (from #1116, updated to the current call-site option sets) pins resolver equivalence at 702 / 714 / 0 permitted divergences across ZIP / PPTX / video.

Refs #1007

### DIFF
--- a/lib/export/classroom-zip-types.ts
+++ b/lib/export/classroom-zip-types.ts
@@ -136,7 +136,7 @@ export type ManifestAction = Omit<Action, 'audioId'> & {
 
 export interface MediaIndexEntry {
   type: 'audio' | 'image' | 'generated';
-  /** Original document ref for audio; ZIP paths and refs are separate namespaces. */
+  /** Original document ref; ZIP paths and refs are separate namespaces. */
   sourceRef?: string;
   mimeType?: string;
   format?: string;
--- a/lib/export/classroom-zip-types.ts
+++ b/lib/export/classroom-zip-types.ts
@@ -136,6 +136,8 @@ export type ManifestAction = Omit<Action, 'audioId'> & {
 
 export interface MediaIndexEntry {
   type: 'audio' | 'image' | 'generated';
+  /** Original document ref for audio; ZIP paths and refs are separate namespaces. */
+  sourceRef?: string;
   mimeType?: string;
   format?: string;
   duration?: number;
--- a/lib/export/classroom-zip-utils.ts
+++ b/lib/export/classroom-zip-utils.ts
@@ -216,11 +216,14 @@ export function rewriteAudioRefsToIds(
   return actions.map((action) => {
     if (action.type === 'speech' && 'audioRef' in action) {
       const { audioRef, ...rest } = action;
-      const mapped = audioRef
-        ? audioRefMap instanceof Map
-          ? audioRefMap.get(audioRef)
-          : (audioRefMap as Readonly<Record<string, unknown>>)[audioRef]
-        : undefined;
+      const mapped =
+        typeof audioRef === 'string'
+          ? audioRefMap instanceof Map
+            ? audioRefMap.get(audioRef)
+            : Object.hasOwn(audioRefMap, audioRef)
+              ? (audioRefMap as Readonly<Record<string, unknown>>)[audioRef]
+              : undefined
+          : undefined;
       const audioId = typeof mapped === 'string' ? mapped : undefined;
       return {
         ...rest,
--- a/lib/export/classroom-zip-utils.ts
+++ b/lib/export/classroom-zip-utils.ts
@@ -1,17 +1,19 @@
 import type { Action, DiscussionAction, SpeechAction } from '@/lib/types/action';
 import type { ManifestAction } from './classroom-zip-types';
-import { db } from '@/lib/utils/database';
+import { db, mediaFileKey } from '@/lib/utils/database';
+import type { AssetManifestEntry } from '@openmaic/dsl';
 import type { AudioFileRecord, MediaFileRecord } from '@/lib/utils/database';
 import type { Scene } from '@/lib/types/stage';
 import { resolveAudioBlob } from '@/lib/media/resolve-audio-bytes';
 import { fetchMediaUrl } from '@/lib/media/fetch-media-url';
 import { mapWithConcurrency } from '@/lib/media/convert-legacy-asset-refs';
-import { mediaRefFromRecordId, resolveStoredBytes } from '@/lib/media/resolve-stored-bytes';
+import { resolveStoredBytes } from '@/lib/media/resolve-stored-bytes';
 
 // ─── Export: Collect Media ─────────────────────────────────────
 
 export interface CollectedAudio {
   zipPath: string;
+  sourceRef: string;
   record: AudioFileRecord;
 }
 
@@ -21,30 +23,30 @@ export interface CollectedMedia {
   elementId: string;
 }
 
-export async function collectAudioFiles(scenes: Scene[]): Promise<CollectedAudio[]> {
-  const audioIds = new Set<string>();
-  for (const scene of scenes) {
-    for (const action of scene.actions ?? []) {
-      if (action.type === 'speech' && (action as SpeechAction).audioId) {
-        audioIds.add((action as SpeechAction).audioId!);
-      }
-    }
-  }
+/**
+ * Collect the bytes of every audio entry selected by the classroom export.
+ * That set includes speech narration and reconstructable slide-audio refs, so
+ * no table scan runs here and an orphan audio row cannot ride into the archive.
+ * Bytes come only from the shared resolver (pool-first, with the
+ * compatibility-row fallback inside it); the row read here supplies the
+ * archive's format/duration/voice metadata.
+ */
+export async function collectAudioFiles(
+  entries: readonly AssetManifestEntry[],
+): Promise<CollectedAudio[]> {
   const collected: CollectedAudio[] = [];
-  for (const audioId of audioIds) {
-    const record = await db.audioFiles.get(audioId);
+  for (const entry of entries) {
+    const audioId = entry.ref;
     // The pool answers first: after a stable-id regeneration whose mirror write
     // failed, the row holds the superseded narration.
     const blob = await resolveAudioBlob(audioId);
     // A row with no usable bytes -- an evicted row (empty blob, no pool
-    // resolve) -- must not ship an empty audio file. It produces no zip
-    // path, so the URL rescue in collectLegacyAudioForExport sees the id as
-    // missing and fetches the live co-present URL instead.
-    if (blob && blob.size > 0) {
-      const ext = record?.format || 'mp3';
-      const resolved = (record ? { ...record, blob } : { id: audioId, blob }) as AudioFileRecord;
-      collected.push({ zipPath: `audio/${audioId}.${ext}`, record: resolved });
-    }
+    // resolve) -- must not ship an empty audio file.
+    if (!blob || blob.size === 0) continue;
+    const record = await db.audioFiles.get(audioId);
+    const ext = record?.format || 'mp3';
+    const resolved = (record ? { ...record, blob } : { id: audioId, blob }) as AudioFileRecord;
+    collected.push({ zipPath: `audio/${audioId}.${ext}`, sourceRef: entry.ref, record: resolved });
   }
   return collected;
 }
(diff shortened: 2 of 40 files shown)


wyuc:
Codex review converged: 2 rounds. Round 1 caught the slide-audio gap (PPTAudioElement src was not enumerated, so manifest-driven collectors could never archive those bytes) — fixed in 9e8360a6 with manifest + test coverage. Round 2 is clean.

Local validation: root tsc / eslint / prettier clean; dsl 235 tests and 745 export/video/media tests green; package-version gate passes.

Note: stacked on #1116; CI runs after it retargets to main down the chain.

wyuc:
Requesting review. Orientation for the reviewer:

**What this is.** Epic #1007 part 3: the standardized asset manifest (`@openmaic/dsl` 0.10, `enumerateAssetManifest`) and the three export paths (classroom ZIP, PPTX, video) rebuilt on it and on the shared byte resolver from #1116. Stacked on #1101/#1116, rebased onto current `main`; CI green at this head.

**Pre-review state.** Five single-vendor red-team rounds (Codex) with fix rounds to closure before requesting review; zero open findings. Every behavioral fix landed with red→green proof. The full round-by-round account and all stated behavior changes are in the PR description.

**Where to look hardest:**

1. **ZIP semantics** — archives now contain exactly the manifest-referenced media; the compatibility row is a pool-miss byte fallback (relocation pinned by the harness at exactly 702 null→bytes divergences, `ok`-row only, `row-blob` identity, identical call trace); stage whiteboards are excluded as unreconstructable.
2. **The manifest contract** — one entry per `(ref, kind)` pair because `AssetRef` is unconstrained and a ref can span kinds; owner keys are traversal positions, not user ids, in both the DSL and the app ownership consumer.
3. **The ninth commit** (`fix(media): harden ref-keyed lookups...`) — prototype-safety across every ref-keyed lookup the branch introduces, incl. the centralized `lookupMediaTask`; a `buildPptxBlob` regression drives a re-keyed `__proto__` placeholderRef end to end.

Reproduce locally: `pnpm exec vitest run tests/media/resolve-stored-bytes-equivalence.test.ts` (divergence derivations are in-file), plus `tests/export tests/video-export tests/media tests/import` and the `@openmaic/dsl` package suite.

cosarah:
Two remaining review points:

1. The DSL manifest walker duplicates the media-role classification already maintained in `lib/media/slide-media-slots.ts` and `lib/media/collect-stage-asset-refs.ts`. Adding a new media role now requires synchronized changes in multiple walkers; consider exposing a canonical read-only slot descriptor and layering the app-side writer/category projection on top.

2. `AssetRef` is an unconstrained string, but ZIP paths interpolate raw refs (`audio/${ref}.${ext}` and `media/${ref}.${ext}`). A ref containing `..` is normalized by JSZip, while `mediaIndex` retains the unsanitized path, so import cannot find the entry. Generate safe archive names independently of refs and preserve the original ref through an explicit `sourceRef` mapping for every media kind.

Scope clarification: I am withdrawing the earlier export-convergence point. For this PR, PPTX convergence is limited to shared resolver semantics, while video may still walk speech actions for timeline-specific context and speech-only audio behavior. Any broader unification of those traversals can be handled separately.

wyuc:
All three points are addressed on head `b830b628` (CI green at this head).

**1. Canonical slot descriptor — done.** `@openmaic/dsl` exports a canonical read-only slide-media slot descriptor; the app's slot walker (`slide-media-slots.ts`), the ownership/ref collection (`collect-stage-asset-refs.ts`), and the manifest walker all derive from it. Roles: background, image, video-src, video-mediaRef, poster, audio. A parity test pins the three consumers to the same role set; adding a role is now one edit, not three synchronized walkers.

**2. Safe archive names + sourceRef — done, and taken further.** Archive names are sequential safe paths (`media/asset-<n>.<ext>`, `audio/audio-<n>.<ext>`) that never interpolate refs; `mediaIndex` keys are the safe paths; every media kind (media, posters, audio, legacy narration) preserves the original ref through an explicit `sourceRef` mapping and import remaps all of them. Extensions are allowlisted per kind with canonical fallbacks, so no metadata-derived component of an archive path can carry a separator or dot segment — including the video-export planner, which previously stripped punctuation only. Red→green for `../evil`, `a/b`, collision-shaped refs, and malicious `format`/MIME values; backward import of ref-interpolated archives is preserved.

**3. Manifest-derived consumers — done.** PPTX media reference candidates derive from the manifest (element iteration remains only for layout/coordinates/z-order/video binding); video narration ids derive from the manifest's audio entries as a membership gate, retaining `speechPairs` order, speech-only loading, and evicted-row CDN retention.

Your points also triggered a broader hardening pass on archive-path/metadata coherence, now enforced by a 42-cell matrix (3 kinds × 6 metadata classes × 7 surfaces) asserting canonical path + serialized metadata + import classification per cell, and by a stated coherence contract: labels are a canonical function of the authoritative kind metadata and mutually coherent; the exporter neither inspects nor transcodes payload bytes, so label↔byte truth is explicitly out of contract (pinned by a boundary test with mutation evidence). Cross-kind ownership in the video planner is now keyed by `(id, kind)` so a ref used as both audio and video keeps independent paths and presence.

Nine red-team rounds total; the final convergence gate passed with zero findings. Full account in the PR description. One process note: when satisfied, please submit an actual review (approve or findings) — issue comments alone don't move the PR out of `REVIEW_REQUIRED`.

cosarah:
Two findings remain on head `b830b628`:

1. **P1: task-generated poster fallback can now abort PPTX export.** When a video element has no explicit `poster` but its media task has `task.poster`, `resolveVideoMediaForElement` returns that runtime URL as `posterRef` while leaving `posterTask` undefined (`lib/media/media-task-resolution.ts`). The manifest guard in `lib/export/use-export-pptx.ts` then rejects the URL because it is not a document ref and there is no `task.objectUrl` exemption, so `buildPptxBlob` throws instead of using the existing runtime poster fallback. Please either carry a poster task binding for this case or explicitly allow the task-owned poster URL, and add coverage for an element with no poster plus a task-provided poster.

2. **P2: the stated explicit `sourceRef` contract is still partial.** Primary audio/media entries preserve `sourceRef`, but `legacyAudioMediaIndexEntry` serializes no source ref, and sibling poster files are written without their own `mediaIndex` entry/source mapping (`lib/export/classroom-zip-utils.ts`, `lib/export/use-export-classroom.ts`). Import currently reconstructs these indirectly from actions/slides, so either complete the explicit mapping for these cases or narrow the PR description to match the actual contract.

wyuc:
Both findings are addressed on head `168b4654` (rebased onto current `main`; CI green at this head).

**1. Task-owned poster fallback — fixed (your option 1).** `resolveVideoMediaForElement` now carries the poster task binding whenever `task.poster` is the effective poster — i.e. when the element has no poster of its own or its poster is an opaque ref; a concrete explicit poster stays element-owned. The task-owned poster URL therefore satisfies the guard's existing task-ownership exemption end to end; the guard predicate itself is unchanged (extracted verbatim into the exported `isPptxManifestForeignRef` so its rejection behavior stays directly testable). Coverage: an element with no poster plus a task-provided poster fails at `b830b628` with the guard's exact error and passes with the fix (video bytes and task poster embedded); a companion test pins that a genuinely unrelated URL is still rejected.

**2. sourceRef contract — completed for legacy narration, narrowed for poster siblings.** `legacyAudioMediaIndexEntry` now serializes an explicit source ref (the legacy URL itself — the natural source ref for legacy narration), and import's generic sourceRef handling round-trips it; the coherence matrix asserts this for every legacy audio cell. For poster siblings the claim is narrowed instead: a sibling (`media/asset-<n>.poster.<ext>`) is a legacy byte copy written next to its video, not an independently referenced document asset — a poster that IS a document asset already has its own indexed entry with a sourceRef, and giving the byte copy its own entry would require threading document access into the collector and would create duplicate allocations on import. Its identity derives from its parent video entry, and import reconstructs it by sibling-path derivation, reusing the poster's own indexed allocation when one exists. The PR description now states exactly this.

Battery at this head: 985 passed / 5 skipped scoped, `@openmaic/dsl` 238 passed, tsc / eslint / prettier clean.

Process note again: when satisfied, please submit an actual review (approve or findings) so the PR moves out of `REVIEW_REQUIRED`.

cosarah (approved):
LGTM

PR #1119, opened 2026-08-13
feat(openmaic-skill): add secondary-development (二开) flow

### LINKED ISSUE #1118: feat(openmaic-skill): add a secondary-development (二开) capability line
## Problem

The `openmaic` skill today only covers **using / generating** classrooms — its description triggers on "use OpenMAIC" / "generate a classroom", and Phase 0 branches into Live Demo vs. local setup. There is **no path** for a user who wants to extend / build on / customize OpenMAIC or consume the `@openmaic/*` SDK ("二次开发" / 二开). Such requests don't trigger the skill, and even when loaded, the SOP has no 二开 branch.

## Proposal

Add 二开 as a **3rd parallel top-level branch** in Phase 0 (alongside Live Demo / local), non-invasively:

- Phase 0 gets option 3 → loads a new router reference and skips setup/generation phases (mirrors the existing Live Demo short-circuit).
- 3 new references split by mental model: **router** + **product cookbook** + **SDK consumption**. The agent loads at most 2 (router + one leaf), never all at once.
- The existing 5 references and Core Rules are untouched; the use/generate flow is unchanged.

## What changes

- `skills/openmaic/SKILL.md` (modified): description gains 二开 trigger words (extend / build on / customize / SDK / secondary development / 二开); Phase 0 option 3 + short-circuit rule; the `accessCode` auto-shortcut now yields to 二开 intent so returning Live Demo users aren't mis-routed.
- `skills/openmaic/references/extend.md` (new): 二开 SOP router — charter, read-before-modify rules, dev environment (= local-deploy env, reused not duplicated), fork-vs-clone decision (conditional, not mandatory), route to cookbook or sdk, general gotchas.
- `skills/openmaic/references/extend-cookbook.md` (new): product-layer tasks with precise `file:line` entry points — swap/add provider, storage backend, branding/theme, new page/API route, embed renderer/editor.
- `skills/openmaic/references/extend-sdk.md` (new): consuming `@openmaic/*` in a standalone app — 6-package quickref, exact import paths, storage subpath table (incl. the Runtime-barrel asymmetry), version pinning.

## Non-goals

- No changes to the existing 5 references or Core Rules.
- No changes to `packages/@openmaic/*` SDK source.
- No new tests (the repo has no skill-content tests; validation = link checks + the ClawHub publish dry-run in CI).

## Validation

- All `Load [references/...]` links resolve; frontmatter stays valid YAML; `name: openmaic` unchanged.
- All `file:line` entry points and package versions/symbols verified against source.
- Awaiting: the `publish-openmaic-skill.yml` preview job (`clawhub skill publish --dry-run`) as the final content gate.

CXuPercy:
Closes #1118.

## What

Adds a non-invasive **secondary-development (二开)** capability line to the `openmaic` skill — a 3rd top-level branch in Phase 0 (alongside Live Demo / local) for users who want to extend / build on / customize OpenMAIC or consume the `@openmaic/*` SDK. Previously these requests did not trigger the skill and had no SOP path.

## Design

- **Entry point = Phase 0 option 3**, mirroring the existing Live Demo short-circuit (load a reference, skip setup/generation phases). Zero changes to the use/generate flow.
- **3 new references** split by mental model, so the agent loads at most 2 (router + one leaf), never all at once:
  - `extend.md` — router: charter, read-before-modify rules, dev environment (= local-deploy env, reused not duplicated), conditional fork-vs-clone + CI, route, gotchas.
  - `extend-cookbook.md` — customizing the product: 5 task types with precise `file:line` entry points (provider / storage / branding / routes / embed renderer-editor).
  - `extend-sdk.md` — consuming `@openmaic/*` in a standalone app: 6-package quickref, exact import paths, storage subpath table (incl. Runtime-barrel asymmetry), version pinning.
- **Routing safety:** the `accessCode` auto-shortcut in Phase 0 now yields to 二开 intent, so a returning Live Demo user who wants to do 二开 is routed correctly instead of being sent back to Live Demo.

## Files

- `skills/openmaic/SKILL.md` (modified) — description trigger words + Phase 0 option 3 + short-circuit + accessCode exception.
- `skills/openmaic/references/extend.md` (new)
- `skills/openmaic/references/extend-cookbook.md` (new)
- `skills/openmaic/references/extend-sdk.md` (new)

## Non-goals

The existing 5 references, Core Rules, and `packages/@openmaic/*` source are untouched.

## Validation

- All `Load [references/...]` links resolve to real files; frontmatter stays valid YAML; `name: openmaic` unchanged.
- All `file:line` entry points and package versions/symbols verified against source.
- Locally tested.
- CI gate: the `publish-openmaic-skill.yml` preview job (`clawhub skill publish --dry-run`) runs on this PR.


wyuc (changes requested):
Thanks for adding a dedicated secondary-development path. The overall routing direction looks good, but several instructions are currently inaccurate enough to send users down broken paths:

1. **New-provider `.env.local` setup is incomplete.** The cookbook says that after adding a new ID to `BuiltInProviderId` and `PROVIDERS`, its key can be configured through either `.env.local` or `server-providers.yml`. However, env loading is controlled by the fixed `LLM_ENV_MAP` in `lib/server/provider-config.ts`. Without adding the new provider there, `<NEW_PROVIDER>_API_KEY` is never loaded. Please either add that step explicitly or state that arbitrary new providers must use `server-providers.yml`.

2. **The storage-backend description does not match the implementation.** The cookbook describes `lib/persistence/asset-byte-store.ts` as switching between “S3 vs local bytes” and frames the task as “PG / S3 / Local”. The actual fallback is `PgAssetByteStore`, not a local-file store, and the storage package does not expose a local-file backend. Please describe the supported topology accurately: browser-local storage by default; server persistence through PostgreSQL; and asset bytes in PostgreSQL or S3.

3. **The `dist/` guidance incorrectly claims to apply to both paths.** A source checkout/workspace needs postinstall or package builds to produce its ignored `dist/`, but published npm packages ship built `dist/` artifacts. Saying that every package “ships source only” under “Apply To Both Paths” is false for standalone SDK consumers. Please scope this gotcha to monorepo/fork development.

4. **A monorepo fork cannot generally be consumed as an individual package through npm `overrides`/`resolutions`.** A Git dependency points at the repository root package, not `packages/@openmaic/<name>`. The reliable options here are packing the modified subpackage, publishing it under a private/different package identity, or otherwise producing an installable package artifact. Please remove or qualify the fork override recommendation.

Two smaller accuracy issues are also worth correcting:

- Fork pushes do not universally make both publish workflows fail. Their triggers depend on branch, path, and dispatch conditions; ordinary feature-branch pushes may not run them at all.
- This branch is currently seven commits behind `main`, and at least one promised precise line reference has already drifted (the Bedrock region helper is now around line 1760 rather than 1746). Please sync and revalidate the `file:line` references before merging, or prefer symbol-oriented references where possible.

CI is green, but these are content-correctness issues that the publish preview does not validate.

CXuPercy:
Thanks @wyuc — all fair points; each was a real accuracy bug. Force-pushed fixes (single squashed commit, now on top of latest `main`).

**1. New-provider env wiring.** You're right that env loading is gated by `LLM_ENV_MAP` in `lib/server/provider-config.ts`. Cookbook Task 1 now spells it out: to use `.env.local` for a new provider you must add a `PREFIX: 'id'` entry to `LLM_ENV_MAP` (otherwise `<PREFIX>_API_KEY` is silently ignored), **or** configure it via `server-providers.yml` (keyed by id, no map change needed). Bedrock's separate `applyBedrockProviderConfig` path is noted too.

**2. Storage topology.** Corrected. Task 2 now reads: browser-local by default; opt into server persistence via `NEXT_PUBLIC_PERSISTENCE=1` (client-side `bootstrap.ts` → HTTP stores → `/api/persistence`); the server persists documents + runtime to PostgreSQL, and asset bytes to PostgreSQL by default (`PgAssetByteStore`) or S3 when `ASSET_S3_BUCKET` is set. Explicitly states there is **no local-file backend**.

**3. `dist/` scope.** Scoped. The "ships source only / dist gitignored" gotcha is now tagged `[product/fork]` with an explicit note that published npm packages ship built `dist/`, so it doesn't apply to the SDK path. All gotchas are now tagged `[product/fork]` / `[SDK]` / `[both]` rather than "applies to both".

**4. Fork → package consumption.** Removed the git-URL `overrides`/`resolutions` recommendation. extend-sdk now says: produce an installable artifact for just that subpackage (`npm pack` + install the tarball, or publish under a private name), and explicitly warns that a Git dependency / `overrides` pointing at the repo won't resolve — a Git dep points at the root package, not `packages/@openmaic/<name>`.

**Smaller points:**
- **CI triggers qualified.** The fork-push-CI note now states the two workflows' actual triggers (`publish-packages.yml`: push to main + a `packages/@openmaic/*/package.json` change; `publish-openmaic-skill.yml`: PR/push to main touching `skills/openmaic/**`), and that ordinary feature-branch pushes that don't match won't run them — so disabling is conditional, not universal.
- **Drift / file:line.** Synced to latest `main` (was 7 behind; only `lib/ai/providers.ts` had moved, which is why the Bedrock line shifted). Per your suggestion, I converted all entry points to **symbol-oriented** references (file + symbol name, no line numbers), so they don't drift going forward.

CI is re-running on the updated commit. Happy to adjust further.

wyuc (changes requested):
Thanks for the thorough revision — I verified that the four primary findings from the previous review are now substantially addressed. Two factual inaccuracies remain:

1. **An invalid `ASSET_S3_BUCKET` does not fall back to PostgreSQL.** `configuredS3Bucket()` throws for an invalid bucket name, and `lazyAssetByteStore()` propagates that failure to the asset request while clearing the cached promise for a retry. Only an **unset/empty** bucket selects `PgAssetByteStore`; a valid bucket selects S3; an invalid bucket makes asset operations fail. Please change the `unset/invalid ⇒ PgAssetByteStore` statement accordingly.

2. **A matching pull request does not reach the token-bearing ClawHub publish job.** `publish-openmaic-skill.yml` is triggered for a PR touching `skills/openmaic/**`, but on a PR it runs the compatibility/preview jobs and skips `publish`; the token-bearing job only runs on a push, or a non-dry-run manual dispatch from `main`. The current text says matching pushes/PRs run a publish job and fail for missing `CLAWHUB_TOKEN`, which is not true for PRs. Please distinguish workflow triggering from the conditions that actually enable the publish job.

Once these two sentences are corrected, the issues from my review will be resolved.

CXuPercy:
Both points fixed and verified against source this time (a6d52f11):

**1. `ASSET_S3_BUCKET` semantics.** Rewritten as the strict three-way behavior: **unset/empty** ⇒ `PgAssetByteStore`; **valid** ⇒ S3; **invalid** ⇒ asset operations **fail** — `configuredS3Bucket()` throws and there is no fallback to PG. Also noted `lazyAssetByteStore()`'s no-poisoned-singleton behavior (failed construction isn't cached, so the next asset request retries, and only asset traffic is affected — document/runtime requests keep working).

**2. PR vs publish-job gating.** The CI note now separates the two layers explicitly:
- *Trigger layer*: `publish-packages.yml` — push to `main` + a `packages/@openmaic/*/package.json` change only (PRs don't trigger this workflow at all); `publish-openmaic-skill.yml` — PR or push to `main` touching `skills/openmaic/**`.
- *Job layer*: on a **PR**, `publish-openmaic-skill.yml` runs only the bash-3 compatibility + preview (dry-run) jobs — no token needed; the token-bearing `publish` job (`if: github.event_name == 'push' || (workflow_dispatch && ref==main && !dry_run)`) runs **only on push** (or a non-dry-run dispatch from `main`). Same push-only gating for `publish-packages.yml` (`if: push || inputs.dry_run == false`).

So the doc now says token red-X's can only come from pushes that match the triggers, never from PRs.

I also did a full self-audit of every remaining factual claim in the three new references against current `main` (toolchain pins, postinstall/build wiring, workflow conditions, gitignore, tsconfig alias, renderer peer set, package exports, symbols like `BuiltInProviderId`/`PROVIDERS`/`PRESET_THEMES`, the persistence route's `PgRuntimeStore`/`PgDocumentStore`/`PgAssetStore` wiring) — all verified at HEAD.

wyuc (approved):
All review findings have been addressed and verified against the implementation. The updated documentation now accurately describes provider env wiring, persistence backends, package distribution, fork consumption, S3 validation behavior, and publish workflow gating.

PR #1132, opened 2026-08-17
fix(renderer): preserve literal text line breaks

xuyuanwei678:
## Summary
- render literal line endings in text elements and shape labels with `white-space: pre-line`
- keep the style scoped to static content so injected editors are unaffected
- bump `@openmaic/renderer` to `0.1.1`

## Validation
- `pnpm --filter @openmaic/renderer test`
- `pnpm --filter @openmaic/renderer typecheck`

### DIFF
--- a/packages/@openmaic/renderer/src/elements/text/BaseTextElement.tsx
+++ b/packages/@openmaic/renderer/src/elements/text/BaseTextElement.tsx
@@ -23,6 +23,7 @@ export function BaseTextElement({ elementInfo, target, renderContent }: BaseText
       style={{
         position: 'relative',
         pointerEvents: target === 'thumbnail' ? 'none' : undefined,
+        whiteSpace: 'pre-line',
       }}
       dangerouslySetInnerHTML={{ __html: elementInfo.content }}
     />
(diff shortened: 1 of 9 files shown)


packages/@openmaic/renderer/src/elements/text/BaseTextElement.tsx
```diff
@@ -23,6 +23,7 @@ export function BaseTextElement({ elementInfo, target, renderContent }: BaseText
       style={{
         position: 'relative',
         pointerEvents: target === 'thumbnail' ? 'none' : undefined,
+        whiteSpace: 'pre-line',
```
wyuc: [P3] Consider limiting `pre-line` to plain-text content. The importer and editor currently serialize compact HTML, so this is unlikely on the main paths, but `content` is still a public HTML-string contract. External consumers or generated DSL may provide formatted markup such as `<p>A</p>\n<p>B</p>`; Chromium preserves that inter-tag whitespace here and adds an anonymous line box. Either normalize/document compact HTML as an invariant, or apply newline preservation only when the content is plain text. The same consideration applies to shape labels.

wyuc (commented):
Non-blocking compatibility note.

xuyuanwei678:
Thanks for this feedback — it has been addressed in the latest commits.

Resolution:
- `pre-line` is now used only for plain-text content without HTML markup, so genuine literal line endings are preserved.
- Content containing HTML tags, comments, or declarations keeps the default whitespace behavior, preventing source-formatting newlines in markup such as `<p>A</p>\n<p>B</p>` from becoming extra visible lines.
- Empty paragraphs (`<p></p>`) continue to retain the blank lines created in the editor through the static renderer placeholder rule.

The update also adds coverage for text and shape labels, plus boundary cases for tags, comments, doctypes, HTML entities, custom elements, and ordinary text containing `<`.

Thanks again for highlighting this compatibility issue!

wyuc (approved):

PR #1133, opened 2026-08-17
feat(pi): add Native Child Web Search

LING-6150:
## Summary

- add server-only `web_search` to the Native Child tool inventory
- reuse OpenMAIC's existing registered Web Search provider system and return results through the same Child tool-result continuation
- preserve OpenMAIC as the sole Agent model/provider-resolution authority while keeping the independently selected Web Search provider responsible for retrieval

## Behavior

- carries the Toolbar-selected Web Search provider identity and narrow configuration fields through the Pi request, while preserving server-managed configuration precedence and secret containment
- resolves the selected registered provider through the existing Web Search configuration path and executes it through `searchWeb()`
- keeps the tool registered for valid Native Child delegations; missing provider configuration returns the existing explicit `not_configured` tool result
- validates Native arguments before Pi coercion, including strict plain-object, own-property, prototype, field, query, and `maxResults` checks
- propagates caller cancellation through the Native tool and Web Search adapters while preserving the search-service timeout and Native deadline boundaries
- treats source text as untrusted tool-result data and returns normalized HTTP(S) source URLs to the same Child continuation
- adds an inventory-gated Native prompt policy for current/recent information or explicit external-verification requests when attached evidence is insufficient; ordinary course content and timeless facts remain excluded unless verification is explicitly requested
- retains lower-level provider-transport, deadline, cancellation, duplicate-call, and protocol-anomaly bounds without adding a fixed search/action business budget
- emits no classroom action event and adds no Browser delivery, acknowledgement, or rendering contract

## Reviewer-request closure

- replaced the parallel Responses-only provider/config path with the existing registered Web Search provider system
- removed the Native-Web-Search-specific feature flag and dedicated Responses Web Search environment surface
- removed fixed search/action/tool-execution budgeting from this capability
- verified the real user path with a Toolbar-selected Bocha provider: a natural current-information request invoked `web_search`, continued in the same Native Child, and returned the correct date with a complete official Apple Developer URL

## Validation

- targeted Native/Pi/Web Search regression tests passed
- Prompt policy tests: 27/27 passed
- TypeScript, ESLint, Prettier, and `git diff --check` passed
- independent read-only code reviews and targeted closure reviews passed
- GitHub CI: all 5 checks passed on `a5d170429b2dbfe307eab66a809c0f427d4e6958`
- real-provider E2E GIF recorded; evidence will be attached in the reviewer follow-up

## Scope

This PR does not add Native whiteboard tools, RuntimeStore coupling, Browser Web Search, Browser ACK/receipt, a second Agent model/provider path, server-side pricing, benchmark usage accounting, or later parity work. Legacy Child, Director Web Search, Spotlight, Browser, and whiteboard behavior remain unchanged.

### DIFF
--- a/components/chat/use-chat-sessions.ts
+++ b/components/chat/use-chat-sessions.ts
@@ -160,6 +160,7 @@ export type ChatRequestTemplate = {
   piSessionBoundary?: PiSessionBoundaryContext;
   webSearchProviderId?: WebSearchProviderId;
   webSearchApiKey?: string;
+  webSearchBaseUrl?: string;
   webSearchModelId?: string;
   baiduSubSources?: BaiduSubSources;
 };
@@ -220,12 +221,16 @@ export function withPiWebSearchSettings<T extends ChatRequestTemplate>(
   const request = { ...requestTemplate };
   delete request.webSearchProviderId;
   delete request.webSearchApiKey;
+  delete request.webSearchBaseUrl;
   delete request.webSearchModelId;
   delete request.baiduSubSources;
   return {
     ...request,
     webSearchProviderId: providerId,
     ...(providerConfig?.apiKey ? { webSearchApiKey: providerConfig.apiKey } : {}),
+    ...(providerConfig?.baseUrl && !providerConfig.isServerConfigured && providerId !== 'searxng'
+      ? { webSearchBaseUrl: providerConfig.baseUrl }
+      : {}),
     ...(providerId === 'claude' && providerConfig?.modelId
       ? { webSearchModelId: providerConfig.modelId }
       : {}),
--- a/components/chat/use-chat-sessions.ts
+++ b/components/chat/use-chat-sessions.ts
@@ -19,7 +19,7 @@ import type { UIMessage } from 'ai';
 import type { ThinkingConfig } from '@/lib/types/provider';
 import { useStageStore } from '@/lib/store';
 import { useCanvasStore } from '@/lib/store/canvas';
-import { useSettingsStore } from '@/lib/store/settings';
+import { useSettingsStore, type SettingsState } from '@/lib/store/settings';
 import { useUserProfileStore } from '@/lib/store/user-profile';
 import { useAgentRegistry } from '@/lib/orchestration/registry/store';
 import { useI18n } from '@/lib/hooks/use-i18n';
@@ -38,6 +38,7 @@ import { createLogger } from '@/lib/logger';
 import { isPiChatEnabled } from '@/lib/config/feature-flags';
 import type { CleanupSource } from '@/lib/playback/auto-resume';
 import { nanoid } from 'nanoid';
+import type { BaiduSubSources, WebSearchProviderId } from '@/lib/web-search/types';
 
 const log = createLogger('ChatSessions');
 const SOFT_CLOSE_TIMEOUT_MS = 15_000;
@@ -157,6 +158,10 @@ export type ChatRequestTemplate = {
   thinkingConfig?: ThinkingConfig;
   directorState?: DirectorState;
   piSessionBoundary?: PiSessionBoundaryContext;
+  webSearchProviderId?: WebSearchProviderId;
+  webSearchApiKey?: string;
+  webSearchModelId?: string;
+  baiduSubSources?: BaiduSubSources;
 };
 
 /**
@@ -198,6 +203,34 @@ export function withPiInclassWhiteboardTools<T extends ChatRequestTemplate>(requ
       piEnableWhiteboardTools: true,
     },
   };
+}
+
+type PiWebSearchSettings = Pick<
+  SettingsState,
+  'webSearchProviderId' | 'webSearchProvidersConfig' | 'baiduSubSources'
+>;
+
+/** Snapshot only the selected provider fields accepted by the classroom resolver. */
+export function withPiWebSearchSettings<T extends ChatRequestTemplate>(
+  requestTemplate: T,
+  settings: PiWebSearchSettings,
+): T {
+  const providerId = settings.webSearchProviderId;
+  const providerConfig = settings.webSearchProvidersConfig[providerId];
+  const request = { ...requestTemplate };
+  delete request.webSearchProviderId;
+  delete request.webSearchApiKey;
+  delete request.webSearchModelId;
+  delete request.baiduSubSources;
+  return {
+    ...request,
+    webSearchProviderId: providerId,
+    ...(providerConfig?.apiKey ? { webSearchApiKey: providerConfig.apiKey } : {}),
+    ...(providerId === 'claude' && providerConfig?.modelId
+      ? { webSearchModelId: providerConfig.modelId }
+      : {}),
+    ...(providerId === 'baidu' ? { baiduSubSources: { ...settings.baiduSubSources } } : {}),
+  } as T;
 }
 
 export function shouldAwaitPresentationAction(actionName: string): boolean {
@@ -1151,9 +1184,13 @@ export function useChatSessions(options: UseChatSessionsOptions = {}) {
         const piRequestTemplate = firstRequestContext
           ? { ...requestTemplate, storeState, piSessionBoundary: firstRequestContext }
           : { ...requestTemplate, storeState };
+        const piRequestWithWebSearch = withPiWebSearchSettings(
+          piRequestTemplate,
+          useSettingsStore.getState(),
+        );
         await runPiSingleRequest(
           sessionId,
-          withPiInclassWhiteboardTools(piRequestTemplate),
+          withPiInclassWhiteboardTools(piRequestWithWebSearch),
           controller,
           sessionType,
           createStatelessStreamConsumer,
--- a/lib/chat/pi/tools/web-search.ts
+++ b/lib/chat/pi/tools/web-search.ts
@@ -391,6 +391,7 @@ async function executeNativeWebSearch(
       apiKey: config.apiKey,
       maxResults: opts.params.maxResults ?? 5,
       baseUrl: config.baseUrl,
+      ...(opts.signal ? { signal: opts.signal } : {}),
       ...(config.baiduSubSources ? { baiduSubSources: config.baiduSubSources } : {}),
       ...(config.claudeModelId ? { claudeModelId: config.claudeModelId } : {}),
     });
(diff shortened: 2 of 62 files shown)


wyuc (changes requested):
Thanks — the Native Child tool-call/continuation path itself looks reasonable, but I think the integration should be simplified before we merge this.

1. **Reuse the existing Web Search provider system instead of creating a Responses-only path.**

   OpenMAIC already has `WEB_SEARCH_PROVIDERS`, `searchWeb()`, `resolveClassroomWebSearchConfig()`, and a dedicated Web Search provider selector in the Toolbar (`webSearchProviderId` plus its provider config). This PR currently bypasses that abstraction and introduces a second source of truth through `RESPONSES_WEB_SEARCH_*`.

   The expected path should be:

   ```
   Toolbar-selected Web Search provider/config
     -> Pi request
     -> server-side Web Search config resolution
     -> Native Child web_search
     -> searchWeb()
     -> selected registered provider
   ```

   The Toolbar's LLM provider/model selection should continue to drive the Agent Runtime model; the separately selected **Web Search provider** should drive the search backend. Server-managed provider config must remain authoritative, following the existing resolver rules.

2. **Please avoid a separate per-tool feature flag/config path.**

   I don't think we need `OPENMAIC_ENABLE_PI_NATIVE_CHILD_WEB_SEARCH` or a dedicated set of Responses Web Search env vars here. Once the Native runtime is enabled, `web_search` can be part of its tool inventory and resolve availability/configuration through the registered Web Search provider system. If the selected provider is not configured, returning the existing explicit `not_configured` tool result is sufficient.

   More generally, adding a feature flag/config surface for every Native tool will make the runtime unnecessarily difficult to operate and reason about.

3. **Do not make “at most N searches/actions” an upper-layer business policy.**

   Search should participate in the normal Native tool loop. The safety boundary should be the runtime's lower-level turn/transport/deadline/cancellation limits, rather than a separate fixed search/action budget. A hardcoded search count is arbitrary and may prevent legitimate multi-step research without providing a meaningful architectural guarantee.

4. **Please attach an end-to-end test GIF before merge.**

   The GIF should show the real user path: select/configure a Web Search provider in the Toolbar, ask a question that requires current information, observe the Native Child invoke search, and receive the streamed answer with source URL(s). This will verify the provider-selection plumbing and actual UX, which the current mocked tests do not cover.

So the main requested change is to treat this PR as adding a Native `web_search` tool on top of the existing Web Search abstraction, rather than introducing a parallel provider/configuration subsystem.

components/chat/use-chat-sessions.ts
```diff
@@ -200,6 +205,34 @@ export function withPiInclassWhiteboardTools<T extends ChatRequestTemplate>(requ
...
+  const providerConfig = settings.webSearchProvidersConfig[providerId];
+  const request = { ...requestTemplate };
+  delete request.webSearchProviderId;
+  delete request.webSearchApiKey;
+  delete request.webSearchModelId;
+  delete request.baiduSubSources;
+  return {
+    ...request,
+    webSearchProviderId: providerId,
+    ...(providerConfig?.apiKey ? { webSearchApiKey: providerConfig.apiKey } : {}),
```
wyuc, line 230: Blocking: this request snapshot drops the selected provider baseUrl. The existing Toolbar configuration and /api/web-search path support allowlisted alternate endpoints, including the Bocha and MiniMax regional hosts, but the Pi route then resolves only the server or default endpoint. A valid Toolbar configuration can therefore send its credentials to the wrong host and fail. Please carry a dedicated Web Search base URL and validate and resolve it at the server boundary with the existing allowlist and managed-provider precedence; do not accept arbitrary client URLs.

lib/chat/pi/tools/web-search.ts
```diff
@@ -276,3 +338,172 @@ export function buildDirectorWebSearchTool(opts: {
...
+        retrievedAt,
+        sourceCount: 0,
+        sources: [],
+      },
+      isError: true,
+    };
+  }
+
+  try {
+    const result = await (opts.search ?? searchWeb)({
```
wyuc, line 388: Blocking: the signal is checked before and after searchWeb, but it is never passed into the provider request. executeWithAbort lets the Native run return promptly, while the underlying HTTP or SDK search can keep running after user cancellation or the 60-second Child deadline, consuming connections and provider quota. Please thread the AbortSignal through searchWeb and the adapters, and test that the provider request itself observes cancellation rather than only rejecting a wrapper after a late provider completion.

wyuc (changes requested):
Thanks — the revised architecture now matches the direction we discussed: Native Child reuses the registered Web Search provider system, follows the Toolbar-selected Web Search provider, removes the per-tool feature flag, and does not impose an upper-layer search/action count policy.

Two blocking correctness issues remain in the integration, called out inline below. Please also update the PR body: it still describes the old Responses-only, default-OFF, attempt-limit design, so it is currently the opposite of the implementation. The requested end-to-end GIF is also still missing; please attach the real Toolbar provider selection → Native Child search → streamed answer with source URL flow.

After those changes, please update the branch from main because it is currently behind, let CI rerun on the new head, and re-request review.

LING-6150:
@wyuc   Updated the branch in commit `4bb602`.

The two integration issues found during closure are now fixed:

1. The Toolbar-selected Web Search configuration now carries the selected
provider base URL through the Pi request for unmanaged providers. Client-provided
URLs are validated against the existing server allowlist before the Pi loop
starts, while server-managed credentials and base URLs remain authoritative.

2. Caller/request cancellation now propagates from the Native tool through
`searchWeb()` into every registered provider adapter and the underlying HTTP or
AI SDK request. An aborted search remains caller-owned cancellation rather than
being translated into an ordinary tool failure.

This continues to use the existing Web Search provider selector and
`searchWeb()` abstraction. It does not introduce a Native-specific provider
path, per-tool feature flag, or search/action budget.

Validation completed:

- 7 targeted files / 113 tests passed
- TypeScript passed
- ESLint passed
- Prettier passed
- `git diff --check` passed

The real-provider E2E GIF is being handled separately.

LING-6150:
@wyuc I have completed the requested amendments and the real-provider E2E verification on the updated head `a5d1704`.

## Reviewer-request closure

The implementation now:

- reuses the existing registered Web Search provider system through `searchWeb()`;
- uses the independently Toolbar-selected Web Search provider with server-managed configuration precedence;
- removes the Responses-only provider/config path;
- removes the Native-Web-Search-specific feature flag;
- removes the fixed search/action/tool-execution budget;
- preserves same-Child tool-result continuation and the lower-level transport, deadline, cancellation, duplicate-call, and protocol bounds.

## Prompt amendment and A/B evidence

While running the requested natural-user-path E2E, I found one additional tool-selection gap.

Using the same classroom, Agent model, Toolbar-selected Bocha provider, and natural Apple verification question, I ran a controlled A/B test:

- On the committed head without the Prompt amendment, `web_search` was registered, but the Native Child stated that it could not verify the website. No `/v1/web-search` request occurred.
- With the Prompt amendment, the same natural request produced a real `/v1/web-search` request, waited for the tool result, continued in the same Native Child, and returned the correct dates with a complete official Apple Developer URL.

The isolated Prompt amendment is limited to:

- `lib/chat/pi/prompts.ts`
- `tests/lib/chat/pi/prompts.test.ts`

The policy is inventory-gated and deliberately narrow:

```text
attached evidence is insufficient
AND
(current/recent information OR explicit external-verification request)
→ call web_search before answering
```

Ordinary course-content questions and timeless facts are explicitly excluded unless the user asks for external verification.

This amendment does not change:

- provider or configuration authority;
- tool authorization;
- runtime accounting;
- Browser behavior;
- the same-Child continuation contract.

## Real-provider test environment

- Agent model: `gpt-5.4`, resolved through the existing OpenMAIC model/provider path;
- Web Search provider: Toolbar-selected `Bocha`, resolved through the existing server-managed Web Search configuration path;
- PR head: `a5d17042`.

No runtime-specific model/provider path or client-supplied credential was introduced for this test.

## Real-provider E2E

The GIF below shows the final natural user path on the current PR head:

```text
Toolbar-selected Bocha provider
→ natural external-verification question
→ Native Child web_search
→ searchWeb()
→ same-Child continuation
→ correct date and complete official Apple Developer URL
```

PR1133-Native-Web-Search-E2E-HD

The natural user question shown in the GIF is:

> According to Apple’s official Developer page, when was WWDC 2026 held?  
> Please verify it online. Reply only with the dates and one complete Apple source URL.

The Native Child returned:

> June 8–12, 2026  
> https://developer.apple.com/cn/wwdc26/

## Additional natural-query verification

I also tested a second natural current-information question on the same final PR head, Agent model, and Toolbar-selected Bocha provider:

> Who won the 2026 FIFA World Cup final, and what was the score? Please verify it online and include one complete official FIFA source URL.

The Native Child returned:

> Spain beat Argentina 1–0 after extra time.

It also returned an official FIFA source URL:

https://www.fifa.com/it/articles/classifica-finale-mondiali-2026

This result is consistent with FIFA’s official
[match report](https://www.fifa.com/en/tournaments/mens/worldcup/canadamexicousa2026/articles/spain-argentina-final-report-highlights)
and
[final tournament standings](https://www.fifa.com/en/tournaments/mens/worldcup/canadamexicousa2026/articles/final-tournament-standings).

This second check used no explicit `MUST call web_search` instruction, tool name, or tool-specific query.

PR1133 FIFA natural-query verification

During an earlier exploratory run before the final Prompt amendment, a differently worded FIFA query successfully invoked the Web Search provider, but its top-ranked retrieval evidence did not contain the final result and therefore did not yield a reliable answer.

I did not add FIFA-specific prompting, query rewriting, provider fallback, or result-ranking logic. The final-head result above is a fresh real-provider verification using the same narrow, inventory-gated Native Web Search policy.

## Validation

- GitHub CI: **5/5 checks passed**
- Prompt policy tests: **27/27 passed**
- TypeScript: **PASS**
- ESLint: **PASS**
- Prettier: **PASS**
- `git diff --check`: **PASS**
- Targeted independent Prompt closure review: **PASS**

Could you please re-check the updated head and the E2E evidence? If everything looks good, this PR should be ready to merge.

wyuc (approved):
The original provider endpoint and cancellation blockers are resolved, the registered Web Search provider integration matches the agreed architecture, the real-provider E2E evidence is attached, and all checks pass on the main-synchronized head. The remaining Native prompt source-fidelity ambiguity is accepted as non-blocking for this scope and can be tightened separately.

PR #1136, opened 2026-08-18
fix(media): share legacy URL probe budget

### LINKED ISSUE #1129: [Task] Asset conversion follow-ups: ownership alignment, conditional document writes, aggregate probe budgets
Three items were deliberately deferred out of #1101 (asset reference conversion, #1007 part 2c). They are recorded here so the reasoning is not buried in a review thread.

## 1. Align document and asset ownership

#1101 converts legacy handles into allocated asset ids. Documents are stored globally with no partition, while the asset pool partitions entries by an asset principal. To keep the two consistent, the development authenticator now issues **one shared asset principal** for every request (`lib/persistence/server-auth.ts`), so any browser that can load a document can also resolve its assets — the same semantics the document store already had.

That is a stopgap that matches today's development authenticator, not a model. The client-supplied `x-learner-key` never provided isolation (the module's own docstring says so), so nothing real is given up, but production authentication must derive the principal from server-controlled claims. At that point the question returns properly: **if assets are partitioned per principal, documents must be too**, or a shared document will again reference assets its reader cannot resolve.

The work: derive the asset principal from verified session claims, and decide the document's partition in the same model — either documents become principal-scoped, or assets referenced by a shared document are owned by the document rather than by whoever minted them.

## 2. Conditional writes for the document store

Conversion is lazy: opening a document can rewrite and save it. #1101 narrows the overwrite window to a single load+save (the save-back reloads first, converts any newer content on top, and treats a missing document as a concurrent deletion), but it does not eliminate it — the HTTP document store has no compare-and-swap.

The work: add a revision or ETag to the document contract and make `saveDocument` conditional, so a write that lost a race fails instead of overwriting. This belongs to the store contract, not to the converter, which is why it was kept out of #1101.

## 3. Aggregate wall-clock budget for legacy URL probes outside the converter

The conversion pass bounds its own legacy URL work with a shared 60-second budget on top of the per-probe 15-second timeout (`URL_PROBE_BUDGET_MS` in `lib/media/convert-legacy-asset-refs.ts`).

Two other paths fetch legacy URLs and are bounded only by concurrency (4) and the per-probe timeout, with no aggregate deadline:

- `collectLegacyAudioForExport` (`lib/export/classroom-zip-utils.ts`) — ZIP export of a document whose narration could not convert.
- the legacy-URL fallback in `lib/video-export-app/timeline-deps.ts`.

Worst case is therefore `ceil(uniqueUrls / 4) * 15s`: a document with many unique stalled legacy URLs turns best-effort recovery into a multi-minute stall. Both paths only run for documents that failed to convert (bytes unavailable, or a transient probe failure), so this is a tail case, but the bound should be the same shared deadline the converter already uses.

The work: hoist the aggregate budget into the shared probe helper so every legacy URL path — conversion, export, and video timeline — abandons the remainder once the budget is spent, and reports the partial result rather than stalling.

Refs #1007, #1101

HFY-aa:
## Summary

- add a shared wall-clock budget for legacy URL probes
- apply the budget to document conversion, classroom ZIP export, and video timeline export
- cap each fetch by the time remaining and skip queued requests after the deadline
- cover remaining-time and deadline-expiry behavior

## Why

Each legacy URL path previously applied only a per-request timeout. A document with many unreachable URLs could therefore stall conversion or export for multiple timeout windows.

## Validation

- Added focused unit coverage for the shared budget.
- Local test execution was not possible because this machine's Git checkout connection resets before dependency installation. Please rely on CI for the full suite.

Closes #1129 (item 3 only).

### DIFF
--- a/lib/export/classroom-zip-utils.ts
+++ b/lib/export/classroom-zip-utils.ts
@@ -7,6 +7,7 @@ import { isConcreteMediaAddress } from '@/lib/media/resolve-media-ref';
 import { resolveAudioBlob } from '@/lib/media/resolve-audio-bytes';
 import { fetchMediaUrl } from '@/lib/media/fetch-media-url';
 import { mapWithConcurrency } from '@/lib/media/convert-legacy-asset-refs';
+import { createLegacyUrlProbeBudget } from '@/lib/media/legacy-url-probe-budget';
 import { withAssetUrl } from '@/lib/media/use-asset-url';
 
 // ─── Export: Collect Media ─────────────────────────────────────
@@ -141,9 +142,12 @@ export async function collectLegacyAudioForExport(
   }
   const blobs: LegacyAudioBlob[] = [];
   const audioUrlToPath = new Map<string, string>();
+  const probeBudget = createLegacyUrlProbeBudget();
   const fetched = await mapWithConcurrency([...uniqueLegacyUrls], 4, async (url) => {
+    const timeoutMs = probeBudget.nextTimeoutMs();
+    if (timeoutMs === null) return { url, blob: null };
     try {
-      const response = await fetchMediaUrl(url, 15_000);
+      const response = await fetchMediaUrl(url, timeoutMs);
       if (!response.ok) return { url, blob: null };
       const blob = await response.blob();
       // Zero-byte responses are not narration: skip the entry (the same
--- a/lib/media/convert-legacy-asset-refs.ts
+++ b/lib/media/convert-legacy-asset-refs.ts
@@ -42,22 +42,22 @@
  * this converter never reached.
  */
 
-import type { AssetMeta, Slide } from '@openmaic/dsl';
-import { createLogger } from '@/lib/logger';
-import type { AppDocument } from '@/lib/document-store/persistence-types';
-import type { Action, SpeechAction } from '@/lib/types/action';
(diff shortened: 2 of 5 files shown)


CXuPercy (changes requested):
Cross-reviewed by two independent reviewers; the budget module itself is approve-quality — the shared wall-clock design is clean (per-fetch timeout capped by remaining time, queued probes skipped after the deadline, injectable clock), and both reviewers found no correctness issues in it.

What blocks merge is hygiene, not content: CI is red on both the lint job and E2E. The lint failure is an editor-driven quote-style rewrite churned through `convert-legacy-asset-refs.ts` (single → double quotes; the repo's prettier enforces `singleQuote: true`), which buries the real ~100-line change under formatting noise. The branch is also six weeks behind main.

Please: run `prettier --write` to revert the quote churn, rebase onto current main, and re-run CI. With that, the remaining review ask is small — coverage for timeout propagation and deadline-skipped probes at the call sites, which the current tests only partially pin.

wyuc:
@HFY-aa — the change-request review from 2026-09-04 has been waiting a week, and the asset lifecycle work now starting under #1007 (newest comment) will touch the same probe paths. Could you say by 2026-09-18 whether you plan to pick this up? If not, no problem at all — I will take the remaining review items over on a branch that credits this PR, and close this one when that lands.

PR #1140, opened 2026-08-18
fix(tts): stop TTS reading LaTeX math delimiters literally (#394)

### LINKED ISSUE #394: [Bug]: 中文语音播报会读出包裹 LaTeX 公式的 $ 符号，且无法正确识别角标与数学符号
### Bug Description

1.  读取`$`包裹的行内LaTeX公式时，会将`$`符号读为“美元”；
2.  无法正确识别并播报公式中的角标（如下标`$x_i$`）、特殊数学符号（如`$\lceil$`、`$\frac$`等），直接按照纯文本的方式播报。

例如：读取 $n^2$ 时，会错误播报为「n 两美元」，而非正确的「n 的平方」。

### Steps to Reproduce

使用中文语音播报至任意公式部分时即可复现。

### Expected Behavior

1.  不播报包裹公式的`$`，仅播报公式内容；
2.  正确识别并播报角标、数学符号的数学含义，而非以纯文本的方式朗读。

### Actual Behavior

1.  播报时将包裹公式的`$`读为“美元”；
2.  无法正确识别角标、特殊数学符号，仅以纯文本方式播报，无法还原数学含义。

### Deployment Method

Other

### Browser

_No response_

### Operating System

_No response_

```shell

```

### Additional Context

_No response_

SomSamantray:
## Summary

Fixes #394. Narration/discussion text that still contains LaTeX math (`$...$`, `$$...$$`, `\frac{a}{b}`, `^`, `_`) was passed verbatim to the TTS providers, so Chinese (and other) TTS engines read out the literal `$` delimiters and math tokens. This PR normalizes that text into a spoken, pronounceable form at the single shared server-side synthesis entry point before any provider reads it.

## Related Issues

Fixes #394

## Changes

- **`lib/audio/tts-text-normalize.ts` (new)** — pure, dependency-free `normalizeSpokenText(text)` that:
  - converts paired `$...$` / `$$...$$` math spans (only when the interior looks like math; a lone/currency `$` stays literal) and unambiguous LaTeX command tokens (`\frac`, `\times`, `\cdot`, `\leq`, ...) into spoken words
  - verbalizes superscripts (`^`), subscripts (`_`), square roots, and math symbols (`= + - × ÷ ≤ ≥ ≠ % < >`) inside math spans
  - leaves ordinary prose (stray backslashes, currency, `3+4`, `C++`) byte-identical — no regression
- **`lib/audio/tts-providers.ts`** — `generateTTS` now calls `normalizeSpokenText(text)` before the provider switch. `generateTTS` is the shared synthesis entry used by both the HTTP route (`/api/generate/tts` -> narration, discussion, agent-bar, preview) and the server-side classroom media path, so every server-side TTS request is normalized from one choke point.
- **`tests/audio/tts-text-normalize.test.ts` (new)** — vitest table tests (11 cases): inline/display math, fractions, exponents/subscripts, sqrt, comparisons, bare commands, lone-`$`/currency, and no-math passthrough.

## Type of Change

### Steps to reproduce / test

1. `pnpm vitest run tests/audio/tts-text-normalize.test.ts` (11 tests pass)
2. `pnpm vitest run tests/audio` (27 files / 170 tests pass)
3. `pnpm lint` + `pnpm check` on the changed files, and `tsc --noEmit` (0 errors)
4. Dev-server smoke: root loads (HTTP 200); `/api/generate/tts` returns the expected `400` for a missing-fields body, confirming the route + `generateTTS` + normalizer import chain is intact

The full audio unit suite passes with no regression in TTS-consuming providers, and the app type-checks clean. Verifying "no literal `$` in the synthesized audio" requires external provider credentials, so it is covered at the unit level by the `normalizeSpokenText` tests rather than live synthesis.

### Evidence

- [x] `pnpm lint` + `pnpm check` + `tsc --noEmit` clean
- [x] Manually tested locally
- [ ] Screenshots / recordings attached (N/A — no UI change)

## Checklist


wyuc (changes requested):
Thanks for working on this. I do not think a provider-agnostic regex rewrite in the shared TTS entry point is the right boundary. It hard-codes Chinese output for every language and provider, does not cover browser speechSynthesis, and already reverses fraction semantics: \frac{a}{b} becomes a 分之 b.
Please generate and cache a separate localized speechText alongside the display text, preferably during content generation, with a deterministic fallback for legacy content. Provider adapters can then use native LaTeX or SSML when supported, and plain speech text otherwise. I suggest redesigning this PR around that model instead of expanding the regex table.

LING-6150:
Thanks for working on this. After maintainer discussion, we’ve decided not to pursue provider-agnostic regex normalization in the shared TTS path. Math speech behavior depends on both locale and provider capabilities, and this approach would introduce Chinese-specific semantics across all providers while leaving the browser speech path inconsistent.
Since addressing this properly requires a different design rather than incremental changes to this implementation, I’m closing this PR. Thanks again for the contribution.

PR #1151, opened 2026-08-20
feat(storage): allow path-style addressing for S3-compatible asset st…

weinaike:
## What

  `loadS3AssetByteStore` constructed its S3 client with empty options, so a deployment pointing the AWS SDK env chain (`AWS_ENDPOINT_URL_S3`) at an
  S3-compatible server was stuck with virtual-hosted addressing. MinIO, Ceph, and other self-hosted gateways generally only answer path-style — and the AWS
  SDK exposes no environment variable for the flag.

  - `loadS3AssetByteStore(bucket, { forcePathStyle })` passes the flag through. The package still reads no environment itself: region / endpoint / credentials
  stay owned by the SDK's own chain, matching the seam's existing rule.
  - The app's byte-store selection reads `AWS_S3_FORCE_PATH_STYLE=1` and forwards it — the same division as today, where the app owns `ASSET_S3_BUCKET` while
  the SDK owns `AWS_*`.
  - `.env.example` documents the variable next to `ASSET_S3_BUCKET`.

  ## Verification

  - `@openmaic/storage` unit suite: 856 passed (4 skipped files unchanged).
  - S3 contract suite against a real MinIO instance (`S3_CONTRACT_ENDPOINT` + `S3_CONTRACT_BUCKET` + `STORAGE_S3_CONTRACT_REQUIRED=1`): 3 passed.
  - `prettier --check` / `eslint` / `tsc --noEmit` clean on the touched files.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -402,6 +402,12 @@ DEFAULT_MODEL=
 # are resolved through the standard AWS SDK environment / credential chain.
 # ASSET_S3_BUCKET=
 
+# Use path-style bucket addressing (endpoint/bucket/key). S3-compatible servers
+# reached through AWS_ENDPOINT_URL(_S3) — MinIO, Ceph, other self-hosted
+# gateways — generally require it; the AWS SDK offers no variable of its own
+# for this flag. Leave unset for Amazon S3.
+# AWS_S3_FORCE_PATH_STYLE=1
+
 # Opt into indirect asset byte egress: answer asset byte GETs with a short-lived
 # signed S3 URL (a 302, or a JSON descriptor for the packaged client) instead of
 # the bytes. Unset or "direct" keeps direct egress (the safe default). Requires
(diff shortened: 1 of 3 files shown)


CXuPercy (changes requested):
Cross-reviewed by two independent reviewers. The fix itself is correct and the passthrough design is honest (typed client-config slice, documented env exception, no SDK-owned settings duplicated). Three mechanical items before merge:

1. **Version gate**: the PR modifies `packages/@openmaic/storage` sources without a version bump — CI's package-version check fails ("Every @openmaic publishable package change must ship with a new package version"). Please bump `@openmaic/storage`.
2. **Two exact-args assertions break with the new second argument** (latent — the version gate fails before tests run): `tests/persistence/asset-collector-schedule.test.ts:251` and `tests/persistence/route.test.ts:324` both assert `toHaveBeenCalledExactlyOnceWith('asset-bucket')`. Please update them for the options object.
3. Nice-to-have: a unit test asserting `AWS_S3_FORCE_PATH_STYLE=1` reaches `S3Client({ forcePathStyle: true })` — the MinIO contract test constructs the client directly and doesn't exercise this path.

wyuc:
Thanks @weinaike for this contribution. I'm closing it to keep the queue tidy, since there hasn't been an update since the changes requested on 2026-09-04. It isn't a rejection of the idea: if you'd like to pick it back up, you're welcome to reopen it or open a fresh PR against the current `main` that addresses the review, and we'll take another look.

PR #1152, opened 2026-08-20
feat(pi): add Native Child whiteboard tools

LING-6150:
## Related issue

- Related to #1049

## Summary

Implements the minimal RuntimeStore-backed Native Child whiteboard slice defined in #1049:

- `wb_read` reads the authoritative learner whiteboard state and current best-effort Browser visibility.
- `wb_open` / `wb_close` send UI-only, best-effort visibility effects without mutating durable state.
- `wb_draw_text` commits through the existing RuntimeStore CAS and replay contract.
- Committed state is projected to the Browser, which performs an authoritative refetch.
- Tool results return to the same Native Child for continuation.

`wb_read` also returns a derived, non-persistent `nextMutation.expectedLastSeq` hint. Mutation tools must copy that value exactly, including `0`; closed Browser visibility does not block durable drawing.

## Architecture boundaries

- RuntimeStore remains the only durable whiteboard authority, partitioned by `(stageId, learnerKey)`.
- Reuses the existing learner/session plumbing and PostgreSQL-backed RuntimeStore provider.
- Preserves mandatory `expectedLastSeq` CAS behavior; the host does not override model-supplied sequence values or retry mutations automatically.
- Returns `STALE_STATE` when the observed sequence is outdated.
- Projection failure never rolls back a committed RuntimeStore mutation.
- Does not write or dual-write Legacy `Stage.whiteboard`.
- Does not route Native drawing through Legacy ActionEngine.
- Does not introduce a whiteboard-specific authorization/context system, RuntimeStore implementation, provider path, feature flag, capability registry, retry manager, runtime budget, or completion exception.
- Native `AgentTool` descriptions and schemas remain the tool-contract authority; the Native prompt only reflects the actual tool inventory.
- Generic Native completion semantics and existing provider/model resolution remain unchanged.

## Production flow

```text
Classroom Browser
→ POST /api/chat/pi
→ Director
→ Native Child
→ wb_read
→ optional wb_open
→ wb_draw_text(expectedLastSeq)
→ RuntimeStore append / CAS / fold
→ committed tool result
→ best-effort Browser projection
→ authoritative refetch
→ same-Child continuation
→ optional wb_close
```

## Verification

- Final PR HEAD: `5105e616`
- GitHub CI: **5/5 passed**
  - Lint, Typecheck & Unit Tests
  - E2E Tests
  - PostgreSQL Storage Contract
  - Render Service
  - CodeQL
- Targeted Native whiteboard, route-wiring, visibility, RuntimeStore, and payload regression: **51/51 passed**
- ESLint, Prettier, and `git diff --check`: passed
- CAS handoff coverage includes:
  - `lastSeq = null`
  - `lastSeq = 0` without falsy coercion
  - closed Browser visibility remaining non-blocking
  - stale `null` against authoritative sequence `0`
  - no additional RuntimeStore record after stale rejection

## Browser E2E

A real Browser E2E was recorded from the PR implementation using a real tool-calling model, PostgreSQL-backed RuntimeStore, and the production `/api/chat/pi` route—not mocks.

pr4-native-whiteboard-e2e

The recording verifies:

```text
real classroom request
→ Director delegation
→ Native wb_read / wb_open / wb_draw_text / wb_close calls
→ RuntimeStore commit
→ same-Child continuation
→ Browser projection and authoritative refetch
→ rendered whiteboard
```

## Cross-review result

Independent Codex and Claude reviews checked the implementation against Epic #1049, the Frozen Implementation Note, and the merged RuntimeStore, Shared Pi transport, Native Child, and Native Web Search foundations.

The reviews found:

- no remaining architecture or correctness blocker;
- no whiteboard-specific Runtime, authorization, provider, configuration, budget, or completion subsystem;
- correct reuse of the existing RuntimeStore service, CAS, replay, and reconciliation behavior;
- correct learner/Stage isolation;
- correct visibility pending cleanup and stale Stage response rejection;
- no Native write to Legacy `Stage.whiteboard`;
- no PR5 drawing tools or unrelated scope included in this PR.

The final CAS handoff amendment was separately reviewed as a narrow tool-contract reliability fix with no architecture change.

### DIFF
--- a/app/api/chat/pi/route.ts
+++ b/app/api/chat/pi/route.ts
@@ -25,6 +25,9 @@ import { apiError } from '@/lib/server/api-response';
 import type { ThinkingConfig } from '@/lib/types/provider';
 import type { StatelessChatRequest } from '@/lib/types/chat';
 import { resolveClassroomWebSearchConfig } from '@/lib/server/web-search-config';
+import { authenticatePersistenceHeaders } from '@/lib/persistence/server-auth';
+import { getServerPersistenceProvider } from '@/lib/persistence/server-provider';
+import { createWhiteboardRuntimeService } from '@/lib/whiteboard/runtime/store';
 
 const log = createLogger('Pi Chat API');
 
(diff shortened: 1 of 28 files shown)


wyuc (changes requested):
There is one blocking projection-recovery issue.

A RuntimeStore mutation may commit successfully while its projection SSE is lost, the request is aborted, or the event-triggered authoritative read fails transiently. In those cases, the Whiteboard component currently refreshes only when stage.id changes, and manually opening the board does not trigger another read. The durable mutation can therefore remain invisible until a Stage switch or page reload.

Please reconcile from RuntimeStore when the whiteboard is opened, with regression coverage for a failed or missed projection followed by a manual open. This can remain a simple authoritative refetch; it does not require Browser ACKs, durable delivery, or a retry subsystem.

LING-6150:
@wyuc   Thanks for your review; I've addressed in `884f1745`.

The Whiteboard now performs a one-shot authoritative RuntimeStore refetch on a `closed → open` transition. This covers both manual opening and Native `wb_open`, since both update the same presentation state.

I also coalesced Stage preload and open-transition reconciliation into a single effect, so a simultaneous Stage change and open triggers only one read.

Regression coverage now verifies:

- a failed/missed projection followed by manual open triggers a fresh authoritative read;
- a transient read failure does not prevent a later successful refetch;
- a Stage change and open transition in the same render issue only one refetch.

No ACK, retry/polling, durable-delivery mechanism, or new Runtime state was added. Projection failure remains independent of the already-committed RuntimeStore mutation.

Local validation:

- focused tests: 32/32 passed;
- full suite: 4,870 passed, 61 skipped;
- TypeScript, ESLint, Prettier, and `git diff --check`: passed.

wyuc (approved):
The projection-recovery finding is resolved in 884f1745. The whiteboard now performs an authoritative RuntimeStore refetch on open without introducing ACKs, polling, or durable UI-delivery machinery. CI is green.

Issue #1153, opened 2026-08-21
[RFC] Uploaded documents as first-class assets: a material library over the asset pool


wyuc:
## Goal

Uploaded source documents are one-shot parse fuel today: the file is read into a
buffer, an extraction artifact comes back, and the original bytes are discarded.
Everything downstream of that moment — the extracted images, the generation flow,
the produced course — works on transient copies, and the heaviest of those copies
(images) end up inlined into the DSL document as base64 data URLs.

This RFC makes uploaded documents and the media derived from them **first-class
assets in the pool #1007 built**: the original file is ingested at upload and gets
an allocated `asset_id`; extraction becomes an id-addressed, cacheable derivation;
extracted images become derived assets with recorded lineage; and generation
references materials by allocated id instead of inlining bytes. One new object ties
it together: a per-principal **material library** — a manifest of the assets a user
has imported, with the metadata an agent needs to select among them.

The pool itself needs no new mechanics. Allocation, content-addressed dedup, the
server backend, principal partitioning and logical-byte quota all shipped under
#1007; this RFC is the pool's second ingest client, plus the one manifest object.

## Motivation — where we are today

**The original file has no durable identity.** `app/api/extract-document` reads
the upload into a buffer, runs the extractor, and returns the artifact; the server
never persists the original bytes. The client does stash each selected source in a
session-scoped IndexedDB blob store (#844) — but as transport between the upload
page and generation preview: a cleaned-up cache with no stable id, no server-side
existence, and no cross-course meaning. A document can therefore never be re-parsed
(extractors improve; the input is effectively gone), never be reused by a second
course without a second upload, and never be consulted again by any later step.

**Extracted images travel and land as base64.** Extraction returns images as data
URLs on `DocumentAsset.data` (`lib/document/pdf-compat.ts:104-121`). On the client
they sit in the session-scoped `imageFiles` Dexie cache — a workaround for the
sessionStorage 5MB limit (`lib/utils/image-storage.ts`), and explicitly a non-goal
of #1007. Generation ships them to the model through `imageMapping` (id → data
URL), and `resolveImageIds`
(`packages/@openmaic/generation/src/scene-generator.ts:327`) then writes the data
URL itself into `PPTImageElement.src`. The persisted document carries the raw
base64: a source image used on five slides is stored five times, and every
load/save of the document moves those bytes through the document store instead of
the byte layer.

**The pool's benefits stop at generated media.** The #1101 converter ingests
[...]

PR #1154, opened 2026-08-21
feat(document): ingest uploaded sources into the asset pool and extract by id (#1153 part 0)

wyuc:
## Summary

Part 0 of RFC #1153: uploaded course-material source files become first-class assets. Each selected file is ingested into the asset pool at upload time and gets an allocated asset id; the extract-document route accepts an asset id as input (the server resolves the bytes from the server asset store) alongside the existing multipart byte form, which is preserved byte-for-byte; every extractor provider now declares a `version`.

This is the step that removes the "one-shot" property the RFC's motivation describes: a source document now has a durable identity that later parses, later courses, and later milestones (lineage, the material library) can refer to.

## Related Issues

Part 0 of #1153. Builds on the extractor boundary from #621 (see the RFC's "Relationship to MAIC ETL" section); the ETL computation stages are unchanged.

## Changes

- **Upload ingest** (`app/page.tsx`, `lib/document/extract-source.ts`): selected files are ingested into the asset pool via `putAsset` at selection time; the generation session carries the allocated `assetId` per source (legacy `storageKey` sessions still load). The material set and provider config are frozen for the duration of generate-prep; pending ingests are awaited with a 15 s bound (timed-out sources proceed byte-path, late-resolving ids are released); removal releases the pool entry, with the pending-ingest map as the durable release handle.
- **Extract by id** (`app/api/extract-document/route.ts`, `lib/persistence/resolve-server-asset.ts`): the route accepts an `application/json` form `{ assetId, fileName, mimeType, ...providerConfig }` when the deployment's asset pool is server-backed; bytes are resolved server-side (identity read first — an over-cap recorded length answers 413 without materializing bytes). The multipart form is untouched and remains the path for browser-backed (self-deploy) pools. JSON-path inputs are type-validated; JSON-path error responses are generic (no caller input echoed, no raw internal error text); the SSRF guard applies to both forms.
- **Client fallback** (`lib/document/extract-source.ts`): if the asset-id form fails before extraction ran (store unavailable, asset missing, auth, unconfigured), the client falls back to the byte form it already holds; `PARSE_FAILED` responses are terminal so a paid extraction is never re-billed by the retry.
- **Provider version** (`lib/document/types.ts`, extractors): `DocumentExtractorProvider` and `MediaExtractorProvider` declare `version: string` (consumed by nothing yet; part 1's extraction-artifact cache keys on it).
- **Tests**: new suites for the route's JSON form (validation, error hygiene, SSRF, 413 ordering, provider auto-selection), `resolve-server-asset` (real authenticator, identify-first cap), the client fallback decision table, ingest await/release helpers, session back-compat, and provider version declarations.

## Type of Change

### Steps to reproduce / test

1. `pnpm check` → clean; `pnpm lint` → 0 errors; `tsc --noEmit` → clean; `pnpm exec vitest run` → 462 files / 4990 tests passed.
2. Self-deploy (browser-backed pool): upload materials, generate — multipart path, behavior unchanged.
3. Server-backed pool (`NEXT_PUBLIC_PERSISTENCE=1` + `DATABASE_URL`): upload materials, generate — extraction runs by asset id (no byte re-upload); stop the DB mid-flow and generation still completes via the byte fallback.

- The server asset store's shared-principal development authenticator is unchanged; real per-user principals are RFC #1153 part 3 (the #1129 ownership item). An in-line comment marks the decision at the resolve call.
- A tab closed while an ingest is in flight can leak one pool entry (no reliable browser teardown hook); commented at the ingest site. Management/visibility of pool entries arrives with the RFC's material-library milestone.
- Browser-backed pool entries persist after a successful run — durability is the point of the RFC; the material library provides deletion.


app/api/extract-document/route.ts
```diff
@@ -104,240 +198,423 @@ function formatTimestamp(ms: number): string {
...
+      provider ||
+      selectDocumentExtractorProvider({
+        mimeType,
+        requiredCapabilities: { text: true },
+      });
+  } catch (error) {
+    return apiError(
+      'INVALID_REQUEST',
+      400,
+      error instanceof Error ? error.message : `Unsupported course material type "${mimeType}"`,
```
CXuPercy: **The JSON path can still echo caller-controlled input here**, contrary to the PR description ("no caller input echoed") and the `validateJsonPathProvider` doc comment ("that path does not echo caller input").

Reachability: `normalizeDocumentMimeType` rule 3 passes unrecognized MIMEs through verbatim (correctly, to avoid extension spoofing), and with no `providerId` there is no MIME pre-validation at all — `validateJsonPathProvider` returns `null` immediately. `selectDocumentExtractorProvider` then throws ``No document extractor supports MIME type "${mimeType}" …`` (registry.ts:52), and this catch returns `error.message` as-is.

Verified empirically against this branch — POSTing `{"assetId":"ast_probe","fileName":"probe.bin","mimeType":"application/x-echo-probe"}` (asset resolved) answers:

```json
{"success":false,"errorCode":"INVALID_REQUEST","error":"No document extractor supports MIME type \"application/x-echo-probe\" with the requested capabilities"}
```

Severity is low — JSON content type (no XSS), reflected only to the caller who supplied it, and `normalizeDocumentMimeType` lowercases and strips anything after `;`. But it falsifies the form's headline invariant, and no test covers this branch (the closest test uses `text/markdown`, which auto-selects successfully rather than throwing).

Suggested fix: mirror the catch-all 500 handler — when `isAssetIdForm`, answer a generic static message here (or pre-check the MIME against `SUPPORTED_COURSE_MATERIAL_MIME_TYPES` on the JSON path before `runExtraction`).

app/page.tsx
```diff
@@ -463,27 +479,108 @@ function HomePage() {
...
+
   const addCourseMaterials = (files: File[]) => {
+    // The set is frozen for the duration of generate-prep: adding is inert
+    // while `preparingGenerate` is set (the toolbar affordance is disabled
+    // via the same state), so nothing can slip into the set mid-prep.
+    if (preparingGenerate) return;
+    // Compute the additions before touching state: the ingest loop must not
+    // run inside the setForm updater, which React may invoke more than once —
+    // each replay would putAsset again and orphan an allocated id.
+    const dedupedFiles = dedupeCourseMaterialFiles(form.courseMaterials, files);
```
CXuPercy, line 532: Nit — the dedupe moved from inside the updater (the old code read `prev.courseMaterials`) to this closure snapshot. Two `addCourseMaterials` calls landing in the same render batch now dedupe the second call against a stale list: the same file can enter twice under two nanoid ids (the updater's filter is id-based only), costing two pool ingests and a duplicate extraction in the bundle. The old code was race-free by construction.

The move itself is right — StrictMode replays updaters, and an ingest inside one would double-`putAsset`. If you want to close the window, an extra fingerprint filter inside the updater composes with the existing id filter:

```ts
const missing = additions.filter(
  (a) =>
    !prev.courseMaterials.some(
      (item) =>
        item.id === a.id ||
        courseMaterialFingerprint(item) === courseMaterialFingerprint(a),
    ),
);
```

Rare in practice (needs two adds in one batch) — fine to defer.

lib/document/extract-source.ts
```diff
@@ -0,0 +1,213 @@
...
+        return options.fetchers.submitByteForm();
+      }
+      // PARSE_FAILED: the extractor already ran and will fail identically on a
+      // byte retry, re-billing the provider. Surface the response unchanged.
+      options.logWarning(
+        `Asset-id extraction returned ${response.status} PARSE_FAILED; surfacing the error without a byte retry.`,
+      );
+      return response;
+    } catch (error) {
+      options.logWarning('Asset-id extraction failed; falling back to byte upload:', error);
```
CXuPercy, line 114: Design note — this catch retries with the byte form on *any* throw, including a network failure that arrived while the server-side extractor was already running; in that window the byte retry re-runs a paid extraction. `shouldRetryWithByteUpload`'s docstring states the principle ("free to retry only when the failure provably happened BEFORE extraction ran"), and this path is the one exception to it (the `fetchExtractionResponse` docstring does document the network-error retry, so this is a tension between the two docstrings rather than an oversight).

Either is fine IMO: (a) a sentence in `shouldRetryWithByteUpload`'s docstring acknowledging the network-error exception and its billing cost, or (b) not retrying non-abort network errors. Favoring user success over a rare double bill is defensible — it just deserves to be stated where the principle is claimed.

tests/document/extract-document-route.test.ts
```diff
@@ -215,3 +221,505 @@ describe('POST /api/extract-document', () => {
...
+      success: false,
+      errorCode: 'ASSET_NOT_FOUND',
+    });
+    // The response must stay generic — the caller-controlled asset id is not
+    // echoed back into the body.
+    expect(json.error).toContain('No course material asset');
+    expect(json.error).not.toContain('ast_missing');
+  });
+
+  it('returns 400 when the server asset store fails (generic 500 message, real error logged only)', async () => {
```
CXuPercy: Nit: the test name says "returns 400" but the assertions check `expect(res.status).toBe(500)` — the assertions are right, just the name.

CXuPercy (dismissed):
Cross-review from my side: I verified this PR against the modules it builds on — the storage package's asset contract, the dev authenticator (`server-auth.ts`), `getServerPersistenceProvider`, the MIME normalization table, and the extractor registry — and re-ran the route under test to check the JSON path's error-hygiene invariant empirically.

**Claims that check out:**

- 413-before-materialization is contract-backed: `identify` returns the recorded `byteLength` without reading bytes, and the test asserts `resolve` was never called. The post-resolve length backstop is also in place.
- Double-release is moot by contract: `remove` is an indistinguishable no-op for unknown/already-removed ids, so the timeout-release + later user-removal overlap is harmless.
- No existence oracle: unknown vs. foreign ids are reported identically, and ids carry 128 bits of randomness.
- The shared-principal dev authenticator is the existing platform pattern (the pi routes already use it); reusing it and marking the decision in-line at the resolve call is right for part 0.
- The multipart path's observable behavior is faithfully frozen through the `runExtraction` refactor (messages preserved, legacy tests untouched and passing).
- The fallback decision table is fully enumerated in tests; `PARSE_FAILED` terminality prevents re-billing on the response path; the SSRF test exercises the metadata IP on the JSON path.
- Moving the ingest out of the `setForm` updater is correct for StrictMode purity (a replayed updater would double-`putAsset`).

**One invariant gap** (low severity — inline below with a repro) and a few nits inline. Two more minor notes, no action needed:

- A timed-out ingest that resolves late patches its (already released) `assetId` into form state, so the *next* generate pays a 404→byte-fallback roundtrip for that source. Self-healing; skipping the patch once the id was released would avoid the wasted trip.
- Self-deploy: each selected material now lands in two IndexedDB stores (`imageFiles` stash + `maic-asset-pool`), growing unboundedly until the material library lands. Acknowledged in the PR description — just confirming it's intended for part 0.

Note the branch is behind `main` and will need an update before merge.

app/api/extract-document/route.ts
wyuc: Fixed in c43a9892 — the asset-id form now answers this catch with a generic static message (multipart keeps the registry's interpolated message byte-for-byte). Your exact probe is pinned on both sides: JSON form asserts the MIME string is absent (and no extractor is consulted), multipart asserts it is still present. Thanks for the empirical repro — this was exactly the branch the earlier tests missed.

app/page.tsx
wyuc: Applied your composition in c43a9892: the pure updater now filters by id AND content fingerprint (reusing the courseMaterialFingerprint helper that landed on main), so two adds in one render batch can no longer double-enter the same file. Unit tests added for the fingerprint/dedupe pair.

lib/document/extract-source.ts
wyuc: Took option (a) in c43a9892: shouldRetryWithByteUpload's docstring now names the network-error exception one level up, its double-bill cost, and why user success wins. Behavior unchanged.

tests/document/extract-document-route.test.ts
wyuc: Renamed in c43a9892.

wyuc:
@CXuPercy all four of your inline findings are addressed in c43a9892 (replies inline on each thread), and the branch is updated with main per your note — CI is green on the result. The push auto-dismissed your approval, so this needs one more click from you when you have a moment. The diff since your review is small: the generic-message catch + its two pinned tests, the fingerprint filter in the updater, one docstring paragraph, and a test rename.

PR #1149, opened 2026-08-19
feat(pi): expose benchmark LLM usage

LING-6150:
## Summary

Adds request-scoped LLM usage observability to Pi Chat for benchmark consumers.

- emits additive `llm_call_start`, optional `llm_usage`, and `llm_call_end` SSE events for OpenMAIC-visible `streamLLM` invocations;
- includes stable call identity, phase, provider/resolved model, optional actual model, status, and normalized/raw usage;
- aggregates request usage into the existing terminal `done` / best-effort `error` payload;
- covers Director, Native Child, the current single-transport Legacy Child, same-Child continuations, and Director compaction;
- preserves completed call usage across later cancellation or timeout;
- ignores the additive lifecycle events in the classroom client before buffer allocation.

## Contract boundaries

- OpenMAIC remains the only provider/model-resolution authority.
- Retry visibility is explicitly `openmaic_only`; hidden SDK/provider HTTP retries are not claimed as measured.
- This PR does not calculate cost, add provider/fetch instrumentation, collect independent Web Search provider usage, or add durable telemetry.
- Existing text, action, cue-user, replay, and done behavior remains unchanged.

## Companion consumer

Companion MAIC-Mono PR #1868 consumes these events, persists execution usage, reconstructs incomplete lifecycle prefixes, and performs benchmark-local versioned pricing.  https://github.com/CogEvol/MAIC-Mono/pull/1868

## Validation

- Expanded Pi/runtime regression gate: **29 files / 275 tests passed**
- Targeted ESLint: passed
- Prettier: passed
- `git diff --check`: passed
- Independent final code reviews: **PASS**

Full workspace `tsc --noEmit` remains affected by existing current-main workspace dependency/build resolution outside the changed files; no TypeScript diagnostic was attributed to this diff. GitHub CI is the required final gate.

PR #1156, opened 2026-08-21
feat(pi): add Native Child additive whiteboard tools

LING-6150:
## Summary

Implements the additive-drawing half of PR5 from #1049 by extending the merged PR4 Native whiteboard adapter with:

- `wb_draw_shape`
- `wb_draw_chart`
- `wb_draw_latex`
- `wb_draw_table`
- `wb_draw_line`
- `wb_draw_code`

All six tools reuse the existing PR1 RuntimeStore `element_added` operation and the PR4 append/CAS/reconciliation/projection path. Native tool schemas remain the parameter authority, and tool results return to the same Native Child for continuation.

The Browser renderer changes are limited to two issues found during real E2E validation:

- keep category-axis labels visible for small Runtime-projected charts while retaining automatic thinning for dense charts;
- recover delayed code lines when an authoritative refetch replaces the animated projection.

The Native Child guidance also makes the existing visibility contract explicit: for a user-visible drawing request, call `wb_open` before the first mutation without requiring the learner to name whiteboard tools.

## Boundaries

- RuntimeStore remains the only durable learner-whiteboard authority, partitioned by `(stageId, learnerKey)`.
- Every mutation carries mandatory `expectedLastSeq`; operation, element, table-cell, and code-line IDs remain host-owned and stable for exact replay.
- Confirmed commits project through the existing best-effort Browser refetch path; projection or rendering failure does not roll back durable state.
- `wb_open` remains a separate UI-only best-effort effect. Draw tools do not mutate visibility themselves.
- Does not write or dual-write Legacy `Stage.whiteboard` and does not route Native drawing through Legacy ActionEngine.
- Does not change generic Native completion, transport deadlines, cancellation, or tool budgets.
- Does not add a WB-specific runtime, provider, authorization/context system, configuration path, manager, registry, ACK, receipt, or recovery framework.
- `wb_clear`, `wb_delete`, `wb_edit_code`, and Legacy cleanup are intentionally deferred to the next independently reviewable PR5 slice.
- #1011 remains a historical reference only; #1149 is not an authority or dependency.

## Next PR5 slice

The remaining mutation tools are intentionally excluded from this Draft PR:

- `wb_delete` — removes a targeted element;
- `wb_clear` — removes all learner whiteboard elements;
- `wb_edit_code` — modifies an existing code element in place.

Unlike the additive tools in this PR, these operations have destructive or state-replacing semantics. The next PR will review their target validation, `expectedLastSeq`/replay behavior, projection results, and Legacy isolation independently before performing the final parity cleanup.

## Verification

- Focused Native WB, route, prompt, consumer, and renderer tests: **102/102 passed**.
- TypeScript: passed.
- ESLint: passed.
- Prettier: passed.
- Package-version validation: passed.
- `git diff --check`: passed.
- Real Browser E2E used `deepseek-v4-flash` with ordinary learner language, not mocks or tool-name instructions.

## Cross-review

Independent Codex and Claude reviews checked the exact PR diff against Epic #1049, the PR1 RuntimeStore/CAS contract, and the merged PR4 Native whiteboard adapter and visibility/projection boundaries.

Both reviews found no architecture drift: RuntimeStore remains the sole durable authority, settlement stays centralized in the shared mutation helper, same-Child continuation and generic completion are unchanged, and no parallel runtime/provider/configuration or business-level tool budget was introduced.

The reviews identified local input/rendering correctness gaps only. The resulting narrow fixes now:

- reject blank `wb_draw_code` payloads before append;
- reject zero-length `wb_draw_line` payloads before append;
- preserve all labels for small charts while allowing ECharts to thin dense category axes;
- retain the previously added chart/radar validation, Native table escaping, and deferred-tool prompt guards.

Negative tests confirm rejected mutations neither append RuntimeStore records nor emit Browser projection events.

### Browser E2E

The recording verifies independent Q&A turns for Table, vertical Bar Chart, LaTeX, a red arrow Line, and a two-line Code block. The whiteboard was opened through the Native tool flow and each committed element was rendered from the authoritative RuntimeStore projection.

pr5-native-whiteboard-additive-tools-e2e

Screenshot 2026-08-21 at 4 28 47 PM

Screenshot 2026-08-21 at 4 28 56 PM

## Review focus

- Canonical element construction for all six additive tools.
- CAS, exact replay, uncertain-settlement, and stable nested-ID behavior.
- Chart schema/renderer compatibility and code-line projection recovery.
- Preservation of PR4 visibility, Legacy-isolation, same-Child continuation, and generic completion boundaries.

## Related issue

- #1049


lib/chat/pi/prompts.ts
```diff
@@ -223,6 +223,17 @@ export function buildNativeChildPrompt(
     availableTools.length === 0
       ? 'No Native tools are available. Respond with speech only.'
       : availableTools.map((tool) => `- ${tool}`).join('\n');
+  const nativeWhiteboardGuidance = availableTools.some((tool) => tool.startsWith('wb_'))
+    ? [
+        '',
+        '# Native whiteboard behavior',
+        '- A `wb_read` visibility result of `closed` means the Browser whiteboard is currently hidden; it does not mean whiteboard tools are unavailable or durable drawing is blocked.',
+        '- For every mutation, copy `nextMutation.expectedLastSeq` from the latest `wb_read` result exactly into `expectedLastSeq`; use `null` only when that value itself is `null`. After `STALE_STATE`, read again and use the new value.',
+        '- If the user explicitly requests a visible whiteboard drawing, call `wb_open` before the first mutation even when you have not observed the current visibility; then call `wb_read` and the required `wb_draw_*` tools. Do not wait for the user to ask you to open the whiteboard.',
```
wyuc, line 232: [P2] Keep the Native whiteboard inventory coherent

Per-agent whiteboard authorization is useful, but once an agent is allowed to perform any whiteboard mutation, the base control plane should be complete. Today such an agent can receive `wb_read` plus `wb_draw_*` while `wb_open` and `wb_close` are omitted because they are independently gated by `allowedActions`. This instruction then requires a tool that is not callable. Please automatically inject `wb_read`, `wb_open`, and `wb_close` whenever any whiteboard mutation is enabled, while continuing to gate mutation tools per agent if needed. This preserves the ability to give some agents no whiteboard capability without creating partial inventories.

components/slide-renderer/components/element/ChartElement/chartOption.ts
```diff
@@ -45,11 +45,14 @@ export const getChartOption = ({
...
-        color: textColor,
-      }
-    : undefined;
+  const axisLabel = {
+    show: true,
+    color: textColor ?? '#333333',
+  };
+  const categoryAxisLabel = {
+    ...axisLabel,
+    interval: data.labels.length <= 8 ? 0 : ('auto' as const),
```
wyuc: [P2] Validate chart data before reading labels

`categoryAxisLabel` now reads `data.labels.length` before the existing `data?.series` guard. For malformed or legacy chart payloads with missing `data` or `labels`, this throws instead of returning `null` as the previous defensive path did. Please move category-label construction after validation or access labels defensively. The mirrored package renderer has the same ordering and needs the same fix.

wyuc (changes requested):
Two blocking issues remain. The Native whiteboard inventory can be internally incomplete for a valid per-agent configuration, and the chart renderer now throws on malformed or legacy payloads that were previously handled defensively. The targeted changed-path suite passes (69 tests), but neither regression is covered by those tests.

LING-6150:
@wyuc 
Thanks for finding two blockings; I've addressed both blocking issues in `d00f615c`:

1. Any agent authorized for a Native WB mutation now automatically receives the complete `wb_read` / `wb_open` / `wb_close` control plane, while individual mutation tools remain gated by `allowedActions`.
2.  Both the app and package chart renderers now validate missing or malformed `data`, `series`, and `labels` before reading category labels, preserving the previous defensive behavior.

Added focused inventory, route-wiring, and mirrored-renderer regression tests. All 5 CI checks now pass. Ready for re-review , thanks in advance.

wyuc (approved):
The two requested blockers are resolved: Native whiteboard mutation capability now carries a coherent read/open/close control plane, and both chart renderers restore defensive handling for malformed legacy data. Focused regression coverage was added, the branch is current with main, and all required checks pass.

PR #1158, opened 2026-08-23
feat(settings): warn users before submitting API keys 

### LINKED ISSUE #441: [Security/UX] Add warning when users enter API keys on untrusted instances
## Problem

When users configure their AI provider API keys through the settings UI, there is no indication that the key will be transmitted to and processed by the server. This is the expected architecture — the server needs the key to call AI APIs on behalf of the user.

However, OpenMAIC instances can be deployed as public services by anyone. Users who discover these instances (e.g. via search engines) may not understand that entering their API key grants the instance operator full access to that key. This differs from platforms like ChatGPT where users authenticate against the official provider directly.

## Proposal

Add a brief security notice near the API key input field in provider settings. Something like:

> ⚠️ Your API key will be sent to the server hosting this instance. Only enter your key if you trust the operator of this deployment.
This should be:
- Visible but not intrusive (e.g. a small info tooltip or muted text below the input)
- Localized (zh-CN / en at minimum)
- Shown for all provider types that accept user-supplied keys

## Context

This is an inherent limitation of the proxy architecture, shared by all similar projects (ChatGPT-Next-Web, LobeChat, etc.). It cannot be "fixed" technically — but user awareness can be improved significantly with a simple UI change.

No API or behavior changes needed. This is purely a UX/security-awareness improvement.

*That said, attackers can easily revert it*

yipwingtim:
Closes #441.

## Problem

OpenMAIC sends user-supplied provider credentials to the server so the server can call external AI services on the
user's behalf.

On publicly deployed instances, users may not realize that entering an API key gives the deployment operator access to
that credential.

## Changes

- add a small security notice near user-editable provider credential fields
- explain that credentials are sent to the server hosting the current instance
- remind users to enter credentials only when they trust the deployment operator
- show the notice for:
  - LLM providers
  - TTS and ASR providers
  - web search providers
  - image and video providers
  - PDF providers
- keep server-configured provider flows unchanged
- add localized copy for English and Simplified Chinese
- use the English notice as the fallback for other supported locales

## Scope

This is a UX/security-awareness change only.

It does not change:

- credential storage or transmission behavior
- provider APIs
- server-side configuration
- authentication or authorization

Configurable notice content and client-side direct mode remain out of scope.

## Validation

- ESLint passed
- TypeScript check passed
- i18n key alignment check passed
- `git diff --check` passed


CXuPercy (changes requested):
Cross-reviewed by two independent reviewers. Structure and coverage are good — the reusable notice component, consistent placement, and all 12 locales present, with server-managed flows correctly excluded. One verified accuracy problem needs fixing before merge: the notice currently renders on panels that never accept an API key, so it warns about a key being sent where none can be:

- `components/settings/pdf-settings.tsx:117` — gated on `needsRemoteConfig`, which includes the self-hosted MinerU flow (base URL only, no key; the file's own comment says "for self-hosted: requires base URL").
- `components/settings/tts-settings.tsx:402` — rendered for all non-server-configured providers, including `voxcpm-tts` (`requiresApiKey: false`, base-URL-only local backend).
- `components/settings/web-search-settings.tsx:80` — `showCredentialFields` also covers `requiresApiKey === false` providers.

A security notice that over-warns on keyless flows erodes its credibility exactly where it matters. Please scope the notice to panels/conditions that actually render a credential input (or make the copy conditional on one being present). If practical, a small test over the provider-condition matrix (server-managed / key-required / optional-key / keyless) would pin it.

yipwingtim:
Addressed the review feedback.

The API key security notice is now limited to credential-bearing flows:

- PDF: shown only for MinerU Cloud and AliDocMind; hidden for self-hosted MinerU.
- TTS: hidden for VoxCPM and other keyless providers.
- Web Search: shown only when the provider requires an API key.

Validation:

- Prettier passed for all three changed files.
- ESLint passed for all three changed files.
- `npx tsc --noEmit` passed with Node.js 24.14.0.
- The broader audio test suite still has two unrelated existing failures in `tests/audio/narrator-pin-
fallback.test.ts`.
- No production API behavior was changed.

yipwingtim:
> Addressed the review feedback.
> 
> The API key security notice is now limited to credential-bearing flows:
> 
> * PDF: shown only for MinerU Cloud and AliDocMind; hidden for self-hosted MinerU.
> * TTS: hidden for VoxCPM and other keyless providers.
> * Web Search: shown only when the provider requires an API key.
> 
> Validation:
> 
> * Prettier passed for all three changed files.
> * ESLint passed for all three changed files.
> * `npx tsc --noEmit` passed with Node.js 24.14.0.
> * The broader audio test suite still has two unrelated existing failures in `tests/audio/narrator-pin- fallback.test.ts`.
> * No production API behavior was changed.



> Cross-reviewed by two independent reviewers. Structure and coverage are good — the reusable notice component, consistent placement, and all 12 locales present, with server-managed flows correctly excluded. One verified accuracy problem needs fixing before merge: the notice currently renders on panels that never accept an API key, so it warns about a key being sent where none can be:
> 
> * `components/settings/pdf-settings.tsx:117` — gated on `needsRemoteConfig`, which includes the self-hosted MinerU flow (base URL only, no key; the file's own comment says "for self-hosted: requires base URL").
> * `components/settings/tts-settings.tsx:402` — rendered for all non-server-configured providers, including `voxcpm-tts` (`requiresApiKey: false`, base-URL-only local backend).
> * `components/settings/web-search-settings.tsx:80` — `showCredentialFields` also covers `requiresApiKey === false` providers.
> 
> A security notice that over-warns on keyless flows erodes its credibility exactly where it matters. Please scope the notice to panels/conditions that actually render a credential input (or make the copy conditional on one being present). If practical, a small test over the provider-condition matrix (server-managed / key-required / optional-key / keyless) would pin it.

Addressed the review feedback.

The API key security notice is now limited to credential-bearing flows:

PDF: shown only for MinerU Cloud and AliDocMind; hidden for self-hosted MinerU.
TTS: hidden for VoxCPM and other keyless providers.
Web Search: shown only when the provider requires an API key.
Validation:

Prettier passed for all three changed files.
ESLint passed for all three changed files.
npx tsc --noEmit passed with Node.js 24.14.0.
The broader audio test suite still has two unrelated existing failures in tests/audio/narrator-pin- fallback.test.ts.
No production API behavior was changed.

wyuc (changes requested):
Thanks @yipwingtim, and sorry for the slow turnaround after your update. This is still wanted: #441 is open, and nothing on `main` tells end users that a key they enter reaches whoever operates the instance (`SECURITY.md` covers operator-side assumptions only). The change applies cleanly.

The latest fix (no notice for providers that don't require a key) moves in the right direction, but gating on `requiresApiKey` gets it wrong both ways:

1. **Under-warning where an optional key is still sent (P2).**
   - **Web search, Brave:** `components/settings/web-search-settings.tsx:83` gates on `provider.requiresApiKey`, which is `false` for `brave`, yet the key input stays enabled and a typed key is sent and used (`app/api/web-search/route.ts` → `lib/web-search/brave.ts`).
   - **TTS, Lemonade:** `components/settings/tts-settings.tsx:405` (`requiresApiKey || isCustom`) hides the notice for `lemonade-tts`, but the rendered key input is sent via `getBackendAuthHeaders(config.apiKey)` (`lib/audio/tts-providers.ts:331`).
2. **Inconsistent across panels (P3).** `image-settings.tsx:217` and `asr-settings.tsx:230` still show it unconditionally for keyless providers (`comfyui-image`, `lemonade`, `funasr-asr`, `lemonade-asr`), so the rule holds on only some panels.

A single rule fixes both: **show the notice whenever a key/credential input is actually rendered**, and ideally phrase it for the moment a value is present (e.g. only emphasise it once the field is non-empty). Please apply it consistently across the settings panels.

Also:

3. **A small regression test.** The earlier review asked for a matrix test of the gating, and nothing in `tests/` pins it yet. A table of `(panel, provider, key-field rendered?) → notice shown?`, including Brave and Lemonade TTS, would stop this from drifting again.
4. **Copy (minor).** The notice says "Your API key…", but it also covers AliDocMind's AccessKey ID/Secret (`pdf-settings.tsx:120`) and Doubao's App ID/Access Key. Something like "Credentials you enter here are sent to the server hosting this instance" would fit all of them.

Happy to re-review once these are in.

PR #1160, opened 2026-08-24
feat(tts): add Qwen TTS voice cloning

wyuc:
## Summary

Adds voice cloning to the existing `qwen-tts` provider through the provider-neutral `VoiceRegistrationAdapter` seam (DashScope `qwen-voice-enrollment`).

- Users create a cloned voice in TTS settings from an uploaded or recorded reference clip (24 kHz mono WAV, 1-60 s) plus its verbatim transcript; cloned voices are then selectable for narration and discussion agents.
- Cloned synthesis uses a dedicated voice-clone model (default `qwen3-tts-vc-2026-01-22`, overridable via `TTS_QWEN_VOICE_CLONE_MODEL`) and reuses `TTS_QWEN_API_KEY` — no new credentials.
- Course generation now pins the narrator to the user's currently selected voice (the LLM assigns voices only to the other agents), with graceful fallback to an enabled voice when a bound voice becomes unusable.

## Design notes

- **The model follows the voice.** Catalog voices always synthesize on their own model group; cloned voices always on the operator-resolved voice-clone model, enforced at every synthesis entry point (server-side resolution plus dispatch defense-in-depth). The voice-clone model is hidden from manual model selection — it is only ever reached through a cloned voice.
- **Registration-based cloning only.** Enrollment returns a vendor-authoritative voice id; re-registration is idempotent (content-addressed, account-scoped memo with LRU/TTL eviction).
- **Security posture.** Vendor-side voice deletion only runs with a caller-supplied API key (server-managed keys perform local-only removal, so users on a shared deployment cannot delete each other's enrolled voices). Operator model pins cannot be bypassed by client-sent model ids. Result-audio download is hardened — host allowlist verified against live vendor responses, redirect refusal, 50 MB streaming cap — and now also covers the previously-raw catalog download path.
- Fixes a resolver gap where registered voices outside the static catalog were silently dropped to the default voice (generalizes the previous VoxCPM-only special case).
- `@openmaic/dsl` gains optional `AgentVoiceConfig.modelId` (version bumped).

## Testing

- Full unit suite passes, with extensive new coverage across audio, resolver, routes, and settings (registration idempotency, pin enforcement, download hardening, narrator pinning and fallback).
- i18n keys aligned across all 12 locales.
- A live-vendor smoke test confirmed the enrollment/synthesis endpoints and the result-URL host shape backing the download allowlist.
- Manually verified end-to-end on a production build: clone creation (upload and record), preview, narrator and discussion playback, deletion semantics.

Squash-merge recommended (the branch carries its review-fix history as separate commits).


lib/hooks/use-scene-generator.ts
```diff
@@ -263,63 +278,169 @@ export async function generateAndStoreTTS(
...
-      return data;
-    },
-    {
-      label: `tts "${requestId}"`,
-      shouldRetryResult: (result) => !result.success || !result.base64 || !result.format,
-      ...retryOptions,
-      signal,
-    },
-  );
+      if (globalDiffers) {
```
CXuPercy: **Blocking: this `globalDiffers` retry branch has no recursion guard** — the sibling `bound == global` branch below guards with `if (!overrideVoice)`, but this one recurses unconditionally whenever the teacher binding differs from the global voice.

When the voice that just failed is the **global** one (the resolver already fell back to it after the bound voice was marked unavailable), the recursive call resolves to the same dead global voice, fails identically, and hits this branch again — forever.

Verified empirically on this branch: teacher bound to `clone-bound-dead`, global selection `clone-global-dead`, every synthesis answering `QWEN_VC_VOICE_NOT_FOUND`:

```
FETCH CALLS: 8
VOICES SENT: ['clone-bound-dead', 'clone-global-dead' x 7]
```

(The probe only stopped because its 8th mocked response was a breaker — in production nothing stops it.) Impact: narration generation hangs indefinitely and hot-loops failing `/api/generate/tts` requests until the user aborts. Healthy behavior is exactly 2 calls (bound fails → global fails → surface the error).

Reachability: user changes the global voice after generation (bound ≠ global), and the global clone is later deleted vendor-side. Narrow, but the consequence is a wedged generation flow.

Suggested fix — only recurse when the failed attempt actually used the bound binding:

```ts
const failedBindingKey = voiceBindingKey({ providerId: ttsProviderId, voiceId: ttsVoice });
if (globalDiffers && failedBindingKey === boundKey) {
  return generateAndStoreTTS(...);
}
throw error;
```

so the second failure (already on the global voice) falls through to `throw` — or alternatively applies the deterministic fallback once, mirroring the bound == global branch.

Test gap: `narrator-pin-fallback.test.ts` covers only bound == global. Minimal repro to pin the fix (reuses that file's harness):

```ts
it('surfaces the error when the global voice is dead and bound differs', async () => {
  const { generateAndStoreTTS } = await import('@/lib/hooks/use-scene-generator');
  mocks.settingsState.mockReturnValue({
    ttsProviderId: 'qwen-tts',
    ttsProvidersConfig: { 'qwen-tts': { apiKey: 'k', modelId: QWEN_TTS_VOICE_CLONE_MODEL } },
    ttsVoice: 'clone-global-dead',
    ttsSpeed: 1,
  });
  mocks.isTTSProviderEnabled.mockReturnValue(true);
  mocks.pickNarratorAgent.mockReturnValue({
    id: 'teacher', role: 'teacher',
    voiceConfig: { providerId: 'qwen-tts', voiceId: 'clone-bound-dead' },
  });
  const notFound = jsonResponse(400, {
    errorCode: 'QWEN_VC_VOICE_NOT_FOUND',
    error: 'The cloned Qwen voice no longer exists.',
  });
  mockFetch.mockResolvedValue(notFound); // every attempt fails

  await expect(generateAndStoreTTS('probe', 'Hello')).rejects.toThrow();
  expect(mockFetch).toHaveBeenCalledTimes(2); // bound once, global once — no loop
});
```

(Today this test cannot pass: the call count grows without bound. I ran a variant of it with a breaker response after 7 calls and watched it retry the same dead voice 7 times.)

CXuPercy (changes requested):
Cross-review from my side: I read the full diff and verified the security claims against the actual implementation, re-running the affected code paths under test on this branch.

**Claims that check out (verified, not just read):**

- **Result-audio download hardening.** Anchored host-regex allowlist (`dashscope-result-*.oss-*.aliyuncs.com`), `redirect: 'error'`, non-default port rejection, http→https upgrade, custom-endpoint origin matching, and a dual 50 MB enforcement (content-length precheck + streaming cap with `reader.cancel()`). The previously-raw catalog download path (plain `fetch` + unbounded `arrayBuffer()`) is now covered too.
- **Delete semantics.** `managed || !callerKey` → localOnly response, vendor deletion unreachable; vendor deletion only ever runs under the caller-supplied key. Cross-user deletion on a shared deployment is not possible.
- **Model-follows-voice, no bypass.** `resolveTTSModel` forces the operator-resolved VC model for clone voices (client model ignored); pinned deployments reject unlisted models with a 400; the provider re-derives clone-ness from the voice itself, so a client-forged `providerOptions.qwenVoiceClone` cannot route a catalog voice onto the clone path; the VC model is excluded from manual pickers; the VC-model+catalog-voice wedge self-heals on both the client and the server.
- **WAV validation.** Strict server-side RIFF parse (PCM/mono/24 kHz/16-bit, chunk walk, 1–60 s) re-validating even though the client normalizes — proper defense in depth.
- **Registration memo.** Content- and account-addressed (sha256 includes the API key), LRU 256 / TTL 1 h, in-flight coalescing, exists precheck avoiding redundant enrollments.
- **Resolver generalization** (non-catalog Qwen voices no longer dropped), plus a tightening: LLM voice tokens now resolve only against the advertised list, where the old code `split('::')`-accepted anything.
- **agent-profiles pinning**: teacher always pinned, the schema example is always an advertised token, and the model segment never comes from the LLM.
- **Discussion fallback**: one-shot notice + requeue with the fallback voice — no retry loop on that path.

**One blocking finding** — inline below, with an empirical repro (unbounded recursion in `generateAndStoreTTS`). Everything else is approve-worthy; I'll flip to approve as soon as it's addressed.

**Non-blocking notes:**

- Clone synthesis silently forces speed 1× (route and provider both). A user-set narration speed is quietly ignored for clone voices — consider surfacing that in the UI.
- The exists precheck only swallows timeouts; a vendor 5xx on the list action fails the whole registration even though a re-register would succeed. Minor UX.
- Rotating `TTS_QWEN_VOICE_CLONE_MODEL` makes `qwenVoiceExists` miss on `target_model` and re-enroll a duplicate vendor voice for the same audio. Bounded, operational.

CI is green (5/5), which matches what I saw locally.

lib/hooks/use-scene-generator.ts
wyuc: Fixed in 890d417c — and took your suggested guard one step further: the whole retry state machine now carries an explicit MAX_NARRATOR_VOICE_FALLBACK_HOPS = 1 bound, so any dead-voice chain (bound → global → deterministic) terminates within 2 total synthesis attempts regardless of which link is dead. Your repro is included as a regression test (asserts rejection + exactly 2 fetch calls). Thanks for the empirical probe — this was the mirror blind spot of the bound==global guard and our own review rounds only covered the equal-branch.

wyuc:
Re the non-blocking notes:

- **Silent 1x speed for clones**: the settings speed slider is already disabled with a hint while a clone voice is active; agreed the config popover path could surface it too — happy to follow up if you feel it matters.
- **Existence precheck vs vendor 5xx**: fixed in a070b6ce — 5xx and network errors on the list action now count as 'unknown' and fall through to the idempotent cached-clip re-register; 401/403 still fail loudly.
- **VC model rotation re-enrolling duplicates**: acknowledged as a bounded operational limitation (the exists check is model-scoped by design so a rotated model re-enrolls once per voice); vendor-side cleanup remains available via BYOK delete.

Branch is updated with main and back to ready — the two new commits are the recursion bound and the 5xx tolerance.

CXuPercy (approved):
Re-verified on the new head — flipping to approve.

**The recursion fix (890d417c) checks out empirically.** I re-ran my original probe scenario (teacher bound to `clone-bound-dead`, global selection `clone-global-dead`, every synthesis answering `QWEN_VC_VOICE_NOT_FOUND`) against this head:

```
FETCH CALLS: 2
VOICES SENT: ['clone-bound-dead', 'clone-global-dead']
OUTCOME: rejected: The cloned Qwen voice no longer exists.
```

Exactly the healthy behavior — bound attempt once, global attempt once, error surfaced, no loop. The fix is better than my one-line suggestion: the `voiceBindingKey(resolvedVoice) === boundKey` guard plus the `MAX_NARRATOR_VOICE_FALLBACK_HOPS = 1` bound means even a pathological chain terminates within two synthesis attempts. Your regression test adopting the repro passes, as does the rest of the narrator fallback suite.

The bonus precheck fix (a070b6ce) is also the right shape: transient vendor failures (5xx/transport) during the existence list query now report `unknown` and fall back to the idempotent re-register path, while real client errors (401/403) still surface — that addresses the note from my first review.

Full `tests/audio` + `tests/server` suites pass locally (458/458). LGTM.

PR #1161, opened 2026-08-24
fix(editor): allow multi-tab edit mode by dropping the cross-tab edit lock

wyuc:
## Summary

Removes the cross-tab edit-mode lock (`edit-mode-lock.ts`, `useEditModeLock`, `MultiTabEditConflictPrompt`) and allows entering edit mode from multiple tabs.

## Why

The lock is an **entry gate only** — the storage write layer underneath has always been last-write-wins (`DocumentStore` has no compare-and-swap; `saveStageDataIncremental` writes per-scene or whole-document with the current tab's snapshot). And because the lock lives in `localStorage`, it only ever covered **same-browser, same-profile multi-tab** — never multiple devices, browsers, or users. It blocked the narrowest slice of concurrent editing while signaling that edit conflicts were handled.

Removing it trades that partial gate for a simpler, honest model: edits are last-write-wins, from any tab.

## What can now race (stated plainly)

- Two tabs editing the **same scene**: the later flush wins; the earlier tab's changes are silently lost.
- **Structural changes** (insert/delete/reorder scenes) escalate to a whole-document save carrying that tab's full `scenes` snapshot — tab A inserting a slide can silently roll back scene edits tab B already saved, and vice versa. This is the easiest case to underestimate.
- Whole-value keys (`currentSceneId`, `outlines`, `generationComplete`) are last-write-wins even from non-editing tabs (a tab paging through playback overwrites `currentSceneId`).
- `order` renumbering computes from each tab's stale snapshot; concurrent inserts can produce duplicate/gapped order values.
- The 15-second stale-lock window after a crashed tab is gone; a reloaded tab and a surviving tab coexist immediately.
- On server-backed document stores the write surface was always multi-device; the lock never covered that — this change removes the appearance that it did.

## What is preserved

- Chat sessions keep their cross-tab observed-sessions baseline and merge (unaffected).
- Regeneration vs. manual edits remains guarded by the separate regen lease.
- The auto-exit effect (leaving edit mode when the stage becomes non-editable) is unchanged.

## Follow-up path

`DocumentStore` has no revision/etag, so a "remote changed" prompt has nothing to hook onto today — every save reports success. If optimistic concurrency is added to the store later, silent overwrite can be upgraded to detectable conflict. That is deliberately out of scope here.

## Changes

- Commit 1 extracts `enterEditMode` (pure function) from the Pro toggle, behavior unchanged.
- Commit 2 deletes the lock (3 source files + its test), rewires `components/stage.tsx`, and adds replacement tests including re-add fences (source grep, file absence, locale keys).
- Commit 3 removes the `edit.multiTab` strings from all 12 locales and updates stale comments.

Squash-merge recommended.


lib/store/stage.ts
```diff
@@ -545,8 +545,8 @@ const useStageStoreBase = create<StageState>()((set, get) => ({
   insertSceneAfter: (anchorSceneId, scene) => {
     // Pro mode slide management entry point — inserts after the anchor and
     // rebalances `order` so PPTX export / array position stay consistent.
-    // Edit mode is gated against active regeneration (see useEditModeLock),
-    // so rewriting `order` here is safe — no outline matcher is racing us.
+    // Edits are last-write-wins and concurrent tabs may interleave, so
```
CXuPercy: Nit — this rewritten justification is weaker than (and doesn't actually support) the claim. Last-write-wins with interleaving tabs is precisely the context in which rewriting `order` from a local snapshot is *risky*, not safe; concurrent inserts from two tabs can produce duplicate/gapped `order` values, as the PR description itself states.

The real guard for "no outline matcher is racing us" is the regeneration lease (the PR description correctly notes regen vs. manual edits stays protected by the separate lease). The old comment attributed that safety to the wrong mechanism (`useEditModeLock`); this rewrite swapped in a non-sequitur instead of the right one. Suggest:

```ts
// Regeneration is gated by the regen lease, so no outline matcher is
// racing us here; cross-tab `order` collisions are last-write-wins.
```

CXuPercy (dismissed):
Read the full diff and verified the rationale against the deleted code.

- The lock was indeed an entry gate only over an always-last-write-wins storage layer, and localStorage-scoped (same browser + profile) — it never covered multi-device, while signaling that concurrent editing was handled. Removing it in favor of the honest model is the right call, and the "what can now race" section in the description is exactly the kind of trade-off documentation a removal PR should carry.
- The `enterEditMode` extraction is faithful (teardown + preload in parallel, activate only on success, stay in playback on failure) and is properly covered by the new functional tests.
- The deletion is clean — no dangling references, comments updated at every touchpoint.
- The re-add fences (source grep, file-absence, locale-key sweep across all 12 locales) are a nice touch to keep the removal from silently regressing.

Approving as a conscious ratification of the product trade-off: silent last-write-wins in same-browser multi-tab (worst case, structural changes rolling back the other tab's saved scene edits) in exchange for multi-tab availability and a simpler, honest model — with optimistic concurrency correctly deferred until `DocumentStore` grows a revision/etag.

One documentation nit inline; nothing blocking.

lib/store/stage.ts
wyuc: Applied verbatim — thanks, the regen lease is indeed the mechanism doing the work here; the previous rewrite had swapped in a non-sequitur.

wyuc:
Branch updated with main and your suggested comment wording applied verbatim (docs-only change since your approval — the dismissal was the stale-review automation). Could you re-stamp when convenient?

CXuPercy (approved):
Re-stamping. Verified the delta since my approval is exactly one docs commit (`1d98352d` — `lib/store/stage.ts`, 2 lines: the order-rewrite comment now attributes safety to the regen lease, applied verbatim); everything else in the range is `main` flowing in through the branch update (#1154, #1156). No PR-owned behavior changed.

PR #1162, opened 2026-08-24
test: stop loading .env.local into unit tests by default

wyuc:
## Summary

Unit tests no longer load `.env.local` by default. Loading becomes opt-in via `TEST_LOAD_LOCAL_ENV=1`.

## Why

`tests/setup-env.ts` used to inject the developer's entire `.env.local` into `process.env` for every unit test. CI has no `.env.local` at all, so any test that needed it would already be failing there — loading it cannot make a test pass anywhere it wasn't already passing. It can only invent failures that exist on one machine and nowhere else: keys like `DATABASE_URL` or `STORAGE_PROVIDER` steer suites onto entirely different code paths than CI exercises.

The suite now runs hermetic by default, matching CI. For the rare case that genuinely wants local credentials (a smoke test against a real endpoint), opt in explicitly:

```
TEST_LOAD_LOCAL_ENV=1 npx vitest run <suite>
```

Shell-exported variables are untouched either way — this only governs whether the file is read.


CXuPercy (approved):
Verified before approving:

- The setup file is genuinely wired into the unit suite (root `vitest.config.ts` `setupFiles`), and the package sub-suites (editor/renderer/storage) use their own setups — the scope is exactly what the title claims.
- CI behavior is bit-identical: no `.env.local` exists there, and the flag only governs whether the file is read. The change can only remove local↔CI divergence, never add it.
- Shell-exported variables keep precedence either way (`if (!process.env[key])` fills only unset keys), matching the description.
- No other workflow references the old behavior — no `TEST_LOAD_LOCAL_ENV` consumers, no CI flag, and CONTRIBUTING's `.env.local` guidance is about running the app, not the unit suite.
- The opt-in escape hatch for real-endpoint smoke runs is documented right where someone hitting a missing-key failure will land.

Clean, minimal, well-argued. Approve.

PR #1163, opened 2026-08-24
feat(storage): add an agent-session store with PG backend and layered contracts

wyuc:
## Summary

Adds an **agent-session store** to `@openmaic/storage`: persistence for long-running, multi-process background agent sessions — session lifecycle, lease-based claiming with stale-lease stealing, an append-only event log with generation fencing, an append-only session entry tree, and a per-owner replayable event projection.

This is the storage foundation for the agent workbench line landing on `integration/agent-workbench`; nothing in the app consumes it yet (package-only change plus the CI trigger for the integration branch).

## Shape

Follows the package's existing house patterns throughout:

- **Backend-agnostic core**: the PG backend depends only on the injected `Queryable` / `WithTransaction` contract — no driver dependency (node-postgres, PGlite, and host adapters all fit).
- **Pinned schema**: `AGENT_SESSION_PG_SCHEMA` is a gold-standard constant (every statement `CREATE TABLE|INDEX IF NOT EXISTS`, `;`-splittable for `ensureSchema`), pinned by the schema-contract test alongside the existing stores. Table names are remappable via a `tableNames` option.
- **Layered contracts** (a deliberate extension of the single-contract pattern): `runAgentSessionStoreContract` covers single-process semantics and runs on every backend; a separate concurrency contract covers what only a real database can prove — competing claims, stale-lease stealing, in-lock watermark reads, and deterministic results under contention — and runs on PostgreSQL. The PG16 suite enforces `STORAGE_PG_CONTRACT_REQUIRED=1` so it cannot skip silently on CI.
- **Injected host hooks** (`resolveFinalOwner`, `onSessionCreated`, `onUserMessagePosted`) let a host attach identity resolution and side-band bookkeeping without the store knowing those concepts; the transaction and lock-order contract for hooks is documented on the types.

## Key semantics (argued in the docblocks)

- Claiming is two-phase: an unlocked candidate scan followed by an in-lock recheck, so claim reasons and event watermarks always come from locked reads.
- Appends are generation-fenced: an event carrying a stale attempt is dropped rather than interleaved into a newer run.
- The owner projection deliberately swallows its own failures (business writes must not be vetoed by projection breakage; clients repair via full resync) — the one place the store is not fail-loud, and the docblock argues why.
- Soft-deletion is a tombstone; child data is retained.

## Testing

- Full storage package suite passes (PGlite + in-memory paths).
- Real PostgreSQL 16 contract suite passes (store contract + concurrency contract), verified locally in Docker; the `storage-pg-contract` CI job covers it on every push.
- Repo-root typecheck, prettier, eslint clean.

### DIFF
--- a/packages/@openmaic/storage/src/agent-session/pg.ts
+++ b/packages/@openmaic/storage/src/agent-session/pg.ts
@@ -802,9 +802,19 @@ export class PgAgentSessionStore
            AND EXISTS (SELECT 1 FROM ${this.table('sessions')} s
                        WHERE s.id = $1 AND s.deleted_at IS NULL)
          ORDER BY e.seq LIMIT $3
+       ), span AS (
+         -- One extra row on each side of the page so lag/lead see the true
+         -- neighbors across page boundaries instead of NULL at the edges; the
+         -- anchor row (seq = $2) is the cursor's own row, discarded below.
+         SELECT e.seq, e.type
+         FROM ${this.table('events')} e
+         WHERE e.session_id = $1 AND e.seq >= $2
+           AND EXISTS (SELECT 1 FROM ${this.table('sessions')} s
+                       WHERE s.id = $1 AND s.deleted_at IS NULL)
+         ORDER BY e.seq LIMIT $4
        ), ranked AS (
          SELECT seq, lag(type) OVER (ORDER BY seq) AS prev_type,
-                lead(type) OVER (ORDER BY seq) AS next_type FROM page
+                lead(type) OVER (ORDER BY seq) AS next_type FROM span
        )
        SELECT p.seq, p.ts, p.attempt, p.type, p.data, p.scanned
        FROM page p INNER JOIN ranked r ON r.seq = p.seq
@@ -812,7 +822,7 @@ export class PgAgentSessionStore
           OR r.prev_type IS DISTINCT FROM 'message_update'
           OR r.next_type IS DISTINCT FROM 'message_update'
        ORDER BY p.seq`,
-      [sessionId, afterSeq, limit],
+      [sessionId, afterSeq, limit, limit + 2],
     );
     return {
       events: result.rows.map((row) => ({
--- a/packages/@openmaic/storage/src/agent-session/pg.ts
+++ b/packages/@openmaic/storage/src/agent-session/pg.ts
@@ -1,0 +1,1284 @@
+/**
+ * PostgreSQL backend for durable agent sessions.
+ *
+ * The backend imports no database driver. A host supplies a direct queryable
+ * and a transaction hook that checks out and pins one connection for the
+ * complete callback. READ COMMITTED isolation is assumed by the parent-row
+ * lock followed by max-plus-one child allocation used by both append logs.
+ */
+import { randomUUID } from 'node:crypto';
+
+import type { QueryResult, Queryable, WithTransaction } from '../runtime/pg.js';
+import {
+  AGENT_SESSION_LIFECYCLE,
+  AgentSessionAccessError,
+  AgentSessionEntryTreeError,
+  AgentSessionLeaseLostError,
+  type AgentSessionEntry,
+  type AgentSessionEntryTree,
+  type AgentSessionEntryTreeHandle,
+  type AgentSessionEventLog,
+  type AgentSessionHooks,
+  type AgentSessionMeta,
+  type AgentSessionStore,
+  type AgentSessionTransaction,
+  type AgentSessionUserMessage,
+  type ClaimedAgentSession,
+  type ClaimAgentSessionOptions,
+  type CreateAgentSessionInput,
+  type FinishAgentSessionPatch,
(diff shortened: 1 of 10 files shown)


packages/@openmaic/storage/src/agent-session/pg.ts
```diff
@@ -0,0 +1,1316 @@
...
+           AND EXISTS (SELECT 1 FROM ${this.table('sessions')} s
+                       WHERE s.id = $1 AND s.deleted_at IS NULL)
+         ORDER BY e.seq LIMIT $3
+       ), ranked AS (
+         SELECT seq, lag(type) OVER (ORDER BY seq) AS prev_type,
+                lead(type) OVER (ORDER BY seq) AS next_type FROM page
+       )
+       SELECT p.seq, p.ts, p.attempt, p.type, p.data, p.scanned
+       FROM page p INNER JOIN ranked r ON r.seq = p.seq
+       WHERE p.type <> 'message_update'
```
CXuPercy, line 821: Non-blocking, verified empirically: the "keep only a `message_update` run's first and last" compaction holds **within a page** but leaks at page boundaries, because `lag`/`lead` see NULL beyond the page edges.

Repro on this branch (PGlite): `before` + 10× `message_update` + `after`, then `readEventsAfterForReplay(…, 0, 6)` and `(…, 6, 6)` returns ids `[1, 2, 6]` and `[7, 11, 12]` — the run's survivors are **2, 6, 7, 11** (four frames instead of the ideal 2 and 11), and 6→7 are consecutive mid-run updates, precisely what the compaction exists to collapse.

Impact is low — replay frames are idempotent-shaped and `message_update` isn't written by this package yet (it's the host's future type) — but the contract test name ("compacts only middle message updates on replay") doesn't hold across pages. Either carrying one row of lookback across pages or documenting the boundary behavior in the docblock would pin it.

packages/@openmaic/storage/src/agent-session/pg.ts
```diff
@@ -0,0 +1,1316 @@
...
+        text: input.text,
+        seq,
+        delivery,
+        clientRequestId,
+      });
+      let requeued = false;
+      if (parent.status === 'queued') {
+        await tx.query(
+          `UPDATE ${this.table('sessions')}
+           SET attempt = 0, error = NULL, cancel_requested_at = NULL, updated_at = now()
```
CXuPercy, line 709: Nit (docs only): posting to an already-`queued` session silently clears `cancel_requested_at` — a pending cancel request is wiped by the next user message. That reads like a deliberate choice (a new message means the user re-engaged, so the cancel is moot), but it's a real semantic decision that `postUserMessage`'s docblock doesn't state. One sentence on the interface (e.g. "a message to a queued session clears any pending cancel request") would keep hosts from rediscovering it the hard way.

CXuPercy (approved):
Cross-review from my side: read the full implementation (types + PG backend, 1.7k lines) and the layered contracts, probed the risky surfaces empirically, and ran the package suite locally.

**Verified (not just read):**

- **Identifier injection surface**: `quoteIdentifier` rejects all seven attack shapes I threw at it (SQL injection, comment terminator, uppercase, leading digit, dash, space, embedded quote) — at both `ensureAgentSessionSchema` and construction. The custom-`tableNames` schema rewrite is safe under the validated charset, and its index-name derivation is pinned by the author's own test.
- **Two-phase claiming**: the in-lock recheck is authoritative, claim reasons and `claimSeq` come from locked reads, self-stealing is excluded, and the PG concurrency contract covers single-winner claims, the lock-order smoke test, and merge-under-contention result consistency. The contract's commentary about what the race can and cannot prove (deadlock freedom following from the lock-order invariant rather than the probe) is exactly right.
- **Generation fencing**: both append paths check worker identity *and* attempt under the parent row lock — a stale attempt is dropped (run events) or throws `LeaseLostError` (tree).
- **Projection swallowing**: SAVEPOINT-isolated, business write commits on projection failure, with the reconciliation obligation stated on the type. `mergeOwner`'s deterministic lock order, the resolver-collapsed self-merge guard, and the above-high-water renumbering all check out.
- The MAX+1 child allocation is always preceded by the parent `FOR UPDATE`, consistent with the documented READ COMMITTED assumption.

One local-run note: my machine sees one failing test in `http-asset-store.test.ts` (multipart parser error-message mismatch), but it reproduces identically at the merge-base — pre-existing and environment-specific (Node 24), not from this PR; CI is green.

Two non-blocking inline comments. Nothing here blocks merge — this is excellent infrastructure documentation practice.

packages/@openmaic/storage/src/agent-session/pg.ts
wyuc: Fixed in d23767e8 — the replay window now carries one row of lookback/lookahead across page boundaries, and your exact repro is a contract test (survivors [2, 11] across two pages of 6). The queued-clears-cancel semantics is now stated on the postUserMessage docblock as suggested.

PR #1164, opened 2026-08-24
feat(document): derived assets with lineage and an extraction cache (#1153 part 1)

wyuc:
## Summary

Part 1 of RFC #1153: extracted media become derived pool assets with recorded lineage, and extraction becomes a cacheable derivation. After a successful extraction, every extracted image is ingested into the asset pool and the extraction artifact (with inline image bytes replaced by their pool asset ids) is stored as a pool asset; a derivation record — lineage plus cache index in one KV entry — is written under a key of `content digest × domain × extractor id@version × endpoint fingerprint`. Re-importing the same bytes (same course or another) rebuilds the parse result from the pool and skips the paid extraction entirely.

The extract route is untouched: cache lookup and write live entirely on the client side of the seam, in one new module (`lib/document/extraction-cache.ts`).

## Related Issues

Part 1 of #1153 (design sections 2–3; settles open question 1: the artifact is a pool asset, the lookup index is a KV record that doubles as the lineage record). Builds on part 0 (#1154). The provider `version` field part 0 introduced is now consumed by the cache key.

## Changes

- **Derivation records** (`lib/document/extraction-cache.ts`, new): one KV entry per (digest × domain × extractor@version × endpoint fingerprint) holding `sourceDocAssetId`, the artifact asset id, per-image lineage (pool asset id, page number, description, dimensions), and alias identities. Key `v3` shape documented in-module, including the accepted residual staleness (in-place engine upgrade / account change behind one endpoint) and the per-principal constraint the future KV server backend (#1000 Part B) must honor.
- **Asymmetric trust with a TTL**: a hit under the record's own identity is permanent; a hit that validates only via an alias (written when the provider that actually ran differs from the one the lookup expected, e.g. the route's self-host-MinerU → MinerU-Cloud fallback) is honored for 24 h — expiry re-extracts and converges the cache after a server-side reconfiguration the client cannot observe.
- **Write path is best-effort and non-blocking**: the user's extraction result returns immediately; ingest → artifact → record runs detached, releases its allocations on partial failure (release outcomes logged), re-checks for a concurrent writer before the final set, and never gates or fails generation. Route-level KV unavailability (today: every server-backed deployment, until #1000 Part B) disables caching for a bounded 10-minute window with a single warning and zero asset churn.
- **In-run dedupe**: two same-byte sources in one generation run share a single extraction (memo keyed by the full cache key).
- **Digest at upload** (`app/page.tsx`): SHA-256 computed alongside the pool ingest and carried through the session (`contentDigest` on `SelectedCourseMaterial` / `SessionDocumentSource`); digest failure only skips caching.
- **Hit fidelity**: the rebuilt result is field-for-field the route's wire shape for document (including zero-image), images-only, and media responses; a hit is refused on any inconsistency (record/artifact image mismatch, expired alias, malformed `createdAt`).
- **Tests**: 60 cases in `tests/document/extraction-cache.test.ts` covering key symmetry and aliasing, TTL asymmetry, fingerprint normalization, in-run dedupe, concurrent-writer adoption, degradation windows, release accounting, and wire-shape fidelity.

## Type of Change

### Steps to reproduce / test

1. `pnpm check` / `pnpm lint` (0 errors) / `tsc --noEmit` / `pnpm exec vitest run` → 465 files, 5083 tests passed.
2. Upload a PDF, generate, then re-import the same file into a new course: the second run logs a cache hit and issues no extract request; images render identically.
3. Change the extractor's base URL and re-import: cache miss (endpoint is part of the key), fresh extraction.

- Server-backed deployments cache nothing until the KV server backend lands (#1000 Part B): the KV route answers 404 today, which the module treats as a bounded disable window. Artifact/image assets still ingest into the (existing) server asset store; only the index is missing.
- Derived assets and records are cache-owned: no eviction and no cascade when a source material is removed — management arrives with the material library milestone of #1153.
- An engine upgraded in place behind the same endpoint (or an account change on it) can serve a stale hit; the extractor `version` field is the governance lever, and the 24 h alias TTL bounds the cross-identity case.


lib/document/extraction-cache.ts
```diff
@@ -0,0 +1,1275 @@
...
+      baseUrl: options.baseUrl,
+      fetchImpl: options.fetchImpl,
+    });
+    if (cached) return { data: cached, cacheHit: true };
+  }
+
+  // 2. Real extraction (asset-id JSON form with the legacy byte fallback).
+  const response = await fetchExtractionResponse(options);
+  if (!response.ok) {
+    const errorData = (await response.json()) as { error?: unknown } | null;
```
CXuPercy: Nit — empirically, a non-ok response with a non-JSON body surfaces as a raw `SyntaxError` instead of `parseFailedMessage`. Probed on this branch: returning `<html>502 Bad Gateway</html>` with status 502 from the byte form yields

```
SyntaxError: Unexpected token '<', "<html>502 "... is not valid JSON
```

This is behavior-parity with the old page code (same unguarded `json()`), so not a regression — but the new module's own docblock for `parseFailedMessage` promises "Localized fallback for a response that carries no usable parse data", and a proxy error page is exactly a response carrying no usable parse data. The success-path `json()` at line 1161 has the same shape. A `.catch(() => null)` on both reads would make the fallback universal:

```ts
const errorData = (await response.json().catch(() => null)) as { error?: unknown } | null;
```

CXuPercy (dismissed):
Cross-review from my side: read the full module (1.3k lines), the page integration, and the browser-safe manifest, then ran your 60-case suite plus three probes of my own against this branch — everything passes.

**Verified:**

- **Cache key soundness**: digest × domain × extractor@version × endpoint fingerprint; a version bump or endpoint change misses. The `domain` split correctly keeps alidocmind's dual registration from collapsing document and media derivations of the same bytes into one key.
- **Fingerprint normalization** (probed): `MinerU.example` / `:443` / trailing-slash spellings share one bucket.
- **Asymmetric trust**: primary hits permanent, alias hits bounded to 24 h, and the write path *supersedes* an expired alias so a reconfigured deployment converges — the loop closes correctly.
- **Write path accounting**: record written last; releases on failure and on losing the concurrent-writer race; an alias-write failure deliberately does not release assets the primary record references. The residual get→set race window is honestly documented.
- **Degradation window**: route-level 404 (≠ `KEY_NOT_FOUND`) → 10-minute disable, checked before any ingest, one warn per episode.
- **Hit fidelity**: record/artifact image cross-check (K6) plus per-image resolution — any inconsistency is a miss.
- **Browser-safe manifest**: providers spreading their entry from the manifest (drift impossible by construction), insertion order pinned, and a guard test against future Node-only imports — a clean solution to keeping `sharp`/cloud SDKs out of the client bundle.

**One observation to put on record** (probed): the config fingerprint excludes URL credentials — `https://user:secret@host` and `https://host` hash identically. That is consistent with the documented residual staleness ("an account change on the same endpoint"), so it's confirmation of the stated posture rather than a new issue; noting it here so it's deliberate on the record.

One small inline nit — non-blocking, and notably *not* a regression (the old page code had the same unguarded `json()`).

lib/document/extraction-cache.ts
wyuc: Fixed in 808de545 — both body reads now carry .catch(() => null), so a proxy error page (or any non-JSON body) on either the error or the success read surfaces the localized fallback; two tests pin exactly your 502-HTML probe and the non-JSON success body. Your credentials-in-URL observation is now also on the record in this thread as deliberate — it falls under the documented account-change-on-same-endpoint residual. Thanks for the empirical probes again.

PR #1168, opened 2026-08-24
feat(materials): material library manifest and id-addressed generation images (#1153 part 2)

wyuc:
## Summary

Part 2 of RFC #1153: the material library — a per-principal, KV-backed manifest of imported source documents — and id-addressed generation images. Every imported source now has a durable, library-owned pool reference (surviving selection removal), and on server-backed deployments the generation flow references extracted images by allocated asset id end-to-end: the client sends `imageMapping` as id → asset id, the generation routes resolve bytes server-side at prompt-assembly time, and `resolveImageIds` writes the allocated id into `PPTImageElement.src` (the renderer already resolves allocated ids through the pool). Browser-backed deployments keep today's base64 flow byte-for-byte.

## Related Issues

Part 2 of #1153 (design §4–5). Builds on parts 0 (#1154) and 1 (#1164): the library points at part 1's derivation records for lineage instead of duplicating it, and `listMaterials()` / `readMaterial()` are the two agent-facing atomic tools §4 describes.

## Changes

- **Material library** (`lib/materials/library.ts`, new): entries keyed by content digest (`assetId`, name, MIME, size, `addedAt`, derivation pointers). The library OWNS its pool reference — upsert allocates its own entry from the file (physically free under content addressing), so selection removal and prep-timeout releases never dangle it; re-import advances the reference and releases the previous one only after the entry points at the new id, with an ambiguous-commit re-read guard (leak-over-dangling when a lost response may have committed). Derivation-pointer merges are read-merge-write. Failures degrade with the part-1 bounded-window pattern.
- **Id-addressed vision** (`app/api/generate/scene-content` + `scene-outlines-stream`, `lib/persistence/resolve-vision-images.ts`): the routes resolve the vision slice BEFORE prompt assembly from a shared slice-ordering helper (route and generator cannot drift), refilling past unresolvable ids so a drop admits the next image with its resolution; unresolvable ids are stripped from the mapping, so text mentions, attachments, and element references stay consistent (a hallucinated reference to a dropped id removes the element instead of writing a dangling src). The resolution phase is bounded (15 s aggregate budget raced per probe + a 3-consecutive-failure fuse; either stop degrades to fewer/no images, never a failed request) and size-capped via the shared 50 MB constant at the identity read (no oversized materialization).
- **Cache integration** (`lib/document/extraction-cache.ts`): on server-backed runs, a cache hit rebuilds `imageMapping` as asset ids with no byte materialization, while per-image existence probes (bounded concurrency 8, aggregate budget) keep the reclaim-is-miss invariant mode-independent; a miss learns the derived ids from the (bounded-await, still detached) cache write, falling back per source to data URLs.
- **Per-source transport**: a bundle can mix id-fed and byte-fed sources; `imageMapping` mixes asset ids and data URLs and `resolveImageIds` dispatches by shape. Classroom resume builds the same merged mapping.
- **Tests**: ~170 new assertions across library lifecycle (durability after removal, re-import swap, write-failure release/keep/skip matrix), multi-drop refill parity, hallucinated-id element removal, mixed mappings, timeout fallbacks, probe budgets, oversized-asset drop, and an import-graph guard (now also scanning re-exports) for the new client-reachable modules.

## Type of Change

### Steps to reproduce / test

1. `pnpm check` / `pnpm lint` (0 errors) / `tsc --noEmit` / full vitest (5243 passed) / generation package (136 passed) / `pnpm build` — all green (Node 22).
2. Browser-backed: upload → generate — unchanged base64 flow.
3. Server-backed: upload → generate — network tab shows `imageMapping` carrying `ast_…` ids and no image-byte upload; slides render source images via pool resolution; removing the selected file after upload and calling `readMaterial(assetId)` still resolves (library-owned reference).

- The library manifest and cache index live in KV, which server deployments cannot serve until #1000 Part B; until then both degrade to browser-local with bounded-window probes (same posture as part 1).
- Library entries have no deletion/management surface yet — that UI and the entry-release path it implies are the next milestone of #1153; library-owned allocations accumulate until then.
- The agent-facing tools ship as the two pure functions (`listMaterials` / `readMaterial`); wiring them into an agent tool registry is intentionally left to the consumer.

### DIFF
--- a/app/api/extract-document/route.ts
+++ b/app/api/extract-document/route.ts
@@ -24,13 +24,13 @@ import {
 } from '@/lib/persistence/resolve-server-asset';
 import { apiError, apiSuccess } from '@/lib/server/api-response';
 import { validateUrlForSSRF } from '@/lib/server/ssrf-guard';
+import { MAX_EXTRACT_DOCUMENT_FILE_SIZE_BYTES } from '@/lib/constants/generation';
 
 // The asset-id path resolves bytes from the server asset store, which lives in
 // the PostgreSQL persistence backend; it needs the Node runtime, not the edge.
 export const runtime = 'nodejs';
 
 const log = createLogger('Extract Document');
-const MAX_EXTRACT_DOCUMENT_FILE_SIZE_BYTES = 50 * 1024 * 1024;
 
 /**
  * A normalized extraction input, independent of how the bytes arrived: either
--- a/app/api/generate/scene-content/route.ts
+++ b/app/api/generate/scene-content/route.ts
@@ -28,6 +28,7 @@ import { resolveModelFromRequest } from '@/lib/server/resolve-model';
 import { resolveVocationalActive } from '@/lib/config/feature-flags';
 import { MAX_VISION_IMAGES } from '@/lib/constants/generation';
 import { sortDocumentImagesForVision } from '@/lib/document/bundle';
+import { DEFAULT_INGEST_AWAIT_TIMEOUT_MS } from '@/lib/document/extract-source';
 import {
   resolveVisionImagesForPrompt,
   type VisionPromptImage,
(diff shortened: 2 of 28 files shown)


CXuPercy (approved):
Full cross-review done on 10859782 (everything below was verified there), re-checked on this head: the only delta is the one-line 0.3.1 bump, and the version guard passes locally (I had reproduced the CI failure it fixes).

**Verified:**

- **Library-owned reference model**: the upsert allocates its own pool entry; the re-import swap is ordered allocate-new → write entry → release-old, so an entry never names a released id; and the ambiguous-commit re-read guard (a `set` that rejected may have committed) chooses leak-over-dangling with a well-argued rationale — a leaked allocation is recoverable by re-import, a dangling one breaks `readMaterial`.
- **N7 merge-on-write**: derivation pointers unioned by identity on re-read, so concurrent writers cannot drop each other's pointers.
- **Resolve-with-refill**: shared `partitionImagesForVision` (route and generator cannot drift); the 15 s aggregate budget is raced *per probe*, so a single hanging store round-trip cannot outlive it; the 3-consecutive-failure fuse stops an all-fail phase; drops — including unprobed candidates after a stop — are stripped from the mapping so a hallucinated id takes the remove-element path; one summary warn; degrade, never fail.
- **Id-addressed transport**: `resolveImageIds` writes mapping values verbatim — the transport is decided entirely by the mapping's shape, no flag threading — and the in-closure resolution remains as a defensive no-op for package consumers. The 50 MB cap applies at the identity read, and unresolvable ids are dropped server-side, never echoed and never handed to the LLM as opaque strings.
- **Cache integration**: the asset-id hit mode materializes no bytes while the bounded (concurrency 8 + aggregate budget) existence probes keep the reclaim-is-miss invariant; the per-source byte fallback replaces the bundle-level all-or-nothing decision, so one source's failed cache write no longer silently drops its images.
- **Tests**: all new suites pass locally (root 219/219, generation package 136/136).

The concurrent-same-digest-upsert orphan (loser's allocation) is documented and accepted — same class as part 1's get→set window, resolved by the part-3 deletion surface.

Approving now that the version bump unblocks CI; remaining checks are re-runs.

PR #1172, opened 2026-08-24
feat(media): convert inline base64 images to pool assets on load (#1153 part 4)

wyuc:
## Summary

Part 4 of RFC #1153 — the last delivery-plan item: a lazy converter in the #1101 mold that, on document load, rewrites inline base64 image `src` values into pool-backed allocated ids. Pre-part-2 documents carry source images as inline data URLs (the same image on five slides is stored five times, and every load/save moves those bytes through the document store); after conversion the bytes live once in the asset pool and the document carries ids. Pure slimming: no DSL version bump is needed because the slot type already admits both shapes — data URLs are concrete addresses, ids resolve through the pool (part 2's model) — so an unconverted document stays fully valid.

## Related Issues

Part 4 of #1153 (delivery plan). Mirrors the #1101 converter's structure and reuses its narrowed load+save window, orphan-rollback ledger, and load-path hooks (migration + classroom fetch, running after the reference converter).

## Changes

- **`lib/media/convert-inline-image-assets.ts`** (new): enumerates every slide media slot via `slideMediaReferenceSlots` (canvas images, whiteboards, backgrounds, video posters); converts only unambiguous `data:image/*` values — the `;base64` marker is detected per the fetch spec (case-insensitive, space-tolerant, final-parameter-only), and the converter's accept set is a proven strict subset of the browser's, so every divergence lands on "keep inline", never on a wrong decode path. Zero-byte and oversized payloads (estimated decoded bytes over the shared 50 MB bound, checked before decoding) stay inline.
- **Dedup and accounting**: one allocation per sha256 of the decoded bytes (identical images across slides — even across encodings — collapse to one registry entry); pool write → Dexie compatibility mirror → document rewrite, with each failure point releasing exactly what it allocated and clearing the in-flight memo so sibling slots retry; fresh allocations ride the shared conversion ledger so a failed save-back or superseded load rolls everything back through the existing helper.
- **Bounded work**: a 15 s aggregate budget per pass (expiry keeps progress; the next open continues — including for classroom documents, which now re-run the pass on every open like the migration path), bounded whiteboard concurrency, and immediate abort propagation when a load is superseded.
- **Degradation**: `crypto.subtle` unavailable (non-secure contexts) → the pass is a no-op with one warning, per-pass, never a failed load.
- **Reclaim parity**: converted ids are stage-referenced assets — enumerated, reference-counted, and reclaimed on stage deletion exactly like part-2's id-addressed images (the Dexie mirror row carries the reclamation).
- **Tests**: 40 across a unit suite (fetch-spec marker matrix incl. safe-direction shapes, dedup/encoding collapse, budget, abort, failure accounting, idempotency, non-target shapes untouched) and a wiring suite against the real pool + real Dexie (save-back window, rollback, deletion reclaim, classroom birth + retry paths).

## Type of Change

### Steps to reproduce / test

1. `pnpm check` / `pnpm lint` (0 errors) / `tsc --noEmit` / full vitest (5290 passed) / `pnpm build` — all green (Node 22).
2. Open a pre-part-2 document with repeated inline images: after load, `PPTImageElement.src` values are allocated ids, the pool holds one entry per unique image, and a second open is a no-op (no save).
3. Kill the pool mid-conversion: the document keeps its inline values for the failed slots and converts them on the next open.

### Review notes (verified during an adversarial pass, recorded to save re-derivation)

- Two-tab concurrent conversion is safe: id spaces are disjoint, the reload-before-save re-converts on top of the winner, and a losing tab's rollback can only touch its own allocations.
- Converting an SVG data URL does not change its security context (both shapes feed the same `` sink, scripts inert either way).
- Slot pairs (video src/mediaRef/poster) cannot half-convert; `shape.pattern` is not enumerated and deliberately stays inline.

### DIFF
--- a/lib/classroom/load-classroom.ts
+++ b/lib/classroom/load-classroom.ts
@@ -275,9 +275,10 @@ export async function fetchClassroomFromApi(
   // degrades to the app's lock-free route instead of silently no-op'ing the
   // cold load; the degraded pass still keeps its ledger+rollback discipline,
   // so a failed unlocked attempt leaves nothing behind.
-  const [{ mutateDocument }, converter] = await Promise.all([
+  const [{ mutateDocument }, converter, inlineConverter] = await Promise.all([
     import('@/lib/document-store'),
     import('@/lib/media/convert-legacy-asset-refs'),
+    import('@/lib/media/convert-inline-image-assets'),
   ]);
   try {
     return await mutateDocument(
@@ -300,19 +301,30 @@ export async function fetchClassroomFromApi(
             shouldConvert,
             allocated,
           );
+          // The inline base64 image converter (#1153 part 4) runs AFTER the
+          // legacy converter — the two are independent, and this order lets
+          // the gen-placeholder converter see its expected shapes first. Both
+          // share one allocation ledger, so a superseded or failed load rolls
+          // every fresh allocation back through the single rollback call.
+          const inlined = await inlineConverter.convertInlineImageAssets(
+            converted.document,
+            undefined,
+            shouldConvert,
+            allocated,
+          );
           // The converter rechecks liveness at its commit boundaries; this is
           // the final gate before the document write, so a load superseded in
           // the last window never commits side effects it cannot hand over.
           if (!shouldConvert()) {
             throw new converter.LegacyConversionAbortedError(allocated);
           }
-          if (converter.containsClassroomMediaUrls(converted.document)) {
+          if (converter.containsClassroomMediaUrls(inlined.document)) {
             throw new Error('server classroom media conversion was incomplete');
           }
-          await store.saveDocument(converted.document);
+          await store.saveDocument(inlined.document);
           return {
-            stage: converted.document.stage,
-            scenes: converted.document.scenes,
+            stage: inlined.document.stage,
+            scenes: inlined.document.scenes,
           };
         } catch (error) {
           // Every failure mode -- a liveness abort or an ordinary converter
--- a/lib/classroom/load-classroom.ts
+++ b/lib/classroom/load-classroom.ts
@@ -286,11 +286,45 @@ export async function fetchClassroomFromApi(
       async (existing, store) => {
         if (existing) {
           // The document already owns its media (a concurrent cold load, or a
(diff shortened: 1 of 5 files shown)


CXuPercy (approved):
Cross-review from my side: read the full converter and both wiring points, ran your 41 tests (all pass locally), and wrote my own probe suite for the parts the docblock makes claims about. Everything checks out — including two probe cases where my own first expectations turned out to be wrong and the implementation was right, which says something about the grammar's rigor.

**Verified empirically:**

- **Fetch-spec marker grammar, full matrix**: case-insensitive, spaces between `;` and `base64` tolerated, final-parameter-only enforced. `base64` followed by trailing spaces correctly does NOT match (the spec only tolerates spaces before the marker; the browser percent-decodes that shape) — my probe initially expected otherwise and the code was right. A stray `;base64` token stays inline, never guessed.
- **Strict-subset direction holds**: every divergence between this decoder and the browser's lands on keep-inline — invalid UTF-8 percent payloads throw in `decodeURIComponent` and are kept; binary percent-encoded payloads (outside the real-world corpus; the docblock correctly scopes percent payloads to UTF-8 SVG) are kept. No reachable wrong-decode path.
- **Cross-encoding dedup**: the same SVG as base64, as percent-encoding, and repeated collapses to exactly one allocation (probed: one id, three rewrites).
- **Estimate math**: base64 padding excluded (`QQ==` estimates 1 byte, not 3); percent payloads use the raw length as a safe upper bound — and the gate runs pre-decode, so oversized payloads never allocate decode buffers.
- **Failure accounting**: the in-flight *promise* memo (not settled values) is the right call for concurrent whiteboard slides naming one image; the pool write living inside the guarded region means a transient failure clears the memo so sibling slots retry; an abort between allocation and mirror compensates the allocation.
- **Wiring**: the migration path runs this pass after the #1101 converter with a shared ledger and independent best-effort degradation; the classroom path re-runs on every open with liveness gates on both sides of the save-back and ledger compensation on failure.

Clean approve — nothing to nit. This closes out RFC #1153's delivery plan parts 0/1/2/4 through review.

PR #1173, opened 2026-08-24
feat(whiteboard): add destructive runtime operations

LING-6150:
## Summary

Add RuntimeStore-backed destructive and incremental code-edit operations to the Native Child whiteboard tool set, building on the read/visibility composition from #1156.

The mutation path keeps RuntimeStore as commit authority: exact replay is resolved before CAS and transition, each fresh operation appends at most once, uncertain append outcomes receive at most one read-only reconciliation, and projection is emitted only for confirmed commits.

## Related Issues

Related to #1049

Builds on #1152 and #1156

## Changes

- Add versioned Runtime whiteboard operations for element deletion, clearing all elements, and code-line insert/delete/replace edits.
- Add strict payload validation and deterministic folding for the new operation vocabulary.
- Add `wb_delete`, `wb_clear`, and `wb_edit_code` to the existing Native whiteboard inventory.
- Share operation-agnostic mutation settlement across additive and destructive Native tools.
- Preserve exact replay, CAS, definitive transition errors, one reconciliation at most, and confirmed-commit-only projection.
- Keep missing/already-empty clear as a successful zero-append, zero-projection no-op.
- Expose `wb_edit_code` as a root object JSON Schema while retaining strict operation-specific branches for provider portability.
- Add Browser/Pg RuntimeStore parity coverage and production-route integration coverage.

## Type of Change

- [ ] Documentation update
- [x] CI/CD or build changes

### Steps to reproduce / test

1. Run the focused Native/RuntimeStore suites listed below.
2. Run the PostgreSQL contract suite through the `storage-pg-contract` workflow.
3. In a classroom Q&A, create two elements, delete exactly one, then create a code block and perform line replacement/insertion/deletion without redrawing the block.

- Focused suites: 109 passed; the PostgreSQL suite was collected and skipped locally when `PG_CONTRACT_URL` was absent.
- Native tool and production-route suites passed, including the production-shaped additive replay regression.
- PostgreSQL 16 storage contract passed in CI.
- ESLint passed for every changed TypeScript source/test file.
- Prettier and `git diff --check` passed.
- Browser production path manually verified clear, exact element deletion, nonexistent-target handling, and code-line replace/insert/delete while preserving unrelated board content.
- Full Vitest in the physical validation copy reached 5204 passing tests. Eight unrelated publish-shell cases require Git metadata unavailable in that copy; two unrelated timeout cases passed when rerun in isolation (42/42).
- Full workspace typecheck retains existing renderer package-export diagnostics; no changed-file diagnostic appeared.

### Cross-review

- Independent read-only full-diff review found one reachable P2: replaying an additive operation after a later delete/clear could misreport a confirmed historical commit as failed.
- Fixed at `6b1556b9`: exact additive replay now derives its affected result from the exact operation payload while preserving the current authoritative state and record tail.
- Production-shaped regression covers draw → delete → replay: original `committedSeq`, current `lastSeq`, no additional append, and no element restoration.
- Full-diff re-review at `6b1556b9` reported no actionable P0–P3 findings and approved the change.
- A separate `WHITEBOARD_RUNTIME_CODE_ELEMENT_NOT_CANONICAL` candidate was withdrawn after tracing production reachability: Legacy elements are normalized before append and non-canonical Runtime snapshots are rejected by payload validation.

### Browser E2E evidence

1 Codex Image Aug 24, 2026, 03_39_39 PM

2Codex Image Aug 24, 2026, 03_40_01 PM

In-place code-line replacement and insertion preserve the code block's identity, position, and style.

3 Codex Image Aug 24, 2026, 03_40_12 PM

Before/after element deletion: the selected blue circle is removed while the code block, red rectangle, labels, and other board content remain.

Deleting the two inserted code lines restores the original four-line code block without redrawing or moving it.

### Evidence

- [x] Manually tested locally

## Checklist

### DIFF
--- a/lib/whiteboard/runtime/validate.ts
+++ b/lib/whiteboard/runtime/validate.ts
@@ -245,11 +245,13 @@ function validateElement(element: PPTElement): string | null {
   return schema ? validateSchema(element, schema, 'element') : 'unknown element type';
 }
 
-function validateIdentifierArray(value: unknown, label: string): asserts value is string[] {
+function validateCodeLineTargetIds(value: unknown, label: string): asserts value is string[] {
   if (!Array.isArray(value) || value.length === 0) throw new Error(`${label} must be non-empty`);
   const ids = new Set<string>();
   for (const id of value) {
-    if (!isSafeIdentifier(id)) throw new Error(`${label} contains an invalid id`);
+    // Existing CodeLine.id values are plain strings in the frozen DSL contract. Targets must
+    // remain able to address every line accepted by legacy import, including an empty ID.
+    if (typeof id !== 'string') throw new Error(`${label} contains an invalid id`);
     if (ids.has(id)) throw new Error(`${label} contains a duplicate id`);
     ids.add(id);
   }
@@ -480,18 +482,18 @@ export function validateWhiteboardRuntimePayload(
         if (!hasExactOwnKeys(edit, CODE_LINES_INSERT_EDIT_KEYS)) {
           throw new Error('edit is invalid');
         }
-        if (!isSafeIdentifier(edit.lineId)) throw new Error('lineId is invalid');
+        if (typeof edit.lineId !== 'string') throw new Error('lineId is invalid');
         validateCodeLines(edit.lines, 'edit.lines', { nonEmpty: true });
       } else if (edit.kind === 'delete_lines') {
         if (!hasExactOwnKeys(edit, CODE_LINES_DELETE_EDIT_KEYS)) {
           throw new Error('edit is invalid');
         }
-        validateIdentifierArray(edit.lineIds, 'edit.lineIds');
+        validateCodeLineTargetIds(edit.lineIds, 'edit.lineIds');
       } else if (edit.kind === 'replace_lines') {
         if (!hasExactOwnKeys(edit, CODE_LINES_REPLACE_EDIT_KEYS)) {
           throw new Error('edit is invalid');
         }
-        validateIdentifierArray(edit.lineIds, 'edit.lineIds');
+        validateCodeLineTargetIds(edit.lineIds, 'edit.lineIds');
         validateCodeLines(edit.lines, 'edit.lines', { nonEmpty: true });
       } else {
         throw new Error('edit kind is invalid');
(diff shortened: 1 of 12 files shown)


lib/whiteboard/runtime/validate.ts
```diff
@@ -422,6 +460,42 @@ export function validateWhiteboardRuntimePayload(
...
+        throw new Error('operation is invalid');
+      }
+      if (!isSafeIdentifier(operation.elementId)) throw new Error('elementId is invalid');
+      const edit = objectValue(operation.edit);
+      if (!edit) throw new Error('edit is invalid');
+      if (edit.kind === 'insert_after' || edit.kind === 'insert_before') {
+        if (!hasExactOwnKeys(edit, CODE_LINES_INSERT_EDIT_KEYS)) {
+          throw new Error('edit is invalid');
+        }
+        if (!isSafeIdentifier(edit.lineId)) throw new Error('lineId is invalid');
```
wyuc: [P2] Permit edits to target legacy-valid line IDs

The frozen persisted code-line contract still accepts empty line IDs, and this PR explicitly tests that existing add/import operations remain valid with them. However, this validator and the `wb_edit_code` tool schema reject every edit targeting an empty ID. A valid persisted code block can therefore be exposed by `wb_read` but cannot be inserted around, deleted, or replaced through the new API. Please either accept target IDs according to the existing `CodeLine` contract, or deterministically migrate legacy IDs before exposing editing.

wyuc (changes requested):
One compatibility issue blocks merging this as-is. The persisted code-line contract and the new editing contract need to agree so every readable whiteboard remains editable.

lib/whiteboard/runtime/validate.ts
LING-6150: @wyuc Fixed in e889b5d6.
The issue was that the persisted CodeLine.id contract allows empty strings, while wb_edit_code had narrowed target IDs to non-empty safe identifiers.
The fix:
- accepts existing target lineId / lineIds according to the persisted string contract;
- keeps elementId and newly written line IDs strict;
- generates deterministic host-owned IDs instead of reusing invalid legacy IDs during replacement;
- preserves historical duplicate IDs and only rejects conflicts introduced by the current edit.
Added a production-shaped BrowserRuntimeStore → Runtime service → Native tool regression covering insert and replace on a legacy code block with duplicate empty line IDs.
Verification: 110 passed, 1 skipped (PG_CONTRACT_URL unset); ESLint, Prettier, and git diff --check passed. Please re-review when convenient.

wyuc (approved):
The compatibility issue is resolved. The legacy target-ID contract remains readable and editable, while newly written line IDs stay strict and deterministic.

PR #1175, opened 2026-08-25
feat(media): resolve image, video, and ASR models from server config

wyuc:
## Problem

The image generation chain never consulted the server-side `IMAGE_<PREFIX>_MODELS` config: the route only forwarded the client `x-image-model` header, and each adapter silently fell back to a hardcoded vendor default (`gpt-image-2`, `qwen-image-max`, `image-01`, ...). A managed provider could therefore generate with a model the operator never chose, and when the client sent no provider preference the route hardcoded `seedream`. This was the one media section whose `_MODELS` env was never consulted — TTS already resolves its server-managed settings properly.

## Change

- **Server-side resolvers**: `resolveImageModel` / `resolveVideoModel` / `resolveASRModel` plus `resolveServerImageProviderId` / `resolveServerVideoProviderId` in `lib/server/provider-config.ts`, with one shared allowlist semantics: when the operator pins `<CAP>_<PREFIX>_MODELS`, an allowlisted client choice wins and the first entry is the managed default; with no pin, the client model wins.
- **Fail loud at the routes**: when neither the client nor the server config yields a provider/model, the generate and verify routes return `400 MISSING_PROVIDER` / `400 MISSING_MODEL` (existing error envelope) instead of silently picking a vendor default.
- **Adapters require an explicit model** (`lib/media/require-model.ts`) as a defense-in-depth backstop; connectivity probes keep their fixed probe payloads since they are auth handshakes, not user-facing generation.
- **Classroom media generation** (internal caller with no HTTP response) falls back to the provider's first catalog model when nothing is pinned, preserving the existing behavior of key-only deployments; a server pin stays authoritative when set.
- Model lists from YAML are normalized like env lists (trim + drop empties), and the transcription route trims the client `modelId` like the other routes trim their headers.

## Deliberately out of scope

- ASR keeps its adapter-level default model when nothing resolves (pre-existing, documented behavior of `asr-providers.ts`); only the resolver and trimming were added there.
- The `sora` video catalog entry has no adapter case in the dispatch switch (pre-existing gap, unchanged).

## Tests

`tests/server/provider-config.test.ts` pins the resolver precedence matrix for all three capabilities; `tests/server/classroom-media-generation.test.ts` adds key-only-no-pins regression tests for both image and video; adapter tests updated to pass explicit models. `tsc`, prettier, eslint, and the `tests/media` + `tests/server` suites pass (694 tests).


CXuPercy (approved):
Cross-review from my side: read the resolvers, both fail-loud routes, the classroom fallback, `require-model`, and sampled the adapter changes; ran the media + server suites locally (460 passing, including your precedence matrix and the key-only regressions).

**Verified:**

- The three capability resolvers share exactly the stated semantics — pinned → allowlisted client choice wins, first entry is the managed default; unpinned → client model wins — consistent with the TTS governance pattern, appropriately without its qwen-specific complexity.
- `resolveServerImageProviderId` / `resolveServerVideoProviderId` replace the hardcoded `seedream` for no-preference requests with the first operator-configured provider, and nothing configured fails loud with `MISSING_PROVIDER`.
- The `MISSING_MODEL` boundary is precise: it fires only when no model resolves *and* the provider has a catalog — workflow providers (comfyui) that legitimately run model-less are not collateral damage.
- `requireModel` is wired into all fourteen adapters while connectivity probes keep their fixed payloads (auth handshakes, not user-facing generation) — the exemption is the right call.
- The classroom internal path keeps key-only deployments generating (pin's first entry authoritative, else first catalog model) with the adapter backstop explicitly documented as never the primary failure mode.
- `normalizeModelList` turning whitespace-only pins into "not configured" is a nice catch — an empty-string entry can never act as a truthy-but-garbage managed default, and it applies to both the YAML and env paths.
- Scope honesty: the ASR adapter-level default and the sora dispatch gap are genuinely untouched.

One local-run note: the two `convert-*-wiring` test files fail inside a full-batch run on my machine (localStorage environment bleed between workers) but pass 12/12 in isolation — pre-existing test-ordering fragility, not this PR; CI is green.

Clean approve — the governance gap this closes (a managed provider silently generating on a model the operator never chose) was real, and the fix is tight.

PR #1181, opened 2026-08-25
feat(providers): uniform capability force-off and consistent missing-key contract

wyuc:
## Problem

Capability enable/disable was inconsistent across providers:

- Only TTS had an explicit operator force-off (`TTS_<PREFIX>_ENABLED=false`); image, video, ASR, and web search were purely "key present ⇒ enabled" with no way to turn a capability off without deleting credentials.
- A keyed TTS provider with no key failed as **500 GENERATION_FAILED** (library throw into the generic catch), while the image route returns a clean **400 MISSING_API_KEY** — two contracts in the same repo.

## Change

- **Force-off generalized to image, video, ASR, and web search**, mirroring the TTS semantics exactly (only an explicit `false` disables; unset is no opinion; an env `true` overrides a YAML disable; the flag can never enable a keyless provider). Implemented as one per-section disable map derived from the existing env maps rather than four copies; client-only keyless providers (browser-native ASR/TTS, ComfyUI image) are coverable too.
- Disabled providers are skipped by server-default selection (including the web-search preference chain and classroom media generation), rejected with **403 PROVIDER_DISABLED** by the generate/verify/transcription/web-search routes, reported `{disabled: true}` by `/api/server-providers`, and excluded from client availability, fallback re-pointing, and first-run auto-select.
- **TTS missing-key is now a 400 MISSING_API_KEY pre-guard** matching the image route; keyless providers are never pre-empted and unknown providers keep their existing path.
- Review-pass hardening: the built-in Doubao web-search provider is now mapped (previously it could not be disabled at all), browser-native ASR honors the server force-off on the client (it never reaches the transcription route), and a request with no ASR provider now selects the first enabled server backend — or fails 400 MISSING_PROVIDER — instead of hard-defaulting to a vendor.
- `.env.example` documents the new `<CAP>_<PREFIX>_ENABLED` knobs per section.

Backward compatible: with no new env set, behavior is unchanged except the TTS missing-key contract fix and the removal of the ASR vendor default.

## Tests

1,373 tests across the affected suites (server/media/web-search/store/audio/settings), including per-section force-off semantics, route 403s, disabled-skipping server defaults, client re-pointing, and the classroom only-provider-disabled degradation. tsc/prettier/eslint clean.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -265,6 +265,9 @@ SEARXNG_BASE_URL=
 # Dedicated MiniMax web-search vars avoid conflicting with the LLM MINIMAX_* endpoint.
 WEB_SEARCH_MINIMAX_API_KEY=
 WEB_SEARCH_MINIMAX_BASE_URL=https://api.minimaxi.com
+# Dedicated Doubao web-search vars avoid conflicting with the Doubao LLM provider.
+WEB_SEARCH_DOUBAO_API_KEY=
+WEB_SEARCH_DOUBAO_BASE_URL=https://open.feedcoopapi.com
 # Claude (Anthropic) native web search. Dedicated vars avoid conflicting with
 # ANTHROPIC_* LLM provider vars. Optional WEB_SEARCH_CLAUDE_MODELS pins the
 # search model server-side (first entry wins), e.g. claude-sonnet-5.
@@ -274,6 +277,7 @@ WEB_SEARCH_CLAUDE_MODELS=
 
 # Operators can force-disable any built-in web search provider. Examples:
 # TAVILY_ENABLED=false
+# WEB_SEARCH_DOUBAO_ENABLED=false
 # SEARXNG_ENABLED=false
 
 # Server-only, default-OFF selector for the Native Child execution harness.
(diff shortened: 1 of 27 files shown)


CXuPercy (approved):
Cross-review from my side: read the disable-map implementation, all four route guards, the TTS pre-guard, the server-default selection changes, and the client-side flow for browser-native ASR; ran the affected suites locally (321 passing across the force-off, route, missing-key, provider-config, web-search, settings-sync, and classroom suites).

**Verified:**

- The disable map is one implementation for all five sections (not four copies), with the right precedence semantics: YAML `enabled: false`, env `true` overrides, and — the detail I'd have asked about — unset/empty env values are "no opinion" and never silently clear an explicit YAML disable. LLM/PDF stay credential-driven, as documented.
- `PROVIDER_DISABLED` 403 fires before managed/key resolution on every generate/verify/transcription/web-search route, mirroring the TTS contract.
- The TTS missing-key fix is precisely scoped: `requiresApiKey && !apiKey` → 400 `MISSING_API_KEY`; keyless providers are never pre-empted and unknown providers keep their existing path.
- Server-default selection (`resolveServer*ProviderId`) skips disabled entries everywhere, including the web-search preference chain and classroom media generation.
- The client flow closes the loop end-to-end: `/api/server-providers` reports `{disabled: true}`, the settings store carries `serverDisabled`, and browser-native ASR — which never reaches the transcription route — is gated client-side.
- The Doubao web-search mapping means every built-in provider in that section is now coverable.

The head commit closing self-found force-off bypasses is a good sign the review passes were real. Clean approve.

PR #1182, opened 2026-08-25
feat(server): validate model routing config at startup

wyuc:
## Problem

The repo has no startup validation (`.env.example` says so explicitly): a typo'd MODEL_ROUTES stage, an unregistered provider prefix, a route whose provider has no key, or a stray `<PREFIX>_MODELS` pin is discovered at request time — or, for the agent driver stage, as a per-session failure.

## Change

- New `lib/server/config-validation.ts` with `validateServerConfig()`, wired into `instrumentation.ts` (nodejs runtime only, dynamically imported so nothing new reaches the edge bundle). **Warn-only and provably non-throwing** — a broken config never prevents boot; `[config]`-prefixed warnings name exactly what is wrong.
- Warning matrix: MODEL_ROUTES JSON validity and shape; unknown stage keys (typo detection, lists valid stages); unknown provider prefixes; routed providers without a configured key (fatal at request time since routed stages ignore client keys); DEFAULT_MODEL checks with softer wording (unrouted resolution still honors client-supplied keys); `<PREFIX>_MODELS` pins on unconfigured providers (keyless providers with a default base URL are fine and not flagged).
- **Deprecation, warn-first**: a bare model id (no `provider:` prefix) in MODEL_ROUTES or DEFAULT_MODEL warns that the silent `openai` fallback is deprecated. The fallback behavior is unchanged this release; removal can follow after a deprecation period. The warning fires only for config-derived sites — deliberately never from `parseModelString` itself, where request-controlled strings could drive log volume.

## Tests

19 cases pinning the full warning matrix (including the no-warnings-on-valid-config and never-throws-on-garbage properties, YAML-configured providers, keyless defaults) plus the request-path never-warns contract. Full `tests/server/` suite green; tsc/prettier/eslint clean.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -321,9 +321,11 @@ DEFAULT_MODEL=
 # model as x-model. Unset = identical to today (everything uses DEFAULT_MODEL).
 # Unlisted stages fall back to x-model then DEFAULT_MODEL, so you can override
 # just one or two and leave the rest to the client/default.
-# Point a stage only at a server-configured provider (its key resolvable); a
-# route to an unconfigured/invalid model fails at request time for that stage,
-# the same way a bad DEFAULT_MODEL would (no startup validation).
+# At boot the server validates MODEL_ROUTES / DEFAULT_MODEL / <PREFIX>_MODELS
+# and prints [config] warnings for unknown stages, unregistered providers,
+# providers with no API key, and bare model ids (which still default to openai
+# but are deprecated — write provider:model). Warnings only: the server starts
+# regardless, so a bad value is caught here instead of at request time.
 # Routable stages: scene-outlines-stream, scene-content, scene-actions,
 # agent-profiles, quiz-grade, pbl-chat, chat-adapter, generate-classroom,
 # web-search-query-rewrite.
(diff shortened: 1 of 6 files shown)


CXuPercy (approved):
Cross-review from my side: read the validator and its wiring, checked the deprecation-warning placement across the repo, and attacked it locally with hostile configs (non-JSON, arrays, null, numeric route values, `:::` as DEFAULT_MODEL) alongside your 19 cases — 21/21 passing, nothing throws, every warning carries the `[config]` prefix.

**Verified:**

- Warn-only is real, not aspirational: a top-level catch plus per-parse guards means even unexpected internal errors degrade to one warning, and my garbage-config probe confirmed no input shape can take boot down.
- The warning matrix matches the description, with two touches worth calling out: unknown stages list the valid stage names (typo recovery without a doc dive), and the routed vs unrouted missing-key wording distinguishes a guaranteed request-time failure from a client-key-can-still-save-it note.
- The `<PREFIX>_MODELS` pin check correctly exempts keyless providers with a default base URL.
- The deprecation warning lives in exactly one non-test call site (config validation), and `parseModelString` explicitly does not warn — with a comment naming the reason (request-controlled strings must not drive logging or dedupe-set growth). That's the right boundary; the head commit that moved it there is the crux of this PR.
- Edge-bundle safety: `NEXT_RUNTIME` gate plus dynamic import keeps fs/js-yaml out of the edge.

Clean approve — filling the no-startup-validation gap with something that provably cannot make boot worse is exactly the right shape for a first pass.

PR #1193, opened 2026-08-26
test(render): wait for project cleanup instead of racing it

wyuc:
`render-coordinator.test.ts` asserted `await expect(access(dir)).rejects.toThrow()` immediately after `waitForJob` reported a terminal status — but project cleanup runs *after* the status transition, so the assertion raced the removal. It failed intermittently in CI (twice today on unrelated PRs, green on same-commit rerun each time), which makes every red build ambiguous until someone reruns it.

Replaces the three instantaneous cleanup assertions with a `waitForCleanup` poll shaped like the existing `waitForJob` helper (100 × 5 ms, then a clear timeout message).

Verified: 30 consecutive local runs of the suite, zero failures.

### DIFF
--- a/render-service/test/render-coordinator.test.ts
+++ b/render-service/test/render-coordinator.test.ts
@@ -50,6 +50,22 @@ async function waitForJob(
   throw new Error(`Timed out waiting for job ${id}`);
 }
 
+/**
+ * Project cleanup runs after the job reaches its terminal status, so a bare
+ * `access` right after `waitForJob` races the removal.
+ */
+async function waitForCleanup(path: string): Promise<void> {
+  for (let attempt = 0; attempt < 100; attempt += 1) {
+    try {
+      await access(path);
+    } catch {
+      return;
+    }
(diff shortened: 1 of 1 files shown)


CXuPercy (approved):
Verified before approving:

- The race is real, not guessed: `render-coordinator.ts` updates the job to its terminal status *before* `cleanupProject()` runs, so an `access` assertion fired immediately after `waitForJob` observes the transition races the removal — matching the intermittent-red-on-unrelated-PRs symptom exactly.
- All three flaky assertions are converted; the only remaining `access(` in the file is inside the new helper itself.
- The fix mirrors the existing `waitForJob` shape (bounded poll, clear timeout naming the path), so it reads like the file's established idiom rather than a one-off.
- 10 consecutive local runs of the suite pass (on top of your 30).

Small diff, outsized value: an intermittently red job makes every red build ambiguous until rerun — this restores "red means red". Clean approve.

PR #1195, opened 2026-08-26
test(providers): guard the provider-neutral layers against vendor leakage

wyuc:
A vendor-neutral base only stays neutral if the neutrality is enforced. Today it is enforced by review, which is why several generic layers already know concrete vendors: the model resolver owns one vendor's credential policy, the agent driver infers its transport from an OpenAI-named set, the voice route picks one vendor's clone model and maps that vendor's error class, TTS/ASR/web-search execution are central switches, and object-storage selection is an if/else over two vendors' environment variables.

This adds a CI guard for the mechanically-checkable half of that problem: **a vendor identifier must not appear in a module declared provider-neutral.**

## Design

- **Vocabulary is derived from the registries themselves** — the LLM, TTS, ASR, image, video, web-search, PDF, document/media extractor and voice-registration registries, plus the storage package's `./asset/*-bytes` exports. A hand-written vendor list would rot silently as vendors are added; this cannot. The test fails if a registry declaration disappears or changes shape, so the vocabulary can never go quietly empty.
- **The protected surface is explicit** (routing/resolution/driver/orchestration, shared provider config, the capability routes, search dispatch/config, extractor registries, both storage entry points). Catalogs, composition roots and concrete adapters are deliberately exempt — those are the correct homes for vendor knowledge.
- **TypeScript-aware scanning** of identifiers and string/template tokens, so comments don't trip it and a diagnostic names `file:line:column`, the vendor, the offending token, and the remediation (move it into an adapter, or add a neutral contract method).
- **Existing debt is a ratchet, not a suppression.** Every exception pins an exact file + vendor + occurrence count. Adding an occurrence fails; removing one makes the pin stale and also fails, so debt can only shrink. No entry suppresses a file or a vendor wholesale.
- **The red path is exercised**: a fixture route containing a new vendor ternary is scanned in-memory and the diagnostic is asserted. A guard whose failure branch never runs is not a guard.

Verified additionally by hand: adding one `'qwen-tts'` literal to an already-pinned file turns the suite red with the exact line and remediation.

## Scope

Test-only — no production behavior changes, and no existing violation is fixed here (each belongs with its own seam's refactor). The guard's value is that from now on the set cannot grow silently.

156 tests green across the guard and the representative provider suites; tsc/prettier/eslint clean.


packages/@openmaic/storage/test/pg-agent-session-store.pg.test.ts
```diff
@@ -0,0 +1,78 @@
...
+  let store: PgAgentSessionStore;
+
+  beforeAll(async () => {
+    pool = new Pool({ connectionString: contractUrl, max: 16 });
+    await ensureAgentSessionSchema(pool as Queryable);
+  });
+
+  beforeEach(async () => {
+    await pool.query(
+      `TRUNCATE agent_session_entries, agent_session_events,
```
CXuPercy, line 55: **CI-red root cause (reproduced locally against a fresh PG16 container).** This TRUNCATE lists the agent-session tables, but `src/material/pg.ts:52` adds `agent_session_materials` with `session_id REFERENCES agent_sessions(id)`. The PG16 CI suite runs every `*.pg.test.ts` against one shared database, so once `pg-agent-session-material.pg.test.ts` has ensured its schema, this `TRUNCATE … agent_sessions` is rejected with `cannot truncate a table referenced in a foreign key constraint` — all 24 cases in this file fail from that point.

Reproduction: full `vitest run` with `PG_CONTRACT_URL` → same 24 failures as CI; this file alone → green. So it's a cross-suite isolation bug, not store logic.

Any of these fixes works: add the materials table to this TRUNCATE, use `TRUNCATE … CASCADE` here (the schemas are all test-owned, so cascading into the FK-dependent table is safe), or give each PG suite its own schema/database. Whatever is chosen, the same review should ask whether the *next* FK-owning table will break this again — per-suite isolation fixes the class, not just this instance.

CXuPercy (changes requested):
Requesting changes for reasons that have nothing to do with the guard itself — and one of them is the point.

## 1. This PR's description does not describe this PR

The title and body describe exactly one commit: the vendor-leakage guard, 2 files, +873 lines. That guard is genuinely well-designed (registry-derived vocabulary that cannot silently rot, TS-aware scanning that ignores comments, a ratchet where every debt pin is file+vendor+count-exact so debt can only shrink, and an exercised red path) — **split it out and I will approve it immediately.**

But this branch is `integration/agent-workbench`, and the PR as opened carries **168 files / +28,159 lines beyond the guard**: the background session runner, the driver model contract, the skills system, `web_search` and `fetch_url` tools, the per-session URL trust gate, session materials, and the stage-document ownership scope. Those landed through 14 PRs (#1165–#1191) merged into the integration branch **with zero reviewers** (verified individually: every one shows `reviewers: []`, base `integration/agent-workbench`). The body's "Test-only — no production behavior changes" is true of the final commit and false of this pull request. Approving on the description would land ~28k lines of never-human-reviewed production code on `main` — including exactly the surfaces that deserve the most scrutiny (a `fetch_url` tool and URL trust gate are SSRF-adjacent; the ownership scope is a principal-isolation boundary).

Two ways forward, either fine:

- **(a) Split**: land the guard as its own honestly-described PR now; open the integration branch separately with a real description and per-feature review of the unreviewed line.
- **(b) Keep this PR** as the integration landing, but retitle and rewrite the description to say so, and give the 14 embedded PRs (at minimum #1186, #1189, #1190, #1191) real review passes before approval.

Note also that the previously-reviewed pieces have drifted: #1163's store evolved here (#1174 dropped the active-stage API; #1186/#1191 added surfaces), so "already reviewed" no longer covers the file contents.

## 2. CI is red — reproduced locally, root cause confirmed

`pg-agent-session-store.pg.test.ts` fails 24 cases with `cannot truncate a table referenced in a foreign key constraint`. Root cause (inline): the material store's table (`src/material/pg.ts:52`) references `agent_sessions(id)`, the PG16 CI suite shares one database across test files, and once `pg-agent-session-material.pg.test.ts` has ensured its schema, the agent-session suite's `beforeEach` TRUNCATE is rejected by that FK. I reproduced it exactly: full suite against a fresh PG16 container → the same 24 failures; the agent-session file alone → all pass. The fix is mechanical (truncate the materials table too, `TRUNCATE … CASCADE`, or per-suite schema isolation) but this must land green.

## 3. What I am not blocking on

The guard's own mechanics — I read the test; the vocabulary derivation, protected-surface enumeration, and red-path fixture are solid. It is the one part of this PR I would merge today, as its own PR.

wyuc:
You are right on both counts, and thank you for checking the reviewer count on all fourteen instead of taking the description at face value.

**On the base branch — my mistake.** I cut this branch from `integration/agent-workbench` and targeted `main`, so the diff is the whole integration delta, not my two files. The body I wrote ("test-only, no production behaviour changes") is true of the commit and false of the pull request. Approving it would have put ~34k lines of never-human-reviewed code on `main` under a test-guard title — including exactly the SSRF-adjacent and principal-isolation surfaces you named. The rule I should have followed, and will from now on: **the base branch must be the branch I intend to merge into**, and I check `git diff --stat <target>...HEAD` before writing a description.

Taking your option (a):

1. The guard is being re-cut on `main` as its own change. Its protected-file list has to shrink to the surfaces that actually exist on `main` (the agent-runtime files live only on the integration branch), and every debt pin is being recomputed against `main` rather than carried over — with a comment naming the surfaces to add when they land, so the trimming is visible and actionable rather than silent.
2. The integration branch will land through its own pull request, honestly described, with real review on the unreviewed line — at minimum the four you named (#1186 trust gate, #1189 skills, #1190 fetch_url + materials, #1191 ownership scope). Your point that "already reviewed" no longer covers #1163's store is taken: it has since been reshaped by #1174/#1186/#1191, so the file contents need fresh eyes, not credit for an earlier pass.

**On the FK/TRUNCATE failure — real defect, mine, being fixed separately.** Your root cause is exactly right and I could reproduce the shape from your description: the material store is the third table to reference `agent_sessions(id)`, and the shared-database suite's `beforeEach` truncate cannot survive that. I am fixing it so the suite is order-independent *structurally* rather than by adding one more table name to a list that the next store author will forget — a per-file list that must be edited every time a store gains an FK is the same bug waiting to recur.

I will close this PR once the `main`-based guard PR is open, and link them.

wyuc:
Correction to my previous comment, on the review model rather than the base-branch mistake.

The integration branch exists to make this migration fast, and per-slice human review is not the bar for it — **cross-review (an independent AI reviewer pass) on the security- and contract-relevant slices** is. So the plan is narrower and more honest than what I wrote:

- **Integration slices**: cross-reviewed where it matters, self-merged otherwise. Auditing my own record against that bar: only two of the fourteen actually got a cross-review pass (skills, and fetch_url/materials) — and each of those found something real, a quota race and a redirect that bypassed the trust gate. Three of the un-reviewed ones should not have been: the per-session URL trust gate, the stage ownership scope, and the strict SSRF helpers. A cross-review of those three is running now, against the merged state rather than the original diffs, and anything confirmed becomes a follow-up fix on the branch.
- **The landing PR to `main`**: one PR, honestly described, reviewed by main's code owners the way any other PR is. Not fourteen retrospective reviews.

That is the real gap your review surfaced — not "no human looked at it" but "the slices where an independent pass had the highest expected yield are exactly the ones that skipped it." Thank you for the push; the guard-on-main PR follows shortly and I will close this one against it.

wyuc:
Closing in favour of #1197, which is the same guard correctly based on `main`: one file, +517 lines, and the diff stat now matches the description. The protected-surface list is trimmed to what exists on `main` (with a comment naming the agent-driver surface to add back when it lands) and every debt pin was recomputed against `main` rather than carried over from the integration branch.

The two substantive points from your review are being handled separately, as they should be: the FK/TRUNCATE failure in the shared PG16 database is a real defect in merged code and is getting a structural fix (the material store was the third table to reference `agent_sessions`, so adding one more name to a per-file truncate list would just queue up the same failure for the next store), and a cross-review of the three security-relevant slices that never got one — the URL trust gate, the ownership scope, and the strict SSRF helpers — is running against their merged state.

PR #1179, opened 2026-08-25
fix(build): scope Next typecheck to production sources

YizukiAme:
## Summary

A clean Docker production build aborts during the Next.js TypeScript-checking phase (#1157). The optimized compile succeeds, then the type-check worker reaches the builder image's default V8 heap limit (~2,096 MiB) and dies with `SIGABRT`.

This scopes the production type-check to production sources. `next build` now reads a `tsconfig.build.json` that extends the root config and additionally excludes the test and eval trees. The saving comes less from the test files themselves than from the test-only `node_modules` type declarations they pull into the program (vitest, testing-library and friends): the type-check program drops from 7,960 to 6,915 files and from 410k to 253k lines of TypeScript.

No Dockerfile or `NODE_OPTIONS` change is included. Raising the heap ceiling would relocate the failure rather than detect it, and it is a build-resource policy decision independent of this fix.

## Related Issues

Fixes #1157

## Changes

- `next.config.ts`: point `typescript.tsconfigPath` at `tsconfig.build.json` when `NODE_ENV` is `production`, and at the root `tsconfig.json` otherwise.
- `tsconfig.build.json`: new config extending the root one, additionally excluding `tests`, `eval` and `packages/@openmaic/*/test`. TypeScript replaces an inherited `exclude` instead of merging into it, so the root entries are repeated and carry a comment saying they must stay in sync.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [x] CI/CD or build changes

### Steps to reproduce / test

1. On the parent commit `9b54baa6`, run `docker build --no-cache --progress=plain -t openmaic-1157:baseline .` — the build dies during `Running TypeScript ...`.
2. On this branch, run `docker build --no-cache --progress=plain -t openmaic-1157:final .` — the build completes and exports the image.
3. Start a container from that image and request `/api/health`.

### What you personally verified

Environment: Docker (aarch64), 18 CPU, 8,319,213,568 B VM memory; builder stage `node:22-alpine` (Node 22). The default V8 heap limit inside that image is 2,197,815,296 B (~2,096 MiB), as measured in #1157.

| Source | Builder heap | Result |
| --- | --- | --- |
[...]

PR #1196, opened 2026-08-26
fix(providers): three silent-behavior fixes from the provider audit

wyuc:
Three independent findings from a provider-decoupling audit, one commit each so they can be reverted separately.

## 1. A self-hosted extractor was silently switched to a cloud vendor (HIGH)

`app/api/extract-document/route.ts`: when self-hosted MinerU was unavailable, the route looked up MinerU **Cloud** and, if a cloud key happened to exist, silently substituted it — an operator who deliberately chose self-hosted extraction (so documents stay in their own infrastructure) would have those documents sent to a third-party cloud with no message. A test even pinned that silent behavior.

Now: 422 with an actionable error naming what was configured, what was unavailable, and both remedies. The cloud fallback still exists but requires an explicit `ALLOW_MINERU_CLOUD_FALLBACK` opt-in, documented in `.env.example` and **default off**. Tests pin the loud path, the opt-in path, and that the cloud parser is not called without the opt-in.

## 2. A selectable video provider had no adapter (MEDIUM)

`sora` was in the catalog (selectable in settings and via `VIDEO_SORA`) with an empty model list and no case in either the connectivity or generation switch, so choosing it failed only at execution time. Removed from the catalog, the id union, the UI maps, the store defaults, and the env map — no stub adapter.

The new `video-provider-catalog` suite pins the invariant forward: every catalog id must reach its own adapter in both switches, so adding an entry without dispatch now fails.

## 3. The nominal generic storage entry point was a silent no-op (HIGH)

`lib/storage/index.ts`'s `getStorageProvider()` unconditionally returned a no-op provider whose methods returned empty values, and the module swallowed errors into `null` — a caller believing it had storage got silence. A repo-wide grep found **zero callers**, so the abstraction (`index.ts`, `types.ts`, `providers/noop.ts`) is deleted; a would-be caller now gets a loud resolution failure instead. `lib/storage/client.ts` is kept — it has a live caller and a documented, deliberately-handled contract.

Tests pin the deletion, the loud failure, and the surviving client helper.

741 tests green across document + media; 279 across runtime and provider suites; tsc/prettier/eslint clean.


CXuPercy (approved):
Cross-review from my side: read all three fixes against the code they replace, independently verified the zero-caller claim, and ran the three suites (43/43). (Resubmitted after the branch update: the delta since my review is only a merge of main — #1193 and #1179 — with zero changes to this PR's own files.)

**Fix 1 (MinerU silent cloud substitution)** — the important one, and it's done right. The gate is strict (`ALLOW_MINERU_CLOUD_FALLBACK` parsed as exactly `'true'`/`'1'`, default off), and the 422 is exemplary operator messaging: it names what was configured (self-hosted), what is missing (its base URL), states explicitly that documents are *not* sent to MinerU Cloud automatically, and gives both remedies. This was a real privacy defect — an operator who deliberately kept extraction on their own infrastructure would have had documents handed to a third-party cloud without a message. The tests pin the loud path, the opt-in path, and that the cloud parser is never invoked without the opt-in. I also checked the client seam: the 422 carries `INVALID_REQUEST` (not `PARSE_FAILED`), so the byte-form fallback retries once, hits the same 422, and surfaces it — no loop, no double-bill.

**Fix 2 (sora)** — removed from the catalog, union, UI maps, store defaults, and env map with no stub adapter, and the new catalog invariant test (registry-derived ids must reach their own adapter in *both* switches) means the class of bug, not just the instance, is now pinned — same design philosophy as the neutrality guard.

**Fix 3 (no-op storage entry point)** — I re-ran the caller search independently: zero non-test references to `getStorageProvider` or the `@/lib/storage` root; the only live import is the kept `client.ts` helper (used by PBL submission). A would-be caller now fails loud instead of getting a silent no-op.

One trivial leftover (not blocking): the twelve locale files still carry the now-unreferenced `providerSora` translation keys — dead weight after fix 2, deletable whenever convenient.

Approving all three; the one-commit-per-fix structure with independent revertability is exactly the right shape for audit findings.

PR #1197, opened 2026-08-26
test(providers): guard the provider-neutral layers against vendor leakage

wyuc:
Re-opened correctly based on `main` — the previous attempt (#1195) was branched off an integration branch while targeting `main`, so its diff carried that whole branch instead of this test. This PR is **one file, +517 lines, test-only**, and `git diff --stat main...HEAD` says exactly that.

## What it guards

A vendor identifier must not appear in a module declared provider-neutral. Today several do — the model resolver owns one vendor's credential policy, the voice route picks one vendor's clone model and maps its error class, TTS/ASR/web-search execution are central switches — so this cannot be a clean bar on day one. It is a ratchet instead.

- **Vocabulary is derived from the registries themselves** (LLM/TTS/ASR/image/video/web-search/PDF/document-extractor registries and the storage package's byte-store exports). A hand-written vendor list would silently stop covering vendors added later; this cannot. If a registry declaration disappears or changes shape the test fails rather than quietly scanning for nothing.
- **The protected surface is explicit**, and trimmed to what exists on `main`: the agent-runtime driver surface lives only on the integration branch, so it is not listed — with a comment naming it and saying to add it back the day it lands, rather than leaving a silent hole.
- **TypeScript-aware**: identifiers and string/template literals only, so comments don't trip it; every failure names `file:line:column`, the vendor, the token, and the remediation.
- **Debt is a ratchet, not a suppression**: each exception pins file + vendor + exact count, recomputed against `main` rather than carried over. Adding an occurrence fails; removing one makes the pin stale and also fails. Debt can only shrink.
- **The red path is exercised** both by an in-memory fixture test and by hand: adding `'qwen-tts'` to an already-pinned file turns it red with the exact line, while the same token in a comment is correctly ignored.

1,584 tests green across the provider suites; tsc/prettier/eslint clean.

Two follow-ups already in flight that will retire pins here: the voice route's vendor knowledge moving behind its adapter, and #1196's removals (`sora` leaves the env map, the no-op storage module is deleted).

### DIFF
--- a/tests/providers/provider-neutrality-guard.test.ts
+++ b/tests/providers/provider-neutrality-guard.test.ts
@@ -1,0 +1,517 @@
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+import ts from 'typescript';
+import { describe, expect, it } from 'vitest';
+
+interface RegistrySource {
+  file: string;
+  declarations: readonly string[];
+}
+
+interface VendorLeak {
+  file: string;
+  line: number;
+  column: number;
+  vendor: string;
+  token: string;
+}
+
+interface AllowedLeak {
+  file: string;
+  vendor: string;
+  token: string;
+  count: number;
+  reason: string;
+}
+
+interface AllowedVendorDebt {
+  file: string;
+  vendor: string;
+  count: number;
+  reason: string;
+}
+
+/** Catalogs and adapter registries are composition roots, so their keys are authoritative vocabulary. */
+const REGISTRY_SOURCES: readonly RegistrySource[] = [
+  { file: 'lib/ai/providers.ts', declarations: ['PROVIDERS'] },
+  { file: 'lib/audio/constants.ts', declarations: ['TTS_PROVIDERS', 'ASR_PROVIDERS'] },
+  { file: 'lib/media/image-providers.ts', declarations: ['IMAGE_PROVIDERS'] },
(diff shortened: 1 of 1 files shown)


CXuPercy (approved):
As promised in my #1195 review: the guard split out onto main is an immediate approve from me.

Verified before approving:

- **The delta from the #1195 guard commit is exactly the main adaptation** — 28 lines: the protected surface drops `agent-driver-model.ts` (not on main) with an explicit comment to re-add it the day the agent driver lands, and the ratchet drops that file's temporary debt pins, which would have been stale-pin failures on main. Nothing else is carried.
- **Clean tree**: 3/3 green locally.
- **The red path is real, verified against the actual scanner on an actual file**: appending a `'qwen-tts'` literal to a protected file fails with `lib/server/resolve-model.ts:190:20 contains vendor token "qwen" in "'qwen-tts'". Move vendor behavior into an adapter, or add a provider-neutral contract method.` — exact file:line:column, vendor, token, remediation.
- **The ratchet semantics hold on that same file**: its pinned existing debt is tolerated while the *new* occurrence I injected fails — debt can only shrink.

Registry-derived vocabulary, TS-aware scanning, exact-count ratchet pins, an exercised failure branch — this is the right shape for a decoupling guard, and landing it on main now means the integration line arrives at a gate that is already armed. Clean approve.

PR #1209, opened 2026-08-26
fix: restore the reference's stage access model

wyuc:
A previous pass of mine replaced the reference's stage-access layer with an invented owner-scoped store. That invention is why the workbench could not work, and it changed product behaviour I had no business changing. This restores the reference's model.

## The split-brain, which is why this is a blocker

The agent runner bound the document store to the session owner; the browser's document path used the **unbound** store and served it with everything authorised. Different partitions. So the agent created a course, handed over its link, and the browser reported it missing. There is a regression test for exactly this: save through the agent's store, then load through the browser path with the same anonymous cookie, and get the document back.

## The policy I had wrong

The reference treats **the stage id as a read capability**: any holder may READ, only the owner may WRITE. I had made reads owner-only, which silently broke shared course links — they would render "not found" upstream while working in the reference. Restored:

- read by id is allowed; a foreign mutation is refused with the existing forbidden shape
- the generic document list is refused; the library route lists only the caller's own stages
- deletion tombstones through `stage_meta` and the id is retired — it cannot be written or recreated, and repeated owner deletion is idempotent

## The tool layer needed the same fidelity, not a blanket rule

The reference injects a **three-state** access probe into the toolsets (owned / exists-but-not-owned / absent) and each tool branches on it: folder mutations require ownership, while a stage read only requires readability. I had collapsed that into one owner-bound store, which is why two cross-owner tests failed once the HTTP layer was fixed — their premise encoded my invented policy. The probe is ported and each tool's guard now mirrors the reference per tool rather than following one rule of mine.

768 tests green across agent-runtime and persistence; the storage version gate passes.


wyuc:
Superseded: #1215 included this branch as its base and has been squash-merged, so the stage-access fidelity work is now on the integration branch in full. Closing as merged-via-1215.

PR #1224, opened 2026-08-27
feat(chat): add single PPT element references to Pi

LING-6150:
## Summary

- Add one global **Reference slide** entry to playback Pi chat.
- Support direct single-element picking for all 10 canonical PPT element types: Text, Image, Shape, Line, Chart, Table, LaTeX, Video, Audio, and Code.
- Resolve the browser-submitted element identity against the request-start Scene snapshot in the Host, then project bounded, type-specific evidence for the Director and Child.
- Keep the existing classic `/api/chat` path unchanged.

## Contracts

- `PlaybackChromeRoot` owns the draft selection and send snapshot; the local overlay is a UI transport seam only.
- Legacy and new renderers are picked through their real paint-node geometry, including current-transform remeasurement, Scene stacking order, and canonical fallback for elements without measurable paint nodes.
- The request contains identity only (`sceneId` and `elementId`). The Host re-resolves the element and rejects missing, duplicate, cross-scene, non-slide, or unsupported references before model execution.
- All 10 element projectors are deterministic and bounded. Truncation is Unicode-safe, item boundaries are preserved, and `truncatedFields` / `omittedItems` truthfully describe lost evidence.
- Chart routing hints include the projected numeric series before labels and legends, so the Director can delegate value and trend questions while the Child still receives the full evidence packet.
- Selected-element evidence is request-scoped and take-once, remains independent from pending `read_scene` evidence, and is available to both Legacy and Native Child paths.
- A grounded Child failure remains owned by that Child and does not fall back to an ungrounded replacement answer.
- The selection pill is retained when a request is rejected or not accepted. It is cleared only after the matching request receives `X-OpenMAIC-Element-Reference-Accepted: 1`; a late receipt cannot clear a newer draft.
- Typed sends, voice sends, and playback-engine interrupts freeze one selection snapshot and issue exactly one POST.
- Navigation and composer lifecycle cleanup prevent stale cross-page references while allowing same-scene Spotlight behavior to remain fail-closed.

## UI and localization

- Candidate menus and reference pills use localized element type names across all 12 locales.
- Shape summaries use embedded text, then element name, then a localized no-text label.
- Candidate and pill summaries share one bounded presentation contract; media and Line keep their existing truthful metadata/type descriptions.

### Global entry, element picking, and draft pill

Entering **Reference slide** mode exposes the selectable PPT elements. Selecting one creates a single, visible, draft-scoped reference pill.
And a Shape without embedded text remains selectable and is presented with the localized `Shape · No text` summary instead of a bare canonical type.

Reference slide entry, element picking, and draft reference pill

### Text and LaTeX grounding

Selected text or formula content is resolved by the Host and reaches the delegated classroom agent as grounded evidence.

Text and LaTeX element grounding in classroom Q&A

### Image metadata reference

Image elements can be selected and grounded through bounded metadata without claiming OCR or visual understanding in this phase.

Image element metadata reference

### Table structured evidence

Table rows and cells are projected as bounded structured evidence for the classroom answer.

Table element structured evidence in classroom Q&A

### Chart numeric-series grounding

The values `180`, `145`, `112`, and `88` reach both routing and Child evidence, allowing the answer to report the exact values and decreasing trend.

Chart numeric series values and grounded trend answer

### Code grounding

The grounded answer correctly explained the referenced Code element line by line, including the gravity constant, buoyancy calculation, and console output.

code -WechatIMG267

### Line grounding

The grounded answer correctly identified the referenced Line element as an upward arrow and explained that it represents the direction of buoyant force.

Line-WechatIMG268

### Audio metadata isolation

The grounded answer received only the Audio element ID, canonical type, position, and size. It correctly stated that `440Hz → 660Hz` and `4 seconds` came from nearby caption text—not from the referenced Audio element—and therefore could not confirm the audio content..

audio-WechatIMG269

### Video metadata isolation

The grounded answer received the Video element ID, canonical type, name, position, size, `autoplay=false`, `ext=mp4`, and `srcKind=reference`. It correctly stated that codec, bitrate, duration, resolution, and visual content were not provided, and that the nearby caption was not part of the referenced Video element’s metadata.

video- WechatIMG271

## Independent cross-review

- A read-only full-diff cross-review was repeated after the final fixes and found no actionable P0–P3 issues.
- The previously reported Chart routing gap is closed: valid numeric series reach the bounded Director hint and the full Legacy/Native Child evidence packet.
- Renderer-tolerated legacy Charts with missing data now produce bounded empty evidence instead of crashing the picker or Host route.
- `<script>` and `<style>` blocks are removed from normalized HTML evidence together with their contents.
- The review found no parallel Context system, persistent selection registry, new Native tool, renderer callback framework, or generic Runtime policy in the diff.

## Verification

- Focused Vitest suite: 11 files, 119 tests passed.
- TypeScript: `pnpm exec tsc --noEmit --incremental false` passed.
- ESLint: `pnpm lint` passed with 0 errors (15 pre-existing warnings).
- Prettier: `pnpm check` passed.
- Locale key alignment: `pnpm check:i18n-keys` passed for all 12 locales.
- Patch integrity: `git diff --check` passed.
- Real Browser validation covered Text/LaTeX, Shape, Line, Table, Image, Chart, Code, Audio, and Video flows. Audio and Video validation also confirmed that nearby captions are not treated as element-owned metadata.
- A production-shaped Route → Director → real `call_agent` regression verifies Chart series values `180`, `145`, `112`, and `88` reach both the Director prompt and Child evidence.

## Scope

This PR is limited to playback Pi chat and PPT Slide Scenes. It does not add Whiteboard or interactive-page selection, free-region marquee selection, multi-select, OCR/vision, editing authorization, renderer callbacks, a selection registry, a new Native tool, or generic Runtime policy.


wyuc (changes requested):
Thanks for the thorough implementation and test coverage. The code-level review did not find a separate regression, and I am fine with keeping the current navigation-clears-reference behavior for the initial A/B test.

I think the evidence ownership and failure flow still need one architectural adjustment before merge.

The selected element evidence should be immutable, request-scoped context shared by every relevant child call in the current user turn. It should not use take-once semantics. With the current pendingElementReferenceEvidence / takeElementReferenceEvidence flow, whichever valid call_agent happens first consumes the full evidence. A later retry or handoff receives only the ordinary conversational context, so grounding depends on the Director choosing the right child first and on the first child accurately restating the element.

The grounded-child failure rule appears to compensate for that one-shot ownership: once the only grounded child fails, the Director has no grounded retry path, so the whole loop is terminated. This adds a feature-specific terminal rule to an otherwise general agent loop and prevents the Director from retrying or handing off normally.

Please keep the resolved evidence available as read-only context for the lifetime of this request and inject it into each relevant call_agent invocation. Then a child failure can remain a normal tool failure: the Director may retry or hand off, and the next child is still grounded in the same element. The request should fail only when the ordinary loop exhausts its available recovery paths, without a separate groundedChildSucceeded termination branch.

In short, request-scoped shared evidence removes both the information-loss issue and the need for the special failure control flow.

LING-6150:
@wyuc     Thanks for the review , I’ve addressed the grounding lifecycle issue and pushed the update in 17db36d5.
The selected-element evidence is now retained as immutable, request-scoped context and passed to every valid Legacy or Native Child delegation. This preserves grounding across retries and agent handoffs without introducing a separate take/clear lifecycle.
I also removed the element-reference-specific failure control flow, so Child failures now follow the existing Director retry, handoff, and loop-limit behavior. The independent read_scene evidence remains take-once as before.
Spotlight remains fail-closed: selected-element evidence can authorize Spotlight only when it belongs to the request-start current Scene. I added a regression test covering repeated Native delegations with cross-scene evidence; both delegations receive the grounding context, but neither exposes nor dispatches Spotlight.
Verification is green:
- Pi suite: 20 files, 224 tests passed
- TypeScript passed
- ESLint passed
- Prettier passed
- git diff --check passed

Could you please take another look when convenient?

wyuc (approved):
Re-reviewed at 17db36d5. The selected-element evidence is now shared as immutable request-scoped context across Legacy and Native retries and handoffs, while the element-specific Director termination path has been removed. The targeted regression suite and all required checks are green.

PR #1227, opened 2026-08-27
fix(storage): asset writes self-deadlock against pooled PostgreSQL

wyuc:
## Symptom

Importing a course ZIP (or any flow that stores new asset bytes) hangs forever at "writing media files" on a deployment with pooled PostgreSQL and database-backed asset bytes. Reported independently by two users; both investigations reached the same root cause.

## Root cause

`lazyAssetByteStore()` (introduced in #1089) returns, in the no-S3-bucket case, a literal `{ write, read, delete }` object that does not forward the `writeWith`/`readWith` methods of the underlying `PgAssetByteStore`. `PgAssetStore.put()` locks the `asset_blobs` row inside its transaction, then — because the `hasTransactionalWriter` duck-check fails on the wrapper — falls back to plain `write()` on a **second pooled connection**, which blocks forever on the very row lock its caller holds. PostgreSQL cannot detect it (one side is idle-in-transaction). Single-connection test databases cannot reproduce it, the PG contract tests use `PgAssetByteStore` directly, and the route tests mock the byte store — so the real wrapper+registry combination was never covered.

## Fix

1. **Root**: the lazy wrapper forwards `writeWith`/`readWith`/`deleteWith` in the database-backed case, typed against the real signatures so it cannot drift. Object-store (S3) semantics are unchanged and it declares the new `writesOutsideRegistryDatabase` capability flag.
2. **Fail loud, never deadlock**: the registry (and the offline collector) refuse a same-database byte store that has neither transactional methods nor the out-of-registry declaration — a clear configuration error before any row is claimed, instead of a hang. Registry write transactions also set a 30 s `lock_timeout` so any future lock-wait bug surfaces as an error.

## Verification

- Red→green proven by restoring the pre-fix files: the wrapper capability pin (`hasTransactionalWriter(lazyAssetByteStore(undefined, db))`), the loud refusal with zero byte writes and zero claimed rows, the collector refusal, and the write-statement sequence (lock-budget prefix + in-transaction byte `UPDATE`) all fail before and pass after.
- `tsc --noEmit` 0 errors; `prettier --check` clean; version gate passes (`@openmaic/storage` 0.4.0 → 0.5.0).
- Storage suite 858 passed; app persistence 74 passed (includes the new wrapper coverage); document suite 299 passed.


CXuPercy (approved):
Verified locally against the PR head (worktree + full targeted test runs). Details and evidence for each point below.

**Root cause and fix confirmed along the full call chain.** `lib/persistence/server-provider.ts` wires the pool into `lazyAssetByteStore`; in the no-bucket case the wrapper dropped `writeWith`/`readWith`, so `PgAssetStore.put()` fell back to the plain `write()` on a second pooled connection against the very blob-row lock its own transaction had just taken — the self-deadlock. With the forwarding, byte traffic goes through the transaction-pinned queryable (the `SET, INSERT, UPDATE, INSERT` sequence asserted by the new test is exactly right).

**Guard placement is sound.** `assertByteWriteIsCoordinatable()` runs before the try/catch in both `put()` and `replace()`, so the configuration error is not collapsed into a generic registry failure; the collector guard throws before any row is locked and propagates through `collect()`. `SET LOCAL lock_timeout` is valid under both production wiring (`nodePostgresTransaction` uses an explicit BEGIN) and the PGlite test path, so it cannot leak across pooled sessions.

**Tests reproduced:** new wrapper tests 3/3; `tests/persistence/` 74/74; storage suite 857 passed + 1 failure in `http-asset-store.test.ts` (multipart parser error precedence) that also fails on current main with the same local Node — pre-existing and unrelated, and the totals match the PR description.

**The neutrality-guard reconciliation is necessary and exact, not a widening.** On current main the guard test is red three ways: it crashes with ENOENT on the removed `lib/storage/types.ts`; the extract-document route has 15 AST mineru leaks against an allowance of 9; and the sora allowance is stale at 0 occurrences. I verified this by running the guard test on origin/main (fails) and on this branch (passes), and by confirming the exact-pin semantics (lowering mineru to 14 fails with exactly one remaining leak). So beyond the deadlock fix, this PR also repairs main's red guard test — worth mentioning in the description as an argument for a prompt merge.

Two non-blocking notes:

1. The 30s `lock_timeout` is a behavior change for S3 deployments: the byte upload runs inside the registry write transaction while holding the blob-row lock, so a concurrent same-hash `put()` waiting behind an upload that exceeds 30s now fails loudly instead of eventually succeeding. The window looks narrow (the quota advisory lock is not wired in this app, the import flow writes sequentially, and the collector is re-runnable), and loud beats hang — but please confirm that trade-off is intended, or consider exempting stores that declare `writesOutsideRegistryDatabase`.
2. Nit: consider exporting `hasTransactionalWriter` from the package so `tests/persistence/asset-byte-store.test.ts` doesn't have to hand-copy the predicate it explicitly promises to keep in sync with `pg.ts`.

CXuPercy:

PR #1259, opened 2026-08-27
feat(agent): let generate_scene pass widgetType and widgetOutline for interactive pages

### LINKED ISSUE #1258: generate_scene agent tool cannot pass widgetType; interactive pages always fall back to simulation
## Problem

In the v1.0.0 agent workbench, the `generate_scene` tool cannot express which widget an interactive page should use. Every interactive scene generated through the agent ends up as a `simulation` widget, even when a `diagram`, `code`, `game`, or `visualization3d` widget would fit the brief.

## Root cause

1. The tool parameter schema has no widget fields. `SceneParams` in `lib/server/agent-runtime/generation-tools.ts` only accepts `stageId / order / title / type / brief / instruction / materialFacts / media`; `type` is limited to `slide | quiz | interactive | pbl`.
2. The handler builds a `SceneOutline` without `widgetType` / `widgetOutline`. A `widgetType` is only inherited when the scene happens to match a planned outline in the snapshot (`outlineFromScene`), which freshly generated pages never do.
3. The generation pipeline then falls back: `packages/@openmaic/generation/src/scene-generator.ts` (`generateSceneContent`) logs `Interactive outline "..." has no widgetType, falling back to simulation` and forces `widgetType: 'simulation'`.

The UI outline-editor flow is unaffected: it passes the full outline (including `widgetType`) through to `/api/generate/scene-content`.

There is also a capability mismatch: agent skills can declare `requiredWidgetTypes` (`lib/server/agent-runtime/skills.ts`), but no agent tool can satisfy that constraint for newly generated interactive scenes.

## Expected behavior

`generate_scene` should accept an optional `widgetType` (the ordinary widget types: `simulation`, `diagram`, `code`, `game`, `visualization3d`) plus an optional `widgetOutline`, validate that they are only used with `type: 'interactive'`, and pass them through to the outline so the widget generator produces the requested widget kind.

## Proposed fix

- Extend `SceneParams` with optional `widgetType` (enum of the 5 ordinary types) and optional `widgetOutline`.
- In the `generate_scene` handler, reject `widgetType` on non-interactive types with a clear tool error, and include the fields in the constructed `SceneOutline`.
- Keep `procedural-skill` out of the tool surface (it stays gated behind task-engine mode, same as `allowProceduralSkill` in the generation pipeline).
- Add tests covering: widget type passthrough, default simulation fallback when omitted, and rejection of widgetType on non-interactive scenes.

wyuc:
## Summary

Fixes #1258.

The `generate_scene` agent tool had no widget parameters, so every interactive page generated through the workbench fell back to a `simulation` widget (`generateSceneContent` logs "has no widgetType, falling back to simulation").

## Changes

- `SceneParams` gains optional `widgetType` (`simulation` / `diagram` / `code` / `game` / `visualization3d`) and optional `widgetOutline` (loose JSON passthrough, same `Type.Unknown` pattern as the dsl tools). `procedural-skill` stays gated behind task-engine mode and is rejected by the schema — the agent runtime has no vocational signal, so accepting it here would ungate the feature in every deployment.
- Non-interactive page types reject widget fields (checked by definedness) with a `widget-requires-interactive` tool error before any mutation; a malformed `widgetOutline` (non-object) is rejected with `invalid-widget-outline`.
- For interactive pages the fields are carried into the constructed `SceneOutline`. When only one of the two is given, the handler mirrors the generator's own defaults (`widgetType: 'simulation'` / `widgetOutline: { concept: title }`) so half-specified calls still generate instead of failing inside `generateWidgetContent`.
- Tool description updated so the model knows the widget kinds.

## Tests

Six new cases in `tests/agent-runtime/generation-tools.test.ts` (15/15 pass):

- interactive + `widgetType: 'diagram'` + widgetOutline → persisted scene carries `content.widgetType === 'diagram'` and the diagram prompt receives the widgetOutline (pinned to a non-default `diagramType`).
- interactive without widget fields → unchanged simulation fallback.
- `type: 'slide'` + widgetType → tool error, no scene written.
- malformed widgetOutline (null / string) → `invalid-widget-outline`, nothing written.
- single-field defaults: widgetType-only and widgetOutline-only calls.
- schema validation through TypeBox `Value.Check`, including the procedural-skill gate.

## Validation

- `tsc --noEmit`, `eslint`, `prettier --check`: clean.
- Full vitest run: only pre-existing environment flakes in `chat-storage` / `quiz runtime` (unrelated to this change's dependency graph, fail the same on a clean checkout).


wyuc:
**Cross-review summary**

Ran two rounds of single-reviewer cross-review (DeepSeek harness) over the full branch diff.

- Round 1 (a764b68b): no P1/P2; 7 P3 findings around widgetOutline shape enforcement, guard truthiness, the procedural-skill/schema mismatch, and three test-effectiveness gaps.
- Fix commit a9a5097f closed 6 of 7 (procedural-skill added to the union, invalid-widget-outline validation, definedness guards, mindmap assertion, single-field default tests, Value.Check schema coverage).
- Accepted limitation: a widgetOutline-only call keeps the outline but forces widgetType 'simulation'; documented in the schema description.
- Round 2 re-review of the full diff: all prior findings verified closed, no new actionable findings.

CXuPercy (changes requested):
Verified locally against the PR head (worktree + full targeted test runs). The design of the widget params itself is solid — guards fire before any write, the half-specified defaults correctly mirror the generator's own fallbacks (I confirmed `generateWidgetContent` returns null without a `widgetOutline`, so the `{ concept: title }` default is load-bearing), and the new tests assert prompt-level behavior, not just shapes (16/16 pass locally). One blocking issue, plus a documentation ask.

## Blocking: `procedural-skill` is ungated here while every other path on main gates it

The evidence chain, each step verified against the actual refs:

1. **The first commit (a764b68b) excluded it, matching the PR description** ("procedural-skill stays gated behind task-engine mode"). The harden commit (a9a5097f) added `Type.Literal('procedural-skill')` to the union unconditionally — so the "schema/handler mismatch" was resolved in the ungate direction.
2. **main deliberately gates this widget in three places**, all converging on `resolveVocationalActive` (per-course `taskEngineMode` **and** the deployment flag `OPENMAIC_ENABLE_VOCATIONAL`):
   - `app/api/generate/scene-outlines-stream/route.ts` — `sanitizeNonTaskEngineOutline` forces procedural-skill back to a diagram in ordinary mode, with the comment "to protect ordinary MAIC generation";
   - `app/api/generate/scene-content/route.ts` — `applyOutlineFallbacks({ allowProceduralSkill: vocationalActive })`;
   - `lib/server/classroom-generation.ts` — same `allowProceduralSkill: vocationalActive`.
3. **The agent runtime has no vocational signal at all** (no `taskEngineMode`/`requirements` concept reaches `GenerationToolDeps`), and `generate_scene` is registered unconditionally in the standard course toolset — so the new literal is reachable in every session, including deployments where the vocational feature flag is off.
4. The pre-existing `allowProceduralSkill: true` at this call site (main, generation-tools.ts:329) is **not** a precedent for ungating: on main no agent tool could author a `widgetType` at all — `generate_scene` built its outline inline without widget fields, `outlineFromScene` is only used by `generate_actions`, and `patch_stage` explicitly blocklists `widgetType` (dsl-tools.ts:505). That `true` was never exercised in practice.

Net effect: after this PR, a model in an ordinary session (even with `OPENMAIC_ENABLE_VOCATIONAL` unset) can mint procedural-skill "practice" pages — exactly what the three gates above exist to prevent.

**Suggested fix**, cheapest first:
- (a) Drop the `procedural-skill` literal from the union (back to the first commit's semantics), remove/adjust the two procedural-skill tests and the tool-description mention. This also makes the code match the PR description again.
- (b) If the intent is for vocational agent sessions to use it, plumb a vocational/task-engine signal into the tool deps and gate the literal on it — a larger change since no such signal reaches the agent runtime today.

## Non-blocking: please update the PR description to match the final diff

Three spots read as stale relative to a9a5097f: it says "Three new cases" (the final diff adds 7), "12/12 pass" (the file now has 16 tests), and "procedural-skill stays gated" (with option (a) applied this becomes true again — otherwise please correct it).

Everything else — the `widget-requires-interactive` / `invalid-widget-outline` guards before mutation, the definedness fixes, the `Value.Check` schema coverage, and the loose-outline passthrough with graceful generator defaults — looks good to me.

wyuc:
Thanks for the careful check — you're right, and I've applied option (a) in f965da83:

- `procedural-skill` is back out of the `widgetType` union. Your evidence chain settled it: every main path gates it behind `taskEngineMode` + `OPENMAIC_ENABLE_VOCATIONAL`, and no vocational signal reaches the agent runtime, so the literal would have ungated the feature fleet-wide. The schema description now says it stays gated, and the Value.Check schema test pins the rejection so a future regression fails loudly. Ungating for vocational agent sessions should come back as its own change with a task-engine signal plumbed into the tool layer (your option b).
- The passthrough test is removed; the schema test now asserts `procedural-skill` → false.
- PR description updated to match the final diff (six new cases, 15/15).

tsc / eslint / prettier clean; targeted tests 15/15.

CXuPercy (approved):
Verified f965da83 locally: the literal is out of the union, the schema description now states procedural-skill stays gated, the passthrough test is removed, and the schema test pins the rejection (Value.Check → false) so a regression fails loudly. Targeted tests 15/15, CI green. Thanks for the clean turnaround — the gated-rejection test is a particularly nice touch. Option (b) as its own change with a task-engine signal sounds right.

CXuPercy:

PR #1261, opened 2026-08-27
fix(dsl): migrate away legacy rotate/height fields on line elements

### LINKED ISSUE #1260: Legacy line elements with a stray rotate field break every patch_stage on old classrooms
## Symptom

Editing a classroom created by a pre-1.0 build fails on the first `patch_stage`, even when the op targets an unrelated element. The error is a wall of union-schema noise:

```
patch_stage rejected at op 1: patch rejected: canvas/elements/11: must have required properties
height, content, defaultFontName, defaultColor; canvas/elements/11: must not have additional
properties; canvas/elements/11/type: must be equal to constant; ...
```

Import succeeds and the scene renders fine; only agent edits are rejected.

## Root cause

1. Documents written by older runtimes can carry `line` elements with a `rotate` field (and in principle `height`). The contract omits both from `PPTLineElement` (`Omit<PPTBaseElement, 'height' | 'rotate'>`), but pre-versioning writes were not schema-checked, so stray fields survived into storage and exports.
2. Since 1.0.0, `patch_stage` validates the whole canvas after every op against a closed schema (`validateSlideCanvas`, `additionalProperties: false`). The line variant does not list `rotate`, so one legacy line element fails all ten union branches. Because validation is whole-canvas, any op on that scene is rejected.
3. Import does not run this validation, which is why the classroom opens and renders fine and only fails later. The renderer never reads `rotate` on a line, so display is unaffected.

Two distinct populations are affected:

- **Documents written before version stamping existed.** They carry no `dslVersion` stamp, so the `DSL_MIGRATIONS` ladder lifts them on every load. A new ladder step can clean them.
- **Documents imported on 1.0.0.** The import pipeline writes legacy scenes through the document store, which stamps the current `DSL_VERSION` on save. The ladder never runs for these, so import-time sanitation is needed.

## Proposal

1. Append a `0.2.0 → 0.3.0` step to `DSL_MIGRATIONS` stripping `rotate` (defensively also `height`) from `type: 'line'` elements under `scenes[*].content.canvas.elements`. Pure transform, other scene kinds pass through. Bump `DSL_VERSION` and the `@openmaic/dsl` npm version per the enforced release rule (`0.10.x` → `0.11.0`).
2. Export the strip as a pure helper from `@openmaic/dsl` and call it in the import pipeline before `saveDocument`, covering the one path the ladder cannot reach.

Stripping is lossless: line geometry is fully determined by `left`/`top`/`width` plus `start`/`end`, and no 1.0.0 reader or writer uses `rotate` on lines.

Happy to pick this up.

wyuc:
Closes #1260.

## What happened

Classrooms written before schema enforcement can carry `line` elements with a stray `rotate` (and in principle `height`) field. The contract omits both from `PPTLineElement`, and since 1.0.0 `patch_stage` validates the whole scene canvas against the closed schema after every op, so one legacy line element fails all ten union branches and makes EVERY edit to its scene rejected. Import does not validate, so such classrooms open and render fine and only break on their first agent edit.

Two document populations are affected, and both are healed by the migration ladder alone:

1. Documents written before version stamping existed (no `dslVersion` field). The ladder lifts them on every aggregate read.
2. Documents stored by 1.0.0 (stamped `0.2.0` at save time, including previously imported classrooms). The new ladder step lifts them on the next read.

## What this PR does

- Adds `stripLegacyLineGeometry` to `@openmaic/dsl`: a pure, dependency-free transform that strips `rotate`/`height` from `type: 'line'` elements on every slide surface of every migratable envelope — scene canvases (`scenes[*].content.canvas.elements`), interactive whiteboard slides (`scenes[*].whiteboards[*].elements`), stage-level explainer boards (`stage.whiteboard[*].elements`), at document, Scene-row, and Stage-row roots. It never mutates, returns the input by identity when nothing needs stripping, shares untouched subtrees, and gates the canvas walk on the slide discriminant so non-slide scene kinds pass through untouched. The strip is lossless: line geometry is fully determined by `left`/`top`/`width` plus `start`/`end`, and no reader or writer uses `rotate` on lines.
- Appends the `0.2.0 → 0.3.0` ladder step wiring the helper, bumps `DSL_VERSION` to `0.3.0`, and bumps `@openmaic/dsl` to `0.11.0` per the enforced release rule.
- Hardens the two paths the bump touches:
  - `exportDatabase`'s legacy-only fallback now builds the document unstamped and runs `migrate`, so the exported stamp reflects an actually-run migration instead of being hand-assigned.
  - Server tools' incremental scene writes go through `putSceneBringingCurrent`, whose `not-current` fallback reloads the migrated aggregate and full-saves it with the scene spliced in — the server-side counterpart of the app autosave's catch-and-full-save, so the first agent edit on a stale-stamped course migrates it instead of erroring.

## Accepted limitation

`putSceneBringingCurrent`'s fallback reload → splice → save spans two transactions, so a writer committing another scene inside that window would be pruned by the snapshot save. Cross-session server writers are excluded by the per-stage lease; the remaining competitor is the browser autosave, which already replaces the whole document last-writer-wins (the pre-existing document-level concurrency model), and its data is periodic — the editor re-saves live state on the next tick. The window is one-shot per document: once the stamp is current, later writes take the incremental fast path. Documented on the wrapper.

## Testing

- 252 DSL tests: ladder invariants, every envelope/surface combination, discriminant gating, purity, identity no-ops, idempotency.
- New `document-writes` unit tests: fast path, splice-and-sort fallback, replace-by-id, vanished-document and `future` rethrows.
- `database-chat-cutover`: legacy-only export runs the ladder (stray `rotate` stripped, legacy rows untouched).
- Real pre-1.0 classroom export verified end to end: all dirty line elements cleaned, every other element deep-equal.
- `check-package-version-bumps` passes (`0.10.1 → 0.11.0`).

## Review

Five review rounds (independent review passes + targeted audits of the version-envelope state space). Every P1/P2 finding fixed; the one remaining item is the accepted limitation above.

### DIFF
--- a/lib/import/use-import-classroom.ts
+++ b/lib/import/use-import-classroom.ts
@@ -19,7 +19,6 @@ import { canonicalizeLegacyScene, mutateDocument, type AppDocument } from '@/lib
 import { isConcreteMediaAddress } from '@/lib/media/resolve-media-ref';
 import { isGeneratedMediaPlaceholder } from '@/lib/media/media-ref';
 import type JSZip from 'jszip';
-import { stripLegacyLineGeometry } from '@openmaic/dsl';
 import type { Slide } from '@openmaic/dsl';
 import type { Stage } from '@/lib/types/stage';
 
@@ -510,13 +509,6 @@ export function useImportClassroom(onSuccess?: (importedStageId: string) => void
           }),
         };
 
-        // Imported legacy scenes never pass through the DSL migration ladder:
-        // the store stamps the current version on save, so a document written
-        // here reads as current forever. Strip the stray legacy line fields
-        // (rotate/height) at write time instead -- the 1.0.0 closed canvas
-        // schema would otherwise reject every patch_stage on this classroom.
-        const sanitizedDocument = stripLegacyLineGeometry(document) as AppDocument;
-
         // The document is the commit point: one aggregate write under its
         // per-stage lock. Wholesale replacement: the imported aggregate
         // overwrites the whole document, so eager conversion of whatever
@@ -524,7 +516,7 @@ export function useImportClassroom(onSuccess?: (importedStageId: string) => void
         // replaces.
         await mutateDocument(
           newStageId,
-          async (_existing, store) => store.saveDocument(sanitizedDocument),
+          async (_existing, store) => store.saveDocument(document),
           {},
           { mode: 'replace' },
         );
--- a/lib/import/use-import-classroom.ts
+++ b/lib/import/use-import-classroom.ts
@@ -19,6 +19,7 @@ import { canonicalizeLegacyScene, mutateDocument, type AppDocument } from '@/lib
 import { isConcreteMediaAddress } from '@/lib/media/resolve-media-ref';
 import { isGeneratedMediaPlaceholder } from '@/lib/media/media-ref';
 import type JSZip from 'jszip';
+import { stripLegacyLineGeometry } from '@openmaic/dsl';
 import type { Slide } from '@openmaic/dsl';
 import type { Stage } from '@/lib/types/stage';
 
@@ -509,6 +510,13 @@ export function useImportClassroom(onSuccess?: (importedStageId: string) => void
           }),
         };
 
+        // Imported legacy scenes never pass through the DSL migration ladder:
+        // the store stamps the current version on save, so a document written
+        // here reads as current forever. Strip the stray legacy line fields
+        // (rotate/height) at write time instead -- the 1.0.0 closed canvas
+        // schema would otherwise reject every patch_stage on this classroom.
+        const sanitizedDocument = stripLegacyLineGeometry(document) as AppDocument;
+
         // The document is the commit point: one aggregate write under its
         // per-stage lock. Wholesale replacement: the imported aggregate
         // overwrites the whole document, so eager conversion of whatever
@@ -516,7 +524,7 @@ export function useImportClassroom(onSuccess?: (importedStageId: string) => void
         // replaces.
         await mutateDocument(
           newStageId,
-          async (_existing, store) => store.saveDocument(document),
+          async (_existing, store) => store.saveDocument(sanitizedDocument),
           {},
           { mode: 'replace' },
         );
--- a/lib/server/agent-runtime/course-edit/tools.ts
+++ b/lib/server/agent-runtime/course-edit/tools.ts
@@ -5,6 +5,7 @@ import type { AgentTool } from '@earendil-works/pi-agent-core';
 import type { AppDocumentOutline } from '@/lib/document-store/persistence-types';
 import type { Scene } from '@/lib/types/stage';
 import { matchOutlineEntries } from '../course-outline-union';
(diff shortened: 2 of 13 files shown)


wyuc:
Review process: five rounds of independent review passes over the full branch diff, each followed by targeted fixes and a fresh audit of the changed surfaces.

- Round 1: no findings from the automated pass; a targeted contract audit surfaced two scope gaps (whiteboard slides uncovered; a redundant import-site call whose rationale comment misdescribed the store's migrate-on-save). Fixed by covering whiteboards and dropping the redundant call.
- Round 2: the terminal audit surfaced two more contract-level findings (stage-level explainer boards uncovered; the traversal assumed the document envelope while the migration contract is envelope-agnostic). Fixed by walking every line-element surface of every migratable envelope.
- Round 3: surfaced two more (the lock-free backup fallback hand-assigned the current stamp without running the ladder; the canvas walk lacked a slide-discriminant gate). Fixed: exports earn their stamp via `migrate`; the walk gates on `content.type`.
- Round 4: surfaced the most consequential one — incremental scene writes reject stale-stamped documents, so the first agent edit on an older stored course would have failed. Fixed with `putSceneBringingCurrent`, whose fallback full-saves the migrated aggregate (mirroring the app autosave's existing catch-and-full-save).
- Round 5: audited the new fallback and surfaced one concurrency window (non-atomic reload → splice → save). Accepted with analysis and documented on the wrapper: cross-session server writers are lease-excluded, the remaining competitor is the periodic autosave whose model is already last-writer-wins, and the window is one-shot per document.

Final state: all P1/P2 findings fixed or explicitly accepted, full test suites green, real legacy classroom exports verified end to end.

cosarah (approved):
Approved.

Verified at the reviewed commit: dsl tests (252 passed), agent-runtime document-writes + database-chat-cutover (45 passed), version-bump check, dsl typecheck/build, root `tsc --noEmit`, and lint (0 errors) are all green. All server agent-runtime `putScene` call sites are wrapped; the fallback fires correctly on the owner-bound PG store's typed `not-current` error.

On the brief's three questions, I agree with the implemented design: the expanded scope is necessary (whole-canvas validation means any dirty surface keeps `patch_stage` rejected), the one-shot reload → splice → save window is acceptable under the existing last-writer-wins model and per-stage lease, and keeping `stripLegacyLineGeometry` internal preserves the ladder as sole cleanup owner.

Three non-blocking observations:

- Legacy-only export now runs the ladder instead of hand-stamping, so a malformed version stamp in legacy Dexie data makes the export throw where it previously succeeded. Intentional per the fail-loud policy, but a user-visible behavior change.
- If `loadDocument` itself throws inside the fallback (other than returning `null`), the original `not-current` error is replaced by the load error. Edge case; the vanished-document case is correctly rethrown.
- The fallback save validates the whole aggregate, so schema violations beyond line `rotate`/`height` would still fail it — within the declared non-goal of general document repair.

Forward-looking note: the HTTP store reconstructs only the `future` variant as a typed error, so the stale-stamp fallback would not fire over HTTP. No current code path is affected.

PR #1264, opened 2026-08-28
fix(instrumentation): remove process.once Edge warnings

### LINKED ISSUE #1262: [Bug]: running pnpm dev triggers two warnings at instrumentation.ts.
### Bug Description

pnpm dev → two warnings at instrumentation.ts.

### Steps to Reproduce

Run pnpm dev

### Expected Behavior

pnpm dev and pnpm build succeed. Node.js shutdown logic is excluded from the Edge bundle and runs only in the Node runtime.

### Actual Behavior

pnpm dev shows two warnings: "A Node.js API is used (process.once)" from instrumentation.ts:98. pnpm build is unaffected.

### Deployment Method

Local development (pnpm dev)

### OpenMAIC Version

latest

### Affected Area

Deployment / infrastructure

### Browser

Not applicable

### Operating System

Not applicable

### Relevant Logs

```shell
⚠ ./instrumentation.ts:98:3
Ecmascript file had an error
   96 |
   97 |   process.once('SIGTERM', () => void shutdown());
>  98 |   process.once('SIGINT', () => void shutdown());
      |   ^^^^^^^^^^^^
   99 | }

A Node.js API is used (process.once at line: 98) which is not supported in the Edge Runtime.
Learn more: https://nextjs.org/docs/api-reference/edge-runtime
```

### Screenshots or Recordings

_No response_

puxiao:
## Summary

This removes two build-time Edge warnings that occur during pnpm dev, while preserving the existing Node-side shutdown behavior.

## Related Issues

Fixes #1262

Changes

- Add `lib/server/shutdown-signals.ts` with `registerShutdownSignals(shutdown)`.
- In `instrumentation.ts`, replace the inline `process.once` calls with a dynamic import and call:
  `await import('@/lib/server/shutdown-signals')` → `registerShutdownSignals(shutdown)`.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Run pnpm dev — before the fix, this triggered two warnings at instrumentation.ts
2. After the fix, pnpm dev runs without the warnings, and pnpm build remains unaffected.


LING-6150 (changes requested):
Thanks for the focused patch. The implementation itself looks appropriately scoped, and I don’t have any code-level blockers.
Before merging, please correct the PR title/body and #1262 so they match the observed behavior.
With Next.js 16.1.2, the inline process.once calls produce two Edge Runtime warnings, but next build still completes successfully with exit code 0. This patch removes those warnings; it does not fix a compile failure.
I also reproduced the warnings with a Node-runtime proxy.ts and an empty Edge middleware manifest, so having an explicit Edge runtime entry is not a required condition.
Please update:
- The PR title and summary to describe removing Edge build warnings rather than fixing a compile failure.
- The verification section to state that the build succeeds before and after the patch.
- #1262’s title/body and reproduction conditions accordingly.
- The malformed Closes #1262" reference.
An accurate framing would be: this removes two build-time Edge warnings while preserving the existing Node-side shutdown behavior.  In addition, please also modify the title as "fix(instrumentation): remove process.once Edge warnings" . 
No code changes are requested.

LING-6150:
Thanks for addressing the review feedback so promptly. I’ve confirmed that this patch removes the two non-blocking Edge warnings while preserving the existing Node-side behavior.
After considering the maintenance tradeoff, we don’t plan to take a warning-only cleanup with no functional or user-facing impact at this time. The build already succeeds without this change, so I’m going to close the PR.
We can revisit this if the warning becomes blocking or needs to be addressed as part of a broader Next.js upgrade. Thanks again for the contribution.

PR #1267, opened 2026-08-28
feat(agent): make generate_video asynchronous with a placeholder ref

### LINKED ISSUE #1266: generate_video is synchronous: blocks the agent turn for minutes and its 15-min budget is pre-empted by the 10-min tool cap
## Problem

`generate_video` (`lib/server/agent-runtime/generate-video.ts`) runs the whole provider submit → poll → download → persist cycle inside the tool call. Video providers routinely take minutes to render, so one call can stall the agent's turn for a long time, and it can be killed before its own budget is reached.

## Details

**1. Budget mismatch: the 15-minute internal budget never survives the 10-minute outer cap.**

Every agent tool is wrapped by `withAgentToolTimeout` (`lib/agent/runtime/tool-timeout.ts`, default budget 10 min). `AGENT_TOOL_TIMEOUT_OVERRIDES` raises the budget to 15 min for `generate_scene`, `generate_actions`, and `extract_material`, but **not for `generate_video`**, whose own `GENERATE_VIDEO_TIMEOUT_MS` is 15 min ("the longest provider poll budget is 15 minutes"). So the tool is aborted at 10 minutes and its internal timeout is dead code in practice.

The error reporting also degrades on this path: the outer signal's reason is an `AgentToolTimeoutError`, not the `TimeoutError` DOMException that `isTimeout()` checks for, so the catch block falls through to the caller-abort branch and the agent ends up with the generic "exceeded its execution budget" message with no provider/model context.

**2. The agent loop is blocked for the whole duration.**

The pi agent loop awaits tool execution serially, so while a video renders (minutes to tens of minutes), the rest of the turn (page edits, follow-up messages) is stuck behind it and the user stares at a spinner. The tool emits no progress via `onUpdate`. The session itself survives (the lease keeps heartbeating, and the timeout rejection becomes a structured tool error the agent can react to), but the turn is unproductive for the entire wait.

**3. Aborted calls leave zombie polling behind.**

When the outer timeout fires, `awaitWithSignal` rejects immediately, but the provider-side poll loop in `runPolledTask` (`lib/media/polled-task.ts`) has no signal handling (plain `setTimeout` delays) and keeps polling until its own budget runs out, burning provider quota after the result is already discarded.

## Existing precedent for async media

The codebase already has the full pattern for asynchronous media: the classic generation flow uses `image:pending` / `video:pending` placeholders plus the media-generation task store (`lib/store/media-generation.ts`, pending/generating/done/failed) and render-time resolution (`lib/media/media-task-resolution.ts`), so pages ship first and media fills in later. The agent side already understands these placeholders (`generation-tools.ts` rejects unresolved placeholder srcs).

## Options

- **A. Align the budgets (stopgap).** Add `generate_video` to `AGENT_TOOL_TIMEOUT_OVERRIDES`, fix `isTimeout` so the timeout error carries provider/duration context, and emit progress via `onUpdate`. Small change, but the turn still blocks.
- **B. Make the tool truly asynchronous.** Submit the provider task and return a placeholder ref immediately so the agent can `patch_stage` the page and keep working; move polling to a background task. On completion either resolve the placeholder through the existing client media-task mechanism (reusing the classic flow) or inject a steering message telling the agent the video is ready with its src. Requires a server-side task registry, placeholder-to-element binding, and a completion path.
- **C. Hybrid: A now, B as a follow-up.**

wyuc:
## Summary

Fixes #1266.

`generate_video` awaited the whole provider submit → poll → download → persist cycle inside the tool call, blocking the agent turn for minutes — and in practice capping it at the 10-minute tool budget, below its own 15-minute internal budget. The tool is now asynchronous, aligned with the classic flow's placeholder pattern.

## How it works

1. The tool validates as before, mints a `gen_vid_<id>` placeholder (same scheme as the outline flow), and returns immediately. The agent `patch_stage`s the ref onto a video element's `mediaRef`/`src` and keeps working; the element renders the existing skeleton while pending.
2. A detached background job runs the provider cycle on its own 15-minute timeout, deliberately decoupled from the caller's abort — a cancelled chat must not silently orphan a billable provider job (the classic orchestrator accepts the same caveat).
3. On success the job (a) patches the persisted document — every slide video element still referencing the placeholder gets the concrete `/api/classroom-media/...` src, via the same `runStageMutation` discipline as the generation tools, skipped silently if the element changed meanwhile — and (b) appends a `media_ready` lifecycle event to the session's durable log via the session-level control channel (`appendControlEvent`, valid post-run; the runner's `appendRunEvent` is lease-guarded to a live run). On failure it emits `media_ready` with `status: 'failed'` and a structured error code.
4. The workbench folds `media_ready` into the media generation store, live and on replay, so the skeleton resolves instantly on any connected or re-attached client.

## Tests

- Rewrote `tests/agent-runtime/generate-video.test.ts` on deferred-promise provider mocks: immediate return shape, background completion patching the document, silent skip when unreferenced, failed/timeout events, caller-abort-doesn't-kill-job, provider-neutral payloads.
- New `tests/agent-runtime/pending-media.test.ts` (registry) and `tests/workbench/media-lifecycle.test.ts` (event fold + a regression pin that a placeholder-only video element renders the skeleton with no store task).
- `tsc` / `eslint` / `prettier` clean; `tests/agent-runtime` + `tests/media` + `tests/workbench`: 2256 passed, 0 failed.

## Accepted v1 limitations

- Pending tasks live in a process-local registry; a server restart orphans in-flight jobs (placeholder keeps its skeleton). A PG-backed table + recovery sweep is a possible follow-up.
- The background patch scans slide-canvas video elements only.
- `generate_image` stays synchronous; the same pattern can be applied later.


wyuc:
**Cross-review summary**

Five single-reviewer cross-review rounds (DeepSeek harness) over the full branch diff:

- Round 1 found a P1 (the completion patch wrote through the run-lease-fenced store, so every post-run patch threw and surfaced as failed) and a P2 (stale whole-scene rewrite), plus 3 P3s — all fixed in 99337b79 via a dedicated lease-free owner-bound store and per-scene fresh reads.
- Round 2 verified the closures; 2 further P3s fixed in 935ba4e0, 2 accepted (defensive crash-guard path untestable by design; bounded N+1 reads).
- Round 3 (full audit) found no P1/P2; 15c8c4d4 isolated the patch onto its own 60s budget so a persisted asset is never rebranded failed, and stopped the patch from clobbering a user's own concrete src.
- Two delta rounds closed a non-discriminating test and widened the src guard for regeneration (previous generated src, empty src, absolute-form URLs), hardened in 8e4a110c + 60a84532.

Accepted v1 limitations: process-local pending registry (restart orphans in-flight jobs), failed-state fold requires an attached chat stream, and the sub-millisecond read-write window between two concurrent jobs on one scene (client fold renders either way).

CXuPercy (approved):
Verified locally against the PR head (worktree + the four touched test files, 37/37). This is a well-executed architectural change — I traced every load-bearing claim through the code and they all hold:

**The detached write path is correctly unfenced.** `mediaJobStore` differs from the run's `ownerScopedStore` by exactly the `assertActiveLease` call (runner.ts), keeping owner binding and the stage-mutation fence; `runStageMutation` is a module-level AsyncLocalStorage, so the fence works minutes after the run ended. The lease-loss throw the comments warn about is real, and this is the right escape.

**The completion event genuinely reaches browsers, live and on replay.** `appendControlEvent` appends to the same durable events table the SSE route replays (valid post-run; the null-on-missing-session path is handled), `insertEvent` fires the transactional NOTIFY that wakes attached streams, and the route forwards every durable event with `event: ${event.type}`. The client subscription list derives from `LIFECYCLE` so `media_ready` is auto-included — and the regression test pinning it against the EventSource named-event drop is a smart touch given that failure mode has bitten this codebase before.

**The patch discipline holds up.** Fresh re-read per candidate scene (with tests simulating both a concurrent edit and a concurrent element removal), skip-silently when unreferenced, the `isReplaceableSrc` guard scoped to the stage's own media root so a user's cross-stage pick survives while a stale previous generation is replaced, and the patch on its own 60s budget so an exhausted job signal can't rebrand a persisted, downloadable asset as failed. The registry's first-settle-wins idempotency, the synchronous crash guard on the detached promise, and the never-crashing emit wrapper are all tested.

Two non-blocking notes:

1. The process-local registry means a restart orphans in-flight jobs — a placeholder keeps its skeleton forever and a billable provider job finishes with nothing landing. You've documented this as a v1 limitation; I'd just encourage keeping the PG-backed table + recovery sweep near the top of the follow-up list rather than letting it settle, since it's the one gap users will actually hit on any redeploy.
2. Failures are invisible to the model: `media_ready` goes to the client, so the agent can't apologize or retry in-conversation. Inherent to the async design and fine for v1 — worth an eventual follow-up note in the skill/prompt layer if agent-side failure awareness ever becomes wanted.

The tool description and `DSL_TOOLS_PROMPT` guidance both match the new semantics, and the abort trade-off (a cancelled chat still lands its billable job) is deliberate, documented, and pinned by a test. Nice work.

CXuPercy:

PR #1269, opened 2026-08-28
fix(agent): bound import_pptx parse with size cap and timeout

### LINKED ISSUE #1268: import_pptx can hang forever on large PowerPoint files
## Problem

`import_pptx` (`lib/server/agent-runtime/import-pptx.ts`) can hang the Pro agent turn indefinitely while parsing a large PowerPoint. The worker that runs `@openmaic/importer` has no deadline, does not listen to the tool AbortSignal, and is only bounded after a successful parse (`MAX_IMPORT_SLIDES = 80`). A deck that is merely large, or whose embedded media explode in memory as data URLs, never reaches that cap.

## Details

**1. `parsePptxIsolated` waits forever.**

The helper creates a `worker_threads` Worker and `await`s `message` / `error` with no timer. There is no `AbortSignal` on that Promise, so session cancel, lease loss, and `withAgentToolTimeout` (default 10 min in `lib/agent/runtime/tool-timeout.ts`) cannot `worker.terminate()`. The JS race in `withAgentToolTimeout` may reject the tool call, but the worker keeps parsing and holding memory. `throwIfAborted` in the tool body only runs before/after `parsePptx`, not during it.

**2. Size is checked too late.**

`MAX_IMPORT_SLIDES` slices the parsed slide array after `importPptx` returns. A tens-of-MB deck with many embedded images is fully parsed first. The worker also converts each asset to an in-memory `data:` URL before the parent rewrites URLs to object storage.

**3. Synthesized scenario.**

A layout-preserving import of a large lecture `.pptx` (tens of MB, image-heavy) starts `import_pptx`, emits a "parsing …" trace, and never returns `tool_execution_end`. The stage stays empty. The runner lease keeps heartbeating. Smaller decks (a few MB, ~20 slides) finish in about a second.

## Proposed fix

Keep this in the durable tool, not a new pipeline:

1. **Byte cap before parse.** Add `MAX_IMPORT_BYTES` next to `MAX_IMPORT_SLIDES`. If `record.bytes` exceeds it, return a structured `isError` tool result (ask the user to compress or split the file) and do not start the worker. Suggested cap: 8 MiB, above known-good few-MB decks and well below tens-of-MB hang cases.
2. **Bound `parsePptxIsolated`.** Honor the tool AbortSignal and a parse timeout (60–90 s). On either, `worker.terminate()` and fail the parse so the existing `Failed to parse the PowerPoint: …` path runs. Do not leave a zombie worker.
3. **Tests** in `tests/agent-runtime/import-pptx.test.ts`: oversized material is rejected without calling `parsePptx`; a hung `parsePptxIsolated` terminates the worker on timeout/abort.

Out of scope: rewriting worker media to skip data URLs; changing `import-pptx-worker.mjs` beyond what parent-side terminate already covers.

## Files

- `lib/server/agent-runtime/import-pptx.ts`
- `tests/agent-runtime/import-pptx.test.ts`

wyuc:
Fixes #1268.

`import_pptx` could hang the Pro turn while parsing a large PowerPoint: the isolated worker had no deadline, ignored AbortSignal, and `MAX_IMPORT_SLIDES` only applied after a successful parse.

This PR:
- Rejects decks larger than 8 MiB before parse (`MAX_IMPORT_BYTES`)
- Terminates the parse worker on abort or a 90s timeout (`PARSE_PPTX_TIMEOUT_MS`)
- Keeps the existing parse-failed tool error path

Tests cover oversized reject-without-parse, hung-worker timeout, and abort.

### DIFF
--- a/lib/server/agent-runtime/import-pptx.ts
+++ b/lib/server/agent-runtime/import-pptx.ts
@@ -39,6 +39,8 @@ export { isPptxMaterial, PPTX_MIME } from './pptx-mime';
 export const IMPORT_PPTX_TOOL_NAME = 'import_pptx';
 export const IMPORT_PPTX_REQUIREMENT_PREFIX = 'import_pptx:';
 export const MAX_IMPORT_SLIDES = 80;
+export const MAX_IMPORT_BYTES = 8 * 1024 * 1024;
+export const PARSE_PPTX_TIMEOUT_MS = 90_000;
 
 /** Next-step hint after import. Inspection and repair come before TTS. */
 export const AFTER_IMPORT_NEXT_STEP =
@@ -58,12 +60,25 @@ const ImportParams = Type.Object({
   ),
 });
 
+export interface ParsePptxWorker {
+  once(event: string, listener: (...args: unknown[]) => void): unknown;
+  terminate(): unknown;
+}
+
+export interface ParsePptxOptions {
+  upload?: OssUpload;
+  signal?: AbortSignal;
+  /** Parse deadline; tests may inject a shorter value. */
+  timeoutMs?: number;
+  Worker?: new (filename: string | URL, options?: { workerData?: unknown }) => ParsePptxWorker;
+}
+
 export interface ImportPptxToolDeps extends CourseToolDeps {
   getMaterial?: (sessionId: string, materialId: string) => Promise<AgentSessionMaterial | null>;
   readMaterialBytes?: (
     record: AgentSessionMaterial,
   ) => Promise<{ bytes: Buffer; mime: string } | null>;
-  parsePptx?: (buffer: ArrayBuffer, options?: { upload?: OssUpload }) => Promise<Slide[]>;
+  parsePptx?: (buffer: ArrayBuffer, options?: ParsePptxOptions) => Promise<Slide[]>;
   uploadImportedMedia?: OssUpload;
 }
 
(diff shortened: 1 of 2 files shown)


CXuPercy (approved):
Verified locally against the PR head (worktree + targeted test runs, 31/31 pass). The implementation is solid — details below — with one calibration question worth an answer before or right after merge.

**Implementation verified:**
- The settlement pattern is correct: all four paths (message / worker error / abort / timeout) funnel through `finish()` with the `settled` guard into `cleanup()`, so there is no double-settlement, no leaked timer, and no leaked abort listener. I also checked the declaration-order hazard (`cleanup` referencing the later-declared `onAbort`) — unreachable in the actual call order, and the pre-registration `signal?.aborted` re-check closes the race window.
- Abort semantics line up: entry fast-fail before the buffer copy, per-call `signal ?? deps.abortSignal` at the tool layer, and the `'aborted'` message rethrows into the existing session-cancel path.
- Blast radius is contained: `parsePptxIsolated`'s only production caller is this tool (`parsePptxBuffer` is an exported wrapper with no other call sites), so the 90s deadline changes nothing else, and it sits well inside the 10-min tool timeout.
- Worker-side rationale confirmed by reading `import-pptx-worker.mjs`: every asset becomes a base64 data URL (~1.33× inflation) on a thread sharing the process heap, so a byte cap — not a timeout — is the real OOM defense. The two mechanisms are complementary and both belong here.

**One calibration question (main comment):**

`MAX_IMPORT_BYTES = 8 MiB` vs the pptx upload limit of 50 MiB. pptx is accepted by the workbench upload policy and is not in `MEDIA_MIME_TYPES`, so it gets `DOCUMENT_UPLOAD_LIMIT = min(maxDocumentBytes, maxUploadBytes)` — 50 MiB by default. Net effect: a user can upload a 12 MiB deck successfully and then `import_pptx` refuses it forever, with the failure surfacing in the conversation rather than at upload time. The direction of the cap is right; the calibration is the question. Either raising the import cap to 16–32 MiB per your memory math, or tightening the pptx upload limit by MIME (earlier failure, clearer message) would resolve it. At minimum the "pptx for import must be < 8 MiB" bound should be visible somewhere user-facing.

**Two nits (non-blocking):**

1. The size check runs before the `importRequirementFor` idempotency lookup, so a deck that was imported before this cap existed (> 8 MiB) now gets `too_large` on retry instead of the already-imported report. Moving the check after the digest, or noting in the message that previously imported pages are unaffected, would smooth that edge.
2. The worker `exit` event isn't listened to, so a hard crash (e.g. OOM kill) waits out the full 90s before surfacing — bounded and acceptable, but worth knowing. Also, the `timeoutMs > 0` escape hatch (0 disables the timer) deserves a one-line comment since nothing documents it.

Thanks — the settled/cleanup shape and the injected-worker tests are exactly how I'd want this bounded.

CXuPercy:

PR #1273, opened 2026-08-28
fix(workbench): persist Pro conversation titles

### LINKED ISSUE #1263: [Bug]: Pro conversation rename UI calls a missing PATCH endpoint and cannot persist titles
### Bug Description

The Pro workbench exposes manual conversation rename actions in both the conversation rail and pane header. The client sends `PATCH /api/agent/sessions/:id` with `{ "title": ... }`, but the matching route only implements `GET`.

On the current local `main`, the PATCH request returns HTTP 405, so a manual conversation title cannot be persisted. The storage model and PostgreSQL schema also do not currently contain a session title field.

### Steps to Reproduce

1. Check out commit `aa2bfb3c1d406c47100c6744d90e788abdf1f6d5` and start the app with `pnpm dev`.
2. Send:
   ```sh
   curl -i -X PATCH \
     -H 'content-type: application/json' \
     --data '{"title":"Renamed conversation"}' \
     http://127.0.0.1:<port>/api/agent/sessions/probe-session
   ```
3. Observe `HTTP/1.1 405 Method Not Allowed`.

The same request is issued by `renameWorkbenchSession()` when a user renames a Pro conversation from the UI.

### Expected Behavior

- A manual conversation title is stored for the owned session.
- Session list/detail responses return the stored title.
- The title survives reload.
- Sending `null` clears the override and restores the existing first-prompt fallback.

### Actual Behavior

- The client issues a PATCH request, but `app/api/agent/sessions/[id]/route.ts` exports only `GET`.
- The request returns 405.
- `AgentSessionMeta` and the `agent_sessions` schema have no title field to persist or return.

### Deployment Method

Local development (pnpm dev)

### OpenMAIC Version

v1.0.0 / `aa2bfb3c1d406c47100c6744d90e788abdf1f6d5`

### Affected Area

Storage / persistence

### Browser

Not applicable (route-level reproduction; the UI caller is browser-independent)

### Operating System

macOS 26.6.2

### Relevant Logs

```shell
PATCH /api/agent/sessions/probe-session
HTTP/1.1 405 Method Not Allowed
```

### Additional Context

Relevant code paths:

- `lib/workbench/session-store.ts`: `renameWorkbenchSession()` sends the PATCH request.
- `app/api/agent/sessions/[id]/route.ts`: only `GET` is implemented.
- `packages/@openmaic/storage/src/agent-session/types.ts`: `AgentSessionMeta` has no title.
- `packages/@openmaic/storage/src/agent-session/pg.ts`: `agent_sessions` has no title column.

Suggested acceptance boundary: add owner-scoped persistence for an optional manual title, expose it from list/detail reads, and cover rename, clear, reload, and owner isolation with tests.

This report is verified against the latest local `main`. The access-gated test deployment was not used as evidence. Automatic title generation is intentionally out of scope and can be discussed separately.

YizukiAme:
## Summary

Persist manual Pro conversation titles end to end so a rename survives reloads and cannot be rolled back by later owner-session status updates.

## Related Issues

Fixes #1263

## Changes

- Add a nullable agent_sessions.title column with an idempotent upgrade path and expose an additive AgentSessionTitleStore capability without changing AgentSessionStore.
- Add the missing owner-scoped PATCH /api/agent/sessions/:id route with strict validation, trim/120-character normalization, blank-to-null clearing, and uniform not-found behavior.
- Update the OwnerSessionClient authoritative snapshot during optimistic rename, server settle, and rollback; request one no-loading reconciliation after a changed successful write.
- Fence session-detail title bootstraps with a per-session revision so an older response cannot overwrite a rename or explicit clear made while it was in flight.
- Add real PGlite migration/isolation coverage, route boundary tests, and a shell regression test for the late status-event rollback race.
- Bump @openmaic/storage from 0.26.0 to 0.26.1.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Rename a Pro conversation from the rail or attached header and reload the workspace.
2. Clear the title and verify the original prompt-derived fallback returns.
3. Deliver a later owner session_status update and verify it does not restore the previous title.
4. Delay the initial session-detail response, rename before it arrives, and compare the rail, attached header, and stored title after release.

- Previous-head full verification: Node 22 package/dependency gates, Prettier, root TypeScript, i18n, ESLint, production build, root Vitest (7051 passed, 26 skipped), and package suites all passed.
- Current-head focused verification: all 77 workbench files passed (990 tests); root TypeScript, changed-file ESLint, Prettier, and diff checks passed.
- Manual browser verification against local PostgreSQL 16: rename + reload passed; clear + reload restored the prompt fallback.
- Deterministic delayed-response verification passed: after an older session-detail response carrying `Race Start` was captured, the title was changed to `Race New`; releasing the old response left the rail, attached header, and database on `Race New`. The same check passed for an explicit clear, preserving the prompt fallback instead of restoring the stale title.

### Evidence

- Local verification results are summarized above; all four GitHub checks passed for the latest head.
- [x] Manually tested locally, including delayed session-detail responses during rename and clear

## Checklist

### DIFF
--- a/components/workbench/workspace/WorkspaceShell.tsx
+++ b/components/workbench/workspace/WorkspaceShell.tsx
@@ -104,7 +104,7 @@ import {
 import type { WorkbenchCourseSummary } from '@/lib/workbench/panel-context';
 import { useWorkbenchStore, type WorkbenchMaterial } from '@/lib/workbench/session-store';
 import { renameWorkbenchSession } from '@/lib/workbench/session-store';
-import { commitSessionRename } from '@/lib/workbench/session-title';
+import { commitSessionRename, createSessionRenameQueue } from '@/lib/workbench/session-title';
 import type { ElementRef } from '@/lib/workbench/element-refs';
 import type { CourseRef } from '@/lib/workbench/course-refs';
 import { useStageFreshnessSync, useWorkbenchStream } from '@/lib/workbench/use-workbench-session';
@@ -146,6 +146,16 @@ async function fetchSessions(): Promise<ProHomeSessionItem[]> {
   }
 }
 
+function syncAttachedSessionTitle(
+  sessionId: string,
+  title: string | null,
+  forceRevision = false,
+): void {
+  const store = useWorkbenchStore.getState();
+  if (store.sessionId !== sessionId || (!forceRevision && store.sessionTitle === title)) return;
+  store.setSessionTitle(title);
+}
+
 /**
  * Route seam: capture the deep link exactly once. Native History API writes
  * update Next's search-param readers, while the controller below keeps its own
@@ -166,14 +176,10 @@ function WorkspaceShellController({ initialPanes }: { readonly initialPanes: Wor
   const courses = useHomeDiscovery({ mode: 'discover-only' });
   const [sessions, setSessions] = useState<ProHomeSessionItem[]>(EMPTY_SESSIONS);
   /**
-   * The latest list, readable from a callback without making that callback a
-   * new function on every poll (the rail would then see a changed prop every
-   * few seconds for no reason). Synced in an effect, never during render.
+   * The latest list, readable from queued callbacks without waiting for React
+   * to commit another render. The owner publisher updates it before setState.
    */
   const sessionsRef = useRef(sessions);
-  useEffect(() => {
-    sessionsRef.current = sessions;
-  }, [sessions]);
   const [sessionState, setSessionState] = useState<HomeDiscoveryState>('loading');
   const [composerReset, setComposerReset] = useState(0);
   /**
@@ -187,6 +193,7 @@ function WorkspaceShellController({ initialPanes }: { readonly initialPanes: Wor
    */
   const [newConversationRequested, setNewConversationRequested] = useState(false);
   const ownerSessionClient = useRef<OwnerSessionClient | null>(null);
+  const [sessionRenameQueue] = useState(createSessionRenameQueue);
 
   const rootRef = useRef<HTMLDivElement>(null);
   const railWidth = useRailWidth(rootRef);
@@ -331,11 +338,26 @@ function WorkspaceShellController({ initialPanes }: { readonly initialPanes: Wor
     ownerSessionClient.current?.requestFullFetch(showLoading);
   }, []);
 
+  const publishOwnerSessions = useCallback(
+    (next: readonly ProHomeSessionItem[], source?: 'incremental' | 'snapshot') => {
+      const snapshot = [...next];
+      sessionsRef.current = snapshot;
+      setSessions(snapshot);
+      const attachedId = useWorkbenchStore.getState().sessionId;
+      const attached = snapshot.find((session) => session.id === attachedId);
+      if (attached) {
+        syncAttachedSessionTitle(attached.id, attached.title ?? null, source === 'snapshot');
+      }
+    },
+    [],
+  );
+
   useEffect(() => {
     const client = new OwnerSessionClient({
       fetchSessions,
       createEventSource: (url) => new EventSource(url),
-      onSessions: (next) => setSessions([...next]),
+      onSessions: publishOwnerSessions,
+      onSessionTitle: (sessionId, title) => syncAttachedSessionTitle(sessionId, title, true),
       onState: setSessionState,
     });
     ownerSessionClient.current = client;
@@ -344,7 +366,7 @@ function WorkspaceShellController({ initialPanes }: { readonly initialPanes: Wor
       ownerSessionClient.current = null;
       client.stop();
     };
-  }, []);
+  }, [publishOwnerSessions]);
 
   useEffect(() => {
     if (
@@ -529,12 +551,10 @@ function WorkspaceShellController({ initialPanes }: { readonly initialPanes: Wor
    * which reads its own copy out of the session store.
    */
   const applySessionTitle = useCallback((sessionId: string, title: string | null) => {
-    ownerSessionClient.current?.updateSessions((current) =>
-      current.map((session) => (session.id === sessionId ? { ...session, title } : session)),
-    );
-    if (useWorkbenchStore.getState().sessionId === sessionId) {
-      useWorkbenchStore.getState().setSessionTitle(title);
-    }
+    syncAttachedSessionTitle(sessionId, title, true);
+    const client = ownerSessionClient.current;
+    const revision = client?.updateSessionTitle(sessionId, title) ?? null;
+    return { client, revision };
   }, []);
 
   /**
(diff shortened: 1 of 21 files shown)


components/workbench/workspace/WorkspaceShell.tsx
```diff
@@ -531,6 +531,7 @@ function WorkspaceShellController({ initialPanes }: { readonly initialPanes: Wor
         apply: (title) => applySessionTitle(sessionId, title),
         save: (title) => renameWorkbenchSession(sessionId, title),
       });
+      if (outcome === 'renamed') ownerSessionClient.current?.requestFullFetch();
```
wyuc: [P2] Fence concurrent rename completions. The rail and pane can submit two renames for the same session before the first request settles. Every completion and failure rollback is applied unconditionally, while the backing UPDATE has no revision or CAS condition. For A start, B start, B success, A failure, A restores the pre-A title in the UI; if A reaches SQL after B, it also becomes the durable title. Serialize renames per session or fence both client settlement and persistence with a monotonic revision.

lib/workbench/use-workbench-session.ts
```diff
@@ -112,6 +112,7 @@ export function useWorkbenchStream(sessionId: string | null): void {
 
   useEffect(() => {
     if (!sessionId) return;
+    const expectedTitleRevision = useWorkbenchStore.getState().sessionTitleRevision;
```
wyuc, line 117: [P2] Fence metadata by attachment instance. The current session-id check is insufficient for A to B to A navigation. attach() resets sessionTitleRevision to zero, so the first A request and the new A request can carry the same revision; the obsolete first response can then arrive last and overwrite a newer rename or explicit clear. Abort metadata work from cleaned-up effects or use a non-resetting attachment/request epoch.

components/workbench/workspace/WorkspaceShell.tsx
```diff
@@ -497,7 +497,7 @@ function WorkspaceShellController({ initialPanes }: { readonly initialPanes: Wor
    * which reads its own copy out of the session store.
    */
   const applySessionTitle = useCallback((sessionId: string, title: string | null) => {
-    setSessions((current) =>
+    ownerSessionClient.current?.updateSessions((current) =>
```
wyuc: [P2] Preserve title mutations across owner snapshots and clients. This local update is not represented in the owner event journal, so a full fetch that started before the rename can later replace it with an older snapshot. If the queued reconciliation fails, the rail remains stale until a periodic fetch. The same missing signal leaves another open tab’s attached header stale indefinitely because session metadata is fetched only on attachment. Add an owner title event/local mutation generation and synchronize the attached store from authoritative title updates.

wyuc (changes requested):
Three correctness races remain in title reconciliation. They can lose a newer rename, admit stale metadata after navigation, or leave open clients showing an obsolete title. Please address these interleavings and add deterministic regression coverage before merging.

components/workbench/workspace/WorkspaceShell.tsx
YizukiAme: Fixed by routing both rename entry points through a per-session queue, so same-session PATCHes are handled in submission order while different sessions remain independent. Settlements are revision-fenced, and an explicit restore queued behind an ambiguous write is still persisted. Regression coverage includes both failure orders and newer owner-title events.

lib/workbench/use-workbench-session.ts
YizukiAme: The metadata request is now bound to its attachment effect’s lifetime. Once the user navigates away, that request is invalidated even if they later return to the same session. The title revision fence remains in place for renames and explicit clears, with a delayed A → B → A response covered by a regression test.

components/workbench/workspace/WorkspaceShell.tsx
YizukiAme: Local renames now update the owner-session client directly, with mutation revisions protecting them from older list responses. Persisted changes are also published as session_title owner events, keeping the rail and attached header in sync across tabs. A full reconciliation repairs missed events or lost responses.

YizukiAme:
Thanks for the careful review! I’ve addressed all three threads and added regression coverage for the relevant request ordering, navigation, snapshot, and cross-tab scenarios.

lib/workbench/owner-session-client.ts
```diff
@@ -296,11 +344,28 @@ export class OwnerSessionClient {
...
-        let merged: readonly ProHomeSessionItem[] = newestFirst(snapshot);
+        let merged: readonly ProHomeSessionItem[] = newestFirst(snapshot).map((session) => {
+          const mutation = this.titleMutations.get(session.id);
+          if (!mutation) return session;
+          if (mutation.revision !== titleRevisions.get(session.id)) {
+            // This decision happened after the request began, so the response
+            // cannot contain it yet.
+            return { ...session, title: mutation.title };
+          }
+          // This request began after the local decision and its row is now the
```
wyuc: [P2] Keep the local title authoritative while its PATCH is still pending. A reconciliation can start after the optimistic mutation but race ahead of the PATCH and return the old title. This branch then deletes the mutation and invalidates its revision, so the later successful PATCH settlement is ignored. If its owner event or follow-up fetch is unavailable, the UI remains reverted until periodic reconciliation. Track write settlement separately and do not let a snapshot fence an unresolved mutation.

wyuc (changes requested):
One title-reconciliation race remains: a snapshot can start after the optimistic decision but still read before its PATCH commits. Please keep unresolved writes authoritative until their settlement is known and add the corresponding ordering test.

lib/workbench/owner-session-client.ts
YizukiAme: Mutations now carry a `settled` flag, and `runFullFetch` captures the settled revisions when the request starts — a row can only retire a decision that was already settled before that read began and whose revision has not changed since. Anything unresolved keeps its local title and its revision, so the later PATCH settlement still lands. Covered by `keeps a pending title over a snapshot started before its PATCH settled` and `keeps the settled result over a snapshot that began while its PATCH was pending`; `lets a snapshot begun after a settled …` keeps the retirement path reachable, including `null → null`.

YizukiAme:
Addressed — write settlement is tracked separately from the optimistic decision now, so a snapshot that started before the PATCH resolved can no longer fence it, and the ordering tests cover both directions. Also in this range: a title event hidden by an unconfirmed mutation schedules its own reconciliation instead of waiting for the periodic fetch, a post-settlement snapshot that omits the session retires its mutation, and the rename queue survives a shell remount. `26a0b01` is documentation only — it records that a title change counts as rail activity, the same semantics `rename_stage` already has for courses, plus a two-session ordering test.

wyuc (approved):
The reported title-reconciliation races are addressed with explicit write-settlement tracking and deterministic ordering coverage. CI is green.

PR #1274, opened 2026-08-28
feat(agent): add fact-check skill

### LINKED ISSUE #1272: [Feature]: Add Fact Check skill for course factual reliability
## Problem or Motivation

AI-generated courses may contain obvious factual errors, knowledge errors, or serious hallucinations. Users need a lightweight way to improve course reliability without turning generation into a slow, exhaustive audit.

## Proposed Solution

Add a built-in agent skill:

- `name`: `fact-check`
- Display name: `事实核查` / Fact Check

The skill should support factual checking both while creating a course and after a course has been generated. It should use available materials and web search when needed while remaining fast and lightweight.

## Scope

- This is a general-purpose skill applicable to courses across different knowledge domains.
- It only checks objectively verifiable factual errors, such as dates, regulations, scientific definitions, and wording made misleading or imprecise by those errors.
- It does not make subjective judgments about opinions, interpretations, teaching styles, or preferences.

## Acceptance Criteria

- [ ] The Fact Check skill is displayed and works correctly in the frontend
- [ ] It can be used while creating a course and after a course has been generated
- [ ] It can be invoked both explicitly and implicitly by the agent
- [ ] It produces stable, useful factual-checking results
- [ ] Existing agent-runtime skill tests and type checks pass

cc @wyuc

sophietao20-star:
## Scope

Adds general-purpose fact checking for completed course content. It covers verifiable objective claims and factual imprecision, not subjective judgments.

## Changes

Adds the `fact-check` agent-runtime skill, localized labels, and tests. It checks completed pages before delivery without altering planning or generation inputs. For existing courses, it verifies only high-signal claims, reports concise evidence-backed findings, and lets users choose what to fix. Corrections never silently override an approved outline, uploaded material, or supplied fact; detected source conflicts are left unchanged and surfaced in an `ask_user` choice card.

AI-assisted. Closes #1272


wyuc (changes requested):
The creation flow does not currently guarantee that the agent reads the completed course content before claiming to fact-check it.

`list_scenes` only returns each page id, order, title, and type. Unlike the review flow below, this section never explicitly calls `read_stage`, so “read their text” is not an executable instruction and may leave the model checking only page metadata.

Please make the creation flow explicitly read every completed page with `read_stage` using `detail: "text"`, including narration and `nextOffset` pagination, before correcting or reporting factual issues.

sophietao20-star:
Fixed in `80d77ef4`: the creation flow now explicitly reads every completed page with `read_stage` using `detail:"text"`, follows `nextOffset`, and includes both visible text and narration. The test now scopes these assertions to the creation section so the review flow cannot satisfy them accidentally. @wyuc

wyuc (approved):
Approved. The remaining concerns are accepted as broader agent-runtime capability limitations rather than blockers specific to this skill.

PR #1276, opened 2026-08-29
perf(classroom): speed up classroom loading (media hydration, sidebar thumbnails, media range requests)

wyuc:
## Summary

Entering a media-heavy classroom currently blocks on hydrating every restored
media blob and eagerly renders a full slide canvas for every scene in the
sidebar. This PR moves the expensive work off the load path.

- **Deferred media hydration**: classroom load now restores media metadata
  synchronously and only materializes object URLs for the scene it opens on
  (persisted cursor, else first scene). All other restored blobs hydrate in
  idle-time chunks afterwards. Consumers resolve a known-but-not-yet-hydrated
  task as pending and re-render when the URL lands, so nothing is regenerated
  or broken. (`lib/classroom/load-classroom.ts`)
- **Lazy sidebar thumbnails**: the playback scene sidebar now gates
  `SlideThumbnail` rendering with the near-viewport IntersectionObserver
  already used by the editor nav rail (extracted to
  `lib/hooks/use-near-viewport.ts`). Off-screen scenes render the cheap
  placeholder and skip their `<video preload="metadata">` fetches.
- **Lazy/async images + Range requests**: renderer images get
  `loading="lazy" decoding="async"` (the off-screen `slideToPng` snapshot tree
  forces eager to keep exports correct), and the classroom media route answers
  single byte-range requests with proper 206/416 semantics so hosted
  audio/video can stream progressively and seek.

## Verification

- `tsc --noEmit`, `eslint`, `prettier --check` clean
- `vitest run`: new tests for deferred media restore, near-viewport hook,
  range parsing, and the media route; full suite green except pre-existing
  environmental timeouts that reproduce identically on the base commit

No user-facing behavior changes: same content, same persistence semantics,
just less work before first paint.


wyuc:
## Cross-review summary

Ran 6 codex review rounds over the full diff; converged at zero actionable
findings. Fixed along the way (each with a regression test):

- deferred hydration could attach stale bytes to a task that regeneration had
  restarted — now guarded by task status
- idle-scheduled hydration had no timeout — now bounded at 1s per slice
- hydration kept running for an abandoned classroom — now cancelled by a
  restore epoch composed with the classroom load token
- a reload of the same stage could receive bytes read by the older load —
  covered by the same epoch
- 416 range errors were marked immutable-cacheable — now `Cache-Control:
  no-store`
- element-id-keyed and legacy-singleton-recovered media of the opening scene
  were deferred despite being visible — both are classified as priority now

CI green on the final head (lint/typecheck/unit, E2E, render service, storage
contract).

LING-6150 (changes requested):
@wyuc  Overall, this looks good to me. I found one edge case around deferred video hydration that may leave `play_video` stuck:

**Deferred videos can miss `play_video` before their URL is hydrated**

This introduces a restored task state where `status === 'done'` but `objectUrl` is still undefined. The renderer correctly treats this state as pending, but `ActionEngine.executePlayVideo()` still treats `status === 'done'` as immediately playable.

For an imported or legacy classroom, a video on a non-opening scene can still be deferred when the user reaches that scene. In that case, `play_video` sets `playingVideoElementId` while the renderer is showing a skeleton and no `<video>` exists. When hydration later supplies the URL, the component is preserved and its playback effect does not depend on the resolved source, so `video.play()` is not retriggered. Playback can then remain stuck on this action until the user intervenes or the five-minute timeout expires, after which the video is skipped.

Could we make the readiness checks in `executePlayVideo()` use the same renderability contract as the renderer? In particular, the initial check, subscription exit condition, and post-subscription recheck should not treat this new done-without-URL state as ready. It would also be useful to cover this with an engine-level regression test: start with `done` and no `objectUrl`, verify playback has not started, then add the URL and verify that it starts.

wyuc:
Good catch — fixed in 5b6064c7.

`executePlayVideo` readiness now follows the renderer's contract via `isPlayableVideoTask`: only `done` with an `objectUrl` counts as playable, so the deferred done-without-url state (which `resolveMediaRef` maps to pending) waits instead of starting. Applied at all three points you flagged: the initial check, the subscription exit condition, and the post-subscription recheck. I also moved the failed-skip outside the wait so a task failed at entry still skips playback.

Added the engine-level regression test in the shape you suggested (`tests/action/play-video-media-resolution.test.ts`): start `done` without `objectUrl` → playback does not start → hydration lands the URL → playback starts. Mutation-checked: dropping the `objectUrl` condition fails the test.

LING-6150 (approved):
Re-verified at afbdfbad. The deferred video playback issue I raised is resolved: ActionEngine now waits for a completed task to have its playable URL before starting play_video, including the initial check, subscription condition, and race recheck.
I also reviewed the subsequent stage-whiteboard priority update and did not find any additional actionable issues. Thanks for the quick fixes!

packages/@openmaic/renderer/src/elements/image/BaseImageElement.tsx
```diff
@@ -58,6 +58,12 @@ export function BaseImageElement({ elementInfo, renderImage }: BaseImageElementP
...
         draggable={false}
         data-soft-edge={softEdge || undefined}
+        // Lazy + async decode: thumbnail surfaces (sidebar, nav rail, course
+        // cards) mount far more images than the viewport needs. In-viewport
+        // images are unaffected — the browser fetches them immediately.
+        // slideToPng forces these back to eager in its off-screen tree.
+        loading="lazy"
+        decoding="async"
         style={{
           position: 'absolute',
```
thuyvylenguyen9691-jpg, line 68: lenguyenthuyvy663@gmail.com

PR #1281, opened 2026-08-30
feat(playback): reference static GenUI components

LING-6150:
## Summary

- Reuse the global **Reference courseware** entry in playback for HTML-backed GenUI / Interactive scenes.
- Allow one source-authored DOM component to be selected by a stable unique `id`, with one-shot picking, an exact-element persistent outline, and `PlaybackChromeRoot` remaining the sole draft and receipt owner.
- Resolve the selector again on the Host against the request-start Scene HTML, sanitize and deterministically bound the static evidence, and share the same request-scoped read-only attachment with the Director and both Legacy and Native Child runtimes.
- Preserve the existing PPT element-reference path, editor picker behavior, Pi-off behavior, classic `/api/chat`, and deferred Whiteboard/live-state/multi-select scope.
- Adds an independent default-off deployment gate for PPT and Interactive courseware references.

## Grounding contract

- The playback Browser seam sends identity only: `sceneId + selector`.
- Only HTML-backed Interactive scenes and stable unique source-authored IDs are accepted.
- Browser-provided text, markup, and live control state are not trusted as evidence.
- The Host resolves the selected component from the request-start Scene snapshot and rejects missing, duplicate, dynamic-only, URL-only, Canvas, and excluded nodes.
- Child evidence is sanitized, bounded to 24,000 Unicode code points, and shared unchanged across Director retries, handoffs, and Legacy/Native Child delegation.
- Interactive references never authorize Spotlight; PPT Spotlight behavior remains unchanged.

## Selection lifecycle

- The global playback toolbar entry arms either the existing PPT picker or the Interactive picker according to the current Scene.
- Interactive picking is single-shot and consumes the owner-owned arm synchronously, preventing queued or stale iframe messages from replacing the selected component.
- The outline remains attached to the exact clicked DOM element and fails closed after replacement, disconnection, or duplicate-ID invalidation.
- Successful accepted responses clear the pill and outline; failed or rejected requests retain them for retry. Manual clear and Scene navigation clear both projections consistently.

## Browser validation

- Dragged the density slider from its authored value `1000` to live value `1200`.
- Selected the slider through the global **Reference courseware** entry.
- Confirmed the pill and persistent outline, and confirmed later clicks could not replace the single selection.
- The grounded answer reported selector `density-slider`, `min=600`, `max=1400`, `step=50`, and source value `1000`, while explicitly stating that the current live slider position was unknown.
- Confirmed accepted-response cleanup and manual-clear cleanup for both pill and outline.

The automated Chromium regression also exercises real mouse input against a scaled iframe and verifies transform-aware picking.

## Feature gate
This PR adds NEXT_PUBLIC_COURSEWARE_REFERENCE_ENABLED, an independent, default-off build-time gate for the unified courseware-reference entry in Pi playback. The gate covers both PPT and Interactive scenes, while Pi chat itself and the existing editor element-reference workflow remain unaffected. CI and the Playwright development server enable the gate explicitly. Changing this value in a deployed application requires rebuilding the application or Docker image.

## Demo
OpenMAIC-demo

The recording demonstrates the global **Reference courseware** entry selecting one source-authored GenUI component, retaining the exact-element outline, grounding only the component's static source definition rather than its live control state, and clearing the reference after an accepted response.

## Independent cross-review

Two independent reviewers inspected the complete pinned diff and then challenged and reconciled each other's findings.

- The initial CI coverage blocker was closed by enabling the Pi feature flag for both the Playwright server and the CI production build.
- The stale/duplicate iframe-pick race was closed with one-shot shim disarm plus synchronous owner authority.
- The invalid-handle RAF lifecycle and the React pre-commit stale-pick/cancel windows are covered by focused regression oracles.
- The final full-diff review found **no actionable P0-P3 findings** and returned **APPROVE**.

This review closure supplements, but does not replace, the automated CI and Browser evidence below.

## Test plan

- `pnpm exec vitest run tests/components/edit/playback-chrome-root-element-reference-owner.test.ts tests/i18n/ppt-element-reference-locales.test.ts tests/lib/chat/pi/element-reference-route-l2.test.ts tests/lib/chat/pi/element-reference.test.ts tests/lib/chat/pi/interactive-component-reference.test.ts tests/lib/chat/pi/native-child-consumer.test.ts tests/lib/chat/pi/route-model-thinking.test.ts tests/lib/utils/iframe-element-picker.test.ts tests/lib/utils/iframe-playback-element-picker.test.ts tests/lib/utils/iframe-playback-projected-selector.test.ts tests/scene-renderers/interactive-iframe-picker.test.ts` — 11 files, 108 tests passed
- `pnpm exec playwright test e2e/tests/interactive-component-reference.spec.ts --project=chromium` — 1 passed
- `pnpm lint` — 0 errors (18 existing warnings)
- `pnpm check` — passed
- `pnpm check:i18n-keys` — 12 locales passed
- `git diff --check` — passed

Manual browser verification covered both configurations:
- PI_CHAT=true, COURSEWARE_REFERENCE=true: the unified reference entry, selection pill, retained outline, and accepted-response cleanup work for PPT and Interactive scenes.
- PI_CHAT=true, COURSEWARE_REFERENCE=false: the reference entry and reference state are absent, while ordinary Pi chat remains available.

## Scope

This PR intentionally does not add live runtime-state grounding, arbitrary/dynamic DOM selection, URL-only or Canvas/WebGL component references, Whiteboard references, multi-select, marquee selection, visual inference, editing authorization, a new Context/store/registry, a Native tool, or generic Runtime policy.


wyuc (changes requested):
The overall direction makes sense. Reusing the selection interaction while keeping playback references read-only and grounded in the request-start source snapshot is a reasonable distinction from the Pro Mode editor reference flow.

There is one resource-boundary issue I would like addressed before approval. The 24,000-code-point limit bounds only the final evidence packet: the resolver first parses the complete Scene HTML, deep-clones the selected subtree, walks all of its text, and serializes its full `outerHTML` before truncation. A large authored root can therefore cause substantial CPU and memory amplification even though only a small packet reaches the model. Please add an explicit input/work bound before `parseHTML` (or implement bounded traversal/serialization), together with a regression test for oversized Interactive HTML.

Please also attach a screenshot or recording from the actual E2E run, covering component selection, the reference pill and persistent outline, and cleanup after the request is accepted. This will make the UI behavior easier to verify beyond the green CI result.

LING-6150:
@wyuc 
Addressed in 777cf712.
- Added an explicit 256,000 UTF-16-code-unit source HTML bound before trim() and parseHTML(), preventing unbounded parsing, cloning, traversal, and serialization work.
- Added regression coverage for the exact limit, limit + 1, and oversized whitespace input.
- Updated the Chromium E2E to cover scaled component selection, the reference pill, persistent outline tracking, submitted sceneId + selector identity, and automatic pill/outline cleanup after an accepted receipt.
The screenshots below were captured from the E2E flow. The browser test uses a deterministic Pi response carrying X-OpenMAIC-Element-Reference-Accepted: 1 to isolate and verify the UI receipt contract. The real route → Director → Legacy/Native grounding path remains covered by the route/L2 tests, while the existing MaaS recording demonstrates the full real-model flow.
All CI checks are now passing. Could you please take another look?
PR1281-E2E-selected-component
PR1281-E2E-accepted-cleanup

LING-6150:
@wyuc   I added one focused follow-up commit:

[`eac2bf84`]`fix(pi): bound interactive reference HTML processing`

The reason for this follow-up is that the existing 24,000-code-point limit bounded only the final evidence packet. A large authored Interactive root could still require the Host to parse, traverse, clone, and serialize substantially more content before that final truncation was applied.

The commit keeps the existing Host resolver and adds a single bounded processing path:

`4M raw ceiling → one parse5 compaction pass → bounded compact HTML → linkedom → existing evidence projection`

During the compaction pass, content that cannot contribute valid component evidence—such as comments, excluded element payloads, template/raw-text payloads, and inline Base64 attributes—is removed before linkedom. The retained document is also bounded by node count, aggregate attributes, retained content, and depth. Boundary violations fail with the existing 400 response before model or Director execution and without the accepted receipt.

This is intentionally not a new sanitizer or parsing framework:

- only the existing Host-side Interactive resolver was changed;
- parse5 was already a direct repository dependency;
- no UI, picker, request schema, Prompt, Director/Child, runtime-state, feature-flag, or product behavior was added;
- the other two changed files contain focused regression coverage for the resource boundaries and Route fail-closed behavior.

Current verification:

- Pi focused suite: 21 files, 274 tests passed;
- `git diff --check` passed;
- Browser E2E confirmed component selection, the reference pill and persistent outline, and automatic cleanup after an accepted request.

I also successfully imported a separate exported experiment course archive of 40,511,198 bytes (about 38.6 MiB). This validates normal large-course compatibility. The archive size is distinct from the 4M limit, which applies only to the raw HTML of an individual Interactive Scene rather than to the complete course package and its media assets.

The attached GIF shows the production browser flow using the imported circuit experiment course.

PR1281-courseware-reference-e2e

wyuc (changes requested):
The bounded processing path lands as agreed: the 4 M raw ceiling is checked before any work, the compaction pass and structural budgets hold under measurement (worst accepted input about 260 ms and 130 MB transient across deep-nesting, wide-sibling, raw-text and base64 cases; every over-limit input is rejected pre-parse or inside 50 ms), the grounding contract survived selector-injection and compacted-vs-raw id probes, Spotlight stays PPT-only, and the branch merges cleanly with main. 416 tests, tsc, lint, prettier and the i18n key check pass on eac2bf84. Three implementation gaps before approval:

1. **`<template>` content is still billed against the 512 K retained-content budget although compaction discards it.** `isExcludedParent` requires an element parent, and parse5 parents template children under the template's `DocumentFragment`, so `insertText` retains and counts the payload that the post-parse walk then drops. A valid scene with a large data template next to a small target 400s with "exceeds the 512000-unit retained-content limit". Treat the fragment parent as excluded (or skip accounting for content under a `<template>`), and add a regression with an oversized template beside a small selected component.
2. **The iframe picker shim and the Host disagree on excluded tags.** The shim's list lacks `noembed`, `noframes`, `plaintext` and `xmp`, so the browser can arm and outline a pick that the Host always rejects with 400 (verified server-side for all four). Share one list.
3. **The `instruction` locale key still says "Click a slide element"** in en-US and nine other locales (only zh-CN and zh-TW say courseware element), and nothing surfaces it for Interactive scenes because the pick overlay that renders it only mounts for slide scenes. Update the copy in all twelve locales and give the Interactive arm the same instruction and Esc affordance.

Non-blocking notes: `<canvas>` fallback descendants resolve (only the canvas element itself is excluded), so the PR text's "Canvas rejected" is slightly wider than the code; `Object.freeze` on the runtime attachment is shallow (metadata is mutable; every consumer is read-only today); CI and Playwright now set `NEXT_PUBLIC_PI_CHAT_ENABLED` for the whole e2e suite rather than only the new spec; and the Scene HTML the Host resolves against comes from the request body's `storeState`, the same stateless contract the PPT path already uses, so "request-start snapshot" should not be read as server-side authority.

LING-6150:
@wyuc Thanks for the detailed review. I addressed all three implementation gaps in f4349586569753c7b1a618fcbeadd4953c1fffb3.

1. Template content is now excluded consistently from retained-content settlement, including descendant attributes. The raw-size, node, attribute, attribute-work and depth bounds remain enforced. Retained content is accounted once over the compacted tree, avoiding the per-text-token ancestor walk and its performance regression.

2. The Host resolver and iframe picker now use one shared excluded-tag list, including noembed, noframes, plaintext and xmp, so the Browser cannot select an element that the Host will always reject.

3. The instruction copy now refers to a “courseware element” in all 12 locales. Interactive scenes show the same instruction and Esc affordance as PPT scenes, using the existing picker owner state rather than introducing another selection state.

Verification:
- 27 focused test files, 335 tests passed
- TypeScript passed
- ESLint for the changed files passed
- Prettier passed
- 12-locale key alignment passed
- git diff --check passed
- A real exported course of approximately 39 MB imported successfully, and Interactive component selection worked normally

I have also attached the Browser E2E GIF showing the Interactive selection instruction, component selection, persistent outline, reference pill, Escape behavior and manual cleanup. 

PR1281-interactive-reference-e2e-final

wyuc (approved):
Round 3 on f4349586: all three gaps closed.

1. Retained-content accounting now happens once in the post-parse walk and `<template>` payloads (HTML and foreign namespace, text and attributes) are detached before billing, so a 600 K template beside a small selected component resolves while the same payload in a plain `<div>` still trips the 512 K limit; nested template / script / style inside a template are neither billed nor resolvable as targets. The four "does not bill discarded" tests go red on the previous accounting.
2. `lib/interactive/element-reference-policy.ts` is the single excluded-tag source for both the Host resolver and the iframe shim (15 tags, now including noembed / noframes / plaintext / xmp); the serialised form is a JSON object of lowercase tag names so it cannot break out of the injected script, and the shim test evaluates the patched source and checks the four new tags.
3. `chat.elementReference.instruction` reads as courseware copy in all twelve locales with an exact-string test; the Interactive host renders the instruction while armed, and Escape cancels the pick from both the page (capture handler, pinned by a test that fails when the effect is disabled) and the iframe (shim onKey → disarmed → cancel).

402/402 across the affected suites, tsc, eslint, prettier and the i18n key check clean; the commit touches exactly the 21 listed files, no lockfile.

One cosmetic note, pre-existing and harmless in direction: the shim's `excludedTags` is a plain object, so a tag literally named `constructor` is blocked by the shim while the Host would accept it. `Object.create(null)` would remove the asymmetry; not worth another round.

PR #1283, opened 2026-08-30
fix(asr): bypass AI SDK for custom providers to handle extra response…

### LINKED ISSUE #1277: [Bug]: Invalid JSON response Error when setting the custom Siliconflow API address
### Bug Description

OpenMAIC 的自定义 ASR provider 确实会把 apiKey 带出去（回声里 authorization: Bearer sk-...，model 也正确转发）。所以集成机制没问题。不是限流,是响应格式解析问题——ai-sdk 收到的不是它能 parse 的 JSON
直连 SiliconFlow 返回的是合法 JSON（{"duration":1.5,"text":"嗯嗯","usage":...}），但 OpenMAIC 的 ai-sdk 报 Invalid JSON response。而我的回声服务只回 {"text":"echo-ok"} 时 OpenMAIC 就成功。怀疑是 SiliconFlow 响应里多了 duration/usage 字段把解析器搞挂了。验证一下：让回声服务回 SiliconFlow 的原样 body，看 OpenMAIC 能否接受。
echo2 回原样 body 时 OpenMAIC 竟然成功了（{"success":true,"text":"..."}）！说明 body 格式没问题。那真·SiliconFlow 失败一定是传输层差异（SiliconFlow 对 OpenMAIC 的 fetch 回了不同东西）。写个转发代理，把 SiliconFlow 真实返回的原样抓出来看。
根因 100% 锁定了！ 对比两次：
echo2 回 {"duration":...,"text":...,"usage":...}（无 segments）→ OpenMAIC 成功
代理回 SiliconFlow 真实响应 {"duration":...,"text":"嗯嗯","segments":[...],"usage":...}（多了 segments）→ OpenMAIC 失败
SiliconFlow 的响应比 OpenAI 标准多了一个 segments 数组，OpenMAIC 的 ai-sdk 转录解析器（Zod 严格模式）吃到这个字段直接抛 Invalid JSON response。这是 SiliconFlow 与 OpenAI 格式不兼容，不是 key 或网络问题。
修复方案：一个薄代理，转发到 SiliconFlow 后剥掉 segments 再回给 OpenMAIC。

### Steps to Reproduce

自定义ASR api地址为siliconflow

### Expected Behavior

fixed

### Actual Behavior

OpenMAIC 的自定义 ASR provider 确实会把 apiKey 带出去（回声里 authorization: Bearer sk-...，model 也正确转发）。所以集成机制没问题。不是限流,是响应格式解析问题——ai-sdk 收到的不是它能 parse 的 JSON

### Deployment Method

Docker

### OpenMAIC Version

latest

### Affected Area

Model / provider integration

### Browser

Edge

### Operating System

Windows 10 LTSC 2021

### Relevant Logs

```shell

```

### Screenshots or Recordings

_No response_

### Additional Context

_No response_

maheshsingh20:
## Summary

Custom ASR providers (e.g. SiliconFlow) return OpenAI-compatible JSON but include extra fields like `segments`, `duration`, and `usage` that the Vercel AI SDK's strict Zod response schema does not expect, causing it to throw an `Invalid JSON response` error even when the `text` value is perfectly valid.

## Related Issues

Fixes #1277

## Changes

- Introduce `transcribeCustomOpenAICompatibleASR()` in `lib/audio/asr-providers.ts` — a raw `fetch`-based implementation that POSTs multipart FormData to `/audio/transcriptions` and extracts only `data.text`, making it tolerant of any extra fields a provider returns
- Route all `custom-asr-*` providers through this new function instead of `transcribeOpenAIWhisper()`, bypassing the AI SDK's strict Zod response schema entirely
- The `openai-whisper` built-in case is unchanged and continues to use `experimental_transcribe` from the Vercel AI SDK
- Follows the same raw-fetch pattern already used by `funasr-asr` and `lemonade-asr`

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Add a custom ASR provider in OpenMAIC settings with SiliconFlow's base URL (`https://api.siliconflow.cn/v1`) and a valid API key
2. Select a SiliconFlow Whisper model (e.g. `FunAudioLLM/SenseVoiceSmall`)
3. Record audio and submit — previously got `Invalid JSON response`; now returns the transcribed text correctly

- Traced root cause: SiliconFlow's response `{"duration":...,"text":"...","segments":[...],"usage":...}` contains a `segments` array not present in the OpenAI Whisper spec; the SDK's strict schema rejects it
- Confirmed that stripping `segments` from the response body (via proxy) made OpenMAIC succeed — proving the field was the sole cause
- New function extracts only `data.text`; all extra fields (`segments`, `duration`, `usage`) are silently ignored
- All existing provider paths (`openai-whisper`, `qwen-asr`, `azure-asr`, `funasr-asr`, `lemonade-asr`) are untouched
- All three call sites covered: live mic recording (`use-audio-recorder.ts`), settings test button (`asr-settings.tsx`), and server-side media extraction (`local-media.ts`) — all route through `transcribeAudio()`
- TypeScript diagnostics: clean (no errors or warnings)

### Evidence

- No UI changes; fix is server-side only

- [ ] Manually tested locally

## Checklist

### DIFF
--- a/lib/audio/asr-providers.ts
+++ b/lib/audio/asr-providers.ts
@@ -336,12 +336,21 @@ async function transcribeCustomOpenAICompatibleASR(
     audioBlob = new Blob([arrayBuffer], { type: 'audio/webm' });
   }
 
+  // Most OpenAI-compatible transcription endpoints treat `model` as a
+  // required field.  Custom providers have no built-in default, so we
+  // require the caller to supply one and fail fast with a clear message
+  // rather than sending a request that will almost certainly return a 400.
+  if (!config.modelId?.trim()) {
+    throw new Error(
+      'Custom ASR provider requires a model ID (e.g. FunAudioLLM/SenseVoiceSmall). ' +
+        'Please set the model in the provider settings.',
+    );
+  }
+
   const formData = new FormData();
   // Use 'file' key — required by the OpenAI audio/transcriptions spec.
   formData.set('file', audioBlob, 'audio.webm');
-  if (config.modelId) {
-    formData.set('model', config.modelId);
-  }
+  formData.set('model', config.modelId);
   formData.set('response_format', 'json');
   if (config.language && config.language !== 'auto') {
     formData.set('language', config.language);
--- a/lib/audio/asr-providers.ts
+++ b/lib/audio/asr-providers.ts
@@ -198,7 +198,7 @@ export async function transcribeAudio(
 
     default:
       if (isCustomASRProvider(config.providerId)) {
-        return await transcribeOpenAIWhisper(config, audioBuffer);
+        return await transcribeCustomOpenAICompatibleASR(config, audioBuffer);
       }
       throw new Error(`Unsupported ASR provider: ${config.providerId}`);
   }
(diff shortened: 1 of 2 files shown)


maheshsingh20:
CI is now green across all checks, including the previously-failing lint/typecheck/unit test job. This is ready for review whenever a maintainer gets a chance to take a look.

CXuPercy (changes requested):
Cross-reviewed by two independent reviewers. The overall approach is right — bypassing the AI SDK's strict Zod validation so providers returning extra fields (SiliconFlow's segments/duration/usage) keep working, with only `data.text` extracted — and the test coverage (11 scenarios) is thorough. One regression needs fixing before merge:

**Missing model fallback.** The new path sets `model` only when `config.modelId` is truthy:

```ts
if (config.modelId) {
  formData.set('model', config.modelId);
}
```

The path this replaces (`transcribeOpenAIWhisper`) explicitly falls back — `config.modelId || ASR_PROVIDERS['openai-whisper'].defaultModelId` (verified on main). With this PR, a custom provider configured without a model id now sends no `model` field at all, and OpenAI-compatible transcription endpoints generally treat `model` as required, so that configuration goes from working to failing with a 400. Please restore an equivalent fallback (or fail fast with a descriptive error if you'd rather make modelId mandatory for custom providers — but then the settings UI contract needs to match).

Minor, non-blocking: the hardcoded `audio/webm` fallback for Buffer inputs matches the file's existing convention, so no change needed there.

maheshsingh20:
Hi @CXuPercy,
Thanks for the thorough review — you caught a real regression.

. Here's what changed in the new commit (`3e850698`):

**Root cause confirmed:** the original `transcribeOpenAIWhisper` path fell back to `ASR_PROVIDERS['openai-whisper'].defaultModelId` when `config.modelId` was absent. The new raw-fetch path had no equivalent fallback — it silently dropped the `model` field from the FormData, which causes a `400` on any endpoint that treats `model` as required (SiliconFlow and most OpenAI-compatible providers do).

**Fix:** explicit guard before building the FormData:
```ts
if (!config.modelId?.trim()) {
  throw new Error(
    'Custom ASR provider requires a model ID (e.g. FunAudioLLM/SenseVoiceSmall). ' +
      'Please set the model in the provider settings.',
  );
}
formData.set('model', config.modelId); // now unconditional
```

**UI contract check:** I verified `components/settings/asr-settings.tsx` and `add-audio-provider-dialog.tsx`. A custom provider can be saved with zero models (the creation dialog has no model field for ASR), but the settings panel already shows an amber warning — _"Please add at least one model below before using this provider."_ — so the UI is already aligned with the fail-fast behaviour. No UI change needed.

**Tests:** all 12 test cases now supply a `modelId`, and a dedicated regression test `'throws when no model ID is configured'` has been added to prevent this regressing silently. All 12 pass locally and CI is green across all 4 checks.

CXuPercy (approved):
The follow-up addresses the regression exactly right. Verified in the diff and against CI: the fail-fast guard fires before any FormData is built (`requires a model ID … set the model in the provider settings`), `model` is now set unconditionally, and the dedicated regression test asserts the throw plus that fetch was never called. I also checked your UI-contract reasoning and agree with the fail-fast choice over a default fallback: the old path's fallback sent OpenAI's whisper model id, which custom endpoints would reject anyway, so a descriptive error aligned with the settings panel's existing "add at least one model" warning is the better contract. Cross-reviewed by two independent reviewers; the second reviewer had no further findings. CI green.

PR #1285, opened 2026-08-30
feat(render-service): add POST /preview endpoint (fixes render_scene_preview 404)

### LINKED ISSUE #1284: render-service: add POST /preview endpoint (fixes render_scene_preview 404)
## Problem

The `render_scene_preview` agent tool (`lib/server/agent-runtime/scene-preview.ts`) calls a `POST /preview` endpoint on the render-service, but the render-service does not implement that route. The tool therefore 404s — a broken link — so persisted pages can't be visually inspected during generation.

## Proposal

Add a `POST /preview` endpoint to the render-service that renders a single scene to a PNG, with its own admission control so previews don't compete with the (minute-scale) MP4 export queue:

- **`PreviewGate`** — independent admission control (global in-flight cap + per-identity cap), reject-before-buffering, idempotent release.
- **Preview renderer** — renders `{scene, stage, viewport}` to a PNG via `SlideCanvas` + static markup, reusing the existing Chromium executable configuration, egress policy, and semaphore infrastructure, with a deadline/AbortSignal.
- **`POST /preview`** — declared-size rejection, gate admission before buffering, byte-capped payload parsing, scene/stage/viewport validation, deadline/cancellation, PNG response, and 400/413/429/504/500 error mapping.
- Config knobs (`previewDeadlineMs`, `previewMaxInFlight`, `previewMaxPerUser`) and resource-profile viewport pixel limits.

This makes the existing `render_scene_preview` tool functional end-to-end.

See the linked PR for the implementation.

wyuc:
Fixes #1284.

## What

Adds a `POST /preview` endpoint to the render-service that renders a single scene to a PNG. This fixes the broken link from the existing `render_scene_preview` agent tool, which previously 404'd because the render-service had no `/preview` route.

## Changes

- **`preview-gate.ts`** — `PreviewGate`: independent admission control (global in-flight cap + per-identity cap), reject-before-buffering, idempotent release. Previews get their own gate so they never compete with the minute-scale MP4 export queue.
- **`preview-renderer.ts`** — renders `{scene, stage, viewport}` to a PNG via `SlideCanvas` + static markup, reusing the existing Chromium executable configuration, egress policy, and semaphore infrastructure, with a deadline/AbortSignal.
- **`main.ts`** — wires `POST /preview`: declared-size rejection, gate admission before buffering, byte-capped payload parsing, scene/stage/viewport validation, deadline/cancellation, PNG response, and 400/413/429/504/500 mapping.
- **`config.ts` / `resource-profile.ts`** — `previewDeadlineMs` / `previewMaxInFlight` / `previewMaxPerUser` config and resource-profile viewport pixel limits.
- **Tests** — gate tests + endpoint tests (14 files, 85 tests pass).

## Validation

- `pnpm install` + `npm ci --ignore-scripts` (render-service) — pass
- `npm run typecheck` — pass
- `npm test` — 14 files, 85 tests pass
- ESLint on changed files — pass
- `pnpm check` (prettier) — pass


wyuc:
**Review context** — rebased onto current main (incl. #1351) and through an internal dual-reviewer pass (3 rounds); final state differs substantially from the original diff:

- Preview is decoupled from the render queue: try-acquire on the shared Chromium execution slot → 429 `capacity_busy` when busy (synchronous previews never queue behind long renders; video admission never stalls behind previews); extraction permit held only across JSON body parse.
- Machine-readable 429 reasons following the #1351 contract (`preview_queue_full` / `preview_per_user_limit` / `capacity_busy`); the deadline actually bounds resource holds (bounded `protocolTimeout`, SIGKILL on abort/close timeout); client disconnect aborts work promptly.
- Dedicated 32 MiB JSON body cap; per-owner concurrency limit (default 2, RENDER_PREVIEW_* knobs).
- Fail loud: undrawable or non-self-contained scenes reject with 422 before Chromium launches (egress-locked container can only load inline/`data:` content); in-page mount crashes fail fast; interactive iframe gets an in-memory storage shim.
- README documents the endpoint, env knobs, 429 reasons, and capacity interaction with /render and /health.

Known limitation (documented): self-containment detection is best-effort; end-to-end fidelity for persisted scenes (asset resolution + dependency inlining, like video export does) is the caller's responsibility — `render_scene_preview` follow-up.

CXuPercy (commented):
Full cross-review done (ZCode + independent second reviewer + local test run: 131/131, plus adversarial probes). Architecture verified end to end — admission ordering (gate before any body read), dual byte caps, profile pixel budget, the no-queue execution slot shared with video renders, and the browser lifecycle with bounded close + SIGKILL fallback all check out.

Two robustness gaps surfaced on the validation path, both confirmed experimentally — sharing so you can decide whether to fix here or as a follow-up:

1. **`canvas.elements: [null]` passes `validateScene`, then throws in `previewabilityError`** — `Cannot read properties of null (reading 'type')` from the slot-descriptor walk, so the route's generic handler answers **500** where the taxonomy promises 400/422. Repro: any slide whose `canvas.elements` contains a null.

2. **Deeply nested interactive HTML overflows the recursive walk** — 60k nested `<i>` tags produce `RangeError: Maximum call stack size exceeded` in `findNonSelfContainedInteractiveReferences`'s `visit()` → again an uncontrolled 500. V8 catches it and the service survives (the gates release, and 32 MiB + in-flight caps bound the cost), but there's no depth/element-count ceiling, so adversarial payloads land as 500s instead of controlled rejections.

Reviewed and NOT flagging as blocking: the `x-openmaic-client` identity model — it inherits the same trusted-proxy assumption the `/render` per-user guard already documents, and the global preview in-flight cap bounds total damage regardless of identity spoofing. One doc-level nit alongside: the README could state more prominently that a running preview can delay a queued video render by up to the preview deadline.

How do you want to handle 1 and 2 — fix here (both look like small changes on the validation path, each with an adversarial test) or land as an immediate follow-up? Either works for me; the architecture itself reads as approve-quality.

wyuc:
Both gaps fixed in this PR (3807b730), not deferred:

1. **Malformed `canvas.elements` entries** — new `invalidSlideCanvasElementError` rejects `null`/`undefined`/missing-`type` elements at payload parse time (400, with the exact `/content/canvas/elements/<i>` path) and again inside `previewabilityError` (422). `countNonSelfContainedSlideMediaReferences` is now non-throwing as defense in depth. No 500 path remains for these shapes.

2. **Unbounded HTML walk** — `findNonSelfContainedInteractiveReferences` is now iterative with explicit ceilings: `MAX_INTERACTIVE_HTML_DEPTH = 256` elements, `MAX_INTERACTIVE_HTML_ELEMENTS = 10,000`. Violations raise a limit error that `previewabilityError` maps to a 422 naming the exceeded ceiling; other errors still rethrow. Child-before-template-content traversal order preserved.

Tests: adversarial route + unit cases for both, each asserting the controlled status (never 500), the error path/limit message, that `render` was not invoked, and that the fixture actually exceeds the ceiling so the cases cannot silently degrade. Suite is 131 → 141 passing on Node 22, typecheck clean.

@CXuPercy if this closes your two findings, a formal approve would land it.

CXuPercy (dismissed):
Both findings verified closed against 3807b730 — re-ran the original adversarial probes directly:

- `canvas.elements: [null]` now answers a controlled parse-time error (`Invalid scene at /content/canvas/elements/0: canvas element must be an object with a non-empty type`), with the non-throwing slot walk as defense in depth;
- 60k-deep nested HTML now answers `Interactive HTML exceeds the maximum DOM depth of 256`, from the iterative bounded traversal (depth 256 / 10k elements ceilings).

Suite 131 → 141 with the new adversarial cases, all passing locally; the render-service CI job is green. The architecture verification from the earlier review stands (admission ordering, dual byte caps, pixel budget, no-queue execution slot, browser lifecycle).

One residual before merge: the main lint job is red on prettier formatting in `render-service/src/main.ts` and `render-service/test/preview-route.test.ts` — a `prettier --write` commit closes it.

Approved from my side once that lands.

PR #1294, opened 2026-08-31
fix: expose branded media providers in settings

ChristianGrubbs:
## Summary

- expose operator-defined display names for server-managed image and video providers without changing adapter IDs
- synchronize server-managed media model allowlists into settings and generation selectors
- avoid showing an upstream vendor logo when a compatibility adapter has a different operator identity

## Verification

- 193 focused tests passed
- ESLint passed for changed files
- TypeScript check passed
- production build passed
- browser-verified both Image Generation and Video Generation settings against the rebuilt local server

### DIFF
--- a/components/generation/media-popover.tsx
+++ b/components/generation/media-popover.tsx
@@ -66,9 +66,19 @@ const TABS: Array<{ id: TabId; icon: LucideIcon; label: string }> = [
 
 function providerModels<T extends { id: string; name: string }>(
   builtInModels: T[],
-  config?: { customModels?: T[]; replaceBuiltInModels?: boolean },
+  config?: {
+    customModels?: T[];
+    replaceBuiltInModels?: boolean;
+    serverModels?: string[];
+  },
 ): T[] {
   const customModels = config?.customModels || [];
+  if (config?.serverModels?.length) {
+    return config.serverModels.map((id) => {
+      const knownModel = [...builtInModels, ...customModels].find((model) => model.id === id);
+      return knownModel ?? ({ id, name: id } as T);
+    });
+  }
   if (config?.replaceBuiltInModels && customModels.length > 0) {
     return customModels;
   }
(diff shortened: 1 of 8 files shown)


CXuPercy (changes requested):
Cross-reviewed by two independent reviewers. The direction is right — operator branding for compatibility adapters is a real self-hosting need, and the `serverModels` separation is the better of the two designs floating around for server model lists. Three things before merge:

**1. Rebase and unify on `serverModels` as the single mechanism.** This PR predates #1298's merge, so it now sits on a main where the same sync blocks inject server models into `customModels` + `replaceBuiltInModels`. Two mechanisms on the same fields can't ship together. Please adopt this PR's `serverModels` field as THE mechanism and remove the #1298-style injection path.

Doing so also fixes a verified bug on current main: when a server-pinned model list is later removed, the previously injected `customModels` + `replaceBuiltInModels: true` are never cleared, so the client keeps displaying the stale allowlist after the operator lifts the restriction. Your separation (own field, cleared on reset) closes exactly that. Please add removal/empty-list transition tests for image and video so this stays fixed.

**2. Generalize `displayName` across all server-managed provider families.** We're accepting operator display names as a general capability of server-managed providers, not an image/video feature — so please build it that way: the config surface (YAML field + `<PREFIX>_DISPLAY_NAME` env) and the `/api/server-providers` payloads for LLM, TTS, ASR, PDF, image, video, and web-search alike, with the client sync and the logo-suppression rule following the same generic path. Doing it once now avoids a trail of "me too" PRs per family later. If the LLM panel's surface makes one PR unreasonable, land the generic mechanism plus the non-LLM families here and propose the LLM extension explicitly — but the mechanism itself should be family-agnostic from the start.

**3. UI verification before final approval.** This change is user-visible (names, logo suppression, model pickers), so we'll ask for browser screenshots or a local pass on the reworked branch before approving — same bar as #1342.

The verified positive-path tests you have are good; they just need the transition and generalization coverage above.

PR #1297, opened 2026-08-31
fix(skills): normalize paths to POSIX before passing to pi-agent-core loader

### LINKED ISSUE #1295: [Bug]: Built-in skill loading throws on Windows paths
### Bug Description

On Windows, the built-in agent skill loader throws before it can return any recursively discovered skill. OpenMAIC pins `@earendil-works/pi-agent-core` 0.78.0; that version's relative-path helper only normalizes `/`, while the Node execution environment returns child paths with `\` on Windows. The resulting absolute-looking path is passed to the `ignore` package, which rejects it.

This is on the production `listSkills()` path (`lib/server/agent-runtime/skills.ts`), and it also prevents the real-skill runner tests from reaching agent setup.

### Steps to Reproduce

1. Check out OpenMAIC on Windows at commit `dfebbcf3`.
2. Install dependencies with Node 24.14.0 and the repository-pinned pnpm 10.28.0.
3. Run:

   `pnpm exec vitest run tests/agent-runtime/skills.test.ts -t "every shipped skill directory actually loads"`

4. Observe the loader exception.

### Expected Behavior

The shipped skill directories load successfully and `listSkills()` returns the built-in skills on Windows, as it does on POSIX systems.

### Actual Behavior

The focused test fails before loading the first recursive skill directory.

### Deployment Method

Local development (pnpm)

### OpenMAIC Version

`dfebbcf3`

### Affected Area

Other — agent runtime / built-in skills

### Browser

Not applicable

### Operating System

Windows 11

### Relevant Logs

```shell
RangeError: path should be a `path.relative()`d string, but got "C:\repo\OpenMAIC\skills\agent-runtime\build-personal-skill/"
  at loadSkillsFromDirInternal (.../@earendil-works/pi-agent-core/dist/harness/skills.js:112:27)
  at listBuiltinSkills (lib/server/agent-runtime/skills.ts:143:35)
```

### Screenshots or Recordings

Not applicable

### Additional Context

I found no matching OpenMAIC issue or pull request after searching for the error, `loadSkills`, Windows skill loading, and `pi-agent-core`.

The dependency's upstream project tracked the same Windows path defect in [earendil-works/pi#7427](https://github.com/earendil-works/pi/issues/7427) and fixed it in [`6a14037`](https://github.com/earendil-works/pi/commit/6a140371d6fa1069f89724741de912ab2f266ef8). OpenMAIC currently pins both `pi-agent-core` and `pi-ai` to 0.78.0, so any dependency update should keep the pair compatible and be validated against OpenMAIC's full TypeScript and test suite. I have not yet verified a full version bump.

mturac:
## Summary

Fixes #1295 — built-in skill loading throws on Windows because `NodeExecutionEnv` returns backslash-separated paths that the `ignore` package rejects.

**Root cause:** `pi-agent-core`'s skill loader passes paths from `NodeExecutionEnv.listDir()` and `fileInfo()` to the `ignore` package, which requires POSIX-style relative paths. On Windows, Node's `path.resolve` and `fs.readdir` return `\`-separated paths, causing:

```
RangeError: path should be a `path.relative()`d string, but got "C:\repo\OpenMAIC\skills\agent-runtime\build-personal-skill/"
```

**Fix:** Wrap `NodeExecutionEnv` with `PosixNormalizingEnv` that converts all `FileInfo.path` values to forward slashes in `fileInfo()` and `listDir()` overrides. Normalize `skillsDir` itself before passing it to the loader.

## Changed files

- `lib/server/agent-runtime/skills.ts` — add `toPosixPath` helper, `PosixNormalizingEnv` wrapper class, normalize `skillsDir` in `listBuiltinSkills()`
- `tests/agent-runtime/skills.test.ts` — 3 regression tests for POSIX path normalization

## Test plan

- [x] `vitest run tests/agent-runtime/skills.test.ts` — 76 passed, 0 failed
- [x] `tsc --noEmit` — clean, exit 0


HaningZS (commented):
Thanks for picking this up. I found one Windows-specific gap that the current tests do not exercise.

`pi-agent-core` 0.78.0's `NodeExecutionEnv` derives both `FileInfo.path` **and** `FileInfo.name` with forward-slash-only logic. On Windows, `fileInfoFromStats()` therefore returns a `name` containing the full backslash-separated absolute path. This patch normalizes only `info.path`; `info.name` remains incorrect.

That matters because the recursive loader checks `entry.name === 'SKILL.md'`. With path-only normalization, the ignore matcher no longer rejects the path, but the nested `SKILL.md` entry is still never recognized, so `listSkills()` returns no built-in skills.

The added Windows tests cannot catch this on the Linux CI host: the function under test is a no-op when `sep === '/'`, and the other assertions exercise a separate inline `split('\\').join('/')` expression rather than `toPosixPath()` or the production `listSkills()` boundary.

I reproduced this deterministically by mocking `NodeExecutionEnv` to return the same Windows-shaped `FileInfo` values while reading the real shipped skill tree. The current implementation loads zero skills; normalizing both `path` and the basename makes all 23 load. A tested reference implementation and production-boundary regression are available in [`3e099686`](https://github.com/HaningZS/OpenMAIC/commit/3e099686).

Suggested changes:

1. Recompute `FileInfo.name` from the normalized path in both `fileInfo()` and `listDir()` results.
2. Replace or supplement the platform-conditional unit tests with a deterministic test that drives OpenMAIC's real `listSkills()` path using Windows-shaped environment results.

I validated that version with 74 focused skill tests, 1,230 related agent-runtime tests, the full 7,083-test root suite, package CI gates, and a Node 22 production build. I have not run it on a native Windows host, so that limitation should remain explicit.

mturac:
Thank you for the thorough review and the reference implementation. You're right — normalizing only `path` leaves `name` carrying the full Windows-style absolute path, so `entry.name === 'SKILL.md'` never matches on Windows.

Addressed in e187896:
- Replaced the path-only normalization with `normalizeSkillFileInfo()` that derives `name` from the normalized path (matching your approach in 3e09968)
- Added deterministic unit tests for `normalizeSkillFileInfo` covering Windows paths, trailing separators, POSIX no-op, and property preservation
- Added a `listSkills()` production-boundary test asserting built-in skills load successfully

Tests: 81/81 pass, typecheck clean, prettier clean.

CXuPercy (approved):
Cross-reviewed by two reviewers. The approach is good — subclassing NodeExecutionEnv to normalize results at the boundary is cleaner than post-processing, and I verified it covers the only loadSkills call site on main (listBuiltinSkills), so nothing else still feeds backslash paths to the ignore package. Approving with two nits from the second reviewer, both valid but low-severity:

1. `normalizeSkillFileInfo()` replaces backslashes unconditionally, even on POSIX where `\` is a legal filename character — an exotic-but-real corruption path. Consider gating the replacement on `sep === '\\'` (i.e. only normalize where Node actually produces backslash paths).
2. The `toPosixPath` Windows/mixed-separator tests manually perform `split('\\').join('/')` instead of calling the helper, so they'd pass even if the helper's Windows branch were broken. Make them assert against `toPosixPath()` output (and ideally add an integration test exercising the normalizing env with Windows-shaped results).

Happy to see these addressed in a follow-up commit here rather than a separate PR.

PR #1298, opened 2026-08-31
fix(providers): apply server-pinned models for image and video providers

### LINKED ISSUE #1251: [Bug]: Server-configured image/video model allowlist is not applied, causing UnsupportedModel
### Bug Description

Configuring an image or video provider via `server-providers.yml` (or env) with a `models` allowlist has no effect. The server exposes those models but the client never receives them, so it falls back to the built-in dated model IDs. On endpoints that require different IDs — e.g. the Volcengine Ark Agent Plan, which uses dotted aliases like `doubao-seedream-5.0-lite` instead of `doubao-seedream-5-0-260128` — generation fails with a 404 `UnsupportedModel`.

### Steps to Reproduce

- Add a server-managed image provider with a custom model list:
  
  ```yaml
  image:
    seedream:
      apiKey: ...
      baseUrl: https://ark.cn-beijing.volces.com/api/plan/v3
      models:
        - doubao-seedream-5.0-lite
  ```
  
- Start the app and open the image generation UI.
  
- Trigger an image generation.

### Expected Behavior

The server-pinned model `doubao-seedream-5.0-lite` becomes the selectable/active model, and generation succeeds.

### Actual Behavior

The client still uses the built-in `doubao-seedream-5-0-260128`, and generation returns:

```
Seedream generation failed (404): {"error":{"code":"UnsupportedModel","message":"The requested model does not support the agent plan feature. ..."}}
```

### Deployment Method

### OpenMAIC Version

latest

### Affected Area

Classroom generation

### Browser

Not applicable

### Operating System

Windows11

### Relevant Logs

```shell

```

### Screenshots or Recordings

_No response_

### Additional Context

`/api/server-providers` returns `image.models`, but `getServerVideoProviders()` returns no models at all. In `lib/store/settings.ts`, `fetchServerProviders` marks image/video providers as `isServerConfigured: true` but **discards** the `models` payload. The LLM path already consumes `providers.models` correctly; image/video do not.

mturac:
## Summary

Fixes #1251 — server-configured image/video model allowlists are not applied to the client, causing generation to fall back to built-in model IDs that fail on endpoints requiring different aliases.

**Root cause:** Two gaps in the server-to-client model pipeline:

1. `getServerVideoProviders()` omitted the `models` field entirely from its response, despite the server resolving them internally via `resolveVideoModel()`.
2. `fetchServerProviders()` on the client discarded the `models` payload for both image and video providers — it only set `isServerConfigured` and `serverDisabled`, never populating `customModels`.

This meant the client always used built-in dated model IDs (e.g. `doubao-seedream-5-0-260128`) even when the server pinned different ones (e.g. `doubao-seedream-5.0-lite`), causing `UnsupportedModel` errors on endpoints that require the pinned aliases.

**Fix:**
- Server: `getServerVideoProviders()` now exposes `models`, mirroring `getServerImageProviders()`.
- Client: `fetchServerProviders()` stores the server model list as `customModels` with `replaceBuiltInModels: true` for both image and video, matching the existing LLM provider model-filtering pattern.

## Changed files

- `lib/server/provider-config.ts` — `getServerVideoProviders()` now includes models in response
- `lib/store/settings.ts` — `fetchServerProviders()` processes `models` for image and video
- `tests/server/provider-config.test.ts` — 2 regression tests for model exposure
- `tests/store/settings-server-sync.test.ts` — 3 regression tests for client-side model merge

## Test plan

- [x] `vitest run tests/server/provider-config.test.ts` — 109 passed
- [x] `vitest run tests/store/settings-server-sync.test.ts` — 82 passed
- [x] `tsc --noEmit` — clean, exit 0

### DIFF
--- a/lib/server/provider-config.ts
+++ b/lib/server/provider-config.ts
@@ -893,10 +893,16 @@ export function resolveImageModel(providerId: string, clientModel?: string): str
  * (presence = managed flag) plus operator force-disabled providers
  * (`{ disabled: true }`), mirroring the TTS listing — disable wins (#665).
  */
-export function getServerVideoProviders(): Record<string, { disabled?: boolean }> {
+export function getServerVideoProviders(): Record<
+  string,
+  { models?: string[]; disabled?: boolean }
+> {
   const cfg = getConfig();
-  const result: Record<string, { disabled?: boolean }> = {};
-  for (const id of Object.keys(cfg.video)) result[id] = {};
+  const result: Record<string, { models?: string[]; disabled?: boolean }> = {};
+  for (const [id, entry] of Object.entries(cfg.video)) {
+    result[id] = {};
+    if (entry.models && entry.models.length > 0) result[id].models = entry.models;
+  }
   for (const id of cfg.disabled.video) result[id] = { disabled: true };
   return result;
 }
--- a/lib/store/settings.ts
+++ b/lib/store/settings.ts
@@ -1410,7 +1410,7 @@ export const useSettingsStore = create<SettingsState>()(
               asr: Record<string, { disabled?: boolean }>;
               pdf: Record<string, Record<string, never>>;
               image: Record<string, { models?: string[]; disabled?: boolean }>;
-              video: Record<string, { disabled?: boolean }>;
+              video: Record<string, { models?: string[]; disabled?: boolean }>;
               webSearch: Record<string, { disabled?: boolean }>;
               generation?: { parallelSceneConcurrency?: number };
             };
@@ -1559,6 +1559,12 @@ export const useSettingsStore = create<SettingsState>()(
                     ...newImageConfig[key],
                     isServerConfigured: !info.disabled,
                     serverDisabled: info.disabled === true,
+                    ...(info.models?.length
+                      ? {
+                          customModels: info.models.map((id) => ({ id, name: id })),
+                          replaceBuiltInModels: true,
+                        }
+                      : {}),
                   };
                 }
               }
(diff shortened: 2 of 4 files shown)


CXuPercy (approved):
Verified locally against the PR head (worktree + both touched test files, 191/191). This is the complete fix — I traced the full pipeline on main to confirm both gaps:

- **Server-side**: `getServerVideoProviders()` omitting `models` is real (the image counterpart already exposes it), and this hunk mirrors that implementation exactly.
- **Client-side** (the gap that actually blocks users): on main, `fetchServerProviders()` reads only `disabled` and discards `models` for both image and video, so an operator's allowlist never reaches the model pickers regardless of what the server returns. Mapping the server models onto `customModels` + `replaceBuiltInModels` is the right call — it's the exact mechanism the Token Plan flow (`apply-token-plan.ts`) already uses, and `media-popover.tsx` consumes those fields to swap the list.

End-to-end it closes: the picker now offers the operator-pinned IDs, and `resolveVideoModel()` server-side already validates against the pinned list (falling back to its first entry). The server-precedence semantics — server models replacing a user's custom list for a server-managed provider — also matches the established merge discipline in this store (mirroring the TTS merge). Tests cover both provider kinds on both sides plus the no-models no-clobber edge.

One tiny non-blocking nit: consider updating the function's doc comment to say "allowed models only, no base URLs … mirroring the image listing" — the legacy text still describes the old shape. Thanks for the clean fix.

CXuPercy:
@wyuc Approved from my side — over to you to confirm and merge. (Do check the #1299 close note too: same issue, superseded by this PR.)

wyuc (approved):

PR #1309, opened 2026-08-31
fix(workbench): render LaTeX in assistant messages

### LINKED ISSUE #1301: [Bug]: Pro Workbench assistant messages render LaTeX as raw text
## Bug Description

In the Pro Workbench (`/workspace`), assistant-message Markdown does not render LaTeX math. Inline math such as `$x^2$` and block math such as `$$\frac{a}{b}$$` are displayed verbatim in the chat transcript.

This is specific to the conversational workbench. Generated slides, whiteboards, and quizzes use separate KaTeX rendering paths and can render formulas, so the same course-making flow behaves inconsistently.

## Steps to Reproduce

1. Open the Pro Workbench at `/workspace`.
2. Start or continue an agent conversation.
3. Ask the assistant to return:

   ```markdown
   Inline: $x^2$

   $$
   \frac{a}{b}
   $$
   ```

4. Observe the assistant message in the transcript.

A deterministic renderer-level reproduction on current `main` produces ordinary paragraph text containing the delimiters:

```html
<p>inline $x^2$ and block:</p>
<p>$$\frac{a}{b}$$</p>
```

## Expected Behavior

Inline and block LaTeX in assistant messages should render as formatted math, consistent with the rest of OpenMAIC's course-authoring surfaces.

## Actual Behavior

The LaTeX source and its `$` / `$$` delimiters remain visible as plain text. No KaTeX math DOM is produced.

## Deployment Method

Local development (`pnpm dev` / renderer-level reproduction)

## OpenMAIC Version

`main` at `9d16e68296dbfdf6002e0572fef98ea3689fee29`

## Affected Area

Other — Pro Workbench / multi-agent chat

## Browser

Not browser-specific; reproduced at the React/Streamdown renderer level.

## Operating System

macOS 26.6.2

## Relevant Logs

No runtime error is emitted. The assistant-message rendering chain ends at `TextBlock`, which configures Streamdown with CJK plus its default remark plugins but does not register a math plugin:

- [`components/workbench/chat/text-block.tsx`](https://github.com/THU-MAIC/OpenMAIC/blob/9d16e68296dbfdf6002e0572fef98ea3689fee29/components/workbench/chat/text-block.tsx#L22-L76)
- [`components/workbench/chat/chat-timeline.tsx`](https://github.com/THU-MAIC/OpenMAIC/blob/9d16e68296dbfdf6002e0572fef98ea3689fee29/components/workbench/chat/chat-timeline.tsx#L279-L287)

The root layout already imports KaTeX CSS, but CSS alone does not parse math markup:

- [`app/layout.tsx`](https://github.com/THU-MAIC/OpenMAIC/blob/9d16e68296dbfdf6002e0572fef98ea3689fee29/app/layout.tsx#L4-L8)

## Additional Context

Git history suggests this is an initial Workbench integration gap rather than a recent regression: the math plugin has not been wired into `TextBlock` since the component was introduced. There is also no focused test covering LaTeX rendering in assistant messages.

Suggested acceptance criteria:

- Inline `$...$` and block `$$...$$` math render in both streaming and settled assistant messages.
- Existing CJK, GFM, links, and fenced-code rendering remain unchanged.
- A focused `TextBlock` rendering test prevents regression.

YizukiAme:
## Summary

Render inline and display LaTeX in Pro Workbench assistant messages through the official Streamdown math plugin, while keeping ordinary currency and shell-variable dollars as literal text.

## Related Issues

Fixes #1301

## Changes

- keep Streamdown's official `remark-math` pipeline for display math and mdast conversion
- validate single-dollar candidates inside the same micromark parser, rolling non-formulas back to ordinary Markdown text
- preserve unescaped currency and shell variables such as `$5`, `$HOME`, and `$OPENAI_API_KEY` in settled and streaming messages
- recover later formulas after literal dollars without rewriting source or reparsing the document
- support standard multiline `$$` display fences in streaming and settled messages
- disable only generic KaTeX delimiter completion so dollar signs inside fenced code remain literal
- include the math plugin in Tailwind source discovery and contain wide formulas with horizontal scrolling
- preserve KaTeX inside unresolved workspace course-link labels without overriding their text/MathML accessible name
- add focused regressions for currency, shell syntax, mixed literal/formula messages, escapes, Markdown/GFM boundaries, parser padding, and long-message performance

Block math follows standard `remark-math` syntax: the opening and closing `$$` fences are on their own lines, as in the issue reproduction. Same-line `$$...$$` still renders as math using upstream inline semantics.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Run `pnpm install --frozen-lockfile` with Node 22.
2. Run `pnpm test`.
3. Open a Pro Workbench session whose assistant response contains `$x^2$`, `$$...$$`, ordinary prices, and shell variables.

### Verified locally on the final tree

- Focused renderer/lifecycle coverage: 96/96 tests passed.
- Full suite: 653 test files passed (9 skipped); 7,315 tests passed (26 skipped).
- Prettier, TypeScript, i18n alignment, package-version, frozen-lockfile, and diff checks pass.
- ESLint reports 0 errors and the existing 18 warnings.
- Production build completes; its three existing Turbopack warnings remain unchanged.
- Final independent adversarial review found no blocking issues; 10,000 randomized parser inputs completed without exceptions or invalid positions.
- A 52 KB price-heavy input parses in about 11 ms in the parser benchmark, with near-linear scaling.

### Visual evidence

| Before — raw LaTeX | After — rendered KaTeX |
| --- | --- |
| Pro Workbench before LaTeX rendering fix | Pro Workbench after LaTeX rendering fix |

- [x] Manually tested locally
- [x] Screenshots attached

## Checklist

- [x] I have performed a self-review
- [x] I have added or updated documentation where needed

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -8,6 +8,7 @@
    Tailwind must scan its dist so those classes get generated. */
 @source '../node_modules/streamdown/dist/*.js';
 @source '../node_modules/@streamdown/code/dist/*.js';
+@source '../node_modules/@streamdown/math/dist/*.js';
 @source "../node_modules/@openmaic/renderer/dist/**/*.{js,cjs}";
 /* slide-renderer-demo is gitignored (local sandbox), which makes Tailwind v4's
    auto source detection skip it. Re-include it explicitly so the demo's
--- a/components/workbench/chat/course-link.tsx
+++ b/components/workbench/chat/course-link.tsx
@@ -60,7 +60,7 @@ export function CourseLink({
    * workspace's own name for the course wins when it has one — the rail, the
    * tab and this link must not disagree about what a course is called.
    */
-  readonly label?: string;
+  readonly label?: React.ReactNode;
   /**
    * What to render with no workspace around us: the classic workbench hosts
    * this same timeline and has no right pane to open anything into, so an
@@ -74,12 +74,19 @@ export function CourseLink({
   if (!navigation) return <>{fallback ?? label ?? courseId}</>;
 
   const summary = navigation.lookupCourse(courseId);
-  const name = summary?.name || label || t('workspace.untitledCourse');
+  const fallbackName = t('workspace.untitledCourse');
+  const name = summary?.name || label || fallbackName;
   const pageCount = summary?.pageCount ?? null;
   // One sentence, whatever the pane is currently showing — the press is the
   // same in every case, so describing it three ways only invited the reader to
   // look for a difference that is not there.
-  const hint = t('workspace.courseLinkHint', { name });
+  const richLabel = label != null && typeof label !== 'string';
+  // Rich labels carry their own accessible content (including KaTeX MathML).
+  // Do not replace it with an inaccurate "untitled" accessible name.
+  const hint =
+    richLabel && !summary?.name
+      ? undefined
+      : t('workspace.courseLinkHint', { name: summary?.name || label || fallbackName });
 
   return (
     <button
--- a/components/workbench/chat/text-block.tsx
+++ b/components/workbench/chat/text-block.tsx
@@ -21,191 +21,19 @@
  *    there is no right pane — the pill falls back to the plain anchor.
  */
 import { createMathPlugin } from '@streamdown/math';
-import { Streamdown, defaultRemarkPlugins, type StreamdownProps } from 'streamdown';
+import { Streamdown, defaultRemarkPlugins } from 'streamdown';
 import remarkCjkFriendly from 'remark-cjk-friendly';
 import { courseIdFromHref } from '@/lib/workbench/course-link';
 import { CourseLink } from './course-link';
 
-interface MarkdownNode {
-  type: string;
-  value?: string;
-  meta?: string | null;
-  data?: Record<string, unknown>;
-  position?: {
-    start: { line?: number; column?: number; offset?: number };
-    end: { line?: number; column?: number; offset?: number };
-  };
-  children?: MarkdownNode[];
-}
-
-function displayMathData(value: string): Record<string, unknown> {
-  return {
-    hName: 'pre',
-    hChildren: [
-      {
-        type: 'element',
-        tagName: 'code',
-        properties: { className: ['language-math', 'math-display'] },
-        children: [{ type: 'text', value }],
-      },
-    ],
-  };
-}
-
-function nodeSource(node: MarkdownNode, source: string): string | undefined {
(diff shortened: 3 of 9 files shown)


LING-6150 (changes requested):
Thanks , the LaTeX rendering direction is correct, but enabling singleDollarTextMath introduces a content regression in ordinary assistant messages.
Production-shaped TextBlock rendering currently interprets inputs such as Plans cost $5 and $10 per month., Set $HOME and $PATH..., and Use $OPENAI_API_KEY or $ANTHROPIC_API_KEY. as KaTeX, in both settled and streaming states. For example, the first case turns 5 and  into math and removes the dollar delimiters.
The current regression test only covers explicitly escaped currency (\$5), but assistant output cannot be assumed to escape literal dollar signs. The repository already treats ordinary $5 and $7 as non-math in tests/quiz/math-text.test.ts.
Could you add disambiguation at the Workbench parsing boundary and regression coverage for ordinary currency and shell variables, while preserving $x^2$ inline math? Requesting changes for this content-correctness regression.  
There is already a repository precedent for the expected behavior: lib/quiz/math-text.ts supports $x^2=4$ while preserving ordinary text such as Cost is $5 and $7; both cases are covered in tests/quiz/math-text.test.ts. The Workbench uses the remark/micromark pipeline rather than that custom scanner, so this is a behavioral reference, not necessarily code that can be reused directly.

YizukiAme:
Thanks for catching this — fixed in `4cfeff42`. The Workbench now validates single-dollar candidates inside micromark itself: rejected currency and shell spans roll back to ordinary Markdown, while `$x^2$` and later formulas still render. I added settled/streaming regressions for all three reported examples plus mixed Markdown, GFM, and escape boundaries; focused 96/96 and the full 7,315-test suite pass locally. Ready for re-review.

LING-6150:
Screenshot 2026-09-01 at 4 04 31 PM


Re-reviewed on 4cfeff42. The original finding is closed.
I independently verified the updated implementation at both the renderer and browser boundaries. Ordinary currency and shell variables remain literal, inline and display LaTeX render correctly in light and dark mode, and wide display formulas scroll horizontally without causing page-level overflow.
Focused tests pass 96/96, additional adversarial cases pass, and CI is green. Screenshot attached.
No remaining findings. Approving. CC @wyuc for UI/UX visibility.

LING-6150 (approved):

PR #1311, opened 2026-08-31
fix(settings): turn TTS and media generation on from the Settings page

### LINKED ISSUE #1288: [Bug]: TTS/image/video enable flags are unreachable — one-shot auto-config plus a never-mounted control

official-burak:
## Summary

The global `ttsEnabled` / `imageGenerationEnabled` / `videoGenerationEnabled` flags only auto-set on the first provider sync. After that, Settings could store a working API key (Test TTS plays) while the flags stayed off, so course generation skipped narration and media.

This puts the existing enable switches on the Settings pages and turns each flag on when that provider first becomes usable (empty credential to a real key). Later edits do not force the flag back on if the user turned it off.

A provider configured server-side only after that first sync still does not auto-flip the flag. The Settings toggle is the recovery for that case.

## Related Issues

Closes #1288

## Changes

- Add Settings toggles for TTS, image generation, and video generation (existing i18n strings).
- When a hosted TTS / image / video provider first becomes usable, turn the matching global flag on. Image and video use `isUsableMediaProvider` (force-disabled providers stay off; keyless ones can come up from a baseUrl).
- Cover the store behavior in `tests/store/settings-server-sync.test.ts`.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Open the app once with no TTS/image/video server providers so first-run auto-config is consumed.
2. Settings, Speech Synthesis: paste a Qwen (or other hosted) TTS key and click Test TTS. Confirm it plays.
3. Generate a course and play it. Before this change, slides advance with no narration. After it, speech audio is generated.
4. Confirm Settings now has an Enable Text-to-Speech switch, and the same pattern for image and video.

- `pnpm exec vitest run tests/store/settings-server-sync.test.ts tests/store/settings-validation.test.ts` (128 passed), including: key paste turns the flag on, a force-disabled media provider gaining a key does not, a keyless provider with a baseUrl does, empty key and browser-native TTS do not, later key edits do not re-enable after the user turns the flag off, and a later server sync still does not force-enable a user-disabled image flag.
- I did not run the full app UI against a live TTS provider in this environment.

### Evidence

Store tests in `tests/store/settings-server-sync.test.ts` (`settings media enable flags (#1288)`).

## Checklist

- [x] I have tested these changes locally
- [x] I have linked related issues
- [ ] I have updated documentation (if needed)
- [x] I have added/updated tests (if applicable)
- [x] My code follows the project's style guidelines
- [x] I have considered the impact on existing features (playback, generation, export, i18n, etc.)
- [ ] I have provided screenshots or recordings (for UI changes)

### DIFF
--- a/components/settings/image-settings.tsx
+++ b/components/settings/image-settings.tsx
@@ -102,11 +102,17 @@ export function ImageSettings({ selectedProviderId }: ImageSettingsProps) {
   const requiresApiKey = currentProvider?.requiresApiKey ?? true;
 
   const handleApiKeyChange = (apiKey: string) => {
-    setImageProviderConfig(selectedProviderId, { apiKey });
+    setImageProviderConfig(selectedProviderId, {
+      apiKey,
+      ...(apiKey.trim() ? { enabled: true } : {}),
+    });
   };
 
   const handleBaseUrlChange = (baseUrl: string) => {
-    setImageProviderConfig(selectedProviderId, { baseUrl });
+    setImageProviderConfig(selectedProviderId, {
+      baseUrl,
+      ...(baseUrl.trim() ? { enabled: true } : {}),
+    });
   };
 
   const handleTest = async () => {
@@ -188,7 +194,11 @@ export function ImageSettings({ selectedProviderId }: ImageSettingsProps) {
             {t('settings.imageGenerationDisabledHint')}
           </p>
         </div>
-        <Switch checked={imageGenerationEnabled} onCheckedChange={setImageGenerationEnabled} />
+        <Switch
+          checked={imageGenerationEnabled}
+          onCheckedChange={setImageGenerationEnabled}
+          aria-label={t('settings.enableImageGeneration')}
+        />
       </div>
 
       {/* Server-configured notice */}
--- a/components/settings/image-settings.tsx
+++ b/components/settings/image-settings.tsx
@@ -4,6 +4,7 @@ import { useState, useCallback, useMemo, useEffect } from 'react';
 import { Label } from '@/components/ui/label';
 import { Input } from '@/components/ui/input';
 import { Button } from '@/components/ui/button';
+import { Switch } from '@/components/ui/switch';
 import { Dialog, DialogContent, DialogTitle, DialogDescription } from '@/components/ui/dialog';
 import { useI18n } from '@/lib/hooks/use-i18n';
 import { useSettingsStore } from '@/lib/store/settings';
@@ -37,6 +38,8 @@ export function ImageSettings({ selectedProviderId }: ImageSettingsProps) {
 
   const imageModelId = useSettingsStore((state) => state.imageModelId);
   const imageProvidersConfig = useSettingsStore((state) => state.imageProvidersConfig);
+  const imageGenerationEnabled = useSettingsStore((state) => state.imageGenerationEnabled);
+  const setImageGenerationEnabled = useSettingsStore((state) => state.setImageGenerationEnabled);
   const _setImageModelId = useSettingsStore((state) => state.setImageModelId);
   const setImageProvider = useSettingsStore((state) => state.setImageProvider);
   const setImageProviderConfig = useSettingsStore((state) => state.setImageProviderConfig);
@@ -178,6 +181,16 @@ export function ImageSettings({ selectedProviderId }: ImageSettingsProps) {
 
   return (
     <div className="space-y-6 max-w-3xl">
+      <div className="flex items-center justify-between rounded-lg border border-border/60 bg-background px-3 py-2.5">
+        <div className="min-w-0 pr-3">
+          <p className="text-sm font-medium">{t('settings.enableImageGeneration')}</p>
+          <p className="text-[11px] text-muted-foreground">
+            {t('settings.imageGenerationDisabledHint')}
+          </p>
+        </div>
+        <Switch checked={imageGenerationEnabled} onCheckedChange={setImageGenerationEnabled} />
+      </div>
+
       {/* Server-configured notice */}
       {isServerConfigured && (
         <div className="rounded-lg border border-blue-200 bg-blue-50 dark:border-blue-800 dark:bg-blue-950/30 p-3 text-sm text-blue-700 dark:text-blue-300">
--- a/components/settings/tts-settings.tsx
+++ b/components/settings/tts-settings.tsx
@@ -91,6 +91,8 @@ export function TTSSettings({ selectedProviderId }: TTSSettingsProps) {
   const ttsVoice = useSettingsStore((state) => state.ttsVoice);
   const ttsSpeed = useSettingsStore((state) => state.ttsSpeed);
   const setTTSSpeed = useSettingsStore((state) => state.setTTSSpeed);
+  const ttsEnabled = useSettingsStore((state) => state.ttsEnabled);
+  const setTTSEnabled = useSettingsStore((state) => state.setTTSEnabled);
   const ttsProvidersConfig = useSettingsStore((state) => state.ttsProvidersConfig);
   const setTTSProviderConfig = useSettingsStore((state) => state.setTTSProviderConfig);
   const activeProviderId = useSettingsStore((state) => state.ttsProviderId);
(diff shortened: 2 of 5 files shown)


CXuPercy (changes requested):
Cross-reviewed by two independent reviewers. **The architecture here is the right one** — edge-triggered enablement inside the config setters (usable-transition detection with `shouldTurnOn`, "do not re-enable on later edits if the user turned the flag off") is exactly what the competing approach to #1288 got wrong, and it shows: your regression tests prove user-off survives provider edits and server syncs, CI is green, and all six i18n keys you reference already exist on main. Two correctness gaps need fixing before merge, both found by the second reviewer and verified against main:

**1. Route the image/video edge through `isUsableMediaProvider` (blocking).** The TTS side correctly uses `isTTSProviderEnabled()` for old/new usability, but the image/video blocks inline a narrower credential check:

```ts
const hadImageCredential = !!(state.imageProvidersConfig[providerId]?.apiKey || ...?.isServerConfigured);
const hasImageCredential = !!(mergedProvider.apiKey || mergedProvider.isServerConfigured);
```

This misses two cases the existing helper (same file, `isUsableMediaProvider`) already handles:
- **`enabled: false`**: a provider the user force-disabled in its own config still counts as "gaining a credential" here, so adding a key to it flips the global flag on — the helper returns false for that case.
- **Keyless providers**: `comfyui-image` and `lemonade` are `requiresApiKey: false` and become usable through a `baseUrl` alone; your check never fires for them, so configuring those from Settings reproduces the exact #1288 symptom for the keyless family.

The fix is mechanical: compute `wasUsable`/`nowUsable` via `isUsableMediaProvider(IMAGE_PROVIDERS[providerId], oldConfig)` / `(…, mergedProvider)` (and the video equivalent), keeping your `shouldTurnOn` semantics.

**2. Test the new cases**: a disabled (`enabled: false`) media provider gaining a key must NOT turn the flag on; a baseUrl-only keyless provider becoming configured SHOULD. Also worth one test for whitespace-only credentials if you want the edge to be robust.

Minor, non-blocking: the three new switches have no explicit accessible name — an `aria-label` (or an associated `<Label>`) would give screen readers something to announce.

The uncovered path (a provider configured server-side *after* a user's first sync doesn't auto-flip the flag) is acceptable as you've scoped it — the toggle is now reachable on the Settings page, which is the recovery the issue asks for — but it's worth a line in the PR description so maintainers know the boundary.

official-burak:
Image and video now go through `isUsableMediaProvider`, same usable-transition as TTS. A force-disabled provider that gets a key stays off, and lemonade/comfyui turn the flag on from a baseUrl. Added tests for those plus whitespace-only credentials, and named the three Settings switches. A provider that only shows up server-side after the first sync still does not auto-flip; the Settings toggle is the recovery there.

CXuPercy (approved):
The follow-up addresses everything from my review, verified locally: image/video enablement now routes through `isUsableMediaProvider` for the old/new usability edge, the helper itself was hardened (whitespace-only credentials no longer count; keyless providers like comfyui-image/lemonade become usable through a baseUrl — both gaps from my review are covered), the three switches carry `aria-label`s, and the branch adds 121 lines of behavior tests. I ran the full settings-server-sync suite against this head: 91/91, including the regression tests the competing approach to #1288 failed — user-disabled flags survive provider edits and server syncs, and first-time credentials turn the flags on. CI green.

One scope note for the record, not a change request: tightening `isUsableMediaProvider` also affects the disable-fallback path (a keyless provider with no configured baseUrl no longer counts as a fallback candidate). That direction is more honest — the old behavior could fall back to a localhost default that isn't running — but it's a behavior shift beyond #1288 that the added tests now pin.

With #1291 closed, this is the only carrier for #1288.

Issue #1305, opened 2026-08-31
generate_scene hides non-PBL content-generation failure causes


YizukiAme:
## Problem

The agent-facing `generate_scene` tool collapses every non-PBL content-generation `null` into the same error:

```text
Page content generation failed; nothing was written.
```

The result has `details: {}`, so the agent cannot distinguish a prompt construction/loading failure from a model response that could not be parsed into the expected page content, such as malformed slide/quiz JSON or an interactive response with no extractable HTML. This prevents evidence-based recovery and can lead the agent to retry the same page while guessing at the cause.

The PBL path already returns structured `error`, `statusCode`, and `cause` fields, so the two generation paths expose materially different diagnostics.

## Root cause

- The slide, quiz, and interactive generation helpers return `null` for several distinct failures, including prompt construction/loading and model-output parsing or extraction failures.
- `lib/server/agent-runtime/generation-tools.ts` maps every falsy `content` returned by `generateSceneContent` to the generic message above with empty details.
- This tool path does not inject a generation logger, so the package defaults to `noopGenerationLogger`; the lower-level reason is unavailable both to the agent and in server logs.

## Expected behavior

Keep the no-partial-write behavior, but return a safe structured failure classification for valid non-PBL generation requests, sufficient to distinguish at least:

- prompt construction/loading failure;
- model output that is missing or malformed for the expected page type, including invalid slide/quiz payloads and interactive output with no extractable HTML.

Include stable context such as page order, title, and scene type. Raw model output should remain server-only and sanitized.

Add focused tests across representative slide, quiz, and interactive failure paths proving that they no longer collapse to `details: {}`.

YizukiAme:
Working on it.

PR #1319, opened 2026-08-31
fix(agent): bound generate_scene retries per page

### LINKED ISSUE #1308: generate_scene can retry the same failing page indefinitely
## Problem

In the durable server workbench, a failed `generate_scene` result returns control to the model. If the model calls `generate_scene` again for the same semantic page target, nothing keys the attempt to that target: the runtime holds no per-page consecutive-failure state, and the only guidance that mentions retrying lives in `curriculum-planner`, which a single stage build never loads and which is written per lesson rather than per page.

This lets the agent keep retrying the same `stageId + order` while changing the brief and guessing at the cause. In the observed run, page 4 failed four consecutive times with:

```text
Page content generation failed; nothing was written.
```

The agent said it would retry once, then kept retrying the same page after the second failure. Each call consumed generation time and made no durable progress.

That run predates #1316; the message above is now the unclassified fallback only, and `prompt-unavailable` / `invalid-model-output` are reported distinctly.

## Why nothing stops it today

- `lib/agent/runtime/build-agent.ts` normalizes a tool result with `isError: true`, but an error does not terminate the Pi loop unless an `afterToolCall` hook also returns `terminate: true`.
- `lib/server/agent-runtime/runner.ts` counts tool calls, but its request-scoped hook only terminates after a successful `ask_user`; the count is diagnostic.
- The runner's `maxAttempts` limits unattended session starts/resumptions, not repeated tool calls inside one live run.
- Retrying with a new tool-call ID or a modified brief is not a protocol duplicate, even when it targets the same persisted page position.

## Expected behavior

`generate_scene` recovery is page hygiene on the build floor, so the rule belongs in `stage-design`: every stage build goes through it, including a single classroom outside a series.

- Identity is `stageId + order`. Changing the title or brief, or generating another page in between, does not reset the attempt.
- `prompt-unavailable`, and other deterministic failures: do not repeat the same call.
- `invalid-model-output`: one retry, with a changed brief that keeps the settled teaching intent. A clearly transient provider failure: one retry.
- Once that budget is exhausted: leave the page unwritten, continue with later pages **in the same stage**, and name the gap when wrapping up. Do not `ask_user` after every failed page, and do not silently drop the page from the settled plan.
- Reconcile against `list_scenes` before reporting completion. A stage with a missing page can be wrapped up, but it is not done.

The gap is a missing page inside one stage, and filling it later is an ordinary follow-up: `generate_scene` on that order.

`curriculum-planner` needs one line delegating page recovery to `stage-design`, plus the fact that a failed page does not abort the rest of the series.

## Out of scope

- Runtime per-tool counters — the approach in #1319, closed.
- A generic run-level budget. If a backstop is still wanted once the skill text exists, it should be run-level and generic rather than a per-tool rule, and tracked separately from this issue.

## Related

- #1305 — the failure cause `generate_scene` reports; landed via #1316. This was the prerequisite: without a cause, the model could not tell a bad brief from an unavailable prompt, which is why it retried near-identical calls.
- #1370 — implementation.

Scope settled in [this comment](https://github.com/THU-MAIC/OpenMAIC/issues/1308#issuecomment-5538743323) and [the one before it](https://github.com/THU-MAIC/OpenMAIC/issues/1308#issuecomment-5528281602); this description was rewritten to match. The original "Expected behavior" asked for enforcement in the runtime.

hichipli:
## Summary

Bound repeated `generate_scene` failures for the same page within one server-workbench run. The initial generation and one retry may execute; further consecutive calls for the same `stageId + order` are rejected before reaching the generation provider.

## Related Issues

Fixes #1308.

## Changes

- add a request-scoped retry budget keyed by the semantic page target (`stageId + order`)
- preserve the original failure content/details while adding structured retry state and clear recovery guidance
- block calls after the second consecutive failure, before another generation request can run
- reset the budget after a successful generation or when generation moves to another page
- keep unrelated tool calls outside the budget and keep the shared tool allowlist authoritative
- add focused policy, Agent-hook, and runner-wiring regression coverage

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Run the focused retry-policy, Agent-hook, and runner-wiring tests.
2. Run the complete agent-runtime test directories under Node 22.
3. Run the complete repository test suite under Node 22.
4. Run formatting, lint, and TypeScript checks.

- the first failed page generation leaves exactly one real retry available
- a second consecutive failure returns exhausted-state guidance and a third call is blocked before tool execution
- a successful generation resets the budget
- moving to a different page target resets the budget
- unrelated tools neither consume nor reset the active page budget
- existing error details are preserved when retry metadata is added
- the shared allowlist still runs before the new request-scoped pre-call hook
- Node 22 full suite: 652 test files passed, 9 skipped; 7,170 tests passed, 81 skipped
- Node 22 agent-runtime suites: 85 test files passed, 4 skipped; 1,044 tests passed, 11 skipped
- local lint completed with 0 errors; 18 unrelated baseline warnings remain

### Evidence

- Before: every new tool-call ID for the same failing page could reach generation again.
- After: only the initial attempt and one retry can execute consecutively for a page; later same-target calls return a deterministic blocked result while later page targets remain available.
- [x] Manually verified locally with focused and complete automated suites
- [ ] Screenshots / recordings attached (no UI changes)

## Checklist

### DIFF
--- a/lib/agent/runtime/build-agent.ts
+++ b/lib/agent/runtime/build-agent.ts
@@ -4,8 +4,8 @@
  * Stands up a pi `Agent` with:
  * - injected StreamFn (-> OpenMAIC connector),
  * - request-scoped tools supplied by the route,
- * - a caller-supplied `beforeToolCall` allowlist gate,
- * - a `afterToolCall` quota hook (v0 stub: unlimited).
+ * - the shared allowlist gate plus an optional request-scoped pre-call policy,
+ * - the shared quota hook (v0 stub: unlimited) plus an optional post-call policy.
  */
 import {
   Agent,
@@ -14,6 +14,8 @@ import {
   type AgentMessage,
   type AgentOptions,
   type AgentTool,
+  type BeforeToolCallContext,
+  type BeforeToolCallResult,
   type StreamFn,
 } from '@earendil-works/pi-agent-core';
 import type { Api, Model } from '@earendil-works/pi-ai';
@@ -55,6 +57,11 @@ export interface BuildAgentOptions {
   transformContext?: (messages: AgentMessage[], signal?: AbortSignal) => Promise<AgentMessage[]>;
   /** Optional Pi message conversion, required when a context transform emits custom roles. */
   convertToLlm?: AgentOptions['convertToLlm'];
+  /** Optional request-scoped policy applied after the shared allowlist gate. */
+  beforeToolCall?: (
+    context: BeforeToolCallContext,
+    signal?: AbortSignal,
+  ) => Promise<BeforeToolCallResult | undefined> | BeforeToolCallResult | undefined;
   /** Optional request-scoped hook composed with the shared quota hook. */
   afterToolCall?: (
     context: AfterToolCallContext,
@@ -64,6 +71,7 @@ export interface BuildAgentOptions {
 
 export function buildAgent(opts: BuildAgentOptions): Agent {
   const quotaHook = makeQuotaHook({ remaining: () => Number.MAX_SAFE_INTEGER });
+  const allowlistGate = makeAllowlistGate(opts.allowedToolNames);
   const agent = new Agent({
     streamFn: opts.streamFn,
     transformContext: opts.transformContext,
@@ -79,7 +87,11 @@ export function buildAgent(opts: BuildAgentOptions): Agent {
       // conversation in context — without this the agent is stateless per turn.
       ...(opts.history && opts.history.length > 0 ? { messages: opts.history } : {}),
     },
-    beforeToolCall: makeAllowlistGate(opts.allowedToolNames),
+    beforeToolCall: async (context, signal) => {
+      const allowlistResult = await allowlistGate(context);
+      if (allowlistResult?.block) return allowlistResult;
+      return opts.beforeToolCall?.(context, signal);
+    },
     afterToolCall: async (context, signal) => {
       const markerIsError =
         typeof context.result === 'object' &&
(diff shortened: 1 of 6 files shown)


wyuc:
Thanks for this — the composition is right (allowlist stays authoritative, request-scoped state, blocking before the expensive call rather than terminating the run), and the coverage is thorough.

I'm not merging it, for architectural reasons rather than anything about the code. A per-tool retry policy doesn't belong in the agent runtime. The runtime is a mechanism layer; the allowlist lives there because it's a capability boundary. "This page is worth one retry, then abandon it" is task-level judgment that already has a home at the skill layer — `curriculum-planner` states that convention today. Putting it in the runtime also removes the model's discretion at a bad moment, since #1305 means it can't see *why* generation failed: a provider blip and an invalid brief are identical through `isError`. Extending shared `buildAgent` for that policy is the part I'd push back on hardest. Product questions are on #1308.

One thing you'd want to know regardless: the budget is a single slot, so "A fails twice → generate B → back to A" gives A a fresh budget — and the guidance you return tells the model to move to later pages, which is exactly what clears it. In a multi-page run nothing is actually bounded. A `Map` keyed on `stageId:order` would fix that, and would also give you the "report the missing page" part.

If you want to keep going here, #1305 is the useful next thing and a prerequisite for any policy. I'd review it promptly.

hichipli:
Thanks, I’ve moved the proposed product behavior to #1308 for alignment. I’ll keep this PR in draft and won’t continue with the per-tool runtime approach.

hichipli:
Closing this runtime-based approach to avoid parallel review. The replacement draft is #1370, limited to the curriculum-planner layer as discussed here; the remaining product behavior is being aligned in #1308.

Issue #202, opened 2026-03-21
[Feature]: Per-Scene Editing & Regeneration


Custowner:
### Problem or Motivation

After generating a classroom lesson, there is no way to edit or regenerate individual scenes. If I want to fix the text on slide 7, change an image on slide 12, or rephrase a quiz question on slide 4, my only options are:

Regenerate the entire lesson from scratch — which loses all the good content on other slides
Export to PPTX and edit manually in PowerPoint — which loses all interactive classroom features (narration, quizzes, simulations, whiteboard)

This is especially frustrating for longer lessons. I generated a 23-slide presentation on energy transition, and while 20 slides were excellent, 3 needed text adjustments and 1 needed a different image. I had to regenerate the whole thing multiple times to get all slides right, wasting tokens and time.
The current workflow forces an all-or-nothing approach that doesn't match how professionals actually create presentations: generate → review → refine specific parts → finalize.

### Proposed Solution

Proposed Solution
A per-scene editing system with three levels of control:
Level 1 — Direct Text Editing (highest priority)

Click any text element in a scene to edit it inline
Changes save to scene data and reflect in both classroom playback and PPTX export
Covers: typo fixes, rephrasing, updating numbers/names/terminology

Level 2 — AI-Powered Scene Regeneration

Add an "Edit with AI" button on each scene
User types a natural language instruction, e.g.:

"Make this slide more visual and less text-heavy"
"Replace the chart with a comparison table"
"Simplify the language for a non-technical audience"

System calls the LLM to regenerate only that one scene, using surrounding scenes as context for coherence
Suggested API: POST /api/regenerate-scene with classroomId, sceneIndex, instruction, and optional directEdits

Level 3 — Image & Layout Editing

Swap images (upload custom or request AI regeneration)
Adjust layout (e.g., single-column → two-column)
Modify visual elements (colors, chart types)

Why this is technically feasible: The Stage 2 generation pipeline in lib/generation/ already generates each scene independently. Exposing per-scene regeneration as a separate endpoint reuses the existing logic. The components/scene-renderers/ already render scenes independently, so adding an edit overlay per scene should be contained. The PPTX export in packages/pptxgenjs/ processes slides individually, so updated scenes would automatically appear in re-exports.

### Alternatives Considered

Full regeneration with seed/lock: Lock specific scenes as "approved" and only regenerate unlocked ones. Simpler than a full editor but still requires re-running the generation pipeline. Doesn't help with small text fixes.
[...]

PR #1325, opened 2026-09-01
fix(interactive): contain scrolling and regenerate code

KimYeongHyeon:
## Summary

Fixes runaway scrolling in generated interactive content and adds a host-owned recovery control that regenerates failed Python/code widgets without replacing narration or scene actions.

## Related Issues

Related to #314 and #202. No existing issue exactly matched the runaway root-scroller behavior shown in the reproduction.

## Changes

- enforce a single bounded root scroller inside generated iframes, with the viewport patch injected after authored head styles
- expose widget type to the host and add an owner-only Regenerate code control outside the sandbox
- regenerate content from the persisted outline while preserving scene identity, narration, and actions
- guard read-only/edit-locked/stale responses and localize the control in all 12 locales
- add unit, real-Chromium, and headed desktop/mobile regression coverage

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Open a generated interactive code scene whose authored CSS sets `html, body` as root scrollers and whose content exceeds the viewport.
2. Confirm scrolling stays inside the iframe and does not grow the page indefinitely.
3. Click Regenerate code and confirm only the code-widget content is replaced.

- 28 targeted unit tests passed.
- Real Chromium root-scroll regression passed.
- Headed Playwright passed at 1280x900 and 390x844, including scroll ownership, regeneration, clean page console/API responses, and before/after screenshots.
- Full `tsc --noEmit`, targeted ESLint/Prettier, and 12-locale key alignment passed.
- Read-only, edit-lock, missing/stale generation boundaries are covered by hard gates/tests.

### Evidence

- [x] Manually tested locally

## Checklist

### DIFF
--- a/components/scene-renderers/InteractiveIframeHost.tsx
+++ b/components/scene-renderers/InteractiveIframeHost.tsx
@@ -17,6 +17,10 @@ import { intersectClientBoxes } from '@/lib/edit/visible-client-rect';
 import { useCanvasStore } from '@/lib/store/canvas';
 import { useElementRefsStore } from '@/lib/store/element-refs';
 import { useI18n } from '@/lib/hooks/use-i18n';
+import { RefreshCw } from 'lucide-react';
+import { toast } from 'sonner';
+import { regenerateCodeSceneContent } from '@/lib/interactive/regenerate-code-scene';
+import { useStageStore } from '@/lib/store/stage';
 import {
   ELEMENT_REF_SELECTOR_MAX,
   ELEMENT_SNAPSHOT_MAX,
@@ -155,11 +159,13 @@ interface PooledIframeProps {
  */
 function PooledIframe({ sceneId, entry, visible }: PooledIframeProps) {
   const { t } = useI18n();
+  const [regeneratingCode, setRegeneratingCode] = useState(false);
   const iframeRef = useRef<HTMLIFrameElement>(null);
   const registerIframe = useWidgetIframeStore((s) => s.registerIframe);
   const getSendMessage = useWidgetIframeStore((s) => s.getSendMessage);
   const pickTarget = useCanvasStore.use.pickTarget();
   const refs = useElementRefsStore.use.refs();
+  const canRegenerate = useStageStore((s) => s.isOwner && !s.readOnly);
   const armed = pickTarget?.purpose === 'element-ref' && pickTarget.sceneId === sceneId;
   const selectors = useMemo(
     () =>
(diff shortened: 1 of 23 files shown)


KimYeongHyeon:
Automated Codex review status: unavailable. The `chatgpt-codex-connector` replied that this GitHub account must be connected to Codex before it can review, so no code finding or severity was emitted. This is an external authentication limitation, not a deferred code finding. All four required GitHub Actions checks passed, and the branch also completed the local Chromium/headed regression checks documented in the PR body.

CXuPercy (changes requested):
Cross-reviewed by two independent reviewers. The scrolling containment design and the E2E coverage (real editor drive, dual viewport, scroll-metric assertions) are strong, and the recovery-control concept fits the failure surface well. One verified race needs fixing before merge:

**A late regeneration response can overwrite newer user edits.** The guard after `fetchContent` resolves checks stage id, `generationEpoch`, and scene existence — but `updateScene` does not increment `generationEpoch` (on main it only bumps on stage load/clear and the explicit `bumpGenerationEpoch` action). So while a regeneration is pending, any user edit to the same scene (or a second regeneration completing first) passes the guard, and `after.updateScene(sceneId, { content })` replaces the newer content wholesale. Reproduce: start a code regeneration, edit the scene before it resolves, observe the edit being lost when the response lands.

Suggested fix: capture a scene revision/content identity when the request starts and compare before applying, or invalidate in-flight regenerations on scene updates (bumping a per-scene token on `updateScene` would also serve). Please add a regression test for the concurrent-edit case.

The UI-side capability addition (owner-gated regenerate control) is acceptable scope from our side — this RC is only about the race.

CXuPercy (changes requested):
Local end-to-end pass completed on the reworked branch (46e110ad). The race fix is verified good — the reference-equality guard on the scene object is the right shape, and the concurrent-edit regression test pins it. The scrolling containment also worked as intended locally.

However, the pass surfaced **three blocking UX problems** with the regenerate control — as an actual user in the playback UI, its semantics cannot be understood from the interface itself:

**1. The label "Regenerate code" / 「重新生成代码」 is ambiguous (blocking).** Regenerate *what* code? The user never wrote any code — the button acts on AI-generated widget content. The naming is written from the implementation's perspective (`widgetType === 'code'`), not the user's intent. An intent-oriented label ("Redo this activity" / 「重做此互动页」 or similar) is needed, and the copy must state the actual scope: the interactive content is rewritten from the original outline while narration and page placement are preserved — that preservation guarantee is the feature's core selling point and currently appears nowhere in the UI.

**2. No scope information and no confirmation (blocking).** One click spends an LLM call and wholesale-replaces the widget content, with no tooltip, no hover hint, and no confirmation. Please add a tooltip stating scope at minimum, and a lightweight confirm (or an explicit "replacing current interactive content" step) on click.

**3. Mode mismatch (blocking).** The button appears in **playback mode** on the widget corner, but "regenerate" is an editing action — users in playback have no mental model for destructive regeneration appearing there. Either move the control into edit mode, or demote it in playback (e.g. behind a menu) so the playback surface stays read-only in feel.

These are blocking because the feature ships a new user-facing affordance whose meaning, cost, and consequences are invisible at the point of use. Happy to re-review — the underlying regeneration machinery itself is in good shape.

wyuc:
Thanks @KimYeongHyeon for this contribution. I'm closing it to keep the queue tidy, since there hasn't been an update since the changes requested on 2026-09-07. It isn't a rejection of the idea: if you'd like to pick it back up, you're welcome to reopen it or open a fresh PR against the current `main` that addresses the review, and we'll take another look.

PR #1330, opened 2026-09-01
feat(providers): add Codex subscription provider

### LINKED ISSUE #924: [Feature]: Sign in with ChatGPT (Codex subscription) as an LLM provider
### Problem

Self-hosted OpenMAIC users may already have a ChatGPT plan with Codex usage, but currently need a separately billed OpenAI Platform API key (or another provider). A "Sign in with ChatGPT" option would let them connect the entitlement they already pay for, without managing an API key by hand.

Similar OSS projects have shipped this — [OpenCode](https://github.com/anomalyco/opencode/pull/7537) added browser PKCE and Codex Responses to core, then [added device-code](https://github.com/anomalyco/opencode/pull/10890) for headless; [OpenClaw](https://docs.openclaw.ai/providers/openai) documents Codex subscription OAuth. OpenAI officially [documents ChatGPT sign-in for Codex](https://learn.chatgpt.com/docs/auth), with device-code recommended for remote/headless.

### Proposal

An experimental provider, scoped to local / self-hosted single-user deployments:

- "Sign in with ChatGPT" action in provider settings — account status, available models, sign-out.
- Browser PKCE by default; Codex device-code as the headless fallback.
- Tokens stay server-side, refresh tokens rotate safely, sign-out clears stored credentials.
- Model list reflects what the signed-in account actually exposes; expired-session, workspace, quota, and rate-limit errors surface as clear UI states.
- Shows as unavailable on serverless / read-only deployments rather than failing at runtime.
- Existing API-key and third-party providers stay untouched.

Initial scope:

- Text generation only. Image via `gpt-image-2` has working third-party prior art but would be a separate follow-up.
- Local single-user only. No multi-tenant or public multi-user.

### Two implementation paths

**Native TypeScript** — implement PKCE, device-code, token refresh, and the Codex adapter in-repo. Follows OpenCode's direction. No extra binary; upstream protocol drift is on us to track.

**Managed [CLIProxyAPI](https://github.com/router-for-me/CLIProxyAPI) sidecar** — bundle and supervise a pinned binary. Less protocol code to own, and it already handles Codex-backed `gpt-image-2` for a possible image follow-up. Cost is binary verification, cross-platform packaging, and process lifecycle.

### Notes

Usage counts against the user's ChatGPT plan limits, not an OpenAI Platform API balance. UI should say so, and note it's subject to OpenAI's terms.

### Questions for maintainers

1. Is ChatGPT-subscription-backed sign-in an acceptable direction?
2. Preference between native adapter and managed sidecar?

white638:
## Summary

- add an opt-in, server-only `codex` language-model provider backed by the official Codex App Server
- add ChatGPT browser/device-code sign-in, sign-out, model discovery, and subscription rate-limit display
- add an AI SDK v3 adapter for text, image input, streaming, structured JSON output, reasoning effort, and usage reporting
- isolate provider state in a required dedicated `CODEX_PROVIDER_HOME`
- run every generation in an ephemeral, no-environment thread with commands, file access, plugins, MCP, web access, approvals, and dynamic tools disabled
- document persistent-Node, shared-quota, and access-control requirements

## Screenshot

Codex Subscription provider settings

## Rationale

The earlier native-adapter discussion in #949 concerned direct use of an unofficial Codex OAuth/backend path. This implementation uses OpenAI’s documented [Codex App Server](https://developers.openai.com/codex/app-server), which is specifically intended for rich-client embedding. The pinned official package owns OAuth, token refresh, model discovery, and quota reporting; OpenMAIC never parses or stores OAuth tokens.

Closes #924.

## Verification

- [x] frozen-lock dependency install
- [x] 7 focused test files, 41 tests
- [x] TypeScript (`tsc --noEmit`)
- [x] ESLint (0 errors; 18 pre-existing repository warnings)
- [x] production Next.js build
- [x] GitHub CI: lint/typecheck/unit, E2E, render service, and PostgreSQL storage contract
- [x] all 12 locale files have aligned keys
- [x] Prettier check for every changed, supported file
- [x] production `/api/codex` status route and settings UI smoke test
- [x] official App Server accepted the hardened ephemeral `thread/start` configuration
- [x] account-authenticated generation with `gpt-5.6-luna` and live ChatGPT Pro quota read

Repository-wide `prettier . --check` currently reports pre-existing formatting/line-ending differences across 2,371 files. The package-version gate also fails unchanged `HEAD` because the current publish workflow omits five packages from `on.push.paths`; this patch does not modify publishable package sources or versions.

## Deployment notes

- disabled unless `CODEX_PROVIDER_ENABLED=true`
- requires an absolute, dedicated `CODEX_PROVIDER_HOME`
- requires a persistent local Node.js server; serverless deployment is unsupported
- the ChatGPT login belongs to the server, so shared deployments must enable access control
- OpenMAIC tool calls are not exposed through this transport

## AI assistance

This contribution was developed with AI assistance. The final diff was manually reviewed, security boundaries were checked against the official App Server protocol/source, and only the verification listed above is claimed.


pascalnshoboza-create (approved):

wyq0x55:
Is it possible to use it to generate images and speech?

white638:
> Is it possible to use it to generate images and speech?

This doesn't seem to be officially permitted by Codex. You can only use it for reasoning.

wyuc:
Thank you — this is a real step up from the earlier unofficial path, and the hardening is careful: the pinned official package owns login and token refresh so OpenMAIC never touches tokens, agent capabilities are switched off rather than trusted, and tool calls fail loudly rather than degrading silently.

I'm not merging it, and the main reason isn't in the code: whether a third-party app may serve model calls from a user's ChatGPT subscription is still unanswered by OpenAI. See openai/codex#8338 — asked since May, every follow-up from developers building third-party apps unanswered, and the "known clients list" the README points at has no working intake. Your own note above, that image and speech "don't seem to be officially permitted," is the same uncertainty from the other side. More code can't settle it. Reasoning on #924.

Three things independent of that:

- `language-model.ts` throws on any tool call, and generation here is tool-driven — so this cannot drive the workbench, which is most of what a self-hoster would want it for.
- `@openai/codex` is a hard dependency pulling six platform binaries, so every install downloads a coding-agent CLI whether or not `CODEX_PROVIDER_ENABLED` is ever set.
- "shared deployments must enable access control" is a README line, not a gate. The login belongs to the server, so a multi-user deployment silently bills every user's generation to the owner's plan.

For the underlying need today: point your own Codex session at `skills/agent-runtime/` — `slide-dsl`, `stage-dsl`, `curriculum-planner`, `stage-design`. Codex keeps the capabilities this PR has to switch off, runs under your own subscription in its intended local use, and there is no integration to maintain. Documenting that workflow would be a genuinely useful contribution.

PR #1336, opened 2026-09-01
harden: LLM streaming, classroom Postgres fallback

yanyushuai:
## Summary

Reliability fixes for OpenMAIC.

### Runtime
- **ai**: stream named OpenAI-compatible providers (deepseek/qwen/kimi/glm) when `OPENAI_COMPAT_USE_STREAMING_CHAT=true`, avoiding NAT/firewall idle-timeout disconnects that surface as `Failed to process successful response`.
- **ai**: log the LLM failure `Error.cause` chain so real root causes (ECONNRESET, zod, truncated body) reach server logs.
- **classroom**: fall back to the Postgres DocumentStore to load Pro workbench courses, so pure-Postgres courses open by link instead of 404.

### Deps / local
- **deps**: patch @earendil-works/pi-agent-core for Windows path handling; add `/debug/` to .gitignore.

> Note: the cross-platform `rm -rf -> node fs.rmSync` workspace build-script change was split out of this PR (it triggered the package-version-bump gate and is unrelated to the runtime work here).

## Verification
- Root app typecheck (`tsc -p tsconfig.json --noEmit`): passes, 0 errors.
- ESLint on changed files: passes. Prettier: clean. i18n-keys: passes.
- `tests/ai/openai-provider.test.ts`: 52/52 pass.
- Note: `check-package-version-bumps` / `check-internal-dependency-ranges` still fail on a **pre-existing** `publish-packages.yml` drift (5 packages omitted from the workflow) that reproduces on clean `origin/main`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 .pnpm-debug.log*
+/debug/
 
 # env files
 .env*
--- a/app/api/classroom/route.ts
+++ b/app/api/classroom/route.ts
@@ -4,6 +4,7 @@ import { apiSuccess, apiError, API_ERROR_CODES } from '@/lib/server/api-response
 import {
   buildRequestOrigin,
   isValidClassroomId,
+  loadClassroomFromDocumentStore,
   persistClassroom,
   readClassroom,
 } from '@/lib/server/classroom-storage';
@@ -65,11 +66,20 @@ export async function GET(request: NextRequest) {
     }
 
     const classroom = await readClassroom(id);
-    if (!classroom) {
-      return apiError(API_ERROR_CODES.INVALID_REQUEST, 404, 'Classroom not found');
+    if (classroom) {
+      return apiSuccess({ classroom });
     }
 
-    return apiSuccess({ classroom });
+    // Fallback: Pro workbench (agent-runtime) courses live in the PostgreSQL
+    // DocumentStore, not `data/classrooms/*.json`. A JSON miss is NOT a
+    // "course missing" verdict — consult the store before declaring 404, so a
+    // Postgres-only course opens by link like any JSON classroom.
+    const stored = await loadClassroomFromDocumentStore(id);
+    if (stored) {
+      return apiSuccess({ classroom: stored });
+    }
+
+    return apiError(API_ERROR_CODES.INVALID_REQUEST, 404, 'Classroom not found');
   } catch (error) {
     log.error(
       `Classroom retrieval failed [id=${request.nextUrl.searchParams.get('id') ?? 'unknown'}]:`,
--- a/lib/ai/llm.ts
+++ b/lib/ai/llm.ts
@@ -275,6 +275,29 @@ export interface LLMRetryOptions {
 
 const DEFAULT_VALIDATE = (text: string) => text.trim().length > 0;
 
+/**
+ * Extract the Error.cause chain from an LLM call failure.
+ *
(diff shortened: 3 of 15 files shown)


liuxusummer (commented):
The new PostgreSQL fallback appears to bypass the existing tombstone semantics for deleted courses.

`loadClassroomFromDocumentStore()` obtains `owner_id` from `document_stages` and then calls the raw `provider.documentStore.forOwner(owner).loadDocument(id)`. However, course deletion is implemented by setting `stage_meta.deleted_at`; the corresponding `document_stages` row remains present. Since the raw `PgDocumentStore` does not consult `stage_meta`, a deleted Pro Workbench course may become readable again through `/api/classroom?id=...` when its ID is known.

Could this path resolve access through the existing `resolveStageAccess(id)` first, or otherwise reject tombstoned stages before loading the document? A regression test covering a tombstoned PostgreSQL course through the classroom fallback would also help prevent this from returning.

CXuPercy (changes requested):
Cross-reviewed by two independent reviewers. The four changes are individually sensible — the opt-in streaming extension for named OpenAI-compatible providers (with on/off tests), the cause-chain logger (cycle-guarded, depth-capped), the classroom Postgres fallback (JSON miss → DocumentStore → 404), and the Windows skills fix. Two things need resolving before merge:

**1. The pi-agent-core patch duplicates an already-approved fix.** #1297 fixes the same Windows path bug via a POSIX-normalizing subclass of NodeExecutionEnv — no dependency patching, no version coupling. The pnpm patch here pins `@earendil-works/pi-agent-core@0.78.0` exactly and must be re-verified on every bump of that package. Please drop the patch (and the patchedDependencies/package.json churn) from this PR and let #1297 carry the fix — or, if you believe patching upstream is the right long-term answer, make that case explicitly and coordinate so we don't merge both.

**2. The classroom fallback is a new read path with zero tests.** `loadClassroomFromDocumentStore` introduces a dynamic import, an owner lookup, a `forOwner` load, and an `as unknown as` shape boundary between the DSL document shapes and the classroom payload — the kind of path `tests/` exists for. At minimum: found-in-store returns the classroom, absent → null → 404, no DATABASE_URL → null, and store error → null (never throws). The streaming tests only cover deepseek; a loop over the named compatible providers would also cheaply pin the generalization.

Minor: this supersedes #1335 without explanation in the description — a line about what changed between the two would help reviewers of the future.

wyuc (changes requested):
Thanks for putting these fixes together. The LLM streaming compatibility path can stay: it is opt-in, preserves the existing non-streaming response contract, and is independent of the blockers below. The cause-chain logging change also looks reasonable.

Before merging, please address the following:

1. Remove the Windows `pi-agent-core` patch, including the `patchedDependencies` entry and related lockfile churn. This is now duplicated by #1297.
2. Rework or split out the classroom fallback. As written, the fallback can bypass the existing tombstone/access semantics, so it needs to preserve those checks and include regression coverage.
3. For the expanded OpenAI-compatible streaming path, please add parameterized coverage for every intended named provider (for example DeepSeek, Qwen, Kimi, and GLM), and verify that each accepts `stream_options.include_usage`. If support differs, gate that option by provider rather than sending it universally.

With those changes, the streaming compatibility and cause-chain logging portions are fine to keep.

wyuc:
Thanks @yanyushuai for this contribution. I'm closing it to keep the queue tidy, since there hasn't been an update since the changes requested on 2026-09-02. It isn't a rejection of the idea: if you'd like to pick it back up, you're welcome to reopen it or open a fresh PR against the current `main` that addresses the review, and we'll take another look.

PR #1338, opened 2026-09-01
fix: replace rm -rf with cross-platform fs.rmSync in package scripts

liangb612:
## Summary

The build/clean scripts in 6 workspace packages used the Unix `rm -rf dist` command, which fails on Windows (cmd.exe has no `rm`), breaking `pnpm install` on Windows at the postinstall step.

This PR replaces `rm -rf dist` with Node's built-in `fs.rmSync`, which is cross-platform (Node >= 20.9 is already required by this repo). No dependencies changed; the lockfile is untouched.

## Packages changed

- packages/@openmaic/dsl
- packages/@openmaic/generation
- packages/@openmaic/storage
- packages/@openmaic/importer
- packages/@openmaic/renderer
- packages/@openmaic/editor

## Verification

`pnpm install` completes successfully on Windows (postinstall builds all packages).

### DIFF
--- a/packages/@openmaic/dsl/package.json
+++ b/packages/@openmaic/dsl/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openmaic/dsl",
-  "version": "0.11.0",
+  "version": "0.11.1",
   "description": "The MAIC slide DSL: the pure, dependency-free contract (types + JSON Schema + validators + version/migration helpers) that @openmaic/renderer and @openmaic/importer both depend on.",
   "type": "module",
   "main": "./dist/index.js",
--- a/packages/@openmaic/dsl/package.json
+++ b/packages/@openmaic/dsl/package.json
@@ -20,8 +20,8 @@
   ],
   "sideEffects": false,
   "scripts": {
-    "clean": "rm -rf dist",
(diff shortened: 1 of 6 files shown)


puxiao:
https://github.com/THU-MAIC/OpenMAIC/pull/1110
I don't know why they choose to turn a blind eye to this issue.

liangb612:
> #1110 I don't know why they choose to turn a blind eye to this issue.

Same here, maybe they only consider the users of Linux or Unix. I wish this PR can be merged.
Meanwhile, if you need, just install from my branch directly.

wyuc (approved):
Thanks for flagging the overlap with #1110. That PR had the right diagnosis and essentially the same implementation, but remained blocked by the required package-version bumps and did not receive a timely maintainer follow-up. Apologies for the delayed response.

I have verified that this PR carries the fix onto current main, includes the required patch version bumps, leaves dependencies and the lockfile unchanged, and passes all required checks.

Thanks @liangb612 for completing the update, and @niitbeo for the original diagnosis and implementation. Approving #1338 as the current merge candidate. We will preserve the attribution when closing #1110 and resolving #973.

PR #1344, opened 2026-09-02
fix(agent-runtime): resolve VoxCPM Auto Voice for generate_tts

FireSpoonYZ:
## Summary
- resolve VoxCPM `voxcpm:auto` from the Agent Runtime's persisted teacher roster
- reuse the existing deterministic voice ID and voice-registration adapter, registering the bootstrap clip only when absent
- pass `registeredVoiceId` into server-side `generateTTS` so Workbench `generate_tts` matches the browser narration path

## Root cause
Agent Workbench's `scene-tts.ts` was added after the Auto Voice register-once flow. It forwarded `voice: voxcpm:auto` without `voicePrompt` or `registeredVoiceId`, so `generateVoxCPMTTS` rejected every speech action before making an HTTP request.

## Verification
- `pnpm exec vitest run tests/agent-runtime/scene-tts.test.ts`
- `pnpm exec next typegen && pnpm exec tsc --noEmit`
- `pnpm exec eslint lib/server/agent-runtime/scene-tts.ts tests/agent-runtime/scene-tts.test.ts`
- `pnpm exec prettier lib/server/agent-runtime/scene-tts.ts tests/agent-runtime/scene-tts.test.ts --check`
- `pnpm build`

### DIFF
--- a/lib/server/agent-runtime/scene-tts.ts
+++ b/lib/server/agent-runtime/scene-tts.ts
@@ -1,6 +1,13 @@
 import { DEFAULT_TTS_MODELS, DEFAULT_TTS_VOICES, TTS_PROVIDERS } from '@/lib/audio/constants';
 import { generateTTS, TTSRequestTimeoutError } from '@/lib/audio/tts-providers';
 import type { TTSProviderId } from '@/lib/audio/types';
+import { getDeterministicVoiceId, type VoiceDesign } from '@/lib/audio/voice-design';
+import { getVoiceRegistrationAdapter } from '@/lib/audio/voice-registration';
+import {
+  buildAutoVoxCPMVoicePrompt,
+  VOXCPM_AUTO_VOICE_ID,
+  VOXCPM_TTS_PROVIDER_ID,
+} from '@/lib/audio/voxcpm';
 import { BROWSER_NATIVE_TTS_PROVIDER_ID } from '@/lib/audio/provider-enablement';
 import type { LegacySpeechAction, SpeechAction } from '@/lib/types/action';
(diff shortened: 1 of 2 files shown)


CXuPercy (changes requested):
Cross-reviewed by two independent reviewers. The parity direction is right — the browser narration path already resolves `voxcpm:auto` through the deterministic voice ID + registration adapter, and this wires the same machinery into the server-side path, reusing the existing components rather than inventing new ones. One concern from review was refuted (the adapter interface does declare `signal?` on all three registration methods — typecheck is green), but one real gap needs a fix before merge:

**Registration failure is fatal to narration.** `resolveVoxCPMAutoVoiceOptions` runs before the synthesis loop with no try/catch: if `voiceExists` misbehaves or `bootstrapReferenceClip`/`registerVoice` throws (registration endpoint down while TTS itself is fine), the whole scene's narration aborts — a regression for deployments where narration previously worked without any registration traffic. Please wrap the resolution so any failure falls back to `{ voicePrompt }` — the function's own graceful-degradation branch (no design / no baseUrl / unsupported backend) already points at that intent.

Secondary: tests cover the register path only; please add the reuse path (voiceExists → true → no registration calls) and one fallback case (registration throws → narration still synthesizes with voicePrompt).

Verified: 3/3 tests locally, CI green, no UI/UX surface changes (server-side only).

wyuc:
Thanks @FireSpoonYZ for this contribution. I'm closing it to keep the queue tidy, since there hasn't been an update since the changes requested on 2026-09-03. It isn't a rejection of the idea: if you'd like to pick it back up, you're welcome to reopen it or open a fresh PR against the current `main` that addresses the review, and we'll take another look.

Issue #527, opened 2026-05-04
Normalize image/video generation constraints per provider and model


wyuc:
## Context

A runtime OpenAI image generation error exposed a broader media pipeline issue:

`OpenAI image generation failed (400): Invalid size '1024x576'. Requested resolution is below the current minimum pixel budget.`

The immediate failing case was OpenAI, but the root problem is not provider-specific. Today we often convert `aspectRatio` into generic `width/height` or pass coarse registry values before validating against the selected provider/model's actual API constraints.

## Problem

Media generation constraints are currently scattered across:

- provider registry metadata
- generic helpers like `aspectRatioToDimensions`
- adapter-specific request shaping
- implicit provider API defaults

This makes it easy to send invalid requests, especially when constraints differ by model rather than provider.

Examples from docs:

- OpenAI `gpt-image-2` supports arbitrary `size`, but only within pixel/edge/ratio constraints.
- Seedream size limits differ between 3.0, 4.0, and 4.5.
- Qwen Image 2.0 and Qwen Image Max/Plus use different allowed size strategies.
- MiniMax Image supports both `aspect_ratio` and `width/height`, but model behavior differs.
- xAI image/video supports native `aspect_ratio` / `resolution`, but adapters do not fully pass them today.
- Video providers have model-specific duration/resolution matrices, especially Veo, MiniMax, and Seedance.

## Proposal

Add a model-aware media capability normalization layer before adapter execution.

### Scope

1. Introduce a central capability registry for image and video generation keyed by `providerId + modelId`.
2. Represent provider/model constraints explicitly, for example:
   - image fixed size presets
   - image arbitrary size constraints
   - native aspect ratio support
   - native resolution support
   - minimum/maximum total pixels
   - edge divisibility requirements
   - supported video durations
   - supported video resolutions
   - valid duration/resolution combinations
3. Normalize user/request options before calling adapters.
4. Ensure adapters receive already-normalized options and only handle provider-specific wire format mapping.
5. Log both requested and normalized media parameters for easier debugging.
6. Add tests covering at least:
   - OpenAI `gpt-image-2` invalid low-pixel 16:9 normalization
   - Qwen fixed-preset vs arbitrary-size behavior
   - Seedream model-specific size constraints
   - MiniMax Image aspect ratio and width/height precedence
   - xAI image/video aspect ratio and resolution pass-through
   - Veo duration/resolution/aspect ratio combinations
   - Seedance and MiniMax video duration/resolution matrices

[...]

PR #1346, opened 2026-09-02
fix: normalize GPT Image 2 generation sizes

bigshezhang:
## Summary
- normalize GPT Image 2 image requests to the OpenAI Images API canonical square, landscape, or portrait size before provider execution
- apply the same normalization to both the interactive image endpoint and classroom-generated media
- preserve requested arbitrary dimensions for other providers and models

## Why
A generic 16:9 request currently becomes `1024x576`, which GPT Image 2 rejects. This is the model-aware normalization gap tracked in #527.

## Validation
- `pnpm exec vitest run tests/server/image-sizing.test.ts tests/server/classroom-media-generation.test.ts tests/server/capability-force-off-routes.test.ts --maxWorkers=1`
- `pnpm exec tsc --noEmit`
- `pnpm build`

The new regression coverage verifies the normalized API request and classroom media request, including the versioned GPT Image 2 model ID.


CXuPercy (approved):
Cross-reviewed by two independent reviewers. Verified correct and scoped: the normalization only fires for `openai-image` + `gpt-image-2*` (snapshots included), snapping to the three canonical sizes; all other providers keep arbitrary dimensions. The one concern raised in review — that requests without a model header would skip normalization — is refuted by the code flow: sizing runs against the *resolved* model (`resolveImageModel` falls back to the catalog's first entry, which is gpt-image-2), and empty model resolution is rejected upstream. Both entry paths (interactive route, classroom media generation) covered by tests at three levels; 28/28 locally, CI green. The known trade-off — a 16:9 request yields a 3:2 image, since the model supports only the canonical buckets — is the right call over failing outright.

Bug fix, no boundary change: the model was already in the catalog and selectable; this makes it actually work.

PR #1356, opened 2026-09-02
feat(media): add OpenRouter image and video providers

### LINKED ISSUE #1355: Add OpenRouter as an image and video generation provider
### Problem

OpenMAIC ships seven image providers and six video providers (Veo, Kling, Seedance, MiniMax, Grok, HappyHorse). Each needs its own account and its own API key, even though several route to the same upstream model families.

OpenRouter fronts those same families behind a single key, and is already a supported LLM provider here — but it is not selectable under Image Generation or Video Generation.

### Proposal

Add `openrouter-image` and `openrouter-video` providers using OpenRouter's dedicated media endpoints (not chat-completions):

- Image: `POST /images` → `{ data: [{ b64_json }] }`
- Video: `POST /videos` → 202 job, poll `GET /videos/{id}`, then `GET /videos/{id}/content`

OpenRouter currently lists 48 image and 28 video models and adds more over time, so the model list should be read live from `GET /images/models` and `GET /videos/models` rather than pinned in the registry — a hardcoded shortlist would decide for the operator which models exist.

### Notes

I have an implementation ready with adapter contract tests (stubbed fetch, no billable calls) and will open a PR against this issue.

WizKid1968:
Closes #1355

## What

Adds `openrouter-image` and `openrouter-video` as selectable providers under Image Generation and Video Generation.

## Why

OpenMAIC ships seven image providers and six video providers, each needing its own account and key. OpenRouter fronts the same model families (Veo, Kling, Runway, Seedance, Hailuo, Wan, Sora, Grok Imagine on video; FLUX, Seedream, GPT Image, Gemini, Qwen Image, Recraft, Krea on image) behind one key, and is already a supported LLM provider here.

## How

Both adapters use OpenRouter's dedicated media endpoints, not chat-completions:

- **Image** — `POST /images` → `{ data: [{ b64_json }] }`
- **Video** — `POST /videos` → `202 { id, status }`, poll `GET /videos/{id}`, then `GET /videos/{id}/content` for the mp4 bytes, returned as a data URL like the Veo adapter

**Live model list.** OpenRouter lists 48 image and 28 video models today and adds more, so the picker reads `GET /images/models` and `GET /videos/models` through a new `/api/openrouter-models` route (10-minute server cache) instead of a pinned array — a hardcoded shortlist would decide for the operator which models exist. The registry keeps a three-entry seed as an offline fallback, and the existing custom-model UI still accepts any model id. Both catalogs answer unauthenticated, so the picker fills before a key is pasted; a key is forwarded when present, for proxied base URLs.

## Testing

- New `tests/media/openrouter-adapters.test.ts` — request shape, empty-response handling, and the video job state machine including terminal failure. `fetch` is stubbed; **no test performs a billable call.**
- `tests/media/video-provider-catalog.test.ts` — the repo's existing guard that every catalog id dispatches to its own adapter; extended with the new provider.
- `pnpm test tests/media/`: 356 passed.
- `tsc --noEmit`, `eslint`, `prettier --check` (repo-wide): clean.
- `check-i18n-keys`: passes; both keys added to all 12 locales.
- `.env.example` documents `IMAGE_OPENROUTER_API_KEY`/`_BASE_URL` and `VIDEO_OPENROUTER_API_KEY`/`_BASE_URL`.
- No publishable `packages/@openmaic/*` file touched, so no version bump.

## Caveat

Request and response shapes were verified against OpenRouter's published OpenAPI spec and both live catalog endpoints, but I have not yet run a paid end-to-end generation through either adapter. Worth a maintainer sanity check on a real key before merge.


WizKid1968:
Pushed two follow-up commits from configuring this end to end. Three of the five fixes are pre-existing bugs that are not specific to OpenRouter — flagging them so they can be split out if you would rather take them separately.

**Specific to this PR**

1. Both connectivity probes hit the model catalogs, which answer 200 unauthenticated — `Test Connection` reported success for any string, including an invalid key. Now probes `GET /key`, which actually rejects a bad key.
2. The settings Models list is a read-only catalog for every provider; the real picker is the media popover, and that built its groups from the static registry array. OpenRouter therefore offered only the three-entry seed while settings showed the full live catalog — visible but not selectable. The popover now reads the same live catalog, fetched only once the provider is usable.

**Pre-existing, surfaced here**

3. The settings field is labelled "Base URL" but the panel echoes it back as "Request URL", so pasting the full endpoint (`https://openrouter.ai/api/v1/images`) is the natural mistake — it built `/api/v1/images/images` and 404'd. The adapters now trim a trailing slash and a trailing `/images`/`/videos`; a proxy path that merely contains the word is left alone. Covered by a test.
4. `image-settings` and `video-settings` read `data.message` on a failed connectivity test, but failures answer with `error` (`apiError`) and only successes carry `message`. Every failing test — for **any** provider — rendered "connection failed: undefined" instead of the reason.
5. `imageProviderId`/`videoProviderId` are empty until a provider is chosen, which first-run auto-config leaves blank when the server reports no media provider. Opening the settings panel then selected nothing: the header rendered the missing key as `settings.undefined`, and Test Connection posted a blank `x-image-provider`/`x-video-provider`, failing with "No image/video provider configured" whatever key was typed. Now falls back to the first catalog entry.

Checks still green: `tsc --noEmit`, `eslint`, `prettier --check`, `check-i18n-keys`, and `vitest tests/media/` (356 passed). The caveat in the description stands — no paid end-to-end generation has been run through either adapter.

CXuPercy (changes requested):
Cross-reviewed by two independent reviewers, with each finding verified against the code. The overall structure is good (12-locale coverage, hook/settings integration following conventions, real-user demand in #1355), but one critical security issue blocks merge:

**1. Critical — `/api/openrouter-models` is an SSRF and server-key exfiltration path.** The route takes a client-controlled `x-base-url` (highest precedence) with **no `validateUrlForSSRF`**, while `apiKey` prefers the **server env key** (`IMAGE/VIDEO_OPENROUTER_API_KEY`) — so a malicious caller can direct the server to send the operator's OpenRouter key as an `Authorization: Bearer` header to an arbitrary URL. The route's own comment says it follows "the same pattern `/api/verify-image-provider` uses" — that pattern includes `validateUrlForSSRF` for client base URLs (verified present on both reference routes). Fix direction: the env key must only ever be sent to the env-configured/default base URL; a client-supplied base URL may only be paired with the client's own `x-api-key`, after SSRF validation. Please also add `redirect: 'manual'` and a timeout, and tests proving server keys never reach client-selected URLs.

**2. High — the discovery route skips base-URL normalization.** The adapters normalize trailing `/images`/`/videos` via `openRouterBaseUrl`; the route builds `${baseUrl}/${kind}s/models` raw, so a settings URL ending in `/images` — explicitly supported by the adapter — yields `/images/images/models` in the picker. Reuse the same helper.

**3. High — generated image MIME is discarded.** The adapter's response type accepts `media_type` but only returns `b64_json`; the orchestration layer wraps base64 as `data:image/png` unconditionally (`media-orchestrator.ts:306` on main). JPEG/WebP results get mislabeled. Preserve the MIME type (or normalize the bytes).

**4. CI red: provider-neutrality guard.** The Lint job fails on four new `openrouter` tokens in `lib/server/provider-config.ts` — the guard debt (`['openrouter', 2]`) wasn't updated. Please raise it or refactor the mapping.

**5. Medium:** adapters' generation/poll requests omit `redirect: 'manual'` (the same adapter's `/key` probe has it); video polling sleeps aren't abort-aware; the route's per-URL cache is unbounded for client-supplied URLs; and the highest-risk paths (credential precedence, SSRF blocking, MIME, cache isolation) have no tests.

The provider addition itself is welcome — this is about landing it safely.

WizKid1968:
Thanks — the critical finding is correct, and the diagnosis was exact. Pushed 4290f1e.

**1. SSRF / server-key exfiltration.** Fixed as directed. The boundary is now explicit: the server env key travels only to the operator's own base URL. A client-supplied `x-base-url` is `validateUrlForSSRF`-checked and paired **only** with that caller's own `x-api-key`; the server key is dropped entirely. The unauthenticated public-catalog fallback no longer forwards a credential chosen for a different host, all reads use `redirect: 'manual'` (a redirect would otherwise carry the Authorization header off-host and reopen the hole), and each read is bounded by a 10s timeout.

The cache is now keyed by destination **and** a hash of the credential, and bounded to 64 entries with oldest-first eviction — so client-supplied URLs cannot grow it without limit, and one caller's key-authorised catalog is never served to another.

New `tests/api/openrouter-models-credentials.test.ts` pins exactly what you asked for: which credential reaches which URL, that the server key never appears in a request to a client-chosen host (with or without a caller key), that an SSRF-rejected destination is **never contacted**, that the fallback is unauthenticated, and cache isolation between callers. 8 tests.

**3. Image MIME discarded.** Fixed. The adapter now returns a data URL carrying the reported `media_type`, so `media-orchestrator.ts`'s unconditional `data:image/png` wrap no longer mislabels jpeg/webp. Two tests cover the reported type and the png fallback.

**5. Medium items.** `redirect: 'manual'` added to the adapters' generation and poll requests, matching the `/key` probe that already had it. `runPolledTask` now takes an optional `AbortSignal` so the sleep between polls is cancellable — previously a cancelled generation still slept out a full 10s interval — and the video adapter passes the caller's signal through.

**2 and 4 were already fixed**, in d553a08, pushed the day after the review was submitted. The route reuses `openRouterBaseUrl` (so a settings URL ending in `/images` no longer yields `/images/images/models`), and the neutrality-guard debt was raised from 2 to 6 — each env entry contributes both its key and its value, so the two new mappings add four tokens. CI is green on that commit.

Local checks on 4290f1e: `prettier --check` and `tsc --noEmit` clean, eslint 0 errors, 622 tests passing across `tests/api`, `tests/media`, `tests/providers`, `tests/audio`.

One thing I should flag rather than leave implied: the adapters' request and response shapes were verified against OpenRouter's published OpenAPI spec and both live catalog endpoints, but I still have not run a paid end-to-end generation through either adapter. Worth a maintainer sanity check with a real key before merge.

CXuPercy:
The remaining open item is the one you flagged yourself: no paid end-to-end generation has been run through either adapter. Since you have the OpenRouter account and key, please run that smoke yourself before merge:

1. One **image** generation through the adapter with a real key — cheapest image model is fine — and post the result (screenshot of the generated media, or the log line with the request id). This also directly exercises the MIME fix: please confirm a non-PNG result (jpeg/webp) carries its real `media_type`, not the hardcoded `data:image/png`.
2. One **video** generation the same way, including a poll cycle, confirming the abort-aware sleep doesn't affect a normal run.

Everything else on this PR is verified from my side: all five review findings resolved on `4290f1e`, the credential-boundary tests pass (15/15), and I ran a local settings pass — both panels render, an invalid key is correctly rejected with 401 through the `/key` probe, and the live catalog loads through the secured route. Once your smoke evidence is up, this is ready for final approval.

WizKid1968:
Paid end-to-end smoke tests completed on the current rebased head `0c62073` using a real OpenRouter account/key and the adapters in this PR.

### Image adapter

- Model: `black-forest-labs/flux.2-klein-4b`
- OpenRouter request trace: `a3ba2274bdd2602a-IAD`
- Result: 175,366-byte, 1024×1024 JPEG
- Adapter data URL prefix: `data:image/jpeg;base64,`
- Verified independently with `file` as JPEG/JFIF bytes.

This exercises the non-PNG path requested in review: OpenRouter reported `image/jpeg`, and the adapter preserved that real media type rather than wrapping it as `data:image/png`.

### Video adapter

- Model: `bytedance/seedance-2.0-fast`
- Request: 4 seconds, 480p, 1:1
- OpenRouter request trace: `a3ba22832d8c602a-IAD`
- Normal polling completed after 10 poll requests.
- Result: 739,690-byte `video/mp4`; `ffprobe` reports 4.096 seconds.

The submit → abort-aware wait → poll → content-download path completed normally without cancellation or timing errors.

Post-merge local validation also passes:

- `pnpm vitest run tests/media/openrouter-adapters.test.ts tests/api/openrouter-models-credentials.test.ts` — 15/15 passing
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm exec tsc --noEmit` — clean

WizKid1968:
Vercel deployment follow-up:

- The feature preview initially failed after a successful Next.js build because it exceeded Vercel Hobby's limit of 12 bundled Serverless Functions.
- The cause was the global `outputFileTracingIncludes` rule force-including every Sharp/libvips native library variant in Vercel function traces. That rule exists for self-hosted standalone/Docker output; Vercel handles its own runtime dependency packaging.
- `fbfe183` now applies the Sharp tracing workaround only to non-Vercel standalone builds. Self-hosted behavior is preserved.
- I also removed the extra model-catalog proxy function in `2548c1e`; OpenRouter's public image/video catalogs are CORS-enabled and were verified directly from a browser origin (HTTP 200, 52 image and 29 video models). Custom gateways receive only the caller-supplied key.

Verification:

- Vercel Hobby preview is Ready: https://open-maic-43xgqem1y-rudys-projects-ad8c1dfb.vercel.app/
- TypeScript: clean
- Prettier: clean
- OpenRouter media/provider tests: 538 passed
- The prior paid image and video adapter smoke evidence remains unchanged because the generation adapters were not altered by these deployment fixes.

CXuPercy:
### Local UIUX Verification：

Model list is working properly：
20260916155617_rec_

The key error corresponds to the correct error message：
image

CXuPercy (approved):
The smoke evidence closes everything from my side. Both adapters verified end-to-end with a real account — the JPEG result exercising the non-PNG MIME path exactly as requested (`data:image/jpeg` preserved, independently checked with `file`), and the video path completing a normal 10-poll cycle with the abort-aware wait intact. The two deployment commits are sound as well: `fbfe183` gates the sharp-libvips tracing include to non-Vercel builds (consistent with the existing `output: standalone` fork in the same file — and it fixes a real Vercel regression the merged #1407 rule introduced), and `2548c1e` deleting the server catalog proxy is a net simplification: the browser fetches OpenRouter's public CORS-enabled catalog directly with no credentials involved, removing the server-side credential-pairing surface the earlier review had hardened — better than hardening it is not needing it. 538 tests green on the head, plus a live preview link.

I also re-ran the settings UI pass on the final head locally — both panels render, the live catalog loads through the new direct browser fetch, and an invalid key still surfaces the real 401 through the `/key` probe.

All five original findings fixed and verified, the credential boundary pinned by tests, a local UI pass done, and paid smoke evidence in. @wyuc approved from my side — please confirm and merge.

wyuc (approved):
Approving. The OpenRouter image and video adapters follow the existing media-provider patterns (dedicated `/images` and `/videos` endpoints, `redirect: 'manual'` on every authenticated request), the earlier server-side catalog proxy that could send the operator key to a client-chosen host is gone, and CXuPercy's real-account smoke test covered both the image (non-PNG) and the polled video path. The optional abort signal in `runPolledTask` is backward compatible for the existing video adapters, and the non-Vercel gate keeps the sharp-libvips tracing fix for standalone builds.

One note for anyone running a self-hosted OpenAI-compatible TTS server: `generateOpenAITTS` now always sends `response_format: 'mp3'`. That's the OpenAI default, but a server that can't produce mp3 would now error instead of returning its own default format. Thanks!

PR #1358, opened 2026-09-02
fix(render): preserve plan audio during chunk assembly

### LINKED ISSUE #1352: [Bug]: Chunked video export (RENDER_CHUNK_EXECUTION) silently drops audio
### Bug Description

With `RENDER_CHUNK_EXECUTION=true`, the render service produces a **video-only MP4 even when the composition has audio**. The mixed audio track is fully rendered and present in the plan directory, but never muxed into the final output. There is no error, no warning, and the job reports `succeeded`. The non-chunked path (`executeRenderJob`) exports the same project with audio intact.

Root cause is a filename mismatch between the render service and `@hyperframes/producer` 0.7.107:

- `producer.plan()` writes the mixed audio as `planDir/audio.m4a` (`MIXED_AUDIO_FILENAME` renamed from `audio.aac`; the legacy name is only kept for reading).
- `render-service/src/chunk-executor.ts` (assembly step of `executeRenderChunks`) still probes the legacy name:
  ```ts
  const audioPath = join(plan.planDir, 'audio.aac');
  ... (await exists(audioPath)) ? audioPath : null
  ```
  `exists()` is always false → `assemble()` receives `null` → the audio pad/trim + mux steps are skipped entirely.

The chunk executor landed in #1115 together with the producer bump 0.7.60 → 0.7.107, so the chunked path has been silently audio-less since it merged. This directly violates the Definition of Done in #1091 ("No OOM, blank frames, **missing audio**, or unplayable output in the supported profile").

### Steps to Reproduce

1. Run the render service (current `main`) with `RENDER_CHUNK_EXECUTION=true`, `RENDER_CHUNK_COUNT=4`, `RENDER_MAX_PARALLEL_CHUNKS=4` (any composition with an audio element; I used a manually constructed 16s/480-frame project with a looping `<audio data-start="0">`).
2. `POST /render` with the project ZIP, wait for `succeeded`, download the MP4.
3. `ffprobe` the output.

Control: same project with `RENDER_CHUNK_EXECUTION=false` exports with audio.

### Expected Behavior

- The chunked export contains the mixed AAC audio stream (as the non-chunked path does).
- If the producer declares `plan.json hasAudio: true` but no audio artifact can be found, the job should **fail loudly** (or at minimum warn) rather than silently ship video-only.

### Actual Behavior

- Chunked output: single `h264` stream, no audio stream.
- Plan dir after the run: `audio.m4a` present and `plan.json hasAudio: true` — the audio was rendered, then discarded.
- Job status: `succeeded`, no warning anywhere.

### Fix Direction

The producer already exports `resolvePlanAudioPath` / `PLAN_AUDIO_RELATIVE_PATH` (+ legacy constant) from `@hyperframes/producer/distributed`; use those instead of hardcoding `audio.aac`, and fail/warn when `hasAudio` is true but no artifact exists. Add a unit test at the executor seam (the `assemble` dependency is injectable) covering: `audio.m4a` picked up, legacy `audio.aac` still honored, missing-artifact-with-`hasAudio` handled.

### Area

Import / export

### Additional Context

Verified on `main` (post-#1115, producer 0.7.107): A/B render of the same audio project, chunked vs non-chunked; ffprobe streams and plan-dir artifacts captured. Related: #1094 (original chunk executor task), #1115 (merge), #1091 (epic).

MohammedThowfiq:
## Summary
Preserve the producer's audio artifact when assembling rendered chunks. Resolve the artifact through `resolvePlanAudioPath`, so both current `audio.m4a` and legacy `audio.aac` plans retain their soundtrack. Treat `plan.json.hasAudio` as authoritative so a stale artifact from a failed planning attempt is never attached to a later silent plan. Fail explicitly when a plan declares audio but its artifact is missing.

## Related Issues
Fixes #1352

## Changes
- Resolve and pass the producer audio artifact only when `plan.json` declares audio.
- Reject missing declared audio before assembly instead of silently exporting a video without sound.
- Cover M4A, AAC, silent plans, missing declared audio, and stale audio after a partial planning failure.

## Type of Change

## Verification
Run from `render-service` after rebasing onto current `main`:
- `vitest run test/chunk-executor.test.ts` — 15 tests passed.
- `tsc --noEmit` — passed.
- `git diff --check origin/main...HEAD` — passed.

The regression cases exercise artifact selection and assembly error handling with mocked producer calls, including an `audio.m4a` left in the plan directory while the current `plan.json` contains `hasAudio: false`.

Full application formatting/lint/typecheck, a live Chromium/FFmpeg export, and manual end-to-end regression checks have not been completed. This PR remains a draft pending those checks and human review.

- [ ] Manually tested locally
- [ ] Screenshots / recordings attached (not applicable)

## AI assistance and review
AI-assisted contribution: Codex implemented the change and regression tests and performed an additional agent review of the diff. This is not a claim of human review; human review is pending.

## Checklist
- [x] Changed files follow the project's formatting style
- [x] Agent self-review completed
- [x] No documentation change needed for this internal bug fix
- [ ] Full application checks establish no new warnings

### DIFF
--- a/render-service/src/chunk-executor.ts
+++ b/render-service/src/chunk-executor.ts
@@ -14,6 +14,7 @@ import {
   plan as producerPlan,
   renderChunk as producerRenderChunk,
   hashProjectDir,
+  resolvePlanAudioPath,
   type AssembleResult,
   type ChunkResult,
   type DistributedRenderConfig,
@@ -844,11 +845,17 @@ export async function executeRenderChunks(
     let assembly: AssembleResult;
     const assembleStartedAt = Date.now();
     try {
-      const audioPath = join(plan.planDir, 'audio.aac');
+      const audioPath = resolvePlanAudioPath(plan.planDir);
+      const producerMetadata = JSON.parse(
+        await readFile(join(plan.planDir, 'plan.json'), 'utf8'),
+      ) as { hasAudio?: boolean };
+      if (producerMetadata.hasAudio && !audioPath) {
+        throw new Error('Plan declares audio but its audio artifact is missing');
+      }
       assembly = await (dependencies.assemble ?? defaultDeps.assemble)(
         plan.planDir,
         results.map((result) => result.outputPath),
-        (await exists(audioPath)) ? audioPath : null,
+        audioPath,
         plan.outputPath,
         { abortSignal: request.signal },
       );
--- a/render-service/src/chunk-executor.ts
+++ b/render-service/src/chunk-executor.ts
@@ -845,10 +845,10 @@ export async function executeRenderChunks(
     let assembly: AssembleResult;
     const assembleStartedAt = Date.now();
     try {
-      const audioPath = resolvePlanAudioPath(plan.planDir);
       const producerMetadata = JSON.parse(
         await readFile(join(plan.planDir, 'plan.json'), 'utf8'),
       ) as { hasAudio?: boolean };
+      const audioPath = producerMetadata.hasAudio ? resolvePlanAudioPath(plan.planDir) : null;
       if (producerMetadata.hasAudio && !audioPath) {
         throw new Error('Plan declares audio but its audio artifact is missing');
       }
--- a/render-service/test/chunk-executor.test.ts
+++ b/render-service/test/chunk-executor.test.ts
@@ -185,6 +185,29 @@ describe('local bounded chunk executor', () => {
     expect(assembled).toBe(false);
   });
 
+  it('ignores stale audio from a failed planning attempt when the current plan has no audio', async () => {
+    const paths = setup();
+    await materializeProject(paths.projectDir);
+    let assembledAudio: string | null | undefined;
+    const dependencies = deps();
+    const assemble = dependencies.assemble!;
+    await executeRenderChunks(
+      { ...paths, options },
+      deps({
+        plan: async (...args) => {
+          const result = await fakePlan(...args);
+          await writeFile(join(paths.planDir, 'audio.m4a'), 'stale-audio');
+          return result;
+        },
+        assemble: async (...args) => {
+          assembledAudio = args[2];
+          return assemble(...args);
+        },
+      }),
+    );
+    expect(assembledAudio).toBeNull();
+  });
+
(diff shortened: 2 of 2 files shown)


LING-6150 (changes requested):
The resolver direction is correct, but plan.json.hasAudio needs to remain authoritative.
This is reachable after a partial planning failure: createRenderPlan only removes planDir when the adjacent local envelope already exists. If producer has promoted audio.m4a and planning then fails before that envelope is written, the next attempt reuses the directory. Producer does not remove the root audio artifact when the new plan has no audio, so assembly can receive stale audio despite hasAudio: false.
I reproduced that two-attempt lifecycle. Please only resolve/pass audio when hasAudio is true, and add a regression for the partial-plan retry case.

MohammedThowfiq:
Addressed in c0ff3c2. Assembly now resolves and passes audio only when the current plan metadata has `hasAudio: true`. I added a regression that leaves `audio.m4a` in the reused plan directory while the current plan declares no audio and verifies assembly receives `null`. The branch is rebased onto current `main`; the focused executor suite passes 15/15 and `tsc --noEmit` passes.

LING-6150 (approved):
Thanks for the update. The stale-audio blocker is resolved at c0ff3c25: plan.json.hasAudio now controls whether audio is passed to assembly.
I independently verified M4A and legacy AAC output using the real producer assembly and FFmpeg. A silent retry after a partial planning failure no longer inherits stale audio, and missing declared audio fails explicitly. The 15 executor tests and typecheck also passed.
Verification covered assembly and output artifacts, not the full Chromium capture/HTTP export flow. No remaining blocking findings. Approved.   cc @wyuc   @cosarah

wyuc (approved):
Approving on the strength of @LING-6150's review: the plan.json.hasAudio blocker is resolved at c0ff3c25 and the M4A/legacy AAC assembly was independently verified with the real producer and FFmpeg. CI is green on the updated branch; merging.

wyuc (approved):
Approving on the strength of @LING-6150's review: the plan.json.hasAudio blocker is resolved at c0ff3c25 and M4A / legacy AAC assembly was independently verified with the real producer and FFmpeg. CI is green on the updated branch.

PR #1359, opened 2026-09-02
fix(render): enforce chunk worker cap

### LINKED ISSUE #1353: [Bug]: Chunk executor exceeds the 1-worker profile cap (2 capture workers per chunk) and misreports actualWorkers
### Bug Description

With chunked execution enabled, the number of capture workers actually spawned **exceeds every configured worker knob**, and the job metrics report the configured value instead of the observed one. On the `standard` resource profile — which pins `producerWorkers=1`, `maxChunkWorkers=1` — a chunked render runs **2 capture workers per chunk, multiplied by `RENDER_MAX_PARALLEL_CHUNKS`**, inside a container sized for one worker.

Two stacked causes:

1. **Per-chunk floor overrides the cap (producer side).** In `@hyperframes/producer` 0.7.107, `renderChunk` sizes capture with `calculateOptimalWorkers(framesInChunk, undefined, cfg)` — `requested` is never passed. `PRODUCER_MAX_WORKERS`/`RENDER_CHUNK_WORKERS=1` only bounds `effectiveMaxWorkers`, but any chunk ≥ 120 frames (default `minParallelFrames`) trips the `minWorkersForJob = 2` floor, and `finalWorkers = Math.max(2, Math.min(1, optimal)) = 2`. The non-chunked path passes `workers: 1` explicitly (render-service `buildProducerJobConfig`), where sizing is exact — so the two paths disagree for identical config.
2. **Metrics report the plan knob, not reality (render-service side).** The chunked branch of `InProcessExecutor.execute` (`render-service/src/render-executor.ts`) sets `actualWorkers = result.plan.chunkWorkers` (= 1) and `performance.workers` the same. Nothing observes what the chunk workers actually used.

Measured on `main` (4 chunks × 120 frames, `RENDER_CHUNK_WORKERS=1`, `RENDER_MAX_PARALLEL_CHUNKS=4`, container `--cpus 6 --memory 8g`):

- Every chunk's `.perf.json` sidecar records `"workers": 2` → 8 concurrent capture browsers.
- Process sampling inside the container: peak **54 chromium processes / 22 browser mains** (non-chunked control on the same project: 6 / 2).
- Job metrics nonetheless report `requestedWorkers: 1, actualWorkers: 1`.

So the standard profile's resource contract (a browser-budget of 1) is silently multiplied ×8, while observability claims compliance.

### Steps to Reproduce

1. Render service on `main` with `RENDER_CHUNK_EXECUTION=true`, `RENDER_CHUNK_COUNT=4`, `RENDER_CHUNK_WORKERS=1`, `RENDER_MAX_PARALLEL_CHUNKS=4`, any composition ≥ 480 frames at 30 fps.
2. During the render, count chromium processes in the container; after the render, read the chunk sidecars (`<planDir>.chunks/*.perf.json` → `"workers"`) and `GET /render/:id` → `metrics`.

### Expected Behavior

- With `RENDER_CHUNK_WORKERS=1` / `PRODUCER_MAX_WORKERS=1`, each chunk uses at most 1 capture worker (matching the non-chunked path's behavior for `workers: 1`).
- `metrics.actualWorkers` reflects observed worker usage (e.g. from chunk sidecars), not the configured knob.

### Actual Behavior

- 2 capture workers per chunk (sidecar `"workers": 2`), 8 concurrent browsers with 4 parallel chunks.
- `metrics.actualWorkers = 1` regardless.

### Fix Direction

- Propagate `PRODUCER_MIN_PARALLEL_FRAMES` (set high, e.g. `999999`) to the forked chunk workers alongside `PRODUCER_MAX_WORKERS` so the min-parallel floor cannot exceed the cap — the same technique already required to force 1 worker on the monolithic path. (Longer term: let `renderChunk` accept an explicit worker count.)
- Aggregate observed workers from the chunk results/sidecars into `metrics.actualWorkers` (and consider exposing total concurrent capture workers = per-chunk workers × parallel chunks, which is what the profile actually budgets).

### Area

Import / export

### Additional Context

Verified empirically on `main` (post-#1115, producer 0.7.107); evidence from an A/B run against the same project. Related: #1094, #1115, #1091 (epic — resource-bounded chunk executor), #1041 (earlier worker-defaults bug).

hydraxman:
## Summary

- Keep one-worker chunk renders at one capture worker even above the producer's parallelization threshold.
- Report the maximum worker count observed in chunk results instead of echoing the plan setting.
- Preserve reusable plans when the producer omits executor-only worker controls from `encoder.json.runtimeEnv`.

AI-assisted and self-reviewed before submission.

## Related Issues

Fixes #1353

## Changes

- Apply and restore the producer's serial-capture threshold around one-worker chunk execution.
- Derive chunked render metrics and performance data from observed chunk worker counts.
- Keep plan reuse keyed by the executor-owned `chunkWorkers` field rather than expecting non-snapshotted environment variables in producer metadata.
- Add regression coverage using the real producer runtime snapshot shape.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Run `pnpm --dir render-service test`.
2. Run `pnpm --dir render-service typecheck`.
3. Run `pnpm exec prettier --check render-service/src/chunk-executor.ts render-service/src/render-executor.ts render-service/test/chunk-executor.test.ts render-service/test/render-executor.test.ts`.

- The original worker-cap regression failed on the PR base: chunk execution retained `PRODUCER_MIN_PARALLEL_FRAMES=120`, and metrics reported configured workers `1` when chunk results observed `2`.
- The cache-reuse regression failed before the review fix with two planner calls instead of one when `encoder.json.runtimeEnv` matched producer 0.7.107's actual filtered shape.
- The same regression passes after removing the invalid runtime snapshot predicates while retaining the executor-owned `chunkWorkers` cache key.
- The current-main render-service suite passes: 142 tests.
- Render-service type checking, Prettier 3.8.1, and `git diff --check` pass for the complete rebased patch.

### Evidence

- [x] Manually tested locally

## Checklist

### DIFF
--- a/render-service/src/chunk-executor.ts
+++ b/render-service/src/chunk-executor.ts
@@ -25,6 +25,7 @@ import { config } from './config.js';
 export const CHUNK_PLAN_SCHEMA_VERSION = 1 as const;
 export const DEFAULT_CHUNK_COUNT = 1;
 export const DEFAULT_CHUNK_WORKERS = 1;
+const SERIAL_CAPTURE_MIN_PARALLEL_FRAMES = String(Number.MAX_SAFE_INTEGER);
 
 export type ChunkFailureCode =
   | 'missing_chunk'
@@ -619,7 +620,11 @@ export async function createRenderPlan(
   await mkdir(planDir, { recursive: true });
   const dimensions = await readCompositionDimensions(projectDir);
   const previousWorkers = process.env.PRODUCER_MAX_WORKERS;
+  const previousMinParallelFrames = process.env.PRODUCER_MIN_PARALLEL_FRAMES;
   process.env.PRODUCER_MAX_WORKERS = String(chunkWorkers);
+  if (chunkWorkers === 1) {
+    process.env.PRODUCER_MIN_PARALLEL_FRAMES = SERIAL_CAPTURE_MIN_PARALLEL_FRAMES;
+  }
   let result: PlanResult;
   try {
     result = await (dependencies.plan ?? defaultDeps.plan)(
@@ -638,6 +643,8 @@ export async function createRenderPlan(
   } finally {
     if (previousWorkers === undefined) delete process.env.PRODUCER_MAX_WORKERS;
     else process.env.PRODUCER_MAX_WORKERS = previousWorkers;
+    if (previousMinParallelFrames === undefined) delete process.env.PRODUCER_MIN_PARALLEL_FRAMES;
+    else process.env.PRODUCER_MIN_PARALLEL_FRAMES = previousMinParallelFrames;
   }
   const resolvedChunkCount = result.chunkCount;
   const chunksJson = JSON.parse(
@@ -814,7 +821,14 @@ export async function executeRenderChunks(
   const plan = await createRenderPlan(request, dependencies);
   const planMs = Date.now() - planStartedAt;
   const previousWorkers = process.env.PRODUCER_MAX_WORKERS;
+  const previousMinParallelFrames = process.env.PRODUCER_MIN_PARALLEL_FRAMES;
   process.env.PRODUCER_MAX_WORKERS = String(plan.chunkWorkers);
+  if (plan.chunkWorkers === 1) {
+    // Producer's distributed renderChunk path does not accept an explicit
+    // worker count. Disable its two-worker minimum so the configured cap is
+    // also honored by forked chunk workers.
+    process.env.PRODUCER_MIN_PARALLEL_FRAMES = SERIAL_CAPTURE_MIN_PARALLEL_FRAMES;
+  }
   const results = new Array<ChunkResult>(plan.chunks.length);
   try {
     const chunksStartedAt = Date.now();
@@ -869,6 +883,8 @@ export async function executeRenderChunks(
   } finally {
     if (previousWorkers === undefined) delete process.env.PRODUCER_MAX_WORKERS;
     else process.env.PRODUCER_MAX_WORKERS = previousWorkers;
+    if (previousMinParallelFrames === undefined) delete process.env.PRODUCER_MIN_PARALLEL_FRAMES;
+    else process.env.PRODUCER_MIN_PARALLEL_FRAMES = previousMinParallelFrames;
   }
 }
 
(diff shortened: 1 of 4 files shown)


LING-6150 (changes requested):
The worker-cap mechanism looks correct, but the new cache predicate checks PRODUCER_MAX_WORKERS and PRODUCER_MIN_PARALLEL_FRAMES in encoder.json.runtimeEnv.
Producer 0.7.107 snapshots only keys prefixed with PRODUCER_RUNTIME_ or PRODUCER_RENDER_, so these two values cannot appear in the real encoder.json.runtimeEnv. As a result, an otherwise reusable plan never satisfies the predicate. I reproduced this with the producer’s actual runtimeEnv shape: creating the same plan twice invoked the planner twice.
Please remove these predicates, or persist an executor-owned policy/version field explicitly, and add a cache-reuse regression using the real producer snapshot shape.

hydraxman:
@LING-6150 Thanks for catching this. The producer snapshot contract makes those two runtime predicates invalid, so I removed them and kept plan reuse keyed by the executor-owned `chunkWorkers` field. I also changed the fixture to match the real filtered `encoder.json.runtimeEnv` shape and added a regression that fails with two planner calls before the fix and passes with one after it. The branch is rebased onto current `main`; the render-service suite is 142/142 and typecheck/Prettier/diff checks pass.

LING-6150 (approved):
Thanks, I rechecked the current head. This resolves my earlier cache-reuse finding: the invalid runtimeEnv predicates are gone, the fixture now matches the producer snapshot contract, and the regression proves that the second execution reuses the existing plan.
I also verified the one-worker environment propagation/restoration and observed-worker metrics. The full render-service suite passes (142/142) along with typecheck. No further blocking findings from me.
@wyuc This looks ready from my side.

wyuc (approved):
Verified against @hyperframes/producer 0.7.107: renderChunk builds its config through resolveConfig() on every call and reads PRODUCER_MIN_PARALLEL_FRAMES from process.env, and the two-worker floor is exactly totalFrames >= minParallelFrames ? 2 : 1, so raising the threshold while chunkWorkers === 1 is the right lever; the chunk workers are forked after the env is set, so both paths honor it. The metric change is unambiguous. Thanks for the fix and for the follow-up on the cache predicate.

PR #1363, opened 2026-09-03
fix: use configured Qwen TTS model for course audio

hwzhangd:


CXuPercy (changes requested):
Cross-reviewed by two independent reviewers, with findings verified against the diff and main. The bundle has real value but needs work on several fronts before merge:

**1. Hardcoded Chinese UI strings (blocking).** The voice-catalog UI ships nine non-i18n strings ("Qwen 官方音色", "搜索名称、ID 或语言", "未找到匹配的官方音色", "我的复刻音色", "Qwen 导入", "官方", "已选择官方音色：…", "已使用", "使用"). This repo internationalizes all settings-surface text across 12 locales — please route these through the i18n system.

**2. Server sync overwrites the user's model choice (blocking).** `settings.ts` sets `modelId: info.models[0]` on **every** sync when server models exist, clobbering a user-selected model even when it's still within the pinned list — the same user-choice regression class we've fixed before. Preserve the current selection when it remains valid; default only when absent/invalid.

**3. The remote-import feature is dead code as submitted.** `syncRemoteVoices` is defined and returned by the hook but has no call site — the pool route and the import flow can never trigger from the UI. Either wire it (with a button/entry point) or drop it from this PR.

**4. Coordination: this implements TTS `serverModels` piecemeal.** #1294's review direction (see its thread) is to build server-model sync as a family-agnostic mechanism across all provider kinds. Please align there rather than landing a TTS-specific variant.

**5. Pool route (verify intent).** No timeout/abort, and up to 100 sequential page requests per call. More importantly, in managed mode any caller can enumerate the server-configured Qwen account's voice IDs — if that exposure is intended, please state it in the PR description; if not, gate it.

**6. Disclosure asks (non-blocking):** the `resolveTTSModel` relaxation (out-of-pin catalog model now coerces to the pinned catalog model instead of throwing) reads as a sound self-heal — please document it explicitly in the PR description since the test change from `toThrow` to `toBe` is otherwise easy to misread. The widened DashScope audio-host regex looks acceptably constrained; also worth a line.

Finally: this PR has no linked issue or discussion, and it bundles five separable changes (model sync, voice catalog UI, pool route, base-URL translation, resolution semantics). Please add an issue (or expand the description) covering motivation and scope — and consider splitting the UI piece, which will also need a local browser-verification pass before approval per our UI-change bar.

wyuc:
Thanks @hwzhangd for this contribution. I'm closing it to keep the queue tidy, since there hasn't been an update since the changes requested on 2026-09-04. It isn't a rejection of the idea: if you'd like to pick it back up, you're welcome to reopen it or open a fresh PR against the current `main` that addresses the review, and we'll take another look.

PR #1364, opened 2026-09-03
fix: keep long Grok relay generations alive and inline image bytes

ciclou1:
## What changed

Rebased onto latest `main`. The `instrumentation.ts` timeout change is dropped —
#1404 now gives LLM calls their own undici headers/body timeouts, which covers
the same failure without raising the process-wide defaults.

Two changes remain, each fixing a distinct failure seen when Grok is reached
through a relay (a custom base URL) instead of `api.x.ai` directly.

### 1. `lib/ai/providers.ts` — long chat completions cut off by the relay

**Failure:** with `OPENAI_COMPAT_USE_STREAMING_CHAT=true`, a long non-streaming
chat completion through the relay was killed with a 504 at the gateway's idle
timeout (~5 min). Nothing is sent upstream until the model has composed the
entire answer, so the connection sits idle for the whole generation.

**Fix:** add `grok` to the existing streaming-compat path so the upstream call
streams and bytes keep flowing across the idle window. The SSE is buffered back
into a normal JSON response for the caller, so nothing downstream changes.

### 2. `lib/media/adapters/grok-image-adapter.ts` — image fetched from an unreachable host

**Failure:** `response_format: 'url'` returns a link on the relay's CDN host
(`imgen.x.ai`). That host may be unreachable from the server's network, so every
image generation failed at the follow-up fetch through `/api/proxy-media` — even
though the image had been generated successfully.

**Fix:** request `b64_json` instead, which inlines the bytes in the response and
removes the second hop entirely. `ImageGenerationResult` already carries an
optional `base64` field, and `lemonade-image-adapter.ts` already uses this exact
request shape.

### Verification

- `eslint` clean on both files
- `tsc --noEmit` reports no errors in either file

### DIFF
--- a/instrumentation.ts
+++ b/instrumentation.ts
@@ -15,6 +15,19 @@ export async function register(): Promise<void> {
   // want; the persistence stack is Node-only.
   if (process.env.NEXT_RUNTIME !== 'nodejs') return;
 
+  // undici's default headers timeout is 5 minutes. Slow LLM providers can take
+  // longer than that to return the first byte on big generations (e.g. scene
+  // content), which surfaces as "AI_APICallError: Headers Timeout Error".
+  // Raise it process-wide. The npm undici package's setGlobalDispatcher targets
+  // the same global symbol that Node's built-in fetch reads, so this applies to
+  // every global fetch call in the server.
+  try {
+    const { Agent, setGlobalDispatcher } = await import('undici');
+    setGlobalDispatcher(new Agent({ headersTimeout: 15 * 60_000, bodyTimeout: 15 * 60_000 }));
+  } catch (error) {
+    console.warn('[instrumentation] Failed to raise fetch timeouts', error);
+  }
+
   // Imported dynamically so the Edge bundle never pulls in `pg`.
   const { startAssetCollectorSchedule } =
     await import('@/lib/persistence/asset-collector-schedule');
--- a/lib/ai/providers.ts
+++ b/lib/ai/providers.ts
@@ -1835,7 +1835,11 @@ function usesCustomOpenAIBaseUrl(baseUrl?: string): boolean {
 
 function shouldUseOpenAIStreamingChatCompat(providerId: ProviderId, baseUrl?: string): boolean {
   return (
-    providerId === 'openai' &&
+    // 'grok' added: relays (custom base URL) with long non-streaming generations
+    // hit gateway idle timeouts (~5 min 504). Streaming upstream keeps bytes
+    // flowing so the gateway never cuts the connection; the SSE is buffered
+    // back into a normal JSON response client-side.
+    (providerId === 'openai' || providerId === 'grok') &&
     usesCustomOpenAIBaseUrl(baseUrl) &&
     process.env.OPENAI_COMPAT_USE_STREAMING_CHAT === 'true'
   );
--- a/lib/media/adapters/grok-image-adapter.ts
+++ b/lib/media/adapters/grok-image-adapter.ts
@@ -67,7 +67,10 @@ export async function generateWithGrokImage(
       model: requireModel(config.model, 'Grok Image'),
       prompt: options.prompt,
       n: 1,
-      response_format: 'url',
+      // Request base64 instead of a URL: the relay/CDN host (imgen.x.ai) may
+      // be unreachable from the server's network, so inline data avoids a
+      // second fetch through /api/proxy-media entirely.
+      response_format: 'b64_json',
     }),
   });
 
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -2,3 +2,6 @@ packages:
   - "packages/*"
   - "packages/@openmaic/*"
   - "!packages/docs" # standalone docs sub-app: own lockfile, build, deploy
+
+# Windows 下用 Git Bash 执行构建脚本（脚本里用了 rm 等 POSIX 命令）
+scriptShell: "E:\\新建文件夹 (2)\\Git\\bin\\bash.exe"
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -2,6 +2,3 @@ packages:
   - "packages/*"
   - "packages/@openmaic/*"
   - "!packages/docs" # standalone docs sub-app: own lockfile, build, deploy
-
-# Windows 下用 Git Bash 执行构建脚本（脚本里用了 rm 等 POSIX 命令）
-scriptShell: "E:\\新建文件夹 (2)\\Git\\bin\\bash.exe"


ciclou1:
小修

LING-6150:
Thanks ,  before reviewing the scope, could you share the actual error logs or reproduction for these changes?
In particular:
- which provider/base URL produced the Undici Headers Timeout Error;
- which provider/relay returned the 504;
- which deployment/network could not reach imgen.x.ai;
- and which individual change you verified fixed each failure.
A single request attempt cannot end as both a headers timeout and a 504 response, so these appear to be separate failures or successive troubleshooting attempts. Without the original evidence, I can’t determine whether the process-wide dispatcher change is necessary or speculative.  I

wyuc:
Thanks @ciclou1. The timeout part of this is now covered on `main` by #1404, which gives LLM calls their own undici headers/body timeouts instead of raising the process-wide defaults. Could you rebase onto the latest `main` and keep only the remaining changes (the grok streaming-compat addition and the image-fetch change), with a short note on the failure each one fixes? Otherwise we'll close this in a couple of weeks.

ciclou1:
@wyuc Done — rebased onto main (f67e708) and force-pushed. Dropped instrumentation.ts
since #1404 covers the timeout; the commit now carries only the two remaining
changes, each with the failure it fixes in both the message and the PR
description. Thanks for pointing at #1404 — per-call undici timeouts are a
better seam than raising the process-wide defaults.

ciclou1:
@LING-6150  — apologies for the slow reply, and you were right to push back on
this. They were two separate failures on two different relays/models, about
2.5 hours apart — not one request ending two ways.

Scope has since been reduced (rebased onto main, f67e708): the process-wide
undici timeout change is **dropped**, since #1404 covers it with per-call
timeouts. What follows is the evidence for the two changes that remain.

## 1. 504 on long chat completions — `lib/ai/providers.ts` (grok)

Deployment: relay in front of Grok, reached via a custom base URL
(`OPENAI_COMPAT_USE_STREAMING_CHAT=true`), model `grok-4.6`.

**Failure** (2026-08-31T09:12:39Z) — a scene whose generated HTML is large:

    [Scene Content API] Scene content generation failed [scene="…", model=grok:grok-4.6]:
    AI_APICallError: Gateway Time-out
    POST /api/generate/scene-content 504 in 5.0min

A smaller scene in the same run completed at 4.7 min, which is what pointed at a
~5 min gateway idle cutoff rather than a client-side deadline.

**After enabling the existing streaming-compat path for `grok`**
(2026-08-31T09:55:00Z), same model, a scene that also crosses the 5-minute mark:

    [Scene Content API] Generating content: "…" (interactive) [model=grok:grok-4.6]
    POST /api/generate/scene-content 200 in 5.0min

Same model, same duration bucket, 504 → 200. Non-streaming sends nothing
upstream until the whole answer is composed, so the relay's idle timer fires;
streaming keeps bytes moving and the SSE is buffered back into a normal JSON
response. This is the only change for this failure — no timeout was raised.

## 2. Unreachable image host — `lib/media/adapters/grok-image-adapter.ts`

Same deployment, relay image model `grok-imagine-image`.

**Failure** — generation succeeds, then the follow-up fetch of the returned URL
fails, so the generation is lost even though the image was produced:

    [ImageGeneration API] Generating image: provider=grok-image, model=grok-imagine-image, …
    POST /api/generate/image 200 in 7.1s
    POST /api/proxy-media 500 in 10.6s

Fetching the URL the relay returned, directly, from that host:

    URL: https://imgen.x.ai/xai-imgen/xai-tmp-imgen-62d5d220-….jpeg
    DOWNLOAD FAILED: 无法连接到远程服务器  (cannot connect to remote server)

So with `response_format: 'url'` the relay hands back a link on **x.ai's own CDN**,
and the deployment's network cannot reach it — while the API call to the relay
itself works fine. The fix asks for `b64_json` instead: the bytes arrive inline
with the response and the second hop (and its dependency on that CDN) disappears.

**Verified:** a `b64_json` request returns the image inline (~650 KB of image
data for a 1024×576 request), and no `/api/proxy-media` call is needed.
`grok-image-adapter.ts` already surfaces `imageData.b64_json` on
`ImageGenerationResult`, and `lemonade-image-adapter.ts` already uses this exact
request shape.

## On "this is a network problem, not a code problem"

Agreed that the *trigger* is environmental — but the fix does not require the
environment to change, and it costs nothing on deployments that aren't
restricted:

- The unreachable host is chosen by the relay, not by this code. With
  `response_format: 'url'` the adapter cannot know whether the link it gets back
  points somewhere the deployment can reach. `b64_json` removes that dependency
  entirely.
- Every deployment already downloads these bytes anyway: the media orchestrator
  fetches the URL through `/api/proxy-media` in order to persist the image as a
  blob (`app/api/proxy-media/route.ts` says as much). Inline bytes make that
  second hop unnecessary for **all** deployments, not only restricted ones — the
  same image, one fewer round trip.
- The pattern is already in the tree: `lemonade-image-adapter.ts` requests
  `response_format: 'b64_json'` for the same class of reason, and
  `grok-image-adapter.ts` already surfaces `imageData.b64_json`, so nothing
  downstream needs to change.
- For any deployment that cannot reach x.ai's CDN, the Grok image provider is
  currently **unusable**: the image is generated (and billed) and then always
  discarded at the proxy step. This is the difference between working and not
  working, not a preference.

The tradeoff is response size — base64 inflates the payload by roughly a third.
That is already the shape the Lemonade adapter uses, and it is well within
limits at these image sizes.

## On the timeout change

Dropped, as you and @wyuc both noted. It was a third, unrelated failure:
`Headers Timeout Error` on a *different* relay (a different custom base URL) with
`glm-5.3-flash`, at 2026-08-31T06:40:31Z, also on scene content at 5.1 min. #1404
addresses that properly by scoping the timeouts to LLM calls rather than raising
the process-wide defaults.

LING-6150 (changes requested):
Thanks for the follow-up. The rebase and removal of the process-wide timeout change address the earlier scope request.

Two points remain at `6edbfdcd`:

1. **Preserve the image format.** The new base64 path stores JPEG bytes as `image/png` in `defaultPersistGeneratedImage()`. I reproduced this with the same JPEG fixture: the base URL path preserves `image/jpeg`, while this head labels it as PNG. Other consumers also assume PNG for base64 results. Please keep the actual bytes and MIME/file extension consistent, with a focused JPEG regression test.

2. **Keep streaming compatibility limited to custom relays.** `usesCustomOpenAIBaseUrl()` only excludes `api.openai.com`, so Grok's native `api.x.ai` endpoint also gets forced into streaming when the flag is enabled. Please exclude Grok's native endpoint and cover both native and relay cases in a regression test.

These are corrections within the two retained changes; no additional functionality is requested. Once addressed, I will ask maintainer for final review.

ciclou1:
@LING-6150 Both addressed in ed923b1.

## 1. Image format is preserved

Inline bytes carry no `Content-Type`, so the base64 path had nothing to read and
recorded `image/png` unconditionally. It now reads the container from the bytes
themselves — `lib/media/image-format.ts` recognises PNG/JPEG/GIF/WebP signatures
and returns `undefined` when it cannot place them, in which case the previous
`image/png` fallback still applies. That path cannot distinguish "a format we do
not know" from "not an image at all", and inventing a third answer there would be
a guess rather than a reading.

You were right that the assumption is spread. Grok now reaches two of them, so
both were fixed rather than only the one you reproduced:

- `lib/server/agent-runtime/generate-image.ts` — `defaultPersistGeneratedImage()`,
  the site in your fixture.
- `lib/media/media-orchestrator.ts` — the classic client path wrapped bare base64
  as `data:image/png;base64,…`, recording the same wrong type one layer up.

Tests: `tests/media/image-format.test.ts` covers signature recognition, the
too-short and unparseable inputs, and the "cannot place it" answer for both entry
points. `tests/agent-runtime/generate-image.test.ts` adds the JPEG regression —
a payload with a real JPEG header now lands in the pool as `image/jpeg` — sitting
next to the existing URL-path test that already expected `image/jpeg` from the
provider's `Content-Type`, plus a case pinning the PNG fallback for unsigned
bytes.

## 2. Streaming compatibility is limited to relays

`usesCustomOpenAIBaseUrl` recognises `api.openai.com` alone, so `api.x.ai` read
as custom and the compat transport was forced onto Grok's own endpoint. Added
`isProviderNativeBaseUrl(providerId, baseUrl)`, which compares against the
registry's `defaultBaseUrl` for that provider (trailing slashes and query strings
normalised away); the compat path now requires the base URL to be a relay.

`usesCustomOpenAIBaseUrl` keeps its behaviour — it is now expressed through the
same endpoint normalisation the new check uses, so there is one definition of
"which endpoint is this" rather than two.

Tests in `tests/ai/openai-provider.test.ts` assert the upstream request body: a
relay base URL gets `stream: true` rewritten into it, and `https://api.x.ai/v1`
(plus trailing-slash and case variants) does not. The observable is the rewritten
body rather than the installed transport, because `usesCompatTransport` is
already true for every non-OpenAI provider and the dialect cannot separate them
(`shouldUseOpenAIResponsesApi` is OpenAI-only).

No other behaviour changed, and `OPENAI_COMPAT_USE_STREAMING_CHAT` still gates
both paths.

LING-6150:
@ciclou1 Thanks for following through on my review. After implementing and testing a smaller alternative, I’d like to revise my earlier guidance for the image portion. I’m sorry for the additional iteration — I should have evaluated this approach before the previous request.
Could you carry the image MIME type from the Grok adapter through ImageGenerationResult, instead of introducing the shared format-sniffing module?
The candidate I tested:
- Adds an optional mimeType field and has the Grok adapter return a typed data URL alongside the base64 bytes, recognizing the JPEG prefix and retaining the existing PNG fallback.
- Uses that MIME type in agent image persistence.
- Preserves the MIME type in the client’s stored media record.
- Uses the matching extension in classroom-media-generation.ts, which still saves Grok JPEG bytes as .png at def8c63c.
I verified this approach locally with JPEG/PNG fixtures through all three consumer paths; eight targeted test suites passed, totaling 153 tests. This did not include live xAI requests or browser E2E.
Please keep the streaming changes as they are and retain focused regression coverage for the image paths. The intent is to finish the same fix with less shared machinery, without adding functionality. If you see a compatibility reason to retain the current helper, please flag it before reworking that part.

ciclou1:
@LING-6150 Done in b50e5a9 — reworked the way you suggested, and no, I don't see a
compatibility reason to keep the helper: it was introduced by this branch, so
nothing depends on it. The module and its suite are deleted rather than left
behind.

## What it looks like now

- `ImageGenerationResult` gains an optional `mimeType`.
- `grok-image-adapter` recognises the JPEG signature in `b64_json` — that is the
  base64 of `FF D8 FF`, so the test is four characters rather than a decoder —
  keeps the PNG fallback, and reports the type both as `mimeType` and in the
  URL: a typed data URL alongside the bytes, which is the shape
  `openrouter-image-adapter` already returns for the same reason. I left a
  comment pointing at it.
- Consumers take the type from there:
  - agent image persistence (`generate-image.ts`): `result.mimeType ?? 'image/png'`
  - the client's stored media row (`media-orchestrator.ts`): the row keeps the
    type its data URL states instead of a constant. This was the other
    hardcoded `image/png` — the `storedMediaType(blob, …)` call just above it
    already read the blob.
  - `classroom-media-generation.ts`: the extension comes from a small
    mime→extension map, so JPEG bytes are written as `.jpg`. You were right that
    this one still saved `.png` at def8c63.

## Coverage

`tests/media/grok-image-adapter.test.ts` is new: inline JPEG → `image/jpeg` in
both `url` and `mimeType`, non-JPEG inline → PNG, a hosted URL → no reported
type, and the `b64_json` request shape.

Each consumer path has a case:
- `tests/agent-runtime/generate-image.test.ts` — the pool entry carries
  `image/jpeg`, plus the PNG fallback when an adapter reports nothing.
- `tests/media/media-orchestrator.test.ts` — a typed data URL lands in the
  classic row as `image/jpeg`. The stub derives the response type from the URL
  the way fetching a data URL does, so a URL claiming PNG fails it; I confirmed
  it goes red without the source change.
- `tests/server/classroom-media-generation.test.ts` — the served path ends
  `.jpg`.

Streaming is untouched since your last review. As with your candidate, none of
this was verified against live xAI or browser E2E.

LING-6150 (commented):
Image

Thanks for following through on the revisions. I revised my earlier image-handling recommendation midway through review to carry MIME information from the Grok adapter through the existing result and consumer paths. The current implementation follows that simpler approach and removes the shared format-sniffing helper. Thanks for accommodating the extra iteration.
Re-reviewed b50e5a9: 118 focused and independent fixture tests passed, covering streaming-mode selection, image MIME preservation, and persistence/file-extension handling.
Browser fixture verification also passed: real JPEG and PNG images went through the adapter and client media pipeline, displayed correctly, and survived an IndexedDB save/reload with their bytes and MIME types intact. Reloading did not trigger another generation request or remote image fetch.
This was scoped browser integration testing, not full classroom E2E or a live xAI/relay test.
No remaining blockers found in the reviewed scope. Approved from my side.  cc @wyuc, ready for your final review.

wyuc (approved):
Thanks @ciclou1, and @LING-6150 for the careful re-review. Approving after a final maintainer pass on b50e5a9:

- **Scope**: only `grok` joins the streaming-compat path; no other provider's request shape changes, and the `b64_json` switch is confined to the Grok image adapter. The new MIME field is additive and optional for every other consumer.
- **Buffered streaming**: the body is drained before parsing, error frames surface with a derived status, a missing `data:` stream falls back to the raw body, and there's no timer or reader to leak.
- **MIME**: the adapter's MIME reaches the stored asset and its file extension; forcing the JPEG branch to PNG turns the new tests red.
- It merges cleanly onto current `main`, and the media/AI suites, lint and typecheck pass.

Small follow-ups, none blocking:

1. A regression test for a relay that drops the connection mid-stream (or an abort / partial SSE chunk) through `fetchCustomOpenAIChat`: behaviour is correct today, but only the error-frame path is pinned, and Grok now depends on this seam.
2. `inlineImageMimeType` treats everything that isn't JPEG as PNG; recognizing WebP would be cheap if a relay ever returns it.
3. Other inline-byte adapters (e.g. `nano-banana-adapter.ts`, which has Gemini's `inlineData.mimeType` available, and `openai-image-adapter.ts`) could populate the new `mimeType` field too.
4. The buffered accumulator reads only `delta.content`, so streamed `reasoning_content` is dropped in the buffered conversion (pre-existing, not a regression).

I'll update the branch and merge once CI is green.

PR #1366, opened 2026-09-03
fix(slide): keep bottom table rows visible in classroom playback

### LINKED ISSUE #1365: [Bug]: Generated slide bottom content is clipped in classroom view
### Bug Description

For an AI-generated slide, the bottom row is visible in the overview thumbnail, but is completely clipped in the single-slide/classroom view.

The content is not missing from the generated slide. It becomes visible after enabling the experimental playback renderer.

### Steps to Reproduce

1. Start OpenMAIC with the default playback renderer configuration.
2. Generate a course containing a slide with content close to the bottom edge.
3. Open the generated course.
4. Compare the overview thumbnail with the single-slide/classroom view.

### Expected Behavior

The single-slide view should display the same complete content as the overview thumbnail.

### Actual Behavior

The bottom row is visible in the thumbnail but is completely missing in the single-slide view.

Image

### Deployment Method

### OpenMAIC Version

latest

### Affected Area

Playback / presentation

### Browser

Microsoft Edge 151.0.4129.107

### Operating System

Windows 11

### Relevant Logs

```shell

```

### Screenshots or Recordings

_No response_

### Additional Context

_No response_

YizukiAme:
## Summary

- align the legacy classroom table geometry with `@openmaic/renderer`
- use DSL-provided cell padding and a deterministic compact line-height instead of legacy host-page defaults
- add a real-browser regression that compares the default classroom canvas with the package-rendered thumbnail
- document that row heights are minimums and clarify the two renderers' text-wrapping behavior

## Root cause

Thumbnails and default classroom playback render the same slide through different table implementations. The legacy table put a fixed `5px` padding directly on every `<td>` and inherited line-height from the host page. HTML row heights are minimums, not hard caps, so those intrinsic cell dimensions expanded every row beyond the slide data's `rowHeights`. Near the bottom of the canvas, the enlarged table crossed the existing clipping boundary and its final rows disappeared.

The slide data is not changed. The legacy path now reuses the package renderer's existing inner-cell geometry: `rowHeight - 4` minimum height, `cell.padding`, `line-height: 1`, and flex-based `cell.vAlign`.

## Scope

This deliberately does not remove canvas clipping, constrain legitimately overfull tables, auto-fit text, migrate stored slide data, or change `@openmaic/renderer`. Those would solve different problems or alter the already-correct thumbnail output.

## User-level regression

The Chromium test opens the real classroom route with a five-row table near the bottom of a 1000 × 562.5 slide. Each row is verified to contain a softly wrapped cell, and the table element is 180px high: between the old legacy intrinsic height (~266px) and the aligned legacy/package height (~146px).

The test locates each renderer explicitly, checks that both render all five wrapped rows, and compares transformed browser rectangles using scale-independent overflow ratios. Its measurements do not depend on the new inner wrapper, so the old implementation fails for the actual geometry defect: the legacy table extends **47.78%** beyond its element. With the fix, both the default classroom canvas and thumbnail keep the table and final row within the element.

## Visual verification

Captured from the same real classroom E2E fixture (not a component-level mock):

| Before: legacy classroom renderer | After: this PR |
| --- | --- |
| Legacy classroom renderer clips the final table row | Fixed classroom renderer shows all five table rows |

## Verification

- GitHub CI: all 4 required checks passed
- Node 22 Chromium classroom regression: 1 passed
- controlled pre-fix run: failed at the table-overflow assertion (47.78%)
- Node 22 renderer contract tests: 6 passed
- Node 22 TypeScript check: passed
- focused ESLint and Prettier: passed
- Node engine, package-version, and internal-dependency checks: passed
- `git diff --check`: passed

Fixes #1365

### DIFF
--- a/components/slide-renderer/components/element/TableElement/StaticTable.tsx
+++ b/components/slide-renderer/components/element/TableElement/StaticTable.tsx
@@ -95,47 +95,64 @@ export function StaticTable({ elementInfo }: StaticTableProps) {
         ))}
       </colgroup>
       <tbody>
-        {tableData.map((row, rowIdx) => (
-          <tr
-            key={rowIdx}
-            style={{
-              height: `${Number.isFinite(tableRowHeights[rowIdx]) ? tableRowHeights[rowIdx] : safeCellMinHeight}px`,
-            }}
-          >
-            {(Array.isArray(row) ? row : []).map((cell, colIdx) => {
-              if (hiddenCells.has(`${rowIdx}_${colIdx}`)) return null;
-              if (!cell) return null;
+        {tableData.map((row, rowIdx) => {
+          const rowHeight = Number.isFinite(tableRowHeights[rowIdx])
+            ? tableRowHeights[rowIdx]
+            : safeCellMinHeight;
 
-              const bgColor = getCellBg(rowIdx, colIdx, cell.style?.backcolor);
-              const headerColor = getHeaderTextColor(rowIdx);
-              const textStyle = getTextStyle(cell.style);
-              const colspan = Number.isFinite(cell.colspan) && cell.colspan > 0 ? cell.colspan : 1;
-              const rowspan = Number.isFinite(cell.rowspan) && cell.rowspan > 0 ? cell.rowspan : 1;
+          return (
+            <tr key={rowIdx} style={{ height: `${rowHeight}px` }}>
+              {(Array.isArray(row) ? row : []).map((cell, colIdx) => {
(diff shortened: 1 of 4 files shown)


wyuc (approved):
LGTM. The classroom table was growing past the element box because of hardcoded 5px padding and inherited line-height; thumbnails already used the package geometry. Aligning the legacy inner cell (`rowHeight - 4`, DSL padding, `line-height: 1`, vAlign) without changing stored data or `@openmaic/renderer` is the right scope.

The Chromium regression actually measures overflow on the real classroom route. The slide-dsl updates match the new playback contract.

PR #1368, opened 2026-09-03
fix(generation): assign unique IDs per media request

### LINKED ISSUE #1367: [Bug]: duplicate media request IDs overwrite generated assets across scenes
### Bug Description

`uniquifyMediaElementIds` does not make every media generation request globally unique when the outline model repeats a placeholder ID. It builds one `oldId -> newId` map, so two distinct requests such as `Image 1 / gen_img_1` and `Image 2 / gen_img_1` receive the same replacement ID.

This causes deterministic last-write-wins behavior downstream:

- the browser media orchestrator indexes tasks and IndexedDB rows by `elementId`;
- the server classroom media path uses `elementId` as both filename and `mediaMap` key.

Both generations can run (and incur cost), but the later result replaces the earlier one, so different slides can display the same generated asset.

The prompt contract already requires each `mediaGenerations` definition to use a course-global unique ID. Cross-scene reuse is represented by a later scene referencing the existing ID in its content **without** adding another generation request, so assigning distinct IDs to duplicate request definitions does not break valid reuse.

### Steps to Reproduce

On current `main` (`6334e9ad`), call `uniquifyMediaElementIds` with two scene outlines whose distinct image requests both contain `elementId: "gen_img_1"`:

```ts
const result = uniquifyMediaElementIds([
  outline(1, { type: 'image', prompt: 'Image 1', elementId: 'gen_img_1' }),
  outline(2, { type: 'image', prompt: 'Image 2', elementId: 'gen_img_1' }),
]);

result[0].mediaGenerations[0].elementId ===
  result[1].mediaGenerations[0].elementId; // true
```

The failure is deterministic. Re-running produces a different random replacement, but both requests within each run still share it.

### Expected Behavior

Every `mediaGenerations` request definition receives its own globally unique replacement ID, while the input outline objects remain unchanged.

### Actual Behavior

Distinct generation requests that repeat an incoming ID are rewritten to one shared ID and overwrite each other's generated result downstream.

### Deployment Method

Local development / package API

### OpenMAIC Version

Current `main` at `6334e9ad`

### Affected Area

Classroom generation

### Browser

Provider-independent; reproduced in the pure generation package before browser rendering.

### Operating System / Node

Windows 11, Node.js 22.19.0. The logic is platform-independent.

### Additional Context

A focused fix can update the existing `outline-media.test.ts` coverage, allocate one ID per request definition, preserve structural sharing for outlines without media requests, and bump the published `@openmaic/generation` patch version as required by `CONTRIBUTING.md`.

ClayExc:
## Summary

Give every generated-media request its own globally unique ID instead of reusing one replacement for repeated model-supplied placeholders. This prevents distinct image or video generations from overwriting each other in task state, IndexedDB, server files, manifests, and media maps.

## Related Issues

Fixes #1367

## Root Cause

`uniquifyMediaElementIds` previously built an `oldId -> newId` map. When the outline model emitted two distinct requests with the same placeholder, both requests received the same replacement even though every downstream consumer treats `elementId` as a unique key.

The prompt contract defines reuse differently: one generation request may be referenced from later content, but a later scene must not add another `mediaGenerations` entry for the same ID.

## Changes

- allocate a fresh type-prefixed ID for each `mediaGenerations` request
- keep the input immutable and preserve the original array when there are no media requests
- make the existing regression deterministic and cover distinct requests that repeat an incoming ID
- bump `@openmaic/generation` from `0.3.5` to `0.3.6`

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Pass two scene outlines with different prompts but the same incoming `elementId` to `uniquifyMediaElementIds`.
2. On `main`, observe that both output requests receive one shared ID.
3. On this branch, observe that each request receives a separate ID with the prefix matching its media type.

- focused regression passes on Node.js 22.19.0 and Node.js 24
- related browser-orchestrator and video-manifest tests: 15/15 passed
- related generation prompt/manifest tests: 9/9 passed
- generation typecheck and package build passed
- Prettier, root TypeScript, i18n, Node engine, package-version, and internal-dependency checks passed
- ESLint completed with 0 errors and the existing 18 warnings
- production build and the generation Node consumer smoke passed
- full generation suite: 144 passed; the only failure is the existing Windows path-separator assertion in `test/assets.test.ts`
- full root suite on Windows: 7,334 passed; remaining failures are existing Windows-only path/shell assumptions and load-sensitive timeouts; the complete Linux CI suite passed

### Evidence

Before, two distinct requests such as `Diagram 1 / gen_img_1` and `Diagram 2 / gen_img_1` received the same random replacement within a run. The deterministic regression now produces:

```text
gen_img_00000001
gen_vid_00000002
gen_img_00000003
gen_vid_00000004
```

- [x] Manually tested locally

## AI Assistance

AI-assisted. The final diff was self-reviewed separately against repository standards, Issue #1367, and adversarial correctness/compatibility concerns; all three reviews reported no findings.

## Checklist

### DIFF
--- a/packages/@openmaic/generation/package.json
+++ b/packages/@openmaic/generation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openmaic/generation",
-  "version": "0.3.5",
+  "version": "0.3.6",
   "description": "Pure generation pipeline contracts and packaged prompt assets for MAIC consumers.",
   "type": "module",
   "main": "./dist/index.js",
--- a/packages/@openmaic/generation/src/outline-media.ts
+++ b/packages/@openmaic/generation/src/outline-media.ts
@@ -1,30 +1,18 @@
 import { nanoid } from 'nanoid';
 import type { SceneOutline } from './outline-types.js';
 
-/** Replace course-local generated-media IDs with globally unique IDs. */
+/** Give every generated-media request its own globally unique ID. */
 export function uniquifyMediaElementIds(outlines: SceneOutline[]): SceneOutline[] {
-  const idMap = new Map<string, string>();
-
-  for (const outline of outlines) {
-    if (!outline.mediaGenerations) continue;
-    for (const mediaGeneration of outline.mediaGenerations) {
-      if (!idMap.has(mediaGeneration.elementId)) {
-        const prefix = mediaGeneration.type === 'video' ? 'gen_vid_' : 'gen_img_';
-        idMap.set(mediaGeneration.elementId, `${prefix}${nanoid(8)}`);
-      }
-    }
-  }
-
-  if (idMap.size === 0) return outlines;
+  if (!outlines.some((outline) => outline.mediaGenerations?.length)) return outlines;
 
   return outlines.map((outline) => {
     if (!outline.mediaGenerations) return outline;
     return {
       ...outline,
-      mediaGenerations: outline.mediaGenerations.map((mediaGeneration) => ({
-        ...mediaGeneration,
-        elementId: idMap.get(mediaGeneration.elementId) || mediaGeneration.elementId,
-      })),
+      mediaGenerations: outline.mediaGenerations.map((mediaGeneration) => {
+        const prefix = mediaGeneration.type === 'video' ? 'gen_vid_' : 'gen_img_';
+        return { ...mediaGeneration, elementId: `${prefix}${nanoid(8)}` };
+      }),
     };
   });
 }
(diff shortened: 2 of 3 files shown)


CXuPercy (approved):
Cross-reviewed by two independent reviewers. The semantic change is right: each media-generation request now gets its own unique ID instead of first-seen placeholder IDs being aliased to one shared replacement — repeated model-supplied placeholders were being collapsed into a single request, which is the bug. Tests are deterministic (mocked nanoid sequence) and pin both uniqueness and input immutability. Version bump 0.3.5 → 0.3.6 satisfies the gate.

One second-reviewer concern checked and set aside: nanoid(8) collisions without retry. The ID space is ~2.8e14; at realistic per-course request counts the birthday probability is ~1e-11, and nanoid(8) without collision-tracking is the repo-wide convention (createElementId, scene-dup). Not a real risk.

PR #1316, opened 2026-08-31
fix(agent): preserve generate_scene failure causes

YizukiAme:
## Summary

Preserve safe, structured failure causes when valid non-PBL `generate_scene` requests reach content generation but fail before a page can be written.

## Related Issues

Fixes #1305.

## Changes

- add an optional typed failure callback to `@openmaic/generation` without changing its content-or-`null` return contract
- classify prompt construction/loading failures separately from malformed or missing model output for slide, quiz, and interactive generation
- return stable error codes and page context from `generate_scene`, while keeping raw model output out of tool results and structured logs
- add positive and negative regression coverage, document the public contract, and bump `@openmaic/generation` to 0.3.2

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Run the generation failure and type-surface tests.
2. Run the agent-runtime `generate_scene` tests.
3. Run package typecheck/build, root typecheck, lint/format/version gates, and both complete test suites under Node 22.

### What you personally verified

- prompt-unavailable and invalid-model-output classifications across slide, quiz, and interactive generation
- exact safe agent messages, structured page context, one sanitized warning, raw-output exclusion, and zero persistence/action/checkpoint side effects
- representative gate, PBL, and provider-exception paths remain outside the new callback classification
- existing content-or-null, PBL, provider exception, abort, and bulk retry behavior remains unchanged
- generation suite: 26 files, 143 tests passed
- repository suite: 651 files passed, 9 skipped; 7,221 tests passed, 26 skipped
- full lint completed with 0 errors; 18 unrelated baseline warnings remain

### Evidence

- Before: non-PBL content failures returned `details: {}` with one generic message.
- After: failures return `details.error` as `prompt-unavailable` or `invalid-model-output` with `order`, normalized `title`, `type`, and optional `sceneId`; unexpected unclassified `null` retains a generic fallback.
- [x] CI passes (`pnpm check && pnpm lint && npx tsc --noEmit`)
- [x] Manually verified locally with focused and complete automated suites
- [ ] Screenshots / recordings attached (no UI changes)

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [x] I have added/updated documentation as needed
[...]

PR #1282, opened 2026-08-30
fix(generation): load micropip before generated imports

HsiangNianian:
## Summary

Generated Pyodide widgets can import `micropip` before Pyodide loads it, which aborts initialization and leaves the Run button disabled. This change teaches the packaged code-content prompt to load `micropip` first and locks that order with a regression test.

## Related Issues

Fixes #1008

## Changes

- add conditional `micropip` loading guidance and an ordered Pyodide example
- verify the packaged prompt loads `micropip` before its first import
- bump `@openmaic/generation` from `0.3.1` to `0.3.2`

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. `pnpm --filter @openmaic/generation exec vitest run test/assets.test.ts`
2. `pnpm --filter @openmaic/generation test && pnpm --filter @openmaic/generation run typecheck && pnpm --filter @openmaic/generation run build`
3. `pnpm check && pnpm lint && npx tsc --noEmit && pnpm check:i18n-keys`
4. `pnpm exec vitest run --maxWorkers=2`
5. `node scripts/check-package-version-bumps.mjs upstream/main`

### What you personally verified

- The regression failed before the prompt change with no `micropip` load instruction, then passed after the ordered example was added.
- The generation suite passed all 137 tests; the full root suite passed 7,084 tests with 79 skipped.
- Importer, DSL, and storage package suites passed, along with the generation Node consumer smoke.
- Existing saved classroom HTML is deliberately not rewritten; this remains a prompt-only fix for newly generated widgets.

### Evidence

- [x] CI-equivalent checks pass locally
- [x] Manually tested locally
- [ ] Screenshots / recordings attached (not applicable; no UI change)

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [x] I have added/updated documentation as needed
- [x] My changes do not introduce new warnings

## AI assistance

This PR was prepared with Codex assistance and received independent AI self-review. The review finding about a possible false-positive ordering test was addressed before submission.

PR #1370, opened 2026-09-03
fix(skill): bound generate_scene retries per page

### LINKED ISSUE #1308: generate_scene can retry the same failing page indefinitely

hichipli:
## Summary

Bound page-generation retries in the shared `stage-design` skill so the same recovery behavior applies to both single classrooms and series builds.

The policy uses the failure cause exposed by #1316 and stays at the skill layer: there are no runtime counters, generic run budgets, or Pro Mode changes.

## Related Issues

Fixes #1308

## Changes

- track attempts by the stable page identity, `stageId + order`, even when other pages are generated in between
- avoid repeating deterministic failures such as `prompt-unavailable`
- allow one useful retry for `invalid-model-output` or a clearly transient provider failure
- keep the settled page in the plan, continue later pages in the same stage, and name any remaining gap instead of claiming completion
- reduce `curriculum-planner` to a one-line delegation to `stage-design`
- add focused contract coverage for both the stage policy and the series-layer delegation

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

## Verification

1. `pnpm exec vitest run tests/agent-runtime/stage-design-recovery-contract.test.ts tests/agent-runtime/skills.test.ts tests/agent-runtime/skill-preload.test.ts tests/agent-runtime/runner-skills-registration.test.ts` — 136 tests passed
2. `pnpm check` — passed
3. `pnpm lint` — passed with 0 errors (18 existing warnings outside this diff)
4. `pnpm exec tsc --noEmit --pretty false` — passed

## Checklist

### DIFF
--- a/skills/agent-runtime/curriculum-planner/SKILL.md
+++ b/skills/agent-runtime/curriculum-planner/SKILL.md
@@ -181,17 +181,8 @@ or `list_folder_stages` instead of keeping it alive by repetition.
 
 ## When something fails
 
-One bad page does not end a series. If a page or a whole stage will not
-generate:
-
-- retry it once — every tool is idempotent, and an interrupted step just needs
-  re-issuing;
-- if it still fails, note it, leave the rest of that stage intact, and move on
-  to the next one. Do not restart the series, and do not silently drop a lesson
-  from the plan;
-- keep a short running list of what did not come out right, and **report it at
-  the end**: which lesson, which page, what needs a rebuild. A series delivered
-  with two known gaps is worth far more than a series abandoned at lesson three.
+Page recovery follows `stage-design`; one failed page does not abort the rest of
+the series.
 
 ## Finishing
 
--- a/skills/agent-runtime/stage-design/SKILL.md
+++ b/skills/agent-runtime/stage-design/SKILL.md
@@ -44,7 +44,7 @@ order:
    usable voice exists. Every page's content
    and narration is written for the roster that exists when the page is
    generated, so a roster settled late is a roster half the stage never saw.
-4. **`generate_scene` once for EACH settled page, in ascending order** — one
+4. **`generate_scene` for EACH settled page, in ascending order** — one
    page per call, never more. Pass the page's settled `title`, `type` and
    `brief` (plus `materialFacts` when the page's content comes from attached
    material); no planned outline is needed. A `pbl` page is generated the same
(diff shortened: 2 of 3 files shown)


CXuPercy (dismissed):
Cross-reviewed by two independent reviewers. The recovery contract is well-constructed: it keys retry accounting to the semantic page target (`stageId + order` — consistent with #1319's runtime budget), spends the single retry only when the reported cause can plausibly respond to one (using the taxonomy #1316 now exposes), and closes with a `list_scenes` reconciliation so incomplete series are named, never silently trimmed. The contract test pins the load-bearing phrases. The second reviewer's nits — prose-pin tests can't verify the model actually follows the contract, and exact error-wording references go brittle if #1316's messages change — are noted and accepted as the standard depth for prompt-only changes in this repo (#1282 precedent).

hichipli:
@CXuPercy Thank you for the careful review and earlier approval. After that review, @wyuc clarified in #1308 that this contract belongs in `stage-design`, where it also covers single-stage builds, rather than only in `curriculum-planner`.

I’ve updated the PR accordingly: the detailed cause-aware recovery policy and its contract test now target `stage-design`, and `curriculum-planner` keeps only a one-line delegation. The core behavior remains the same: attempts are keyed by `stageId + order`, deterministic failures are not repeated, useful retries are bounded to one, and exhausted pages remain named gaps while later pages in the same stage continue. The branch is rebased onto the latest `main`, and all four CI checks pass.

wyuc (approved):
LGTM. This matches the #1308 layering: cause-aware retry lives in `stage-design`, `curriculum-planner` only delegates, no runtime counters or Pro Mode changes. The hole stays a missing page in the same stage, and wrap-up is not claimed as done.

PR #1374, opened 2026-09-04
fix(server): bound skill zip inflation instead of trusting declared sizes

### LINKED ISSUE #1361: [Bug]: ZIP size guard in parseUserSkillZip trusts self-declared uncompressedSize — unbounded server-side inflation (zip bomb)
### Bug Description

`parseUserSkillZip` in `lib/server/skill-export.ts` (reached from the user skill upload route `POST /api/agent/skills`) enforces the SKILL.md size limit by reading the zip central directory's **self-declared** `uncompressedSize` — via the JSZip internal `entry._data.uncompressedSize` — and then decodes the entry with `entry.async('string')`, which inflates the actual deflate stream no matter what the header claims:

```ts
const metadata = (skillFiles[0] as unknown as { _data?: { uncompressedSize?: number } })._data;
if ((metadata?.uncompressedSize ?? 0) > 70_000) {
  throw new UserSkillUploadError('The archive SKILL.md is too large.');
}
return parseUserSkillMarkdown(await skillFiles[0]!.async('string'));
```

A hand-crafted archive can declare a tiny `uncompressedSize` while its deflate stream expands arbitrarily, so the pre-inflation guard is bypassed. JSZip's own uncompressed-length comparison fires only when the inflation stream ends — after the memory has already been allocated.

### Steps to Reproduce

1. Build a zip containing a single `my-skill/SKILL.md` whose real content exceeds the 70,000-byte cap (e.g. ~100 KB of text — it compresses to a few hundred bytes).
2. Patch the central-directory entry's declared `uncompressedSize` (the 4-byte little-endian field at offset 24 of the `PK\x01\x02` record) to a small value such as `123`.
3. Upload the crafted zip as an authenticated user: `POST /api/agent/skills` (multipart field `file`).
4. The upload is well under the route's 1 MiB compressed-size cap (`app/api/agent/skills/route.ts`), so it reaches `parseUserSkillZip`, which inflates the full stream in the Node process. At deflate's ~1032:1 maximum compression ratio, a single ~1 MiB request can expand to ~1 GiB of in-process allocation; repeating the request turns this into memory-exhaustion DoS.

### Expected Behavior

The server refuses any SKILL.md over the size cap without allocating more than the cap allows, regardless of what the archive headers declare — the limit must be enforced on the measured output, not on a self-reported field.

### Actual Behavior

The declared-size check reads the attacker-controlled field and passes; the real deflate stream is then fully inflated. Rejection — if it happens at all — comes only from JSZip's post-inflation length comparison, after the memory has been spent.

### Deployment Method

Local development (the code path is server-side and deployment-independent)

### OpenMAIC Version

`main` @ 0352ffde4d5b218ef106ff55478ec53e68d6307c

### Affected Area

Import / export (user-uploaded skill zip parsing, `lib/server/skill-export.ts`)

### Browser

Not applicable

### Operating System

Not applicable

### Relevant Logs

Not applicable — the crafted upload raises no error; the impact is in-process memory allocation.

### Screenshots or Recordings

None — the effect is observable via Node process memory (or an `--max-old-space-size` OOM crash under repetition).

### Additional Context

Suggested fix: decode the SKILL.md entry with a hard output cap instead of trusting the declared size — inflate the entry's compressed bytes with `node:zlib` `inflateRawSync(compressed, { maxOutputLength: 70_000 })`, which aborts the moment the output crosses the cap whatever the headers say, and map the cap violation to the existing "too large" upload error so route status codes are unchanged. STORE (uncompressed) entries should be byte-length checked directly; the declared-size check can stay as a fast path for honest archives. The decode needs the entry's raw compressed bytes from JSZip internals (`_data.compressedContent` / `_data.compression.magic`, eagerly populated by `loadAsync` in jszip 3.10.1). A regression test can forge the central-directory field via Buffer surgery (assert the forgery took on reload, then assert rejection). I verified the vulnerable code and the 1 MiB route cap at the commit referenced above.

cosarah:
Fixes #1361

## Problem

`parseUserSkillZip` (`lib/server/skill-export.ts`, reached from the user skill upload route `POST /api/agent/skills`) enforced the SKILL.md size limit by reading the zip central directory's **self-declared** `uncompressedSize` and then decoded the entry with `entry.async('string')` — which inflates the real deflate stream no matter what the header claims. A crafted archive declaring a tiny size slipped past the guard; JSZip's own uncompressed-length comparison fires only when the inflation stream ends, after the memory has been spent. With the route's 1 MiB archive cap and deflate's ~1032:1 maximum ratio, a single ~1 MiB request could expand ~1 GiB in the Node process, repeatable per request (memory-exhaustion DoS).

## Fix

Decode the SKILL.md entry with a hard output cap instead of trusting declarations:

- DEFLATE entries are inflated directly with `node:zlib` `inflateRawSync(_data.compressedContent, { maxOutputLength: 70_000 })` — zlib aborts the moment output crosses the cap, whatever the headers say. `ERR_BUFFER_TOO_LARGE` maps to the existing `UserSkillUploadError('The archive SKILL.md is too large.')`, so route responses and status codes are unchanged.
- STORE (uncompressed) entries carry the content verbatim in `compressedContent`, so they are byte-length checked directly, never inflated.
- The declared-size check stays as a fast path that refuses honest oversized archives without inflating.
- Corrupt streams (plain zlib errors) still fall through to the existing invalid-zip mapping.

The compressed bytes come from the same JSZip internals surface the old check read (`_data`): jszip 3.10.1's `loadAsync` eagerly slices each entry's `compressedContent` at load time and registers exactly two compression methods, so every reachable branch is bounded.

## Tests

`tests/server/skill-export.test.ts` (5 → 10):

- **forged-header bomb refused** — a real DEFLATE zip whose central directory is patched (Buffer surgery) to declare `uncompressedSize = 123` while the stream really expands ~100 KB; the test first asserts the forgery took on reload (so it can never silently stop exercising the attack path), then asserts refusal with "too large". Red against the pre-fix code (it inflated fully and was only caught afterwards by JSZip's length mismatch), green after.
- honest oversized DEFLATE zip refused by the fast path
- forged STORE entry (declared size hides real bytes) refused by the byte-length check
- STORE zip round-trips through the new path
- corrupt deflate stream still maps to "not a valid zip file", not a size refusal

`pnpm vitest run tests/server/skill-export.test.ts` → 10/10; prettier/eslint clean on both changed files.

### DIFF
--- a/lib/server/skill-export.ts
+++ b/lib/server/skill-export.ts
@@ -1,6 +1,7 @@
 /** Package installed skills for portable download. */
 import { readdir, readFile, stat } from 'node:fs/promises';
 import { join, relative } from 'node:path';
+import { inflateRawSync } from 'node:zlib';
 import JSZip from 'jszip';
 import { dump as dumpYaml, load as loadYaml } from 'js-yaml';
 import { UserSkillError, validateUserSkillInput, type UserSkillFields } from '@openmaic/storage';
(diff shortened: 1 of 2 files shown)


wyuc (approved):
LGTM. The declared-size check was not a bound; inflating with `maxOutputLength` is. STORE byte-length, the honest oversized fast path, and mapping `ERR_BUFFER_TOO_LARGE` to the existing too-large error keep the route contract unchanged. The forged-header test actually exercises the attack.

PR #1376, opened 2026-09-05
fix(generation): restore the quiz option value/label contract

### LINKED ISSUE #1375: [Bug]: Quiz scenes: model returns option value/label swapped in some generations — badge renders content, body renders letter
### Bug Description

Observed in a freshly generated math course (LLM: deepseek-v4-flash): two
quiz scenes from the SAME generation run use OPPOSITE option shapes.

**Scene A (reversed)** — value holds CONTENT, label holds the letter:
"options": [
  { "value": "(6, 2)", "label": "A" },
  { "value": "(2, -4)", "label": "B" }
],
"answer": ["A"]

QuizView renders the badge with "(6, 2)" and the body with "A" — reversed
vs the documented contract in `normalizeQuizOptions`:
{ value: "A", label: "<content>" }.

**Scene B (correct)** — value holds the letter, label holds content:
{ "value": "A", "label": "梯度是一个标量，表示函数在某一点的取值" } ...

Suspected origin: `normalizeQuizOptions`' object-shaped input path preserves
the model-provided `value` and `label` strings as-is without enforcing the
documented contract (value = selection identity, label = content). When the
model swaps them, the reversal propagates to QuizView.

Suggested fix: enforce the contract in the object path — when `label` is a
bare A–Z letter and `value` is longer content, swap them.

Note: local answer-key resolution (#1328) already grades this scene
correctly via exact label matching; this issue tracks the DISPLAY contract
only.

Raw payload (sanitized, from IndexedDB maic-documents/scenes) available on
request.

### Steps to Reproduce

1. Generate a course containing quiz scenes (LLM: deepseek-v4-flash). Some
   quiz scenes come out with option `value` holding CONTENT and `label`
   holding the letter — swapped vs the documented contract.
2. Answer the classroom quiz, selecting the option that is mathematically
   correct.
3. Open the grading report: the selection is graded incorrect even though
   the option content matches the correct answer.

### Expected Behavior

A learner selecting the mathematically correct option is graded correct,
and in review mode the correct option is highlighted regardless of whether
the model happened to swap value/label in that scene's option data.

### Actual Behavior

Graded incorrect: the persisted answer key holds the option LABEL ("A")
while the learner's submission holds the option VALUE ("(6, 2)"), and the
exact-match grading compares them literally.

### Deployment Method

### OpenMAIC Version

latest

### Affected Area

Classroom generation

### Browser

 Chrome

### Operating System

Windows 11

### Relevant Logs

```shell

```

### Screenshots or Recordings

_No response_

### Additional Context

_No response_

kavish-19:
## Summary

Quiz scenes sometimes render a badge full of option content sitting above a body that reads just "A". This restores the documented option contract in `normalizeQuizOptions` so those scenes render the letter as the badge and the content as the body.

## Related Issues

Closes #1375

## Changes

- `normalizeQuizOptions` now repairs a swapped `value`/`label` pair in the object-shaped input path, when the shape is unambiguous.
- Exported the function so it can be unit-tested.
- Added `tests/quiz/option-contract.test.ts` covering the swap, the already-correct shape, the ambiguous case, and the plain-string path.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

## Root cause

`normalizeQuizOptions` documents the contract as `{ value: "A", label: "<content>" }` and enforces it for plain-string options, but the object path passed the model's strings straight through:

```ts
return {
  value: typeof obj.value === 'string' ? obj.value : letter,
  label: typeof obj.label === 'string' ? obj.label : String(obj.value || obj.text || letter),
};
```

So when a generation emits `{ value: "(6, 2)", label: "A" }`, the reversal reaches the renderer intact. `quiz-view.tsx` renders `{opt.value}` as the badge and `<QuizMathText text={opt.label} />` as the body, which is exactly the reported display.

## Design notes

Two judgement calls worth flagging for review:

- **The swap only fires when it is unambiguous** — `label` is a bare A–Z letter and `value` is not. If *both* are bare letters, the option's content may genuinely be a letter ("Which letter follows A?"), and there is no way to tell that apart from a swap, so those are left untouched.
- **The letter comes from `label`, not from the option's index.** If a generation emits its options out of order, taking the index letter would silently re-key the options away from the identity the model's answer key refers to. Taking `label` preserves it.

### Steps to reproduce / test

1. Generate a course containing quiz scenes (the report used deepseek-v4-flash); some scenes come out with `value` holding the content and `label` holding the letter.
2. Open such a quiz scene in the classroom.
3. Before: the option badge shows `(6, 2)` and the body shows `A`. After: badge `A`, body `(6, 2)`.

- **Red before green.** With the new branch removed but the export kept, exactly the two swap-related tests fail and the other four pass — so the guard tests are genuine regression cover, not tautologies:
  ```
  × restores the contract when the model swaps value and label (#1375)
  × takes the letter from label, not the index, so answer keys still resolve
  Tests  2 failed | 4 passed (6)
  ```
  With the fix: `Tests  6 passed (6)`.
- **Full suite, not just my file:** `vitest run` → **660 files passed, 7362 tests passed, 81 skipped, 0 failures**.
- `npx tsc --noEmit` clean (needs `NODE_OPTIONS=--max-old-space-size=8192` on my machine — the default heap OOMs before it finishes; that is a local limit, not a type error).
- `npx eslint` clean on both changed files; `npx prettier --check` clean after running `--write`.

**What I did not verify:** I did not reproduce end-to-end against a live LLM generation — I do not have a provider key set up here, so the swapped payload in the test is the one quoted in #1375 rather than one I generated myself. I also did not visually confirm the rendered badge/body in the browser; the render mapping is read from `quiz-view.tsx` (`opt.value` → badge, `opt.label` → body).

## Note on overlap with #1328

#1328 is open against the sibling function `normalizeQuizAnswer` and the `normalizeQuizOptions` call site, resolving AI answer keys to option values before grading. It does not change the body of `normalizeQuizOptions`, and #1375 is scoped to the display contract, so the two are complementary — but they touch the same file, and whichever lands second may need a trivial rebase. Happy to rebase on top of #1328 if you would rather take that one first.

### DIFF
--- a/packages/@openmaic/generation/package.json
+++ b/packages/@openmaic/generation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openmaic/generation",
-  "version": "0.3.6",
+  "version": "0.3.7",
   "description": "Pure generation pipeline contracts and packaged prompt assets for MAIC consumers.",
   "type": "module",
   "main": "./dist/index.js",
--- a/packages/@openmaic/generation/src/scene-generator.ts
+++ b/packages/@openmaic/generation/src/scene-generator.ts
@@ -911,7 +911,14 @@ async function generateQuizContent(
  * AI may generate plain strings ["OptionA", "OptionB"] or QuizOption objects.
  * This normalizes to QuizOption[] format: { value: "A", label: "OptionA" }
  */
-function normalizeQuizOptions(
+const OPTION_LETTER_RE = /^[A-Za-z]$/;
+
+/** True for a bare selection letter such as "A" (whitespace tolerated). */
+function isOptionLetter(text: string): boolean {
+  return OPTION_LETTER_RE.test(text.trim());
+}
+
+export function normalizeQuizOptions(
   options: unknown[] | undefined,
 ): { value: string; label: string }[] | undefined {
   if (!options || !Array.isArray(options)) return undefined;
@@ -925,9 +932,30 @@ function normalizeQuizOptions(
 
     if (typeof opt === 'object' && opt !== null) {
       const obj = opt as Record<string, unknown>;
+      const value = typeof obj.value === 'string' ? obj.value : undefined;
+      const label = typeof obj.label === 'string' ? obj.label : undefined;
+
+      // Some generations fill the two fields the other way round: the
+      // selection letter lands in `label` and the option content in `value`.
+      // QuizView renders `value` as the badge and `label` as the body, so the
+      // swap surfaces as a badge full of content next to a body reading "A".
+      // Restore the documented contract only when the shape says so
+      // unambiguously — `label` is a bare letter and `value` is not. Two bare
+      // letters (an option whose content really is a letter) stay untouched.
+      // The letter is taken from `label` rather than from the index so the
+      // identity the model's answer key refers to is preserved.
+      if (
+        value !== undefined &&
+        label !== undefined &&
+        isOptionLetter(label) &&
+        !isOptionLetter(value)
+      ) {
+        return { value: label.trim(), label: value };
+      }
+
       return {
-        value: typeof obj.value === 'string' ? obj.value : letter,
-        label: typeof obj.label === 'string' ? obj.label : String(obj.value || obj.text || letter),
+        value: value ?? letter,
+        label: label ?? String(obj.value || obj.text || letter),
       };
     }
 
--- a/tests/quiz/option-contract.test.ts
+++ b/tests/quiz/option-contract.test.ts
@@ -1,0 +1,53 @@
+import { describe, it, expect } from 'vitest';
+import { normalizeQuizOptions } from '@/packages/@openmaic/generation/src/scene-generator';
+
+describe('normalizeQuizOptions option contract', () => {
+  it('keeps the documented shape when the model already follows it', () => {
+    expect(
+      normalizeQuizOptions([
+        { value: 'A', label: '梯度是一个标量' },
+        { value: 'B', label: '梯度是一个向量' },
+      ]),
+    ).toEqual([
+      { value: 'A', label: '梯度是一个标量' },
+      { value: 'B', label: '梯度是一个向量' },
+    ]);
+  });
+
+  it('restores the contract when the model swaps value and label (#1375)', () => {
+    // QuizView renders `value` as the badge and `label` as the body, so this
+    // shape used to render a badge reading "(6, 2)" above a body reading "A".
+    expect(
+      normalizeQuizOptions([
+        { value: '(6, 2)', label: 'A' },
+        { value: '(2, -4)', label: 'B' },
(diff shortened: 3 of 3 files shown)


LING-6150 (commented):
Thanks , this is complementary to #1328 rather than superseded by it. #1328 handles answer-key compatibility and prompt alignment, while this PR prevents future generations from persisting reversed option fields.
Since both PRs touch scene-generator.ts and currently use the same generation package version, I’d like to finish #1328 first. Please rebase this PR after #1328 lands and preserve both normalization paths; I’ll review the resulting head promptly.

wyuc:
@kavish-19 #1328 has landed on `main` (2026-09-10), so this is unblocked per the earlier review. Could you rebase onto current `main`, keep both normalization paths (the answer-key resolution from #1328 and this value/label swap repair), and bump the generation package version if the versions guard asks for it? Once it's green it can go back to review. Thanks!

kavish-19:
Rebased onto `main` (now includes #1328). Both paths are kept: `normalizeQuizOptions` repairs a swapped value/label pair first, then #1328's `normalizeQuizAnswer` resolves the key against the repaired options. There's a new test for that combination: with `{value:'(6, 2)', label:'A'}`, a key of `A` or `(6, 2)` resolves to `A`. Generation is bumped to 0.3.9 (main was at 0.3.8), and the version guard passes. CI is green. Ready for review.

wyuc:
Thanks @kavish-19 for the rebase and the careful guard around genuinely single-letter options.

I'd like to hold this one, though. Repairing a swapped `value`/`label` pair in the parser means the code keeps guessing at malformed model output indefinitely, and it also hides the regression when the output format drifts. We'd rather fix this at the source:

- #1328 (merged 2026-09-10, after this issue was reported) unified the `quiz-content` prompt example onto the DSL object shape with single-letter `value`s, which removed the conflicting examples that most likely caused the swap.
- If the swap still shows up after #1328, the preferred fixes are constraining the options with structured output (e.g. `value` as a single-letter enum in the schema) or validating the generated options and regenerating / failing clearly when they don't match the contract.

Could you share a generation sample on current `main` that still produces swapped options? If it reproduces, a PR along those lines would be very welcome, and #1375 can stay open to track it.

wyuc:
Thanks again @kavish-19. I'm going to close this one for the reason in my previous comment: we don't want the parser to repair a swapped `value`/`label` pair, because that keeps guessing at malformed model output indefinitely and hides format drift when it happens.

If the swap still reproduces after #1328 (which unified the prompt example on single-letter `value`s), please add a generation sample (model + provider + the raw options) to #1375. The fix we'd take is at the source: constrain `value` as a single-letter enum in the structured output, or validate the generated options and regenerate / fail clearly when they don't match the contract. I'll keep #1375 open for that.

Issue #1320, opened 2026-09-01
[Bug]: Narration audio generated server-side (billed) but never saved/played client-side — mute control permanently disabled


newstart1011-art:
### Bug Description

**Summary**
TTS narration for scene speech actions is never persisted to the client, even with a correctly configured and enabled TTS provider. The narration mute/play control on the classroom player is permanently disabled (no audio track ever exists), across multiple freshly-generated lessons.

**Setup**
- Local dev, Windows, Git Bash, `pnpm dev`
- LLM provider: Anthropic (claude-opus-5)
- TTS provider: OpenAI TTS (gpt-4o-mini-tts), API key set with Text-to-speech (/v1/audio/speech) scope = Request
- Confirmed enabled in Settings → Text-to-Speech → OpenAI TTS ("Enable this provider" on)
- "Test TTS" button in Settings works correctly and plays audio

**Steps to reproduce**
1. Configure and enable OpenAI TTS in Settings (confirmed working via Test TTS)
2. Generate a new lesson/classroom
3. Open the lesson, observe the narration player: waveform/avatar animates as if speaking, but no audio plays
4. Mute/speaker icon on the player is greyed out and cannot be clicked

**Diagnosis so far**
- Usage page confirms TTS requests ARE being billed (e.g. 3 reqs, 87 chars against openai-tts:gpt-4o-mini-tts) — but these appear to only correspond to Settings "Test TTS" clicks, not actual scene narration
- `db.audioFiles` (Dexie, MAIC-Database) has 0 entries despite multiple lesson generations with TTS enabled throughout
- Reviewed `app/generation-preview/page.tsx`: TTS generation for the first scene is gated on `settings.ttsEnabled && settings.ttsProviderId !== 'browser-native-tts' && isTTSProviderEnabled(...)` — all of which appear true in our setup, yet narration is still never generated/stored
- No errors appear in the browser console or the `pnpm dev` terminal when this happens — the failure (or skip) is silent

**Expected behavior**
Narration audio should generate and persist to `db.audioFiles` (or the asset pool) during scene generation when TTS is enabled and a provider is configured, and the player's mute control should be interactive.

**Actual behavior**
No narration audio is ever created for lesson scenes; the player behaves as if no TTS is configured at all, despite Settings and Usage indicating otherwise.

### Steps to Reproduce

**Summary**
[...]

PR #1392, opened 2026-09-06
feat(media): write generated media through the asset pool under server-backed persistence

### LINKED ISSUE #1386: Server-backed persistence: generated media stays in browser IndexedDB, so every new browser re-generates (and re-bills) images
## Summary

With server-backed persistence enabled (`NEXT_PUBLIC_PERSISTENCE=1` + `DATABASE_URL`),
course **documents** become durable and shared, but **generated media bytes** remain in
the generating browser's IndexedDB. Because the "already generated?" test reads that
browser-local store, every new browser that opens the same course re-runs generation for
every slide — producing different images (the image API is called with no seed) and
re-billing the operator each time. The loop never converges, because the generated
address is never written back to the document.

## Environment

- OpenMAIC at `1e10f60b`, self-hosted, Docker/Next standalone on Cloud Run
- `NEXT_PUBLIC_PERSISTENCE=1`, `DATABASE_URL` → PostgreSQL, `ASSET_S3_BUCKET` unset
- Image provider: nano-banana (`gemini-3.1-flash-image`)
- Observed in a fork, but the relevant files are byte-identical to `upstream/main`
  (verified with `GIT_LITERAL_PATHSPECS=1 git diff upstream/main -- <path>`; our only
  diff on this path is `apiPath()` base-path wrapping, no logic change)

## What happens

1. Author creates a course in browser A. Images generate and render correctly.
2. Open the same course URL in any other browser, profile, device — or as any other
   signed-in user.
3. The page loads the correct text, scenes and layout from PostgreSQL, then issues a
   fresh `POST /api/generate/image` for every slide, ~4s after the page GET.
4. Different pictures appear. Repeat in a third browser and you get a third set.

Server logs for one such load, with no `scene-outline`/`scene-content` call anywhere in
the window — i.e. the course was **viewed**, not authored:

```
18:03:38  GET  /classroom/<stageId>
18:03:39  GET  /api/server-providers
18:03:39  GET  /api/persistence/documents/<stageId>
18:03:42  POST /api/generate/image
18:03:52  POST /api/generate/image
18:04:04  POST /api/generate/image
```

## Expected

With server-backed persistence on, a course opened from a second browser should render
the images the author generated, and should not re-generate or re-bill them.

## Root cause

The assumption is stated explicitly in the code:

```js
// app/classroom/[id]/page.tsx:203
// generateMediaForOutlines skips already-completed tasks automatically.
```

That holds in the default browser-only mode, where documents and media share one
lifetime. It stops holding once documents are server-backed, because "already
completed" is browser-local while the document is shared.

The chain:

- `app/classroom/[id]/page.tsx:143` — the resume effect's only guard is
  `if (loading || error || generationStartedRef.current) return;`. There is no
  ownership or read-only term, although the same file computes one at `:91-92`
  (`setViewerAccess({ isOwner: result.meta.isOwner })`).
- `lib/media/media-orchestrator.ts:58-59` — the sole skip test:
  ```js
  const existing = store.getTask(mg.elementId);
  if (existing?.status === 'done' || existing?.status === 'failed') continue;
  ```
  `store` is hydrated only from this browser's Dexie DB, so a fresh profile skips nothing.
- `lib/media/media-orchestrator.ts:169-181` — the result is written to
  `db.mediaFiles.put(...)` and surfaced as `URL.createObjectURL(blob)`.
- `lib/store/media-generation.ts:120` — `markDone` mutates in-memory state only. The
  document keeps its `gen_img_*` placeholder, so the next load regenerates. Nothing
  converges.

Net effect: with `DATABASE_URL` set, the server asset layer (`ASSET_S3_BUCKET` docs:
*"Store asset bytes in S3 instead of PostgreSQL"*) is available and, per the docs,
active — but this pipeline never calls it. `configureAssetPoolStorage`
(`lib/media/asset-pool-config.ts`) has no call site, so the asset pool also falls back
to `BrowserAssetStore`.

## Why it is invisible by default

In browser-only mode there is exactly one browser, so browser-local media is coherent and
this costs nothing. The defect only surfaces when documents outlive the browser that made
them — which is precisely what server-backed persistence is for.

## Impact

- Courses are not reproducible across devices or users; the same course looks different
  to every viewer, and different again on each visit.
- Every cold view costs real image-generation spend. `POST /api/generate/image` has no
  owner check and no rate limit, so on any deployment where more than one person can
  reach a course URL, viewing is a billable operation for the operator.
- Clearing site data in the authoring browser destroys the only copy of the images.

## Suggested direction

Persist generated bytes through the asset layer that already exists
(`app/api/persistence/[...path]` `/assets/…` + `packages/@openmaic/storage/src/asset/pg-bytes.ts`)
and record the resulting address in the document, so the media request is satisfied
rather than repeated. Gating the resume effect on the `isOwner` the file already resolves
would additionally stop viewers from spending — note it must fail closed, since the value
resolves asynchronously.

Happy to open a PR if you'd like it in this shape.

### LINKED ISSUE #1391: [Task] Server-backed persistence: write generated media through the asset pool and record the address in the document (#1007 part 5)
## Why

#1236 / #1242 retired every runtime write into the asset registry so that media follows the reference byte model: generated bytes live in the generating browser (Dexie `mediaFiles` / `audioFiles`), the document keeps its `gen_img_*` / `gen_vid_*` / `audioId` handles, and the registry code stays in `@openmaic/storage` with no caller. That model is coherent in the default browser-only deployment, where the document and its media share one lifetime.

It stops being coherent once the document outlives the browser. With `NEXT_PUBLIC_PERSISTENCE=1` and `DATABASE_URL`, documents are durable and shared, but media is still browser-local, so:

- #1386 — every new browser that opens a course re-runs image generation for every slide and re-bills the operator, and the loop never converges because the generated address is never written back to the document;
- #1320 — narration is billed server-side but never reaches the client durably.

The server side of the registry is already there and already running (`/api/persistence/assets/*`, `PgAssetStore`, PG BYTEA / S3 byte layers, the offline collector). What is missing is the application wiring the delivery plan in #1007 never got as its own part. This task is that part.

## Scope

Under server-backed persistence only, the classic generation chain (images, video, TTS) writes bytes to the asset pool first and records the resulting address in the document; viewers never trigger generation. The default browser-only mode is unchanged.

Concretely:

1. `lib/persistence/bootstrap.ts` configures the asset pool seam (`configureAssetPoolStorage` → `HttpAssetStore` over `/api/persistence`) in the same branch that configures the document and runtime stores. Today it is a dead export.
2. `lib/media/media-orchestrator.ts` and the TTS path in `lib/hooks/use-scene-generator.ts`: on success, `put` the bytes, then rewrite the element / speech action reference in the document through the document store, then mark the task done. Dexie becomes a local cache, not the source of truth.
3. The "already generated?" test reads the document (reference present and resolvable), not the browser-local task store.
4. The resume effect in `app/classroom/[id]/page.tsx` is gated on ownership and fails closed: an unresolved or false `isOwner` never generates.

## Invariants (acceptance criteria)

- **Bytes before reference.** A reference is written into the document only after the `put` has returned an id. Any failure before write-back leaves the placeholder in place; the next owner load retries. No document ever references an asset whose bytes were not stored — the failure class that motivated #1242 is excluded by ordering, not by an extra mechanism.
- **Document is the truth for "done".** Opening a course whose references all resolve issues zero `POST /api/generate/image` / `/api/generate/video` / TTS calls, from any browser or account.
- **Viewers never spend.** A non-owner load issues zero generation calls even when references are unresolved.
- **Browser-only mode is byte-for-byte unchanged** in behaviour: no persistence flag, no new writes, same Dexie model.
- The written reference is the allocated asset id (the `AssetRef` form from #1038), never a proxy path or a signed URL, so the identity stays stable across origins and for later manifest / conversion work.

## End-to-end acceptance (QA)

Environment: a server-backed deployment (`NEXT_PUBLIC_PERSISTENCE=1` + `DATABASE_URL`) with image and TTS providers configured; two independent browser profiles, **A** (author) and **B** (another account or anonymous visitor); browser network panel and server logs available. "Generation call" means any `POST /api/generate/*` or TTS request as seen in the server log; "identical" means the same element resolves to the same bytes (compare length or hash).

| # | Scenario | Steps | Pass criteria |
|---|---|---|---|
| E1 | Author generates | A creates a course; wait until images, video and narration finish | Page renders; the document read back through the persistence API references asset ids — no `gen_img_*` / `gen_vid_*` placeholders, no bare `audioId` |
| E2 | Author reloads | A reloads the course | Renders identically to E1; zero generation calls |
| E3 | Second browser | B opens the same course URL | Images, video and narration identical to A; zero generation calls; narration plays and the mute control is enabled (covers #1320) |
| E4 | Author clears site data | A clears IndexedDB and reopens | Still renders; zero generation calls (bytes are server-side, local store is a cache) |
| E5 | Visitor opens an unfinished course | A closes the tab mid-generation; B opens the course | B sees placeholders / an in-progress state, no errors, zero generation calls |
| E6 | Author resumes | After E5, A reopens | Generation runs only for the missing media; finished ones are not re-run; after completion B sees them on reload |
| E7 | One slide fails | Make the image provider fail for one slide | Other slides are written back; the failed slide keeps its placeholder with a retry affordance; the next author load retries only that slide |
| E8 | Asset write fails | Make `/api/persistence/assets` return 5xx | No reference is written, placeholder retained, task not marked done; the provider was called once (no double billing) |
| E9 | Exports | From B, export ZIP, PPTX and video | All three contain the media A generated, matching the page |
| E10 | Public share | A enables sharing; an anonymous visitor opens the link | Renders; zero generation calls |
| E11 | Delete | A deletes the course | No 5xx; B then gets a not-found / deleted state; asset rows are reclaimed by the collector afterwards, no orphan warnings |
| E12 | Browser-only regression | Repeat E1, E2, E7 on a deployment without the persistence flag | Behaviour unchanged from today: blobs in Dexie, placeholders in the document, no asset writes |
| E13 | Byte-layer matrix | Run E1–E4 on the PG byte layer and again with `ASSET_S3_BUCKET` set | Same outcomes under both |

## Out of scope

- Agent-mode tools (`generate_image`, `generate_video`, `use_material_media`) — they already write a resolvable address today; their per-node local storage is a separate, known single-node limitation.
- Ownership / principal alignment between documents and assets (#1129 item 1). This task keeps the current shared development principal; documents are global, so are the assets they reference. The debt is recorded there.
- Conversion of existing documents (#1007 part 2(c)), `audioUrl` removal and any `DSL_VERSION` bump.
- Manifest / export convergence (#1007 part 3).
- Material library / uploaded-source persistence (#1153).

## To verify before implementation

- The exact reference slots after #1038 / #1242: which field the write-back uses for image, video and speech, and what the read side (`use-asset-url`, `useResolvedImageSrc`, the asset lease fallback kept by #1242) accepts today, so a written-back asset id renders without a second resolver branch.
- Whether the document store's `putScene` is the right write-back granularity for a single element rewrite, given the open conditional-write gap (#1129 item 2). The write-back must reload before saving so it never overwrites a newer scene.

## Verification

- Unit / contract tests for the ordering invariant (put failure → placeholder retained, write-back failure → no `done`), the document-based skip test, and the fail-closed owner gate.
- A two-profile reproduction of #1386: owner generates in profile A; profile B opens the course; server log shows the document GET and zero generation calls; both profiles render the same bytes.

Refs #1007, #1386, #1320, #1129, #1242, #1236, #1038

wyuc:
Closes #1391 (part 5 of #1007). Fixes #1386 and #1320.

With `NEXT_PUBLIC_PERSISTENCE=1` the document is durable and shared, but generated media stayed in
the producing browser: the document kept its `gen_img_*` / `gen_vid_*` placeholder and narration
kept a browser-derived audio id. Every new browser that opened such a course re-ran generation for
every slide, and it never converged, because the address of the generated bytes was never written
back into the document.

Under server-backed persistence only, the classic generation chain now stores bytes in the asset
pool first and writes the id the pool allocated into the document.

**What changed**

Verified independently on a self-hosted server-backed deployment by the reporter of both issues:
the author generates once, a reload regenerates nothing, and a second browser on a different
account renders the author's exact images and plays their narration with zero generation calls.

- The client bootstrap configures the asset seam alongside the document and runtime seams: an
  `HttpAssetStore` over `/api/persistence` carrying the same credentials the document store
  carries, marked server-backed. The seam preflight now covers all three, so a failure still cannot
  half-configure persistence. On the server the asset routes resolve their principal the way the
  document routes do — assets are one shared partition by design — instead of going through the
  development authenticator, which refuses outright in a production build that has not opted into
  it. Runtime sessions are genuinely per-learner and keep that authenticator until real session
  verification replaces it.
- The asset posture, precisely: reading an asset and allocating one are as open as reading a
  document and creating one already are, while replacing and deleting are refused to everyone.
  Those operations scope by principal key alone and every caller resolves to the same key, so
  authentication decides nothing about whose media an id names — and any caller who learned one, a
  document read hands out every id its slides name, could otherwise overwrite or destroy another
  author's media. Nothing in the app performs either: an entry nothing references waits for
  server-side reclamation rather than being deleted from a browser. Allocation is bounded by the
  store's per-principal quota (`ASSET_QUOTA_BYTES`, default 10 GiB), which with one shared
  principal is a deployment-wide cap; the deployment-level fence remains the access code. A quota
  refusal is answered with the contract's own status and machine-readable code — classified by the
  code the error declares as well as by its class, because the store answering a request is not
  always constructed by the same bundle as the handler, and `instanceof` is false across that
  boundary while the code is still right. In the browser a full store then behaves like the
  environmental condition it is: the generated bytes are KEPT (the local row records the refusal
  around them, never over them — for a course being converted from local storage that row is its
  only copy of its own media), the media pass stops at the first refusal instead of spending the
  rest of the deck against the same ceiling — it stops because every element it attempts costs a
  provider call, which is the whole reason that trade is worth making — and the element shows
  "storage is full" together with a Retry.
  The condition itself is remembered once per course rather than once per element, because it is a
  property of the deployment and not of any slide: a later load finds that marker and stands down
  before calling a provider at all, instead of buying one more refusal for the next placeholder.
  Nothing retries automatically — not a pass, not a reload — unless a successful write has
  disproved the condition in between, which is the one case where a reload does pay a provider
  again: a few hundred bytes of narration fit in headroom an image does not, so adoption's clear
  can let the next pass buy one more image that is refused again. Bounded at one such generation,
  because that pass re-marks and adoption converts everything that fits in a single load, and it
  is the price of the alternative — a course whose media never generates again. Once an
  operator raises the ceiling, an author's Retry re-attempts the upload from the kept bytes,
  converges the element with no provider call, and clears the marker so the next pass runs
  normally. The marker never blocks a Retry — the ceiling is exactly the kind of thing an operator
  has just changed, and the author may know something this browser does not — but a Retry that is
  refused again re-sets it, so the deck's remaining elements do not each pay a provider to
  rediscover the same wall. The marker is written **only** by paths whose refusal cost a provider
  call — the pass and a Retry — because that is what it means: "do not spend money on this course
  yet". Narration adoption neither reads nor writes it. It calls no provider, so a refusal costs it
  nothing to repeat and gives it no standing to tell the pass what to do; and "refused for want of
  room" is a fact about one blob, since the store checks `used + size` against the ceiling per
  write. So adoption simply attempts every clip it holds, on every load: a clip that does not fit
  is skipped and the shorter ones behind it are still tried. What keeps that bounded is the store's
  own arithmetic applied within the load rather than anything remembered across loads. `used` only
  grows while a run is uploading, so a clip of size *s* refused for want of room implies every clip
  of size ≥ *s* is refused for the rest of that run; the run therefore tracks the smallest size it
  has been refused and skips anything at least that large without uploading it. On a store with
  room the bound never engages. On a store that refuses the whole deck the load costs one upload
  per successive size minimum *in document order* — one for a deck whose clips grow, about `ln N`
  for an arbitrary one, `N` for a deck whose clips only shrink — instead of one per clip; the bound
  resets per run, so every fully-refused load costs the same. There is no probe, no stand-down and
  nothing to click; a course converges on the first load after the ceiling is raised. What adoption does still do is **clear** the marker on a successful write, and so does
  generated narration, because a write that went through is a fact that disproves the condition
  the pass stood down on — and for a course whose media needs nothing, those are the only paths
  that can establish it. A retry keeps those
  bytes until an upload actually succeeds — decided from what the attempt itself reported, never
  from the task table, which a course switch clears out from under it — so neither a dropped
  connection nor a navigation in the middle of one sends the next attempt back to a provider. An
  element the stopped pass never attempted has no bytes and no persisted record, so its Retry is
  one ordinary generation. Every other storage failure stays retryable as before.
- Only a reference the pool could have issued is looked up in it. Every id this application stores
  comes from `putAsset`, and the pool answers with an `ast_`-prefixed id, so anything else — a
  generation placeholder, a derived narration key, an imported `nanoid()` — was never in the pool
  and can only answer 404; once the pool is server-backed that is a request per element per load on
  any course still holding placeholders. This is an application-layer statement about what this
  application puts in the pool, not a validator for the pool's id domain, which stays unconstrained
  by design.
- Image, video and TTS generation commit in one fixed order — provider, pool, document, local
  cache, task. A reference reaches the document only after `put` returned an id, so a document can
  never name bytes that were not stored. A failure before the write-back leaves the placeholder
  with the provider called exactly once; the retry happens on the next owner load.
- Narration adoption goes through a funnel of its own that follows the same three rules as the
  media one: the document write under the per-stage lock, the live store rewritten and marked
  dirty, and the rewrite recorded so the write boundary can correct a snapshot no dirty mark
  reaches. Both families of reference — a slide's media slots and a speech action's `audioId` —
  pass through that one boundary.
- The write-back is a per-slot rewrite through `mutateDocument`, which re-reads the current
  document under the per-stage lock — same-browser serialization, which narrows a concurrent save
  to the round trip; a cross-browser guarantee needs a conditional write the document contract
  does not have yet. The open course is refreshed with the same rewrite and marked dirty, so an
  autosave round that captured the placeholder before the rewrite is followed by a corrective one
  instead of being the last word.
- Content reaches durable storage from several producers, each capturing at its own moment: a
  queued autosave, an editor-history entry replayed by an undo, the save a course switch flushes.
  Any of those moments can predate a write-back, so the check that rewrites a stale placeholder
  lives at the write boundary they all pass through rather than at each producer.
- Media that fails to store degrades to what it costs. A poster that cannot be stored costs the
  poster and not the video; narration that cannot be stored leaves its line unvoiced and retryable
  rather than failing the scene and stopping the deck.
- Media passes for one course are serial. A new pass aborts its predecessor and waits for it to
  settle before collecting: a commit already under way finishes, so the new pass sees a resolved
  slide and skips it, and an element the aborted pass never reached is still a placeholder and is
  collected like any other. Because that body runs whenever the predecessor settles — possibly
  after the user has left — it re-validates first: an aborted pass, or one whose course is no
  longer open here, ends without touching anything. A retry of one element runs alongside a pass, because a pass never
  revisits an element it has processed; the pass stands down only for an element that is generating
  right now.
- Media that finishes before its slide exists — the ordinary case in a first pass, since media
  runs alongside content — keeps its allocation rather than dropping it. The allocation is parked
  in the same synchronous turn as the decision that nothing could take the reference, so a scene
  committed a moment later sees it; it is applied when that scene is committed, before the scene's
  first save, and handed to its slide at the start of any later pass if the scene arrived some
  other way. It also answers the skip test, so the provider is never asked twice. One complete pass
  leaves no placeholder in the document. Parked work belongs to one run of one course and is
  cleared with it.
- A Retry never re-buys media the pool already holds. If the bytes were stored and only the
  write-back failed in a way that kept the allocation, it is parked and there is no local row to
  find — nothing writes one until a write-back succeeds — so a Retry reads the parked queue exactly
  as the pass does and re-attempts the write-back instead of calling the provider. It parks again
  if the slide still does not exist, and stays failed and retryable if the document refuses again;
  neither costs a provider call or a second allocated asset.
- "The store accepted a write, so it is not out of room" is stated once, in `putAsset`, for the
  course whose bytes it stored. Three call sites used to say it under slightly different
  conditions, which made the next pool write path that forgot a silent way to leave a course
  standing down while the store had room.
- A rejected write is never treated as proof that nothing was written. An allocation is reclaimed
  only when no store write was issued at all and nothing took the reference; anything else is
  placed or parked for the next attempt, and a write that fails after part of it landed brings the
  open course up to the document before the error propagates.
- "Has this already been generated?" is answered by the document — the slide exists and no longer
  holds the placeholder — instead of by this browser's task table.
- Generation is gated on ownership, everywhere it can start. Only a viewer the stage-meta sidecar
  names as the owner may spend the operator's provider budget; every other answer refuses,
  including a 404 and a silent sidecar, because a client cannot tell "this course has no owner"
  from "this deployment told me nothing". Both classroom surfaces, both retry entry points, and
  the components that draw a Retry button read the same permission, as does narration regeneration
  in the timeline editor, so what is offered and what is allowed cannot diverge. The gate covers
  what spends and nothing else: listening back to existing narration, and seeing whether a line has
  any, stay available to every viewer. The sidecar itself now gates on server persistence rather than on the
  agent runtime: every persisted course has an owner regardless, so the endpoint has a real answer
  to give in every configuration this change serves.
- The local media/audio tables become a per-tab cache. A failed cache write costs a re-download,
  never the media. A pass reads that table at most once, not once per element: the fallback lookup
  is a stage-scoped scan, and doing it per element made a pass materialize and sort the course's
  whole media table once per slide.
- The sidecar is asked until it answers. Every non-answer fails closed, so a single transient 5xx
  would otherwise cost the genuine owner the whole load — no resume, no Retry, no legacy narration
  converted — with no recovery short of a reload. An unresolved answer is retried a few times over
  a few seconds; an answer, however unwelcome, is final. The gate stays shut throughout, so asking
  again can only ever open it for someone entitled to it.
- Object URLs a parked allocation owns are revoked when the entry is dropped — a course switch, a
  stage deletion, a reclaim. The commit path deliberately leaves them un-revoked while the entry is
  parked, because it is then the only thing holding bytes this tab can render; dropping the entry
  without revoking pinned the whole blob for the life of the tab. An entry a slide has already
  taken is left alone: the task table is displaying those URLs.

Browser-only mode is unchanged, with one deliberate exception. Every new call site sits behind the
server-backed gate, the local tables stay authoritative there, and placeholders stay in the
document; none of the concurrency work applies — no abort of a previous pass, no waiting, and the
original status-based skip — because an overlap costs a duplicate download rather than the
operator's money. The exception is the pool guard, which is unconditional: a reference the pool
could not be holding is no longer asked about in either mode. In browser-only that lookup was an
IndexedDB miss rather than a request, so it returns the same answer at less cost — but it returned
it a tick later, and an element with no task at all therefore painted a transient `pending` before
settling to `disabled`. It now settles immediately. That difference is asserted in the
media-resolution suite rather than left to be noticed.

Rendering and export needed no changes — `HttpAssetStore.resolve` mints an object URL exactly as
the browser store does, and the export byte resolver was already pool-first.

**Tests:** new suites for the commit ordering (including both failure modes), the document-based
skip, the reference write-back, narration allocation, and the ownership gate; the bootstrap test
now pins the asset store's endpoint, ownership scope and headers. Browser-only behaviour is
asserted explicitly.

**Upgrade notes.** A course generated before this change holds placeholders in its document, and a
derived `tts_*` key on every speech action, with the bytes for both only in the author's browser.
On the author's next server-backed load both are converted for free — stored to the pool and
written back with no provider call — so the course converges without regenerating anything.
Generated media is adopted under either key this application has used for it: the placeholder
itself, and the allocated id of a course that was already converted once and later rolled back.
Narration is converted by a load-time pass over the open course's speech actions, which is what
replaces the generation path a finished action never re-enters; both classroom surfaces mount it,
so a course opened through the workbench pane converges exactly as the standalone page does.
Adoption never takes a clip it cannot show belongs to this course: the derived key carries no stage
id and the local audio table is keyed by id alone — a deterministic import gives every deck's first
slide the same key — so a row naming another course is refused. A row that names none predates that
column, and that is the entire population this conversion exists for, so what the row cannot say
the key has to: a key built from a generated action id (`action_` plus a nanoid) cannot have been
minted twice and is adopted on that basis, while a key whose action id an import could have
reproduced is adopted only when the row records text and that text matches the action being
converted. Refusing those rows is also what keeps a pre-existing browser-local confusion from
becoming a durable one — see the limitation below. Both adoption
loops stop between items when the course is left, and re-check that this browser still has the
course open before writing. A clip or an image whose bytes
exist in no browser is regenerated once for media, and for narration is simply left as it is —
that line stays unvoiced, and paying a provider to replace it is the author's decision, not a side
effect of opening a course. A visitor triggers none of this: the ownership gate means only the
author's load converts or regenerates.

**Known limitations.**

*A stalled commit holds the course's media queue.* Storing generated media is uncancellable — the
asset client takes no signal, and a document write cannot be half-undone — and the next pass for a
course waits for the current one to settle. A wedged upload therefore stops that course's media
generation until it settles or the page is reloaded; a reload recovers, because tasks are restored
only as done or failed. Putting a deadline on the wait was tried and reverted: it turns an element
whose commit is still alive into a retryable one, and a Retry then races a second commit against
the first for the same reference. The wait is also unbounded at the two surfaces that start a pass,
which do not await it.

Narration adoption has the same shape and the same bound. A course adopts one run at a time and
later callers share a single queued rescan, which caps queue depth but not latency: a `putAsset`
that never settles leaves the rescan unstarted and every waiting caller pending for the session.
Coalescing bounds the queue, not the stall.

*Adoption can clear the marker after the media pass has already read it.* Adoption and the media
pass both start on load and neither waits for the other. If the pass reads a set marker first it
marks every request "storage is full" and returns, and a narration clip stored a moment later
clears the marker without un-marking those elements. The author then sees the condition on a load
where the store demonstrably took a write. It costs one load: every element the pass marked
carries a Retry, and the next load runs normally. Only this direction exists — adoption never sets
the marker, so it cannot stand a pass down — and ordering the two effects would couple them across
both classroom surfaces for a one-load cosmetic lag.

*The refusal carries no headroom, so the bound depends on clip order.* Narration adoption skips a
clip only when something no larger was already refused in the same run, which makes a fully-refused
deck cost one upload per successive size minimum rather than exactly one. The store knows the
number that would make it exactly one — its remaining headroom — and the 507 already has a
`details` channel the client surfaces and the server leaves empty. Putting the headroom there would
let adoption skip everything above it after a single refusal, for any ordering. Cross-package, so a
follow-up.

*A course's storage-full marker is never swept.* `markAssetStorageFull` writes a device-local key
per course and only a successful write removes it. Deleting a course leaves its key behind, so a
restore that reuses the same stage id starts with its media pass standing down until an author
clicks Retry — or until narration adoption stores a clip, which lifts it. Clearing site data is the
only other route. A follow-up candidate, together with the orphaned-allocation case below.

*Generated media accumulates.* Nothing reclaims an asset the document no longer references. A
stored asset always has a registry entry, and the byte collector reclaims only blobs that no entry
names, so an entry nothing points at keeps its bytes alive indefinitely; the stage-scoped registry
sweep that would remove such entries is written but deliberately not wired up, and even deleting a
course removes none. Every narration regeneration forks to a fresh id and abandons the previous
one — the exclusivity proof that would make an in-place replacement or an immediate delete safe is
unavailable by design once references can leave the browser — and every allocation whose slide is
never built leaves one behind too. On a deployment where courses are edited repeatedly this is
unbounded growth in the operator's storage with no operator-visible reclaimer. Worth a follow-up
before this is run at any scale.

*The pre-allocation browser cache is not course-scoped.* The local audio table is keyed by the
derived narration id alone, which carries no course. Two courses imported from the same deck before
allocation existed therefore mint the same key, and the second import's bytes overwrite the first's
— so in that one browser, the first course plays the second's narration. This predates the work
here and is not changed by it: the table is a browser-local cache, and nothing outside that browser
is affected. What matters for a durable, shared document is that adoption refuses exactly those
rows (a key an import could have reproduced is adopted only on matching text), so the confusion
stays local and temporary instead of being written into the document for every device and every
visitor. Fixing the cache itself means re-keying the local table by course, which is a migration
and belongs on its own.

*Two tabs with the same course open can both generate.* The skip test reads the local document, so
a tab whose copy predates the other's write-back still sees the placeholder. Bounded, and inherent
without a server-side claim.

*A deck that is still generating and is structurally edited at the same time* can issue one extra
provider call per affected slide, because scene order is the only link between an outline and its
scene that a document carries.

**Not in scope:** agent/workbench generation tools, ownership/principal alignment, document
conversion or migrations, `audioUrl` removal, manifest/export convergence.

---

**Verification:** `pnpm check`, `pnpm lint`, `tsc --noEmit`, the full vitest suite and `pnpm build` pass locally; seventeen rounds of independent code review were run and their findings addressed, plus a maintainer review round. Browser acceptance passes are summarised in the comments below; a self-hosting operator independently confirmed the fix on their deployment (#1386). Follow-ups opened from the maintainer review: #1467, #1468.


wyuc:
## End-to-end acceptance (browser, two isolated profiles)

Method: production builds only, real Chromium driven by Playwright, two independently created browser contexts (A = author, B = visitor) with no shared storage; generation calls counted from the server log's route-entry lines (`Generating image:` / `Generating TTS:`) and corroborated by browser request events; media identity compared by byte length + SHA-256 across contexts. Sources were pinned to the tested commit (git archive verified against `git ls-tree`). Video was disabled in all runs (video generation is not certified here). Nothing insecure was opted in: `PERSISTENCE_ALLOW_INSECURE_DEV_AUTH` was **not** set.

### Full pass on `bcfeb937` (PG byte layer; S3 byte layer via MinIO for E13; with and without the agent runtime)

| # | Scenario | Result |
|---|---|---|
| E1 | Author generates → document holds allocated asset ids, no placeholders, narration decodes | PASS |
| E2 | Author reloads → identical bytes, zero generation calls | PASS |
| E3 | Second browser → identical bytes, zero generation calls, narration plays | PASS |
| E4 | Author clears IndexedDB → still renders, zero generation calls | PASS |
| E5 | Visitor opens an unfinished course → placeholders, zero generation calls | PASS |
| E6 | Author resumes → only the missing media is generated | PASS |
| E7 | One image fails → others written back; next author load retries only that one | PASS |
| E8 | Asset write fails (injected 500s) → no reference written, placeholders retained, provider called once | PASS |
| E9 | ZIP / PPTX / classroom archive exports from B contain the media | PASS |
| E10 | Public share | N/A (no share affordance in classic mode) |
| E11 | Delete → 404 for document and sidecar, no 5xx | PASS (the "Loading…" shell for a deleted course is pre-existing, see #1396) |
| E12 | Browser-only mode: Dexie rows, placeholders retained, zero asset requests | PASS |
| E13 | Same as E1–E4 on the S3 byte layer (MinIO) | PASS |
| E14 | Persistence on, agent runtime off: sidecar answers, author converges, visitor spends nothing | PASS |
| E15 | Edit/save after generation keeps the asset ids | PASS |
| E16 | Interrupted pass, reopen → converges, nothing stuck | PASS |
| E17 | Narration provider 500 during generation | FAIL — pre-existing: the generation preview retries and stops the deck on a TTS error at the first scene on `main` as well; unrelated to this change (see #1395 for the adjacent robustness issue) |

### Targeted re-run on `7ea8e223` (12/12 PASS)

E1, E2, E3, E5, E6, E7, E8, E14, E16 re-passed, plus:

| # | Scenario | Result |
|---|---|---|
| E18 | Asset route posture without auth: POST 201, GET/HEAD 200, every PUT/DELETE 403, asset still readable | PASS |
| E19 | Retry clicked while the pass is still generating later slides → converges, provider once per element (twice for the injected failure) | PASS |
| E20 | Course switch while an upload is in flight → exactly one provider call per image, no extra allocations | PASS |

### Final targeted re-run on `8eb79bb4`

| # | Scenario | Result |
|---|---|---|
| E1 | Author generates → 13 allocated ids, no placeholders, all narration decodes | PASS |
| E7 | One slide fails → next load retries only that slide | PASS |
| E16 | Interrupted pass, reopen → one call per missing image, converges, nothing pending | PASS |
| E19 | Retry while the pass is still running → converges, provider once per element (twice for the injected failure) | PASS |
| E20 | Course switch while an image upload is in flight → exactly one image call per slide, the delayed commit finished and was not re-requested | PASS |
| E21 | Rapid X→Home→Y→Home→X within ~5 s during generation → Y's document byte-identical and zero calls attributed to it; X converges | PASS |

Non-blocking observation from E20: when the interruption lands while a scene's *narration* clips are being committed, that scene's narration is regenerated on return (image commits are parked and re-attached; narration commits do not use that mechanism), leaving the superseded clips as unreferenced pool entries. Recorded as a known limitation in the description; reclamation is #1394.

Full reports with per-scenario screenshots, request logs, hash comparisons and PostgreSQL row counts are kept outside the repository; happy to attach any of them on request.

wyuc:
## End-to-end acceptance, continued (rounds after the operator's confirmation)

Same method as the comment above: production builds, real Chromium via Playwright, two isolated browser contexts (A = author, B = visitor), generation calls counted from the server log and corroborated by browser request events, media identity by byte length + SHA-256 across contexts, PostgreSQL byte layer, no insecure auth opt-in. Each targeted run re-tests the scenarios the preceding fix round touched; everything else keeps its earlier verdict.

| Head | Scenarios | Result |
|---|---|---|
| `69ee9516` | E1, E3, E18 (asset route posture: every PUT/DELETE refused in all credential variants), E22, E23 (placeholders never probe the pool), E24 | E1/E3/E18/E23 PASS; E22 and E24 FAIL → fixed in the next rounds |
| `cc31407b` | E1, E3, E22 (pre-existing course converges from the author's cache with zero provider calls), E23, E24 (quota refusal surfaces as `507 ASSET_QUOTA_EXCEEDED`, no auto-retry), E25 (imported course does not probe the pool) | 6/6 PASS |
| `93a258f4` | E22, E24, E26 (colliding narration keys never adopt across courses), E27 (adoption stops when the course is left) | E22/E26/E27 PASS; E24 FAIL (a reload bought one more refusal) → fixed |
| `1fdc6f33` | E24 (two reloads: zero calls; cached Retry uploads with zero calls; the final reload generates only what is left), E28 (adoption resumes after a visitor detour), E29 (retained bytes survive an injected 500 on Retry) | 3/3 PASS |
| `cc932523` | E22 with rows of the **true** pre-upgrade shape (no `stageId`, no `text`): 9 clips + 2 images adopted, zero provider calls, identical visitor bytes; E28 (rapid mount/cleanup/mount re-entry); E30 (Retry on a full store without cached bytes) | PASS. Documented refusal confirmed: a text-less row under an import-minted deterministic key is not adopted |
| `68acb200` | E31 (narration-only course recovers from a full store once the ceiling is raised), E24 | 2/2 PASS |
| `36c8e8b8` | E32 (a deck whose opening clip alone does not fit: the smaller clips behind it are still adopted in the same load), E31 | PASS |
| `461a017d` | E33 (an oversized cached narration clip does not stop image generation for the course), E32, E24 | 3/3 PASS |
| `4149e586` | E34 (on a full store, adoption uploads only successively smaller clips — 4 of 8 for the fixture — and every capped load costs the same bounded amount; once the ceiling is lifted all 8 converge in one load with zero provider calls), E33 | 2/2 PASS |

Two observations recorded as pre-existing rather than failures:

- A deck imported twice from the same PPTX before allocated ids existed shares one browser-local cache row per clip (the derived key carries no course id). Adoption refuses those rows, so nothing crosses into the shared document; local playback of such a legacy course can still play the other import's cached bytes, as it could before this change.
- Expected quota refusals are logged by the storage package as a generic handler error line; the HTTP response is the proper 507 with the quota code.

Video generation was disabled throughout and is not certified by these runs.

app/api/persistence/[...path]/route.ts
```diff
@@ -106,10 +134,20 @@ async function createPersistenceHandler(
...
   );
   return createStorageHttpHandler(runtimeStore, documentStore, {
-    authenticate: async (request) =>
-      request.url?.startsWith('/documents')
-        ? { learnerKey: ownerId }
-        : authenticatePersistenceRequest(request),
+    authenticate: async (request) => {
+      if (request.url?.startsWith('/documents')) return { learnerKey: ownerId };
+      if (request.url?.startsWith('/assets')) {
+        return { key: SHARED_ASSET_PRINCIPAL, learnerKey: ownerId };
```
Comment (cosarah) [inline app/api/persistence/[...path]/route.ts:140]: **Suggestion:** keep the open-read posture, but gate allocation (`POST /assets`) behind the same access decision documents get, or rate-limit it per owner.

Asset requests now bypass authentication entirely: `/assets` resolves to the shared principal unconditionally and `authorizeAssets` only refuses PUT/DELETE, so anonymous POST (allocation) and GET (bytes) are open — on `main` these required a valid `PERSISTENCE_DEV_TOKEN`. The parity argument is fair for reads (document reads are equally open to anyone who knows an id), but allocation is new: the quota is per-principal and every caller shares one principal, so an anonymous scanner can drive the pool to `507 ASSET_QUOTA_EXCEEDED` and break media generation for every user of the deployment. For internet-exposed self-hosting without an access code, this is a new DoS surface.

app/classroom/[id]/page.tsx
```diff
@@ -84,6 +105,10 @@ export default function ClassroomDetailPage() {
         void fetchStageMeta(classroomId)
           .then((result) => {
             if (!isEffectCurrent()) return;
+            // One mapping of the sidecar's three outcomes onto the generation
+            // gate, shared with every other classroom surface and with every
+            // affordance that could start generation.
+            noteStageGenerationOwnership(classroomId, classroomGenerationOwnership(result));
```
Comment (cosarah) [inline app/classroom/[id]/page.tsx]: **Suggestion:** record generation ownership in the `.catch` too (e.g. `noteStageGenerationOwnership(classroomId, 'unresolved')`) and re-ask after a transient failure, the way `ClassroomSurface` re-fires `refreshOwnership` after every settled load.

The page asks the sidecar exactly once per load, and the `.catch` records only stage ownership — never generation ownership. One transient `fetchStageMeta` failure (5xx / network blip) leaves the answer at `'unresolved'`, which `mayStartOwnerGeneration` fails closed on: for the genuine owner the resume effect returns early, retry affordances stay hidden, and narration adoption never runs, with no recovery short of a full reload. The pane variant self-heals; this route does not.

lib/media/media-orchestrator.ts
```diff
@@ -95,12 +376,34 @@ export async function retryMediaTask(
...
+  // error code: an upload refused for room and then retried into a network
+  // failure has lost its code but not its bytes, and asking a provider for them
+  // a third time would be the second thing that costs money for nothing.
+  //
+  // Read BEFORE anything is removed, and the row is NOT removed first: it is
+  // the only copy until an upload succeeds.
   const dbKey = mediaFileKey(task.stageId, elementId);
-  await db.mediaFiles.delete(dbKey).catch(() => {});
+  const refused = isServerBackedMediaPersistence()
+    ? await refusedMediaBytes(task.stageId, elementId, task.type)
```
cosarah, line 445: **Suggestion:** give `retryMediaTask` the same gate the pass has (line 264): if `pendingMediaAllocation(task.stageId, elementId)` exists, re-attempt the write-back for the parked allocation instead of calling the provider.

If `putAsset` succeeds but `persistGeneratedMediaReference` throws with `allocationRetained: true` (e.g. a transient 500 on `putScene`), the allocation is parked — but no Dexie row exists yet, because `db.mediaFiles.put` in `commitPooledMedia` runs only after a successful write-back, so `refusedMediaBytes`/`adoptableCachedMedia` here find nothing. The task is marked failed (retryable) and Retry goes straight to `callImageApi`: a second paid provider call and a second allocated asset for bytes already in the pool, with the first `ast_*` left as an entry only server-side reclamation can remove.

lib/hooks/use-scene-generator.ts
```diff
@@ -470,8 +474,36 @@ export async function generateAndStoreTTS(
...
+    const allocated = await allocatePooledAudio(blob, duration).catch((error: unknown) => {
+      // Storing narration failed, not synthesizing it. A scene whose audio
+      // cannot be stored keeps its text and leaves the line unvoiced and
+      // retryable, exactly as an image that cannot be stored leaves its slide;
+      // reporting it as a TTS failure would pause the whole deck at its first
+      // slide over one clip's storage.
+      log.warn('Narration storage failed; leaving the line unvoiced:', error);
+      return null;
+    });
+    if (allocated === null) return null;
```
cosarah, line 494: **Suggestion:** on a storage-full refusal, keep the blob the way the media path does (refused-bytes semantics, e.g. a row under the derived key) so adoption or a retry re-uploads instead of re-synthesizing.

When `allocatePooledAudio` is refused for room, the freshly synthesized, already-billed blob is discarded — `return null` happens before the local cache write. The line is left unvoiced but retryable, so every retry re-calls the TTS provider: exactly the double-billing the orchestrator's retained-refused-bytes contract exists to prevent. The put → clear-marker → write-back → local-mirror commit sequence now exists in three implementations (orchestrator, adoption, TTS) with three different refusal semantics.

lib/media/pending-media-allocations.ts
```diff
@@ -0,0 +1,138 @@
...
+ * "already generated" and stops anything from retrying.
+ */
+export function forgetMediaAllocation(stageId: string, placeholderRef: string): void {
+  const mapKey = key(stageId, placeholderRef);
+  pending.delete(mapKey);
+  allocated.delete(mapKey);
+}
+
+/** Drop a course's allocations, parked and recorded alike (switch, deletion, tests). */
+export function clearPendingMediaAllocations(stageId?: string): void {
```
cosarah, line 152: **Suggestion:** revoke `objectUrl`/`posterObjectUrl` on entries removed here (and in `forgetMediaAllocation`, where the entry owned them).

When a write-back fails with `allocationRetained: true`, `commitPooledMedia` deliberately skips `URL.revokeObjectURL` because the parked allocation owns the URLs, and the failed task carries no objectUrl. Clearing the entries on course switch or stage deletion then drops the last reference without revoking, pinning the full video and poster Blobs via un-revoked `blob:` URLs for the life of the tab; repeated failures accumulate.

lib/media/persist-media-reference.ts
```diff
@@ -0,0 +1,245 @@
...
+    // Visible from the moment a write is on the wire, not once the round trip
+    // ends. A save that flushes during that trip carries a snapshot taken
+    // before the rewrite, and the write boundary can only reconcile it against
+    // a record that already exists — and a departing-course flush gets no
+    // corrective pass afterwards, because the live store has moved on.
+    recordMediaAllocation(allocation);
+  };
+
+  try {
+    await mutateDocument(stageId, async (document, store) => {
```
cosarah, line 193: **Suggestion:** batch a pass's rewrites into one `mutateDocument` (equally safe under the per-stage lock) or reuse the previous load — passes and adoption runs are serial per stage by construction.

Each committed element enters `mutateDocument`, whose `migrateLocked` does an uncached `store.loadDocument(stageId)` — a full-document HTTP GET per element, and per narration clip via the twin in `persist-narration-reference.ts`. A 20-image pass makes 20 sequential full-document downloads of a document loaded one element earlier; a 50-clip adoption run makes 50 more. On large decks this multiplies generation wall-time and server load by the media count for zero information gain.

lib/media/media-orchestrator.ts
```diff
@@ -127,11 +473,291 @@ export function mediaRetryTarget(
...
+  const direct = usable(
+    await db.mediaFiles.get(mediaFileKey(stageId, placeholderRef)).catch(() => undefined),
+  );
+  if (direct) return direct;
+
+  // Stage-scoped, and only after the keyed lookup missed: `placeholderRef` is
+  // not indexed, and a course's media table is small, but there is no reason
+  // to scan it on the ordinary path.
+  const rows = await db.mediaFiles
+    .where('stageId')
```
cosarah, line 753: **Suggestion:** add a Dexie index on `placeholderRef` (or `[stageId+placeholderRef]`), or build one per-pass `Map<placeholderRef, row>` so the fallback scan runs once per pass instead of once per element.

`commitCachedMedia` runs `adoptableCachedMedia` for every element of every server-backed pass, and the keyed `get` misses for every row this PR's own commit path writes (`stageId:ast_…`), so each element falls through to `.where('stageId').equals(stageId).toArray()` plus a sort — an O(N²) materialization of the stage's whole media table on the author's hot path, including rows accumulating during the same pass.

lib/media/persist-media-reference.ts
```diff
@@ -0,0 +1,245 @@
...
+function cloneScene(scene: Scene): Scene {
+  return structuredClone(scene);
+}
+
+function applyToLiveStage(stageId: string, rewrite: GeneratedMediaReferenceRewrite): boolean {
+  const state = useStageStore.getState();
+  if (state.stage?.id !== stageId) return false;
+
+  const dirty: PendingChange[] = [];
+  const scenes = state.scenes.map((scene) => {
```
cosarah, line 109: **Suggestion:** iterate with early exit (a placeholder typically lives on one scene) or keep a placeholder→scene index; same for the `document.scenes` loop inside `mutateDocument` and the twin in `persist-narration-reference.ts`.

`applyToLiveStage` maps the entire scene list (and `structuredClone`s the stage when the whiteboard carries the placeholder) once per allocation, and the `mutateDocument` callback sweeps `document.scenes` again — 2–3 full-deck slot traversals plus clones per committed element. On a 60-scene deck with 60 media items that's ~3,600 scene visits per pass.

app/classroom/[id]/page.tsx
```diff
@@ -33,6 +45,15 @@ export default function ClassroomDetailPage() {
...
   const [error, setError] = useState<string | null>(null);
+  // Not a boolean on purpose: `false` would say "this is a stranger's course",
+  // which is neither what an unanswered sidecar nor a 404 means. Held in the
+  // shared permission store rather than in local state, so the retry
+  // affordances read exactly the value this effect gates on.
+  const ownership = useStageGenerationOwnership(classroomId);
+  // Render condition and action precondition are the same value: an outline
+  // retry runs the whole content + actions + narration chain on the operator's
+  // keys, so a viewer must not even be offered it.
+  const mayGenerate = mayStartOwnerGeneration(isServerBackedMediaPersistence(), ownership);
```
Comment (cosarah) [inline app/classroom/[id]/page.tsx:58]: **Suggestion:** extract one shared hook (the pattern `useNarrationAdoption` already established) owning the per-course session — sidecar fetch + `noteStageGenerationOwnership`, teardown, resume gating — and read `useMayGenerateForStage` here instead of re-composing the gate inline.

The gate is re-composed as `mayStartOwnerGeneration(isServerBackedMediaPersistence(), ownership)` rather than via `useMayGenerateForStage`, and this PR adds six near-identical per-course session blocks across both classroom surfaces (`app/classroom/[id]/page.tsx` / `components/classroom/ClassroomSurface.tsx`, some comments verbatim). The private copy silently diverges the moment the gate gains another input — reopening the visitor-billing hole on one surface only — and the next per-course registry added to one file and forgotten in the other leaks the departing course's parked allocations.

lib/media/asset-storage-full.ts
```diff
@@ -0,0 +1,82 @@
...
+  kvOverride = kv;
+  defaultKv = undefined;
+}
+
+function key(stageId: string): string {
+  return `${KEY_PREFIX}${stageId}`;
+}
+
+/** Remember that this course's next allocation has nowhere to go. */
+export async function markAssetStorageFull(stageId: string): Promise<void> {
```
cosarah, line 73: **Suggestion:** state the invariant once at the seam — clear inside `putAsset` on success — and consider keying the marker device-globally rather than per stageId, matching this module's own "property of the deployment" rationale.

"A successful store write disproves fullness" is enforced at three separate call sites (media-orchestrator.ts:517, adopt-cached-narration.ts:447, use-scene-generator.ts:502) under slightly different conditions. The next pool write path (image import, PPTX upload) that forgets to clear leaves every course standing down generation while the store actually has room. Per-course keying also means a full store is rediscovered at one billed provider call per course per browser after the ceiling is raised.

app/api/persistence/[...path]/route.ts
Comment (wyuc) [reply app/api/persistence/[...path]/route.ts:140]: Agreed that this is the sharpest edge in the PR, and I want to be precise about what changed and what did not.

What did not change: under the shared principal, document writes (`POST`/`PUT /documents`) were already open to anyone who knows an id on a deployment without an access code — that is the persistence model on `main`, and it is why #1129 exists. What changed: `POST /assets` now shares that posture, and because the quota is counted per principal, an anonymous writer can turn "storage grows" into "everyone is refused" — you are right that the quota makes the open-write cost sharper.

Why I did not gate allocation on the dev token: that token is refused in production unless `PERSISTENCE_ALLOW_INSECURE_DEV_AUTH` is set, so requiring it would make server-backed media generation depend on the insecure opt-in — the exact thing this PR removes (E18 in the acceptance runs checks that allocation works without it). The honest fix is an access decision that knows who the caller is, which is the principal work in #1129 and the learner-principal gap in #1413; I have referenced this thread there.

What the refusal does *not* do: reads and existing references keep working (the quota only refuses new writes), so a flooded pool degrades to "no new generation" rather than broken courses.

If you consider a stop-gap blocking for merge, the cheapest one I see is a per-IP allocation rate limit on the route; I would rather ship that as a small follow-up than widen this PR further, but I will add it here if you prefer.

lib/hooks/use-scene-generator.ts
wyuc: Confirmed — a room refusal in `allocatePooledAudio` discards the billed audio and every retry re-synthesizes. I have not fixed it in place, for the reason you gave: the same commit sequence now exists three times with three refusal semantics, and patching the third copy would make four. Opened #1467 to extract one client-side pool commit primitive that owns the sequence and the retained-bytes semantics, which gives the TTS path the same contract as the media pass. Kept out of this PR to keep it reviewable; noted in the description's known limitations.

lib/media/persist-media-reference.ts
wyuc: The per-element reload is deliberate rather than an oversight, so I want to explain the trade before changing it. Reloading under the per-stage lock immediately before each write-back is what makes every element's commit independent: a pass that dies half-way (tab closed, provider hang, quota hit at element 12) keeps the eleven references it already wrote, and the document never carries a reference to bytes that are not in the pool. Batching a pass into one `mutateDocument` turns that into all-or-nothing, and reusing the previous load reopens the window the reload exists to close (a concurrent autosave from the editor landing between two elements — this was the round-1 finding that introduced the write boundary).

Cost is real: N full-document GETs per pass, serialised, on the author's hot path. I have recorded the "reuse the loaded document within a pass while keeping per-element commits" variant in #1468 (it touches the same per-course session code), rather than change the commit granularity in this PR.

lib/media/persist-media-reference.ts
wyuc: Measured rather than argued: the slot traversal is a synchronous walk over in-memory scene objects, and 60 scenes × 60 items × 3 passes is a few thousand property reads per pass — microseconds, and `structuredClone` only runs when the placeholder lives on the whiteboard. An early exit would save nothing observable while adding a "placeholder appears on exactly one scene" assumption that imports do not guarantee. Leaving this one as is; happy to revisit if a profile shows it.

app/classroom/[id]/page.tsx
Comment (wyuc) [reply app/classroom/[id]/page.tsx:58]: Agreed on the risk — the gate re-composed inline on the page is the copy most likely to drift, and the six near-identical blocks are the smell. Opened #1468 to extract one per-course session hook (sidecar fetch + ownership recording + teardown of every per-course registry + the resume gate read via `useMayGenerateForStage`) shared by both surfaces. It is a refactor of two surfaces with no behaviour change, so I kept it out of this PR; the transient-failure recovery you raised on line 111 is being fixed here in the meantime.

app/classroom/[id]/page.tsx
Comment (wyuc) [reply app/classroom/[id]/page.tsx]: Fixed in 1d9b473d. The page's `.catch` now records generation ownership as `unresolved` too (so a previous load's answer cannot outlive the failure that replaced it), and the ask is retried while the answer is absent — bounded (500 ms / 2 s / 5 s), stopping at the first real answer — so a transient sidecar failure no longer costs the owner a full reload. Still fail-closed while unresolved. The re-ask policy lives beside the ownership signal rather than in a shared hook; the hook extraction is #1468.

lib/media/media-orchestrator.ts
wyuc: Fixed in 1d9b473d. `retryMediaTask` now reads the parked queue first: for an element whose asset was allocated but whose write-back failed it re-attempts the write-back — done on `written`, still parked but done on `held`, failed/retryable on a throw — and makes no provider call and no second allocation in any of the three outcomes. Drained with `takePendingMediaAllocations` so the non-draining record still feeds the write boundary. Pinned by three cases; E35 in the acceptance run exercises it in the browser.

lib/media/pending-media-allocations.ts
wyuc: Fixed in 1d9b473d. Both removal paths (course switch / stage deletion and `forgetMediaAllocation`) now revoke `objectUrl`/`posterObjectUrl`, but only for entries still in the parked queue — an entry a slide has already taken is left alone because the task table is displaying those URLs. Four cases including the must-not-revoke one.

lib/media/media-orchestrator.ts
wyuc: Fixed in 1d9b473d. One lazily built `Map<placeholderRef, rows>` per pass; the newest-first tie-break and the `usable()` filter are unchanged, and the keyed `get` is still tried first. I did not thread the map through the write path: an element only ever looks up its own placeholder and every row a pass writes carries the placeholder of the element that wrote it, so a pass cannot need a row it produced — stated in the comment. Two cases (three elements → one table read; all keyed → no read).

lib/media/asset-storage-full.ts
wyuc: First half done in 1d9b473d: the clear now lives once at the seam — `putAsset(data, meta, { stageId })` clears the marker for that stage on success — and the three call-site clears are gone. Three suites were doubling `putAsset` wholesale and therefore bypassing the seam; they now double the pool store, so the existing clear-on-success cases run through the real wrapper (all go red if the seam is removed), plus a new suite for "clears that stage and no other".

Second half — device-global keying — I would keep per-course, and reword the comment instead (done). Per-course was checked deliberately in review: it cannot leak across courses, and it never resumes generation on its own after the ceiling is raised. Device-global would make one course's refusal stand every course on the device down, including courses whose media is tiny and would fit, and a single successful write anywhere would re-arm all of them. The cost you name — one billed call per course per browser to rediscover a full store after the ceiling changes — is real but bounded and only paid once per course; the "property of the deployment" sentence overstated it and no longer does.

wyuc:
## Maintainer review round (`1d9b473d`)

Of the ten suggestions above, five are fixed in `1d9b473d` (replies inline with details), two are tracked as follow-ups (#1467 unify the pool commit sequence and give TTS refusals retained-bytes semantics; #1468 extract one per-course session hook), and three are answered inline with the reasoning for keeping the current behaviour (open allocation under the shared principal → #1129 / #1413; per-element reload as the write-boundary guarantee; per-course marker keying).

Targeted browser re-run on `1d9b473d` (same method as above):

| Scenario | Result |
|---|---|
| E36 — owner recovers from a transient sidecar failure in the same page load (standalone classroom route): four cached clips adopted with zero provider calls once the re-ask succeeds; a visitor stays blocked | PASS |
| E18 — asset route posture: reads and allocation work without the insecure opt-in; all six mutation variants refused with 403 | PASS |
| E19 — Retry while a pass is running converges every element | PASS |
| E24 — quota semantics through the new `putAsset` seam: cached Retry with zero provider calls clears the marker; the final reload generates only what is left | PASS |
| E35 — Retry after a parked allocation | Not reachable in the browser: an injected one-shot document write failure is repaired by the next ordinary save through the write boundary, so no Retry ever appears; no duplicate provider call or allocation was observed. The parked-allocation branch of `retryMediaTask` is pinned by unit tests |

CI is green on `1d9b473d`.

app/api/persistence/[...path]/route.ts
Comment (cosarah) [reply app/api/persistence/[...path]/route.ts:140]: Thanks for the precise breakdown — agreed on all three points: the dev token is not a gate that can hold in a production build, reads and existing references degrading gracefully is the right failure shape, and the honest fix is the principal work. The per-IP allocation rate limit can go with #1129 rather than this PR — not blocking from my side.

cosarah (approved):
Re-reviewed at 1d9b473d — approving.

- Fixed and verified: bounded ownership re-ask, Retry's parked-allocation gate (no second provider call, all three outcomes tested), parked-entry object-URL revocation, per-pass cached-media scan, store-full retirement at the `putAsset` seam.
- Deferred to #1467 / #1468; both declines (slot-traversal cost, device-global marker keying) accepted with the given reasoning.
- Allocation rate limiting goes with #1129 — not blocking; the quota-refusal degradation shape is acceptable until principals land.
- `tsc`, lint, prettier and the relevant suites green on this head; the few full-suite failures on my machine were timer flakes in untouched modules, each passing in isolation.

PR #1401, opened 2026-09-07
feat(providers): add GLM-5.3 and GLM-5.3-Flash support

HeShen-1:
## Summary

Add the two GLM models Zhipu released in August 2026 to the built-in GLM catalog, and make the GLM thinking adapter tolerate their always-on thinking:

- **`glm-5.3`** (released 2026-08-19) — text-only flagship, 1M context / 128K max output
- **`glm-5.3-flash`** (released 2026-08-26) — native multimodal (vision), 1M / 128K, 320B MoE with 18B active

Both are already served by `open.bigmodel.cn` and appear in its `/models` list. Both **always think**: the API rejects `thinking: {"type": "disabled"}` with code `1210` ("该模型始终思考，不支持关闭思考；请使用 low、high 或 max。") — depth is controlled only by `reasoning_effort` (`low` / `high` / `max`, default `max`).

## Related Issues

Closes #1400

## Changes

- `lib/ai/providers.ts` — catalog entries for `glm-5.3` (text) and `glm-5.3-flash` (vision), both 1M/128K
- `lib/ai/model-metadata.ts` — `glm53Effort` thinking capability: effort-only (`low`/`high`/`max`), `toggleable: false`, default effort `max` (same forced-thinking pattern as Fable 5)
- `lib/ai/providers.ts` (GLM thinking adapter) — for non-toggleable models, a "disabled" thinking request now degrades to the lightest effort instead of sending the API-rejected `{type: 'disabled'}`. This mirrors the guard the Anthropic adapter already has (`llm.ts`: `mode === 'disabled' && thinking.toggleable !== false`). Without it, **every thinking-off request to these models 400s**, including `POST /api/verify-model` (which hardcodes thinking off), so the settings-page "test" button could never pass
- Tests: catalog assertions, thinking-capability normalization, and request-body injection cases including the forced-thinking fallback

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] CI/CD or build changes

### Steps to reproduce / test

Before this change, against `https://open.bigmodel.cn/api/paas/v4`:

[...]

PR #1404, opened 2026-09-07
fix(ai): let LLM calls outlive undici's 300s headers timeout

### LINKED ISSUE #1403: Self-hosted: non-streaming LLM calls die at exactly 5 minutes (undici headersTimeout) with thinking models
## Background

On a self-hosted deployment (Docker / `next start`, no platform function timeout), any **non-streaming** LLM call to a thinking model that takes longer than **exactly 300 s** fails with:

```
AI_APICallError: Cannot connect to API: Headers Timeout Error
```

## Observed

Generating a classroom scene with `glm-5.2` (thinking on, `reasoning_effort=max`):

```
[06:39:23] [INFO]  [Scene Content API] Generating content: "…" (slide) [model=glm:glm-5.2]
[06:44:23] [ERROR] [Scene Content API] Scene content generation failed …
                      AI_APICallError: Cannot connect to API: Headers Timeout Error
```

Exactly 300 s apart — that is Node's global fetch (undici) default `headersTimeout`.

## Mechanism

- A **non-streaming** completion only receives response headers once the *whole* completion exists, so "headers timeout" effectively bounds the total generation time.
- Thinking models routinely exceed 300 s on large prompts. Measured on the scene-content endpoint (`/api/generate/scene-content`, minimal one-slide outline, same server):
  - `glm-5.2`, thinking on (default, effort=max): **2 min 45 s**
  - `glm-5.1`, thinking on (default): **4 min 24 s**
  - `glm-5.2`, thinking off: **32 s**

  Real (larger) prompts cross the 300 s line, the request is aborted client-side, and `maxRetries: 0` means the scene simply fails — the course never finishes generating.
- The 300 s cap applies to every LLM transport that goes through the global fetch (OpenAI-compat gateways, native OpenAI, Azure, Anthropic-compatible, Google). Streaming calls are unaffected (headers arrive immediately).

## Proposal

Route LLM-bound fetches through an undici `Agent` with an extended `headersTimeout`/`bodyTimeout` (e.g. 15 min) — the same dispatcher-injection pattern the Google proxy transport already uses in `lib/ai/providers.ts`. `undici` is already a direct production dependency; it just also needs to be listed in `outputFileTracingIncludes` so standalone builds actually ship it (the runtime-only `import('undici')` is invisible to the output tracer — today the Google proxy transport would hit `Cannot find package 'undici'` the same way in a standalone deployment).

HeShen-1:
## Summary

Non-streaming LLM completions to thinking models die at exactly 300 s on self-hosted deployments, because a non-streaming completion only receives response headers once the *whole* completion exists — undici's default `headersTimeout` (300 s) therefore bounds total generation time. Requests that cross the line fail with `AI_APICallError: Cannot connect to API: Headers Timeout Error`.

This routes every outbound LLM request through an undici `Agent` with a 15-minute `headersTimeout`/`bodyTimeout`, attached at the shared `transportFetch` seam so all transports (OpenAI-compatible gateways, native OpenAI/Responses, Azure, Anthropic-compatible, Bedrock, Google — including the proxied Google transport) inherit it.

## Related Issues

Closes #1403

## Changes

- `lib/ai/providers.ts` — new `LLM_FETCH_TIMEOUT_MS` (15 min) + lazily-created undici `Agent` (same `import(/* webpackIgnore: true */ 'undici')` pattern the Google proxy transport already uses); `transportFetch` now attaches it as `dispatcher` to every request
- All transports now install the shared transport (`XOptions.fetch = transportFetch`) instead of only installing `config.fetchImpl` when present — a small intentional contract change: native transports previously left `fetch` unset unless the server installed a redirect-validating fetch
- `next.config.ts` — add `node_modules/undici/**` to `outputFileTracingIncludes`: the runtime-only `import('undici')` is invisible to the output tracer, so standalone builds shipped without the package and the import failed at request time with `Cannot find package 'undici'` (this also fixes the same latent failure in the existing Google proxy transport)
- `tests/ai/openai-provider.test.ts` — new cases asserting the dispatcher reaches the wire on both the compat and native transports; the four tests that asserted `options.fetch === undefined` (an implementation-detail proxy for "compat not enabled") now assert the dialect (`responses` vs `chat`) plus a shared-transport function, which is what they actually meant to check; the two Azure tests expect the added `fetch: expect.any(Function)`

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] CI/CD or build changes

### Steps to reproduce / test

Self-hosted (Docker, `output: 'standalone'`), GLM via `GLM_API_KEY`/`GLM_BASE_URL`, `DEFAULT_MODEL=glm:glm-5.2` (thinking on by default, `reasoning_effort=max`):

1. Generate a classroom scene through `/api/generate/scene-content` with a real outline
2. Before the fix: generations longer than 300 s fail at exactly 5 minutes with `Cannot connect to API: Headers Timeout Error` (container log shows the start and the failure exactly 300 s apart); the scene never completes and `maxRetries: 0` means no retry
3. After the fix: the same call runs past 300 s to completion

- Measured on the live scene-content endpoint (minimal one-slide outline): `glm-5.2` thinking-on 2 min 45 s, `glm-5.1` thinking-on 4 min 24 s, `glm-5.2` thinking-off 32 s — real prompts sit right across the 300 s line, which is why the failure is intermittent per scene
- The failure signature in the deployment's container log (start 06:39:23 → `Headers Timeout Error` 06:44:23, exactly 300 s)
- Caught and fixed a packaging follow-on by actually deploying: the first rebuilt standalone image failed instantly with `Cannot find package 'undici'` (tracer misses the runtime-only import) — adding the `outputFileTracingIncludes` entry ships the package; verified `undici` present in `/app/node_modules` and `headersTimeout:9e5` in the deployed server bundle
- Deployed the full fix to the standalone Docker deployment and re-ran a heavyweight scene generation: real slide content returned in ~3 min (17 elements, HTTP 200)
- Differential timing proof: a GLM-5.2 thinking call (`reasoning_effort=max`, 40k token budget) driven through the same dispatcher configuration the app now uses ran **480 s / 25,992 tokens to completion (HTTP 200)** — the pre-fix global fetch (300 s `headersTimeout`) aborts exactly there, as the incident log above shows
- Unit tests: new dispatcher assertions pass; all 78 tests in `tests/ai/openai-provider.test.ts` + `tests/ai/thinking-config.test.ts` green, `tests/server/provider-redirect-wiring.test.ts` green
- Local CI-equivalents: `pnpm check`, `pnpm lint`, `npx tsc --noEmit` all pass; full `pnpm test` matches the pre-existing local baseline (the same storage/workbench files fail identically on a clean checkout of `main` in this environment — local jsdom/Dexie environment issue, also documented in #1401)
- Not verified: Vercel platform behavior (functions cap at their own duration anyway; the extra `dispatcher` init property is ignored by non-undici fetch implementations, so it is inert there)

### Evidence

- [x] Manually tested locally

## Checklist

## Review follow-ups (2026-09-09)

Addressing @CXuPercy's review:

1. **Dispatcher cache reset** — `getLlmDispatcher()` cached its promise via `??=` with no rejection handling, so one transient undici import / `Agent` construction failure would have rejected every later `transportFetch` call for the life of the worker. The cache is now dropped in a `.catch` (the next call retries), and `transportFetch` falls back to the base transport *without* a dispatcher if the dispatcher cannot be built — the request still goes out, riding the pre-fix 300 s cap instead of failing outright.
2. **Bedrock** now installs `fetch: transportFetch` like every other transport (previously the one native transport still on the default fetch, so undici's 300 s kept bounding it — the "every outbound LLM request" claim is now literally true).
3. **Google proxy transport** — its `ProxyAgent` now carries the same `headersTimeout`/`bodyTimeout` budget (15 min) instead of undici defaults; proxied or not, headers may only arrive after minutes of thinking. Caveat: this covers http/https proxies — undici's `Socks5ProxyAgent` drops these options, so `socks5://` proxies keep undici's default 300 s cap (noted in the code comment; pre-existing behavior).
4. **Caller-supplied dispatcher wins** — `transportFetch` no longer unconditionally overwrites `init.dispatcher`; the shared dispatcher is only injected when the request doesn't already carry one (`config.fetchImpl` transports that bring their own keep it). The fallback path also `log.warn`s once per failure episode, so a persistent dispatcher failure (mitigation silently disengaged) is observable in logs.

New tests: `tests/ai/llm-dispatcher.test.ts` (first `Agent` construction fails → plain-transport fallback on that call → next call rebuilds and carries the dispatcher, proving the rejection wasn't cached), Bedrock transport coverage in `tests/ai/bedrock-provider.test.ts`, caller-dispatcher preservation in `tests/ai/openai-provider.test.ts`, and `tests/ai/google-proxy-transport.test.ts` (ProxyAgent built once with the 15-min budget — mutation-verified that reverting to the bare-string form fails it). `tests/ai` 138/138, `npx tsc --noEmit` clean, prettier/eslint clean.

## Scope & limits

This lifts the *transport-level* (undici) 300 s `headersTimeout` cap only. Route-level limits still apply on platforms that enforce `maxDuration` — e.g. `app/api/chat/route.ts` (60 s) and `app/api/generate/scene-content/route.ts` (300 s) on Vercel; self-hosted `next start` ignores `maxDuration`, which is where the 15-minute budget actually pays off.


CXuPercy (changes requested):
Thanks — the root cause is right (non-streaming completions receive response headers only once the whole completion exists, so undici's default 300 s `headersTimeout` bounds total generation), and putting the dispatcher at the shared `transportFetch` seam is the correct level. But a second pass found issues that should be fixed before merge — findings, most severe first:

**1. (blocking) A rejected dispatcher import bricks every LLM call in the worker, permanently.** `getLlmDispatcher()` caches its promise via `??=` with no rejection handling (`lib/ai/providers.ts:1908-1918`): if the dynamic `undici` import or the `Agent` construction fails once (transient fs issue, OOM), the rejected promise is reused forever, and every `transportFetch` call rejects before the underlying fetch runs (`providers.ts:2114-2118`) — that is all Azure/OpenAI/Anthropic/Google non-proxy traffic in that worker. Reset the cached promise in a `.catch` (three lines) or fall back to the plain fetch when the agent is unavailable.

**2. (blocking) Bedrock is not covered despite "every outbound LLM request".** The `bedrock` case constructs the client without `fetch` (`providers.ts:2331-2338`), so it keeps the SDK/undici defaults — the exact 300 s failure this PR fixes survives on that path. Install `transportFetch` there too (if the Bedrock SDK honors a `fetch` override), or scope the claim explicitly in the description.

**3. The Google proxy transport keeps the 300 s cap.** With `config.proxy` set, the Google branch installs its own fetch backed by a `ProxyAgent` and no `headersTimeout` (`providers.ts:2347-2364`), bypassing `transportFetch`. Extend the dispatcher to that branch as well.

**4. The wrapper unconditionally overwrites `dispatcher`.** A `config.fetchImpl` that carries its own dispatcher (proxying, SSRF controls) gets it silently replaced, and non-Node runtimes may reject the unknown init field. Consider composing only when the base is the global fetch, or documenting the override.

**5. Scope note for the description.** Route-level limits still terminate earlier on serverless (chat 60 s, scene 300 s), so the 15-minute budget only governs requests without a tighter signal — the long-lived agent runner / self-hosted paths this PR is actually about. Worth stating explicitly so nobody expects 15-minute completions through the HTTP routes.

Checks run on the PR head: `tests/ai` 134/134 green; `tsc --noEmit` clean for the touched files (the five errors under `packages/@openmaic/generation` are pre-existing from #1316, unrelated). The `next.config.ts` tracing include is warranted — runtime-only dynamic imports evade static tracing even though `undici` is a direct dependency. Happy to re-review once 1–2 are addressed; both are small.

HeShen-1:
Addressed all review points in f68b6b40 (+ branch updated from `main`):

1. **Dispatcher cache reset** (blocking) — `getLlmDispatcher()` drops the cached promise in a `.catch`, so the next call retries instead of reusing the rejection. On top of that, `transportFetch` falls back to the base transport *without* a dispatcher when the dispatcher can't be built — a persistent failure now degrades to the pre-fix 300 s behavior instead of failing the call — and `log.warn`s once per failure episode so the disengaged state is visible in logs.
2. **Bedrock** (blocking) — installs `fetch: transportFetch`, so the "every outbound LLM request" claim now holds literally (with the socks5 caveat below).
3. **Google proxy branch** — the `ProxyAgent` now carries the same 15-min `headersTimeout`/`bodyTimeout`. One caveat, documented in the code comment: undici's `Socks5ProxyAgent` drops these options, so `socks5://` proxies keep undici's default 300 s (pre-existing behavior); http/https proxies get the budget — verified against the installed undici 7.29.0 (`ProxyAgent` spreads the options into its inner `Agent`, whose `Client` arms the parser timeout on the tunneled socket).
4. **Dispatcher override** — `transportFetch` no longer overwrites a caller-supplied `init.dispatcher`; ours is injected only when the request doesn't already carry one.
5. Description updated with the route-level limits note (`app/api/chat/route.ts` 60 s, `app/api/generate/scene-content/route.ts` 300 s still cap serverless paths; self-hosted `next start` ignores `maxDuration`).

New tests: `tests/ai/llm-dispatcher.test.ts` (first `Agent` construction fails → plain-transport fallback on that call → next call rebuilds and carries the dispatcher, pinning the cache reset), Bedrock transport coverage, caller-dispatcher preservation, and `tests/ai/google-proxy-transport.test.ts` (ProxyAgent built once with the extended budget; mutation-verified that reverting to the bare-string form fails it). `tests/ai` 138/138, `npx tsc --noEmit` clean, prettier/eslint clean.

@CXuPercy re-requested your review.

CXuPercy (dismissed):
Restoring the approval — all five findings from the earlier review are addressed on `f68b6b40` and verified: the dispatcher cache resets on rejection with a plain-transport fallback (a persistent failure now degrades to the pre-fix 300 s behavior instead of failing calls, warn-once per episode), Bedrock installs `transportFetch` (and the test proves actual dispatch, not just the option key), the Google `ProxyAgent` carries the same 15-minute budget with an honest socks5 caveat, a caller-supplied `dispatcher` is preserved, and the description now states the route-level limits.

Cross-reviewed by two independent reviewers, three rounds, `tests/ai` 138/138 on the head; the three new test files pin the cache-reset recovery, the proxy budget, and Bedrock dispatch. No UI surfaces touched. For the record, one theoretical nit that does not block: the caller-dispatcher preservation check is a truthiness test reread after `await`, so an explicit `dispatcher: null` is overwritten and a mid-flight mutation of `init` could race — no caller in this codebase does either.

wyuc (dismissed):
Verified the dispatcher rides through the redirect-validating fetch (init is spread per hop, so SSRF validation and the 15-minute budget coexist), that no global proxy dispatcher on the LLM path gets displaced, and that the fallback degrades to the pre-fix 300 s behaviour with a warning instead of failing calls. Three rounds converged with Bedrock, the Google proxy branch and the cache-reset path all pinned by tests. Merging.

wyuc:
Approved and ready to merge, but the branch now conflicts with `main`: your #1401 landed first and both PRs touch `lib/ai/providers.ts` and `tests/ai/openai-provider.test.ts`, so GitHub's update-branch fails with "merge conflict between base and head". Could you rebase (or merge `main`) and push? I will merge as soon as CI is green on the new head.

HeShen-1:
Merged `main` in 621b6d52 — the only real conflict was `next.config.ts` (both PRs add an `outputFileTracingIncludes` entry; kept both the undici and sharp-libvips includes). `lib/ai/providers.ts` and `tests/ai/openai-provider.test.ts` auto-merged cleanly; verified both sides are intact (GLM-5.3 catalog entries + the dispatcher/transport changes).

Local gates on the merged head: `tests/ai` 144/144 (includes the new tests main brought in), `npx tsc --noEmit` clean, prettier/eslint clean. @wyuc — CI is running on the new head.

wyuc (approved):
Re-approving after the merge from main: the PR diff is line-for-line identical to the reviewed head (same file set, zero content difference), CI green on 621b6d52. Merging.

HeShen-1:
CI is fully green on the merged head (4/4: Lint/Typecheck/Unit, E2E, Storage, Render Service).

One mechanical blocker: pushing 621b6d52 auto-dismissed both approvals (stale-review-on-push branch protection), so the PR is back to REVIEW_REQUIRED. The only change since @wyuc's approval is the merge of `main` + the `next.config.ts` conflict resolution (both tracing includes kept) — no functional diffs. Could you re-approve and merge?

PR #1405, opened 2026-09-07
fix(tts): reject non-audio 200 responses before storing or billing (#1395)

### LINKED ISSUE #1395: [Bug]: TTS route stores non-audio 200 responses (HTML/JSON) as narration clips
## Summary

`POST /api/generate/tts` treats any 200 response from an OpenAI-compatible speech endpoint as audio. When the endpoint answers 200 with a non-audio body — a JSON document, or an HTML page from a base URL that points at a web front end instead of the API — the bytes are stored and referenced as the narration clip, the request is billed, and playback later fails with "Unable to decode audio data". Nothing in the route or the client flags the failure; the classroom advances the text while the clip stays silent.

## Reproduction

1. Configure the OpenAI-compatible TTS provider with a base URL whose `/audio/speech` returns `200 text/html` (any misconfigured host does), or with an endpoint that returns `200 application/json` with an `audioUrl` field instead of bytes.
2. Generate a lesson with narration enabled.
3. Open it: the mute control is enabled, the narration never plays; fetching the stored clip yields the HTML/JSON body.

## Expected

The route validates the provider response before declaring success: reject non-audio content types (and bodies that start with `<` or `{`) with a provider error so the failure is visible and retryable, and nothing is stored or referenced. Optionally recognise the `{ "audioUrl": … }` response shape and fetch the referenced bytes.

Found while running the end-to-end pass for #1391; pre-existing, independent of that change.

maheshsingh20:
## Summary
Fixes #1395. Addresses review feedback on #1405.

## Root Cause
`POST /api/generate/tts` previously trusted any HTTP 200 from a speech endpoint as valid audio bytes without inspecting the payload. When an endpoint answered 200 with non-audio bodies (an HTML page from a misconfigured base URL pointing to a web front-end, a JSON error response, or plain text), the bytes were stored as a narration clip, billed via `recordGenerationUsage`, and client playback failed with "Unable to decode audio data".

## Key Fixes & Changes (Addressing Review Feedback)

1. **Dropped the `audioUrl` follower entirely**:
   - Eliminates SSRF bypass vectors: untrusted URLs returned inside a JSON envelope are never followed.
   - Eliminates recursion: no follow-up fetch loop is possible.
   - JSON responses are rejected directly with `TTSInvalidResponseError` (502).

2. **Skip leading-byte sniff for `audio/*` content types (protects PCM, μ-law, A-law)**:
   - Headerless audio formats have no magic numbers and ~1.2% naturally start with `<`, `{`, or `[`.
   - When `Content-Type` starts with `audio/`, the leading-byte sniff is skipped entirely, preventing false-positive rejections for ElevenLabs PCM/μ-law/A-law and OpenAI PCM.

3. **Guaranteed "never billed" invariant**:
   - `tests/audio/tts-invalid-response.test.ts` mocks `@/lib/server/usage-storage` and asserts `recordGenerationUsage` is strictly **not** called on the 502 path, and **is** called on the 200 success path.

4. **Security & Information Leakage Prevention**:
   - Upstream JSON details or plain-text snippets are logged on the server with `log.warn` instead of being echoed in `error.message` to the browser, preventing internal gateway error leakage.
   - Generic, descriptive error messages are returned in the 502 response envelope.

5. **Format Handling & Visibility**:
   - In `generateOpenAITTS`, fallback format defaults to `mp3` (OpenAI does not receive `response_format` and always sends mp3 bytes).
   - `validateTTSAudioResponse` is an internal helper in `lib/audio/tts-providers.ts` (not exported).
   - Provider wiring: `generateOpenAITTS`, `generateLemonadeTTS`, `generateVoxCPMTTS`, `generateAzureTTS`, `generateGLMTTS`, and `generateElevenLabsTTS`.
   - Clean diff against `origin/main`: unrelated `scripts/openmaic-packages.mjs` changes excluded.

## Verification
- Prettier and ESLint pass with 0 errors/warnings on all changed files.
- Vitest suite `tests/audio/tts-invalid-response.test.ts` (10/10 tests pass), including:
  - HTML responses (with text/html, octet-stream, BOM/whitespace).
  - JSON responses (generic message without leaking upstream details).
  - Empty (0 bytes) and text/plain bodies.
  - Headerless PCM starting with `<`, `{`, or `[` accepted when `audio/*`.
  - SSRF-safe rejection of JSON envelopes containing internal metadata URLs.
  - Billing invariant assertion verifying `recordGenerationUsage` is never called on 502.
- Neutrality guard: `tests/providers/provider-neutrality-guard.test.ts` passes.


wyuc (changes requested):
Thanks for this — the core of the fix is right and well tested: sniffing the 200 body before it can be stored or billed is exactly what #1395 asked for, and the wiring into the six raw-bytes providers plus the route-level 502 before `recordGenerationUsage` all check out (verified the ordering in `app/api/generate/tts/route.ts`, and that the two server-side callers in `classroom-media-generation.ts` and `scene-tts.ts` already catch per action, so they benefit too).

The `audioUrl` follower is the part I can't merge as is. Findings below, most severe first; the two blocking ones were reproduced with a stubbed `fetch`.

**1. (blocking) The follow-up `fetch(resolvedUrl)` bypasses the SSRF guard.** `route.ts` runs `validateUrlForSSRF` on a client-supplied base URL, but the URL taken from the JSON envelope is fetched with no validation, `redirect: 'follow'`, and no size cap. A base URL on an attacker-controlled public host (passes the guard) answering `{"audioUrl":"http://169.254.169.254/latest/meta-data/..."}` gets the internal response handed back to the caller through two channels: the `text/plain` snippet that lands in `error.message` (and is returned by `apiError`), and any unrecognised content type which is returned verbatim as base64 "audio". This is a new trust boundary, not the existing one. The repo already solves this exact "provider returns an audio URL" case in `lib/audio/qwen-voice-clone.ts` (`downloadAudio`: host allowlist, protocol/port checks, `redirect: 'error'`, streaming byte cap), and `generate-image.ts` does per-hop `validateUrlForSSRF` with `redirect: 'manual'`. Either reuse one of those, or drop the follower from this PR — none of the six wired providers return that envelope shape, and it isn't needed to close #1395.

**2. (blocking) Unbounded recursion.** `validateAndResolveTTSAudioResponse` calls itself with no depth counter. An envelope whose `audioUrl` yields another envelope loops until the 30 s TTS timeout; with a 1 ms mock latency that is ~730 outbound requests per second per `POST /api/generate/tts`, each level retaining its body. With zero-latency responses the loop is pure microtasks, so the abort timer never runs and the Node process OOMs (observed at 4 GB in ~15 s). A depth cap of 1 fixes it.

**3. The first-byte sniff runs even when `content-type` is `audio/*`, so headerless audio is rejected at random.** Raw PCM, μ-law and A-law have no magic number; ~1.2 % of such responses start with `<`, `{` or `[` and are rejected as HTML or malformed JSON. `elevenlabs-tts` ships `pcm`/`ulaw`/`alaw` (`pcm_44100`, `ulaw_8000`, `alaw_8000`) and `openai-tts` lists `pcm`, so this is reachable today. Skip the sniff when the content type starts with `audio/`.

**4. The "never billed" invariant isn't actually covered.** Moving the `recordGenerationUsage` call above `await generateTTS` in the route leaves all 11 tests green. Please mock `@/lib/server/usage-storage` and assert it is not called on the 502 path.

Minor:
- The `text/plain` snippet and JSON `detail`/`message`/`error` are echoed to the browser; for server-configured providers that can leak an internal gateway's error body. Log it and return a generic message in that case.
- `requestedFormat || 'mp3'`: `generateOpenAITTS` never sends `response_format`, so with a proxy that omits `content-type` a user who picked `flac`/`aac` now gets mp3 bytes labeled `flac`/`aac`, where before the label was correct.
- The `!audioResponse.ok` branch never consumes the body (`await audioResponse.body?.cancel()`).
- `validateAndResolveTTSAudioResponse` is exported but has no external consumer.
- The `scripts/openmaic-packages.mjs` CRLF change is unrelated to #1395; please keep it out of this PR or explain it in the description.

Checks run on the PR head: `pnpm exec tsc --noEmit`, eslint and prettier on the changed files, and vitest over `tests/audio`, `tests/agent-runtime`, `tests/server`, `tests/providers` (148 files, 1688 tests) — all green. Happy to re-review once 1–4 are addressed; the simplest path is probably to remove the follower and keep the validator.

maheshsingh20:
Hi @wyuc 

Thanks for the thorough review and catching these edge cases! All points have been addressed in commit `3bf261f5`:

1. **Dropped the audioUrl follower (Blocking points 1 and 2)**:
   - Completely dropped the audioUrl follow-up fetch logic. Any JSON response returned from a raw-bytes speech provider is directly rejected with `TTSInvalidResponseError` (`502`).
   - This eliminates the SSRF bypass risk on internal addresses (such as `169.254.169.254`) and eliminates recursion/OOM loops entirely.

2. **Headerless audio format protection (Point 3)**:
   - When `Content-Type` starts with `audio/`, the leading-byte sniff (`<`, `{`, `[`) is skipped entirely (only verifying `byteLength > 0`).
   - Headerless formats like raw PCM, μ-law, and A-law from ElevenLabs/OpenAI whose samples naturally start with `<`, `{`, or `[` are no longer falsely rejected. Added a unit test covering this case.

3. **"Never billed" invariant covered (Point 4)**:
   - Mocked `@/lib/server/usage-storage` in `tests/audio/tts-invalid-response.test.ts`.
   - Added explicit assertions that `recordGenerationUsage` is strictly **not** called on the 502 rejection path, and **is** called with the expected parameters on 200 success.

4. **Minor items and cleanup**:
   - **Gateway leak prevention**: Upstream JSON error details and plain-text snippets are logged on the server with `log.warn` rather than echoed in `error.message` to the browser. Safe generic error messages are returned in the response.
   - **Format handling**: OpenAI fallback format in `getAudioResponseFormat` defaults to `mp3` without overriding when `Content-Type` is omitted.
   - **Visibility**: `validateTTSAudioResponse` is internal to `lib/audio/tts-providers.ts` (not exported).
   - **Diff hygiene**: Reverted the unrelated CRLF edit in `scripts/openmaic-packages.mjs`. The PR diff now strictly touches only the 4 relevant files.

All CI workflows on the PR head are green (`Lint, Typecheck & Unit Tests`, `E2E Tests`, `Render Service`, and `Storage Contract`). Ready for re-review!

wyuc (approved):
Round 2 verified: the `audioUrl` follower and the recursion are gone entirely (the only `fetch(` sites left are the provider request calls, none takes a URL from a response body, and the new test pins the 169.254.169.254 case); the billing invariant test goes red when `recordGenerationUsage` is moved above `generateTTS`; all round-1 P2s are fixed, moot, or reverted. 262 tests, tsc, eslint, prettier and `pnpm check` pass on 1f8df7ee. Merging.

Three non-blocking notes, in case you want a small follow-up:

- The `audio/*` early return trusts the header absolutely, so an HTML or JSON body served with `content-type: audio/mpeg` is still stored and billed. A multi-byte signature check on that path (reject only on exact `<!DOCTYPE`, `<html`, `<?xml`, `{"` prefixes after whitespace/BOM) would close it without reintroducing the PCM false positive, since a 5-byte exact match against raw samples is vanishingly unlikely.
- ElevenLabs `format` now follows the response header with the requested format only as fallback; since the output format is pinned in the query string, the request is the authoritative label. Small behaviour change worth a mention.
- `getAudioResponseFormat` has no mapping for `audio/pcm`, `audio/basic`, `audio/L16`, so the new test asserts `format === 'mp3'` for `audio/pcm`; mapping those or falling back to the requested format would avoid locking that in.

PR #1409, opened 2026-09-07
fix(i18n): UI copy that never reaches the translator (4 places)

khunmax2:
Four places where UI copy never reaches the translator, so it shows the same language in all twelve locales. Each commit stands alone.

**`fix(i18n): add the missing stage.proMode label`**

`components/stage/header-controls.tsx` renders `{t('stage.proMode')}` beside the Pro-mode switch, and no locale defines that key — so every language, Simplified Chinese included, shows the literal string `stage.proMode` where a label should be.

`scripts/check-i18n-keys.mjs` cannot catch this, which is why it has survived: the check compares locale files against each other, and a key that exists in none of them is perfectly consistent.

**`fix(i18n): translate the classroom transport controls`**

The playback toolbar labels its controls in English for every language — `Toggle sidebar`, `Playback speed`, `Previous scene`, `Next scene`, `Auto-play`, `Toggle chat`, plus the `Play`/`Pause` and `Mute`/`Unmute` pairs. They sit beside neighbours that already call `t()`, so a Simplified Chinese reader gets a mixed toolbar and a screen reader announces English throughout.

The Play/Pause and Mute/Unmute labels are built from a ternary rather than written as literals, which is why a sweep for `aria-label="…"` walks past them.

**`fix(i18n): translate the media popover's tab labels`**

`Image`, `Video`, `TTS` and `ASR` are written into the `TABS` table as English strings, so they stay English everywhere while the panel below each tab is translated. The table is module scope and `t` lives in the component, so the label becomes a key the tab bar resolves where the translator is in scope.

`TTS` and `ASR` also become words rather than acronyms, matching how the panels below already name the same capabilities.

**`fix(i18n): route the speech recorder's messages through t()`**

`useAudioRecorder` reports failures with nine hardcoded Simplified Chinese strings — no speech detected, microphone unavailable, permission denied, network error, and the rest. The other eleven locales get Chinese for every recording failure.

---

**Deliberately not touched:** `recognition.lang = asrLanguage || 'zh-CN'` in the same hook, and the other Chinese language defaults. Those are decisions about the default *language*, which is yours to make; this PR is only about copy that cannot be translated at all.

Translations for the eleven non-English locales are included. I am most confident in the English, Simplified and Traditional Chinese and Japanese; the rest follow the conventions already in your locale files, and I would rather you correct anything that reads wrong than have me guess twice.

`node scripts/check-i18n-keys.mjs` passes at 12 locale files.


LING-6150 (changes requested):
Thanks , I confirmed that the missing stage.proMode key, English-only toolbar accessibility labels, and Chinese-only recorder errors are real, and the i18n plumbing looks sound.
Two blockers:
1. The required CI check is currently failing because lib/hooks/use-audio-recorder.ts does not pass Prettier.
2. Please preserve the existing TTS/ASR product terminology. Changing the tabs to “Speech”/“Voice” is a terminology change rather than an i18n fix, and it creates ambiguity with existing labels such as the TTS voice selector. Please keep the new labelKey plumbing but use the existing TTS/ASR terminology across locales.
Once these are addressed, I’ll run the user-visible browser checks.

PR #1410, opened 2026-09-07
feat(narration): add Abogen scene-level narration via enableNarration

LIGHTSPEED1699:
## Summary

Adds optional per-scene **Abogen narration** to OpenMAIC classrooms. When `enableNarration: true` is set on `POST /api/generate-classroom`, after the classroom is persisted the server runs `scripts/narration_pipeline.py`, which for each scene:

1. extracts the scene's speech text,
2. synthesizes one M4B via a local **Abogen** (Kokoro TTS) service,
3. waits for the M4B to be **complete** (moov atom present + stable size) so browsers can decode it,
4. copies it into the classroom media dir and stamps `scenes[i].narrationUrl`,
5. re-persists the classroom JSON.

Narration plays once per scene as a voiceover on a dedicated element, separate from per-line speech. It is **best-effort**: a down Abogen or script failure logs a warning and does not fail generation.

## Changes

- `enableNarration` input flag + `generating_narration` progress step
- `scripts/narration_pipeline.py` (env-configurable: `ABOGEN_URL`, `ABOGEN_OUTPUT_DIR`, `ABOGEN_VOICE`)
- `lib/server/classroom-storage.ts` — storage root now env-configurable via `CLASSROOMS_DATA_DIR`; best-effort migration from legacy `<cwd>/data`. Migrated files are copied to a sibling temp name and renamed into place, so a copy that dies mid-write (disk full, crash) can never leave a truncated destination that the mtime guard would then skip forever
- `app/api/classroom-media/[...path]/route.ts` — serve `.m4a`/`.m4b` as `audio/mp4` with `Cache-Control: no-store` (a stable narration URL + `immutable` caching left clients playing stale silent bytes)
- `package.json` build — copy `.next/static` into standalone output (Next.js build does not do this)
- Skill docs in `skills/openmaic/`

## Notes

Requires a **local Abogen** service, e.g. `http://localhost:8808`. For contributors, point `ABOGEN_URL` / `ABOGEN_OUTPUT_DIR` / `CLASSROOMS_DATA_DIR` via `.env.local` (gitignored).

### Known limitation: the boot window

Legacy migration is fired from `instrumentation.register()` as **fire-and-forget** (boot must not block on I/O), so on the *first* boot of a deployment that (a) sets `CLASSROOMS_DATA_DIR` to a new location and (b) has a legacy `<cwd>/data` directory, there is a short window — the first seconds after boot, until the copy finishes — where a classroom that exists only in the legacy directory reads as missing (`readClassroom` → `null` → 404). It is self-healing: the migration completes, writes its marker, and every later read finds the data. The window only exists on the first boot after a storage-path change; the marker makes subsequent boots no-ops.

Accepted as-is rather than making `readClassroom` migrate synchronously on a first miss, because that would put the full copy on the request path and race the marker the background pass is already writing.

### Known behavior: the sound button

The sound/mute button is actionable in classrooms that have neither TTS nor scene narration (it is the existing shared TTS control; there is simply nothing for it to silence).

## Review follow-ups

- **Partial copies can survive forever** — fixed: `copyFileAtomic` (temp + rename, temp removed on failure) with a regression test that fails against the previous in-place `copyFile`.
- **Job download 404 polls to the full timeout** — fixed: `narration_pipeline.py` probes the job page on a 404 and fails immediately when Abogen reports the job gone/failed/cancelled.
- **Narration volume hardcoded to 1** — fixed: scene narration now takes the shared player's volume (and mute), so the volume slider affects it.

### DIFF
--- a/app/api/classroom-media/[classroomId]/[...path]/route.ts
+++ b/app/api/classroom-media/[classroomId]/[...path]/route.ts
@@ -18,6 +18,8 @@ const MIME_TYPES: Record<string, string> = {
   '.wav': 'audio/wav',
   '.ogg': 'audio/ogg',
   '.aac': 'audio/aac',
+  '.m4a': 'audio/mp4',
+  '.m4b': 'audio/mp4',
 };
 
 export async function GET(
@@ -79,7 +81,7 @@ export async function GET(
       headers: {
         'Content-Type': contentType,
         'Content-Length': String(stat.size),
-        'Cache-Control': 'public, max-age=86400, immutable',
+        'Cache-Control': 'no-store',
       },
     });
   } catch (error) {
--- a/app/api/classroom-media/[classroomId]/[...path]/route.ts
+++ b/app/api/classroom-media/[classroomId]/[...path]/route.ts
@@ -63,6 +63,15 @@ export async function GET(
     const ext = path.extname(realPath).toLowerCase();
     const contentType = MIME_TYPES[ext] || 'application/octet-stream';
 
+    // Narration files are rewritten in place under a stable URL, so they must
+    // never be cached (a stale fetch would keep silent bytes). Immutable media
+    // (images, videos, other audio) is written once at generation time and can
+    // keep long-lived cache headers.
+    const isNarration = ext === '.m4a' || ext === '.m4b';
+    const cacheControl = isNarration
+      ? 'no-store'
+      : 'public, max-age=86400, immutable';
+
     // Stream the file to avoid loading large videos into memory
     const stream = createReadStream(realPath);
     const webStream = new ReadableStream({
@@ -81,7 +90,7 @@ export async function GET(
       headers: {
         'Content-Type': contentType,
         'Content-Length': String(stat.size),
-        'Cache-Control': 'no-store',
+        'Cache-Control': cacheControl,
       },
     });
   } catch (error) {
--- a/app/api/generate-classroom/route.ts
+++ b/app/api/generate-classroom/route.ts
@@ -32,6 +32,7 @@ export async function POST(req: NextRequest) {
         ? { enableVideoGeneration: rawBody.enableVideoGeneration }
         : {}),
       ...(rawBody.enableTTS != null ? { enableTTS: rawBody.enableTTS } : {}),
+      ...(rawBody.enableNarration != null ? { enableNarration: rawBody.enableNarration } : {}),
       ...(rawBody.agentMode ? { agentMode: rawBody.agentMode } : {}),
     };
     const { requirement } = body;
--- a/components/canvas/canvas-toolbar.tsx
+++ b/components/canvas/canvas-toolbar.tsx
@@ -195,19 +195,16 @@ export function CanvasToolbar({
             >
               <button
                 onClick={onToggleMute}
-                disabled={!ttsEnabled}
                 className={cn(
                   ctrlBtn,
                   'w-6 h-6',
-                  !ttsEnabled
-                    ? 'text-gray-300 dark:text-gray-600 cursor-not-allowed'
-                    : ttsMuted
-                      ? 'text-red-500 dark:text-red-400'
-                      : 'text-gray-500 dark:text-gray-400',
+                  ttsMuted
+                    ? 'text-red-500 dark:text-red-400'
+                    : 'text-gray-500 dark:text-gray-400',
                 )}
                 aria-label={ttsMuted ? 'Unmute' : 'Mute'}
               >
-                <VolumeIcon muted={!!ttsMuted} volume={ttsVolume} disabled={!ttsEnabled} />
+                <VolumeIcon muted={!!ttsMuted} volume={ttsVolume} disabled={false} />
               </button>
 
               {/* Vertical volume slider (pops up above) */}
--- a/components/edit/PlaybackChromeRoot.tsx
+++ b/components/edit/PlaybackChromeRoot.tsx
@@ -923,6 +923,9 @@ export const PlaybackChromeRoot = forwardRef<PlaybackChromeRootHandle, PlaybackC
     // Sync mute state from settings store to audioPlayer
     useEffect(() => {
       audioPlayerRef.current.setMuted(ttsMuted);
+      // Also silence scene-level narration (separate element) so the
+      // sound/mute button controls it too.
+      engineRef.current?.setNarrationMuted(ttsMuted);
     }, [ttsMuted]);
 
     // Sync volume from settings store to audioPlayer
--- a/components/edit/PlaybackChromeRoot.tsx
+++ b/components/edit/PlaybackChromeRoot.tsx
@@ -925,7 +925,7 @@ export const PlaybackChromeRoot = forwardRef<PlaybackChromeRootHandle, PlaybackC
       audioPlayerRef.current.setMuted(ttsMuted);
       // Also silence scene-level narration (separate element) so the
       // sound/mute button controls it too.
-      engineRef.current?.setNarrationMuted(ttsMuted);
+      engineRef.current?.syncNarrationVolume();
     }, [ttsMuted]);
 
     // Sync volume from settings store to audioPlayer
@@ -934,6 +934,8 @@ export const PlaybackChromeRoot = forwardRef<PlaybackChromeRootHandle, PlaybackC
       if (!ttsMuted) {
         audioPlayerRef.current.setVolume(ttsVolume);
       }
+      // Narration follows the same volume, not a hardcoded 1.
+      engineRef.current?.syncNarrationVolume();
     }, [ttsVolume, ttsMuted]);
 
     // Sync playback speed to audio player (for live-updating current audio)
--- a/components/roundtable/index.tsx
+++ b/components/roundtable/index.tsx
@@ -679,7 +679,7 @@ export function Roundtable({
       ttsEnabled={ttsEnabled}
       ttsMuted={ttsMuted}
       ttsVolume={ttsVolume}
-      onToggleMute={() => ttsEnabled && setTTSMuted(!ttsMuted)}
+      onToggleMute={() => setTTSMuted(!ttsMuted)}
       onVolumeChange={(v) => setTTSVolume(v)}
       autoPlayLecture={autoPlayLecture}
       onToggleAutoPlay={() => setAutoPlayLecture(!autoPlayLecture)}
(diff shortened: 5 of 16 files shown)


CXuPercy (changes requested):
Thanks — the moov-atom/stability wait is the right instinct (mid-write M4Bs were the silent-narration root cause), and wrapping the phase best-effort so a down Abogen never fails generation is the right shape. But this needs another pass — findings, most severe first:

**1. (blocking) The narration subprocess and the storage layer disagree on the env var.** `lib/server/classroom-generation.ts` reads `CLASSROOMS_DIR` (falling back to `<cwd>/data/classrooms`) while `classroom-storage.ts` moved to `CLASSROOMS_DATA_DIR`. With only the latter set, the pipeline can't find the persisted JSON and narration silently no-ops behind the try/catch. Use the exported `CLASSROOMS_DIR` from classroom-storage instead of re-deriving.

**2. (blocking) The legacy migration runs on every read and can overwrite newer data.** `migrateLegacyClassroomData()` is called from `readClassroom`/ensure* with no completion marker, copy-only semantics, and no timestamp comparison: after a storage move, a classroom regenerated into the new location can be reverted by the stale legacy copy on the next read. Even in the default (same-path) configuration it rescans the whole tree on every read — and same-path `copyFile` behavior differs across platforms (no-op on Node 24/macOS; I wouldn't rely on it under the Linux standalone target). Migrate once at startup behind a marker, and never copy over a newer destination.

**3. (blocking) `newest_m4b()` mis-attributes audio under concurrency.** The script identifies each scene's output as "the newest .m4b in the Abogen output dir that differs from the previous one". Any concurrent Abogen job — another classroom, any other user of the local service — swaps in the wrong file, and nothing detects it. Correlate by the job id the API returns instead of mtime scavenging.

**4. Blanket `no-store` for all classroom media.** The route change fixes the stale-narration cache but applies to every image and audio file the classroom serves. Scope it to narration files (or version the narration URL) so immutable media keeps its cache headers.

**5. Pause drops the scene's narration for good.** `pause()` stops and discards `narrationAudio`; `resume()` never restarts it, so any pause mid-scene loses the voiceover until the next scene. At minimum resume from `currentTime`, or state the interaction explicitly in the skill docs.

**6. The skill docs leak a personal deployment and contradict the code.** `skills/openmaic/references/generate-flow.md` references `~/.local/bin/narration_pipeline.py`, `systemctl --user restart openmaic.service`, a host called "ryzen9", and personal backup dirs — while the PR ships the script at `scripts/narration_pipeline.py` (which SKILL.md correctly references). The docs need to describe the repo's layout, not one machine's.

Minor: no tests anywhere in the PR (the storage migration alone deserves some), the `cp -r .next/static` appended to `build` is not Windows-safe and looks orthogonal to narration, and narration overlaps audibly with per-line TTS when both are enabled — worth a sentence in the docs on the intended combination.

Checks run: none on the head — the PR ships no tests; findings are from reading the diff. Happy to re-review once 1–3 are addressed.

LIGHTSPEED1699:
@CXuPercy addressed all six findings — env-var fix, marker-gated migration, job-id correlation, cache scoping, pause/resume, docs + tests. Ready for re-review.

CXuPercy (commented):
Thanks — this rework addresses all six findings and the quality is high: the storage layer now migrates once behind a marker with same-path no-op and an mtime guard, the pipeline correlates by job id through a job-scoped download endpoint with atomic replace (better than the mtime scavenging the review asked for), caching is scoped to narration files, pause/resume keeps `currentTime` with mute wired through `PlaybackChromeRoot`, and the docs now describe the repo's layout. The new migration tests pass on the head (4/4). Three things before approval, plus minors:

**1. (please fix) A partial copy can survive forever.** `copyDirRecursive` writes destinations in place (`lib/server/classroom-storage.ts` copyFile path); if a copy dies mid-write (disk full, crash), the partial file's mtime is newer than the legacy source, the retry skips it via the mtime guard, and the completion marker then freezes it there. You already do temp + `os.replace` on the Python side — do the same here (or unlink the partial destination on failure).

**2. (please note in the description) The boot window.** Migration is fire-and-forget from `instrumentation`, so for the first seconds after boot — only on a deployment that sets `CLASSROOMS_DATA_DIR` while legacy data exists — a classroom that lives only in the legacy directory reads as missing. Acceptable trade, but state it explicitly (or make `readClassroom` run the migration synchronously on a first miss).

**3. (blocking evidence) One real run, please.** The whole pipeline — `/wizard/text` → `/wizard/finish` → `GET /jobs/{id}/download/audio`, the moov wait, the stamping of `narrationUrl`, and the playback UX (voiceover on scene start, pause/resume, mute sync) — has never been shown running against a live Abogen. You're the one with the local service, so please generate one narrated classroom end-to-end and post the pipeline log plus a screenshot of playback with narration active. That closes the contract verification and the UI verification in one pass.

Minor, no rework needed: a 404 on the job download polls until the full timeout instead of failing fast on a known-dead job; narration volume is hardcoded to 1, so the per-line TTS volume slider doesn't affect it (mute does); and the sound button is now actionable in classrooms with neither TTS nor narration — fine to note as known behavior.

CXuPercy:
One more point that should be surfaced before this merges — a product-shape question for the maintainers rather than a code defect: this PR adds the product's first server-side dependency on an external non-Node runtime (`execFile('python3')` + a script that talks to a separately-deployed Abogen). That's a genuine capability boundary expansion (scene-level narration is a third audio pathway, and the `CLASSROOMS_DATA_DIR` + migration machinery touches all classroom storage). The repo already has a pattern for optional local services — Ollama, Lemonade, FunASR are first-class providers with settings UI and health checks. As written, narration is none of that: no settings surface, no health check, no provider abstraction, just an env-and-docs contract. Before merging, @wyuc should weigh in on the intended shape: provider-ize it like the existing local services, keep it out of the server pipeline (skill/docs-level tooling), or accept the current form consciously.

LIGHTSPEED1699:
Thanks — all three addressed, plus the minors. Pushed as `84efa508`.

### 1. Partial copy can survive forever — fixed

Migration copies now go through `copyFileAtomic` in `lib/server/classroom-storage.ts`: copy to a sibling `<dest>.<pid>.<ts>.tmp`, then `fs.rename` into place, and remove the temp if the copy throws. `dest` is therefore either absent or complete, so a copy that dies mid-write can no longer leave a truncated file whose fresh mtime beats the legacy source (which the guard would skip and the marker would freeze).

Regression test in `tests/server/classroom-storage-migration.test.ts` (now 5/5): stubs `fs.copyFile` to write truncated bytes and then throw `ENOSPC`, asserts there is no partial destination, no `.tmp` litter, and **no marker** (so the next boot retries rather than freezing it), then that the retry completes the copy for real. It fails against the previous commit (`Received: "{\"legacy\":"` at the destination) and passes now.

### 2. Boot window — documented

Added to the description as **"Known limitation: the boot window"**: on the first boot of a deployment that both sets `CLASSROOMS_DATA_DIR` and has legacy `<cwd>/data`, reads of a classroom that lives only in the legacy directory return `null` → 404 for the first seconds until the copy finishes, then self-heal (the marker makes later boots no-ops). I took the doc option rather than synchronous migrate-on-first-miss, because the latter puts the whole copy on the request path and races the marker the background pass is already writing — say the word if you'd rather have the synchronous version.

### 3. One real run — done, evidence below

Production standalone build (`npm run build` → `node .next/standalone/server.js`), real `POST /api/generate-classroom` with `enableNarration: true`, against the local Abogen on `:8808`.

```
[07:54:52] POST /api/generate-classroom
submit response: {"success": true, "jobId": "jZob1kfr67", "status": "queued", ...}
[07:59:16] status=running   step=generating_narration progress=99 scenes=3/3 :: Generating Abogen narration audio
[08:00:31] status=succeeded step=completed           progress=100 scenes=3/3 :: Classroom generation completed
  -> classroomId R1tJBM1l0T, 3 scenes
```

Server side, the narration phase and the stdout it execFile'd out of `narration_pipeline.py`:

```
[12:00:26] [INFO] [Classroom] Narration pipeline stdout: wired scene 1: a8606036d57843e58b958511d1aabbbe -> /media/sdddata1/openmaic-data/classrooms/R1tJBM1l0T/media/scene-01.m4b
wired scene 2: 25040a1d6998464d848a10273938173a -> .../media/scene-02.m4b
wired scene 3: 9cfadf2ea8d24ab5844e1e67387f13ca -> .../media/scene-03.m4b
narration wired for 3/3 scenes
```

Stamping + serving (`/wizard/text` → `/wizard/finish` → `GET /jobs/{id}/download/audio`, all three correlated by the job id returned from `finish`):

```
scene 1: type=slide,       narrationUrl=/api/classroom-media/R1tJBM1l0T/media/scene-01.m4b
scene 2: type=interactive, narrationUrl=/api/classroom-media/R1tJBM1l0T/media/scene-02.m4b
scene 3: type=quiz,        narrationUrl=/api/classroom-media/R1tJBM1l0T/media/scene-03.m4b

scene-01.m4b: moov atom at byte 32 (inside first 4096) | duration=120.300000
scene-02.m4b: moov atom at byte 32 (inside first 4096) | duration=42.575000
scene-03.m4b: moov atom at byte 32 (inside first 4096) | duration=28.050000

GET /api/classroom-media/R1tJBM1l0T/media/scene-01.m4b -> 200, content-type: audio/mp4,
cache-control: no-store, content-length: 1679651

job a8606036...: Completed | download/audio -> HTTP 200, 1679651 B
job 25040a1d...: Completed | download/audio -> HTTP 200,  590814 B
job 9cfadf2e...: Completed | download/audio -> HTTP 200,  377240 B
```

Playback, in a real browser against that classroom (the narration element is detached — `new Audio(url)` — so it is wrapped to observe `currentTime`/`volume`/`paused` directly):

```
[playing] {"src":".../scene-01.m4b","currentTime_t1":0.69,"currentTime_t2":3.19,"advanced":true,"duration":120.3,"readyState":4}
[muted]   {"volume":0,"paused":false,"stillAdvancing":true}
[unmuted] {"volume":1}
[paused]  {"before":5.675,"after":5.726,"paused":true,"heldAtCurrentTime":true}
[resumed] {"paused":false,"resumedFrom":7.760,"didNotRestart":true}
```

playback with narration active

Full artifacts (job poll log, server log, browser checks, paused/resumed screenshots) are on a scratch branch: [`evidence/abogen-narration-e2e`](https://github.com/LIGHTSPEED1699/OpenMAIC/tree/evidence/abogen-narration-e2e) — not part of the PR diff.

Two caveats, so the evidence is not read as more than it is: generation ran on a different LLM provider than the repo default because the ollama-cli cloud quota on this box is exhausted (irrelevant to the narration path, but it is not the stock config), and playback was verified in headless Chromium rather than by ear.

### Minors

- **404 polls to full timeout** — fixed in `narration_pipeline.py`: on a download 404 it probes the job page (`/jobs/<id>`), which distinguishes "still running" from "gone/failed/cancelled", and raises immediately in the terminal case instead of polling out the 900 s.
- **Narration volume hardcoded to 1** — fixed: `PlaybackEngine.narrationVolume()` reads the shared player's volume (0 when muted) both when the narration element is created and on live volume/mute changes, via `syncNarrationVolume()` called from the existing sync effects. Mute handling is unchanged.
- **Sound button actionable with no TTS/narration** — noted in the description under "Known behavior", no code change.

### @wyuc — the shape question is yours

This does add the product's first server-side dependency on a non-Node runtime: `execFile('python3')` plus a script that talks to a separately deployed Abogen. Scene narration is a third audio path and the `CLASSROOMS_DATA_DIR` + migration machinery touches classroom storage as a whole, and unlike Ollama/Lemonade/FunASR there is no settings surface, no health check, and no provider abstraction — just env and docs. Three shapes, as I see it: provider-ize it alongside the existing local services, keep it out of the server pipeline entirely (skill/docs-level tooling that post-processes a classroom), or accept the current env-and-docs contract consciously. Happy to implement whichever you pick; I would not merge this on the default answer.

PR #1411, opened 2026-09-08
fix(importer): convert Equation.3 OLE formulas via MTEF v3, surface degrade telemetry

CXuPercy:
## Problem

Legacy courseware stores formulas as Equation 3.0 / MathType OLE objects. Their only renderable form inside the .pptx is a WMF preview picture, which the importer cannot rasterize. Those formulas degraded to a hardcoded 1×1 **`TRANSPARENT_PNG_DATA_URL`** — actually a 50%-alpha red pixel (RGB 255,0,0 α=127) despite the name — rendered as a pink block once stretched over the formula frame, with no way for callers to notice the content loss.

## Changes

1. **Equation.3 / MathType OLE → LaTeX** (new `utils/mtef.ts`): detect the progId, resolve the embedding, parse the OLE compound file (`cfb`), convert its `Equation Native` MTEF v3 stream — fractions, radicals, scripts, fences, big operators with per-family limit variations, embellishments, Symbol-font local character encodings (table verified against URW StandardSymbolsPS AFM + Adobe AGL). Slot orders follow the rtf2latex2e reference implementation and real MathType streams (`[main, lower, upper]`; tmLSCRIPT base = following sibling; tmLIM 0=upper/1=lower), not the archived spec prose. Hardening: 200-deep nesting cap and 64 KiB output cap, both failing loud as `MtefParseError`. Any failure falls back to the existing picture path.
2. **Placeholder fixed**: truly transparent pixel; the legacy red one stays recognized (`isPlaceholderDataUrl` exported).
3. **Degrade telemetry**: optional `ImportPptxOptions.onWarning` receives machine-coded warnings (`media-unconvertible` / `formula-fallback-image` / `formula-degraded` / `element-dropped`) at every placeholder consumption point (image, background, shape/text pattern fills, math fallback, video poster). A throwing sink is isolated so telemetry can never fail an import. A formula whose fallback picture is also a placeholder keeps its plain text as a text element.

## Testing

- **37 new tests** (suite: 85). Highlights: three REAL `Equation Native` stream fixtures round-tripped from a legacy deck (pinning the font-local Symbol encoding and the big-operator slot order); spec-conformance synthetic streams incl. one-sided fences asserted against KaTeX `throwOnError`; script-nesting perf guard (linear, bounded); truncation/depth-cap/`MtefParseError` contract; warning emission and sink isolation.
- Fuzz: 8k random payloads behind a valid v3 header — all rejected as `MtefParseError`, zero other exception types, zero bogus LaTeX.
- End-to-end on a real Equation 3.0 deck: 4 formulas → 4 KaTeX-renderable latex elements, 0 placeholders, 0 warnings; real raster images unaffected.

Review round 1 (thanks @wyuc) findings all addressed in 97909e64 — LSCRIPT base duplication, one-sided fence null delimiters, the Symbol table corrected against the AFM, tmLIM role inversion, depth/output caps.

Version: **0.1.4 → 0.2.0** (new public option/type/export + `Math.degraded` field).

Reviewers please focus on `utils/mtef.ts` (binary grammar) and the warning plumbing in `import-pipeline/`.

### DIFF
--- a/packages/@openmaic/importer/package.json
+++ b/packages/@openmaic/importer/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openmaic/importer",
-  "version": "0.1.5",
+  "version": "0.2.0",
   "description": "A javascript tool for parsing .pptx file",
   "type": "module",
   "main": "./dist/index.cjs",
--- a/packages/@openmaic/importer/package.json
+++ b/packages/@openmaic/importer/package.json
@@ -48,6 +48,7 @@
   "dependencies": {
     "@openmaic/dsl": "workspace:^",
     "@xmldom/xmldom": "^0.9.9",
+    "cfb": "1.2.2",
     "jpegxr": "^0.3.0",
     "jszip": "^3.10.1",
     "katex": "^0.16.33",
--- a/packages/@openmaic/importer/src/adapter/types.ts
+++ b/packages/@openmaic/importer/src/adapter/types.ts
@@ -298,6 +298,9 @@ export interface Math {
   picBase64: string;
   order: number;
   text?: string;
+  /** MTEF conversion succeeded but approximated a construct (PILE/MATRIX …);
+   *  the LaTeX renders but is semantically flattened — surfaces a warning. */
+  degraded?: boolean;
 }
 
 export type BaseElement = Shape | Text | Image | Table | Chart | Video | Audio | Diagram | Math;
--- a/packages/@openmaic/importer/src/import-pipeline/index.ts
+++ b/packages/@openmaic/importer/src/import-pipeline/index.ts
@@ -20,7 +20,7 @@ import type { Output } from '../adapter/types';
 import { parseZip } from '../parser/ZipParser';
 import { buildPresentation } from '../model/Presentation';
 import { toPptxtojsonFormat } from '../adapter/toPptxtojson';
-import type { ImportContext } from './types';
+import type { ImportContext, ImportWarning } from './types';
 import { transformParsedToSlides } from './transformParsedToSlides';
 import { createMockImportContext } from './mockContext';
 
@@ -42,6 +42,13 @@ export interface ImportPptxOptions {
    * keeps an in-memory `blob:` URL (valid only for the current tab).
    */
   upload?: OssUpload;
+  /**
+   * Degrade-not-fail telemetry sink. The import never fails on partial
+   * content loss by design; this is the only channel through which callers
+   * can learn it happened (formulas degraded to placeholders, unconvertible
+   * WMF/EMF media, elements dropped by DSL normalization).
+   */
+  onWarning?: (warning: ImportWarning) => void;
 }
 
 /**
@@ -57,7 +64,7 @@ export async function parsedToSlides(
   json: Output,
   options: ImportPptxOptions = {},
 ): Promise<Slide[]> {
-  const baseCtx = createMockImportContext(buildContextOverrides(options.upload));
+  const baseCtx = createMockImportContext(buildContextOverrides(options.upload, options.onWarning));
   // Drive the transform-time width clamp from the deck's own pixel width
   // (pt → px via ratio) so 16:9 widescreen decks (960pt → 1280px) don't
   // get text elements truncated to the legacy 4:3 default of 960.
@@ -79,7 +86,7 @@ export async function parsedToSlides(
   // viewportRatio / theme are filled at construction); the contract's
   // `normalize` pass at this output boundary is the safety net for anything the
   // transform missed on an exotic deck, mirroring the generator's wiring.
-  return normalizeImportedSlides(slides);
+  return normalizeImportedSlides(slides, options.onWarning);
 }
 
 /**
@@ -109,8 +116,34 @@ const normalizeImportedSlide = normalizeSlideWith({
   },
 });
 
-export function normalizeImportedSlides(slides: Slide[]): Slide[] {
-  return slides.map(normalizeImportedSlide);
+export function normalizeImportedSlides(
+  slides: Slide[],
+  onWarning?: (warning: ImportWarning) => void,
+): Slide[] {
+  if (!onWarning) return slides.map(normalizeImportedSlide);
+  return slides.map((slide, slideIndex) =>
+    normalizeSlideWith({
+      onInvalid: 'drop',
+      onDropped: (element, error) => {
+        console.warn(
+          `[@openmaic/importer] dropping element that fails DSL normalization: ${
+            error instanceof Error ? error.message : String(error)
+          }`,
+        );
+        try {
+          onWarning({
+            code: 'element-dropped',
+            slideIndex,
+            message: `A ${String((element as { type?: string }).type ?? 'unknown')} element failed DSL normalization and was dropped: ${
+              error instanceof Error ? error.message : String(error)
+            }`,
+          });
+        } catch (sinkErr) {
+          console.error('[@openmaic/importer] onWarning sink threw (ignored):', sinkErr);
+        }
+      },
+    })(slide),
+  );
 }
 
 /**
@@ -131,15 +164,20 @@ export async function importPptx(
   return parsedToSlides(json, options);
 }
 
-function buildContextOverrides(upload: OssUpload | undefined): Partial<ImportContext> {
-  if (!upload) return {};
-  return {
-    uploadBase64Image: async (base64, filename, dir) => {
+function buildContextOverrides(
+  upload: OssUpload | undefined,
+  onWarning?: (warning: ImportWarning) => void,
+): Partial<ImportContext> {
+  const overrides: Partial<ImportContext> = {};
+  if (upload) {
+    overrides.uploadBase64Image = async (base64, filename, dir) => {
       const blob = await dataUrlToBlob(base64);
       return upload(blob, filename, dir);
-    },
-    uploadBlobMedia: (blob, filename, dir) => upload(blob, filename, dir),
-  };
+    };
+    overrides.uploadBlobMedia = (blob, filename, dir) => upload(blob, filename, dir);
+  }
+  if (onWarning) overrides.onWarning = onWarning;
+  return overrides;
 }
 
 async function toArrayBuffer(input: File | Blob | ArrayBuffer): Promise<ArrayBuffer> {
@@ -152,7 +190,7 @@ async function dataUrlToBlob(dataUrl: string): Promise<Blob> {
   return response.blob();
 }
 
-export type { ImportContext, TransformResult } from './types';
+export type { ImportContext, ImportWarning, TransformResult } from './types';
 export { transformParsedToSlides } from './transformParsedToSlides';
 export { createMockImportContext } from './mockContext';
 export type { Output } from '../adapter/types';
--- a/packages/@openmaic/importer/src/import-pipeline/transformParsedToSlides.ts
+++ b/packages/@openmaic/importer/src/import-pipeline/transformParsedToSlides.ts
@@ -914,6 +914,7 @@ export async function transformParsedToSlides(
             autoplay: false,
             poster: el.src || '',
           };
+          warnUnconvertibleMedia(ctx, slideIndex, 'A video poster', el.src);
           slide.elements.push(videoElement);
 
           // 上传到 OSS
--- a/packages/@openmaic/importer/src/import-pipeline/transformParsedToSlides.ts
+++ b/packages/@openmaic/importer/src/import-pipeline/transformParsedToSlides.ts
@@ -29,6 +29,7 @@ import type {
 import { SHAPE_PATH_FORMULAS } from '../openmaic/configs/shapes';
 import { getSvgPathRange } from '../openmaic/utils/svgPathParser';
 import { parseVideoCodec, isVideoCodecSupported } from '../openmaic/utils/videoCodec';
+import { isPlaceholderDataUrl } from '../utils/mediaWebConvert';
 import type { ImportContext, TransformResult } from './types';
 
 type ParsedPptxJson = Awaited<ReturnType<typeof parsePptxDefault>>;
@@ -364,6 +365,46 @@ const shouldPreserveRotatedTextFrame = (
   );
 };
 
+/**
+ * Degrade-not-fail telemetry: report media that arrived as the blank
+ * placeholder (WMF / vector-only EMF cannot be converted in this environment).
+ * No-op unless the caller wired `ctx.onWarning`.
+ */
+/** Telemetry must never fail the import: a throwing sink is logged and swallowed. */
+function emitWarning(
+  ctx: ImportContext,
+  warning: { code: string; slideIndex: number; message: string },
+): void {
+  if (!ctx.onWarning) return;
+  try {
+    ctx.onWarning(warning);
+  } catch (err) {
+    console.error('[@openmaic/importer] onWarning sink threw (ignored):', err);
+  }
+}
+
+function warnUnconvertibleMedia(
+  ctx: ImportContext,
+  slideIndex: number,
+  what: string,
+  src: string | undefined,
+): void {
+  if (!isPlaceholderDataUrl(src)) return;
+  emitWarning(ctx, {
+    code: 'media-unconvertible',
+    slideIndex,
+    message: `${what} uses a media format that cannot be converted (WMF / vector EMF); it stays a blank placeholder`,
+  });
+}
+
+function escapeHtmlText(text: string): string {
+  return text
+    .replace(/&/g, '&amp;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;');
+}
+
 export async function transformParsedToSlides(
   json: ParsedPptxJson,
   ctx: ImportContext,
@@ -392,7 +433,7 @@ export async function transformParsedToSlides(
   // base64 图片上传改为并发
   const limitUpload = createConcurrencyLimiter(6);
   const uploadTasks: Promise<unknown>[] = [];
-  for (const item of json.slides) {
+  for (const [slideIndex, item] of json.slides.entries()) {
     const { type, value } = item.fill;
     let background: SlideBackground;
     if (type === 'image') {
@@ -404,6 +445,7 @@ export async function transformParsedToSlides(
           size: 'cover',
         },
       };
+      warnUnconvertibleMedia(ctx, slideIndex, 'Slide background image', value.picBase64);
       if (value.picBase64 && value.picBase64.startsWith('data:')) {
         const bg = background.image!;
         uploadTasks.push(
@@ -536,6 +578,7 @@ export async function transformParsedToSlides(
             let pattern: string | undefined;
             if (el.fill.type === 'image') {
               pattern = el.fill.value.picBase64;
+              warnUnconvertibleMedia(ctx, slideIndex, 'A text frame image fill', pattern);
             }
             const shapeEl: PPTShapeElement = {
               type: 'shape',
@@ -618,6 +661,7 @@ export async function transformParsedToSlides(
           }
         } else if (el.type === 'image') {
           const imageSrc = el.src;
+          warnUnconvertibleMedia(ctx, slideIndex, 'An image element', imageSrc);
 
           const element: PPTImageElement = {
             type: 'image',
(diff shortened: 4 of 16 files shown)


wyuc (changes requested):
The approach is right and the safety net holds: every failure path from OLE detection through MTEF decoding lands in the bare `catch` in `mathSerializer.ts` and degrades to the picture path (30,000 random payloads behind a valid v3 header: 30,000 `MtefParseError`, zero other exception types, zero bogus LaTeX); `cfb` rejects corrupted, truncated and cyclic compound files in under a millisecond; both placeholder PNGs decode to exactly what the PR says; the four warning codes fire at all five listed consumption points and a throwing sink cannot fail the import. 79 importer tests, root tsc, eslint, prettier and the version-bump check pass on 4a4c7944. What blocks is the renderer, not the parser.

**P0-1. `src/utils/mtef.ts` tmLSCRIPT branch: a script slot is re-emitted as the base group, corrupting every left-script formula.**
tmLSCRIPT has two slots (sub, sup); the base is the following sibling CHAR. The template string ends with `{${parts[2]?.latex ?? parts[1]?.latex ?? parts[0]?.latex ?? ''}}`, and `parts[2]` is always undefined, so a script slot is emitted twice. Isotope notation (var 2, sub `6`, sup `12`, then CHAR `C`) renders as `{}_{6}^{12}{12}C`; var 0 renders `{}^{12}{12}C`. KaTeX accepts both, so nothing falls back, nothing warns, the slide just shows a wrong formula. The PR's own test at `test/mtef.test.ts:494` pins the bug (`'{}_{1}{}S'`: stray empty group, base detached). Also, `renderList` routes a tmLSCRIPT that follows an atom into `renderScript`, so `[CHAR '=', LSCRIPT, CHAR 'C']` becomes `=_{6}^{12}C`.

**P0-2. Same line: output doubles per nesting level, and a 221-byte stream OOM-kills the import.**
Nesting tmLSCRIPT inside its own sup slot: 167 B → 131,065 chars, 203 B → 2,097,145, 221 B → 8,388,601, 257 B → 134,217,721, 275 B → `RangeError: Invalid string length`. The LaTeX is then rendered server-side by `katex.renderToString(..., { throwOnError: true })` in `transformParsedToSlides.ts`, which runs inside the `import_pptx` worker; measured amplification is about 59×, and the 221-byte case dies with `FATAL ERROR: Reached heap limit`, exit 134, under a 4 GB heap in about 10 s (well inside the 90 s parse timeout). `parsePptxIsolated` starts the worker without `resourceLimits`, so the parent survives but the pod takes a multi-GB heap spike; in the browser `use-import-pptx.ts` runs `importPptx` on the main thread with no worker at all. `MAX_RECORDS = 20_000` is a record-count cap and the 221-byte stream uses about 140 records; the "perf guard" test nests tmPAREN, which is linear, so it never exercises this. Fixing P0-1 removes the doubling; please also add an output-length cap (throw `MtefParseError` past, say, 64 KiB of LaTeX) and a nesting-depth cap, and turn the perf-guard test into one that nests a script template.

**P0-3. Fence branch: one-sided fences always produce KaTeX-invalid LaTeX with `degraded` still false.**
`open = variation === 2 ? '' : fence[0]` / `close = variation === 1 ? '' : fence[1]` leaves an unmatched `\left` or `\right`. `tmBRACE` var 1 (the standard piecewise-function brace in legacy Equation 3.0 decks) renders `\left \{ a` → `KaTeX parse error: Expected '\right', got 'EOF'`; same for tmPAREN/tmBRACK var 1 and 2. No `formula-degraded` fires, and the formula then falls into the ladder in `transformParsedToSlides.ts` and degrades to flat text or a blank placeholder, which is the outcome this PR set out to fix. Fix is `\left\{ … \right.` and `\left. … \right\}`. `test/mtef.test.ts:434` and `:451` pin the broken strings and never run them through KaTeX, unlike the neighbouring tmDIRAC/tmOARC tests.

**P1-1. `SYMBOL_FONT_LATEX` contradicts the encoding its own header comment names, and has no entries above 0x9F.**
The comment says Adobe Symbol (`0x61 = α`, which is right, and the Greek half matches exactly). Under Adobe Symbol `0x3c`/`0x3e` are `<`/`>` (table says `\le`/`\ge`), `0x5b`/`0x5d` are `[`/`]` (table says arrows), `0x22`/`0x24`/`0x5c` are `∀`/`∃`/`∴`, while the operator range where Adobe Symbol actually keeps `≤ ≥ × ÷ ∞ → ≠ √` (0xA3, 0xB3, 0xB4, 0xB8, 0xA5, 0xAE, 0xB9, 0xD6) is unmapped and falls through to the raw Latin-1 byte: `a ≤ b` renders as `a £ b`, `a < b` as `a \le b`. KaTeX accepts all of it, so no fallback and no telemetry. There is no real `Equation Native` fixture in the repo; all 25 `mtef.test.ts` cases are hand-built from the implementation's own model, and that model is internally inconsistent (`mathSerializer.oleEquation.test.ts:31` encodes `·` as fnSYMBOL + Unicode 0x22C5 but `=` as fnSYMBOL + font-local 0x3D). Please either correct the table to the encoding it documents or add one real legacy deck round-tripped as a fixture that proves the current mapping; either settles both this and the slot-order claims.

**P1-2.** `tmLIM`: `sup = variation === 0 ? '' : parts[2]?.latex` has no fallback to index 1 when the writer omits the lower slot, unlike the big-operator branch right below it which handles exactly that. Upper-only `\lim` silently loses its limit. `tmLIM` has no test. PLAUSIBLE on what real writers emit, but the asymmetry is a defect either way.

**P2.** (a) `transformParsedToSlides.ts` video `poster: el.src || ''` is the one placeholder consumption point not routed through `warnUnconvertibleMedia`. (b) The module JSDoc promises `MtefParseError`, but depth ≥ ~1000 fence nesting throws `RangeError: Maximum call stack size exceeded` and depth ≥ 26 script nesting throws `RangeError: Invalid string length`; the bare catch still degrades correctly, so this is a contract-text and depth-cap issue, not a safety one. (c) `test/mtef.test.ts:215` is titled `[main, upper, lower]` while its bytes and the implementation are `[main, lower, upper]`. (d) `pnpm-lock.yaml` carries about 20 lines of unrelated `eslint-plugin-import` / `eslint-import-resolver-typescript` peer-ID churn from a different pnpm version. (e) PR says 26 new tests; it is 43.

On the rest: `cfb@1.2.2` is exact-pinned in `dependencies` with two SheetJS transitives, last published 2022, bundled into dist like the existing deps, fine. 0.1.4 → 0.1.5 passes the bump script; it adds a public option, a public type, a new export and a new `Math.degraded` field, so 0.2.0 would be the honest number, but that is a judgement call, not a gate.

CXuPercy:
@wyuc All findings addressed in 97909e64:

- **P0-1** tmLSCRIPT no longer re-emits a script slot as the base group (isotopes now render `{}_{6}^{12}C`) and no longer steals the previous atom via the trailing-script lookahead; tvLSUPER single-slot writers fall back to parts[0].
- **P0-2** The fix above removes the doubling (16 levels: 524,281 → 81 chars); additionally added a MAX_DEPTH=200 nesting cap and a 64 KiB output cap, both throwing `MtefParseError`. Perf test now nests script templates (120 levels + length assertion) plus a 500-level throw assertion.
- **P0-3** One-sided fences render the missing side as a null delimiter (`\left.` / `\right.`); tmBRACE/tmPAREN/tmBRACK × var1/var2 all pass `throwOnError` assertions (tests no longer pin invalid strings).
- **P1-1** Symbol table rewritten against URW StandardSymbolsPS AFM + Adobe AGL: 0x3C/0x3E/0x5B/0x5D restored to literal `< > [ ]`; the actual operator block (0xA3 ≤, 0xB3 ≥, 0xB4 ×, 0xB8 ÷, 0xA5 ∞, 0xAE →, 0xB9 ≠, 0xD6 √, plus ± ← ≡ ≈) now mapped; added Chi/vartheta/varsigma, the phi/varphi split, and 0x5E = `\perp`. **Added 3 real Equation Native stream fixtures** (base64-embedded, from an actual legacy deck) pinning the font-local encoding and big-op slot order.
- **P1-2** tmLIM slot fallback added (a later review round also caught the variation roles inverted: 0=tvULIM/1=tvLLIM; corrected per spec + eqn.c).
- **P2** Video poster joined the warning points; depth/output caps throw `MtefParseError` instead of RangeError; test title slot-order corrected; lockfile re-anchored on main with pnpm@10.28.0 (only the cfb additions, no churn); PR description test count corrected (85 total, 37 new).
- Version bumped to **0.2.0** per your suggestion.

Re-verified: 85/85 tests, 8k random fuzz all `MtefParseError` with zero other exception types, real-deck end-to-end unchanged.

Also thank you for the methodological demonstration — validating "is the output correct as math" (isotope constructions, byte-sized nesting bombs, checking against the actual font encoding table) exposed the blind spot of spec-conformance-only review. I've added that approach to our review packets.

wyuc (changes requested):
Round 2 on 97909e64. The three P0s are closed: the LSCRIPT base re-emission is gone in all variations with and without a preceding atom (nested scripts now grow linearly, depth 90 → 451 chars), the output cap is enforced on the final LaTeX and also stops width-driven growth (15,000 sibling macro CHARs throw `MtefParseError`), depth 1000 throws `MtefParseError` instead of `RangeError`, and all 14 fence selectors × variations 0–3 pass KaTeX `throwOnError`. The three fixtures decode as genuine `EQNOLEFILEHDR` + MTEF v3 streams (cbObject matches payload, version 3, platform 1) and render sensible math; every mapped Symbol code I checked (100+, including the phi/varphi split, Chi, vartheta, varsigma, omicron) matches the Adobe Symbol encoding; the lockfile diff is exactly the three cfb packages; the bump script accepts 0.2.0; 85/85 tests, tsc, prettier, eslint and `pnpm check` pass; a 20,000-payload fuzz produced only `MtefParseError`. Three things in the rewritten renderer block, the rest is small.

**P1. tmLIM hardcodes `\lim`, appends the main slot after the limits, and emits no terminating space.** `src/utils/mtef.ts` tmLIM branch. The reference (rtf2latex2e `39.1 = limit: lower, #1 #2`) emits the main slot followed by the limits and no function name, because Equation Editor users type the function into the main slot. Probes: var1 main=`lim` lower=`n` → `\lim_{n}lim` (renders "lim" twice); var1 main=`max` lower=`i` → `\lim_{i}max` (wrong function); var2 → `\lim_{n}^{2}lim`. All KaTeX-valid, `degraded=false`, so the wrong formula ships silently. Worse, with an empty or absent limit slot and a main slot starting with a letter, `\lim` glues to it: `[0x03,39,1,0, LINE 'x' END, END]` → `\limx` → `KaTeX parse error: Undefined control sequence`, and since `transformParsedToSlides.ts` renders with `throwOnError: true` the whole equation drops to the blank placeholder. Reachable on variations 0, 1 and 2. Fix: emit `${main}${sub}${sup}` with no injected operator name (or `\operatorname{}` only when the main slot is empty), and add tests for main=`lim`/`max` and for the empty-limit case through KaTeX.

**P1. Unmapped Adobe Symbol high-half codes still emit raw Latin-1 that KaTeX accepts, so wrong-but-plausible math ships.** `SYMBOL_FONT_LATEX` covers 12 of the ~90 defined positions ≥ 0xA0; everything else falls through to `String.fromCharCode`. Probed: 0xF7 ∫ → `÷`, 0xD7 ⋅ → `×`, 0xE5 ∑ → `å`, 0xD5 ∏ → `Õ`, 0xCE ∈ → `Î`, 0xC8/0xC7 ∪/∩ → `È`/`Ç`, 0xB6/0xD1 ∂/∇ → `¶`/`Ñ`, 0xDE/0xDB ⇒/⇔ → `Þ`/`Û`, 0xA2 ′ → `¢`, 0xBC … → `¼`, 0xE1/0xF6 ⟨/⟩ → `á`/`ö`. `degraded` is set but only yields the advisory `formula-degraded` warning; the latex element is still pushed and KaTeX's default `strict: 'warn'` renders the wrong glyph. ∫ → ÷ and ⋅ → × are the dangerous ones since both look like legitimate operators. Either map the full 0xA0–0xFF block (the AFM you already used has it) or refuse to emit a latex element when an unmapped font-local code is hit so the picture fallback takes over. Same function: font-local 0x60 (radicalex) → literal backtick with `degraded=false`; 0x80–0x9F → raw C1 control characters in the LaTeX.

**P2. tmDIRAC var1/var2 render the wrong bra/ket.** Reference `45.1 = \left\langle #1\right|`, `45.2 = \left| #1\right\rangle`. PR: var1 `[a]` → `\langle a\rangle`; var2 `[a]` → `\langle a \mid \rangle` (content on the left, empty right), both `degraded=false`.

**P2. Two of the three caps have no test.** Setting `MAX_LATEX_LENGTH` or `MAX_RECORDS` to `Number.MAX_SAFE_INTEGER` leaves `test/mtef.test.ts` green (31/31); only `MAX_DEPTH` is pinned. A test asserting the throw at ~15,000 sibling CHARs would cover the width case the output cap exists for.

**P2. Fixture claim and test title.** The fixtures describe-block says the streams pin "the big-op slot order", but none of the three contains a selector in 21–43 (only tmROOT, tmPAREN, tmFRACT, tmSCRIPT). The big-op `[main, lower, upper]` reading is nevertheless correct against rtf2latex2e's template table (`#2` → sub, `#3` → sup), so drop the claim rather than the code. `test/mtef.test.ts:215` is still titled `[main, upper, lower]` while its body and the code use `[main, lower, upper]`; only the inline comment changed.

**P3. Embellishment lists bypass the record budget.** The `for (;;)` embellishment loop in `TAG_CHAR` never touches `budget.count`, and the 64 KiB check runs after the full string is built and whitespace-flattened: 1 MB of embellishment bytes → 251 MB RSS, 10 MB → 1.06 GB before the `MtefParseError`. Always falls back, so availability only; counting embellishments against the budget is one line.

Also noted, non-blocking: the PR description says the perf test nests 120 levels and throws at 500, but the code uses `nestedScripts(90)` + `nestedFences(80)` with the 500-level throw in a separate test; a `CHAR + tmSCRIPT + tmSCRIPT` sequence yields `x^{b}^{a}` (KaTeX "Double superscript"), which I could not tie to a real writer, so PLAUSIBLE only.

wyuc (changes requested):
Round 3 on a8516783. Both P1s are closed: tmLIM now emits the main slot as the operator (36-case matrix over var0/1/2 × main ∈ {lim, max, x, empty} × limits present/empty/absent, all KaTeX-valid; empty main gives `\lim ` with the space), and the Symbol high half was machine-compared against the Adobe Symbol encoding vector for all 256 codes: every position named in round 2 is now correct, undefined positions and 0x80–0x9F throw `MtefParseError` into the picture fallback, the three caps all turn a test red when mutated, DIRAC bra/ket are right, and embellishments are budgeted (10 MB of embellishment bytes now peaks at 88 MiB instead of 1 GB). Fuzz 20,000 → only `MtefParseError`; full KaTeX sweep clean except the one regression below. 91/91 tests, tsc, eslint, prettier, `pnpm check`, bump script and the lockfile diff all fine.

**P1 (regression introduced in this round). `0xd6: '\\sqrt '` must be `'\\surd '`.** 97909e64 had `\surd`; a8516783 changed it to `\sqrt`, which takes a mandatory argument. Adobe Symbol 0xD6 is the standalone `radical` glyph. A Symbol CHAR 0xD6 as the last token of any group produces `\sqrt` → KaTeX `Expected group as argument to '\sqrt'` under `throwOnError`, and because this is a render-time failure rather than `MtefParseError`, the bare catch in `mathSerializer.ts` never fires: the importer emits a latex element that dies in the classroom. `a` + 0xD6 + `b` silently becomes `a\sqrt b` with `degraded=false`. No test pins 0xD6, so the suite stayed green. One-token fix plus a test that runs 0xD6 through KaTeX.

**P2. `0xf3: '\\\\ '` emits a LaTeX line break.** 0xF3 is `integraltp`, the top piece of a multi-piece integral. `a` + 0xF3 + `b` renders as two lines; inside `\frac` it breaks the fraction. By the file's own rule that extender pieces map to their base operator, this should be `\int ` (and 0xF4 `integralex` / 0xF5 `integralbt` currently throw, so the three pieces of one integral behave three different ways).

**P2. Test title still wrong.** `test/mtef.test.ts:215` still reads `renders [main, upper, lower] in the right roles`; the body comment and the code use `[main, lower, upper]`. Only the comment above it was reworded.

Non-blocking: the comment at the top of the table says 0xF7 is the integral extender, but 0xF7 is `parenrightex` (the test file's comment has it right; the integral extender is 0xF4); tmDIRAC var1/var2 discard the unused slot without setting `degraded` while var0 does; the header claims the high half is "complete" while 21 defined positions (`\Im \Re \wp \Upsilon`, the four suits, ®/©/™ serif and sans, `integralex`/`integralbt`, `carriagereturn`, `arrowvertex`/`arrowhorizex`) still abort the whole equation even though KaTeX has macros for most of them; approximations such as extenders → `|` or `)`, `florin` → `f`, `fraction` → `/` do not set `degraded`. Fine to leave for a follow-up, but the comment should not say complete.

CXuPercy:
@wyuc All round-2 findings addressed in dade8486:

- **P1-1 tmLIM**: main slot now emits first, limits follow, no injected operator name — matching the reference `39.1 = limit: lower, #1 #2`. `main=lim, lower=n` → `lim_{n}`; `main=max, lower=i` → `max_{i}`; empty main + `lower=x` → `\lim _{x}` (KaTeX valid, no glue). Added tests for all three probe shapes.
- **P1-2 Symbol high-half**: completed 0xA0–0xFF from the URW AFM (~70 positions: ∫ ∑ ∏ ⟨⟩ ∂ ∇ ⇒ ⇔ ⋅ ′ ∅ ⊆ ⊇ ∈ ∉ ∪ ∩ …). Unmapped font-local codes (Euro, Ifraktur, card suits, radicalex, C1 controls) throw `MtefParseError` so the picture fallback takes over. `0xF2 → \int` (was ÷), `0xD7 → \cdot` (was ×), `0xE5 → \sum` (was å). Additionally: `0xD6` uses `\surd` not `\sqrt` (bare radical CHARs have no argument), and `0xF3` uses `\int` not a line break — both caught by a cross-review round on this fix.
- **P2 tmDIRAC**: var1 = `\left\langle L\right|`, var2 = `\left| R\right\rangle` per the reference. KaTeX-validity for all variations.
- **P2 cap tests**: output-length width case (15k sibling macro CHARs throws), record-count case (20k throws) — both have dedicated tests alongside the existing depth test.
- **P2 fixture claim**: corrected (the three real streams contain no big-op selector; that reading is pinned by the spec-conformance tests); test title reads `[main=k, lower=i=1, upper=n]`.
- **P3 embellishment budget**: the `for(;;)` loop now increments `budget.count`.

Suite: 91/91 (6 new). 8k fuzz unchanged (all `MtefParseError`, zero other exceptions). Real-deck e2e unchanged.

wyuc:
Round 4 verified on dade8486: 0xD6 → `\surd` and 0xF3 → `\int`, both KaTeX-clean standalone and in context, 0xF4 still throws into the picture fallback, 91/91. To save a round-trip I pushed f0ab791a on top with the three leftovers from the last review: a test that runs 0xD6/0xF3 through KaTeX, the BigOp test title corrected to `[main, lower, upper]`, and the table comment no longer claiming the high half is complete. Merging once CI is green on that head.

wyuc:
afc7f1d5 is what I want to merge, but #1424 just landed on main and touches the same importer files (`package.json` version line now 0.1.5 on main vs 0.2.0 here, plus `mediaWebConvert.ts` / `imageSerializer.ts`), so the PR is CONFLICTING. Could you merge `main` into the branch and resolve (keep 0.2.0)? I will merge as soon as CI is green on the resolved head.

wyuc (dismissed):
Four rounds converged on afc7f1d5: LSCRIPT duplication and the nesting blow-up removed with depth/output/record caps all pinned by tests; one-sided fences, tmLIM, DIRAC and the Adobe Symbol table verified against the encoding vector and KaTeX, with 0xD6/0xF3 tested through KaTeX; three genuine Equation Native fixtures; fuzz clean. Merging at 0.2.0.

wyuc (approved):
Re-approving after the rebase onto main (linear, PR content unchanged, #1424's importer fixes preserved, version 0.2.0). Four rounds converged: LSCRIPT duplication and the nesting blow-up removed with depth/output/record caps pinned by tests; one-sided fences, tmLIM, DIRAC and the Adobe Symbol table verified against the encoding vector and KaTeX; three genuine Equation Native fixtures; fuzz clean. Merging.

PR #1414, opened 2026-09-08
fix(export): surface video render rejection reasons

### LINKED ISSUE #1349: video export: surface the actual render-service rejection reason in the UI
### Problem or Motivation

When the render service rejects an export submission (429 queue full, 429 per-identity limit, 413 archive too large), the frontend **parses the reason correctly but discards it at the presentation layer**:

- `lib/store/video-render.ts` extracts `data.error` / `data.details`, records `submitStatus`, and deliberately distinguishes real rejections (429/413/5xx) from "service unavailable" (501 / fetch threw) — the failure lands in `store.error`.
- But the toast is hardcoded to the generic key `export.videoFailed` ("Video export failed" / "视频导出失败"), and `store.error` is never rendered anywhere (the export dialog only shows the progress bar).

The user cannot tell "the queue is busy, retry in a minute" from "your export is broken". After a queue-full rejection they get no actionable signal, even though the store deliberately kept the failure "honest and retryable".

### Proposed Solution

Requires the machine-readable `reason` code from #1348 (`queue_full` | `per_identity_limit`).

1. **Proxy passthrough**: `app/api/export-video/render/route.ts` currently collapses any upstream 429 to `errorCode: RATE_LIMITED` with prose in `details`. Forward the upstream `reason` as a distinct field so the client can branch on it.
2. **Reason → localized message mapping** in the video-render store's failure path:
   - `queue_full` → new key `export.videoQueueFull` ("Render queue is busy — try again shortly")
   - `per_identity_limit` → new key `export.videoRenderInProgress` — **neutral wording**: under the default shared `direct` identity bucket, the in-progress render may belong to a different user, so the message must not accuse the reader of already having one running
   - 413 → `export.videoTooLarge`
   - unknown / 5xx → existing `export.videoFailed` fallback
3. **Render `store.error`** (toast description or in-dialog line) so non-rejection failures also show their actual message instead of only the generic label.

### Acceptance Criteria

- [ ] A submit rejected with `queue_full` shows the queue-busy message in the user's locale
- [ ] A submit rejected with `per_identity_limit` shows the render-in-progress message (neutral wording)
- [ ] A 413 shows the too-large message
- [ ] Other failures still show the generic `export.videoFailed`
- [ ] All 12 locales gain the new keys
- [ ] Existing behavior unchanged: 501 / unreachable still degrades to the ZIP download with `export.videoServiceUnavailable`; a failure after the render started still cancels the server job

### Alternatives Considered

- **Show the raw English prose from the service** — rejected: untranslated text leaking into a 12-locale UI is why the generic key was chosen in the first place.
- **String-match the prose to pick a message** — rejected: brittle; this is what the machine-readable `reason` code exists for.

### Area

Import / export

### Additional Context

See #1348 for the `reason` field. Frontend submit parsing lives in `lib/store/video-render.ts` (`submitStatus` / `data.error` / `data.details`); toasts are sonner-based and fire from the store, i18n keys live in `lib/i18n/locales/*.json` under `export.*`.

YizukiAme:
## Summary

When video submission is rejected because the queue is full, an identity already has a render, or the archive is too large, the export toast explains the condition in the selected language. Unrecognized failures use the existing generic localized message; raw upstream diagnostics are not displayed in toasts.

Fixes #1349. Uses the admission reasons introduced by #1351.

## Changes

- Forward only `queue_full` and `per_identity_limit` on HTTP 429, using the shared `apiError()` envelope with an optional `reason`.
- Map those reasons and HTTP 413 to specific messages in all 12 locales. Keep identity-limit wording neutral for direct deployments that share an identity.
- Keep unknown submission, compilation, polling, download, and timeout errors generic in user-facing toasts. Diagnostic responses/state remain available for troubleshooting.
- Rebase onto current `main`, retaining #1456's ZIP retry cache and observer lifecycle, including counter decrement and observer release after compilation failure.
- Preserve ZIP fallback for an unconfigured/unreachable submission endpoint and cancellation after an accepted job fails.

## Type of Change

## Verification

On Node 22.23.2 with the frozen lockfile:

- **85 focused tests passed** across the route, store, locale, polling-driver, and ZIP retry-cache suites. Covers known/unknown/invalid reasons, local and upstream size limits, generic failure presentation, retry, cancellation, ZIP fallback, and observer cleanup.
- Updated regression assertions failed against the pre-fix implementation because raw diagnostic descriptions and an unknown reason were still exposed, then passed with the fixes.
- **Full root unit suite: 8,491 passed, 43 skipped; 741 files passed, 12 skipped.**
- `pnpm check`, `pnpm lint`, `pnpm exec tsc --noEmit`, and `pnpm check:i18n-keys` passed. Lint reports 18 warnings in unchanged files and no errors.

The live render service/encoder was not exercised. GitHub CI status is shown in the checks below.

### Manual verification steps

1. Enable video export and open the export dialog with a slide available.
2. Submit with HTTP 429 and `reason: "queue_full"`, HTTP 429 and `reason: "per_identity_limit"`, or HTTP 413. Check the specific localized toast and that retry is available.
3. Return an unknown rejection or HTTP 502 with internal paths/URLs in its diagnostic details. Check that only the generic failure toast is shown, without a diagnostic description.
4. Fail an accepted job or its poll/download request. Check the generic toast and cancellation request. Retry a rejected submission to check cached ZIP reuse.

## Checklist

- [x] Code follows the project's coding style
- [x] Self-review completed
- [x] Documentation considered; no operator configuration changes
- [x] No new lint warnings

AI-assisted with Codex. Self-review covered response compatibility, localization, ZIP fallback, retry-cache lifecycle, and cancellation behavior.


wyuc (changes requested):
Thanks @YizukiAme — surfacing the admission reasons is exactly what #1349 asked for. The three localized messages are complete across all 12 locales, the ZIP fallback and post-accept cancellation paths are unchanged, and removing the reason forwarding or the 413 mapping turns the new tests red. Two things to address before merge:

### 1. Rebase onto `main` (conflict in `lib/store/video-render.ts`)

#1456 (reuse the compiled ZIP when retrying video renders) landed after this branch and added the cache / observer lifecycle (`cachedZip`, `activeRenders`, `releaseIdleObservers`, …). The resolution is easy to get subtly wrong:

- Near the top of the file, keep **both** main's cache / observer block and your `rejectionMessageKey()`.
- In the compile-failure `catch`, keep your `set({ status: 'failed', error: message })` and toast `description`, **keep** main's `activeRenders -= 1; releaseIdleObservers();` (the compile path returns before the render `try` whose `finally` decrements, so dropping it leaks the counter and observers), and **drop** main's `set({ error: 'compile' })` (it would overwrite your message and fail your compile-error test).

With that resolution the route/store typecheck and your 51 declared tests pass.

### 2. Don't show raw upstream error text to end users

For unrecognized rejections, 5xx responses and failures after acceptance, the toast `description` now shows the render service's free-text error verbatim — no length bound, no redaction. Real messages the service can produce include `… Check PRODUCER_HEADLESS_SHELL_PATH and Chromium compatibility.` (an internal env var name), the raw producer / ffmpeg `Error.message` (container paths, stderr), and `Archive entry too large: <entry name>`. `render-coordinator` deliberately doesn't even log that free text, and this would put it in front of users (and in their screenshots). Please keep the user-facing text generic for unrecognized failures — at most a short, length-capped summary with paths and URLs stripped — and leave the full detail to the server logs / the network response.

### Non-blocking

- `reason` is forwarded as any string; allow-listing the known codes (`queue_full`, `per_identity_limit`) would keep the contract explicit.
- The route builds the `{ success, errorCode, error, details }` envelope by hand and then adds `reason`; extending `apiError()` with an optional `reason` would avoid drifting from the shared envelope.

YizukiAme:
@wyuc Addressed both blocking items and the two non-blocking suggestions in the updated branch (head: `6ac3e375`).

- Rebased onto `main`, retaining #1456's ZIP cache and observer lifecycle. The compile-failure path still decrements `activeRenders` and releases idle observers without overwriting the diagnostic state.
- Removed raw error descriptions from failure toasts, including compilation and post-accept failures. Known rejections keep their localized messages; unknown failures use the existing generic message. Diagnostic responses/state remain available for troubleshooting.
- Restricted forwarded admission reasons to `queue_full` and `per_identity_limit`, and reused `apiError()` with an optional `reason`.

Validation: 85 focused tests passed, including cache/observer cleanup and generic error presentation; the full root suite passed with 8,491 tests and 43 skipped. Formatting, lint, typecheck, and i18n checks passed, and all six GitHub CI checks are green on the updated head. The live render service/encoder was not exercised.

Ready for another review.

wyuc (approved):
Thanks @YizukiAme — both blocking items are resolved, verified independently at 6ac3e375:

1. **Rebase over #1456**: the cache / observer lifecycle is fully retained, and the compile-failure path keeps your message while still decrementing `activeRenders` and calling `releaseIdleObservers()` exactly once. Removing either on that path turns `video-render-cache.test.ts` red.
2. **No raw service text in toasts**: every user-visible path (known rejections, unknown reasons, 4xx/5xx, compile failure, post-accept failure) now shows a localized or generic message; raw text stays in diagnostics and logs only. Re-adding a raw `description` turns the store tests red.

Reason forwarding is allow-listed to `queue_full` / `per_identity_limit` on 429 only, the 413 mappings remain, and the optional `reason` on `apiError()` leaves every other caller's response unchanged. Full suite, i18n key check, lint, typecheck and `next build` all pass.

Optional follow-up: type `reason` as the allow-listed union instead of `string`, so a future caller can't emit an arbitrary value.

Approving — I'll update the branch and merge once CI is green.

PR #1418, opened 2026-09-08
fix(workbench): show effective upload limits in friendly errors

### LINKED ISSUE #1415: [Bug]: Show a friendly Pro upload size error with the effective server limit
## Bug Description

Pro material uploads expose a technical error when a file exceeds the upload size limit. The shared composer displays the backend message verbatim, for example `upload exceeds 52428800 bytes [requestId=…]`, rather than a localized explanation of the allowed file size.

Classic mode already uses a localized “file too large” message. Pro should follow that wording style while displaying its actual configured limit.

## Steps to Reproduce

1. Open a Pro launch composer or an existing Pro conversation on a deployment with material uploads enabled.
2. Attach a supported file larger than the effective per-file limit (50 MiB by default).
3. The upload endpoint returns HTTP 413 and the composer surfaces the raw error.

These steps are derived from source inspection of upstream main at `64621adfa047097f9594bb535e4a190c8d7a32eb`; browser reproduction has not been performed.

## Expected Behavior

- Show a localized message such as **“File too large. Please select a file no larger than 50 MB.”** / **“文件过大，请选择不超过 50MB 的文件。”**
- Populate the number from the server's effective limit for the rejected file, rather than hardcoding 50. If an operator sets the effective limit to 100 MiB, the message should show 100 MB.
- If a 413 response has no limit metadata (for example, an intermediary rejects the request), show a localized fallback: **“File too large. Compress or split the file and try again.”**
- Apply the same handling to both Pro composers and all supported locales.

## Suggested Change

Include a structured byte limit in the application's 413 response and use it when formatting the localized upload error. Keep the request ID available for diagnostics without placing it in the size-limit message.

The default is `50 * 1024 * 1024` bytes. Documents/images use `min(MATERIALS_MAX_DOCUMENT_BYTES, OPENMAIC_AGENT_MAX_UPLOAD_BYTES)`; audio/video use `OPENMAIC_AGENT_MAX_UPLOAD_BYTES`. The displayed value should reflect that effective per-file limit.

Scope: error presentation only. No upload-button text, tooltip, layout changes, or new frontend pre-upload checks are needed.

## Source Evidence

- [Pro upload limits](https://github.com/THU-MAIC/OpenMAIC/blob/64621adfa047097f9594bb535e4a190c8d7a32eb/lib/server/agent-runtime/config.ts#L45-L48)
- [Effective document/image limit](https://github.com/THU-MAIC/OpenMAIC/blob/64621adfa047097f9594bb535e4a190c8d7a32eb/app/api/materials/route.ts#L69-L73)
- [Declared-length rejection](https://github.com/THU-MAIC/OpenMAIC/blob/64621adfa047097f9594bb535e4a190c8d7a32eb/app/api/materials/route.ts#L193-L200) and [streamed-body rejection](https://github.com/THU-MAIC/OpenMAIC/blob/64621adfa047097f9594bb535e4a190c8d7a32eb/app/api/materials/route.ts#L297-L307)
- [Client preserves raw message and request ID](https://github.com/THU-MAIC/OpenMAIC/blob/64621adfa047097f9594bb535e4a190c8d7a32eb/lib/workbench/session-store.ts#L2156-L2164)
- [Shared composer displays the error](https://github.com/THU-MAIC/OpenMAIC/blob/64621adfa047097f9594bb535e4a190c8d7a32eb/components/workbench/compose-extras.tsx#L374-L381)

Affected area: Pro material uploads / UI. Deployment, browser, and OS: not runtime-tested; this report is based on source inspection.

## Follow-up

#1416 discusses whether Classic mode should also support deployment-configurable upload size limits. That decision is independent of this error-message fix.

YizukiAme:
## Summary

Pro uploads that exceed the server limit currently show raw errors such as `upload exceeds 52428800 bytes [requestId=…]`. This change shows a localized file-size message using the effective server limit, with a generic friendly fallback when a 413 has no limit metadata.

For example, the default configuration shows “文件过大，请选择不超过 50MB 的文件。” A deployment with an effective 100 MiB limit shows 100 MB instead.

## Related Issues

Closes #1415.

## Changes

- Include `maxBytes` in both declared-length and streamed-body size rejections, using the existing per-file policy for documents/images and audio/video.
- Preserve the status, raw error, and request ID for diagnostics while formatting 413 messages for the shared Pro composer.
- Add localized limit and fallback messages for all 12 supported languages.

## Type of Change

### Steps to reproduce / test

1. In either Pro composer, attach a supported file larger than the configured limit and check the localized maximum size message.
2. Change the effective server limit and verify the displayed number follows it. Documents/images must use the smaller of the document and upload limits; audio/video must use the upload limit.
3. Return an HTML 413 or a 413 without valid limit metadata and check the generic localized fallback.

Automated validation under Node 22.23.2:

- Five focused suites passed: material client, upload scheduling, workbench locales, materials route, and Pro entry points (147 tests).
- Regression cases cover 50/100/12.5 MB copy, per-class configured limits, both backend size-rejection paths, missing/invalid metadata, HTML proxy 413s, and retained diagnostics.
- Full Prettier check, TypeScript (`tsc --noEmit`), i18n key alignment, and package version checks passed. Full ESLint passed with 18 warnings in unchanged files.

Browser upload reproduction has not been performed. The attachment control and layout are unchanged.

### Evidence

```text
Test Files  5 passed (5)
     Tests  147 passed (147)
```

- [ ] CI passes
- [ ] Manually tested locally in a browser

## Checklist

- [x] Documentation changes are not needed for this error-presentation fix
- [x] My changes do not introduce new lint warnings


wyuc (commented):
Thanks for this — the core design is right: the number shown comes from the threshold the route actually enforced (both the declared-length and streamed-body paths, including the `min(document, upload)` case), so the UI can't drift from the server. Two things before merge:

**1. The limit is rendered as a raw binary fraction.** `lib/workbench/session-store.ts` passes `this.maxBytes / 1024 / 1024` straight into the translator, which stringifies it. With a decimal setting such as `OPENMAIC_AGENT_MAX_UPLOAD_BYTES=100000000` the toast reads:

> File too large. Please select a file no larger than 95.367431640625 MB.

Only exact MiB multiples render cleanly today. Suggest formatting with `Intl.NumberFormat(locale, { maximumFractionDigits: 1 })` (rounding down so we never overstate the limit), using the active locale so de/fr etc. get the right decimal separator. A test with a non-MiB-multiple value would pin it.

**2. The user-visible wiring is untested.** `components/workbench/compose-extras.tsx` is where `WorkbenchMaterialUploadError` is turned into `err.userMessage(t)` for the toast, but nothing exercises that branch — replacing it with `err.message` leaves all related suites green. Could you add a test that stubs a 413 upload through `useComposerMaterials`/`addFiles` and asserts `toast.error` receives the localized message with the limit?

Non-blocking:
- The header contract comment in `app/api/materials/route.ts` doesn't mention the new `maxBytes` field on the 413 body — worth documenting.
- `tests/workbench/workbench-i18n.test.ts` only checks per-locale overlay coverage for `tool.*` keys, so a locale later dropping `material.fileTooLargeWithLimit` would silently fall back to English. All 12 locales carry both keys today, so this is just robustness.
- Non-413 errors (415 unsupported type, 429 quota) still surface raw server text — out of scope for #1415, just noting it.

wyuc:
@YizukiAme, friendly nudge on this one in case the review above slipped past. The two points in my review from 09-18 are the only things left (formatting the limit instead of showing a raw binary fraction like `95.367431640625 MB`, plus the second item there). The core design is already right. Thanks!

YizukiAme:
@wyuc Thanks for the nudge — this one got buried under other emails 😂 I'll go back through the review shortly and get it finished.

YizukiAme:
@wyuc Both points are addressed in 7afc6f85, and CI is green on it.

**1. Limit formatting**

The limit is now formatted with the active locale before it reaches the translator: at most one fractional digit, rounded down, no grouping. `100000000` bytes now shows `95.3 MB` (`95,3` in de/fr).

A few things go slightly beyond the suggestion:

- `maximumFractionDigits: 1` on its own rounds half-expand, so `95.367…` would become `95.4`. I set `roundingMode: 'floor'` and check `resolvedOptions()`. If a runtime can't floor, or `Intl` throws, the toast uses the generic message rather than risk overstating the limit.
- Caps below 0.1 MiB would floor to `0`, so they also use the generic message. `error.maxBytes` still keeps the exact value.
- I pinned `numberingSystem: 'latn'`. Without it, `ar-SA` would show `٥٠` where it showed `50` before. Happy to drop it if you'd prefer native digits there.

Tests pin a non-MiB cap in en/zh/de/fr/ar, the 0.1 MiB boundary, both fallbacks, and no grouping at 1024 MB.

**2. Toast wiring**

`tests/workbench/composer-materials-toast.test.ts` mounts `useComposerMaterials` under the real `I18nProvider`, stubs a 413 through `addFiles`, and asserts `toast.error` receives the full localized sentence (literal strings for en/zh/de). Replacing `err.userMessage(t, locale)` with `err.message` fails it. It also covers the generic toast for JSON and HTML 413s without `maxBytes`.

**Non-blocking**

- The route header now documents `maxBytes` on the two size-cap 413s. The separate declared-length-mismatch 413 has no `maxBytes`, and a route test pins that.
- The i18n test requires both keys in each base and each raw overlay file, with the expected placeholders, and checks that each overlay is actually registered.
- 415/429 are unchanged. Quota 429s go through the same retry as rate-limit 429s, so they're retried twice before the error shows. I opened #1646 for that and the 415/429 copy.

I also checked it in a browser. On a local workbench with the effective upload limit temporarily set to 1 MiB, I opened `/workspace` in a headless browser and attached a PDF slightly over 1 MiB with the home page attachment button. The toast read `File too large. Please select a file no larger than 1 MB.` in English and `文件过大，请选择不超过 1MB 的文件。` in Chinese, instead of `upload exceeds … bytes`.

wyuc (approved):
Re-reviewed 7afc6f85: both items are fixed and pinned. The limit is now floor-formatted per locale with a guarded fallback, and the composer toast wiring is covered end-to-end through `addFiles`. The limit comes only from the server's `maxBytes`, all 12 locales have both keys, and non-413 errors pass through unchanged. Thanks @YizukiAme!

Non-blocking follow-ups (P3):
- The declared-length-mismatch 413 (`route.ts` ~356) has no `maxBytes`, so it shows the generic "file too large, compress or split" copy even though the file is within the limit. Browsers can't reach it, but a `reason` field (or `maxBytes`) would let the client tell the two apart.
- The label says "MB" but the math is MiB, so a decimal config like 50000000 renders as 47.7 MB. This predates the PR; switching to MiB later would make it exact.
- 415/429 raw text is tracked in #1646.

PR #1419, opened 2026-09-08
fix(ssrf): keep cloud metadata endpoints blocked under ALLOW_LOCAL_NETWORKS

wyuc:
## Summary

`ALLOW_LOCAL_NETWORKS=true` is the documented opt-in for self-hosted deployments that need loopback, RFC1918 or split-horizon provider targets. It returned from `validateUrlForSSRF` before any classification, so on such a deployment a client-supplied base URL of `169.254.169.254` (or `metadata.google.internal`, `100.100.100.200`, `fd00:ec2::254`) was accepted. Local networks and cloud instance-metadata endpoints are different things; this keeps the metadata set blocked regardless of the flag.

Hardening follow-up to GHSA-9m7h-vh2h-rc3w / CVE-2026-86259. It is not a new vulnerability: the flag is documented as unsafe for public deployments, and the default deployment never sets it.

## Changes

- `lib/server/ssrf-guard.ts`: under the flag, literal hosts (including IPv4-mapped IPv6) and the known metadata hostname are rejected with a distinct message; non-IP hostnames are resolved and rejected if any answer is a metadata address. DNS failure under the flag still fails open (split-horizon DNS is a flag use case). Without the flag the DNS path now also rejects answers on metadata addresses that are not RFC1918 (`100.100.100.200`). The IP canonicalisation used by `assertSafeIp` is factored into one helper shared with the new check.
- `.env.example`: the flag comment says metadata endpoints stay blocked.
- Tests: new cases in `tests/server/ssrf-guard.test.ts` for every metadata literal, mapped form, DNS answer, RFC1918-only DNS, loopback/private still allowed, DNS error fail-open, and the flag-off path. The four route tests that used the metadata address as their example of a flag-allowed target now use a private-network address.

## Verification

- `pnpm exec vitest run tests/server/ssrf-guard.test.ts tests/server/*url-guard*.test.ts tests/document/extract-document-route.test.ts tests/agent-runtime/fetch-url*.test.ts` → 7 files, 108 tests passed
- `pnpm exec tsc --noEmit`, eslint and prettier on the changed files → clean

## Type of Change

https://claude.ai/code/session_01GRaNc5E88r41GUWt2Y3uiQ

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -459,7 +459,10 @@ DEFAULT_MODEL=
 
 # --- Local/Self-hosted Deployment ---------------------------------------------
 # Set to "true" to allow private/local network URLs (e.g. localhost, 192.168.x.x).
-# Required for self-hosted models like Ollama. Do NOT enable on public deployments.
+# Required for self-hosted models like Ollama. This flag never unblocks cloud
+# instance metadata endpoints (169.254.169.254, 100.100.100.200, fd00:ec2::254,
+# metadata.google.internal): those stay blocked even when it is enabled. Do NOT
+# enable on public deployments.
 # ALLOW_LOCAL_NETWORKS=true
 
 # Build-time, space-separated CSP frame-ancestor sources in addition to 'self'.
--- a/.env.example
+++ b/.env.example
@@ -462,9 +462,10 @@ DEFAULT_MODEL=
 # Required for self-hosted models like Ollama. Do NOT enable on public deployments.
 # The known cloud instance-metadata and credential endpoints (169.254.169.254,
 # 169.254.170.2, 169.254.170.23, 100.100.100.200, 168.63.129.16, 192.0.0.192,
-# fd00:ec2::254, fd00:ec2::23, metadata.google.internal) stay blocked even with
-# this flag. That is a fixed address list, not a general link-local block, and a
-# hostname whose DNS lookup fails or times out is still allowed through.
+# fd00:ec2::254, fd00:ec2::23, metadata.google.internal) are blocked with or
+# without this flag. That is a fixed address list, not a general link-local
+# block, and under the flag a hostname whose DNS lookup fails, times out or
+# returns no answer is still allowed through.
 # ALLOW_LOCAL_NETWORKS=true
 
 # Build-time, space-separated CSP frame-ancestor sources in addition to 'self'.
--- a/.env.example
+++ b/.env.example
@@ -459,10 +459,12 @@ DEFAULT_MODEL=
 
 # --- Local/Self-hosted Deployment ---------------------------------------------
 # Set to "true" to allow private/local network URLs (e.g. localhost, 192.168.x.x).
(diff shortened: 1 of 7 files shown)


wyuc:
@uziii2208 The advisory thread for GHSA-9m7h-vh2h-rc3w is closed now that it is published, so replying to your last comment there here.

Thanks for the detailed follow-up, and apologies for publishing without replying first. That was a process gap on our side; how we handle severity objections and CVEs going forward is written down in #1417.

**Score.** The published advisory keeps the vector you set (CVSS 4.0, 9.0). We read Attack Requirements the same way you now have it (AT:P), and we are not going to relitigate the Subsequent System metrics on a published advisory. CVE-2026-86259 is now linked from the advisory so the two records match.

**`ALLOW_LOCAL_NETWORKS=true`.** That flag is the documented operator opt-in for self-hosted deployments and `.env.example` says not to enable it on public ones, so we are treating it as a hardening request rather than a separate vulnerability, and we will not open a second advisory for it. You are right that a local network and a cloud metadata endpoint are different things, though, which is what this PR fixes: the metadata addresses stay blocked even when the flag is set. Thanks for flagging it.

Going forward we will request the CVE ourselves at publication time for every accepted advisory, so the description, affected versions and score come from one place.

uziii2208:
> @uziii2208 The advisory thread for [GHSA-9m7h-vh2h-rc3w](https://github.com/THU-MAIC/OpenMAIC/security/advisories/GHSA-9m7h-vh2h-rc3w) is closed now that it is published, so replying to your last comment there here.
> 
> Thanks for the detailed follow-up, and apologies for publishing without replying first. That was a process gap on our side; how we handle severity objections and CVEs going forward is written down in #1417.
> 
> **Score.** The published advisory keeps the vector you set (CVSS 4.0, 9.0). We read Attack Requirements the same way you now have it (AT:P), and we are not going to relitigate the Subsequent System metrics on a published advisory. [CVE-2026-86259](https://github.com/advisories/GHSA-6j8v-fv4p-986p) is now linked from the advisory so the two records match.
> 
> **`ALLOW_LOCAL_NETWORKS=true`.** That flag is the documented operator opt-in for self-hosted deployments and `.env.example` says not to enable it on public ones, so we are treating it as a hardening request rather than a separate vulnerability, and we will not open a second advisory for it. You are right that a local network and a cloud metadata endpoint are different things, though, which is what this PR fixes: the metadata addresses stay blocked even when the flag is set. Thanks for flagging it.
> 
> Going forward we will request the CVE ourselves at publication time for every accepted advisory, so the description, affected versions and score come from one place.

Hi @wyuc.
Thanks for the detailed follow-up and for addressing the metadata/local-network distinction so quickly - the fix looks solid.

No hard feelings on the process side; these things happen. Good to see it documented in #1417 going forward.

Happy to keep an eye on **OpenMAIC** and report anything I find. Good luck with the project.
@uziii2208 - Security Researcher

PR #1420, opened 2026-09-08
fix(renderer): lazy-load optional ECharts runtime

### LINKED ISSUE #1384: echarts is an optional peer but statically imported, costing ~182 KB gzip on chart-less decks
Using `@openmaic/renderer@0.1.6`.

`echarts` is declared an **optional** peer ("install only if your slides use the corresponding element type"), but it is imported **statically**, so it lands in the eager graph whether or not a deck contains a chart element.

The edge is a single one: `SlideElement.js` → `elements/chart/BaseChartElement.js` → `elements/chart/Chart.js`, which does

```js
import { BarChart, LineChart, PieChart, ScatterChart, RadarChart } from 'echarts/charts';
import { LegendComponent } from 'echarts/components';
import { SVGRenderer } from 'echarts/renderers';
// + echarts/core
```

`shiki`, by contrast, is already loaded lazily (two real `import('shiki')` call sites), which is exactly the treatment being asked for here.

**Measured** in a Vite 7 production build of an app that lazy-loads the renderer on one route:

| | gzip |
|---|---|
| renderer chunk total | **257 KB** |
| of which echarts | **~182 KB** |
| renderer's own slide engine | ~79 KB |

So a consumer whose generated slides happen to contain no chart still pays roughly 3× the engine's own cost. For context, that chunk is larger than our entire initial application download (213 KB gzip).

**Suggestion.** Load the chart element the way the code element already loads shiki — a dynamic `import('echarts/...')` inside `Chart.js` behind Suspense (or a small async init), so `BaseChartElement` resolves it only when an element of `type: 'chart'` is actually rendered. That would make the optional peer genuinely optional and take the eager cost down to the ~79 KB engine.

Happy to attempt a PR if you'd like, though the Suspense boundary is your design call.

LeoParkerOu:
Fixes #1384.

Load ECharts only when a chart is mounted, share one runtime-loading promise, keep rendering resilient when the optional peer is unavailable, and wait for chart readiness during slide PNG export. Bumps @openmaic/renderer to 0.1.7 and adds a regression test.

Validation: 15 renderer test files / 53 tests passed; TypeScript check, Rollup build, Prettier and ESLint passed. The generated bundle contains only dynamic import('echarts/...') references.

This PR was prepared with AI assistance and manually reviewed.


LING-6150:
I completed a second code-review pass on the current head (075a7815) and found no blocking issues.
Since this changes chart loading timing and the slide snapshot/export path, I also ran the flow locally:
- the live SlideCanvas reached ready and rendered the complete chart;
- slideToPng reached ready;
- the downloaded PNG matches the live chart, including the title, axes, labels, and values, with no blank or clipped content.
Screenshots of the live/export comparison and the downloaded PNG are attached.
One non-blocking test suggestion: the new test says it verifies that the chart option is applied, but currently only asserts the loading → ready transition and that init is called. It would be useful to also assert that setOption is called after the runtime resolves. This does not need to block the current review.
This touches the same snapshot/export path used by my video-export work, so I checked that consumer path explicitly.
cc @wyuc for a product-facing look at the rendering/export result. From the technical review side, this looks ready to approve.

Screenshot 2026-09-09 at 2 11 32 PM
Screenshot 2026-09-09 at 2 11 43 PM

LeoParkerOu:
Added the setOption assertion as suggested — extracted the spy to module scope so it can be verified directly after the runtime resolves. Local test passes (CI / Render Service check is also green).

LING-6150 (approved):
Thanks for adding the suggested assertion. I rechecked head 475182d8: the new commit only strengthens the chart regression test by asserting that setOption is called after the runtime becomes ready; no production code changed.
This does not invalidate the live rendering and PNG export E2E evidence from my previous review. All current checks are green, and I found no further blocking issues. Approving.
cc @wyuc — the technical review and product-facing rendering/export verification are complete. e2e I have finished on last comments.

wyuc (approved):
Verified the three capture paths that could race the lazy load: the render-service preview bundles echarts into its single iife (no external, so the dynamic import resolves locally inside the locked-down srcdoc), video-export frames go through slideToPng which now waits for chart readiness, and the classroom HTML export ships no app JS. Shared runtime promise, error-state fallback and cleanup all check out. Merging.

PR #1421, opened 2026-09-08
fix(render): keep preview browser evaluation self-contained

### LINKED ISSUE #1412: render-service: page.evaluate() helper leak breaks every real render, and no test can see it
`render-service` cannot render real scenes: every one of 20 randomly sampled production scenes failed, and the preview route tests cannot see it because they never start Chromium.

Found while syncing this module into a downstream fork. Probing the deployed service with 20 random real `document_scenes` rows gave **0 successes**.

## 1. `__name is not defined` — 17/20

```
HTTP 500 {"error":"__name is not defined\npptr:evaluate;waitForDocumentAssets%20(file%3A%2F%2F%2Fapp%2Fsrc%2Fpreview-renderer.ts...)"}
```

`waitForDocumentAssets` in `src/preview-renderer.ts` is handed to `page.evaluate()`. Puppeteer serializes such a function with `toString()` and evaluates the resulting source inside the page. The service runs under `tsx`, whose esbuild transform has `keepNames` behaviour, so the function is emitted wrapped as `__name(function waitForDocumentAssets(){...}, "waitForDocumentAssets")`. That helper exists in the module scope, not in the browser context, so the evaluated source throws immediately.

Any function passed to `page.evaluate()` from a `keepNames`-compiled module has this problem, so a fix should cover the pattern rather than the one call site — for example evaluating a source string, or a build/runtime setting that keeps evaluated functions unwrapped.

## 2. Preview rejects scenes whose media is referenced by URL — 3/20

```
HTTP 422 {"error":"Scene is not self-contained: 1 slide media reference(s) must use data: URLs"}
```

`preview-validation.ts` requires media to be inlined as `data:` URLs. The reasoning is sound under the egress lockdown — Chromium cannot fetch remote media anyway. But it changes the outcome for a scene that references media by URL from "renders without that imagery" to a hard failure, and a deployment that stores media as URLs then cannot preview most of its scenes.

Raising it as a question rather than a defect: is the intent that callers inline media before requesting a preview? If so it is worth stating in the route's contract, since the previous behaviour degraded instead. If not, making the requirement configurable would let a URL-backed deployment keep the older degradation.

## Why CI does not catch #1

`test/preview-route.test.ts` injects a fake `previewRenderer`, so no test in the package ever launches Chromium or renders a scene. The suite passes (18 files, 157 tests), `tsc --noEmit` is clean, and `/health` reports the service healthy — while every real preview returns 500.

A smoke test that renders one real scene through the actual renderer would close this. It is the only check that can distinguish "the service starts" from "the service renders", and both defects above would have been caught by it.

## Reproducing

POST a persisted slide scene (one with canvas elements, and for #2 one whose media is an `http(s)` URL) to `/preview` on a running service. The fixture in `preview-route.test.ts` is not a substitute — it has only ever been fed to a fake renderer, and a real render rejects it at validation.

hydraxman:
## Summary
Fix previews failing with `ReferenceError: __name is not defined` under the render service's `tsx` runtime.

`tsx` adds a module-scoped name helper to the nested asset-settling callback. Puppeteer sends only the callback source to Chromium, where that helper does not exist. Keep the browser program as JavaScript source so both page and iframe evaluation remain self-contained, without changing the asset waits or timeout behavior.

## Related Issues
Fixes #1412

## Changes
- Evaluate the existing font/image/layout settling program as a source IIFE.
- Add a required regression using the production `tsx` loader and an isolated evaluation context.
- Add real Chromium smoke tests for slide, interactive, quiz, and PBL previews. These run when `PUPPETEER_EXECUTABLE_PATH` or `PRODUCER_HEADLESS_SHELL_PATH` is configured.

## Type of Change

### Steps to reproduce / test
From `render-service/`:
1. `npm ci --ignore-scripts --no-audit --no-fund`
2. Set `PUPPETEER_EXECUTABLE_PATH` to a local Chromium executable.
3. `npm test` and `npm run typecheck`.

- On base `29735f10`, the isolated-context regression and all four real-browser renders failed with `__name is not defined`.
- On this head, `npm test` passed **163 tests across 19 files**, including real Chromium renders producing PNG signatures and 640×360 dimensions. Tested with Node 24.14.1 and Microsoft Edge's Chromium engine on macOS.
- `npm run typecheck` passed; changed-file Prettier and `git diff --check` passed.
- Self-contained-media validation, egress policy, abort handling, and browser cleanup are unchanged.
- Without a configured browser, the four Chromium tests skip; the `tsx`/isolated-context regression still runs. The existing CI does not configure a browser. I have not run the Docker image locally or verified production deployment.

### Evidence
Before: five regression failures, all `ReferenceError: __name is not defined`.
After: `Test Files 19 passed (19)` / `Tests 163 passed (163)`.

## Checklist
- [x] My changes do not introduce new warnings in the checks run

AI-assisted; an independent code review found no blocking findings.


LING-6150 (approved):
Thanks ,  I independently reproduced the base failure across the production tsx browser boundary (__name is not defined) and verified the current head with real Chrome.
The focused regression and the slide, interactive, quiz, and PBL Chromium renders all pass. The full render-service suite passes (163/163), along with typecheck. The change preserves the existing asset-waiting behavior and does not alter the media/egress policy.
No blocking findings from me. Approving.

@wyuc The real-browser cases are still opt-in rather than running in regular CI; that can be considered separately as CI hardening.

wyuc (approved):
Verified the trigger (the named inner `done` function is what keepNames wraps; the other four browser callbacks in this file have no named inner functions today) and that the source-IIFE form preserves the waits and timeouts. The tsx-boundary regression covers this call site; a pattern-level guard for every callback sent to the browser is tracked as a follow-up. Merging.

PR #1422, opened 2026-09-09
fix: tolerate malformed authored CSS in classroom exports

CXuPercy:
LLM-generated interactive-scene HTML can carry browser-tolerated CSS syntax errors such as `-- crop-green: #22c55e` (a space between `--` and the custom-property name). Browsers silently drop such declarations so the page renders fine, but the strict `postcss` pass used to inline assets for the resource-pack and classroom-zip exports throws (`interactive-css-parse-failed: ... Unknown word`), failing the entire export — while PPTX export, which never parses interactive HTML, keeps working.

This makes inlining best-effort instead of fatal:

- On strict-parse failure the stylesheet / style attribute / SVG presentation attribute is kept byte-for-byte (matching browser behavior) with a warning, and export continues.
- Remote dependencies inside unparseable CSS (`url()` / `@import`, absolute or relative, quoted with parentheses) are still surfaced through a comment- and string-aware fallback scanner, so exporters warn about un-inlined assets and offline validation is not silently bypassed.

Tested: 10 new regression cases in `tests/export/inline-assets.test.ts` (comment/string awareness, unterminated comments, `@import` with interleaved comments, dedup, function-name boundary); full export suite passes.

### DIFF
--- a/lib/export/css-asset-parser.ts
+++ b/lib/export/css-asset-parser.ts
@@ -7,6 +7,9 @@ export interface CssUrlReference {
   end: number;
 }
 
+/** CSS drops an escaped newline from a token value: url("a\<LF>b") is "ab". */
+const stripEscapedNewlines = (value: string) => value.replace(/\\\r\n|\\\n|\\\r|\\\f/g, '');
+
 export function cssUrlReferences(value: string): CssUrlReference[] {
   const refs: CssUrlReference[] = [];
   valueParser(value).walk((node: CssValueNode) => {
@@ -16,7 +19,11 @@ export function cssUrlReferences(value: string): CssUrlReference[] {
       (raw.startsWith('"') && raw.endsWith('"')) || (raw.startsWith("'") && raw.endsWith("'"))
         ? raw.slice(1, -1)
         : raw;
-    refs.push({ raw: unquoted, start: node.sourceIndex, end: node.sourceEndIndex });
+    refs.push({
+      raw: stripEscapedNewlines(unquoted),
+      start: node.sourceIndex,
+      end: node.sourceEndIndex,
+    });
     return false;
   });
   return refs;
@@ -52,7 +59,10 @@ export function cssImportReference(rule: AtRule): { url: string; conditions: str
         : raw;
   }
   if (!url) return null;
-  return { url, conditions: rule.params.slice(node.sourceEndIndex).trim() };
+  return {
+    url: stripEscapedNewlines(url),
+    conditions: rule.params.slice(node.sourceEndIndex).trim(),
+  };
 }
 
 export function parseCss(css: string, cssUrl: string): Root {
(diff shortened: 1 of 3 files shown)


wyuc (changes requested):
Thanks for moving this upstream. The mechanism is right (strict parse first, token scanner only for the malformed path, no fetching from the fallback, no lockfile change), well-formed CSS is byte-identical on the existing fixtures, and the new tests go red when the tolerance is reverted. One blocking gap in the scanner, then this is good to go.

**Blocking: the fallback scanner runs an unterminated string past unescaped newlines, which hides browser-active assets from both dependency reporting and residual validation.**

`lib/export/css-asset-parser.ts` (string branch in `collectCssAssetReferencesByRegex`, and the quoted-`url(` branch below it): `while (j < n && css[j] !== ch)` only stops at the matching quote or EOF. In CSS an unescaped newline ends the string as a bad-string token, and browsers recover at that point. So for

```css
.bad { content: "oops
;}
.ok { background: url(https://x/ok.png) }
```

Chromium still applies `.ok` with the remote background, and the strict parser on the repaired twin returns that URL, but the scanner and `collectCssAssetReferences` both return `[]`. `inlineHtmlAssets` then keeps the remote URL, performs no fetch, and reports `failed: []`, `inlined: []`, `unresolved: []`; only a generic parse warning is logged. Before this PR that export failed loudly; after it, the export succeeds with an unreported remote dependency, and the video export's offline-completeness check has nothing to reject. An unterminated string is one of the most common malformed-CSS shapes in authored slides, so this is squarely on the path the PR is meant to cover.

Fix: in both string-scanning loops, also stop at an unescaped `\n`, `\r`, or `\f` (treat it as the end of the string and resume scanning from that character), and add a regression test for the input above asserting the URL is collected and reported.

Non-blocking, for the record: an unbalanced unquoted `url(` consumes through the next `)`, merging whatever follows into one garbage URL. Chromium discards the following rule for that input too, so behaviour matches the browser; no change needed.

Verified locally on b14fcf26: `tests/export` 207/207, `tsc --noEmit`, eslint/prettier on the three files, and `pnpm check` all pass.

wyuc (changes requested):
Round 2 on 1a4725cd. The blocking finding is closed: the round-1 input now yields `https://x/ok.png` from the scanner and it shows up in `report.failed` / `unresolved` end to end; both bad-string `break`s are pinned (deleting either one turns two of the new tests red); `tests/export` 210/210, `tsc`, eslint and prettier are clean; scope and lockfile unchanged.

One regression introduced by this round, then it's mergeable.

**Blocking: a quoted string that is unterminated at EOF is now treated like a bad string, so its URL is dropped, while browsers keep it.**

`lib/export/css-asset-parser.ts`, the `if (terminated)` gates in the quoted-`url(` branch (~L166) and the bare-string `@import` branch (~L122). CSS Syntax §4.3.5 only produces a `<bad-string-token>` on an unescaped newline; on EOF it returns the `<string-token>` and the enclosing function / at-rule is auto-closed, so `.a { background: url("https://x/a.png` applies the background and `@import "https://x/a.css` imports (css-tree 3.2.1 parses both the same way). Behaviour now:

- `.a { background: url("https://x/a.png` → scanner `[]`; b14fcf26 returned `[{kind:'css-url', url:'https://x/a.png'}]`. Through `inlineHtmlAssets` the export reports `failed: []`, `unresolved: []` — the same silent-remote-dependency shape round 1 was about, now on the truncated-at-EOF case (which is exactly what an LLM-authored `<style>` block looks like when generation is cut off).
- `@import "https://x/a.css` → `[]`, was `[{kind:'css-import', …}]`.
- `url("a\` at EOF → `[]`; per §4.3.5 a backslash at EOF is consumed and the string is `"a"`.

Fix: distinguish "stopped at an unescaped newline/CR/FF" (suppress, resume at that character) from "ran off the end" (take `css.slice(start, n)` as the string, as before), and add the two EOF cases above to the regression tests.

Non-blocking, for the record: `stripEscapedNewlines` is also applied to *unquoted* `url()` values (strict and fallback paths). Per §4.3.6 a backslash followed by a newline inside an unquoted url is not a valid escape and yields a `<bad-url-token>` the browser rejects, so `url(https://x/a\<LF>b.png)` is now fetched as `https://x/ab.png` although the declaration is dropped by Chromium. It goes through the same proxied fetcher and SSRF validation as everything else, so the cost is an extra fetch and a possibly spurious `failed` entry, not a new reachability; fine to leave or to restrict the stripping to quoted strings.

wyuc (approved):
Round 3 on ca2ecf89 closes the EOF blocker: a quoted string that runs to EOF is kept as a string token (CSS Syntax §4.3.5), only an unescaped newline yields a bad string, the three EOF cases surface end to end in `report.failed` / `unresolved`, the round-1 bad-string recovery still holds, the new tests go red when the EOF branch is reverted, and `tests/export` is 212/212 with tsc/eslint/prettier clean. Thanks for the careful iterations — merging.

PR #1424, opened 2026-09-09
fix(importer): stop PPTX import from hanging in non-browser environments

### LINKED ISSUE #1423: [Bug]: PPTX import hangs in non-browser (worker) environments
### Bug Description

Importing a PPTX hangs indefinitely when the import runs in the non-browser PPTX import worker, which uses a linkedom DOM shim instead of a real browser DOM. 
The hang happens specifically for slides containing:

- a picture with an `a:clrChange` (chroma-key) color-replacement effect, or an EMF image (which the importer rasterizes to PNG via pdfjs-dist + canvas).
  

Both code paths `await` a signal that the linkedom DOM shim never produces, so the whole import stalls.

As reported in #1268, @wyuc said the import would "...turn indefinitely while parsing **a large PowerPoint**."

In fact, a large PowerPoint file isn't necessary; even a single-slide PPTX containing just one image can hang forever.

### Steps to Reproduce

- Create or obtain a PPTX whose content includes:
  - a picture with an `a:clrChange` effect, and/or
  - an EMF image.
- Trigger the import through the server-side worker path (not a real browser).
- Observe that the import never completes.

### Expected Behavior

The import completes. Images are decoded/rendered normally, or the importer degrades gracefully (skips the color replacement / emits a transparent placeholder) instead of hanging.

### Actual Behavior

The import hangs forever. Two paths wait on browser-only signals:

- `loadImageElement` waits on `img.onload`/`img.onerror`, which linkedom never
  fires.
- `emfPdfToPngDataUrl` waits on pdfjs `page.render()`, which cannot rasterize
  because linkedom's canvas exposes no usable 2D context.

### Deployment Method

### OpenMAIC Version

latest

### Affected Area

Import / export

### Browser

Not applicable

### Operating System

Not applicable

### Relevant Logs

```shell

```

### Screenshots or Recordings

_No response_

### Additional Context

Root cause: the linkedom DOM shim used by the import worker does not implement async image `load`/`error` events or 2D canvas rasterization, both of which the browser-facing importer relies on.

puxiao:
## Summary

Fix the PPTX import hanging in non-browser (worker) environments by making the importer's image-loading and EMF-PDF rasterization paths degrade gracefully instead of waiting forever on signals the linkedom DOM shim never produces.

## Related Issues

Closes #1423

## Changes

- `loadImageElement` — prefer the modern `decode()` signal; when it is absent, fall back to a 250ms timeout so the promise always settles to `null` instead of hanging.
- `emfPdfToPngDataUrl` — detect whether the environment has a usable 2D canvas before invoking pdfjs, and return a transparent PNG fallback when it does not.
- Add a 10s render deadline via `Promise.race` so a stuck pdfjs worker degrades instead of hanging forever.
- Move `doc.destroy()` into a `finally` so the pdfjs document is always released.
- Bump `@openmaic/importer` from `0.1.4` to `0.1.5` (patch).

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. In a real browser, import a PPTX containing an `a:clrChange` picture and an EMF image — both should still render normally.
2. In the server-side worker (linkedom) path, import the same PPTX — it should now complete, falling back to no-op color replacement / transparent PNG instead of hanging.

- Browser path still decodes images and applies `a:clrChange` as before.
- Non-browser path settles within the timeout instead of hanging.
- EMF with no usable canvas returns the transparent PNG fallback.

### Evidence

- [ ] Manually tested locally

## Checklist


wyuc (approved):
Verified against linkedom 0.18.13: its image element has no decode() and never dispatches load/error, and its canvas has no usable 2D context without the canvas package, so both fallbacks are the paths the worker actually takes. Merging. One follow-up request: a regression test that imports a PPTX with an a:clrChange picture through the linkedom worker path and asserts completion within a few seconds, so this cannot silently regress.

PR #1427, opened 2026-09-09
fix(generation): carry slide remark onto the canvas as script

### LINKED ISSUE #1383: buildCompleteScene drops GeneratedSlideContent.remark, losing the slide's narration
Using `@openmaic/generation@0.3.6`.

`GeneratedSlideContent` declares `remark`, and slide-content generation produces it — but `buildCompleteScene` never reads it. The canvas is assembled from `elements` and `background` only:

```js
const canvas = {
  id: nanoid(),
  viewportSize: 1000,
  viewportRatio: 0.5625,
  theme: defaultTheme,
  elements: content.elements,
  background: content.background,
};
```

So the narration written for a slide is discarded at assembly and never reaches the persisted scene. (`grep -c remark` gives 1 in `scene-types.d.ts` and 0 in `scene-builder.js`.)

**Why it matters for a consumer.** We assemble scenes with `buildCompleteScene` and store the result. The slides alone carried ~330 words for a whole course, while the teaching prose sat in `remark` and in the `speech` actions — so the stored deck read as bullet points with the lesson missing. It took us a while to work out the prose was being generated and then thrown away, rather than never generated.

**Suggestion.** Carry it onto the canvas. `@openmaic/dsl`'s `Slide` already has a `script` field for speaker notes, which looks like the intended home:

```js
const canvas = { /* ... */, script: content.remark };
```

We currently do this ourselves after `buildCompleteScene` and before `validateScene`, and the scene validates with it set — so the contract already accommodates it.

Happy to open a PR if that shape is agreeable.

apetcu:
## Summary

`buildCompleteScene` assembled the slide canvas from `elements` and `background` only, so the `remark` that slide-content generation produces (the slide's narration / speaker notes) was discarded at assembly and never reached the persisted scene. This carries it onto the canvas as `Slide.script`, the field `@openmaic/dsl` already defines for speaker notes and that the PPTX importer populates.

AI-assisted PR (Claude Code); reviewed and tested locally by @apetcu.

## Related Issues

Fixes #1383

## Changes

- `packages/@openmaic/generation/src/scene-builder.ts`: set `canvas.script` from `content.remark` when a remark was generated. When there is none, the canvas is unchanged (no `script` key), so existing documents and validators see no difference.
- `packages/@openmaic/generation/test/scene-builder.test.ts`: two tests. One asserts a generated remark lands in `canvas.script` and the scene still passes `validateScene`; it fails on `main` with `expected undefined to be 'Explain why…'`. The other pins that no `script` key appears when no remark exists.
- `packages/@openmaic/generation/package.json`: `0.3.6` → `0.3.7` (patch: compatible change to a published package, per CONTRIBUTING).

## Type of Change

### Steps to reproduce / test

1. `cd packages/@openmaic/generation && pnpm vitest run test/scene-builder.test.ts` on `main` — the new remark test fails (`canvas.script` is `undefined`).
2. Same command on this branch — 4 tests pass.
3. Build any slide scene through `buildCompleteScene` with content that has `remark` set and inspect `scene.content.canvas.script`.

- `pnpm test` in `packages/@openmaic/generation`: 26 files, 147 tests pass.
- `pnpm typecheck` in the package (src and test tsconfigs): clean.
- `node scripts/check-package-version-bumps.mjs <upstream/main sha>`: "Package version check passed."
- `prettier --check` and `eslint` on the touched files: clean.
- Not verified: end-to-end classroom playback. The change only adds a field the DSL already accepts; `validateScene` passes with it set, matching what the reporter observed when setting it manually.

### Evidence

- [x] CI passes (`pnpm check && pnpm lint && npx tsc --noEmit`) — package-level typecheck/lint/format run locally as listed above; repo CI will confirm.
- [x] Manually tested locally

## Checklist

- [x] I have added/updated documentation as needed (none required; the field is already documented in `@openmaic/dsl`)

---
Prepared with AI assistance (Claude Code); reviewed and tested locally by @apetcu.

### DIFF
--- a/packages/@openmaic/generation/package.json
+++ b/packages/@openmaic/generation/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openmaic/generation",
-  "version": "0.3.6",
+  "version": "0.3.7",
   "description": "Pure generation pipeline contracts and packaged prompt assets for MAIC consumers.",
   "type": "module",
   "main": "./dist/index.js",
--- a/packages/@openmaic/generation/src/scene-builder.ts
+++ b/packages/@openmaic/generation/src/scene-builder.ts
@@ -49,6 +49,7 @@ export function buildCompleteScene(
       theme: defaultTheme,
       elements: content.elements,
       background: content.background,
+      ...(content.remark ? { script: content.remark } : {}),
     };
     return {
       id: sceneId,
--- a/packages/@openmaic/generation/test/scene-builder.test.ts
+++ b/packages/@openmaic/generation/test/scene-builder.test.ts
@@ -32,6 +32,26 @@ describe('buildCompleteScene', () => {
     expect(first?.id).not.toBe(second?.id);
   });
 
+  it('carries the generated remark onto the canvas as speaker notes', () => {
+    const scene = buildCompleteScene(
+      slideOutline(),
+      { ...content, remark: 'Explain why constructor injection keeps tests honest.' },
+      [],
+      'stage-1',
+    );
+    expect(scene?.type).toBe('slide');
+    const canvas = scene?.content.type === 'slide' ? scene.content.canvas : undefined;
+    expect(canvas?.script).toBe('Explain why constructor injection keeps tests honest.');
+    expect(validateScene(scene)).toEqual({ valid: true });
+  });
+
+  it('leaves script unset when no remark was generated', () => {
+    const scene = buildCompleteScene(slideOutline(), content, [], 'stage-1');
+    const canvas = scene?.content.type === 'slide' ? scene.content.canvas : undefined;
+    expect(canvas).toBeDefined();
+    expect(canvas).not.toHaveProperty('script');
+  });
+
   it('honors an injected id across retries/upserts', () => {
     const first = buildCompleteScene(slideOutline(), content, [], 'stage-1', {
       sceneId: 'stable-scene-id',


LING-6150:
Thanks for the focused patch. Before proceeding, I’d like to confirm the intended persistence contract. Slide.script is currently documented as speaker notes imported from a source PPTX, while PPTX export derives its notes from speech actions. Having the DSL accept the field does not by itself establish that generated remark should populate it by default.
I’m holding approval pending clarification of that mapping and its relationship to speech actions. This is a scope question rather than a request for further implementation changes.

wyuc:
Thanks @apetcu for the focused patch and tests, and @LING-6150 for raising the contract question — I'm going to close this rather than merge.

`Slide.script` is reserved for speaker notes imported from a source PPTX, and a generated course's narration lives in its speech actions (which PPTX export also uses for notes). Writing the slide-content stage's `remark` into `script` would give that field two meanings and store a draft narration alongside the final one, so the two could disagree and export / editing would need precedence rules.

For library consumers (#1383): `buildCompleteScene` is called with the generated content, so `content.remark` is available at that point and can be persisted alongside the scene by the caller. If there's a real need for the generation-stage remark to become part of the scene model, please open an issue proposing a dedicated field and how it relates to speech actions and export — that's a design change rather than a bug fix.

PR #1431, opened 2026-09-09
feat(tts): allow custom endpoint path for custom TTS providers

### LINKED ISSUE #1428: [Feature]: Custom TTS providers cannot configure the request endpoint path
## Problem or Motivation

When a user registers a custom (OpenAI-compatible) TTS provider, OpenMAIC synthesizes speech by POSTing to `{baseUrl}/audio/speech`, and the path suffix is hard-coded in two places:

- Server side: `generateOpenAITTS` in `lib/audio/tts-providers.ts` builds the URL as `` `${baseUrl}/audio/speech` `` with no override.
- Client side: the TTS settings panel computes the same fixed suffix for the "Request URL" preview.

Any relay, gateway, or self-hosted engine that exposes its speech API under a different path — e.g. `https://gateway.example.com/v1/tts`, or an OpenAI-compatible proxy that nests services under per-tenant path prefixes — cannot be used at all, because only the base URL is configurable and the suffix is always appended.

## Proposed Solution

Add an optional **endpoint path** setting for custom TTS providers:

1. A per-provider option (`providerOptions.endpointPath`), defaulting to `/audio/speech` when unset — fully backward compatible.
2. A "Request path" input in the TTS settings panel, shown only for custom providers, with the "Request URL" preview reflecting it live.
3. Thread the option through every TTS invocation path (settings test, scene generation, multi-agent discussion), not just the settings preview — today non-VoxCPM providers drop stored `providerOptions` in `resolveAgentVoiceOptions`, which would silently discard this setting.

## Acceptance Criteria

- [ ] A user can set a custom request path (e.g. `/v1/tts`) on a custom TTS provider in the settings panel
- [ ] The "Request URL" preview reflects the configured path in real time
- [ ] Speech synthesis requests (settings test, scene generation, discussions) are sent to `{baseUrl}{endpointPath}`
- [ ] With no path configured, behavior is unchanged (`/audio/speech` is used)
- [ ] Built-in providers are unaffected

## Alternatives Considered

- **Encoding the full path in the base URL and detecting it**: ambiguous (where does the base end and the path begin?) and breaks the existing convention that the base URL ends at `/v1`. An explicit field is clearer and easier to validate.
- **Per-provider hard-coded adapters**: defeats the purpose of the generic OpenAI-compatible custom provider.

## Area

Model / provider integration

## Additional Context

- Affected code: `lib/audio/tts-providers.ts` (`generateOpenAITTS`), `lib/audio/agent-voice.ts` (`resolveAgentVoiceOptions`), `components/settings/tts-settings.tsx`.
- The `/api/generate/tts` route already forwards `ttsProviderOptions` into the provider config, so no API contract change is required.

Jamartin-create:
## What

Adds an optional, configurable **request endpoint path** for custom (OpenAI-compatible) TTS providers, instead of the hard-coded `/audio/speech` suffix.

## Why

Custom TTS providers previously had their request path hard-coded to `/audio/speech` in both the server-side synthesis path (`generateOpenAITTS`) and the settings-panel "Request URL" preview. Relays, gateways, or self-hosted engines that mount their speech API under a different suffix (e.g. `/v1/tts`, or per-tenant path prefixes) could not be integrated at all — configuring the base URL alone was not enough.

## How

- `lib/audio/tts-providers.ts` — `generateOpenAITTS` reads `providerOptions.endpointPath` (default `/audio/speech`; a leading slash is added when missing) and strips trailing slashes from the base URL before joining.
- `lib/audio/agent-voice.ts` — `resolveAgentVoiceOptions` now passes stored `providerOptions` through for non-VoxCPM providers, so the option takes effect on **every** TTS path (scene generation, multi-agent discussion, agent bar), not just the settings test.
- `components/settings/tts-settings.tsx` — a "Request path" input rendered only for custom providers; the "Request URL" preview and the settings test both honor it.
- 12 locale files — new `settings.ttsEndpointPath` string.

No API contract change: `/api/generate/tts` already forwarded `ttsProviderOptions` into the provider config.

## Test plan

- `pnpm format` ✅
- `npx eslint <changed files> --fix` ✅ (0 errors)
- `npx tsc --noEmit -p tsconfig.json` ✅
- `npx vitest run tests/audio tests/i18n` ✅ (40 files / 299 tests passed)

Screenshots: the new settings field is a small UI addition; screenshots can be added on request.

Closes #1428

### DIFF
--- a/components/settings/tts-settings.tsx
+++ b/components/settings/tts-settings.tsx
@@ -182,7 +182,9 @@ export function TTSSettings({ selectedProviderId }: TTSSettingsProps) {
               ...(ttsProvidersConfig[selectedProviderId]?.providerOptions || {}),
               ...(await getVoxCPMProviderOptions(effectiveVoice, { role: 'teacher', locale })),
             }
-          : undefined;
+          : isCustom
+            ? ttsProvidersConfig[selectedProviderId]?.providerOptions
+            : undefined;
       await startPreview({
         text: testText,
         providerId: selectedProviderId,
@@ -220,7 +222,14 @@ export function TTSSettings({ selectedProviderId }: TTSSettingsProps) {
     (isCustom ? providerConfig?.customDefaultBaseUrl : ttsProvider?.defaultBaseUrl) ||
     '';
   const endpointPath = (() => {
-    if (isCustom) return '/audio/speech';
+    if (isCustom) {
+      const customPath = ttsProvidersConfig[selectedProviderId]?.providerOptions?.endpointPath;
+      if (typeof customPath === 'string' && customPath.trim()) {
+        const trimmed = customPath.trim();
+        return trimmed.startsWith('/') ? trimmed : `/${trimmed}`;
+      }
+      return '/audio/speech';
+    }
     switch (selectedProviderId) {
       case 'openai-tts':
       case 'glm-tts':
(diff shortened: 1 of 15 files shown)


LING-6150:
Thanks for the contribution. We’ve reviewed the scope and are not expanding the generic custom TTS contract to support arbitrary endpoint paths at this stage.
The current custom-provider contract targets the OpenAI-compatible /audio/speech endpoint. This PR would add a persistent UI/configuration surface for non-standard variants, without a concrete supported integration demonstrating that changing the path alone provides full request/response compatibility.
We’re proceeding with dedicated provider integrations where the protocol and E2E target are explicit. Closing this as outside the current maintenance scope; we can revisit a generic endpoint override if a concrete supported integration requires it in the future.

PR #1435, opened 2026-09-09
fix(tts): prepare discussion segments during playback

### LINKED ISSUE #1434: [Bug]: Discussion TTS waits until playback ends to synthesize the next segment
### Bug Description

Discussion TTS waits for the current audio clip to finish playing before requesting the next sealed text segment. When the next segment is already available, this adds its entire synthesis latency to the silence between clips.

The problem is in the client queue and is independent of the selected server TTS provider.

### Steps to Reproduce

1. Enable a server TTS provider and start a classroom discussion that produces at least two text segments.
2. Let segment A start playing while segment B has already been sealed by the incoming stream.
3. Inspect requests to `/api/generate/tts` and the audio playback events.
4. Observe that B's request starts only after A's `ended` event.

A deterministic regression can reproduce this with mocked TTS responses and audio events: enqueue A and B, resolve A's response, leave A playing, and inspect whether B's request has started.

### Expected Behavior

Prepare the next available segment while the current one plays, with bounded lookahead and sequential playback. When B is ready before A ends, B should not incur another synthesis wait at the transition. Preserve pause/resume, cancellation, voice selection, and error fallback behavior.

### Actual Behavior

The queue runs `generate A → play A → generate B → play B`. In `useDiscussionTTS`, `isPlayingRef` is set before synthesis, blocks `processQueue`, and is only cleared when playback ends or fails. Enqueuing B during A's playback changes the indicator to `generating` without actually starting B's request.

Source at the inspected revision: [discussion TTS queue](https://github.com/THU-MAIC/OpenMAIC/blob/29735f10d0081859ac3db1a50a0cc92f46436004/lib/hooks/use-discussion-tts.ts#L195-L344).

### Deployment Method

Local development / deterministic hook reproduction.

### OpenMAIC Version

`29735f10d0081859ac3db1a50a0cc92f46436004` (`main`).

### Affected Area

Playback / presentation — discussion TTS.

### Browser

Chromium (Codex in-app browser) for native HTMLAudioElement verification; jsdom for deterministic hook regression tests. Live-provider listening was not performed.

### Operating System

macOS 26.5; Node.js 22.22.3 (matching CI major version); pnpm 10.28.0.

### Relevant Logs

```text
Current:  request A -> play A -> ended A -> request B -> play B
Expected: request A -> play A -> request B -> [B ready] -> ended A -> play B
```

This is the source-derived event ordering, not a captured provider timing measurement.

### Screenshots or Recordings

Test output and native-browser event timings are included in #1435. The initial three regression tests fail on the base implementation; the fixed implementation passes all 17 cases.

### Additional Context

Searched existing TTS/discussion issues and open PRs. #695 addresses pre-generated scene audio, not the discussion playback queue.

Scope: avoidable inter-segment synthesis waits. Independent requests can still have prosody differences; this fix does not establish a continuous model context or claim to eliminate differences in voice quality.

yanpgwang:
## Summary

When a discussion has already received the next text segment, its TTS request still waits for the current clip's `ended` event. Prepare one queued segment during playback so synthesis latency can overlap the current clip, while audio remains strictly ordered.

AI-assisted with Codex. Self-review completed, including cancellation races, queue ownership, bounded lookahead, and fallback completion accounting.

## Related Issues

Closes #1434

## Changes

- Separate audio generation from playback and retain at most one prepared segment. Server synthesis requests remain sequential; browser-native speech stays on the playback path.
- Consume each prepared result once, including when playback catches up before synthesis finishes. Background preparation does not replace the active speaker's audio indicator.
- Abort current and prefetched work on cleanup, disable, or unmount; ignore late responses and stale media events. Muting cancels lookahead while preserving the current clip and its position for unmute.
- Preserve pause/resume and voice fallback order. Count a missing-clone fallback as the same segment, and advance safely after playback errors or a rejected resume.
- Add 17 hook regression cases covering these transitions.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Run `pnpm exec vitest run tests/hooks/use-discussion-tts.test.ts`.
2. With server TTS enabled, start a discussion with multiple sealed text segments. Confirm B's `/api/generate/tts` request starts while A plays and B never plays before A ends.
3. Pause during generation and resume; stop during lookahead; verify that queued audio does not start while paused or after cancellation.

- The initial three regression cases failed against the base implementation: B was never requested during A playback, and a late response could play after cleanup. All 17 cases now pass.
- Exercised the actual hook with native `HTMLAudioElement` in Chromium, using a local verification page and a deterministic mock TTS endpoint (500 ms response delay, two-second silent WAV clips). Verified A/B/C order, pause during generation followed by resume, cancellation during prefetch, and mute during synthesis followed by unmute without restarting the current clip.
- Node.js 22.22.3 / pnpm 10.28.0: `pnpm check`, `pnpm lint`, `pnpm exec tsc --noEmit`, `pnpm check:i18n-keys`, package-version, dependency-range, and Node-engine checks pass. ESLint reports 18 existing warnings outside the changed files; changed files have no lint findings.
- Full root suite: **671 files passed, 9 skipped; 7,495 tests passed, 26 skipped**.
- Live-provider listening and the full classroom UI were not exercised. This fixes client scheduling; it does not provide cross-request model context or claim to correct voice/prosody differences between independently synthesized clips.

### Evidence

Observed in the native-audio browser verification (times relative to starting the sequence):

```text
   1 ms  request A
 502 ms  response A
 682 ms  playing A / request B
1184 ms  response B
3080 ms  ended A
3108 ms  playing B / request C
3609 ms  response C
5494 ms  ended B
5522 ms  playing C
```

B is ready before A ends; the A-to-B playback transition took 28 ms in this mock scenario instead of waiting for another 500 ms synthesis response. These are verification-page timings, not a live-provider benchmark.

Cancellation case:

```text
505 ms  playing A / request B
702 ms  abort B
703 ms  idle
        no subsequent B playback
```

- [x] CI-equivalent local checks pass (`pnpm check && pnpm lint && pnpm exec tsc --noEmit`)
- [x] Manually tested locally (native-audio verification page; scope described above)
- [ ] Screenshots / recordings attached (no visual UI changes)

## Checklist

- [x] I have added/updated documentation as needed (inline scheduling/lifecycle comments; no public API changes)


wyuc (approved):
Reviewed the full diff with the state machine written out (queue / head / prepared-slot / abort controllers / finish idempotence) and probed the transitions. Ordering holds: the prefetch `.then` only stores a result, the only play triggers are `processQueue` from finish / browser onEnd / resume / seal-while-idle, and `prefetchNext` requires an occupied head with no outstanding synthesis, so lookahead is exactly one and N+2 stays queued. Prepared results are consumed once (read+null in one synchronous block, `processQueue` serialised by `isPlayingRef`), `finish` is idempotent across `ended`+`error`, cleanup aborts both controllers and post-abort responses are stopped by `throwIfAborted` and the catch guard, mute keeps the current clip and unmute re-requests the cancelled lookahead exactly once, and the missing-clone fallback advances the counter once. The removal of the `generating` emission on seal is a fix (it pointed the indicator at the next speaker). 17/17, tsc, eslint, prettier clean; making prepared audio play on resolve turns 6 tests red, skipping the prefetch abort on cleanup turns 3 red. CI green.

Two non-blocking notes for a follow-up:
- The `currentItemRef.current !== item` staleness checks (`use-discussion-tts.ts` ~L325 and ~L366) are not pinned: deleting either, or both, keeps 17/17 green because the covered late-response case is already intercepted by `throwIfAborted` / `signal.aborted`. A test where the identity check is the only thing standing between a late result and playback would pin the "ignores late responses" claim.
- The browser-native `onEnd` guard (~L92) is keyed on provider + `isPlayingRef`, not on the utterance, so a duplicate or late `onend` from a previous utterance while a new one is active would end the new one early (reproduced with a probe; real-browser reachability depends on whether `speechSynthesis.cancel()` fires the old utterance's `onend`, so plausible rather than confirmed). Base had no guard at all, so this is not a regression.

PR #1183, opened 2026-08-26
fix(‎middleware): Rename middleware to proxy

puxiao:
Fixes https://github.com/THU-MAIC/OpenMAIC/issues/1177

Next warning: The "middleware" file convention is deprecated. Please use "proxy" instead. Learn more: https://nextjs.org/docs/messages/middleware-to-proxy

PR #1437, opened 2026-09-10
fix: eliminate Next.js runtime and deprecation warnings

### LINKED ISSUE #1436: Fix Next.js runtime and deprecation warnings in the console
## Problem

When running OpenMAIC locally, the console repeatedly reports Next.js runtime compatibility and deprecation warnings.

Although requests still complete successfully, these warnings create unnecessary noise, make real errors harder to identify, and may become build or runtime failures in future Next.js releases.

The following warnings have been identified:

1. `instrumentation.ts` uses `process.once`, which is unsupported in the Edge Runtime.
2. The project still uses the deprecated `middleware.ts` file convention instead of `proxy.ts`.
3. Other repetitive console output may need review if it provides no diagnostic value.

Image

### Node.js APIs in instrumentation

Next.js compiles `instrumentation.ts` for both the Node.js and Edge runtimes. A runtime check using `NEXT_RUNTIME` does not prevent the Edge compiler from statically detecting `process.once`.

Node.js-specific startup and shutdown logic should be isolated in a separate module and dynamically imported only for the Node.js runtime.

### Deprecated middleware convention

Next.js 16 has renamed the `middleware.ts` convention to `proxy.ts`. The exported `middleware` function should also be renamed to `proxy`.

## Proposed Changes

- Move Node.js-specific startup, background-task, and signal-handling logic into a dedicated module.
- Dynamically import the Node.js instrumentation module only when `NEXT_RUNTIME === "nodejs"`.
- Rename `middleware.ts` to `proxy.ts`.
- Rename the exported `middleware` function to `proxy`.
- Preserve the existing access-code authentication, workbench feature gates, and route matcher behavior.
- Remove other redundant console output where appropriate without suppressing actionable errors.

## Acceptance Criteria

- [ ] Local development no longer reports `process.once is not supported in the Edge Runtime`.
- [ ] Local development no longer reports `The "middleware" file convention is deprecated`.
- [ ] Access-code authentication behavior remains unchanged.
- [ ] Workbench feature-gating behavior remains correct.
- [ ] Unauthorized API requests still return `401`.
- [ ] Disabled workbench routes still return `404`.
- [ ] Relevant unit tests and lint checks pass.
- [ ] The production build completes without these warnings.
- [ ] Real errors are not hidden, swallowed, or downgraded merely to silence the console.

ztygod:
## Summary

Fix Next.js runtime and deprecation warnings reported during local development and production builds.

This PR isolates Node.js-specific instrumentation logic from the Edge bundle and migrates the deprecated `middleware.ts` convention to Next.js 16's `proxy.ts` convention.

## Related Issues

Closes #1436

## Changes

- Moved Node.js-specific startup and shutdown logic into `instrumentation-node.ts`.
- Kept the root `instrumentation.ts` entry compatible with both Node.js and Edge runtimes.
- Dynamically imported Node-only instrumentation when `NEXT_RUNTIME === "nodejs"`.
- Prevented the Edge compiler from analyzing unsupported `process.once` calls.
- Renamed `middleware.ts` to `proxy.ts`.
- Renamed the exported `middleware` function to `proxy`.
- Preserved the existing access-code authentication and route matcher behavior.
- Updated the workbench availability check to use the Node.js runtime available to Next.js 16 Proxy.

## Type of Change

- [ ] Documentation update
- [x] CI/CD or build changes

### Steps to reproduce / test

1. Start the application with `pnpm dev`, or run a production build with `pnpm build`.
2. Access the home page, API routes, and the Pro workbench route.
3. Confirm that the following warnings no longer appear:
   - `A Node.js API is used (process.once) which is not supported in the Edge Runtime`
   - `The "middleware" file convention is deprecated. Please use "proxy" instead`
4. If `ACCESS_CODE` is configured, verify that protected API requests without a valid cookie still return `401`.
5. Verify that `/workbench` returns `404` when either the public workbench flag or server-side agent runtime is disabled.

- Ran the existing instrumentation and asset collector test suite:
  - `./node_modules/.bin/vitest run tests/persistence/asset-collector-schedule.test.ts`
  - Result: 11 tests passed.
- Ran ESLint against the changed files:
  - `./node_modules/.bin/eslint proxy.ts instrumentation.ts instrumentation-node.ts`
- Ran `git diff --check`.
- Confirmed that the deprecated middleware warning is no longer emitted when Next.js starts a production build.
- A complete local production build was not verified because another `next build` process was still active; the full build remains covered by CI.

### Evidence

| Before | After |
| --- | --- |
| Image | image


puxiao:
middleware: https://github.com/THU-MAIC/OpenMAIC/pull/1183
warnings: https://github.com/THU-MAIC/OpenMAIC/pull/1264
They said these aren't needed.

LING-6150:
Thanks for putting this together. The warnings are real, and I checked the current commit: the instrumentation extraction appears mechanically faithful, and I found no code-level regression on the PR’s original base.
However, this PR reopens two same-scope changes that were already reviewed and closed in #1183/#1177 and #1264/#1262. The current maintenance decision has not changed: the process.once messages are non-blocking warnings, while the middleware.ts → proxy.ts migration changes the gateway runtime and Workbench availability semantics. That migration should happen only as a deliberate Next.js cutover with an explicit runtime decision and gateway regression coverage, rather than as warning cleanup.
This PR provides no new blocking failure or updated product decision, adds no Workbench/access-code gateway tests, and now conflicts with the substantially changed main instrumentation and middleware paths. I’m therefore closing it rather than asking for implementation changes.
We can revisit this through a new scoped issue if a future Next.js version makes either warning blocking, or if the project decides to perform the proxy migration with the required runtime contract and regression coverage.

PR #1441, opened 2026-09-10
fix: filter out bookmarked courses from @ mention

qianqiang-del:
## Problem Description

  The original @ feature displays all courses, including:
  - Courses created by the user (isOwner: true)
  - Courses bookmarked from other users (isOwner: false)

  ## Original Drawbacks

  1. Poor User Experience:
     → Bookmarked courses cannot be modified
     → Selecting them leads to confusion as they cannot be edited

  2. Inconsistent Functionality:
     → Displays non-operable options
     → Violates the "only show operable content" design principle

  3. Low Efficiency:
     → Users need to manually identify which courses can be modified
     → Increases cognitive burden

  ## Fix

  Added filter condition in courseOptions calculation:
  .filter((course) => course.isOwner !== false)

  Filters out courses where isOwner is false,
  only displays courses created by the user.

  ## User Benefits

  1. Clearer:
     → Only shows modifiable courses
     → Reduces user confusion

  2. More Efficient:
     → Users don't need to identify course types
     → Can directly select and operate

  3. More Consistent:
     → All displayed courses can be modified
     → Meets user expectations

  ## Modified File

  components/workbench/workspace/WorkspaceShell.tsx

### DIFF
--- a/components/workbench/workspace/WorkspaceShell.tsx
+++ b/components/workbench/workspace/WorkspaceShell.tsx
@@ -952,6 +952,7 @@ function WorkspaceShellController({ initialPanes }: { readonly initialPanes: Wor
   const courseOptions = useMemo(
     () =>
       [...courses.classrooms]
+        .filter((course) => course.isOwner !== false)
         .sort((a, b) => b.updatedAt - a.updatedAt)
         .map((course) => ({ id: course.id, name: course.name })),
     [courses.classrooms],


CXuPercy:
Thanks for the report — the UX observation is fair, but this change reverses an intentional design, so I'm closing it rather than iterating here.

What `@` actually does with a course: it is a reference, not an editing action. `lib/workbench/run-courses.ts:33-46` documents that course refs are "not a permission, not prompt input" — display and navigation only, with the owner-bound store refusing anything the user may not touch. `lib/server/agent-runtime/runner.ts:694-702` resolves each mentioned course against the owner's own library, and a classroom that doesn't resolve (missing, **foreign**, or tombstoned) deliberately degrades to the snapshot title — "the user named it, so the model still learns which one." Bookmarking someone's course and `@`-ing it ("compare my course with @X") is exactly the case that path was written for; the agent never gets access it shouldn't have.

You're right that the value is thin — the course tools are owner-scoped, so a foreign mention carries the course's name but not its content. But the answer to a thin-by-design capability is to make that visible (e.g. marking bookmarked courses as reference-only in the picker), not to remove the option.

Also per repo convention this is a behavior change without a linked issue — if the confusion you describe is something you've actually observed, please open an issue describing it, and we can take the reference-only annotation direction there. Thanks again for digging into the composer.

Closing for now.

PR #1407, opened 2026-09-07
fix: include sharp libvips in standalone tracing

itegel:
## Summary

Fix sharp/libvips load failure in standalone Docker builds. Two sharp versions resolve in the tree (0.34.5 transitive → libvips 1.2.4, 0.35.4 direct → libvips 1.3.3); Next.js standalone tracing omitted the dlopen-loaded libvips and kept the wrong version, so sharp 0.35.4 failed to load at boot (`ERR_DLOPEN_FAILED: libvips-cpp.so.8.18.6` not found). This broke agent runtime startup and PDF/image material extraction on every self-hosted Docker deployment.

## Related Issues

Closes #1406

## Changes

- `next.config.ts`: add `node_modules/.pnpm/@img+sharp-libvips-*/node_modules/@img/sharp-libvips-*/lib/**` to `outputFileTracingIncludes`, force-including all sharp-libvips native lib dirs so the version matching the loaded sharp binary is always present in the standalone output.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

### Steps to reproduce / test

1. Build the Docker image on v1.0.1 **before** the fix → boot logs show `[instrumentation] Agent runtime startup failed ... ERR_DLOPEN_FAILED: libvips-cpp.so.8.18.6: No such file`
2. Apply the fix and rebuild: `docker compose --profile server-persistence up --build -d`
3. Check boot logs and the libvips files inside the container

### What you personally verified

- **Before fix**: agent runtime startup failed; `find /app/node_modules/.pnpm -name "libvips-cpp.so*"` showed only `libvips-cpp.so.8.17.3` (1.2.4) — 8.18.6 missing.
- **After fix**: `[AgentRunner] runner 1867754d:1 started (scan=1000ms, heartbeat=2000ms, leaseTtl=10000ms, maxConcurrent=2, maxAttempts=5)` — no sharp error; `find` now also returns `@img+sharp-libvips-linuxmusl-arm64@1.3.3/.../libvips-cpp.so.8.18.6`; `/api/health` 200, `/api/stages` 200.
- The glob traced all libvips versions (1.2.4 + 1.3.3, musl + glibc variants), so the correct one is present regardless of platform or sharp version.
- **NOT verified**: end-to-end PDF upload through the UI. sharp loading and agent-runner startup were the actual failure points; PDF extraction (`lib/pdf/pdf-providers.ts`) imports sharp, which now resolves.

### Evidence

[...]

PR #1442, opened 2026-09-10
fix(docker): create /app/data with runtime-user ownership so classroom persistence works

### LINKED ISSUE #1438: [Bug]: Docker: named volume /app/data is created root-owned — classroom persistence silently fails with EACCES
### Bug Description

On self-hosted Docker deployments, the app's persistent data directory `/app/data` ends up owned by `root` while the server runs as the unprivileged `nextjs` user (uid 1001). Every write under `/app/data` fails with `EACCES`, which silently breaks classroom persistence (courses never reach disk) and usage recording. The app keeps serving normally, so the failure is only noticed much later — e.g. after an upgrade, when the volume turns out to be empty.

The cause is a gap between the Dockerfile and the compose file:

- `docker-compose.yml` mounts the named volume `openmaic-data` at `/app/data`
- The Dockerfile never creates `/app/data` (no `mkdir`/`chown` anywhere in the file; it only does `adduser nextjs` + `USER nextjs`)
- When Docker first mounts a named volume at a path that does not exist in the image, it creates the mountpoint as `root:root` (0755)
- The runtime user `nextjs` (uid 1001) therefore has no write permission on `/app/data`, and `lib/server/classroom-storage.ts` resolves `CLASSROOMS_DIR = path.join(process.cwd(), 'data', 'classrooms')` to `/app/data/classrooms` inside that directory

Observed on the running container before any fix:

```
/app/data  ->  drwxr-xr-x 2 root root
```

### Steps to Reproduce

1. Deploy v1.0.1 with the upstream `Dockerfile` + `docker-compose.yml` (no postgres profile needed)
2. Let compose create the `openmaic-data` volume (i.e. start from a clean volume)
3. Create a classroom in the UI and save it
4. Inspect the volume:

```sh
docker exec openmaic-openmaic-1 ls -ld /app/data
docker exec openmaic-openmaic-1 find /app/data -mindepth 1 | wc -l   # 0
docker logs openmaic-openmaic-1 | grep -i EACCES
```

### Expected Behavior

Classrooms, classroom media, and usage records persist under `/app/data` across container recreation; the directory is writable by the runtime user out of the box.

### Actual Behavior

Every write under `/app/data` fails with `EACCES`. Classroom saves via `POST /api/classroom` fail; usage recording logs a warning and is skipped (`Failed to record usage (ignored)`); the volume stays permanently empty. Nothing in the UI indicates persistence is broken.

### Deployment Method

Docker Compose (upstream `Dockerfile` + `docker-compose.yml`, default service only).

### OpenMAIC Version

v1.0.1 (also reproducible on 0.2.x — the Dockerfile has never created `/app/data`; I have an empty `openmaic-data` volume dating back to a June 0.2.x deployment as evidence the failure is long-standing)

### Affected Area

Server-side persistence: `lib/server/classroom-storage.ts` (`CLASSROOMS_DIR`, `CLASSROOM_JOBS_DIR`), classroom media under `/app/data/classrooms/<id>/`, usage storage. Platform-independent (any host OS/architecture; it is the image, not the host).

### Browser

Not browser-specific.

### Operating System

Windows 10+ (Docker Desktop, WSL2 backend); root cause is host-independent.

### Relevant Logs

```
[2026-09-09T17:19:26.779Z] [WARN] [UsageStorage] Failed to record usage (ignored): Error: EACCES: permission denied, mkdir '/app/data/usage'
```

### Root Cause

The Dockerfile declares `USER nextjs` but never pre-creates `/app/data`. Docker materializes a named volume's mountpoint as `root:root` when the path is absent from the image, so the unprivileged runtime user cannot write to the very directory the compose file mounts for persistence.

### Proposed Fix

Create the data directory with correct ownership in the final stage, before dropping privileges:

```dockerfile
RUN mkdir -p /app/data && chown -R nextjs:nodejs /app/data

USER nextjs
```

This is durable for fresh deployments because Docker's empty-volume copy-up inherits the image directory's ownership into the new volume.

Note for existing deployments: the one-line fix only affects newly created volumes. Already-deployed volumes need a one-time repair, e.g.:

```sh
docker exec -u root <container> chown -R 1001:1001 /app/data
```

(Side note: while verifying this I also hit #1406/#1407 — the missing sharp libvips in standalone output. That one is already covered by #1407; this issue is a separate, independent defect.)

mianbaofang:
## Summary

Fixes #1438. The final stage never creates `/app/data`, while `docker-compose.yml` mounts the `openmaic-data` named volume there. On a fresh volume Docker materializes the mountpoint as `root:root`, so the unprivileged runtime user (`nextjs`, uid 1001) cannot write to the directory the app persists classrooms, classroom media, and usage records into. The build now creates the directory owned by the runtime user before `USER nextjs`.

## Root cause

Verified against the repo:

- `Dockerfile` final stage has only `adduser` + the two `COPY --chown` of the standalone output + `USER nextjs` — no `mkdir`/`chown` for `/app/data`;
- `docker-compose.yml:40` mounts `openmaic-data:/app/data`;
- `lib/server/classroom-storage.ts` resolves `CLASSROOMS_DIR` to `/app/data/classrooms`, so every classroom save lands inside that directory.

Docker creates a named volume's mountpoint as `root:root` when the path is absent from the image; the runtime user then has no write permission. The write failure is non-fatal by design, so the app keeps serving while persisting nothing:

```
[WARN] [UsageStorage] Failed to record usage (ignored): Error: EACCES: permission denied, mkdir '/app/data/usage'
```

## Changes

- `Dockerfile`: `RUN mkdir -p /app/data && chown -R nextjs:nodejs /app/data` before `USER nextjs`. Empty-volume copy-up inherits this ownership, so fresh deployments work out of the box — no compose change needed.
- `packages/docs/content/docs/deployment.mdx`: a note plus the one-time repair for volumes already created by an older image (the fix cannot retroactively change an existing volume).

## Verification

Built from this branch and ran end-to-end on Docker (Windows host, Docker Desktop / WSL2, `node:22-alpine` / musl):

1. Image contains the directory with the runtime user's ownership, so an empty volume inherits it:

```
$ docker run --rm --entrypoint ls <image> -ld /app/data
drwxr-xr-x    2 nextjs   nodejs        4096 //app/data
```

2. After deploying with a fresh volume, `docker exec <container> ls -ld /app/data` reports `nextjs nextjs` and writes succeed; the `EACCES` log line is gone from a clean boot.
3. The app boots and serves normally (`GET /` -> 200, `GET /api/agent/skills` -> 200) and a repaired pre-existing volume accepts writes after the documented one-time `chown`.

The `EACCES` line before the fix was reproduced on an unchanged v1.0.1 tree and is attached to #1438, including the empty volume dating back to a 0.2.x deployment.

## Notes

- Scope is intentionally minimal; the broader "persistence failure is invisible in the UI" hardening (e.g. a `/api/health` probe) is a separate concern, as noted on #1438.
- No compose or config changes are required for new deployments; only pre-existing volumes need the documented one-time repair.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -105,6 +105,15 @@ COPY --from=builder /app/public ./public
 COPY --from=builder --chown=nextjs:nodejs /app/.next/standalone ./
 COPY --from=builder --chown=nextjs:nodejs /app/.next/static ./.next/static
 
+# The app persists classrooms, classroom media, and usage records under
+# ./data, which docker-compose.yml mounts as the openmaic-data named volume.
+# Nothing above creates the directory, so on first run Docker materializes
+# the mountpoint as root:root and every write from the unprivileged runtime
+# user fails with EACCES — classroom persistence silently stores nothing
+# (THU-MAIC/OpenMAIC#1438). Creating it here makes the empty-volume copy-up
+# inherit the runtime user's ownership.
+RUN mkdir -p /app/data && chown -R nextjs:nodejs /app/data
+
 USER nextjs
 
 EXPOSE 3000
--- a/packages/docs/content/docs/deployment.mdx
+++ b/packages/docs/content/docs/deployment.mdx
@@ -33,6 +33,12 @@ docker compose up --build
 
 The default Compose deployment starts the OpenMAIC app and mounts the `openmaic-data` volume. Configure other providers and features as needed; see [Configuration](./configuration.mdx).
 
+The image creates the volume's data directory with the runtime user's ownership, so fresh volumes persist classrooms and usage records out of the box. A volume created by an older image may be root-owned, which makes every server-side save fail with `EACCES` while the app keeps serving normally (classrooms silently never reach disk; the log shows `Failed to record usage ... EACCES: permission denied`). Repair such a volume once, then restart the app:
+
+```bash
+docker exec -u root <container> chown -R 1001:1001 /app/data
+```
+
 `NEXT_PUBLIC_*` feature flags are injected at Docker build time and cannot be set only in the runtime `.env.local`. For example, to enable video export and experimental PPTX import:
 
 ```bash


CXuPercy (approved):
Cross-reviewed by two independent reviewers, including an empirical Docker probe that reproduces #1438 on a root-materialized mountpoint and confirms the fix makes fresh volumes inherit 1001:1001 ownership. Exactly the fix the issue called for, both halves: the Dockerfile pre-creates `/app/data` with runtime-user ownership before `USER nextjs` (matching the existing `COPY --chown=nextjs:nodejs` convention; `.dockerignore` excludes `data` so nothing is shadowed), and the deployment docs document the one-time `chown` repair for volumes created by older images — uid 1001 consistent throughout, all runtime write paths (classrooms, usage, materials, media) resolve under `/app/data`, and it merges cleanly with main's post-base changes. Thanks also for the end-to-end evidence on #1407 earlier — this and that together make self-hosting v1.0.1 actually work out of the box.

wyuc (approved):
Approving. Creating `/app/data` owned by the runtime user before `USER nextjs` makes a fresh `openmaic-data` volume writable, so classroom saves and usage records stop failing with EACCES, and the deployment docs give a one-time repair for volumes created by older images. Thanks for the clear root-cause write-up!

PR #1443, opened 2026-09-10
fix(settings): maintain asr language state invariant on provider selection (#1082)

### LINKED ISSUE #1082: openai-whisper ASR fails with "unsupported language" when the language setting carries a region subtag

Cham1229:
## What & Why
Fixes #1082. Addresses maintainer feedback on #1083.

Previously, `setASRProvider` enforced that `asrLanguage` is valid for the selected provider (or reset to the provider's default supported language), but automatic selection and rehydration paths bypassed this invariant:
- Server configuration sync (`syncServerConfig` -> `autoAsrProvider` & `validASRProvider`) could switch `asrProviderId` (e.g. from `browser-native` with `zh-CN` or `pt-BR` to `openai-whisper`) without validating `asrLanguage`.
- Store rehydration (`ensureValidProviderSelections`) did not guarantee language validity for the active provider.
- Adding and removing custom ASR providers left stale or unsupported language codes.

This change introduces `getValidASRLanguage(providerId, currentLanguage)` and enforces the state invariant across:
1. Automatic server provider selection (`autoAsrProvider` and `validASRProvider` in `syncServerConfig`).
2. Server sync when provider ID is unchanged but language is stale.
3. Store rehydration (`ensureValidProviderSelections`).
4. Custom ASR provider additions (`addCustomASRProvider`) and removals (`removeCustomASRProvider`).
5. Manual provider switching (`setASRProvider`).

## Verification
- Added comprehensive regression test cases in `tests/store/settings-server-sync.test.ts` covering server sync, fallback, rehydration, prototype pollution defense, and custom provider transitions.
- Ran all audio tests: `pnpm test tests/audio/` (235/235 passed).
- Ran all store tests: `pnpm test tests/store/` (333/333 passed).
- Code formatting & lint: `pnpm eslint` (0 errors) and `npx prettier --check` (clean).


LING-6150 (changes requested):
Thanks for addressing the provider/language invariant at the settings boundary. I independently reproduced the original automatic-selection mismatch on base and confirmed that this PR fixes it.
One regression remains: getValidASRLanguage treats custom ASR providers as supporting only auto, while MediaPopover offers CUSTOM_ASR_DEFAULT_LANGUAGES, including pt, en, and zh. With a configured custom provider and pt selected, both persistence rehydration and server-provider synchronization reset the language to auto on this head. Both paths preserve pt on base.
Please reuse CUSTOM_ASR_DEFAULT_LANGUAGES for custom-provider validation and add regressions covering a supported language surviving rehydration and synchronization with the same provider selected.

Cham1229:
Thanks for catching this! Updated `getValidASRLanguage` to reuse `CUSTOM_ASR_DEFAULT_LANGUAGES` for custom ASR providers and added regression tests ensuring supported languages (`pt`, `en`, `zh`, etc.) survive both store rehydration and server synchronization.
And it's ready for review~

LING-6150:
@wyuc This follows the provider/language invariant approach you suggested in #1083. The previous custom-language preservation blocker is resolved, and I verified the browser selection/reload flows.
The implementation also applies validation during rehydration and same-provider synchronization. I have no remaining technical blockers; could you confirm that these recovery paths are within the intended scope before I approve?

custom-pt-after-reload

wyuc:
@LING-6150 Yes, those recovery paths are in scope. The point of the provider/language invariant from #1083 is that the (provider, language) pair must be valid at every boundary where it can enter the store, not only on an explicit user selection, so validating during rehydration and same-provider synchronization is exactly the intended shape. Please go ahead and approve; I will merge once CI is green on the updated branch.

LING-6150 (approved):
confirmed, no blocking issue from me, approving.  cc @wyuc

wyuc (approved):
Approving with @LING-6150's review: the (provider, language) invariant is now enforced at selection, rehydration and same-provider synchronization, the custom-provider language regression is covered, and CI is green on the updated branch. Thanks for following this through, @Cham1229.

PR #1444, opened 2026-09-10
fix(media): emit keyframe images as data URLs [AI-assisted]

### LINKED ISSUE #1220: [Bug]: 生成失败 Failed to store image bundle at image img_1
### Bug Description

网页版，没有选深度交互，点开始后出现如下错误：
生成失败
Failed to store image bundle at image img_1

### Steps to Reproduce

生成失败
Failed to store image bundle at image img_1

### Expected Behavior

可以正常生成ppt

### Actual Behavior

课程无法正常生成

### Deployment Method

Other

### OpenMAIC Version

v0.3.2（网页在线版）

### Affected Area

Classroom generation

### Browser

Version 152.0.7977.64 (Official Build) (64-bit)

### Operating System

Windows 11

### Relevant Logs

```shell

```

### Screenshots or Recordings

Image

### Additional Context

_No response_

dajiaohuang:
## Summary

> **Maintainer note (scope):** on current `main`, local keyframe assets are consumed by the agent-runtime material-extraction path only; the browser document-bundle path (`storeImages`) does not receive them, since the media branch of `extract-document` returns no images. This PR therefore aligns `assets[].data` on one data-URL contract across producers and consumers (material extraction now decodes both forms); it does not by itself fix the browser report in #1220, hence `Refs`. A follow-up maintainer commit also rejects malformed or empty data URLs instead of decoding garbage bytes.

Local `ffmpeg` keyframe extraction writes the asset's `data` field as **raw base64** instead of a **base64 data URL**. Every downstream consumer of the document bundle treats `assets[].data` as a data URL, so any course generated from an uploaded video that went through local keyframe extraction fails at image persistence with `Failed to store image bundle at image img_1`.

This PR makes `lib/document/extractors/local-media.ts` emit the same `data:<mime>;base64,<payload>` form that every other producer in `lib/document/` already emits, and adds a regression assertion to the existing real-pipeline keyframe test.

## Related Issues

Refs #1220

## Changes

- `lib/document/extractors/local-media.ts` — keyframe asset `data` is now built as `` `data:${prepared.mime};base64,${...}` `` instead of bare `prepared.buffer.toString('base64')`. One line; no other behavior changes.
- `tests/document/local-media-extractor.test.ts` — the existing ffmpeg-backed pipeline test now asserts the keyframe `data` field matches `/^data:image\/webp;base64,[A-Za-z0-9+/]+=*$/`.

## Root cause

`local-media.ts` was the **only** raw-base64 producer under `lib/document/`. The other three follow the data-URL contract:

| Producer | Emitted form |
| --- | --- |
| `lib/document/pdf-providers.ts` | `data:<mime>;base64,...` |
| `lib/pdf/mineru-cloud.ts` | `data:<mime>;base64,...` |
| `lib/pdf/mineru-parser.ts:21` | `value.startsWith('data:') ? value : \`data:image/png;base64,${value}\`` |
| **`lib/document/extractors/local-media.ts`** | **raw base64 — fixed here** |

The value flows, unchecked, into the image persistence layer:

1. `lib/document/pdf-compat.ts:193` forwards it verbatim: `src: asset.data as string`.
2. `app/generation-preview/page.tsx:420` calls `storeImages(bundle.images)`.
3. `lib/utils/image-storage.ts:74` → `decodeBase64DataUrl(img.src)` does `base64DataUrl.split(',')` and takes `parts[1]`. A raw-base64 string contains no comma, so `base64Data` is `undefined`, and `atob(undefined)` throws `InvalidCharacterError: The string to be decoded is not correctly encoded.` — which is rethrown as `Failed to store image bundle at image img_1`, exactly the message in #1220.

There is a **second, silent** defect from the same root cause: `lib/document/pdf-compat.ts:108,118` derives the asset mime via `dataUrlMimeType(image.src)`, which returns `undefined` for a non-data-URL. So even where persistence did not throw, the declared `image/webp` mime was being dropped on the floor.

### Steps to reproduce / test

1. `pnpm install --frozen-lockfile`
2. `npx vitest run tests/document/local-media-extractor.test.ts` (an `ffmpeg`/`ffprobe` on `PATH` is required for the real-pipeline suite to run rather than skip)
3. Before the fix: the keyframe `data` assertion fails with a raw-base64 value. After: it passes with `data:image/webp;base64,...`.

- Reproduced the reported failure mechanism directly against `storeImages`. Feeding it `{ id: 'img_1', src: '<raw base64>' }` rejects with `Failed to store image bundle at image img_1` and cause `InvalidCharacterError: The string to be decoded is not correctly encoded.` — byte-for-byte the error in #1220. Feeding the same bytes as a data URL stores successfully (`session_9riZGIzO5k_img_1`).
- Confirmed the fix produces the correct wire form end-to-end through the **real** ffmpeg pipeline: extraction of a generated fixture emits `data:image/webp;base64,UklGRrYGAABXRUJQVlA4IKoGAAAwxQCdASoABdAC...`.
- Confirmed the new assertion is **not vacuous**: with the one-line source change stashed and everything else identical, the test fails on the raw-base64 value; with it restored, the test passes.
- Re-checked the sibling contract consumer `extractOriginalImageId` (`lib/utils/image-storage.ts:133`) while tracing this; its prefix arithmetic (`'session_'.length` + `SESSION_ID_LENGTH` = index 18) is correct, including for nanoids containing underscores. Not changed.

**Not verified:** I did not drive the full browser course-generation flow (upload video → generate) in a real browser session, so I have not observed the end-user-visible success path myself — only the storage-layer contract either side of it.

### Evidence

Commands run locally on Windows, all green, against the parent commit `29735f10d0081859ac3db1a50a0cc92f46436004`:

  - `pnpm check` (prettier) → `All matched files use Prettier code style!`
  - `eslint` → no findings
  - `npx tsc --noEmit` → exit 0
- [x] Manually tested locally
- [ ] Screenshots / recordings attached (if UI changes) — no UI change

Additional repo gates run locally:

- `node scripts/check-package-version-bumps.mjs 29735f10d0081859ac3db1a50a0cc92f46436004` → `Package version check passed.` (no publishable package touched, so no semver bump is required)
- `node scripts/check-i18n-keys.mjs` → `i18n key alignment check passed (12 locale files, source: en-US.json)` (no user-facing string added)
- `npx vitest run` over the focused set (local-media extractor, image-storage, pdf-compat, document bundle, media registry) → 33 passed, 1 skipped

Note on the one skip: `resolveExecutable()` in `local-media.ts` searches `PATH` for an **extensionless** `ffmpeg`, so on Windows (where the binary is `ffmpeg.exe`) the real-pipeline suite self-skips. CI does not install ffmpeg, so this suite also skips there; the CI-run guard for this change is the material-extraction decoder test. To verify the regression assertion on this machine I injected the real `ffmpeg.exe`/`ffprobe.exe` through the provider's `commands` seam.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

## Checklist

- [x] I have added/updated documentation as needed (none required — no public API or user-facing string change)

---

### AI-assisted disclosure

This PR is **AI-assisted** (Claude Code), as required by the AI-Assisted PRs section of `CONTRIBUTING.md`.

Per that policy I ran an AI self-review over the diff before opening this PR. Findings and how they were resolved:

1. *Is one line enough, or is this papering over a laxer contract?* Checked whether the fix belonged at the consumer instead. Rejected: the four producers under `lib/document/` already establish the data-URL contract, three of them explicitly, so the extractor was the outlier rather than the consumer being over-strict. Fixing the producer also repairs the silently-dropped mime in `pdf-compat`, which a consumer-side fix would not.
2. *Could the value already be a data URL in some path, making this a double prefix?* `prepared.buffer` comes from a `sharp` encode of an extracted frame — it is always raw image bytes, never pre-prefixed. Verified there is no other writer of this field in the extractor.
3. *Is the regression test actually exercising the fix, or does it pass either way?* Verified by stash/restore as described above.
4. *Does the new regex over-constrain the webp output?* It asserts only the `data:` prefix, the mime, the base64 marker, and a valid base64 alphabet; it does not pin encoder output length or content, so it will not break on a libwebp version bump.

I understand and take responsibility for the change.

### DIFF
--- a/lib/document/extractors/local-media.ts
+++ b/lib/document/extractors/local-media.ts
@@ -747,7 +747,7 @@ export async function extractMediaMaterial(
             id: assetId,
             type: 'image',
             mimeType: prepared.mime,
-            data: prepared.buffer.toString('base64'),
+            data: `data:${prepared.mime};base64,${prepared.buffer.toString('base64')}`,
             width: prepared.width,
             height: prepared.height,
             description: `${stem} at ${(timeMs / 1000).toFixed(3)} seconds`,
--- a/lib/server/material-extraction/extract.ts
+++ b/lib/server/material-extraction/extract.ts
@@ -56,6 +56,11 @@ function markerTime(timeMs: number): string {
   return `${String(hours).padStart(2, '0')}:${String(minutes).padStart(2, '0')}:${seconds}`;
 }
 
+function decodeMediaAssetData(data: string): Buffer {
+  const dataUrl = /^data:[^,]*;base64,([\s\S]*)$/i.exec(data);
+  return Buffer.from(dataUrl?.[1] ?? data, 'base64');
+}
+
 export function mediaArtifactText(artifact: MediaArtifact): string {
   return (artifact.transcript ?? [])
     .filter((segment) => segment.text.trim())
@@ -151,7 +156,7 @@ export async function extractClaimedSessionMaterial(
     const images = [];
     for (const asset of artifact.assets ?? []) {
       if (asset.type !== 'image' || !asset.data) continue;
-      const bytes = Buffer.from(asset.data, 'base64');
+      const bytes = decodeMediaAssetData(asset.data);
       const rawAssetId = await putBytes(source.sessionId, bytes, asset.mimeType ?? 'image/webp');
       images.push({
         id: createMaterialId(),
--- a/tests/agent-runtime/material-extraction.test.ts
+++ b/tests/agent-runtime/material-extraction.test.ts
@@ -184,7 +184,7 @@ describe('uploaded material extraction lifecycle', () => {
               id: 'frame-1',
               type: 'image',
               mimeType: 'image/webp',
-              data: Buffer.from('prepared-webp').toString('base64'),
+              data: `data:image/webp;base64,${Buffer.from('prepared-webp').toString('base64')}`,
             },
           ],
         })),
--- a/tests/document/local-media-extractor.test.ts
+++ b/tests/document/local-media-extractor.test.ts
@@ -142,7 +142,13 @@ describe.skipIf(!ffmpegAvailable)('local media extractor real pipeline', () => {
       ]);
       expect(artifact.transcript?.[0].endMs).toBeGreaterThan(1_500);
       expect(artifact.keyframes?.length).toBeGreaterThan(0);
-      expect(artifact.assets?.[0]).toMatchObject({ type: 'image', mimeType: 'image/webp' });
+      const keyframe = artifact.assets?.[0];
+      expect(keyframe).toMatchObject({ type: 'image', mimeType: 'image/webp' });
+      // Keyframe `data` must be a data URL. `pdf-compat` derives the asset mime
+      // from it, and the document bundle forwards it as the `pdfImages[].src`
+      // that `storeImages` decodes with `decodeBase64DataUrl`. Raw base64 here
+      // makes course generation fail with "Failed to store image bundle".
+      expect(keyframe?.data).toMatch(/^data:image\/webp;base64,[A-Za-z0-9+/]+=*$/);
 
       const deadlineProvider = createLocalMediaExtractorProvider({
         transcribe: vi.fn(() => new Promise<{ text: string }>(() => undefined)),


CXuPercy (dismissed):
Thanks for the report and the clear diagnosis — the bug is real (#1220, failure mode exactly as described), but this fix breaks the other consumer of the same field, so it needs another pass:

**1. (blocking) `material-extraction` consumes the same assets as raw base64.** The artifact this producer emits is also consumed by the agent-runtime extraction path: `lib/server/material-extraction/extract.ts:127` runs the same `extractMediaMaterial` output (via the provider registered at `lib/document/extractors/local-media.ts:830`), and at `extract.ts:152` it decodes every `asset.data` with `Buffer.from(asset.data, 'base64')`. Node's base64 decoder ignores non-alphabet characters, so a `data:image/webp;base64,` prefix decodes to garbage bytes in front of the real payload — every image material produced from an uploaded video in the agent runtime would become corrupt. The description's "every downstream consumer of the document bundle treats `assets[].data` as a data URL" doesn't hold: the document-bundle path does (`pdf-compat.ts:148` → `image-storage.ts:74 decodeBase64DataUrl`), the material-extraction path does not. Please make `extract.ts` accept the data-URL form (parse/strip the prefix before decoding) in the same PR, so the field has one contract across both consumers.

**2. The regression guard won't run in CI.** The new assertion sits in `describe.skipIf(!ffmpegAvailable)` and no workflow installs ffmpeg, so it protects the fix only on machines that happen to have ffmpeg. Either add ffmpeg to CI or note the limitation explicitly.

Checks run: static trace of both consumption paths on the PR head; the corrupted-decode behavior follows from Node's base64 decoder semantics (non-alphabet characters ignored).

Happy to re-review once 1 is addressed — with both consumers aligned this is the right fix for #1220.

dajiaohuang:
Updated in 0686b75: material-extraction now strips only a data:*;base64, prefix before decoding, while legacy raw base64 remains accepted. The lifecycle regression fixture now uses the data-URL form. Validation: 6 material-extraction tests passed, TypeScript noEmit passed, Prettier passed, and git diff --check passed. The ffmpeg-gated local-media regression remains skipped in CI because the workflow does not install ffmpeg; I did not add a CI dependency.

wyuc:
Thanks @dajiaohuang for the careful work here, and for the quick follow-up on @CXuPercy's review. I had this independently verified (real ffmpeg keyframes → `storeImages`, plus the full consumer list), and there's one correction plus a couple of test gaps before this can land.

**This doesn't fix #1220.** The browser path in #1220 never sees local-media keyframes: the media branch of `/api/extract-document` returns `images: []` and doesn't forward `MediaArtifact.assets`, so they never reach `storeImages`. The only reader of this `data` field is `lib/server/material-extraction/extract.ts`, which decoded correctly before and after this change. `local-media.ts` also only landed in 1.0.0, while #1220 was reported on v0.3.2. So please change `Fixes #1220` to `Refs #1220` and describe the PR as aligning local-media with the data-URL contract every other producer follows — that's still worth having, so a future consumer of keyframes doesn't hit the same mismatch.

To merge it on that basis, please:

1. **Cover the legacy raw-base64 path.** The second commit switched the only `material-extraction` fixture to the data-URL form, so the "legacy raw base64 remains accepted" branch has no test. A case asserting that both forms decode to identical bytes would pin it.
2. **Make the regression check run in CI.** The new assertion sits inside `describe.skipIf(!ffmpegAvailable)`, and the CI runner doesn't install ffmpeg, so it is always skipped there. Testing how the keyframe asset's `data` is constructed without invoking ffmpeg (e.g. extracting the helper) would keep it covered.
3. *(minor)* `decodeMediaAssetData` falls back to base64-decoding the whole string for a non-base64 data URL (e.g. `data:text/plain;charset=utf-8,…`), producing garbage bytes. Only images reach it today, but throwing on an unexpected data URL would be safer than a silent wrong decode.

dajiaohuang:
Addressed the review feedback in `1ac5f38`:

- changed the issue reference from `Fixes #1220` to `Refs #1220`, since this PR aligns the local-media data contract but does not fix the older browser-path report;
- exported and tested the shared media decoder against both legacy raw base64 and `data:*;base64,` assets, plus an explicit failure for non-base64 data URLs;
- kept the existing `material-extraction` lifecycle coverage and verified the focused suite (7 tests), Prettier, ESLint, TypeScript, and `git diff --check`.

The ffmpeg-backed extractor assertion remains covered only where ffmpeg is available; this change adds the non-ffmpeg contract test so CI exercises the decoding behavior directly.

wyuc (approved):
Thanks @dajiaohuang — the blocking item from the earlier review is resolved, verified independently:

- **One contract across both consumers**: material extraction now decodes both the data-URL and legacy raw-base64 forms, so the prefix no longer corrupts image bytes. Reverting its call site to a naive `Buffer.from(data, 'base64')` turns the lifecycle test red, and that test runs in CI without ffmpeg.
- Every other `DocumentAsset` producer already emits data URLs, so local keyframes are now consistent with them.

I pushed one small follow-up commit (591cc8b0) on top: a string that starts with `data:` but doesn't parse as a data URL now throws instead of falling through to raw-base64 decoding (Node's decoder skips non-alphabet characters, so that path silently produced garbage bytes), and an empty payload is rejected rather than stored as a 0-byte image. Covered by a new test that fails with the guard removed. I also corrected the keyframe test comment and added a scope note to the description: on current `main` local keyframes only feed material extraction, so this is a contract alignment and `Refs #1220` is the right link.

Approving — I'll update the branch and merge once CI is green.

PR #1453, opened 2026-09-10
test(agent-runtime): make the material search budget test deterministic

### LINKED ISSUE #1452: Flaky: material-tools search test races the 100 ms time budget against the character budget
## Symptom

`tests/agent-runtime/material-tools.test.ts` › `stops at the per-execution character budget and reports truncation` fails intermittently in the required `Lint, Typecheck & Unit Tests` job:

```
expected { scannedChars: 786432, truncated: true, ... } to match { scannedChars: 1000000, truncated: true, ... }
```

Seen at least twice on unrelated PR heads: 2026-09-05 on #1328 (author reported the same `786432` value) and 2026-09-10 on #1420 (run 34531005106; a rerun of the same commit passed). Neither PR touches the agent runtime.

## Cause

`search_material` in `lib/server/agent-runtime/material-tools.ts` stops on whichever of two budgets is hit first: `MAX_SEARCH_CHARS_PER_EXEC` (1,000,000 chars) or `SEARCH_TIME_BUDGET_MS` (100 ms of wall clock, read through `performance.now()`). The test feeds 1.1 M chars and asserts `scannedChars === 1_000_000`, which only holds when the scan finishes inside 100 ms. On a loaded CI runner the time budget fires first at a chunk boundary, `truncated` is still `true`, and the exact-count assertion fails. The test is asserting the character budget while the code is free to stop on the time budget.

## Fix direction

Make the clock injectable on the search path (the `wait_for_material` path in the same file already takes an injectable `now`), have the test pin a clock that never advances so only the character budget can stop the scan, and add a separate test that drives a fake clock past `SEARCH_TIME_BUDGET_MS` and asserts the time-budget stop. No production behaviour change; the test becomes deterministic instead of timing-dependent.

wyuc:
## Summary

`tests/agent-runtime/material-tools.test.ts` › "stops at the per-execution character budget and reports truncation" fails intermittently in CI (#1452): `search_material` stops on whichever of `MAX_SEARCH_CHARS_PER_EXEC` (1,000,000) or `SEARCH_TIME_BUDGET_MS` (100 ms wall clock) fires first, and the test asserts the exact character count, which only holds when the scan finishes inside 100 ms. On a loaded runner the time budget wins at a chunk boundary (`scannedChars: 786432`).

## Related Issues

Closes #1452

## Changes

- `lib/server/agent-runtime/material-tools.ts`: the search tool reads the clock through the injectable `now` dependency (`deps.now ?? Date.now`) that `wait_for_materials` already uses, instead of calling `performance.now()` directly at the four deadline checks. No limit or default changes.
- `tests/agent-runtime/material-tools.test.ts`: the character-budget test pins `now: () => 0` so only the character budget can stop the scan; a new time-budget test drives a fake clock that expires after exactly one 16,384-char chunk and asserts `scannedChars: 16_384`, `truncated: true`.

## Type of Change

## Verification

- `vitest run tests/agent-runtime/material-tools.test.ts`: 32/32. The character-budget test looped 20 times with the frozen clock: 20/20.
- Mutation: with `performance.now()` restored in the search path, the new time-budget test fails (`scannedChars` 1,000,000 instead of 16,384), so it pins the injected clock rather than the implementation detail.
- `pnpm exec tsc --noEmit`, eslint and prettier on the two files, and `pnpm check` all pass.

### DIFF
--- a/lib/server/agent-runtime/material-tools.ts
+++ b/lib/server/agent-runtime/material-tools.ts
@@ -402,7 +402,7 @@ export function buildMaterialTools(deps: MaterialToolDependencies): AgentTool<ne
         throw new Error('search_material query must contain 1 to 200 characters');
       }
       const needle = foldCase(params.query);
-      const deadline = performance.now() + SEARCH_TIME_BUDGET_MS;
+      const deadline = now() + SEARCH_TIME_BUDGET_MS;
       let scannedChars = 0;
       let truncated = false;
       const hits: Array<{
@@ -417,7 +417,7 @@ export function buildMaterialTools(deps: MaterialToolDependencies): AgentTool<ne
       for (const record of records) {
         throwIfAborted(signal);
         if (!isSearchableTextRecord(record)) continue;
-        if (scannedChars >= MAX_SEARCH_CHARS_PER_EXEC || performance.now() >= deadline) {
+        if (scannedChars >= MAX_SEARCH_CHARS_PER_EXEC || now() >= deadline) {
           truncated = true;
           break;
         }
@@ -430,7 +430,7 @@ export function buildMaterialTools(deps: MaterialToolDependencies): AgentTool<ne
         const maxDecodeBytes = Math.min(raw.length, remainingCharsBeforeRead * 4);
         const text = raw.toString('utf8', 0, maxDecodeBytes);
         const sourceWasByteTruncated = maxDecodeBytes < raw.length;
-        if (performance.now() >= deadline) {
+        if (now() >= deadline) {
           truncated = true;
           break;
         }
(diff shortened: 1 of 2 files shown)


CXuPercy (approved):
Cross-reviewed by two independent reviewers. The four deadline checks now read the injectable `now` (matching the `wait_for_materials` pattern), no `performance.now` residue in `material-tools.ts`, and the two budget tests are deterministic by construction — the file ran three times on the head (32/32 each round), and the scripted-clock test's read sequence (4 zero reads → one 16,384-char chunk → the 1000 trip) was traced against the loop by both reviewers. Baseline can't reproduce the flake locally, consistent with the loaded-runner diagnosis.

Two for-the-record notes, neither blocking: (1) this switches these checks' production default from monotonic `performance.now()` to wall-clock `Date.now` — an NTP step mid-scan can truncate a search slightly early or late; fine for a best-effort 100 ms budget, and keeping one clock pattern is the right call. (2) `dsl-tools.ts:894/910/949` (`grep_stage`) still calls `performance.now()` directly — the same determinism gap this PR fixes for `search_material`; worth a follow-up when a time-budget test is added there.

PR #1454, opened 2026-09-11
test(render-service): cover preview callbacks across the tsx/browser boundary

### LINKED ISSUE #1451: render-service: guard every browser-evaluated callback against the tsx/keepNames boundary, not just waitForDocumentAssets
## Context

#1412 showed that any function handed to Puppeteer's `page.evaluate()` / `frame.evaluate()` / `waitForFunction()` from a module compiled by `tsx` breaks at runtime if it contains a named inner function: esbuild's `keepNames` wraps the declaration in `__name(...)`, Puppeteer ships only the callback source to Chromium, and the helper does not exist there. #1421 fixed the one call site that had a named inner function (`waitForDocumentAssets`) by evaluating a source string, and added a tsx-boundary regression for that function.

The remaining browser callbacks in `render-service/src/preview-renderer.ts` (the `about:srcdoc` readiness wait, the mounted-marker wait, the slide/viewport injection, the scene-id check) are plain arrow functions with no named inner function today, so they are safe by accident, not by construction. The next named helper inside any of them reintroduces the failure, and the route tests still cannot see it because they never start Chromium.

## Ask

Guard the pattern rather than the one site. Either of:

- a test that loads `preview-renderer.ts` through the production `tsx` loader and runs every callback sent to the browser inside an isolated `vm` context (the approach the #1421 regression uses for one function), so a stray `__name` fails in regular CI; or
- run the real-Chromium preview smoke tests from #1421 in CI by provisioning a headless shell in the render-service job, instead of leaving them opt-in behind `PUPPETEER_EXECUTABLE_PATH`.

A lint rule forbidding function declarations inside `evaluate` callbacks would be a cheap complement but is not sufficient on its own.

## Related

#1412, #1421

YizukiAme:
Extends the existing boundary regression to exercise slide and interactive rendering through the production tsx loader. Browser callbacks execute in isolated VM contexts, covering prop injection, readiness checks, scene selection, and Page/Frame asset waits.

Closes #1451.

Validation: 160 tests passed under Node 22; 4 optional Chromium tests skipped. Typecheck and formatting passed. Each of the four function callbacks was temporarily mutated with a named inner function; all mutations failed with `__name is not defined`. No production code, dependencies, or CI changes.

### DIFF
--- a/render-service/test/preview-browser-boundary.fixture.mjs
+++ b/render-service/test/preview-browser-boundary.fixture.mjs
@@ -1,0 +1,163 @@
+// Run with node --import tsx so callbacks come from the production transform.
+import assert from 'node:assert/strict';
+import { createContext, runInContext } from 'node:vm';
+import { buildSlideClientBundle, ChromiumPreviewRenderer } from '../src/preview-renderer.ts';
+
+const type = process.argv[2];
+assert.ok(type === 'slide' || type === 'interactive');
+const viewport = { width: 640, height: 360, deviceScaleFactor: 1 };
+const canvas = {
+  id: 'canvas',
+  viewportSize: 1000,
+  viewportRatio: 0.5625,
+  theme: { backgroundColor: '#fff', themeColors: [], fontColor: '#111', fontName: 'sans-serif' },
+  elements: [],
+};
+const scene = {
+  id: 'boundary-scene',
+  stageId: 'stage',
+  order: 1,
+  title: 'Preview',
+  type,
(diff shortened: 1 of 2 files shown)


wyuc (approved):
Test-only, drives both preview flows through the production tsx loader with every browser callback executed in a helper-free VM context, and relies only on the existing buildSlideClientBundle / browserLauncher seams. Closes the pattern-level gap from #1451. Merging.

Issue #1350, opened 2026-09-02
video export: stop wasting the browser-side compile when the render queue is full


cosarah:
### Problem or Motivation

The one-click MP4 export spends its entire browser-side compile **before the first contact with admission control**:

1. `useVideoRenderStore.startRender` → `buildExportZip()` — Dexie duration probes, per-slide off-screen geometry measurement, quiz layout probes, asset collection (slide snapshots + narration media), ZIP packaging. Seconds to tens of seconds.
2. Only then `POST /api/export-video/render` reaches the render service, where `RenderCoordinator.reserve()` runs as the very first admission check.

If the queue is full, the `429` arrives after the full compile cost was paid. Worse, `zipBlob` is a local variable inside `startRender` — **nothing is cached**, so a retry recompiles from scratch. The render service itself is designed so "a rejected caller never buffers a byte" (`render-service/src/main.ts`); the client deserves the same courtesy.

The only pre-check today, `/api/export-video/capability`, probes `GET /health`, which reports no admission state (addressed by #1348, which adds an `accepting` flag).

### Proposed Solution

Two independent halves; the second has no service dependency and can land first.

1. **Preflight the admission flag** (requires the `accepting` flag from #1348):
   - The capability endpoint forwards the service's `accepting` boolean.
   - The video-export dialog probes capability on open (it already does, for `enabled`) and disables the "Render MP4" button with a localized busy hint while `accepting === false`. "Download ZIP" and subtitle download stay available.
   - Preflight is UX only — the `429` at submit time remains the authoritative answer; the race between probe and submit is accepted.
2. **Reuse the compiled ZIP on retry**:
   - Cache `{ zipBlob, options, stageRevision }` in the video-render store after a successful compile.
   - A resubmit with unchanged options and unchanged scenes reuses the cached ZIP instead of re-running `buildExportZip`.
   - Invalidate on scene edits or any render-option change (a cheap revision counter over the stage/scenes suffices; options are already held in the store).

### Acceptance Criteria

- [ ] Dialog opened while the queue is not accepting → "Render MP4" disabled with a localized hint; ZIP download and subtitles unaffected
- [ ] A 429-then-retry with unchanged scenes and options does **not** re-run `buildExportZip`
- [ ] Editing scenes or changing options invalidates the cache (next render recompiles)
- [ ] No behavior change when the render service is unconfigured (existing degrade path)

### Alternatives Considered

[...]

PR #1455, opened 2026-09-11
feat(export): check render queue availability before compiling

YizukiAme:
When the render queue is full, users currently discover that only after compiling an export ZIP. Forward the render service health response’s optional `accepting` flag and check it each time the export dialog opens. An explicit busy response disables MP4 with a localized hint and a manual **Check again** action; ZIP and subtitle downloads remain available.

Older healthy services without `accepting` retain their current behavior. Unconfigured or unreachable services still disable MP4, and submission remains authoritative if the queue fills after the check. No polling or queue reservations are added.

Refs #1350 — the preflight portion only. ZIP retry reuse is covered independently in #1456; keep the issue open until both parts are merged.

Validation (Node 22):
- Capability and dialog tests cover busy/ready, older services, unavailable services, manual recheck, reopening, and stale responses.
- Video export and media polling/replacement regression tests, type checking, changed-file lint/format, and all 12 locale key checks pass.
- Chromium smoke test with controlled capability responses verifies busy → manual recheck → MP4 enabled, ZIP/subtitle availability, and reopening. This does not exercise a real saturated render service or complete MP4 rendering.


wyuc (approved):
Reviewed the full diff. The busy predicate exposed as `accepting` is the same predicate the server's `reserve()` rejects on, so a disabled MP4 button always corresponds to a request that would have been 429'd, and unknown/malformed values fall back to the previous behaviour rather than to busy. Verified the nine health-response cases (unconfigured, unreachable, 5xx, true/false, missing, non-boolean, non-JSON, capability route failure), the reopen / double-recheck / close-mid-request races, and all 12 locale bundles. CI green on the rebased head.

Two optional follow-ups, non-blocking: the `checking` reset on dialog reopen is not pinned by a test (removing it keeps the suite green), and that reset does not clear the previous `accepting` value, so a reopened dialog briefly shows the last busy hint until the fresh probe returns. Both only err on the side of reporting busy, never on the side of enabling a submission that would be rejected.

wyuc (approved):
Re-approving after update-branch (merge of main only, PR content unchanged). CI green on e5f0fa864cd39ca96d77efb10979dd605a678b57.

PR #1456, opened 2026-09-11
fix(export): reuse compiled ZIP when retrying video renders

YizukiAme:
A render submission rejected with 429 currently makes the next attempt compile the same ZIP again. Keep one private in-memory ZIP result in the render store so an unchanged retry submits the same Blob, course name, and warning counts without recompiling. FPS and quality remain submission-only options and can change on a cache hit.

Invalidate on course/scene or media-task changes, compilation option or locale changes, committed audio/media table writes, and existing asset-replacement notifications. On a potential cache hit, resolve the current course name through the same DocumentStore helper as the compiler and compare it with the cached result. This catches document-only renames without relying on the legacy Dexie `stages` table. A revision captured before compilation prevents an intervening edit/reset from repopulating the cache. Observers exist only while an export is active or the retry slot is retained.

Keep valid ZIPs after 429, retryable server failures, or a started render failure. Release on success, reset, invalidation, 400/413, or the existing automatic ZIP-download fallback. This preserves current network-error fallback, cancellation, progress handling, and lazy compiler loading; the standalone ZIP download path is unchanged. No dependencies or persistent cache are added. There is no dedicated cache-clear UI: after a retryable failure the single ZIP remains in memory until invalidation, a subsequent successful export, or the session ends; a document-only rename is detected on the next retry.

Refs #1350 — the retry-cache portion only; queue preflight has merged in #1455.

Validation (Node 22):
- 32 cache tests cover same-Blob reuse and new FPS/quality, invalidation (including real Dexie commits under unchanged IDs), document-only renames, compilation-time edits, changes during the async name check, retention/release, downloads, cancellation, and observer cleanup.
- Video export and media polling/replacement regression tests, type checking, and changed-file lint/format pass.
- After merging current main (including #1455), the related suite passes: 465 passed, 5 skipped. A previous Chromium dialog smoke test covered controlled busy/ready capability responses. No real saturated render service or complete MP4 rendering was exercised.

The revision check now exists only at cache population; invalidation already clears the slot. Mutation checks independently remove name comparison, the write revision guard, option invalidation, and pending-name-read invalidation: each makes the tests fail (2, 4, 3, and 1 failures respectively), and the restored code passes. Change-back tests also verify that compilation options release the slot immediately, even if their original values are restored before retry.


wyuc (changes requested):
The cache mechanics hold up well under probing: every compile input I could enumerate (scene content and order, stage object, media tasks, narration and generated-media bytes including real Dexie commits under unchanged ids, asset-pool replacements, locale, resolution, subtitles) invalidates, fps/quality-only changes reuse, observers drain to zero on success / reset / compile failure / network-fallback and stay at exactly one set across five consecutive 429s and three course switches, retry classification matches the existing error mapping, and no existing test expectation was touched. One input is missing, plus a test gap.

**Blocking: the course name is a compile input with no invalidation path.** `buildExportZip` resolves the name through `accessDocument(stage.id)` (document store), while `observe-export-changes.ts` watches only `useStageStore.{stage,scenes}` and the `audioFiles` / `mediaFiles` Dexie prefixes; a name-only write deliberately does not touch the stage store (`use-workbench-session.ts` says so). Probe: 429, rename the course, retry → `build` runs once and the download is `Old Name.mp4`, with the old name also inside the ZIP (`<title>` and README). Rendered frames are unaffected, so it is metadata, but it contradicts the PR's "unchanged retry" contract. Either include `stageName` in the cached record and compare it, or add `stages` to the watched Dexie prefixes, plus a regression.

**Please also pin the revision guard.** Deleting `revision === exportRevision` at the cache write, or `cachedZip.revision !== exportRevision` at the cache read, each leaves all 27 tests green; only deleting both turns four tests red. The two halves are correct belt-and-braces, but a test that starts a second `start()` immediately (no intervening invalidation) and asserts `build` is not re-entered and the slot was never populated would pin the write-side guard on its own. Similarly, the `setOptions` invalidation is only covered because the cache-key comparison masks it.

Non-blocking notes: on a cache hit the first toast is still `export.videoRendering` ("Rendering slides and collecting assets…"), which names exactly the work the hit skips; nothing in the product can release a retained ZIP (`reset()` has no callers) so after a 429 the full export stays in memory until an edit, course switch or successful render, which is acceptable for a single slot but worth a sentence in the description; only 400/413 are treated as "archive is bad", so a 415/422 would retry the same bytes indefinitely (cheap, but a category would be safer than a hardcoded pair).

430 tests in `tests/video-export`, tsc, eslint, prettier and `pnpm check` pass on b508b40c.

wyuc (approved):
Round 2 on 842727b5, both round-1 items closed.

(A) Rename: the document-only rename now recompiles and downloads under the new name (test at `video-render-cache.test.ts` "recompiles a document-only rename…", stage store asserted untouched); dropping `|| currentName !== built.stageName` turns two tests red; `resolveExportStageName` is byte-identical to the old inline fallback chain and the name reaches `<title>` / README through the same IR path.

(B) Revision guard: with the read-side comparison gone, deleting `revision === exportRevision` at the write turns the four "does not repopulate a cache invalidated during compilation" cases red on the build count, so the remaining guard is pinned and load-bearing. Removing the read-side check is sound: the only writer of `exportRevision` is `invalidateCachedZip`, which nulls the slot in the next statement, so a non-null slot can never carry a stale revision.

Also probed the new async name-check window: `reset()`, a course switch, and `scenes: []` during the pending `accessDocument` all force a recompile via `cachedZip?.result !== built` (that check is pinned by the committed `scenes: []` case); duplicate `start()` during the check is blocked before `activeRenders` increments and the observer drains to zero afterwards; a slot compiled while the document store was absent is reused when the document later appears with the same name; the slot written after a switch is keyed by the re-read `compileStageId`. `tests/video-export` 444/444, tsc, eslint, prettier clean; no lockfile change.

Two cosmetic notes, no action needed: the committed tests cover the name-check window only with `scenes: []` (reset / course-switch pass but are not pinned), and on a name mismatch the toast flips from `export.videoRendering` to `export.videoCompiling` under the same id.

PR #1459, opened 2026-09-11
fix(settings): read data.error for image/video provider test results

### LINKED ISSUE #1457: [Bug] Image/Video provider "Test Connection" shows "failed: undefined" — panels read data.message but the API returns error
### Environment

- Version: `main` @ 29735f1 (self-hosted, `pnpm dev`)
- macOS arm64, Node 22

### Description

When an image or video provider connectivity test fails, the settings UI always shows something like:

> Image service connection failed: undefined

instead of the real upstream error. This makes self-hosted provider debugging (wrong key, unactivated model, bad base URL) very hard, because the actual failure reason is hidden.

### Root cause

The verify routes return errors via `apiError()`, which serializes `{ success: false, errorCode, error }` — the human-readable message lives in the **error** field. But the two panels read `data.message`:

- `components/settings/image-settings.tsx` (~line 138): ``${t('settings.imageConnectivityFailed')}: ${data.message}``
- `components/settings/video-settings.tsx` (~line 100): same

All other settings panels (ASR, PDF, `provider-config-panel`, `model-edit-dialog`) correctly read `data.error`, so this looks like an oversight in these two panels only. The success path is unaffected because `apiSuccess({ message })` does include `message`.

### Reproduce

1. Settings → Image generation → pick a provider, enter a wrong API key (or an unreachable base URL)
2. Click "Test Connection"
3. The red box shows `failed: undefined` instead of e.g. `OpenAI Image auth failed (401): ...`

### Fix

Read `data.error` in those two places (2-line change). Verified locally against a real provider (the exact upstream error text shows up). Happy to send a PR.

SeanLiew523:
Fixes #1457

The verify routes (`/api/verify-image-provider`, `/api/verify-video-provider`, ...) return errors via `apiError()`, which serializes `{ success: false, errorCode, error }`. The image and video settings panels read `data.message` instead, so every failed connectivity test renders `failed: undefined` and the real upstream error is hidden from the user.

All other settings panels (ASR, PDF, `provider-config-panel`, `model-edit-dialog`) already read `data.error` — this brings the two remaining panels in line.

### Testing
- Self-hosted `pnpm dev`, macOS arm64
- With a wrong API key, the image provider test now shows e.g. `Seedream auth failed (401): {"error":{"code":"AuthenticationError",...}}` instead of `undefined`
- `pnpm lint` / type check pass on the touched files (2-line change)

Happy to adjust if you'd rather normalize the API error body shape instead of the client readers.

### DIFF
--- a/components/settings/image-settings.tsx
+++ b/components/settings/image-settings.tsx
@@ -135,7 +135,7 @@ export function ImageSettings({ selectedProviderId }: ImageSettingsProps) {
         setTestMessage(t('settings.imageConnectivitySuccess'));
       } else {
         setTestStatus('error');
-        setTestMessage(`${t('settings.imageConnectivityFailed')}: ${data.message}`);
+        setTestMessage(`${t('settings.imageConnectivityFailed')}: ${data.error}`);
       }
     } catch (err) {
       setTestStatus('error');
--- a/components/settings/video-settings.tsx
+++ b/components/settings/video-settings.tsx
@@ -97,7 +97,7 @@ export function VideoSettings({ selectedProviderId }: VideoSettingsProps) {
         setTestMessage(t('settings.videoConnectivitySuccess'));
       } else {
         setTestStatus('error');
-        setTestMessage(`${t('settings.videoConnectivityFailed')}: ${data.message}`);
+        setTestMessage(`${t('settings.videoConnectivityFailed')}: ${data.error}`);
       }
     } catch (err) {
       setTestStatus('error');


CXuPercy (approved):
Cross-reviewed by two independent reviewers. Confirmed on both sides: `ApiErrorBody` carries `{success, errorCode, error}` with no `message` field, every failure path of the two verify routes goes through `apiError()` (no thrown escape or non-JSON failure), the success path's `message` is ignored by both panels anyway, and these were the last two `data.message` reads in the settings surfaces.

Locally verified in the browser on this branch: with a deliberately wrong key, the image panel now surfaces the upstream error verbatim (`Incorrect API key provided: …`) and the video panel surfaces its real validation error (`No model configured for video provider: …`) — both previously rendered as `failed: undefined`.

**Local verification screenshot:** 
image


wyuc (approved):
Verified: apiError() serialises {success, errorCode, error} with no message field, and these were the last two data.message reads under components/settings. Merging.

Issue #1449, opened 2026-09-10
Model discovery timeout does not cover response body consumption


takagibit18:
### Bug Description

The current generic model-discovery path starts a 15-second timeout for each candidate request, but clears the timer once `fetch()` returns a `Response`, before `fetchModels()` consumes its body. A provider that returns headers within the deadline but stalls the body can therefore keep `fetchModels()` pending past `FETCH_TIMEOUT_MS`.

I reproduced this with the real repository code and a loopback-only HTTP server. The affected flow is:

```text
Settings → Fetch Models → POST /api/provider/probe-models
→ fetchModels() → fetchWithTimeout(..., 15_000)
→ fetch() resolves Response → timer cleared → res.json() / res.text()
```

The [helper clears its timer in finally](https://github.com/THU-MAIC/OpenMAIC/blob/29735f10d0081859ac3db1a50a0cc92f46436004/lib/server/fetch-with-timeout.ts#L11-L16), while [model-fetch reads the body afterwards](https://github.com/THU-MAIC/OpenMAIC/blob/29735f10d0081859ac3db1a50a0cc92f46436004/lib/server/model-fetch.ts#L122-L149). Obtaining a Response does not mean its body has finished downloading.

### Steps to Reproduce

1. Use the commit below with the locked dependencies (`pnpm install --frozen-lockfile --ignore-scripts`). Run a local TypeScript harness with `pnpm exec tsx` that imports the real `lib/server/model-fetch.ts`.
2. Start a Node HTTP server on `127.0.0.1` and a random port. Call `fetchModels(baseUrl, 'test-key', { modelsUrlOverride: localModelsUrl })`, so there is only one candidate.
3. Immediately send `200` headers with `Content-Type: application/json`, flush headers, and write only `{"data":[`. Keep the response open without calling `res.end()`.
4. Observe the promise's settlement state after approximately 16.5 seconds; it is still pending.
5. Finish the same response with `res.end('{"id":"late-test"}]}')`. The original promise resolves with `late-test` immediately afterwards.

No real provider or API key is involved. The harness bounds its own waits and closes the local connections; it does not modify production code. The full harness and raw logs are available if useful.

### Expected Behavior

My expectation is that the existing 15-second budget for each candidate covers both obtaining the response and consuming its body. Please confirm whether that is the intended contract; this is not a proposal for a single 15-second budget across all fallback candidates.

### Actual Behavior

[...]

PR #1465, opened 2026-09-11
[AI-assisted] feat(export): pre-empt a render the queue would reject (#1350)

LeoParkerOu:
## Summary

Forward the render service's queue-admission signal to the video export dialog, so a submission the queue would reject with 429 does not first incur a full browser-side compile. The service-side 429 remains authoritative; this is an advisory preflight only.

This PR is AI-assisted and manually reviewed.

## Related Issues

Related to #1350 (Part A: queue admission). The compiled-ZIP retry cache is intentionally deferred as Part B because it overlaps the in-flight #1414.

## Changes

- Change `checkRenderServiceHealth()` to return `{ enabled, accepting }`
- Forward `accepting` through `GET /api/export-video/capability`
- Disable **Render MP4** and show a localized busy hint only when the service explicitly reports `accepting: false`
- Add `export.videoQueueBusy` to all 12 locale files
- Add health-probe, capability-route, and locale-coverage regression tests

## Type of Change

### Steps to reproduce / test

1. Configure a render service whose `GET /health` returns `accepting: false`.
2. Open the video export dialog.
3. Confirm **Render MP4** is disabled and the queue-busy hint appears, while ZIP download remains available.
4. When the service reports `accepting: true`, confirm **Render MP4** is enabled again.

- Health-probe behavior for unconfigured, queue-full, missing-flag, non-JSON, non-2xx, and transport-failure cases.
- Capability-route responses for an accepting and non-accepting service.
- The fallback rule: a missing or malformed `accepting` flag is treated as `true`, preserving compatibility with older services and preventing an advisory probe from locking out a render that could succeed.
- Static wiring review of the dialog: only an explicit `accepting: false` disables **Render MP4**; ZIP export remains available.

I did not run a live render-service instance, so the dialog state was not verified end-to-end in a browser.

### Evidence

- [x] Manually tested locally
- [ ] Screenshots / recordings attached (UI state requires a running render service)

Local results:

- `tests/server/render-service-health.test.ts` and `tests/i18n/video-export-queue-locales.test.ts`: 2 files, 9 tests passed
- Prettier check passed for all changed files
- ESLint passed for changed TypeScript/TSX files
- i18n key alignment check passed for all 12 locale files
- Broader targeted run: 11 files, 111 tests passed

The checkout has unresolved workspace package dependencies (`@openmaic/storage` and `pptxgenjs`) that prevent six unrelated export suites from loading; these failures also occur without this change.

## Checklist


wyuc:
Thanks for this, and sorry for the overlap: Part A of #1350 landed in #1455 (merged today as c460bceb, before this PR could be reviewed) with the same shape — `accepting` forwarded through `/api/export-video/capability`, the MP4 button disabled with a localized busy hint only on an explicit `accepting: false`, unknown/malformed values treated as accepting, all 12 locales, plus the health-probe and capability-route tests. Part B also landed in #1456, and #1350 is closed. Closing this as superseded rather than asking you to rebase onto an identical change. If you want to pick up something in the same area, #1449 (model discovery timeout not covering body consumption) is open and unclaimed.

Issue #1413, opened 2026-09-08
[Task] Learner runtime endpoints are unreachable in production builds without the insecure development authenticator


wyuc:
## Summary

Under server-backed persistence in a production build, the learner runtime endpoints (`/api/persistence/runtime/...`) answer 401 for every request unless `PERSISTENCE_ALLOW_INSECURE_DEV_AUTH` is set, because they still go through the development authenticator, which refuses in production by design (#1387). Documents and, since #1392, assets derive their principal on the server instead, so the symptom is silent: media and documents work, learner state does not — and the client keeps retrying the runtime session lookup with backoff on every load (11 requests in a 65-second visitor session were observed).

The choice an operator has today is a permanently non-functional learner runtime or an authenticator that provides no user isolation.

## Direction

The runtime routes need a server-derived learner principal like documents already have (the anonymous owner cookie is the natural minimum; a verified session claim is the real model, #1129 item 1). Two independent pieces:

1. Route: derive `learnerKey` on the server for `/runtime` requests (the same `withRequestOwnerId` path documents use), keep the development authenticator only for admin/merge operations.
2. Client: stop retrying a 401 from the runtime store — an authentication refusal is terminal for the session, not a transient availability gap.

Reported by a self-hosting operator while verifying #1392 (see #1386). Refs #1129, #1387, #1392.

wyuc:
Cross-reference from the review of #1392: with server-backed media, `POST /api/persistence/assets` (allocation) shares the open-write posture that document writes already have under the shared principal, and because the asset quota is counted per principal, an anonymous writer on a deployment without an access code can drive the pool to `507 ASSET_QUOTA_EXCEEDED` for everyone. Gating allocation on the development token is not an option in production (it requires the insecure opt-in), so the real fix is an access decision that knows the caller — the same principal gap as the learner-runtime 401 here. Discussion: https://github.com/THU-MAIC/OpenMAIC/pull/1392#discussion_r3987775086

Hosuke:
I’d like to take the independent client retry slice (item 2): stop automatic chat autosave retries after a runtime 401, retain unsaved chat state, and keep document saves working. I’ll cover incremental/full/departing-stage saves and recovery on a fresh stage load. The server-derived learner principal remains a separate follow-up, so this PR will leave the issue open.

wyuc:
[...]

Issue #1394, opened 2026-09-06
[Task] Reclaim asset registry entries that no document references


wyuc:
## Summary

Asset pool entries that no document references are never reclaimed on the server.

With #1391 (PR #1392) the classic generation chain writes generated bytes through the asset pool under server-backed persistence. Every `put` creates an `asset_entries` row. The offline `AssetCollector` reclaims only **blobs that no registry entry names**, so a registry entry that lost its last document reference — a regenerated slide, a deleted course, an allocation whose write-back could not be completed — keeps its bytes alive indefinitely. The stage-scoped sweep (`lib/media/reclaim-stage-assets.ts`) exists but is intentionally unwired, and client-side `remove` is only issued on the narrow "provably never referenced" path.

## Impact

Unbounded server-side growth proportional to regeneration and deletion activity. No correctness impact: nothing dangling is ever served.

## Direction

Wire a server-side or offline reference sweep over the document store — enumerate asset ids referenced by live documents per stage, release registry entries no document names (respecting the collection grace period), and let the existing collector reclaim the bytes. This is the document-level half of the two-level reference counting #1007 describes (part 3, manifest), so it should reuse the manifest enumerator rather than a second ad-hoc scan. Deletion of a course should trigger the sweep for that stage.

Refs #1007, #1391, #1392

wyuc:
Superseding note: the reclamation described here is being redesigned rather than wired as written. The amendment on #1007 (newest comment) moves the document → asset reference level into the server — a reference table maintained by the document store, pending → committed allocations with expiry, and an entry pass in the existing collector — so this becomes a query instead of a document walk. This issue stays open as the tracking item and closes with the first two PRs of that amendment.

wyuc:
Closed as superseded by the #1007 amendment, now landed: #1472 (`e824117c`) adds the reference table, the entry pass with backfill, the legacy mark and a standing sweep for entries nothing references, and #1473 (`c605e9e6`) turns it on in the app and releases assets on course deletion. Entries no document references are stamped and reclaimed after the grace period by the collector's entry pass, which is what this task asked for.

PR #1472, opened 2026-09-11
feat(storage): give the server the asset entry lifecycle (#1007 amendment, part 1)

wyuc:
Implements part 1 of the lifecycle amendment on #1007 (https://github.com/THU-MAIC/OpenMAIC/issues/1007#issuecomment-5638002396). Tracking: #1394.

Under server-backed persistence every `POST /assets` creates a registry entry that nothing releases: the collector only ever selects from `asset_blobs`, the client is refused `DELETE`, and the quota #1392 added counts every entry a principal ever allocated. This PR gives the server the entry lifecycle so reclamation and quota become queries over data the server already holds.

### What changed (`@openmaic/storage` 0.29.1 → 0.30.0)

- **Schema** (additive, idempotent, appended to `ASSET_PG_SCHEMA`): `asset_entries.committed_at / expires_at / unreferenced_at`; `document_asset_refs (stage_id, scope, scene_id, asset_id → asset_entries ON DELETE CASCADE)` with a structural `scope` column (`'stage'` | `'scene'`) so no scene id can collide with the stage-level scope; `asset_reference_tracking` (a one-row marker that a reference writer runs on this database); `document_asset_withdrawals` (which documents were retired with their rows kept); supporting partial indexes. `ASSET_PG_SCHEMA` now has a verbatim golden pin like every other schema, and the asset tables are in the PG contract suite's required-tables list with a real insert behind the assertion.
- **Allocation is pending until a document names it.** `put` stamps `expires_at = now() + pendingTtlMs` (new store option, default 24 h). The first document write that references the id commits it (`committed_at`, `expires_at = NULL`) inside that write's own transaction. No new route; lifecycle state is unobservable on every asset route.
- **Reference maintenance** (`src/asset/references.ts`): the only code that writes `document_asset_refs` or the lifecycle columns. A slot value becomes a row only if the registry holds that id (`WHERE EXISTS`); nothing parses an id. Scope is exact: `putScene` touches its scene's rows, `putStage` the stage-level rows, `saveDocument` the whole stage, `deleteScene` / `deleteDocument` remove theirs. An entry that loses its last row is stamped `unreferenced_at`; one that regains a row is cleared.
- **`PgDocumentStore.trackAssetReferences`** (default `false`): opts the five write paths into that maintenance. Off, a deployment without the asset schema is untouched.
- **Collector entry pass** (`documentReferences: true`): before the blob pass, release entries that are pending and expired, or committed and unreferenced for longer than the grace period — same batch cap, each row locked and re-checked in its own transaction. A bounded, resumable **backfill** runs first on an upgraded deployment: legacy entries (`committed_at IS NULL AND expires_at IS NULL`) block every release until every stored document has been enumerated; then they are marked committed and the ones no document names are stamped to drain after grace.
- **Hosts that retire documents by tombstone** (keeping the rows, because an app-owned tombstone cascades from the document row) get `PgDocumentStore.withdrawAssetReferences(stageId)`: owner-scoped, locks the stage row, removes every reference row, stamps entries that lose their last reference, records the withdrawal so the one-time backfill never re-references a retired document; a later save re-establishes references and clears the record, and a tracked `deleteDocument` clears it too.
- **`PgDocumentStore.declareAssetReferenceTracking()`** lets a host that builds every document store with `trackAssetReferences: true` declare it at startup, so the entry level and the backfill are eligible from the first collector pass instead of after the first document write.
- **Quota** now sums logical bytes over entries not yet stamped unreferenced (`unreferenced_at IS NULL`), on both `put` and `replace`; a `replace` on an entry outside the live sum is a no-op for the check.
- **Docs**: the asset contract's reclamation and quota sections now describe both levels; the reference server documents the three options and that a host must schedule `collect()`; the document contract notes the side effect.

### Design notes

- Scene attribution runs the DSL enumerator over exactly one scene (or the stage alone), so it is exact without parsing structural keys; `videoManifest` keys are stage-level and land in the stage scope, the conservative direction.
- `deleteDocument` removes reference rows only after locking the scoped stage row, so a foreign or missing stage cannot drop another scope's rows.
- Marking legacy entries is two statements because the commit and the stamp may touch the same row; the stamp's predicate is deliberately the general "committed, unstamped, unreferenced" condition.
- `collect()` still returns the blob count the existing scheduler logs; entry counts are on `collectPass()`.

### Not in this PR

- App wiring (enabling `trackAssetReferences` / `documentReferences`, surfacing `pendingTtlMs`, deleting the dead client-side registry sweep) is part 2 of the amendment.
- The principal model (#1129 item 1, #1413). The reference table is the data it needs.

### Known limitations

- `documentReferences` on a collector whose document store does not run with `trackAssetReferences` would expire entries live documents still name. A marker table (`asset_reference_tracking`, written by every reference-maintaining transaction) makes the entry pass refuse loudly when no tracking writer has ever run. The marker is one-way: rolling `trackAssetReferences` back while leaving `documentReferences` on is documented as a hazard, not enforced (and a tracking-off delete cannot clear a withdrawal record, for the same reason); part 2 makes the pairing unconditional and declares it at startup.
- Repeated `replace` on an entry that has lost its last reference grows physical storage until the entry is released after grace; it is named in the contract as a physical-headroom input rather than refused, because refusing would make lifecycle state observable on a write route.
- The backfill cursor is per collector instance: a restart re-walks. Safe by construction (releases are gated on a completed walk), not resumable across restarts. Legacy marking locks every committed, unstamped entry for one statement, once per deployment.
- `document_asset_refs` has no FK to `document_stages`, so a stage deleted out of band leaves rows that keep entries alive — the safe direction, not self-healing.
- The typed lock-contention error (`StorageLockUnavailableError`, SQLSTATE 55P03/40P01) is mapped on the document store's five write paths and in the collector; folder mutations and the registry's own write path still surface raw driver errors as before.

### Review rounds

Nine implementation rounds against seven adversarial review rounds with PostgreSQL probes ran before this was marked ready; every finding is fixed and pinned by a test that was checked to fail without the fix. The ones worth a reviewer's attention: a scene whose id is the empty string collided with the stage-level scope sentinel (now a structural `scope` column); `elements: [null]` in a stored document threw out of the enumerator and would have stalled the backfill for the whole deployment (non-object members are filtered); the entry pass's locked `NOT EXISTS` re-check was snapshot-stale against an insert-only writer on another instance (lock first, then re-check in a fresh-snapshot statement — the same discipline applied to legacy marking); the one-time backfill could re-reference a document a host had retired by tombstone (now a recorded withdrawal, deliberately **not** inferred from an entry's lifecycle columns, because a pre-tracking live document sharing an id with a deleted one would otherwise lose its asset); `replace` quota is computed on the post-state; internals are no longer exported. Typed errors added along the way: `StorageLockUnavailableError` (SQLSTATE 55P03/40P01 on the document write paths and the collector), `AssetCollectionFailure` (carries `stageId`), `AssetReferenceTrackingNotEnabledError`, `DocumentAssetReferencesDisabledError`.

### Verification

Against PostgreSQL 16, Node 22: package suite in CI shape (capture baseline → vitest → assert) 1260 passed / 7 skipped (the S3 suite and four agent-session cases, pre-existing), every `*.pg.test.ts` executed, all twelve required tables gained inserts; new `test/asset-lifecycle.test.ts` plus thirty-two real-PG lifecycle cases including the interleavings above; `pg-schema-contract` with a verbatim asset pin; `check-package-version-bumps.mjs` passed; `pnpm check` and `pnpm lint` clean; app `tests/persistence` + `tests/media` 624 passed with no app file changed.


wyuc:
Back to draft briefly: wiring part 2 showed that a host which retires documents by tombstone (keeping the rows, as this app does) has no way to withdraw a document's references, so one additive method is coming — `PgDocumentStore.withdrawAssetReferences(stageId)` — with its PG tests. Will mark ready again once it is in and CI is green; nothing already here changes.

wyuc:
Ready again at `181fe038`. Since the draft toggle: `withdrawAssetReferences` (for hosts whose tombstone cascades from the document row), `declareAssetReferenceTracking` (so a cold or upgraded database's entry level is eligible from the first pass), a recorded-withdrawal table so the one-time backfill never re-references a retired document — and, found on the way, the backfill must always re-reference whatever a live unwalked document names, never skip on an entry's lifecycle columns, or a pre-tracking document sharing an id with a deleted one loses its asset. The PR description's review-rounds section lists the rest; CI is green on every push.

wyuc:
@cosarah gentle ping on this one when you have a moment — part 2 (#1473) is already approved and is waiting only on this part to land.

Since your approval of #1473, `main` moved under this branch (#1499 / #1500), so `723d2f24` merges `main` back in. Two conflicts, both mechanical: `document/pg.ts` keeps both new imports (`encodeJson` from #1499 and `asStorageLockUnavailable` from this branch), and the package version goes to **0.31.0** because `main` already shipped 0.30.1. No code from this PR changed. CI is green on the merged head, and the package suite plus the PG contracts (1293 tests, PostgreSQL 16) pass locally on it as well.

cosarah (commented):
Thanks for this PR — the concurrency reasoning here is excellent, and the verification claims held up locally (full suite incl. real PostgreSQL 16: 1293 passed / 7 skipped). Three findings worth a look before merge:

**F1 (medium-low, worth fixing) — transient entry-level failures skip the blob pass, contradicting the stated invariant**
`collector.ts:406` — any throw from `entryLevelPass` (lock timeout in `markLegacyEntries`, a backfill FK conflict, …) propagates before the blob candidate query at `collector.ts:413`, so that pass's blob reclamation is skipped entirely. Only the missing-marker refusal is deferred to `collector.ts:471` after the blob pass. But the comment at `collector.ts:398-401` and `reference-server.md` promise the blob pass still runs when the entry level has problems. Suggested fix: wrap `entryLevelPass` in the same record-and-throw-after pattern used for `trackingFailure`.

**F2 (medium, operational) — `markLegacyEntries` takes `FOR UPDATE` over every live entry in one transaction**
`collector.ts:686` — step (2) locks all committed, unstamped entries, i.e. every healthy entry on the deployment:
- The "no cycle is possible" comment doesn't hold: the mark holds already-scanned row locks while it waits, and a concurrent document write's `commitReferencedEntries` (`references.ts:301`) takes `FOR NO KEY UPDATE` in plan order — a cycle is possible, and the deadlock victim can be a user save (the app's HTTP layer doesn't retry).
- Even without a cycle, during the upgrade window every document write touching existing entries blocks up to the 30s `lock_timeout`, producing user-visible lock-timeout errors.

One-time and upgrade-window-only, but suggest: (a) batch the mark by id the way `releaseEntries` does; (b) correct the comment; (c) note in the upgrade guidance to run it in a quiet window. On a busy system the mark can time out and retry repeatedly, keeping the legacy gate closed and deferring entry reclamation indefinitely.

**F3 (low, FYI) — new deadlock surface between two user saves**
`references.ts:301` / `references.ts:329` — two stages of the same principal sharing asset ids (a user copying across documents) that save concurrently each lock overlapping entry sets in unordered batches and can deadlock each other; the victim gets a retryable `StorageLockUnavailableError`. Low probability; sorting the id array in JS before `= ANY($1)` would make lock order mostly deterministic, or this can be explicitly accepted. Noting that the app doesn't wire the options yet, so the exposure arrives with part 2.

wyuc:
Thanks for the careful read — all three are fair, and I'm taking them: F1 (entry-level failures must not skip the blob pass), F2 (batch the legacy mark by id, fix the comment, add the quiet-window note), F3 (deterministic lock order by sorting the id arrays). Back to draft while I address them; I'll re-request when the fixes are up with CI green.

wyuc:
## Review round (`8bcb49a0`)

All three findings are addressed in three commits on top of the `main` merge: `2b2a584c`, `d7d82fad`, `8bcb49a0`. An independent adversarial pass was run against each round on real PostgreSQL 16 (deadlock and leak probes, hunk-revert checks on every new test); the second round's findings are why there are three commits rather than one, and they are described below because they changed the shape of the F2 fix.

**F1 — entry-level failures no longer skip the blob pass.** `collectPass` records a throw from the entry level (the marker probe *and* `entryLevelPass`) and re-raises it after the blob pass, the same record-and-throw-after pattern the marker refusal already used (`collector.ts`, the block above the blob candidate query). One thing the finding did not cover, found in review: if the blob pass itself throws while an entry-level failure is recorded, the blob failure is what is thrown (it is what stopped the pass) and the recorded entry-level failure now travels on it as a non-enumerable `entryLevelFailure` (`AssetCollectionEntryLevelFailure`, exported) — nothing is dropped and the thrown types are unchanged. The comment states the precedence. Test: `asset-lifecycle.test.ts` — a real lock failure inside the backfill's `FOR SHARE`; the eligible blob is deleted in the same pass and the pass then rejects with the entry-level error (fails with the fix reverted).

**F2 — the legacy mark is batched; the "no cycle" comment was wrong and is replaced.**
- `markLegacyEntries` now loops over bounded batches of *legacy* ids: each batch is its own `lockBoundedTransaction`, locks at most `batchSize` rows in ascending id order, marks them and stamps the unreferenced ones there, and the next batch starts where the predicate now begins — no cursor, monotonic (a lock timeout loses at most one batch's work and the mark resumes from the database), so a busy system cannot keep the legacy gate closed indefinitely.
- Review caught that fusing mark and stamp per batch dropped the parent's broader stamp over every committed-unstamped-unreferenced entry (a committed orphan became a permanent leak that also counted against quota). That breadth now lives in a standing bounded sweep (`stampUnreferencedSweep`): one `batchSize` batch per pass over committed, unstamped entries, ascending `FOR UPDATE` then a fresh-snapshot stamp, in-memory cursor that wraps; it runs only once the legacy gate is open (before the walk finishes, "no reference row" does not yet mean "no document names it"). An orphan becomes a release candidate within `ceil(entries / batchSize)` passes instead of never, which is strictly more than the one-time stamp did.
- Review also showed the cycle you described was still reachable after batching alone: a save's per-scope reference `INSERT` takes `KEY SHARE` in scope order *before* any ordered lock, so a two-scope save and the mark's ascending batch could still deadlock (reproduced: `40P01`, the collector was the victim). Fix: every reference-maintaining transaction now takes ONE ascending `FOR NO KEY UPDATE` over the union of the ids it will reference and the ids it will stop referencing, before its first write (`lockWriteEntries` in `references.ts`, called first in `syncDocumentAssetReferences` / `syncStageAssetReferences` / `removeDocumentAssetReferences`); the backfill does the same at `FOR KEY SHARE` per document (`lockBackfillEntries`). The commit and stamp statements no longer lock — the rows are already held. The old comment is replaced with the rule the guarantee rests on. Probes on the final head: mark vs two-scope save → no deadlock; two full saves whose scopes cross the same ids → both commit. Both are now PG contract tests (`pg-asset-store.pg.test.ts`), each verified to fail with the lock placement reverted.
- Upgrade guidance added to `docs/reference-server.md`: the one-time mark touches every legacy entry in bounded ascending batches; run it in a quiet window on a busy deployment; concurrent saves during the window may see a retryable `StorageLockUnavailableError`. The sweep's cost (one short transaction over one bounded batch of live rows per pass) is stated next to it.

**F3 — deterministic lock order.** Subsumed by the union lock above: a save acquires all its entry locks in one ascending statement, so two saves sharing ids queue instead of cycling (the residual you noted — two ordered sets in sequence — no longer exists, and the comment says what is now claimed). The reference `INSERT … SELECT … ORDER BY id` is kept as a second line of defence for a future caller that reaches it without the union lock. Unit test asserts exactly one `FOR NO KEY UPDATE` per write, parameterised with the sorted union, before the first `document_asset_refs` write.

**Verification.** CI 6/6 on `8bcb49a0`. Package suite with PG contracts locally: 1308 passed / 7 skipped (was 1293 / 7 at the last review; +15 tests, 0 regressions). `pnpm run check` and `pnpm run lint` clean. Part 2 (#1473) has this branch merged in at `b847c3c0`; its app-domain PG contracts and typecheck pass on that merge.

Ready for another look.

cosarah (approved):
Verified all three findings from my earlier comment are fixed:

- **F1** — entry-level failures are recorded and raised after the blob pass; when both levels fail, the blob failure takes precedence and carries the entry-level one as `entryLevelFailure`.
- **F2** — the legacy mark now runs in bounded, resumable batches; the incorrect "no cycle" comment is replaced; quiet-window guidance added.
- **F3** — every reference-maintaining write takes its entry locks in one ascending statement over the add∪drop union, and the backfill locks the document's union at KEY SHARE first.

Re-ran the full suite against PostgreSQL 16 locally: 1308 passed / 7 skipped; typecheck clean. The follow-up round (standing sweep, union locks, failure precedence) is solid work — thanks!

PR #1473, opened 2026-09-11
feat(persistence): turn on the server-owned asset lifecycle and release assets on course deletion (#1007 amendment, part 2)

### LINKED ISSUE #1394: [Task] Reclaim asset registry entries that no document references

wyuc:
Part 2 of the lifecycle amendment on #1007 (https://github.com/THU-MAIC/OpenMAIC/issues/1007#issuecomment-5638002396). Stacked on #1472 (part 1, `@openmaic/storage` 0.30.0); the base will move to `main` once that merges. Closes #1394 together with #1472.

Part 1 gave the server the asset entry lifecycle behind two options. This PR turns them on in the app and removes the client-side machinery they replace.

### What changed

- **The pairing is a property of the deployment, not a setting.** Every server-side `PgDocumentStore` the app builds (the persistence provider and the owner-bound store) passes `trackAssetReferences: true`, and the collector is built with `documentReferences: true`. There is no configuration in which one runs without the other. At startup, right after the schemas are ensured, the provider calls `declareAssetReferenceTracking()`, so a cold or upgraded database's entry level is eligible from the first collector pass; a failure there fails startup like a failed schema ensure.
- **Deleting a course releases its assets.** The app retires a course by a tombstone in `stage_meta` that cascades from the document row, so the document rows must stay. `OwnerBoundDocumentStore.deleteDocument` now calls `withdrawAssetReferences` inside the same transaction that writes the tombstone, on a store pinned to that transaction's connection (a fresh connection would contend for the rows the transaction already holds). A failure rolls the tombstone back; the second delete withdraws nothing and does not restart the grace period; a course saved again under a lifted tombstone gets its references back.
- **`ASSET_PENDING_TTL_MS`** (default 24 h, validated at startup like `ASSET_QUOTA_BYTES`), documented in `.env.example`, README and README-zh next to the collection grace.
- **Collector schedule** logs entry counts and backfill progress beside the blob count, and handles `AssetReferenceTrackingNotEnabledError` (now reachable only if the provider never initialised or a writer bypasses it) and `StorageLockUnavailableError` (retry next interval) distinctly without stopping the schedule.
- **Dead client code removed.** `lib/media/reclaim-stage-assets.ts` (the stage-scoped registry sweep that was never wired) becomes a 41-line `clearStageMediaCache` that keeps exactly the surviving behaviour: the stage's local `mediaFiles` rows are deleted, and `audioFiles` rows only when no surviving document references them. `collectStageAssetRefs` is now a pure document enumerator (every caller already passed empty rows). Stale comments describing the old sweep are corrected.
- **Tests.** A real-PG app-side contract (`tests/persistence/document-asset-references.pg.test.ts`, own schema, dropped before and after) covers deletion releasing references, idempotence, foreign-owner refusal, re-save under a lifted tombstone, rollback when the withdrawal throws or hits a lock, and a documentation case showing why a raw hard delete is not used (it loses the tombstone and keeps the references). `tests/persistence/asset-reference-backfill.pg.test.ts` builds an upgrade fixture before the provider exists and proves the first pass backfills and marks without raising the tracking error. Both run in the `Storage Contract (PostgreSQL 16)` job.

### Not in this PR

- The principal model (#1129 item 1, #1413).
- Anything in `packages/@openmaic/storage` (part 1).

### Verification

`pnpm check`, `pnpm lint` (0 errors, 18 pre-existing warnings), `tsc --noEmit` clean; app suite 7912 passed / 13 skipped with a database and 7886 / 39 skipped in the CI shape without one; the package contract suite followed by the app-domain step on a fresh database 1260 / 7 skipped then 14 passed with no leftover schema; `pnpm build` passes with and without `NEXT_PUBLIC_PERSISTENCE=1`. One adversarial review round with PostgreSQL probes (unconditional pairing, tombstone ordering against a racing autosave over fifteen iterations, rollback on withdrawal failure, schema isolation) found three low items, all fixed.


wyuc:
Note on the base: this PR is stacked on #1472, and the workflows only run for pull requests against `main`, so the base was pointed at `main` for one run (checks are attached to `4de40a8d` and test the merge of both parts) and then moved back to `feat/asset-lifecycle-server-owned` so the diff shows only part 2. It will be retargeted to `main` once #1472 merges.

cosarah (dismissed):
Approved. Verified locally: tsc clean, lint 0 errors (18 pre-existing warnings), all changed test suites pass, and both new PG contract files pass on a real PostgreSQL 16 in the CI order (package suite 1260 → app contracts, same database, no leftover schema — the schema isolation holds). The unconditional pairing claim is literally true in the code: all three `new PgDocumentStore` constructions set `trackAssetReferences: true` and the sole production `AssetCollector` sets `documentReferences: true`. The pinned-store pattern in `deleteDocument` is sound, and both rollback tests exercise real statement failure and real lock contention rather than mocks.

Three minor notes, none blocking:

1. **Provider memoization key mismatch on whitespace-padded `DATABASE_URL`** — the collector schedule and the shutdown hook pass `process.env.DATABASE_URL?.trim()`, while the persistence route (route.ts:317) and agent-runtime call sites pass the untrimmed value. With stray whitespace you'd get two memoized providers (two pools) in one process. Pre-existing inconsistency in shape, but the new `prepare()` call is the first place both spellings can occur at startup. Harmless in practice.

2. **The tracking alarm's first stated cause can't co-occur with the alarm line** — if provider initialization fails, `prepare()` throws and the generic failure branch logs; `collectPass` (and thus the `AssetReferenceTrackingNotEnabledError` alarm) never runs that tick. The alarm only fires when the provider *did* initialize but the marker is gone (external deletion or a non-tracking writer), so "the failed pass logged just above says why" describes a sequence that can't produce that line. Operator-guidance-level nit only.

3. **Rolling-upgrade exposure** (an old-binary writer continuing after the marker is declared) is inherent to part 1's marker design and documented in `reference-server.md`; this deployment restarts as one compose unit, so it's out of scope here. The unconditional pairing in this PR is exactly the "derive both from one setting" posture that doc recommends.

wyuc:
Thanks for the local verification. Both notes fixed in `1a8a6a27`:

1. `getServerPersistenceProvider` now keys its memo on the trimmed connection string and builds the pool from it, so trimmed and untrimmed callers converge on one provider; the call-site trims stay because they are the blank-vs-unset guard rather than a memo key. `route.test.ts` asserts a padded URL returns the same provider object with no second pool.
2. Agreed the first stated cause cannot co-occur with the alarm — and on inspection the second cannot either: the marker is written `INSERT … ON CONFLICT DO NOTHING` and nothing deletes it, so a non-tracking writer cannot remove one that is already there. The alarm now names the only thing left (the row is not in the database this collector reads: dropped, truncated, restored away, or two different databases), and the test asserts neither impossible cause appears in the message.
3. Out of scope here, agreed; the pairing being unconditional is exactly the posture the package doc recommends.

CI will run again when the base moves to `main` after #1472 merges.

wyuc:
`d1b327d9` merges the updated #1472 branch (which now carries `main` through #1500) into this one. One conflict in `.env.example`, where `OPENMAIC_CLASSROOMS_DIR` from `main` and `ASSET_PENDING_TTL_MS` from this PR were appended at the same spot; both blocks are kept. Nothing else changed. On the merged tree: typecheck and format clean, the three app-domain PG contract files pass (14 tests), and `tests/persistence` + `tests/media` pass (645 tests).

wyuc:
@cosarah #1472 landed as `e824117c`, so this PR's base is now `main` — and GitHub dismissed your approval on the base change. Nothing in the app-side diff changed since you approved: `a171182b` only merges `main` (the squashed part 1 plus #1499/#1500, which resolved cleanly) so the diff against `main` is the 30 app files you reviewed and nothing from the storage package. CI is running on this head against `main`; the app-domain PG contracts and typecheck pass locally on it. Could you re-approve when convenient? Thanks for the quick turnaround on both parts.

cosarah (approved):
Approve. Verified against the latest head (`a171182b`):

- Both new PG contract suites pass on a real `postgres:16` (same image as CI): 13/13 — deletion releases references under the tombstone, idempotent second delete, foreign-owner refusal, rollback on both a real in-transaction statement failure (table rename) and a real lock timeout, backfill on an upgraded database.
- The pinned-store transaction join is correct: the pass-through `withTransaction` makes `withdrawAssetReferences` join the delete's transaction, `SET LOCAL lock_timeout` is valid inside it, and the marker upsert inside `removeDocumentAssetReferences` is what the rollback test trips — the mechanism is real.
- Collector error triage matches the package semantics (`collector.ts` rethrows entry-level failures raw after the blob pass), so the `StorageLockUnavailableError` / tracking-refusal branches are accurate.
- `clearStageMediaCache` is behavior-preserving vs the never-wired sweep; no stale references to the removed API; `deleteFolder('remove')` reaches the same tombstone+withdrawal cascade.
- `tsc --noEmit` clean, eslint clean on changed files, affected unit suites green (1227 tests across media/persistence/store/document).

Non-blocking notes: the two-grace-period worst case is now correctly documented in the README, and the `released === false` branch keeping the tombstone is the right trade-off.

Issue #444, opened 2026-04-16
[Bug]: Multiple Issues on Mobile Version


rajibshrimant100-ctrl:
### Bug Description

When using OpenMAIC on mobile, several problems occur. Phone number registration shows an error, so I had to use email registration instead. Conversations sometimes do not appear after the teacher responds. Audio is delayed and loads much later than the text. The platform feels slow overall on mobile.

### Steps to Reproduce

1. Open OpenMAIC on mobile browser
2. Try registering with phone number
3. Start a classroom discussion with voice enabled
4. Wait for teacher response
5. Observe delayed audio and missing chats

### Expected Behavior

Phone registration should work correctly. Conversations should appear normally after the teacher responds. Audio should load at the same time as the text. The platform should feel smooth and responsiveness on mobile.

### Actual Behavior

Phone registration shows an error. Some conversations do not appear after the teacher responds. Audio is delayed and sometimes plays much later than the text. The website feels slow on mobile.

### Deployment Method

Other

### Browser

Chrome 147 on Android

### Operating System

Android

### Relevant Logs / Screenshots

```shell

```

wyuc:
Thanks for the report, and sorry it went so long without a reply. The app has changed a lot since this was filed, and there isn't enough here to reproduce it on current `main`. Closing for now; if it still happens on the latest version, please reopen (or open a new issue) with: self-hosted or the hosted demo, the version, browser console logs, and a separate description for each symptom.

PR #1477, opened 2026-09-12
fix(audio): reuse one narration element so mobile narration survives the first segment

### LINKED ISSUE #1474: fix(audio): narration goes silent after the first segment on mobile (per-segment new Audio element refused by autoplay policy)
### What happened

On mobile browsers (iOS Safari, Chrome on Android, and the in-app WebViews that wrap them), classroom narration plays the **first** segment and then goes silent for every following segment, while the player keeps advancing visually — scene changes, progress bar moves, no voice. Reloading and tapping Play sometimes restores one more segment, then it goes silent again. Desktop browsers rarely show this, which is why it reads as "works for me".

### Environment

- Deployment: self-hosted, Docker (`docker-compose`), `main` @ 2026-09-12
- Client: Chrome on Android / iOS Safari / WeChat in-app browser; desktop Chrome as the control
- TTS: server-side GLM (zh) narration, pre-generated WAV assets served through `/api/classroom-media/...`
- Autoplay setting in the player: **Auto-play ON**

### Steps to reproduce

1. Open a classroom with pre-generated narration on a **mobile** browser.
2. Tap **Play** once, then leave the page alone (Auto-play ON).
3. Listen: the first narration segment plays; from the second segment on there is no audio while the scene keeps advancing.
4. Refresh and tap Play again: sometimes one more segment is audible, then silence again.

### Expected

Narration for every segment plays without further user interaction after the initial gesture — the gesture that started playback should keep covering the rest of the lesson.

### Root cause

`lib/utils/audio-player.ts` creates a **new** media element for every narration line (`this.audio = new Audio()` inside `play()`), and the per-line `play()` is issued programmatically by the playback engine rather than from a user gesture. Mobile browsers only apply the page's user activation to elements that were activated by that gesture; a fresh element created minutes later is a new, un-blessed element, so `play()` rejects with `NotAllowedError` mid-lesson.

This is the condition already described in #651 ("an expected, recoverable condition (autoplay policy before the first user gesture...)"), but the fix in #652 only revoked the leaked blob URL. The engine still catches the rejection and **falls back to a reading-time timer** — which is exactly the reported symptom: the lesson walks on, the voice is dropped. The user-facing half of that bug has no mitigation, and `NotAllowedError` has no dedicated issue in this repo.

### Proposed fix

Keep one narration element for the whole document and bless it with the first real gesture, then hand later lines to that same element:

1. **One shared element** (`lib/utils/audio-autoplay.ts`) instead of `new Audio()` per line; it is unlocked on the first user gesture (muted, silent-source play) so subsequent `src` swaps on the same element keep playing.
2. **`playWithGestureRetry()`**: if a play is refused anyway, re-issue it on the next user gesture (up to a bounded timeout) instead of dropping the line into permanent silence.
3. `onended` is assigned rather than added, because a shared element would otherwise accumulate one listener per segment.
4. The discussion-TTS path (`lib/hooks/use-discussion-tts.ts`) goes through the same helper.

No UI strings are added, so nothing needs i18n. The touched files are outside `packages/@openmaic/*`, so no package version bump applies.

### Evidence

Instrumented both `HTMLMediaElement.prototype.play` and `window.Audio` in the classroom, clicked **Play once** with a real mouse event, then left the page running for 3 minutes (Chromium, default autoplay policy):

```
# before: play() rejected per segment, engine advances on the reading-time timer
play-FAIL | NotAllowedError | .../tts_s1_action_0nnx3PqX.wav
(no further play() attempts reach audio output; scene keeps advancing)
```

```
# after: one element, every segment plays, no further clicks needed
play-OK | .../tts_s1_action_...wav | els=1
play-OK | .../tts_s2_action_...wav | els=1
play-OK | .../tts_s3_action_...wav | els=1
play-OK | .../tts_s4_action_...wav | els=1
play-OK | .../tts_s5_action_...wav | els=1
play-OK | .../tts_s6_action_...wav | els=1   (8+ consecutive segments, 0 failures)
currentTime advancing per segment: 29.5s → 20.9s → 25.6s → …
```

Also verified on a real phone (Android Chrome, Auto-play ON): the lesson now narrates continuously through multiple scenes without touching the screen.

### Scope / not verified

- iOS Safari and other in-app WebViews were not directly instrumented — the mechanism is the same autoplay policy, so the shared-element path is expected to cover them, but that is an expectation, not a measurement.
- If a lesson is opened and the user never interacts at all, no audio can start: the retry helper then waits for the first tap, which is the best any page can do under these policies.

ttkm2023:
> 🤖 **AI-assisted PR** — written with an AI coding agent and reviewed by a second one (Claude Code) before opening. That review is why this PR is one file: an earlier draft also carried a gesture-retry helper and touched the discussion-TTS path, and both were split out (reasons below). I understand the change and stand behind it.

## Summary

On mobile browsers (iOS Safari, Chrome on Android, and the in-app WebViews that wrap them) classroom narration plays the **first** segment and then goes silent for every following segment, while the player keeps advancing visually. A refresh buys back at most one more segment.

The cause is structural. `lib/utils/audio-player.ts` creates a **new** media element for every narration line. Only the first line is covered by the user's own gesture (the tap on Play); every line after it is programmatic, on a brand-new element, so the autoplay policy refuses it with `NotAllowedError`. The playback engine catches that rejection (as #651 documents) and falls back to the reading-time timer — so the lesson walks on and the voice is gone for the rest of the session. #652 fixed the blob URL **leak** from the same rejection; the user-facing half stayed open, and nothing in this repo tracks it.

An element that a gesture has activated keeps playing afterwards, so the fix is to stop minting elements: **one element per player, handed every line**, stays playable for the rest of the lesson.

## Related Issues

Fixes #1474

Related to #651 / #652 — the same rejection, fixed there only for the leak. Related to #444 (mobile playback).

## Changes

`lib/utils/audio-player.ts` only:

- a per-player `element`, created on first use (`getAudioElement()`) and reused by every line, replaces `new Audio()` inside `play()`;
- `stopAudioElement()` releases the **line's** state instead of the element: `onended` is cleared, `src` is removed and `load()` is called, so a stopped line cannot report speech that is no longer playing and a revoked object URL cannot keep the narration buffer alive until the next `play()`. The element itself is deliberately kept — discarding it is what made each following line a fresh, un-activated element;
- `onended` is **assigned** rather than added with `addEventListener`: the element now outlives a single line, so a listener would accumulate once per segment and call the engine back several times for one line.

`tests/audio/audio-player-element-reuse.test.ts` (new) stubs the mobile policy itself — the first element plays, every element created after it is refused — and asserts lines 2 and 3 still play, that one element is created for three lines, that the end callback is a single assigned handler, that `stop()` releases the line's state, and that a non-autoplay failure still surfaces. **The first two tests fail on `main` and pass with this change** (output below), which is the point: they encode the bug rather than the implementation.

**Deliberately not in this PR** (both from the AI review, and both separable):

- a `playWithGestureRetry` helper that would wait for the next user gesture after a refusal. It has to not hold the engine's `play()` promise open (the reading-timer fallback has a 2 s floor and a typical dwell of 3–10 s, so a long wait turns "silent but advancing" into "silent and stalled"), and it needs cancellation wired to the existing `requestToken`/abort pattern so a pending retry cannot start narration over a paused lesson. That is its own PR, with its own tests, and it is not needed for the fix above: the refusal disappears once the element is reused.
- the discussion-TTS path (`lib/hooks/use-discussion-tts.ts`) — a different player with its own queue lifecycle that does not share the element-reuse half.

No user-facing strings were added (no i18n debt), and nothing under `packages/@openmaic/` changed, so no version bump applies.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Self-host, open a classroom with pre-generated narration in a **mobile** browser (or device emulator) with **Auto-play ON**.
2. Tap **Play** once, then leave the page alone.
3. Before: the first segment is audible, every later segment is silent while the scene advances.
4. After: narration continues through consecutive segments with no further interaction.

- **The new tests fail on `main` and pass with the change** (shown below) — the regression is encoded, not just described.
- **Live deployment, instrumented**: patched `HTMLMediaElement.prototype.play` and `window.Audio` in the classroom page of a self-hosted instance, clicked **Play once** with a real (trusted) mouse event, then left it running for 3 minutes under the default Chromium autoplay policy: **1 element created, 8+ consecutive `play-OK`, 0 `play-FAIL`**, `currentTime` advancing line by line (29.5s → 20.9s → 25.6s → …) with the scene following. Same page and instrument before the change: the per-line `play()` rejected with `NotAllowedError` and the scene advanced with no audio.
- **Real device**: Android Chrome, Auto-play ON — a 16-scene lesson narrated continuously without touching the screen.
- Desktop behaviour is unchanged by design: the element is reused there too, and a `play()` that succeeds never sees a refusal.
- **Not verified**: iOS Safari and other in-app WebViews were not instrumented. They share the policy and the mechanism, so they are expected to be covered — that is an expectation, not a measurement.

### Evidence

New tests against `main` (no patch) — the bug, reproduced:

```
× keeps narration audible after the first line, when the policy refuses new elements
× reuses that one element for every line
× reports the end of each line through a single assigned handler
× releases the line's state when playback is stopped
× still surfaces a failure that is not an autoplay refusal
Test Files  1 failed (1)     Tests  5 failed (5)
```

The same file with the patch:

```
Test Files  1 passed (1)     Tests  5 passed (5)
```

Live instrumentation, before → after (one initial Play click, 3-minute window):

```
play-FAIL|NotAllowedError|.../tts_s1_action_0nnx3PqX.wav        (before: silence for the rest of the lesson)
play-OK|...tts_s1_link...wav|els=1                              (after: 8+ consecutive lines)
play-OK|...tts_s2_link...wav|els=1
play-OK|...tts_s3_link...wav|els=1   …
```

Local checks (Node 24 / pnpm 10):

```
$ npx prettier --check …   # clean
$ npx eslint …             # clean
$ npx tsc --noEmit         # clean
$ npx vitest run tests/audio tests/playback
  Test Files  43 passed (43)      Tests  329 passed (329)
```

- [x] Manually tested locally

## Checklist

- [x] I have performed a self-review of my code (plus an AI review; its findings shaped the scope of this PR)
- [x] I have added/updated documentation as needed (the reasoning lives inline; no user-facing docs change)
- [x] My changes generate no new warnings
- [x] I have added tests that prove the fix works — and that fail without it
- [x] New and existing unit tests pass locally

### DIFF
--- a/lib/utils/audio-player.ts
+++ b/lib/utils/audio-player.ts
@@ -32,6 +32,17 @@ async function resolveBytes(audioId: string): Promise<Blob | null> {
  */
 export class AudioPlayer {
   private audio: HTMLAudioElement | null = null;
+  /**
+   * The element narration plays through, created once per player.
+   *
+   * Playback must not mint a new element per line: mobile autoplay policies
+   * only cover an element that a user gesture has reached, so a fresh element
+   * gets its programmatic `play()` refused with NotAllowedError mid-lesson and
+   * the narration goes silent from the second line on, while the lesson keeps
+   * advancing on the reading-time fallback. One element, handed every line,
+   * stays playable for the rest of the lesson.
+   */
+  private element: HTMLAudioElement | null = null;
   private onEndedCallback: (() => void) | null = null;
   private muted: boolean = false;
   private volume: number = 1;
@@ -66,11 +77,39 @@ export class AudioPlayer {
     if (this.blobUrl === blobUrl) this.blobUrl = null;
   }
 
+  /**
+   * The narration element, created on first use and reused by every line.
+   * Deliberately not dropped by stopAudioElement(): discarding it is what made
+   * each following line a fresh element whose play() the policy refuses.
+   */
+  private getAudioElement(): HTMLAudioElement {
+    if (!this.element) {
+      this.element = new Audio();
+      this.element.preload = 'auto';
+    }
+    return this.element;
+  }
+
   private stopAudioElement(): void {
     if (this.audio) {
       this.audio.pause();
-      this.audio.currentTime = 0;
+      try {
+        this.audio.currentTime = 0;
+      } catch {
+        /* Not seekable yet (nothing loaded): there is nothing to rewind. */
+      }
       this.audio = null;
+    }
+    // The element outlives this line now, so the line's own state has to be
+    // released with it: a stale ended handler would report speech that is no
+    // longer playing, and a revoked object URL still loaded keeps the narration
+    // buffer alive until the next play() replaces it.
+    if (this.element) {
+      this.element.onended = null;
+      if (typeof this.element.removeAttribute === 'function') {
+        this.element.removeAttribute('src');
+      }
+      this.element.load?.();
     }
     // Stop or replacement before natural end must not leak the fetched
     // narration: the element is dropped here, so its URL is released with it.
@@ -141,8 +180,10 @@ export class AudioPlayer {
       this.stopAudioElement();
       if (requestToken !== this.requestToken) return false;
 
-      // Create audio element
-      this.audio = new Audio();
+      // Reuse this player's element rather than creating one per line: a new
+      // element's programmatic play() is what mobile autoplay policies refuse,
+      // which left the narration silent from the second line on.
+      this.audio = this.getAudioElement();
 
       // Set audio source
       const blobUrl = blob ? URL.createObjectURL(blob) : undefined;
@@ -156,10 +197,13 @@ export class AudioPlayer {
       this.audio.playbackRate = this.playbackRate;
 
       // Set ended callback
-      this.audio.addEventListener('ended', () => {
+      // Assignment rather than addEventListener: the element is reused, so a
+      // listener would otherwise accumulate once per segment and fire the
+      // callback several times for one line.
+      this.audio.onended = () => {
         this.releaseBlobUrl(blobUrl);
         this.onEndedCallback?.();
-      });
+      };
 
       // Play. If play() rejects (autoplay policy, decode error, interrupted
       // load) the 'ended' listener never fires, so revoke the blob URL here to
--- a/tests/audio/audio-player-element-reuse.test.ts
+++ b/tests/audio/audio-player-element-reuse.test.ts
@@ -1,0 +1,155 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// Mock the IndexedDB layer so importing AudioPlayer doesn't pull in Dexie.
+const getMock = vi.fn();
+vi.mock('@/lib/utils/database', () => ({
+  db: { audioFiles: { get: getMock } },
+}));
+
+/** A rejection shaped like the browser's autoplay refusal. */
+function notAllowed(): Error {
+  return Object.assign(new Error('play() failed because the user did not interact'), {
+    name: 'NotAllowedError',
+  });
+}
+
+/** Stub URL.createObjectURL/revokeObjectURL while keeping `new URL(...)` working. */
+function stubObjectUrl() {
+  let next = 0;
+  const createObjectURL = vi.fn(() => `blob:fake-url-${++next}`);
+  const revokeObjectURL = vi.fn();
+  class URLStub extends URL {}
+  Object.assign(URLStub, { createObjectURL, revokeObjectURL });
+  vi.stubGlobal('URL', URLStub);
+  return { createObjectURL, revokeObjectURL };
+}
+
+class AudioStub {
+  play = vi.fn(async () => {});
+  pause = vi.fn();
+  load = vi.fn();
+  removeAttribute = vi.fn();
+  addEventListener = vi.fn();
+  removeEventListener = vi.fn();
+  volume = 1;
+  muted = false;
+  preload = 'auto';
+  defaultPlaybackRate = 1;
+  playbackRate = 1;
+  currentTime = 0;
+  duration = 0;
+  src = '';
+  paused = true;
+  readyState = 0;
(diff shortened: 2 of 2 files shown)


LING-6150:
The implementation and lifecycle review look clean, and the relevant audio/playback suites pass.
Before approval, could you attach a raw before/after trace from the affected mobile browser,  preferably iOS Safari or a WKWebView,  and clarify whether tts_s1 in the supplied log was the first attempted narration line? The issue says the first segment played successfully, while that log appears to show the first segment failing.
The unit coverage validates element reuse and cleanup, but it cannot establish the browser-specific autoplay behavior by itself. If iOS/WebView cannot be verified here, please narrow the PR’s stated scope accordingly rather than claiming the full mobile issue is resolved.
Non-blocking: updating the older Audio test stubs to implement load() and removeAttribute() would make them closer to the production element contract.

wyuc (approved):
Approving. Reusing one narration element across lines addresses the mobile autoplay restriction behind #1474, and the stop path clears `onended`, the source and the line's blob URL. The new tests reproduce the failure: restoring a new `Audio()` per line fails all five with `NotAllowedError`. Segment transitions, pause/resume, jumps, scene switches, stop and teardown behave correctly, and the change doesn't overlap the recent playback fixes (#1510, #1521).

Notes for follow-ups, none blocking: iOS Safari/WKWebView wasn't traced on a real device, so a quick confirmation there would be welcome; discussion narration (`use-discussion-tts.ts`) still creates an `Audio` per segment and likely has the same mobile issue; and clearing `onended` when `play()` rejects would avoid a stale handler. Thanks!

PR #1478, opened 2026-09-12
fix: bound model discovery response body timeout

### LINKED ISSUE #1449: Model discovery timeout does not cover response body consumption

LeoParkerOu:
## Summary

Fix model discovery requests that receive response headers but stall while reading the response body.

The existing 15-second timeout was cleared as soon as `fetch()` returned a `Response`, so `response.json()` or `response.text()` could remain pending indefinitely.

## Related Issues

Fixes #1449

## Changes

- Keep the timeout active through model-list response-body consumption.
- Retry one transport or timeout failure.
- Preserve the existing per-candidate timeout budget.
- Preserve 404/405 endpoint fallback behavior.
- Preserve redirect rejection and terminal HTTP error handling.
- Abort stalled requests and release their connections.
- Add regression coverage for stalled headers, stalled bodies, retries, malformed JSON, terminal HTTP errors, fallback candidates, and a real stalled HTTP response.

## Type of Change

## Verification

- 34 focused model discovery and route tests passed.
- Real loopback HTTP test passed: a partial JSON response was aborted after the timeout and the retry succeeded.
- Prettier passed.
- ESLint passed for all changed files.
- Full TypeScript checking still reports existing repository baseline errors; the error output is unchanged by this PR.

### DIFF
--- a/lib/server/model-fetch.ts
+++ b/lib/server/model-fetch.ts
@@ -7,8 +7,6 @@
  * suffix-strip fallback) and try each until one returns a model list.
  */
 
-import { fetchWithTimeout } from './fetch-with-timeout';
-
 /** A model id discovered from a provider's /models endpoint. */
 export interface FetchedModel {
   id: string;
@@ -33,6 +31,13 @@ const KNOWN_COMPAT_SUFFIXES = [
 ] as const;
 
 const FETCH_TIMEOUT_MS = 15_000;
+// Preserve the existing per-attempt allowance, with one retry and a finite
+// budget shared by every candidate and attempt in a discovery operation.
+const DISCOVERY_TIMEOUT_MS = 2 * FETCH_TIMEOUT_MS;
+
+function discoveryTimeout(): DOMException {
+  return new DOMException('Model discovery timed out', 'TimeoutError');
+}
 
 /** Whether the URL's last path segment is an OpenAI-style version segment `/v{N}`. */
 function endsWithVersionSegment(url: string): boolean {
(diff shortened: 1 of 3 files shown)


CXuPercy (approved):
/tmp/pr1478-approve.md

CXuPercy:

wyuc (approved):
Approving. The deadline now covers the whole response (headers and body) for each attempt, stalled requests are aborted and their connections released, retries are limited to one transport/timeout failure across all candidates, and terminal HTTP errors (401/403/5xx) and malformed JSON are never retried. Redirect rejection and the 404/405 fallback are unchanged. The fake-timer cases plus the real stalled-HTTP test cover the stall-in-headers, stall-in-body, slow-but-successful, stalled-error-body, and shared-budget-across-fallbacks paths. Thanks!

Minor follow-up for later: with this change `lib/server/fetch-with-timeout.ts` has no remaining callers.

PR #1484, opened 2026-09-13
fix: enable Pro workbench flag in Docker builds; allow non-TLS localhost cookie

nqthiep:
## Summary

Two fixes discovered while self-hosting OpenMAIC v1.0.0 with Docker Compose (`--profile server-persistence`) over plain HTTP on localhost:

1. Expose `NEXT_PUBLIC_PRO_WORKBENCH_ENABLED` as a Docker build arg so the Pro Agent Workbench can be enabled in Docker deployments.
2. Add a `COOKIE_SECURE` opt-out for the anonymous-owner cookie so deployments without TLS (e.g. `http://localhost`) work in Safari.

### 1. Pro Workbench cannot be enabled in Docker deployments

`NEXT_PUBLIC_PRO_WORKBENCH_ENABLED` is the client-visible gate for the Pro workbench entry (see `lib/config/feature-flags.ts#isProWorkbenchEnabled`), and Next.js inlines `NEXT_PUBLIC_*` values at build time.

The Docker image (`Dockerfile`) already declares build args for `NEXT_PUBLIC_PERSISTENCE`, `NEXT_PUBLIC_PERSISTENCE_TOKEN`, `NEXT_PUBLIC_MAIC_EDITOR_ENABLED`, `NEXT_PUBLIC_ENABLE_PPTX_IMPORT` and friends — but **not** for `NEXT_PUBLIC_PRO_WORKBENCH_ENABLED`. `.dockerignore` also excludes `.env*`, so the flag cannot reach `pnpm build` through the build context either.

Result: a Docker deployment with `OPENMAIC_AGENT_RUNTIME_ENABLED=true` and a database still has no way to turn on the workbench entry — the flag stays off in the compiled bundle.

### 2. Scene saves fail with 403 in Safari on `http://localhost`

Production builds always attach `Secure` to the `anonymous_id` owner cookie:

```ts
const secure = process.env.NODE_ENV === 'production' ? '; Secure' : '';
```

Safari refuses to store `Secure` cookies served over plain `http://localhost` (unlike Chromium/Firefox, it does not special-case localhost). The cookie never lands in the jar, so:

- every request mints a **fresh anonymous owner**,
- owner-scoped document writes (`PUT /api/persistence/documents/{stageId}`) are rejected with `403` (`decideDocumentAccess` owner mismatch),
- scenes never persist, and generation appears to hang after the first scene (the stage row is created, but `document_scenes` stays empty).

Verified end-to-end with Playwright:

| Browser engine | Before fix | After fix |
| --- | --- | --- |
| WebKit (Safari) | cookie never stored; `PUT /documents/… -> 403`; 0 scenes saved | cookie stored; full course generated; 9 scenes saved |
| Chromium | works (localhost special-case) | works |

### 1. `Dockerfile` / `docker-compose.yml`

Add `NEXT_PUBLIC_PRO_WORKBENCH_ENABLED` as a build arg in the builder stage (ARG + ENV pair, mirroring the existing flags) and pass it through from `docker-compose.yml` (`${NEXT_PUBLIC_PRO_WORKBENCH_ENABLED:-}`), so the flag can be supplied via shell env or a Compose `.env` file.

### 2. `lib/server/agent-runtime/owner.ts`

Gate the `Secure` attribute behind an explicit opt-out, exposed once so every
place that mints the cookie agrees:

```ts
export function anonymousCookieSecure(): boolean {
  return process.env.NODE_ENV === 'production' && process.env.COOKIE_SECURE !== '0';
}
```

- Default behavior is unchanged: production builds still send `Secure`.
- `COOKIE_SECURE=0` is opt-in for plain-HTTP / trusted-network deployments; the
  Server Action mint in `lib/workbench/workspace-actions.ts` now shares that
  decision instead of hard-coding the production check.
- Documented in `.env.example` (including the security cost of dropping `Secure`
  and the fact that only the exact value `0` disables it).

## Changes

| File | Change |
| --- | --- |
| `Dockerfile` | Add `ARG NEXT_PUBLIC_PRO_WORKBENCH_ENABLED` + matching `ENV` in builder stage |
| `docker-compose.yml` | Pass `NEXT_PUBLIC_PRO_WORKBENCH_ENABLED` through as a build arg |
| `lib/server/agent-runtime/owner.ts` | `COOKIE_SECURE=0` opt-out for the anonymous-owner cookie |
| `.env.example` | Document `COOKIE_SECURE` |

## Testing

- Built the image (`docker compose --profile server-persistence build`) and ran the full stack (app + PostgreSQL).
- Reproduced the 403 scene-save failure with Playwright WebKit against `http://localhost`, then confirmed the fix: cookie persists, generation runs to completion, all scenes land in `document_scenes`.
- Chromium control run passes before and after (no regression).
- `OPENMAIC_AGENT_RUNTIME_ENABLED=true` + `NEXT_PUBLIC_PRO_WORKBENCH_ENABLED=true` now reachable: `/workspace` returns 200 and `/api/agent/sessions` is served in the Docker build.

## Notes

- The Pro workbench still requires the agent runtime (`OPENMAIC_AGENT_RUNTIME_ENABLED=true` + `DATABASE_URL` + a `maic-agent-driver` entry in `MODEL_ROUTES`); this PR only unblocks the build-time flag.
- If you'd prefer a different mechanism for (2) — e.g. deriving `Secure` from the request scheme instead of an env var — happy to adjust; the env opt-out was chosen because `resolveRequestOwnerId` currently only receives headers, not the request URL.

## Checklist

- [x] No new dependencies
- [x] Default behavior unchanged unless the new flags are set
- [x] Documentation added (`.env.example`)
- [x] Verified locally with Docker Compose + Playwright (WebKit & Chromium)
- [ ] Unit tests — n/a (deployment-config + cookie-attribute change; verified end-to-end instead)


wyuc (changes requested):
Thanks — both fixes are real and I reproduced them: the build arg is wired compose → `ARG` → `ENV` → `next build` and the value is genuinely inlined into the compiled server and edge bundles (no `.js` under `.next` still contains the variable name, defaults stay off), and the `COOKIE_SECURE=0` opt-out is env-only, defaults to `Secure` in production, and leaves every other cookie attribute byte-identical. One blocker and two small follow-ups before merging.

**1. The opt-out misses the second place that mints the same cookie.** `lib/workbench/workspace-actions.ts:27` (the `'use server'` action path) re-implements the `anonymous_id` mint and still does `secure: process.env.NODE_ENV === 'production'` unconditionally. With `COOKIE_SECURE=0` on plain HTTP, a Server Action that runs before any `/api/agent/*` request (e.g. deleting a workspace session) mints a `Secure` cookie, Safari drops it, and the action runs under an ephemeral owner — the same "fresh owner every request" failure, just on a different entry point. Verified with a probe: route path emits no `Secure`, action path still does. Suggested fix: extract the decision once in `owner.ts`, e.g. `export function anonymousCookieSecure() { return process.env.NODE_ENV === 'production' && process.env.COOKIE_SECURE !== '0'; }`, and use it in both modules.

**2. Please add the regression test.** `tests/agent-runtime/owner.test.ts` already pins the `Set-Cookie` attributes; reverting your `owner.ts` change leaves it 6/6 green, so nothing would catch a future edit that drops the opt-out. A symmetric case next to the existing production test is enough:

```ts
it('omits Secure when COOKIE_SECURE=0', () => {
  vi.stubEnv('NODE_ENV', 'production');
  vi.stubEnv('COOKIE_SECURE', '0');
  const responseHeaders = new Headers();
  resolveRequestOwnerId(new Request('http://localhost/agent'), responseHeaders);
  expect(responseHeaders.get('set-cookie')).toMatch(/^anonymous_id=[0-9a-f-]+; Path=\/; HttpOnly; SameSite=Lax; Max-Age=2592000$/i);
});
```

**3. `.env.example` wording.** Two things worth stating next to the flag: the security cost (without `Secure` the owner cookie travels in the clear and can be replayed by anyone on-path, so this is only for trusted networks / local use — the adjacent `PERSISTENCE_ALLOW_INSECURE_DEV_AUTH` block is a good model), and the accepted value: only the exact string `0` disables it, so `COOKIE_SECURE=false` silently keeps `Secure`. Either document that or accept `0`/`false`/`no` case-insensitively like the other boolean flags. The comment in `owner.ts:26-28` ("never forces Secure off") also reads backwards for the `=0` case.

Small PR, right direction — should be quick to close out.

nqthiep:
All three addressed in ef08331.

**1. Second mint site (blocker).** Reproduced your probe and fixed it: `anonymousCookieSecure()` is now exported from `owner.ts` and `lib/workbench/workspace-actions.ts` uses it instead of the hard-coded production check, so the route path and the Server Action path read the same decision.

```ts
export function anonymousCookieSecure(): boolean {
  return process.env.NODE_ENV === 'production' && process.env.COOKIE_SECURE !== '0';
}
```

**2. Regression test.** Added the symmetric case next to the existing production test in `tests/agent-runtime/owner.test.ts` (same assertion shape you drafted). Ran your mutation check in reverse: dropping the opt-out from `anonymousCookieSecure()` fails that test and leaves the other 6 green.

**3. `.env.example` wording + comment.** The block now states both the accepted value (exact `0`; `false`/`no` silently keep `Secure`) and the security cost (without `Secure` the cookie travels in the clear and is replayable on-path — trusted network / local only, referencing the adjacent `PERSISTENCE_ALLOW_INSECURE_DEV_AUTH` block). The comment you flagged as reading backwards is gone: the decision now lives in the documented `anonymousCookieSecure()` docstring, which states the opt-out directly.

`tsc --noEmit` clean. One note for honest coverage: `tests/workbench` shows 4 failures in `workspace-rail-session-rename.test.ts` when the whole directory runs, but the same 4 fail identically on a clean checkout of this branch with my changes stashed — pre-existing cross-file flake, not from this PR (it passes 5/5 in isolation).

wyuc (approved):
Round 2 verified: both mint sites now share `anonymousCookieSecure()`, the Server Action path emits `secure:false` under `COOKIE_SECURE=0`, the regression test goes red when the opt-out is reverted, and the docs state the security cost and the accepted value. One optional follow-up: the regression test pins the route path only, so reverting the Server Action site alone would not fail it; a second assertion through `workspace-actions` would close that. Merging.

PR #1485, opened 2026-09-13
fix(classroom): keep transient load failures off the not-found card

### LINKED ISSUE #1450: Classroom surface renders a transient server error as the terminal "this course does not exist" card
## What happens

`components/classroom/ClassroomSurface.tsx` collapses any completed load that did not land a stage into `outcome = 'unavailable'`, and renders that as `notFound`. `lib/classroom/load-classroom.ts` turns a non-OK `/api/classroom` response into `null`, and the stage-meta sidecar independently reports `'unavailable'` for 5xx or timeout. So a course that exists, whose document store returns null (share-store or browser-store course), plus a momentary 503 from `/api/classroom` and `/api/stage-meta/<id>`, produces the not-found card: a factual claim with no retry affordance, only "back to home". The pane variant reaches the same claim after its ~31 s backoff.

This predates #1399 (the standalone page already went straight to not-found when a load completed without a stage); #1399 made the state machine explicit, which is what makes it easy to see now. Noted during the #1399 round-2 review.

## Expected

Keep "unavailable because of an error" distinct from "absent" (positive 404/410), and either keep probing or give the terminal card a retry when it was reached without a positive answer. The `notFound` docstring in `ClassroomSurface.tsx` ("nothing here to retry") should hold true once fixed.

## Related

- Warm-store variant of the same gap: when the pane already holds the course and the server has since tombstoned it, `loadFromStorage` short-circuits and the tombstone signal is stage-meta alone; a stage-meta timeout or 5xx then renders the stale classroom as loaded. Same fix direction: treat `'unavailable'` on a warm hit as non-terminal.
- #1396, #1399

Duang777:
## Summary

A momentary 5xx / network failure from `/api/classroom` was collapsed into `null`, so `ClassroomSurface` treated it like a missing course and showed the terminal **Course not found** card (no retry). This PR classifies classroom fetch answers the same way stage-meta already does (`found` / `absent` / `unavailable`) and only renders not-found on a positive 404/410 miss.

## Related Issues

Fixes #1450

## Changes

- `fetchClassroomFromApi` returns `found` / `absent` / `unavailable` instead of `ClassroomPayload | null`
- `runClassroomLoad` surfaces that outcome so the UI can branch
- `ClassroomSurface` page: outage → error card with Retry; 404/410 → not-found
- `ClassroomSurface` pane: both gaps still use the bounded availability retry; exhaustion after an outage shows Retry, exhaustion after absence shows not-found
- i18n: `classroom.loadUnavailable` / `loadUnavailableDesc` in all 12 locales
- Regression tests for the fetch seam and loader outcomes

## Type of Change

### Steps to reproduce / test

1. Open a real course id with the document store returning null and `/api/classroom` forced to 503 (or disconnect briefly)
2. Confirm the UI shows a load error with **Retry**, not "Course not found"
3. Force `/api/classroom` 404 for an unknown id → not-found card remains
4. Pane: 503 during availability probing still retries, then lands on Retry rather than claiming the course does not exist

- Red → green on `tests/classroom/classroom-fetch-outcome.test.ts` (503/network → `unavailable`, 404/410 → `absent`)
- `pnpm exec vitest run tests/classroom/classroom-fetch-outcome.test.ts tests/classroom/load-classroom.test.ts` → **47 passed**
- `pnpm check:i18n-keys` → passed
- Did **not** drive a live workbench pane against a real 503 (no server-backed deployment in this environment)

### Evidence

- [x] Focused unit tests red → green
- [ ] CI passes
- [ ] Manually tested locally

## Checklist

## Notes

Related but separate from #1399 / #1396 (tombstone 410 fast-fail). This change only stops transient failures from being labeled as absence; it does not add the 410 contract.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/components/classroom/ClassroomSurface.tsx
+++ b/components/classroom/ClassroomSurface.tsx
@@ -66,7 +66,7 @@ import { isServerBackedMediaPersistence } from '@/lib/persistence/media-persiste
 
 const log = createLogger('Classroom');
 
-type ClassroomLoadOutcome = 'loaded' | 'unavailable' | 'failed' | 'cancelled';
+type ClassroomLoadOutcome = 'loaded' | 'unavailable' | 'absent' | 'failed' | 'cancelled';
 
 // stage_link can become visible shortly before its document. Probe only that
 // explicit availability gap, with a small bounded backoff; media conversion
@@ -81,12 +81,12 @@ export function ClassroomSurface({
   const { loadFromStorage } = useStageStore();
   const loadedClassroomId = useStageStore((s) => s.stage?.id ?? null);
   const { t } = useI18n();
-  // The retry loop below reads the message after async gaps, so it must see
-  // the CURRENT translation (a locale switch may have happened since mount).
-  // Written in an effect, not during render.
-  const notFoundMessageRef = useRef(t('classroom.notFound'));
+  // The retry / error paths below read the message after async gaps, so they
+  // must see the CURRENT translation (a locale switch may have happened since
+  // mount). Written in an effect, not during render.
+  const loadUnavailableMessageRef = useRef(t('classroom.loadUnavailable'));
   useEffect(() => {
-    notFoundMessageRef.current = t('classroom.notFound');
+    loadUnavailableMessageRef.current = t('classroom.loadUnavailable');
   }, [t]);
 
   const [loading, setLoading] = useState(true);
@@ -120,10 +120,9 @@ export function ClassroomSurface({
     async (isEffectCurrent: () => boolean = () => true): Promise<ClassroomLoadOutcome> => {
       const loadToken = claimStageSceneLoadToken();
       const isCurrent = () => isEffectCurrent() && isCurrentStageSceneLoadToken(loadToken);
-      let outcome: ClassroomLoadOutcome = 'loaded';
 
       try {
-        await runClassroomLoad({
+        const loadResult = await runClassroomLoad({
           classroomId,
           loadToken,
           isCurrent,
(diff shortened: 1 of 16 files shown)


wyuc (changes requested):
Thanks for picking up #1450 — the three-outcome seam is the right shape and I verified the important parts: the classification (`found` / `absent` / `unavailable`) is correct, a 5xx or network error ends in a retryable state on both the legacy `/api/classroom` path and the server-backed document path, Retry really re-fetches, and the availability schedule stays bounded (no new polling). The new unit tests are also genuinely sensitive (two mutations of the fix went red). Two blockers and one scope point.

**1. The pane now spins forever on a positive 404 instead of showing the not-found card.** `ClassroomSurface` is mounted in exactly one place, `WorkspaceClassroomPane.tsx:168`, always as `variant="pane"`. After the availability schedule is exhausted on `'absent'`, `loadUntilAvailable` sets `loading=false` and `notFound=true` but never sets `error` (only `'unavailable'` does). The render guard at `ClassroomSurface.tsx:436` is evaluated first:

```tsx
{loading || (variant === 'pane' && !error && loadedClassroomId !== classroomId) ? ( /* spinner */ ) : notFound ? ( /* card */ ) : ...}
```

With `error=null` and `loadedClassroomId=null`, the pane branch is `true` and the not-found card at `:443` is unreachable; the manual Retry path (`:481-487`) has the same problem. Before this PR the pane showed "Course not found" with a working Retry after ~31 s; after it, "Loading classroom…" forever. Suggested fix: `variant === 'pane' && !error && !notFound && loadedClassroomId !== classroomId`, plus a surface-level regression test that drives the pane through exhaustion for a 404 (expect the `classroom-not-found` test id) and a 503 (expect the Retry card). There is no DOM harness in the repo, so a small pure extract of the render decision is fine.

**2. CI is red on Prettier.** `pnpm check` flags `tests/classroom/load-classroom.test.ts` (the three `vi.fn().mockResolvedValue(...)` chains at :236, :297, :426 are split across lines). `pnpm format` fixes it.

**3. The standalone `/classroom/[id]` route still ignores the outcome.** `app/classroom/[id]/page.tsx:73-99` keeps its own copy of the load and does `await runClassroomLoad(...)` without reading the result, so on a 5xx it renders a null stage and on a 404 it never shows a not-found card — the `variant="page"` behaviour described in the PR body is unmounted code. Since #1450 is about the classroom surface as a whole, please apply the same three-outcome branching there (or scope the description to the pane).

Non-blocking:

- `classroom.loadUnavailableDesc` is added to all 12 locales but nothing renders it (the outage card shows `common.errorPrefix + error`). Either render it or drop the key.
- Every non-404/410 status is classified as a transient outage, including 400 from the route's id validation, which yields a Retry that can never succeed. Treating other 4xx (except 408/425/429) as a positive miss would be more honest. Also note the stage-meta client maps only 404 → `absent` and 410 → `unavailable`, so "mirrors stage-meta" is not quite accurate.

Appreciate the thorough PR description; with the guard fix and formatting this is close.

Duang777:
@wyuc Thank you for the detailed review. I pushed ebcbb9f6 and synced the branch with the current main: the pane render decision now gives exhausted not-found/error states priority over the stale-stage loading guard, the standalone route uses the shared three-outcome surface, the Prettier failure is fixed, the unavailable description is rendered in all 12 locales, and 400/404/410/422 are treated as definitive misses while auth/rate-limit/server/transport failures stay on the recoverable error path; manual Retry also refreshes ownership and is guarded by a monotonic load epoch plus the stage load token to prevent stale A-B-A completions. The 83 focused classroom tests, Prettier, ESLint, TypeScript, and i18n checks pass locally, and all five GitHub checks for ebcbb9f6 are now green; please re-review the updated branch.

wyuc (approved):
Re-reviewed at bef7a385: all three round-1 blockers are fixed. The render decision is extracted into resolveClassroomSurfaceView and the pane guard now carries !error && !notFound, so an exhausted 404 shows the not-found card (and manual Retry too) instead of spinning forever; the standalone /classroom/[id] route renders the shared three-outcome surface; Prettier is clean. loadUnavailableDesc is now rendered in all 12 locales, and 400/404/410/422 are definitive misses while auth/rate-limit/server/transport failures stay recoverable. The monotonic load epoch plus the stage load token hold against stale A-B-A completions. A mutation removing !notFound from the pane guard turns the suite red, so the exact spinner-on-404 regression is now caught. LGTM.

PR #1486, opened 2026-09-13
fix: pass reasoning_content back to DeepSeek in thinking mode (multi-turn)

nqthiep:
## Summary

Fix multi-turn failures on DeepSeek with thinking mode enabled. DeepSeek rejects thinking-mode requests whose assistant-history messages lack `reasoning_content`:

```
400 invalid_request_error:
"The `reasoning_content` in the thinking mode must be passed back to the API."
```

This made the Pro Agent Workbench (`maic-agent-driver`, tool-using, multi-turn) unusable: turn 1 succeeded, turn 2 failed, sessions died with no work completed.

## Problem

OpenMAIC normalizes provider-native reasoning for the OpenAI-compatible chat transport in three steps:

1. **Response side** — `reasoning-sse.ts` folds streamed/non-streamed `reasoning_content` into an inline `<think>…</think>` block in `content`.
2. **Split** — `extractReasoningMiddleware({ tagName: 'think' })` turns that block back into first-class reasoning parts for the agent loop and UI.
3. **Request side** — the next turn serializes the stored message through the OpenAI chat schema, which has no reasoning field, so reasoning is dropped.

Kimi K3 had a preservation mechanism for step 3 (marker-encode reasoning parts in `transformParams`, restore them into the native `reasoning_content` field in `compatFetch`). DeepSeek had none, so every multi-turn DeepSeek thinking session failed on the second request.

Three distinct gaps:

**a. Round-trip loss (provider side).** No preservation/restore for DeepSeek's request adapter.

**b. Round-trip loss (driver side).** `lib/agent/runtime/stream-fn.ts` gated `includeReasoning` to `provider === 'kimi.chat' && modelId === 'kimi-k3'`. On the driver, every `thinking` block was dropped before the AI SDK saw the messages, so the middleware had nothing to encode. A reasoning-less backfill alone made requests structurally valid but silently discarded the model's own prior reasoning.

**c. Missing field / effort conflict.** A response that skipped reasoning (direct tool call) leaves an assistant message with no reasoning part, which thinking mode still requires; and the body builder always attached `reasoning_effort` (default `high`) when thinking was enabled, which cannot combine with function tools.

### 1. One predicate for both sides (`lib/ai/providers.ts`)

```ts
export function preservesReasoning(providerId: string, modelId: string): boolean {
  if (providerId === 'kimi' && modelId === 'kimi-k3') return true;
  return getCatalogThinkingCapability(providerId, modelId)?.requestAdapter === 'deepseek';
}
```

- Derived from the **resolved request adapter**, so `atlascloud:deepseek-ai/deepseek-v4-pro` (same adapter) is covered — a literal `providerId === 'deepseek'` check missed it.
- `preservesReasoningForModel()` adapts it for the AI SDK model instance, and **the agent driver's `includeReasoning` gate now uses it** — the two sides cannot drift apart again.

### 2. Request-side wiring (`lib/ai/providers.ts`)

- Preservation middleware + `reasoning_content` restore run for every preserving model (not just Kimi K3).
- Direct request-body gates replaced by the predicate (round-trip, non-streaming normalization, middleware assembly).

### 3. Empty backfill + thinking mode

- When thinking is enabled, assistant messages without `reasoning_content` get `""` (verified accepted by the API).
- Both restore and backfill are gated on the **resolved** thinking mode: a disabled turn carries neither markers nor field, so no `reasoning_content` leaks into a non-thinking request.
- For preserving providers the catalog thinking default is injected when no caller supplied one, so the wire always states the mode explicitly instead of relying on the provider's server-side default (which decides whether the field must round-trip).

### 4. Effort scoping (`case 'deepseek'`)

A tool-carrying request never sets `reasoning_effort` (the transport rejects function tools combined with it) — the toggle goes alone. Every other request carries an effort: the explicit value when the caller set one, else the historical default:

| Request | Body |
| --- | --- |
| tools, no explicit effort | `thinking: {type:'enabled'}` |
| tools, explicit `max` | `thinking: {type:'enabled'}` (effort cannot be combined) |
| no tools, explicit `max` | `thinking: {type:'enabled'}`, `reasoning_effort:'max'` |
| no tools, explicit `enabled` without effort | `thinking: {type:'enabled'}`, `reasoning_effort:'high'` |
| no tools, no caller config (catalog default) | `thinking: {type:'enabled'}`, `reasoning_effort:'high'` |
| Kimi K3, thinking disabled (`{enabled:false}`) | `reasoning_effort:'low'`, no `thinking` object — prior reasoning kept as `reasoning_content` (the request still reasons) |
| DeepSeek, `effort:'none'` | `thinking:{type:'disabled'}` — prior reasoning stripped, no field, no marker |

## Changes

| File | Change |
| --- | --- |
| `lib/ai/providers.ts` | `preservesReasoning` / `preservesReasoningForModel`; all request-side gates use it; empty backfill; explicit thinking default for preserving providers; tool-scoped effort |
| `lib/agent/runtime/stream-fn.ts` | Driver `includeReasoning` uses the shared predicate (was Kimi K3 only) |
| `lib/ai/reasoning-sse.ts` | Generic reasoning-preservation middleware + restore helper (renamed from the Kimi-specific names) |
| `tests/ai/reasoning-sse.test.ts` | Updated to the new helper names |
| `tests/ai/openai-sdk-integration.test.ts` | New DeepSeek recorded-shape suite (6 tests) |
| `tests/ai/openai-sdk-integration.test.ts` | + `OpenCode Go session header` suite: header sent when `OPENCODE_GO_SESSION` is set, absent without the opt-in |

## Testing

- **Unit**: `tests/ai/openai-sdk-integration.test.ts` 11/11 (6 new DeepSeek tests), `tests/ai/reasoning-sse.test.ts` 14/14, `tests/ai/thinking-config.test.ts` 23/23, `tests/agent-runtime` + `tests/lib/agent/runtime` (driver/stream) suites green; `tests/ai` + `tests/agent-runtime` together: 1140 passed (one unrelated flaky timeout in `voice-clone-tools`, passes in isolation on both this branch and `main`).
- **Mutation check**: disabling the predicate fails 5 of the new tests (round-trip, backfill, effort scoping, non-streaming, predicate) — the coverage is real, not a rename.
- **New tests assert**, at the recorded request-body level: request 2 of a tool turn carries the non-empty `reasoning_content` (streaming); a reasoning-less turn gets `""`; tool calls omit `reasoning_effort`; non-tool calls keep `high`/`max`; a non-preserving provider emits no `reasoning_content`; the non-streaming response is folded without leaking into the text; and the predicate covers deepseek, atlascloud deepseek, Kimi K3 but not Kimi K2.6.
- **Type check**: `tsc --noEmit` clean.

## Notes

- **Wire change vs before this PR**: limited to tool-carrying requests, which now send `thinking: {type:'enabled'}` without `reasoning_effort` (the transport rejects function tools combined with it). Non-tool calls are unchanged: they carry the explicit effort, else the historical default `high`.
- The strip/restore decision is derived from the params the body builder actually emits (`disablesThinking`), not from the requested config or the serialized body — so adapters that cannot express "off" (Kimi K3, GLM/doubao forced-effort) always restore, and models whose off-switch is a dedicated field always strip.
- No new dependencies, no schema/DB changes. The marker string is a private in-flight encoding, never persisted or shown to users.
- Also includes opt-in `x-opencode-session` header support for the DeepSeek provider (`OPENCODE_GO_SESSION`), needed by gateways like OpenCode Go that require it; unset means byte-identical behavior.

## Checklist

- [x] No new dependencies
- [x] Tests added for the new behavior, and verified to fail without the fix
- [x] Type check clean
- [x] Round-trip now works on the agent-driver path the fix targets

### DIFF
--- a/lib/ai/providers.ts
+++ b/lib/ai/providers.ts
@@ -33,8 +33,8 @@ import { createAmazonBedrock } from '@ai-sdk/amazon-bedrock';
 import { createGoogleGenerativeAI } from '@ai-sdk/google';
 import { wrapLanguageModel, extractReasoningMiddleware } from 'ai';
 import {
-  createKimiReasoningPreservationMiddleware,
-  restoreKimiReasoningInRequestBody,
+  createReasoningPreservationMiddleware,
+  restoreReasoningContentInRequestBody,
   wrapJsonResponseWithReasoning,
   wrapResponseWithReasoning,
 } from './reasoning-sse';
@@ -1717,7 +1717,12 @@ function getCompatThinkingBodyParams(
       if (mode === 'disabled' || config.effort === 'none') {
         return { thinking: { type: 'disabled' } };
       }
-
+      // No explicit effort: send the thinking toggle alone. Tool-using
+      // transports (the agent driver) cannot combine reasoning_effort with
+      // function tools, and DeepSeek applies its own default effort anyway.
+      if (config.effort === undefined) {
+        return { thinking: { type: 'enabled' } };
+      }
       const effort = config.effort === 'max' || config.effort === 'xhigh' ? 'max' : 'high';
       return {
         thinking: { type: 'enabled' },
@@ -2261,14 +2266,20 @@ export function getModel(config: ModelConfig): ModelWithInfo {
           }
 
           if (
-            providerId === 'kimi' &&
-            config.modelId === 'kimi-k3' &&
             init?.body &&
-            typeof init.body === 'string'
+            typeof init.body === 'string' &&
+            ((providerId === 'kimi' && config.modelId === 'kimi-k3') || providerId === 'deepseek')
           ) {
             try {
               const body = JSON.parse(init.body);
-              restoreKimiReasoningInRequestBody(body);
+              restoreReasoningContentInRequestBody(body);
+              if (providerId === 'deepseek' && body.thinking?.type === 'enabled') {
+                for (const message of body.messages ?? []) {
+                  if (message?.role === 'assistant' && message.reasoning_content === undefined) {
+                    message.reasoning_content = '';
+                  }
+                }
+              }
               init = { ...init, body: JSON.stringify(body) };
             } catch {
               /* leave body as-is */
@@ -2292,7 +2303,7 @@ export function getModel(config: ModelConfig): ModelWithInfo {
           }
           const normalizedReasoningResponse = streaming
             ? wrapResponseWithReasoning(response)
-            : providerId === 'kimi' && config.modelId === 'kimi-k3'
+            : (providerId === 'kimi' && config.modelId === 'kimi-k3') || providerId === 'deepseek'
               ? await wrapJsonResponseWithReasoning(response)
               : response;
 
@@ -2349,13 +2360,15 @@ export function getModel(config: ModelConfig): ModelWithInfo {
       // Split it into first-class reasoning parts so the agent stream and UI can
       // show a thinking panel and the answer text stays clean.
       if (usesCompatTransport) {
-        const middleware =
-          config.providerId === 'kimi' && config.modelId === 'kimi-k3'
-            ? [
-                createKimiReasoningPreservationMiddleware(),
-                extractReasoningMiddleware({ tagName: 'think' }),
-              ]
-            : extractReasoningMiddleware({ tagName: 'think' });
+        const preservesReasoning =
+          (config.providerId === 'kimi' && config.modelId === 'kimi-k3') ||
+          config.providerId === 'deepseek';
+        const middleware = preservesReasoning
+          ? [
+              createReasoningPreservationMiddleware(),
+              extractReasoningMiddleware({ tagName: 'think' }),
+            ]
+          : extractReasoningMiddleware({ tagName: 'think' });
         model = wrapLanguageModel({
           model,
           middleware,
--- a/lib/ai/reasoning-sse.ts
+++ b/lib/ai/reasoning-sse.ts
@@ -61,9 +61,11 @@ function extractKimiReasoning(content: string): { content: string; reasoning?: s
 /**
  * The OpenAI chat adapter drops standardized reasoning prompt parts. Encode
  * them as private text markers until the request reaches our fetch wrapper,
- * where they are restored to Kimi's `reasoning_content` field.
+ * where they are restored to the provider's native `reasoning_content` field.
+ * Shared by Kimi and DeepSeek: both require reasoning_content to be passed
+ * back verbatim on the next turn while their thinking mode is on.
  */
-export function createKimiReasoningPreservationMiddleware(): LanguageModelMiddleware {
+export function createReasoningPreservationMiddleware(): LanguageModelMiddleware {
   return {
     specificationVersion: 'v3',
     transformParams: async ({ params }) => ({
@@ -84,8 +86,8 @@ export function createKimiReasoningPreservationMiddleware(): LanguageModelMiddle
   };
 }
 
-/** Restore private Kimi reasoning markers after OpenAI chat serialization. */
-export function restoreKimiReasoningInRequestBody(body: unknown): void {
+/** Restore private reasoning markers after OpenAI chat serialization. */
+export function restoreReasoningContentInRequestBody(body: unknown): void {
   if (!body || typeof body !== 'object') return;
   const messages = (body as { messages?: unknown }).messages;
   if (!Array.isArray(messages)) return;
--- a/tests/ai/reasoning-sse.test.ts
+++ b/tests/ai/reasoning-sse.test.ts
@@ -1,8 +1,8 @@
 import { describe, it, expect } from 'vitest';
 import {
-  createKimiReasoningPreservationMiddleware,
+  createReasoningPreservationMiddleware,
   createReasoningContentRewriter,
-  restoreKimiReasoningInRequestBody,
+  restoreReasoningContentInRequestBody,
   wrapJsonResponseWithReasoning,
   wrapResponseWithReasoning,
 } from '@/lib/ai/reasoning-sse';
@@ -155,7 +155,7 @@ describe('wrapResponseWithReasoning', () => {
 
 describe('Kimi reasoning preservation', () => {
   it('round-trips reasoning prompt parts through OpenAI-compatible serialization markers', async () => {
-    const middleware = createKimiReasoningPreservationMiddleware();
+    const middleware = createReasoningPreservationMiddleware();
     const params = await middleware.transformParams!({
       type: 'stream',
       model: {} as never,
@@ -187,7 +187,7 @@ describe('Kimi reasoning preservation', () => {
       ],
     };
 
-    restoreKimiReasoningInRequestBody(body);
+    restoreReasoningContentInRequestBody(body);
 
     expect(body.messages[0]).toMatchObject({
       content: null,


wyuc (changes requested):
Thanks for taking this on — the direction is right, and the provider-side pieces (the reasoning preservation middleware, the request-body restore, the `reasoning_content: ""` backfill) all work as intended when driven through the AI SDK directly: a multi-step tool probe on `deepseek:deepseek-v4-pro` produced a second request carrying `reasoning_content: "use the lookup tool"` next to `content: null` + `tool_calls`. Two things block merging as-is.

**1. The round-trip is dead on the agent-driver path this PR exists to fix.** `lib/agent/runtime/stream-fn.ts:404-409` still gates `includeReasoning` to `provider === 'kimi.chat' && modelId === 'kimi-k3'`. The Kimi change added both that gate and the provider-side middleware together; this PR generalized only the provider side. On the driver, every `thinking` block is dropped before the AI SDK sees the messages, so the middleware has nothing to encode and the restore has nothing to restore. Captured wire body for turn 2 through `createCallLlmStreamFn` with a stubbed fetch:

```json
{"role":"assistant","content":null,"tool_calls":[{"id":"call-1","type":"function","function":{"name":"lookup","arguments":"{}"}}],"reasoning_content":""}
```

The request is structurally valid again, but the model loses its own prior reasoning across tool turns, which is what the field is for. Suggested fix: derive one predicate from `lib/ai/providers.ts` (e.g. `preservesReasoning(providerId, modelId)`) and use it in both the driver gate and the provider-side gates so they cannot drift again.

**2. No test covers any of the DeepSeek behaviour.** The test diff is a pure rename. Removing all four DeepSeek gates in `providers.ts` plus the effort-less branch and re-running the three suites the PR cites (`reasoning-sse`, `thinking-config`, `openai-sdk-integration`) still gives 42/42 green. Please add a DeepSeek counterpart to the existing Kimi K3 recorded-shape test asserting: request 2 of a tool turn carries the non-empty `reasoning_content` (streaming); a reasoning-less assistant turn gets `""`; `thinking: {type:'enabled'}` with no `reasoning_effort` when effort is unset and `high`/`max` when set; a non-DeepSeek provider with a reasoning part does not emit `reasoning_content`; and the non-streaming (`wrapJsonResponseWithReasoning`) path.

Smaller points, not blocking on their own:

- `atlascloud:deepseek-ai/deepseek-v4-pro` shares the `deepseek` request adapter (`lib/ai/model-metadata.ts:384`) but the new gates key off the literal `providerId === 'deepseek'`, so it sends `thinking` yet never restores/backfills `reasoning_content`. Branching on the resolved `requestAdapter` instead of the provider id would cover it.
- "Default behavior is unchanged" is not accurate: before this PR, `case 'deepseek'` sent `reasoning_effort: 'high'` whenever thinking was enabled without an explicit effort; now it sends only `thinking: {type:'enabled'}`. That changes reasoning intensity for every effort-less DeepSeek call, not just the driver. Either state it in the description (and confirm the API default is equivalent) or scope the effort-less branch to requests that carry tools.
- The backfill is gated on `body.thinking?.type === 'enabled'`, so a route that pins a DeepSeek model without an explicit `thinking` config sends no `thinking` field and gets no backfill; if the server-side default for that model family is thinking-on, the 400 persists. Conversely the restore runs regardless of thinking mode, so a disabled turn can carry `reasoning_content`. Gating both on the resolved thinking mode would be cleaner. I could not verify the API's default mode offline, so treat this as a question.

Happy to re-review quickly once the driver gate and the tests are in.

nqthiep:
Thanks — both blockers were real, and the driver-path one was a genuine hole in the original PR: I had verified "sessions succeed" rather than the wire, which only proved the 400 disappeared, not that reasoning round-tripped. Addressed in 4e479cc:

**1. Driver gate (blocker).** Added one predicate in `lib/ai/providers.ts`:

```ts
export function preservesReasoning(providerId: string, modelId: string): boolean {
  if (providerId === 'kimi' && modelId === 'kimi-k3') return true;
  return getCatalogThinkingCapability(providerId, modelId)?.requestAdapter === 'deepseek';
}
```

`preservesReasoningForModel()` adapts it to the AI SDK model instance, and `stream-fn.ts`'s `includeReasoning` now uses it — both sides read the same source, so they cannot drift.

**2. Tests (blocker).** Added a DeepSeek recorded-shape suite in `tests/ai/openai-sdk-integration.test.ts` (6 tests) asserting on the captured request bodies: request 2 of a tool turn carries the non-empty `reasoning_content`; a reasoning-less turn gets `""`; tool calls carry `thinking: {type:'enabled'}` with no `reasoning_effort`; non-tool calls keep `high`/`max`; a non-preserving provider (qwen) emits no `reasoning_content`; the non-streaming response folds without leaking the block into the text; plus the predicate itself (deepseek, atlascloud deepseek, Kimi K3 true; Kimi K2.6 false). I ran your mutation check in reverse: disabling the predicate fails 5 of them, so the suite is not a rename.

**Smaller points:**

- **atlascloud** — now derived from the resolved `requestAdapter`, so `atlascloud:deepseek-ai/deepseek-v4-pro` gets the same restore/backfill (locked by the predicate test).
- **"Default behavior unchanged"** — you were right, that was inaccurate. Behavior is now stated explicitly in the PR description, and the effort-less branch is scoped to **requests that carry tools** (the actual transport constraint), so non-tool calls keep the historical default effort. Table in the description.
- **Gating on the resolved thinking mode** — done, with one addition that I think answers your open question rather than guessing it: for preserving providers the catalog thinking default is now injected when no caller supplied a config, so the wire always states the mode explicitly instead of depending on the provider's server-side default. Restore and backfill are both gated on that resolved mode, so a disabled turn carries neither markers nor field.

wyuc (changes requested):
Round 2 on 4e479cc4 — both round-1 blockers are closed, thank you. The driver gate now goes through the shared `preservesReasoning` predicate (I captured turn 2 through `createCallLlmStreamFn` and it carries the non-empty `reasoning_content`, with the qwen control emitting no field), Atlas Cloud is covered via the resolved `requestAdapter`, and the new DeepSeek suite is real coverage (breaking the predicate turns 5 of the 6 tests red). Three small things before merge:

1. **`pnpm check` is red** (CI's Lint/Typecheck job fails on formatting): `lib/ai/providers.ts:2313-2317` (the `roundTripProvider && init?.body …` condition fits on one line) and `tests/ai/openai-sdk-integration.test.ts:473` (line over the print width). `pnpm format` fixes both.

2. **Disabled-mode gating leaks the internal marker.** The preservation middleware is installed whenever the provider preserves reasoning, regardless of thinking mode, so prior reasoning parts are encoded into `content` as `openmaic:kimi-reasoning:<len>:<text>`. The only code that turns that marker back into `reasoning_content` is `restoreReasoningContentInRequestBody`, which is now skipped when `body.thinking?.type === 'disabled'`. Net effect on a disabled turn (`thinkingConfig: {enabled:false}` on `deepseek:deepseek-v4-pro`, with a prior `thinking` block in the transcript):

   ```json
   { "role": "assistant", "content": "openmaic:kimi-reasoning:19:use the lookup tool", "tool_calls": [ ... ] }
   ```

   The provider receives the sentinel verbatim, which contradicts the description's "a disabled turn carries neither markers nor field". The same gate is also ineffective for Kimi K3, whose adapter emits `reasoning_effort` and never a `thinking` object, so its "disabled" turns still send `reasoning_content` (pre-existing, but the new gating was meant to cover it). Suggested fix: key the decision on the resolved `ThinkingConfig` rather than on the serialized body — either do not encode reasoning parts when the resolved mode is disabled, or always strip the marker on that path.

3. **The driver wiring has no test.** Reverting `stream-fn.ts:406` to the old Kimi-K3-only gate leaves every committed suite green; since that line was exactly the round-1 blocker, please add a `createCallLlmStreamFn` recorded-body test (deepseek tool turn → request 2 carries the non-empty `reasoning_content`).

With those in, this is good to go.

nqthiep:
All three in 14f7dc8.

**1. `pnpm check`.** Fixed — `pnpm exec prettier --write` on the touched files. The `roundTripProvider && init?.body …` condition now fits on one line and the long test line is wrapped. `pnpm check` is clean for everything tracked (the only remaining warning is a local `.omo/` continuation file that is not part of the repo).

**2. Marker leak on disabled turns.** You were right on both counts — the gate keyed on the serialized body, and Kimi K3 never emits a `thinking` object so its disabled turns kept sending `reasoning_content`. The decision now keys on the resolved `ThinkingConfig`:

```ts
if (getThinkingMode(thinking) === 'disabled') {
  stripReasoningContentInRequestBody(body);
} else {
  restoreReasoningContentInRequestBody(body);
  if (deepseekAdapter) { /* backfill "" only for the DeepSeek adapter */ }
}
```

`stripReasoningContentInRequestBody` decodes the markers and drops them without emitting the field, so a disabled turn carries neither the sentinel nor `reasoning_content` — including on the Kimi K3 path, where the resolved config now decides instead of the absent body field. Backfill is additionally scoped to the DeepSeek request adapter, so no other provider gains an empty field.

**3. Driver-wire test.** Added to `tests/lib/agent/runtime/stream-fn.test.ts`, driving `createCallLlmStreamFn` with a stubbed fetch and a prior turn containing a `thinking` block + tool call:

- thinking on → the outgoing assistant message carries `reasoning_content: 'use the lookup tool'` (non-empty, so it cannot be satisfied by the backfill);
- thinking off → no `reasoning_content` and `content` contains no `openmaic:kimi-reasoning` sentinel.

Mutation-checked both: reverting `stream-fn.ts` to the Kimi-K3-only gate fails exactly the first test; skipping the strip fails exactly the second.

`tsc --noEmit` clean; `tests/ai`, `tests/lib/agent/runtime`, `tests/agent-runtime` all green (103 files, 4 skipped).

wyuc (changes requested):
Thanks for the round-3 update. The main DeepSeek path is now correct: with thinking on, the prior reasoning round-trips as `reasoning_content`; with thinking disabled on DeepSeek, neither the marker nor the field is sent; `pnpm check` is clean; and the new driver-wire test in `stream-fn.test.ts` genuinely guards the round-1 blocker (reverting the driver gate turns it red). Items 1 and 3 from my last review are closed, and item 2 is closed for non-empty reasoning.

Two issues remain, both small, and I'd like them fixed before merge:

**1. Kimi K3 "disabled" still thinks on the wire, but its reasoning is now stripped.** The new gate in `providers.ts` strips `reasoning_content` whenever `getThinkingMode(thinking) === 'disabled'`. Kimi K3's effort capability is `['low', 'high', 'max']` with no `none` and it isn't toggleable, so `{ enabled: false }` resolves to `reasoning_effort: "low"` and no `thinking` object: the model keeps reasoning, but the request drops the field. `preservesReasoning`'s own doc comment says Kimi K3 "rejects a multi-turn request whose assistant messages drop the field", so this path likely reintroduces the 400 this PR fixes (inferred, not reproduced against the live API). It's reachable via a route `thinking: { enabled: false }` or `LLM_THINKING_DISABLED=true`. Please gate on whether the outgoing request actually disables thinking rather than on the requested mode, e.g. have the body-param builder report a `disabled` boolean (always false for Kimi K3), or always take the restore branch for capabilities without a `none` effort. The same change would also cover `{ effort: 'none' }` on DeepSeek, which today sends `thinking: { type: "disabled" }` but still goes through the restore branch.

**2. Empty reasoning still leaks the private marker into `content`.** An empty reasoning part is encoded as the NUL-prefixed `openmaic:kimi-reasoning:` marker followed by `0:`, and `extractKimiReasoning` returns the original `content` (marker included) when the accumulated reasoning is empty, so both `restoreReasoningContentInRequestBody` and `stripReasoningContentInRequestBody` keep it. A prior assistant turn with an empty thinking block therefore sends that marker as its `content` on both enabled and disabled turns. Normal streaming doesn't produce empty thinking blocks, but persisted or resumed transcripts can. Returning `remaining` whenever a marker was consumed (a `found` flag rather than `reasoning ? ...`), and/or skipping empty reasoning parts before encoding, fixes it.

**Tests:** please add a recorded-body test for Kimi K3 with `{ enabled: false }` (asserting the field is kept, or whatever the final rule is) and a case with an empty prior thinking block. Two current guarantees are also untested: widening the empty-string backfill from the DeepSeek adapter to every adapter, or reverting the gate to read the serialized `thinking` field, both leave the suite green. The Kimi K3 case would pin the second one.

Nothing else from me; once these land I'll re-check only these points.

nqthiep:
One more fix in 8d7a787, found by a post-implementation review pass over the round-2 delta.

**Your round-1 point about "scope the effort-less branch to requests that carry tools" was only partially honored.** The condition was `options.hasTools || config.effort === undefined`, so a NON-tool call that enabled thinking without an explicit effort (e.g. `{enabled:true}`) also dropped `reasoning_effort`, where the historical wire carried `reasoning_effort: 'high'`. That is the same class of reasoning-intensity change you flagged, just on the non-tool path — and the code comment above it claimed the opposite ("without tools the historical default effort is preserved").

Now:

```ts
if (options.hasTools) return { thinking: { type: 'enabled' } };
const effort = config.effort === 'max' || config.effort === 'xhigh' ? 'max' : 'high';
return { thinking: { type: 'enabled' }, reasoning_effort: effort };
```

Only a tool-carrying request omits the effort (the transport constraint); every other request keeps the explicit value or the default.

- Test updated: the non-tool `{enabled:true}` case now asserts `reasoning_effort: 'high'` (it previously enshrined the omission). Tool-carrying `{mode:'enabled',effort:'max'}` still asserts no effort key.
- Mutation-checked: restoring the old condition fails exactly that test.
- `tsc --noEmit` clean, prettier clean, 38/38 in the three touched suites, 103 files green across `tests/ai` + `tests/lib/agent/runtime` + `tests/agent-runtime`.

The PR description's behavior table had the same row wrong; corrected to `thinking: {type:'enabled'}, reasoning_effort:'high'`.

nqthiep:
**Ready for re-review.** Summary of where this PR stands after the post-implementation review pass, so you don't have to reconstruct it from the thread.

**Delta since your round-2 review** (`14f7dc8`) is one commit, `8d7a787`, two files (+7/−4):

```diff
- if (options.hasTools || config.effort === undefined) {
+ if (options.hasTools) {
    return { thinking: { type: 'enabled' } };
  }
```

Rationale: your round-1 point asked to scope the effort-less branch to requests that carry tools. It was only partially honored — the `|| config.effort === undefined` clause also dropped `reasoning_effort` for NON-tool calls that enabled thinking without an explicit effort, where the historical wire carried `'high'`. Now only a tool-carrying request omits the effort; every other request keeps the explicit value, else the default.

Verified:

- `tests/ai/openai-sdk-integration.test.ts` — the non-tool `{enabled:true}` case asserts `reasoning_effort:'high'` (it previously enshrined the omission); the tool-carrying `{mode:'enabled',effort:'max'}` case still asserts no effort key.
- Mutation check: restoring the old condition fails exactly that test.
- 38/38 across the three touched suites; 103 files green across `tests/ai` + `tests/lib/agent/runtime` + `tests/agent-runtime`; `tsc --noEmit` clean; prettier clean; Docker production build clean.
- DeepSeek still returns 402 in this environment, so no live multi-turn run is possible here — the wire-level assertions plus mutation checks are the evidence.

**Also corrected in the PR description** (both were stale after the fix): the behavior table's `no tools, explicit enabled without effort` row now reads `thinking:{type:'enabled'}, reasoning_effort:'high'`, and the Notes bullet now states the only wire change vs before is on tool-carrying requests.

Thanks for the two thorough rounds — the driver-gate catch in round 1 was the one that mattered.

wyuc:
Thanks — the `8d7a787` change (keeping `reasoning_effort` for non-tool requests) looks right. Heads-up that there's a round-3 review posted a bit earlier today that this summary doesn't cover: https://github.com/THU-MAIC/OpenMAIC/pull/1486#pullrequestreview-5236720176

Two items are still open there: (1) the disabled-thinking gate strips `reasoning_content` for Kimi K3 even though its request still reasons (`reasoning_effort: "low"`, no `none` level), which likely brings back the 400; (2) an empty reasoning part still leaks the private marker into `content`. Plus a Kimi K3 `{ enabled: false }` recorded-body test and an empty-thinking case. Details and suggested fixes are in the linked review.

nqthiep:
Both round-3 items fixed in 8d17850, verified the way you verify things — by watching tests fail first.

**Verified your two claims on current code before touching it.** A scratch probe (deleted afterwards) showed: an empty `\u0000openmaic:kimi-reasoning:0:` marker survives both `restore` and `strip` verbatim in `content`, and a Kimi K3 `{enabled:false}` turn drops a non-empty prior `reasoning_content` while the wire carries `reasoning_effort`. Both red, exactly as you described. (`pickThinkingEffort` disabled-fallback → `'low'` confirmed at `thinking-config.ts:51-57`.)

**1. Gate now derives from the params actually written, not the requested mode.** `getCompatThinkingBodyParams` returns `{ params, disablesThinking }`, with `disablesThinking: true` only on branches that provably turn thinking off on the wire (DeepSeek `disabled`/`none`, kimi/xiaomi `disabled`, qwen/siliconflow-toggle/openrouter/hunyuan-no_think/lemonade/gateway off-switches). Notably `false` for the `openai` adapter (incl. Kimi K3), forced-effort GLM/doubao, and budget-only siliconflow — an effort value is not an off switch. The compatFetch gate keys on that boolean, so Kimi K3 `{enabled:false}` restores (wire still reasons at `low`) and DeepSeek `{effort:'none'}` strips (wire says `thinking:{type:'disabled'}`). The serialized-body alternative you were worried about is covered too: a gate of "strip unless `body.thinking` is enabled" fails the new Kimi test, because that body never contains a `thinking` object.

**2. Empty-reasoning leak fixed at both ends.** `extractKimiReasoning` reports a `found` flag (set only after a bounds-checked consume; malformed markers still break out unchanged); restore/strip act on `found`, so an empty marker yields `reasoning_content: ''` + clean content on restore and clean content + no field on strip. The preservation middleware no longer encodes empty reasoning parts, so the degenerate `:0:` marker is never produced. Restore/strip share one internal helper since they now differ by exactly the field assignment.

**Tests** (all in the diff, all mutation-checked with clean reverts):
- Kimi K3 `{enabled:false}` recorded body: `reasoning_effort:'low'` present, no `thinking` object, prior reasoning kept, no marker.
- DeepSeek `{effort:'none'}`: `thinking:{type:'disabled'}`, no field, no marker.
- Driver-wire empty thinking block, both modes (enabled → field `''` + clean; disabled → no field + clean).
- Kimi K3 reasoning-less turn gets no backfill (pins the DeepSeek-only scope).
- Unit cases for the exact `:0:` marker in both functions.

`tsc --noEmit` clean, prettier clean, full `tests/ai` + `tests/lib/agent/runtime` + `tests/agent-runtime` green.

Two honest notes: (a) the one functional gap I could not close by test is a live multi-turn run — DeepSeek is 402 here and there are no Kimi credentials, so the evidence is recorded wire bodies + mutations, the same level you used; (b) NITPICK-level items from the earlier self-review (provider-string mapping fragility, double capability lookup, repeated JSON.parse) are intentionally untouched to keep this diff scoped — happy to split them into a follow-up if you want.

wyuc:
Thanks @nqthiep — I'm re-reviewing `8d17850` against the two round-3 items now and will follow up here.

One request in the meantime: could you move `8ee72a8` ("send x-opencode-session for the deepseek provider when configured") into its own PR? It's an unrelated feature, and keeping this PR scoped to the reasoning round-trip fix makes it easier to review and to revert independently if needed. A force-push that drops that commit from this branch is fine.

nqthiep:
Split done as requested: `8ee72a8` ("send x-opencode-session for the deepseek provider when configured") is out of this branch — it now ends at `8d7a787`, and the PR description no longer mentions the header.

The header work lives on as its own PR: #1593 (`feat/opencode-go-session-header`, one commit, opt-in only, tests included). Keeping the two review threads separate from here on.

Thanks again for catching the scope creep before it fossilized into the history. 🙏

wyuc:
Thanks @nqthiep for splitting the header work into #1593. One problem with the force-push: the branch was reset to `8d7a787`, which is the **parent** of your round-3 fix commit `8d17850` ("derive the round-trip decision from the wire, not the requested mode"). So that fix was dropped along with `8ee72a8`, and both round-3 blockers are back on the current head:

- Kimi K3 with `{enabled: false}` still reasons on the wire (`reasoning_effort: "low"`) but `reasoning_content` is stripped.
- An empty reasoning part sends the private marker into `content`.

The commit still exists, so restoring it should be quick: `git cherry-pick 8d17850` on top of `8d7a787` (it shouldn't touch the header code). Push that, and we'll do the final review.

PR #1487, opened 2026-09-13
feat: auto-detect Vietnamese for browser-native TTS narration

nqthiep:
## Summary

Browser-native TTS auto-detection only distinguished Chinese (CJK ratio) from everything else (`en-US`), so Vietnamese narration was read by an English voice. This PR adds Vietnamese auto-detection to the shared language helper and binds an installed `vi` voice in the playback engine when the user has not picked one explicitly.

## Problem

`playBrowserTTSChunk` in the playback engine set:

```ts
utterance.lang = cjkRatio > CJK_LANG_THRESHOLD ? 'zh-CN' : 'en-US';
```

Vietnamese text therefore got `lang='en-US'` with the auto voice — macOS/Safari reads Vietnamese with English phonetics (near-unintelligible for learners). The "Test TTS" preview path (`inferPreviewLang` in `browser-tts-preview.ts`) had the same CJK-only logic, so the preview was equally wrong.

### 1. Shared `detectSpeechLang()` (`lib/audio/browser-tts-preview.ts`)

Returns `zh-CN` / `vi-VN` / `en-US` fallback, used by both the preview path and the playback engine:

- CJK ratio check first (unchanged behavior).
- A hit on `đ/Đ`, `ơ/Ơ`, `ư/Ư`, or the U+1EA0–U+1EF9 precomposed block (`ớ`, `ừ`, `ồ`, `ế`, `ấ`, …) — codepoints that never occur in French or Romanian Latin — marks Vietnamese immediately. (This distinction matters: tone-marked variants are *different codepoints* from their base letters, so a naive `[đơưêô]` class misses them — caught by test.)
- Bare `ă/â/ê/ô` (which do occur in Romanian, and `ê/ô` in French) only count toward a low ratio (> 2%). Known limitation, documented in code: lone French words like `"fête"` or Romanian prose will still misclassify — acceptable because narration chunks follow the course language.

### 2. Playback engine (`lib/playback/engine.ts`)

- Uses the shared `detectSpeechLang()` instead of the inline CJK-only check (the duplicated `CJK_LANG_THRESHOLD` is removed).
- When the detected language is `vi-VN` and no explicit voice is configured, binds the first installed `vi-*` voice (`utterance.voice` + `utterance.lang`), since browsers otherwise fall back to an English voice reading Vietnamese text. An explicitly configured voice still wins (that path is untouched).

## Changes

| File | Change |
| --- | --- |
| `lib/audio/browser-tts-preview.ts` | New exported `detectSpeechLang()`; existing `inferPreviewLang()` delegates to it |
| `lib/playback/engine.ts` | Use shared helper; auto-bind installed `vi` voice; drop duplicated CJK constant |
| `tests/audio/detect-speech-lang.test.ts` | New: zh, Vietnamese (decisive / broad / real narration lines), English, French non-regression cases |

## Testing

- **Unit**: 5/5 pass — including a case (`"Xin chào cả lớp!"`) that failed on the first implementation and exposed the tone-marked-codepoint gap.
- **Production build**: `docker compose build` passes (Next.js compile + `tsc` clean).
- **End-to-end** (Playwright WebKit against a local Docker deployment, browser-native TTS, no explicit voice): every narration utterance carries `lang: "vi-VN"` with an installed Vietnamese voice auto-selected (`"Linh"`), advancing through all lines with `onend` firing — previously `lang` was `"en-US"` with the auto voice.

## Notes

- Default behavior for Chinese and English is byte-identical to before; only text matching Vietnamese rules changes (from `en-US` to `vi-VN` + voice binding).
- No new dependencies, no schema/migration changes.
- Requires an installed `vi` voice on the listener's OS for the voice binding to take effect (e.g. macOS `System Settings → Accessibility → Spoken Content → Vietnamese`); without one, only the `lang` tag improves, and the browser picks its closest voice.

## Checklist

- [x] No new dependencies
- [x] New unit tests added and passing
- [x] Docker production build passes
- [x] Verified end-to-end in a real browser engine (WebKit)

### DIFF
--- a/lib/audio/browser-tts-preview.ts
+++ b/lib/audio/browser-tts-preview.ts
@@ -18,9 +18,27 @@ function createAbortError(): Error {
 }
 
 function inferPreviewLang(text: string): string {
-  const cjkCount = (text.match(/[\u4e00-\u9fff\u3400-\u4dbf]/g) || []).length;
-  const ratio = text.length > 0 ? cjkCount / text.length : 0;
-  return ratio > CJK_LANG_THRESHOLD ? 'zh-CN' : 'en-US';
+  return detectSpeechLang(text);
+}
+
+// U+0110/0111 (đ), U+01A0/01A1 (ơ), U+01AF/01B0 (ư) and the U+1EA0–U+1EF9
+// precomposed block (ớ, ừ, ồ, ế, ấ, …) never occur in French or Romanian
+// Latin, so one occurrence marks Vietnamese. Bare ă/â/ê/ô do occur in
+// Romanian (and ê/ô in lone French words like "fête"), so they only count
+// toward a ratio — this does not fully exclude Romanian prose, which is
+// acceptable because narration chunks follow the course language.
+const VI_DECIDER_RE = /[đĐơƠưƯ\u1EA0-\u1EF9]/;
+const VI_BROAD_RE = /[ăâêôĂÂÊÔ]/g;
+const VI_BROAD_THRESHOLD = 0.02;
+
+/** Language tag for a narration chunk: zh-CN, vi-VN, or en-US fallback. */
+export function detectSpeechLang(text: string): string {
+  if (!text) return 'en-US';
+  const cjkRatio = (text.match(/[\u4e00-\u9fff\u3400-\u4dbf]/g) || []).length / text.length;
+  if (cjkRatio > CJK_LANG_THRESHOLD) return 'zh-CN';
+  if (VI_DECIDER_RE.test(text)) return 'vi-VN';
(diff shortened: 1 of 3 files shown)


CXuPercy (approved):
/tmp/pr1487-approve.md

wyuc (approved):
Approving. Sharing one `detectSpeechLang()` between playback and the TTS preview fixes Vietnamese narration being read by an English voice, and binding an installed `vi` voice only applies when no voice is configured; Chinese and English detection are unchanged. Minor note for later: very short French/Romanian chunks with several ê/ô/â can cross the 2% ratio and be tagged vi-VN, which you already call out as acceptable. Thanks!

Issue #1091, opened 2026-08-10
[EPIC] Open-source video export performance and reliability


cosarah:
## Outcome

Make OpenMAIC video export reliable and resource-conscious on a normal CPU-only host.

The default deployment should support a validated 1080p / 30fps / standard profile with bounded memory and deterministic output. This Epic does not promise a sub-5-minute single-video SLA or high-concurrency production serving.

## Scope

- Reproducible video-export benchmark corpus and stage-level metrics.
- BeginFrame capture profile and explicit worker/resource configuration.
- Capture-path performance improvements.
- A bounded local chunk executor and assembler.
- Stable RenderExecutor, JobStore, and ArtifactStore seams.
- Validated quality tiers and regression gates.
- Documentation for CPU-only, low-memory, and optional render-service deployments.

## Current evidence

- `NEXT_PUBLIC_ENABLE_VIDEO_EXPORT` already exists and defaults to disabled.
- A 720p short-sample export improved from 28.7s to 16.3s with BeginFrame, about 43% faster.
- A 1080p / 30fps / standard full-classroom sample still took 696.9s with one worker.
- Capture represented about 75.5% of the 1080p runtime; encode represented about 20.6%.
- Increasing CPU without adding capture parallelism cannot meet a five-minute target.
- Static-frame deduplication currently adds verification overhead or is disabled on BeginFrame, so it remains off by default.

## Non-goals

- PostgreSQL or Redis distributed queues.
- Provider-specific production infrastructure.
- GPU as the default rendering path.
- Default support for arbitrary 4K / 60fps combinations.

## Dependency shape

Benchmark → shared executor seam + resource profiles → bounded local chunk executor → quality tiers and release gates.

## Work items

- [ ] #1092 — Add a reproducible video-export benchmark corpus and stage metrics
- [ ] #1093 — Extract a stable RenderExecutor seam from the video render service
- [ ] #1095 — Harden BeginFrame and CPU resource profiles for video export
- [ ] #1094 — Add a bounded local chunk executor and deterministic assembler
- [ ] #1096 — Define supported video tiers and export regression gates

## Parallelism and dependencies

- #1092 has no prerequisite.
- #1093 and #1095 can proceed in parallel after the benchmark baseline is agreed.
- #1094 depends on #1092, #1093, and #1095.
- #1096 depends on the validated profiles and the local chunk path.

## Definition of done

[...]

PR #1492, opened 2026-09-14
feat(render): add admission and per-task resource budgets

LING-6150:
## Summary

Move the opt-in render resource fence out of HyperFrames and into OpenMAIC's
render-service boundary. Resource mode now uses the official, unmodified
`@hyperframes/producer` 0.8.37 package through a lockfile alias; OpenMAIC no
longer carries a HyperFrames source patch or private six-package build chain.

The default `npm start` path remains the existing in-process executor. Resource
mode is explicitly enabled with `npm run start:resources -- <root-owned-config>`
on a separately provisioned Linux/systemd host.

## Runtime and ownership

- `RenderCoordinator` remains the only request-admission and queue owner. Its
  queue and per-identity slots are request controls, not a second resource
  ledger.
- Resource mode is qualified for one render-service instance, one active render,
  and one fixed per-task budget. The tested budget was 1 CPU, 768 MiB memory,
  zero swap, and 256 PIDs/threads. Separate instances using separate state roots
  do not coordinate host capacity.
- Before listening, the bootstrap validates the fixed budget against the
  effective CPU/memory capacity and verifies the root-owned config, state root,
  browser, FFmpeg, worker identity, and empty prior state.
- A dedicated root owner starts each task as a transient systemd service. The
  unit establishes CPU, memory, swap, and PID limits before the task imports
  Producer, so Chrome, FFmpeg, and other descendants share the same task domain.
- The task mount namespace supplies a private tmpfs for `HOME`, `TMPDIR`, media
  intermediates, and the render candidate. The extracted project is mounted
  read-only.
- Closeout drains descendants, scans host procfs for external FD/mmap/cwd/root/exe
  references, transfers the candidate across the private-filesystem boundary,
  ordinary-unmounts, waits for task-cgroup removal, and only then atomically
  publishes the staged artifact.
- Publication, cleanup confirmation, reservation return, and admission closure
  are reported separately. Unverifiable cleanup retains the project/state,
  leaves the reservation unreturned, and closes admission. Once publication is
  explicitly committed, a late cancellation cannot relabel it or delete the
  artifact.
- Before exiting, the task runner captures `memory.current`, `memory.events`,
  `memory.events.local`, `cpu.stat`, `pids.current`, and `pids.events`. Missing
  fields are reported rather than fabricated, and accounting status does not
  alter cleanup status.
- Dynamic local pressure in the shared systemd slice or its memory ancestors
  closes admission. Owner loss or a non-empty state root requires platform audit
  before restart.

## Compatibility and deployment boundaries

- Default startup remains `tsx src/main.ts`, using the existing direct
  `@hyperframes/producer` 0.7.107 dependency and unprivileged runtime path.
- The install graph does change: it adds the official Producer 0.8.37 alias, and
  shared transitive versions follow `render-service/package-lock.json`. The
  alias's HyperFrames transitive packages are not all version 0.8.37.
- Existing Docker and Compose entrypoints keep their current privilege model and
  do not support `start:resources`. Resource mode requires a Linux host with
  systemd, cgroup v2, readable host procfs, mount namespaces, `mount`/`umount`,
  and `setpriv`.
- Config, installed service code, browser, FFmpeg, and state roots are trusted
  root-owned inputs. Each project directory must be exclusively owned by the
  unprivileged worker during its render. The external-reference guarantee does
  not cover a privileged process racing the bounded procfs scan.
- Local MP4 rendering only; chunk execution is rejected. Resource-mode progress
  remains `preparing` until the terminal result. The tested workloads do not
  qualify multi-instance capacity, ARM64, Bun, classroom-scale performance, or
  every media composition.

## Verification

The fixed Linux platform was Ubuntu 22.04 x86_64, Node 22.23.2, systemd 249,
cgroup v2, the existing managed Chromium, and FFmpeg 4.4.2. Evidence and
historical failures are retained outside the product repository.

| Evidence | Scope |
| --- | --- |
| `OUTER-PRODUCT-CLOSEOUT-20260920T214911Z-5a7d1c2e` | Cancellation, deadline, root-owner death, and audio/video temporary-path cases passed. Its first external-reference injection was invalid and remains recorded as a failure. |
| `OUTER-PRODUCT-CLOSEOUT-R2-20260920T221243Z-74c9b1e3` | External FD/mmap injection produced the expected refusal/quarantine; dynamic ancestor pressure closed admission. Its OOM recovery failed on a startup-transient classification and remains recorded. |
| `OUTER-PRODUCT-CLOSEOUT-R3-20260920T222119Z-9c4e7a21` | Task OOM followed by same-owner recovery passed after the bounded MainPID startup-transient fix. |
| `OUTER-PRODUCT-SETTLEMENT-R6-20260920T235417Z-82ced734` | The complete 3,083-file R6 snapshot passed normal export, cleanable task OOM, same-owner recovery, final cleanup, and byte-equivalent accounting across task result, owner IPC, and the product job record. Manifest SHA-256: `018046005c8f9795dcf0069935bccd9c8c9cc0ddaa02f7dc302e243d4e20f58b`. |

The final source is deliberately not claimed byte-identical to R6. After R6, a
four-file local delta changes the coordinator's post-publication cancellation
decision, adds focused regressions for committed publication/quarantine and the
unchanged default-executor behavior, and corrects README deployment claims. No
Linux VM matrix was rerun for that local commit-decision fix.

Current local verification for the post-R6 delta:

- 37 coordinator, settlement, route, and default-executor tests passed.
- 2 task-accounting tests passed.
- Render-service TypeScript typecheck passed.
- Merge decisions must use the checks attached to the PR's current HEAD;
  checks and approvals on `eaf47a8f` are historical and do not cover this
  final tree.

The old `render-service/producer-patch/` tree, private Producer build scripts,
installed-evidence harness, and source/package verification tests are removed.
No experiment archive, credential, VM material, or temporary probe is included
in the product diff.

Related to #1091 and #866.


render-service/src/resource-main.ts
```diff
@@ -0,0 +1,85 @@
...
+  typeof owner.cleanupTimeoutMs !== 'number' ||
+  !Number.isSafeInteger(owner.cleanupTimeoutMs) ||
+  owner.cleanupTimeoutMs <= 0 ||
+  owner.cleanupTimeoutMs > 60_000
+)
+  throw new Error('Invalid worker identity, task PID limit or cleanup bound');
+if (lstatSync(config.tmpDir).uid !== owner.workerUid)
+  throw new Error('Project root must already belong to the unprivileged service user');
+const settingsStat = lstatSync(settingsPath);
+if (settingsStat.uid !== 0 || settingsStat.mode & 0o022)
```
cosarah: **Pre-merge (security):** The settings file is validated here, but the forked owner re-reads the same path as root without re-validation (`resource-owner.mjs:119`), and only the file's own uid/mode is checked — never the containing directory's writability. If the settings directory is worker-writable, a compromised render task can swap the file between validation and the child's read; the swapped `packageRoot` then passes `verifyResourcePackage`, because its anchors (revision/patchSha256 from the public `source.json` plus a self-consistent receipt) bind to nothing root-controlled — arbitrary code execution as root.

Suggest passing the already-validated bytes to the child over IPC, or additionally requiring the settings directory to be root-owned and non-writable.

render-service/src/resource-owner.mjs
```diff
@@ -0,0 +1,177 @@
...
+    let invoked = false;
+    try {
+      if (Buffer.byteLength(JSON.stringify(message)) > 1024 * 1024)
+        throw new Error('Oversized resource request');
+      if (typeof message.deadlineNs !== 'string' || !/^\d+$/.test(message.deadlineNs))
+        throw new Error('Missing monotonic deadline');
+      deadlineNs = BigInt(message.deadlineNs);
+      const project = realpathSync(message.projectDir);
+      if (
+        project !== message.projectDir ||
```
cosarah, line 42: **Pre-merge (availability):** The owner requires `realpathSync(projectDir) === projectDir`, but `makeProjectDir()` builds paths under the *lexical* `config.tmpDir` (`PRODUCER_TMP_PROJECT_DIR`). Any symlinked component (e.g. `/var/run → /run`) makes 100% of renders fail this boundary check while the owner stays up, admission stays open and `/health` stays green — the service becomes a black hole that admits and then fails every job. The config forces `projectRoot` to be the realpath, so nothing flags the mismatch.

Suggest requiring `config.tmpDir === realpathSync(config.tmpDir)` at startup, or having the client send the realpath'd project dir.

render-service/src/resource-owner.mjs
```diff
@@ -0,0 +1,177 @@
...
+          code: cancelled ? 'cancelled' : expired ? 'deadline_exceeded' : 'execution_failed',
+          message:
+            expired && !cancelled ? 'Render exceeded the deadline' : String(error).slice(-8192),
+        },
+        resources: {
+          published: details.published === true,
+          cleanupVerified: details.cleanupVerified === true,
+          reservationReturned: details.reservationReturned === true,
+          admissionClosed:
+            producer.status().closed || (invoked && !(error instanceof BudgetedRenderError)),
```
cosarah, line 110: **Pre-merge (availability):** Any unexpected (non-`BudgetedRenderError`) producer failure after invocation — a plain TypeError, ENOSPC, etc. — reports `admissionClosed: true` even though `producer.status().closed` is false, and the client then sets `available = false` forever (only the one-time `ready` handler ever sets it back). One transient producer bug becomes a full rendering outage (429 `resource_unavailable` + `/health accepting: false`) until process restart, with no recovery path.

Fail-closed is a defensible default here, but consider distinguishing "closed" from "unknown", or at least documenting the restart-only recovery procedure.

render-service/src/resource-client.ts
```diff
@@ -0,0 +1,223 @@
...
+  constructor(
+    private readonly child: ChildProcess,
+    private readonly cleanupMs: number,
+  ) {
+    this.exited = new Promise((resolve) => child.once('exit', (code) => resolve(code)));
+    child.stderr?.on('data', (data: Buffer) => {
+      this.diagnostics = (this.diagnostics + data.toString()).slice(-8192);
+    });
+    this.ready = new Promise((resolve, reject) => {
+      const timer = setTimeout(() => {
```
cosarah, line 48: **Nit:** The hard-coded 10s ready timeout races the owner's pre-ready work — `verifyResourcePackage` hashes every file under the producer install (multi-MB native binaries included) before sending `ready`. A cold page cache or IO pressure at container start fails an otherwise healthy boot. Consider scaling the timeout from `cleanupTimeoutMs` or making it configurable.

render-service/src/resource-client.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  private fail(): void {
+    this.available = false;
+    this.terminal = true;
+    for (const item of this.pending.values())
+      item.finish({
+        status: 'failed',
+        failure: {
+          code: 'execution_failed',
+          message: `Resource owner lost; platform cleanup required. ${this.diagnostics}`,
```
cosarah: **Nit (info leak):** This message embeds the last 8KB of the root owner's raw stderr and ends up verbatim in `job.error` on the job-status route — cgroup/accounting paths and stack traces from a root process exposed to anyone who can poll the jobId. Suggest stripping or redacting the diagnostics from the client-visible message (keep them in service logs).

render-service/src/resource-client.ts
```diff
@@ -0,0 +1,223 @@
...
+      };
+    if (request.chunkExecution)
+      return {
+        status: 'failed',
+        failure: {
+          code: 'execution_failed',
+          message: 'Budgeted rendering supports local file output, not chunk execution',
+        },
+      };
+    await request.onProgress({ progress: 0, stage: 'preparing' });
```
cosarah, line 157: **Nit:** The IPC protocol has no progress channel, so a resource-mode render reports `{progress: 0, stage: 'preparing'}` once and then freezes until the terminal flip — no `framesRendered`/`totalFrames`/`metrics`. The app-side poller (3s interval) and any stage-stagnation heuristic will read this as a wedged job. Documented in `producer-patch/README.md`, but worth a note in the main README too, or a synthetic `rendering` stage.

cosarah (commented):
Reviewed the service-side diff in the worktree (typecheck clean; full `npm test` passes locally: 201 passed / 4 skipped).

Overall: very well engineered — the fail-closed settlement contract, supply-chain anchoring and test coverage are well above par, and the PR description is honest about boundaries. Three findings below are worth addressing before merge; the rest are nits.

**Suggested before merge (inline):**
1. Settings-swap TOCTOU in the root owner (`resource-main.ts`)
2. Symlinked `PRODUCER_TMP_PROJECT_DIR` makes 100% of renders fail while `/health` stays green (`resource-owner.mjs`)
3. Any unexpected producer error permanently closes admission until restart (`resource-owner.mjs`)

**Nits:**
- The README's exhaustive `POST /render` 429 `reason` contract (`render-service/README.md:36-42`) still lists only `queue_full` / `per_identity_limit`; the new `resource_unavailable` emitted from `reserve()` and `submit()` is undocumented. One-line doc fix.
- `npm run start:resources` cannot work in the shipped image: the Dockerfile runner stage copies only `src/`, `package.json`, `tsconfig.json` — not `scripts/` or `producer-patch/`, which `resource-owner.mjs` imports (`../scripts/resource-package.mjs` → `../producer-patch/source.json`). The path is documented as non-Docker, but nothing detects or prevents the attempt; copying the two paths or failing fast with a clear message would help.

render-service/src/render-coordinator.ts
```diff
@@ -362,6 +379,13 @@ export class RenderCoordinator {
         });
       }, abort.signal);
 
+      resources = result.resources;
+      if (
+        result.resources &&
+        (!result.resources.cleanupVerified || !result.resources.reservationReturned)
+      ) {
+        this.retainedProjects.add(projectDir);
```
cosarah, line 387: **Nit (ops):** Quarantined jobs have no reclamation path at all: `reservationReturned === false` records survive every TTL sweep, `onReap` never fires so artifacts are never removed, and `retainedProjects` only ever grows — and it is process memory, so retained dirs are orphaned across restarts. Intentional per the platform-takeover boundary, but an unbounded growth leak in practice; consider a bounded marker plus an operator reclamation hook, or at least expose a quarantined-count metric.

render-service/src/job-store.ts
```diff
@@ -72,7 +72,11 @@ export class InMemoryJobStore implements JobStore {
   private sweep(): void {
     const now = Date.now();
     for (const [id, job] of this.jobs) {
-      if (isTerminal(job.status) && now - job.updatedAtMs > this.ttlMs) {
+      if (
+        isTerminal(job.status) &&
+        job.resources?.reservationReturned !== false &&
```
cosarah, line 78: **Nit (latent):** The sweep retains on `reservationReturned === false` only, while the coordinator quarantines on `!cleanupVerified || !reservationReturned`. A `cleanupVerified: false / reservationReturned: true` settlement would retain the project dir forever while its record — the only thing pointing at it — gets reaped (404 + an untracked dir on disk). Current producer paths always emit the two flags together (`supervisor.ts` sets both to `clean`), so it is not reachable today; a shared `retainsResources(settlement)` helper used by both sites would make the invariant explicit.

LING-6150:
@cosarah Thanks for your detailed review! I’ll work through the concrete correctness and security findings first, with focused regression tests. For the recovery and quarantine suggestions, I’ll clarify the intended fail-closed and platform-takeover boundaries and coordinate with the maintainer before expanding the scope. I’ll follow up here with the changes and verification results. Thanks again.

LING-6150:
@cosarah Thanks for the detailed review! I’ve pushed the fixes in c6515f5e.
- Both privileged entry points now validate the settings file and its full parent-directory chain before reading.
- Invalid project roots fail at startup; equivalent normalized path spellings remain supported.
- Owner diagnostics go directly to service logs, while public task errors use safe summaries.
- TTL retention now matches the cleanup/reservation conditions. I also simplified the client to a single pending request and removed the redundant result-time closed notification.
Unknown settlement still fails closed; I documented the required platform cleanup/takeover before recovery rather than adding automatic reopening. The README now also explains resource_unavailable, progress limitations, and the unsupported Docker resource entrypoint.
Typecheck and 84 focused tests passed locally. On the company Linux VM, 12 startup-boundary checks and six HTTP lifecycle cases passed, including normal/recovery exports, cancellation, deadline, S/G death, and ancestor pressure. Evidence was returned and verified, and runtime cleanup completed.
The Producer patch, dependencies, budgets, and deadlines are unchanged. Please take another look when convenient—thanks!

render-service/src/resource-client.ts
```diff
@@ -0,0 +1,218 @@
...
+          typeof message.id === 'string' &&
+          result(message.result)
+        ) {
+          const item = this.pending;
+          if (!item || item.id !== message.id) {
+            this.fail();
+            return;
+          }
+          if (
+            message.result.resources?.admissionClosed ||
```
cosarah, line 87: **Judgement call (non-blocking):** This "unsettled" predicate now exists in three places with drifting member sets — here (includes `admissionClosed`), `job-store.ts:75-76`, and `render-coordinator.ts:382-384`. A future edit to one will silently diverge from the others. Consider a single `isSettlementIncomplete(settlement)` helper next to `RenderResourceSettlement` in `types.ts` that all three call.

render-service/src/resource-client.ts
```diff
@@ -0,0 +1,218 @@
...
+        this.fail();
+      }, 10_000);
+      const failReady = () => {
+        clearTimeout(timer);
+        reject(new Error('Resource owner exited before ready; see service logs'));
+      };
+      child.once('exit', failReady);
+      child.once('error', failReady);
+      child.on('message', (message: unknown) => {
+        if (!record(message) || Buffer.byteLength(JSON.stringify(message)) > 1024 * 1024) {
```
cosarah, line 59: **Low severity:** A legitimately large `details` payload (per-control-file accounting capped at 64 KB each, plus ancestor-pressure text for up to 64 ancestors × 5 files) is classified the same as owner loss (`this.fail()` → `published: 'unknown'`). Unlikely at realistic cgroup depth, but the failure mode is misclassification rather than truncation. Consider truncating `details` before send, or distinguishing oversize from process exit.

render-service/src/resource-client.ts
```diff
@@ -0,0 +1,218 @@
...
+      };
+    if (request.chunkExecution)
+      return {
+        status: 'failed',
+        failure: {
+          code: 'execution_failed',
+          message: 'Budgeted rendering supports local file output, not chunk execution',
+        },
+      };
+    await request.onProgress({ progress: 0, stage: 'preparing' });
```
cosarah, line 157: **Medium (UX), non-blocking:** The resource path emits exactly one progress event (`preparing`/0) and the worker protocol carries no progress messages, so SSE degenerates to a terminal notification. For scale, the 1080p full-classroom sample in #1091 measured ~697s — an opt-in user would sit at 0% for ~12 minutes. This is documented in `producer-patch/README.md`; suggest a follow-up issue for worker→service progress messages before this mode is production-recommended.

render-service/test/resource-client.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    await dispatched(sent);
+    child.emit('exit', 1);
+    expect(await promise).toMatchObject({
+      status: 'failed',
+      resources: { cleanupVerified: false, reservationReturned: false, admissionClosed: true },
+    });
+    expect(owner.accepting()).toBe(false);
+  });
+
+  it('keeps a published artifact distinct from unreturned reservation', async () => {
```
cosarah, line 91: **Medium (contract wording) — maintainer decision point:** The PR description states "Publication and reservation return remain blocked when cleanup cannot be verified", but after `transaction.commit` a failed cleanup re-check still yields `status: 'succeeded'` + `published: true` — only the reservation is quarantined and the owner stops accepting. The containment itself looks right (halted admission, no falsely-returned reservation, valid MP4 — the leak is disk/budget, not artifact corruption), but the spec sentence overstates the guarantee for this window. Either soften the PR wording to match, or re-verify cleanup before publish if strict pre-publication blocking is intended.

render-service/src/resource-main.ts
```diff
@@ -0,0 +1,84 @@
...
+import { readResourceSettings, assertCanonicalProjectRoot } from './resource-settings.mjs';
+
+if (process.platform !== 'linux' || process.getuid?.() !== 0)
+  throw new Error('Resource bootstrap requires Linux root');
+if (config.chunkExecutionEnabled)
+  throw new Error('Budgeted rendering does not support chunk execution');
+assertCanonicalProjectRoot(config.tmpDir);
+const settingsPath = resolve(process.argv[2] ?? '');
+const value = readResourceSettings(settingsPath);
+if (
```
cosarah, line 17: **Judgement call (non-blocking):** ~30 lines of manual `'workerUid' in owner && typeof owner.workerUid === 'number' …` checks over untyped JSON, and `workerUid`/`workerGid`/`taskPidsMax`/`cleanupTimeoutMs` always travel together. A typed parse-and-validate in `resource-settings.mjs` (already imported by both privileged readers) would replace this cascade and give the clump a type.

.github/workflows/ci.yml
```diff
@@ -222,6 +222,27 @@ jobs:
...
+          node --input-type=module - <<'JS'
+          import { readFileSync, appendFileSync } from 'node:fs';
+          const { revision } = JSON.parse(readFileSync('render-service/producer-patch/source.json', 'utf8'));
+          if (!/^[a-f0-9]{40}$/.test(revision)) throw new Error('Invalid source revision');
+          appendFileSync(process.env.GITHUB_OUTPUT, `revision=${revision}\n`);
+          JS
+
+      - uses: actions/checkout@v5
+        with:
+          repository: heygen-com/hyperframes
```
cosarah, line 237: **Low, non-blocking:** CI now checks out the third-party `heygen-com/hyperframes` repo, so CI availability depends on that repo staying reachable (integrity is already covered by the SHA gates — a tampered source fails closed). Consider fetching a commit-SHA tarball instead of a full clone to shrink the external surface.

cosarah (commented):
Thanks for the thorough PR — the evidence chain and fault-injection coverage are impressive. Reviewed head `c6515f5e`; all findings below are **non-blocking**. Two explicit decision points for maintainers:

1. **Post-commit publication guarantee** (inline on `resource-client.test.ts`): the described "publication blocked when cleanup cannot be verified" only holds pre-commit. Either accept the implemented behavior (published artifact + quarantined reservation + halted admission) and soften the wording, or require cleanup re-verification before publish.
2. **Progress degradation in the resource path** (inline on `resource-client.ts`): a single `preparing`/0 event, so SSE is effectively terminal-only. Acceptable for an opt-in mode, but worth a follow-up issue before this mode is recommended for production.

Minor cleanups, fine as follow-ups (non-blocking):

- The `performance`/`metrics`/`resources` conditional spread repeats at four terminal-write sites in `render-coordinator.ts` (~326, 397, 410) — consider one summary object.
- Failure-result literals (`'Render cancelled'`, deadline prose, owner-unavailable text) are hand-built independently in `resource-client.ts`, `resource-owner.mjs`, and `render-coordinator.ts` — shared result factories in a `.mjs` both sides can import would keep them from drifting.
- Producer version `0.8.37` is pinned in three places (`main.ts`, `resource-owner.mjs`, `source.json`).
- `job-store.sweep()` now inspects settlement internals to decide reaping; that policy arguably belongs beside the settlement type rather than in the store.

LING-6150:
@cosarah Thanks for the review. Updated the README and PR description in eaf47a8f to clarify the pre-/post-commit cleanup guarantees. The publication contract and progress limitation remain explicit maintainer decision points. I’ve deferred the optional refactors to keep this revision focused; no runtime behavior changed

cosarah (approved):
Approving. `eaf47a8f` resolves the contract-wording concern: the README now states exactly what the tests verify (post-commit cleanup failure → published artifact stays available, reservation quarantined, admission closes), and the blocking guarantee is correctly scoped to pre-publication detection.

Remaining items are non-blocking and fine to handle after merge:

- Sync the PR description sentence ("Publication and reservation return remain blocked when cleanup cannot be verified") with the README's post-commit qualification, so readers of only the body aren't misled.
- Follow-up issue for progress events in the resource path — a single `preparing`/0 event means ~697s renders sit at 0% the whole way.
- Minor cleanups from the earlier round whenever convenient: the triplicated "unsettled" settlement predicate (`resource-client.ts` / `job-store.ts` / `render-coordinator.ts`), hand-built failure literals in three places, and the `0.8.37` version pin across three files.

cosarah (dismissed):
LGTM

LING-6150:
@wyuc Following our discussion, I’ve updated #1492 to move the resource fence into OpenMAIC’s outer layer and remove the HF source patch and private build chain.
It now uses the official, unmodified HF package. OpenMAIC handles admission, per-task CPU/memory/PID limits, cleanup, and publication/settlement. The scope remains a single service instance with one active task and a fixed budget.
The updated implementation has Linux lifecycle validation and focused regression coverage. It’s ready for your review when you have time.

wyuc (changes requested):
Thanks @LING-6150 — the redesign is a real improvement. Moving the fence to OpenMAIC's outer layer and dropping the HF source patch and private build chain is exactly the right shape, and several parts hold up well under independent review (two reviewers plus my own read of the core path, at f56bbe14):

- **Official package**: the `@hyperframes/producer@0.8.37` alias resolves to the npm registry tarball with a matching integrity hash; no patch-package, vendored copy or postinstall.
- **Admission**: no way found for two requests to both be admitted, and no double release (`reserve()` is synchronous, the client re-checks after the progress yield, the owner rejects an overlapping task). Concurrency is pinned to 1 and overrides are rejected.
- **Publication contract**: atomic rename with pre-commit fsync is intact; a published artifact survives a late cancel and a post-commit cleanup failure, and the README describes what the tests verify.
- **Fail-closed**: every unverified cleanup latches admission closed; limits sit on the transient unit with `KillMode=control-group`, so they cover the whole tree.
- Opt-in and neutral: default `npm start` is unchanged. Typecheck, render-service vitest (305 passed) and prettier all pass.

Requesting changes for one blocking issue, plus two I'd like addressed or explicitly scoped before merge.

### P1 — the published artifact is created by root with the worker-controlled file mode (blocking)

`render-service/src/resource-task-runner.mjs:308` runs as root and does `copyFileSync(candidate, config.stagePath, COPYFILE_EXCL)`; `resource-systemd-runner.mjs:418` then renames it to `output.mp4`. Node's copyfile applies the source's full `st_mode` to the destination (umask does not apply), and the source path is followed if it is a symlink. The candidate is written by the unprivileged worker, and the destination directory is worker-owned. Two consequences:

1. **Functional**: the unit runs with `UMask=0077` (`resource-systemd-runner.mjs:575`), so the candidate is 0600 and the published `output.mp4` ends up `root:root 0600`. The unprivileged HTTP process then cannot open it for download (`main.ts:525`). Unless I'm missing a later chown, the normal download path is broken in resource mode.
2. **Boundary**: because root copies whatever mode bits the worker set, the worker can influence privileged file attributes on a root-owned file it can then reach. Separately, the copy target is resolved by path inside a directory the worker controls, after validation. Both cross the root/worker boundary the PR sets out to keep.

Suggested fix: open the stage relative to a directory fd with `O_CREAT|O_EXCL|O_NOFOLLOW`, read the candidate with `O_NOFOLLOW` and copy bytes only, then `fchown` to the service/worker uid and `fchmod` to an explicit mode (e.g. 0644) — or perform the copy as the worker via `setpriv`. Please also add an actual download of the published artifact to the Linux lifecycle evidence, since that would have caught (1).

### P2 — `/preview` is outside the fence

In resource mode, `ChromiumPreviewRenderer` (`main.ts:422`) still runs Chromium inside the HTTP service with no CPU/memory/PID budget, and it keeps running after admission has closed for memory pressure or quarantine (`tryRunWithExecutionSlot` does not consult `executor.accepting()`). A heavy preview can OOM the service cgroup and take the root owner down with it, which then closes render admission until an audit. Either gate/disable preview in resource mode, or document it as out of scope.

### P2 — the closeout ordering is not covered by in-repo tests

Mutations to `runResourceTask` / the task-runner `main()` that remove the pre-publish cancel check, the deadline check, the procfs reference-scan gate, the descendant drain, or the cgroup-removal readback all left the suite green. Those are the guarantees the README leads with, and today they rest only on out-of-repo Linux evidence (and the latest delta hasn't been re-run on Linux per the description). `waitForStarted` already takes injected dependencies; the same pattern around the orchestration would let each of those gates be pinned by a unit test.

### Smaller items (non-blocking, fine as follow-ups)

- **Post-publication bookkeeping can delete the published file**: if `jobs.update` / `artifacts.put` throws after the owner reported `published: true`, the coordinator's catch (`render-coordinator.ts:423-437`) runs cleanup and removes `output.mp4`. Unreachable with the in-memory stores, but the store seams are meant for Redis/S3. Retaining the project when `result.resources?.published === true` closes it.
- **Launch failure closes admission permanently**: the `systemd-run` / startup-window catch (`resource-systemd-runner.mjs:646-676`) reports unverified cleanup and leaves the task dir even when no unit ever started, so a misconfiguration turns into "closed until manual audit" and a refused restart. Worth distinguishing "never started" and documenting the recovery step.
- `MAX_PIDS = 4096` in `resource-reference-scan.mjs` makes every scan fail on hosts with more processes, closing admission on the first render — worth a startup probe or a larger bound.
- Limits are never read back from the task cgroup; asserting `cpu.max` / `memory.max` / `pids.max` before spawning the worker would catch a non-delegated controller silently leaving the task unbounded.
- The client lifeline timer (`resource-client.ts:186`) can fire before the owner's worst-case stop bound, mislabelling a slow-but-clean settle as unknown and closing admission.
- The README should state the same-uid trust boundary (the worker shares a uid with the HTTP service).
- `coordinator.submit()`'s slot release on `resource_unavailable` is untested (production is covered by the route's release).
- The lockfile bumps shared puppeteer / devtools-protocol for the default install too; fine if intended, but worth a note.

Linux-only mechanisms (systemd units, cgroup accounting, mounts, procfs scan) were reviewed by reading the code only; everything else above was exercised locally.

LING-6150:
@wyuc Thanks for the careful review. I pushed the focused fixes in `6520244c` without restoring the HF patch or adding another resource owner.

The three merge-scope findings are addressed:

1. **Privileged artifact transfer:** the task runner now launches a fixed copy helper through `setpriv`, dropping to the configured worker UID/GID, clearing supplementary groups and capabilities, and enabling `no_new_privs` before either worker-controlled path is opened. The helper rejects symlinks/non-regular sources, creates the stage exclusively, copies bytes only, and fixes the result to worker ownership with mode `0600`. The copy child is included in the task lifecycle, and evidence is parsed only after stdio `close`.
2. **`/preview` outside the fence:** resource mode now returns `503 resource_mode_unsupported` before reading the request body or starting Chromium. The default executor path retains its existing preview behavior.
3. **Closeout coverage:** the production closeout sequence is now exercised directly. The tests pin cancellation/deadline checks, descendant drain, procfs reference scanning, transfer, unmount, and settlement behavior. `waitForStopped()` is also tested directly for the case where the unit has stopped but its cgroup still exists, so removing the cgroup readback makes the suite fail.

I also covered the related smaller items that were local to this change: post-publication bookkeeping failures no longer delete/relabel the committed artifact, `submit()` releases its slot on `resource_unavailable`, the shared-UID trust boundary and lockfile dependency changes are documented, and resource-mode preview behavior has route coverage. The remaining explicitly non-blocking hardening suggestions (for example the PID scan bound and lifeline timing) are not being folded into this PR.

Verification on the committed tree:

- Full render-service suite: **327 passed, 23 skipped**; typecheck and Prettier passed.
- Final focused run after the stdio timing regression was strengthened: **86 passed**; `git diff --check` passed.
- Linux, using the official unmodified HF `0.8.37` package and one frozen product-source manifest:
  - normal real export and HTTP download succeeded; stage/output were worker-owned `0600`, source and destination were on different filesystems, and downloaded bytes matched the published file;
  - exact executable/PID/task-cgroup checks observed both a Chrome renderer and the encoding FFmpeg;
  - a pre-existing stage symlink produced the expected refusal without modifying its target or publishing, followed by a successful export on the same owner;
  - cancellation after the media-process gate left no published/staged artifact and was followed by a successful export on the same owner;
  - final independent readback found no run-owned unit, cgroup, mount, or process residue.

Two earlier Linux records remain preserved as failures because the external harness first counted its own waiting `systemd-run` client as a leak, then tried to read a temporary project after confirmed cleanup had deleted it. Both stopped subsequent cases as intended; the harness-only assertions were corrected and the missing cases passed under fresh run IDs with an identical product file manifest. I am not counting those records as product PASSes or claiming a full matrix rerun.

The PR now points to `6520244c`. The render-service and CodeQL checks for this head are already green; the remaining CI jobs are still running. Please take another look when convenient.

render-service/src/resource-task-runner.mjs
```diff
@@ -0,0 +1,536 @@
...
+    stopping = true;
+    if (activeChild?.pid) {
+      try {
+        process.kill(activeChild.pid, 'SIGTERM');
+      } catch {}
+    }
+  };
+  process.once('SIGTERM', stop);
+  process.once('SIGINT', stop);
+  let status = 'failed';
```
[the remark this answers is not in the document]
LING-6150: Thanks. The redundant initialization is valid, but I would keep the explicit status = 'failed' in the catch path to preserve fail-closed semantics. If cleaned up, I’ll remove the initializer instead.

render-service/src/resource-task-runner.mjs
```diff
@@ -0,0 +1,536 @@
...
+    privateMounted = closeout.privateMounted;
+    Object.assign(details, closeout.details);
+  } catch (error) {
+    details.failure = bounded(error instanceof Error ? error.message : error);
+    const drained = await drain(group, config.cleanupTimeoutMs);
+    details.failureDrain = drained;
+    if (projectMounted) {
+      try {
+        command('/usr/bin/umount', [privateProject]);
+        projectMounted = false;
```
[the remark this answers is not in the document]
LING-6150: Confirmed. This assignment is not read after the successful cleanup-path unmount and can be removed without changing cleanup or settlement behavior.

render-service/src/resource-task-runner.mjs
```diff
@@ -0,0 +1,536 @@
...
+      } catch (unmountError) {
+        details.projectUnmountFailure = bounded(
+          unmountError instanceof Error ? unmountError.message : unmountError,
+        );
+      }
+    }
+    if (privateMounted) {
+      try {
+        command('/usr/bin/umount', [privateRoot]);
+        privateMounted = false;
```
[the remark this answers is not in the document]
LING-6150: Confirmed. This is also a dead assignment after successful unmount and has no effect on cleanup verification or settlement.

wyuc (changes requested):
Thanks @LING-6150 — 6520244c closes the original issues well, re-verified by two reviewers:

- **Artifact transfer (was P1)**: root no longer copies the candidate. The helper drops to the worker uid/gid through `setpriv` (groups, capabilities and `no_new_privs` handled) before opening either path, uses no-follow / exclusive opens, copies bytes only and fixes the mode. There's no root fallback path, and dropping the privilege flags, allowing symlink sources, or adding a root fallback on launch error each turn tests red. The result stays readable for download, since the service and the worker share a uid.
- **`/preview`** returns 503 `resource_mode_unsupported` in resource mode before touching the body, the gate or Chromium; the default path is unchanged.
- **Closeout coverage**: the deadline, reference-scan, cgroup-readback and `waitForStopped` gates are now pinned. (The pre-publish `signal.aborted` guard and one cgroup readback line are duplicated by earlier checks, so they only go red when removed together; the behaviour is still covered.)
- The post-publication bookkeeping fix works: a failing store write after commit no longer deletes the published artifact, and it's tested.

One new blocking issue, plus one small test gap. The first is older code the first round missed, not something this commit introduced:

### P1: root recursively deletes a worker-controlled path on the unpublished-settlement path

`settleUnpublished` → `removeAndVerify(stagePath)` (`render-service/src/resource-systemd-runner.mjs:459-463`, `:483-484`; reached on cancel, deadline, task failure and the `finalizePublication` failure path) runs `rmSync(stagePath, { recursive: true, force: true })` **as root**. `stagePath` lives inside the worker-owned project directory, so the worker controls what exists at that path. It can make it a directory (which also forces the copy to fail and routes the task into this cleanup), and a recursive, path-based delete over a tree the worker can modify concurrently is race-prone: entries can be swapped for symlinks mid-traversal, redirecting root's unlinks outside the tree. We reproduced the traversal race unprivileged against Node 22's `rmSync` (the same JS on Linux); reaching it through the root owner is from code reading.

Suggested fix: the stage is only ever meant to be a single regular file, so never delete it recursively as root. Either `unlink` it non-recursively (and treat "is a directory" or anything unexpected as unverified cleanup → quarantine and close admission), or delete it through the same privilege-dropped helper as the worker. A test with a directory planted at `stagePath` should pin that root doesn't recurse into it.

### P2: the descendant-drain gate still isn't pinned

In `render-service/test/resource-task-runner.test.ts` (the drain-failure case around L254-278), `scanReferences` is `vi.fn()`, which returns `undefined`. With the drain check at `resource-task-runner.mjs:302` removed, the *next* gate throws a `TypeError` and the test still passes, while a mutant with `scanReferences` returning `CLEAR` publishes `status: 'succeeded'` even though `descendantDrain.drained === false`. Returning `{ status: 'CLEAR', … }` from that fixture (or asserting the specific failure) would pin it. This is test-only; no production change is needed.

### Non-blocking follow-ups

- Root still resolves `projectDir` by path after validation (the `mount --bind` source at `resource-task-runner.mjs:415`, and the owner's directory open plus `rename` at `resource-systemd-runner.mjs:420-425`). Opening it once with `O_DIRECTORY|O_NOFOLLOW` and using fd-relative operations, or re-verifying `dev`/`ino`, would close the remaining path-swap window.
- Several fail-closed checks around the helper aren't pinned by tests: a root fallback on nonzero exit, the evidence identity/device/size checks, malformed evidence, the helper's own uid/groups self-check, and registering the copy child for cancel-kill.
- A launch failure (unit never started) still closes admission permanently and leaves the task directory; worth a "never started" distinction or an explicit README note.
- In the new coordinator catch branch, if the recovery `jobs.update` itself throws, it escapes `run()`, which `pump()` calls as `void this.run(next)`, causing an unhandled rejection; a `.catch` on the recovery write would keep the process alive.

Linux-only behaviour (setpriv, systemd, cgroups, mounts, procfs) was reviewed by reading the code; everything else was exercised locally.

LING-6150:
@wyuc Thanks for the detailed follow-up. Addressed in 5ffb9a1c:
- P1: Unpublished settlement now uses non-recursive unlink through a verified project-directory FD. If the directory identity cannot be confirmed, cleanup retains the task evidence and reservation and closes admission.
- P2: The drain-failure fixture now returns CLEAR from the subsequent reference check. Removing the drain gate makes the test fail.
- Also addressed the follow-ups: FD-based bind/publication, additional helper failure-path and production cancellation tests, containment of recovery-write failures, and documentation of the conservative launch-failure behavior.
Local validation: 352 tests passed, 23 skipped, plus typecheck and formatting checks. Targeted Linux validation passed for FD boundaries and path replacement, cancellation/deadline during the real copy helper, and two consecutive exports on the same owner. Final cleanup readback was clean. The full historical Linux matrix was not rerun.
The updated commit is ready for another look. Thank you!

wyuc (approved):
Re-reviewed 5ffb9a1c (two independent reviewers):

- **P1 fixed.** Unpublished settlement, publication and the bind mount now go through an `O_NOFOLLOW` project-directory fd, checked against the owner-recorded dev/ino. Stage cleanup is a non-recursive unlink and fails closed when identity can't be verified. We tried symlinks, hard links, rename-swap TOCTOU, `..`, concurrent same-name tasks and setpriv ordering, and found no way to make root delete or write outside the worker directory. Reverting the fix turns `never traverses a directory planted at the unpublished staging path` red.
- **P2 fixed.** With `scanReferences` returning CLEAR, removing the drain gate now makes the drain-failure test fail (checked by mutation).

Non-blocking follow-ups (P3), fine in a later PR:
- `stateRoot` ancestors aren't checked. If an operator places `stateRoot` under the worker-owned `projectRoot`, the worker can rename-swap it, and root's recursive `rmSync` of `task-<token>`/`private` would be back on a worker-controlled path. A startup assertion that the whole ancestor chain is root-owned and not worker-writable would close it.
- `Number.isSafeInteger` on dev/ino (`resource-systemd-runner.mjs` ~423) rejects 64-bit inode filesystems, which would make resource mode unusable there. Compare them as bigint (`fstatSync(..., {bigint:true})`).
- The README says the project is mounted read-only, but tasks can still write the project dir via the host path, plus `/tmp` and `/dev/shm`, none of which are budgeted. Add `PrivateTmp=yes` or soften the wording.
- Cleanup checks dev/ino but not uid (inode reuse; very hard to exploit).

Thanks @LING-6150, nice work on this one.

wyuc:
Closing out the open review threads before merge. We checked each one against the current head:

**cosarah's pre-merge items: all addressed**
- *Settings re-read by the forked owner:* `readResourceSettings` now requires the settings file **and every parent directory** to be root-owned, not group/other-writable, and not a symlink, and the owner goes through the same reader. A worker-writable settings directory is refused, which closes the swap window.
- *realpath mismatch black hole:* `resource-main.ts` asserts at startup that `config.tmpDir` is canonical (`assertCanonicalProjectRoot`). A symlinked tmp path now fails the boot loudly instead of admitting and then failing every job.
- *Unexpected producer failure closes admission:* kept fail-closed on purpose. The restart-only recovery (operator audit, then owner restart) is documented in `render-service/README.md` (~342-372).

**Non-blocking, tracked as follow-ups** (fine outside this PR):
- A progress channel for resource mode (it currently sits at `preparing`/0 until done); open a follow-up issue before recommending this mode for production.
- A bounded quarantine marker or metric plus an operator reclamation hook; share one `isSettlementIncomplete` / `retainsResources` helper across the three sites.
- Scale or configure the 10s ready timeout; redact root-owner stderr from the client-visible `job.error`; truncate oversize `details` instead of treating them as owner loss.
- Typed parse of the owner settings clump; fetch hyperframes as a SHA tarball in CI; the three code-quality "useless assignment" notes in `resource-task-runner.mjs`.
- Plus the P3s from my approval: the `stateRoot` ancestor check, bigint dev/ino, `PrivateTmp`, and the README read-only wording.

@LING-6150 could you open one follow-up issue collecting these? Resolving the threads and merging now. Thanks @cosarah for the thorough review.

PR #1498, opened 2026-09-14
fix(upload): resolve the generic Office MIME (application/vnd.ms-office) via filename extension

### LINKED ISSUE #1497: [Bug]: PPTX (and other OOXML) uploads rejected on Linux — old XDG MIME databases report application/vnd.ms-office
### Bug Description

Uploading a `.pptx` course material succeeds on macOS Chrome but is rejected on Kylin OS (Linux) Chrome with the toast **“当前解析器不支持该格式，请切换解析器或选择其他文件”** (“The current parser does not support this format — switch parsers or choose another file”). Same file, same server, same selected extractor (MinerU / AliDocMind — both list PPTX in their capability matrix).

**Root cause.** On Linux, Chrome derives `File.type` from the system XDG `shared-mime-info` database instead of a built-in table. Older or trimmed databases (as shipped on Kylin V10) map `*.pptx` / `*.docx` / `*.xlsx` to the generic OOXML container type `application/vnd.ms-office` rather than the concrete `application/vnd.openxmlformats-officedocument.presentationml.presentation`.

`normalizeDocumentMimeType()` in `lib/document/mime.ts` then routes that value into the wrong bucket:

1. Empty or zip-family MIME (`application/zip`, `application/octet-stream`, …) → **falls back to the file extension** ✓
2. Canonical or curated alias MIME → mapped ✓
3. Any other MIME → **returned verbatim; the extension is deliberately not trusted** (anti-spoofing design, pinned by `tests/document/mime.test.ts` “does not let an unknown MIME masquerade as a supported format”)

`application/vnd.ms-office` lands in bucket 3: it is neither in `GENERIC_DOCUMENT_MIME_TYPES` nor in any `aliasMimes`, so it survives normalization verbatim, fails the provider whitelist in `components/generation/generation-toolbar.tsx` (`isMimeSupportedByProviders` on `{ file.type, file.name }`), and the toast fires. The file picker still offers `.pptx` (the input's `accept` list contains the extension), so the rejection only appears after selection — which makes it look parser-related to users.

The same browser-reported MIME also reaches the server-side gate in `app/api/extract-document/route.ts` (`normalizeDocumentMimeType` on `documentFile.type` / `documentFile.name`), and the workbench material path validates the `Content-Type` header against its own policy (`app/api/materials/route.ts` → `isWorkbenchMaterialMime`, 415 `unsupported_mime`), so both paths share the exposure.

### Steps to Reproduce

1. On a Linux desktop whose MIME database maps OOXML extensions to the generic type (reproduced on Kylin OS V10 — `xdg-mime query filetype slides.pptx` returns `application/vnd.ms-office`), open Chrome.
2. Open the course generation page with an extractor that supports PPTX selected (e.g. MinerU).
3. Upload a `.pptx` via the course-material paperclip (file picker or drag-and-drop).
4. The error toast appears immediately; no request leaves the browser.

Quick check of what the browser reports (any page, DevTools console):

```js
const i = document.createElement('input'); i.type = 'file';
i.onchange = () => console.log(i.files[0].type, i.files[0].name); i.click();
```

On Kylin this prints `application/vnd.ms-office slides.pptx`; on macOS/Windows Chrome it prints the concrete `...presentationml.presentation` MIME, which is why the same upload works there.

### Expected Behavior

When the browser reports a generic Office MIME, the file extension should resolve the concrete format — exactly as the existing fallback already does for `application/zip` and empty types. `.pptx` uploads should work identically on every OS.

### Actual Behavior

The upload is rejected client-side with the “unsupported format” toast because the generic OOXML MIME is passed through verbatim and never normalized to the concrete format.

### Deployment Method

Any — the defect is OS-dependent **client** behavior (observed against a self-hosted deployment; the server plays no role in the rejection).

### OpenMAIC Version

main @ `98765db8` (v1.0.2)

### Affected Area

Classroom generation

### Browser

Chrome (Linux build)

### Operating System

Kylin OS V10; expected on any distro shipping an older `shared-mime-info`. macOS and Windows unaffected.

### Relevant Logs

None server-side — the rejection happens in the browser before any request. Bypassing the UI check would hit the same rejection at the `extract-document` API layer.

### Screenshots or Recordings

–

### Additional Context

**Suggested fix:** treat `application/vnd.ms-office` as a generic upload fallback by adding it to `GENERIC_DOCUMENT_MIME_TYPES` (`lib/document/mime.ts`). The semantics match the existing zip-family entries exactly: the MIME says “some Office file” and carries no format specificity, so the extension decides. No new spoofing surface relative to `application/zip` (a file can already be renamed to `.pptx` with a zip/empty MIME; downstream parsers still validate real content, and unknown extensions still fall through verbatim and stay rejected).

⚠️ Do **not** add it to per-format `aliasMimes` instead: `canonicalFromAlias` iterates `DOCUMENT_FORMATS` in declaration order (`doc` first), so one shared alias would misroute `x.pptx` → `application/msword`, which self-hosted MinerU rejects — worse than the status quo.

Two follow-ups worth deciding alongside:
- `lib/workbench/material-upload-policy.ts` + `app/api/materials/route.ts` need the same value if Linux clients matter on the workbench path (today they get a 415 `unsupported_mime`).
- Kylin boxes with WPS Office installed may report WPS-registered MIME values instead; worth collecting real-world values before deciding whether any deserve the same treatment.

cosarah:
## Summary

On Linux, Chrome derives `File.type` from the XDG shared-mime-info database. Older or trimmed databases (e.g. Kylin OS V10) map every OOXML extension to the generic container `application/vnd.ms-office` instead of the concrete format MIME, so `.pptx`/`.docx`/`.xlsx` uploads were rejected with “当前解析器不支持该格式，请切换解析器或选择其他文件” on the generation toolbar and with `415 unsupported_mime` on `POST /api/materials` — while the same files upload fine on macOS/Windows Chrome.

Closes #1497.

## What changed

Treat `application/vnd.ms-office` as a generic upload fallback alongside the zip family, so the filename extension decides the concrete format:

- **`lib/document/mime.ts`** — add it to `GENERIC_DOCUMENT_MIME_TYPES`. One change covers the toolbar's provider-whitelist check, the `extract-document` API gate, and the classic-mode path in `app/page.tsx` (all call `normalizeDocumentMimeType`). It deliberately does **not** go into per-format `aliasMimes`: `canonicalFromAlias` iterates `DOCUMENT_FORMATS` in declaration order, so a shared alias would misroute `x.pptx` → `application/msword`.
- **`lib/workbench/material-upload-policy.ts`** — new `resolveWorkbenchMaterialMime()` grants missing / `octet-stream` / zip-family / `vnd.ms-office` types the same extension fallback the document path already had. (The workbench gate previously rejected empty and zip-family types outright on **every** OS — a latent bug on the same path, fixed by the same mechanism.)
- **`app/api/materials/route.ts`** — gate, store, and respond with the resolved MIME; the filename header is read before the MIME gate so error precedence is unchanged (415 still fires before the missing-filename 400).
- **`lib/workbench/session-store.ts`** — the upload request now sends the resolved concrete MIME instead of the raw browser value.

## Security

The spoofing surface is unchanged: a specific but unsupported MIME (e.g. `application/x-unknown`) still falls through verbatim for the whitelist to reject — pinned by the pre-existing 415 route test — and a generic MIME with an unknown extension keeps the generic value and is rejected too. Downstream parsers still validate real content.

## Test plan

- [x] `tests/document/mime.test.ts` — OOXML extensions resolve via the fallback; `.ppt` maps to its own canonical MIME; provider whitelists accept `{vnd.ms-office, slides.pptx}` for mineru; unknown extension keeps the generic MIME and stays rejected
- [x] `tests/workbench/material-upload-policy.test.ts` (new) — alias normalization preserved; all four generic families fall back to the extension; `x-unknown` + `lesson.pdf` must not masquerade
- [x] `tests/workbench/material-client.test.ts` — the upload request header carries the resolved concrete MIME
- [x] `tests/agent-runtime/materials-route.test.ts` — `vnd.ms-office` + `slides.pptx` passes the gate and drives the reservation/stored bytes; `vnd.ms-office` + `blob.bin` still 415s
- [x] Full suite: `pnpm test` → 7992 passed / 0 failed; `prettier . --check`, `tsc --noEmit`, eslint clean

## Kylin + WPS Office follow-up

Real-device testing on Kylin OS with Chrome and WPS Office installed showed that Chrome reports `.pptx` uploads as `application/wps-office.pptx`, rather than the standard OOXML MIME or the generic `application/vnd.ms-office` value covered by the original change.

This PR now also maps the format-specific WPS MIME aliases to their canonical OOXML types:

- `application/wps-office.pptx` → PPTX
- `application/wps-office.docx` → DOCX
- `application/wps-office.xlsx` → XLSX

The aliases are handled in both the classroom document-upload path and the workbench material-upload path. Unknown `application/wps-office.*` values remain rejected, preserving the existing anti-spoofing boundary.

### Real-device verification

- Environment: Kylin OS, Chrome, WPS Office installed
- Observed browser MIME: `application/wps-office.pptx`
- Selected extractor: self-hosted MinerU
- Result: PPTX upload accepted and parsing completed successfully
- Automated verification: 49 related tests passed; production build passed


wyuc (approved):
Read through the resolution logic: `application/vnd.ms-office` joins the generic set and falls back to the extension like the zip family, the WPS aliases map to their concrete OOXML types, and an unknown extension keeps the generic value so the 415 boundary is unchanged. Tests cover all four paths and CI is green on the updated branch. One non-blocking note: the workbench now carries its own extension→MIME table next to the document-path one; worth unifying later so they cannot drift. Merging.

PR #1499, opened 2026-09-14
fix(storage): keep jsonb writes valid when model output contains NUL or lone surrogates

### LINKED ISSUE #1493: jsonb writes fail with "unsupported Unicode escape sequence" when model output contains U+0000 or a lone UTF-16 surrogate
### Bug Description

A single NUL character or lone UTF-16 surrogate in an agent response aborts the whole session run, losing every lesson generated so far in that run.

`PgAgentSessionStore.appendTreeEntry` and the event projection in `packages/@openmaic/storage/src/agent-session/pg.ts` pass `JSON.stringify(...)` output straight into a `jsonb` parameter (`encodeJson`). PostgreSQL rejects two escape families that `JSON.stringify` emits verbatim, with SQLSTATE 22P05:

| escape in serialized JSON | PostgreSQL jsonb | note |
|---|---|---|
| `\u0000` (NUL) | **rejected**, 22P05 | occasionally appears in model output |
| lone UTF-16 surrogate (`\ud800`) | **rejected**, 22P05 | half a character; often a truncated emoji |
| `\u0001`, `\u007f` and other control chars | accepted | no handling needed |
| paired surrogate (`\ud83d\ude00`, i.e. emoji) | accepted | must **not** be stripped |

Because the failing write is treated as a *critical* entry write, the runner aborts the run and the session ends `failed` — in our case destroying 3–4 generated lessons per occurrence, repeatedly, during batch course generation.

### Steps to Reproduce

1. Run the Pro workbench against a PostgreSQL-backed agent session store.
2. Make a session whose model output contains a NUL character or a lone surrogate (deterministic version: send a user message whose content includes `\u0000`).
3. `server.log` then shows:

```
[ERROR] [AgentRunner] session <id>: event write failed error: unsupported Unicode escape sequence
[ERROR] [AgentRunner] session <id>: entry write failed error: unsupported Unicode escape sequence
[ERROR] [AgentRunner] session <id>: failed error: unsupported Unicode escape sequence
```

Direct proof that these escapes are what PostgreSQL refuses:

```sql
select '{"t":"a\\u0000b"}'::jsonb;        -- ERROR: unsupported Unicode escape sequence (22P05)
select '{"t":"a\\ud800b"}'::jsonb;        -- ERROR: unsupported Unicode escape sequence (22P05)
select '{"t":"\\ud83d\\ude00"}'::jsonb;   -- OK
```

### Expected Behavior

Session entries whose text contains such characters should still persist — the offending code unit dropped or replaced with U+FFFD — instead of failing the entire run.

### Actual Behavior

The `jsonb` write raises `unsupported Unicode escape sequence`; the runner treats it as a fatal critical-write error and fails the session, discarding all in-flight work for that run.

### Deployment Method

Local / self-hosted `next start` on a standalone build (`output: 'standalone'`), with the embedded PostgreSQL 16.11 managed by the repo.

### OpenMAIC Version

`1.0.2` (commit `98765db`). Also present in `1.0.0` (commit `1e10f60`); `CHANGELOG.md` does not mention it and I found no existing issue.

### Affected Area

Model / provider integration → actually **Agent runtime / storage**: `@openmaic/storage` PostgreSQL backends.

### Browser

Not applicable (server-side).

### Operating System

macOS 27.0 (arm64), Node v24.18.1.

### Relevant Logs

See the three `[AgentRunner]` lines above; `server.log` also records the PostgreSQL statement context for the failed parameter.

### Suggested Fix

Sanitize **string values while serializing**, via the `JSON.stringify` replacer. Four `jsonb` write helpers have the same blind spot — three `encodeJson` (`packages/@openmaic/storage/src/{agent-session,document,runtime}/pg.ts`) plus `encodeMeta` (`packages/@openmaic/storage/src/asset/pg.ts`):

```ts
function sanitizeJsonString(_key: string, value: unknown): unknown {
  if (typeof value !== 'string') return value;
  let out = '';
  let last = 0;
  for (let i = 0; i < value.length; i += 1) {
    const code = value.charCodeAt(i);
    let drop = false;
    if (code === 0) drop = true;
    else if (code >= 0xd800 && code <= 0xdbff) {                  // high surrogate
      const next = value.charCodeAt(i + 1);
      if (next >= 0xdc00 && next <= 0xdfff) { i += 1; continue; }  // valid pair: keep
      drop = true;
    } else if (code >= 0xdc00 && code <= 0xdfff) drop = true;     // lone low surrogate
    if (drop) { out += value.slice(last, i); last = i + 1; }
  }
  return out + value.slice(last);
}

function encodeJson(value: unknown, label: string): string {
  try {
    const encoded = JSON.stringify(value === undefined ? null : value, sanitizeJsonString);
    if (encoded === undefined) throw new TypeError('value is not JSON-serializable');
    return encoded;
  } catch (error) {
    throw new Error(`@openmaic/storage: ${label} is not JSON-serializable`, { cause: error });
  }
}
```

Two notes for whoever picks this up, both learned the hard way:

- Do the check on the **in-memory string via a replacer**, not with a regex over the serialized JSON. A regex over serialized text cannot distinguish a real lone-surrogate escape from the literal characters `\` + `ud800` in the content, and stripping those produces invalid JSON (22P02) — a worse failure than the original bug. We hit exactly that with a first attempt.
- `\u0000` can alternatively be replaced with U+FFFD instead of dropped, if you prefer keeping a placeholder.

wyuc:
## Summary

PostgreSQL `jsonb` rejects the `` and lone UTF-16 surrogate escapes that `JSON.stringify` emits verbatim (SQLSTATE 22P05 / 22P02). In the agent-session store the failed tree-entry or event write is treated as fatal, so one NUL or unpaired surrogate in a model response aborted the whole run and lost every lesson generated so far.

Fixes #1493

## Changes

- New shared `encodeJson` in `packages/@openmaic/storage/src/pg-json.ts`: replaces U+0000 and unpaired surrogates with U+FFFD before `JSON.stringify`, in values and object keys, while preserving valid surrogate pairs (emoji survive). Literal backslash-u text is untouched.
- Every `jsonb` parameter in the storage package now goes through it (agent-session, document, runtime, asset, material backends), plus owner-material registration in `lib/persistence`. The document/runtime/asset paths keep their existing `assertJsonValue` guard, which already failed loudly; the agent-session and material paths had no guard and were the live exposure.
- Exported as the `./pg-json` subpath of `@openmaic/storage`.

Not covered (follow-up): session title / prompt / error are plain `TEXT` columns, where a NUL fails with 22021 before any jsonb is involved.

## Tests

- `test/pg-json.test.ts`: NUL, lone high/low, leading/trailing lone surrogates, valid and adjacent pairs, lone high before a valid pair, nested objects/arrays, keys as well as values, literal `\ud800` / `` text unchanged, `undefined` → `null`.
- PGlite regressions: tree-entry and control-event writes with NUL and a lone surrogate now succeed (they reproduced 22P05 / 22P02 on `main`); material extraction stats and owner-material registration covered the same way.
- Storage package: 1051 passed (PG-contract suites run in CI); root `tsc`, eslint and prettier clean.


LING-6150 (changes requested):
@wyuc The string-value sanitization fixes the reproduced JSONB failure, but I found two data-loss cases in sanitizeJsonValue that should be addressed before merging:
1. Sanitized keys can collide and silently overwrite an entire member. For example, {"a\u0000":{"lesson":"first"},"a\uFFFD":{"lesson":"second"}} becomes {"a�":{"lesson":"second"}}. Please detect collisions and handle them explicitly rather than silently applying last-write-wins.
2. Rebuilding into {} drops a valid own __proto__ member. When another key requires sanitization, rebuilt["__proto__"] = value invokes the prototype setter instead of creating a JSON member. I reproduced this through PgAgentSessionStore append and reopen: the unrelated __proto__ field disappeared. Please use a null-prototype object or explicit own-property creation and add a persistence regression.
These are separate from replacing unsupported characters in text: they discard complete fields. The existing 15 helper tests pass, but both independent cases fail at a1b544f1.
Non-blocking scope note: #1493’s suggested NUL-bearing initial prompt still reaches a TEXT column and remains unsupported, as the PR already acknowledges. Please make that distinction explicit when closing the issue and link a follow-up if tracked separately.

wyuc:
Thanks @LING-6150 — both reproduced and fixed in 756a8e2a:

1. **Colliding sanitized keys.** The first member keeps the sanitized key; each later member whose sanitized form collides with any key emitted so far (sanitized or original) gets a deterministic ASCII suffix (`#2`, `#3`, …) until unique, so nothing is overwritten and member order is preserved. Your example now encodes as two members (`a<U+FFFD>` and `a<U+FFFD>#2`) with their own values; reverse order and three-way collisions are covered too, plus a PGlite append/reopen round-trip through `PgAgentSessionStore`.
2. **Own `__proto__` member.** The rebuilt object now has a null prototype, so `__proto__` stays an own data property; `JSON.stringify` emits it and PostgreSQL round-trips it. Unit test plus a persistence regression (tree entry with an own `__proto__` field next to a NUL-bearing field, reopened intact).

The no-op fast path still returns the original reference, and arrays are unchanged. Storage package suite: 1063 passed; root/package `tsc`, eslint and prettier clean.

On the scope note: agreed. The TEXT-column case (session title / prompt / error → 22021) is now tracked separately in #1502, and I will make that distinction explicit when closing #1493.

LING-6150 (approved):
@wyuc   Thanks, I rechecked 756a8e2 and verified that both previous blockers are resolved, including collision/suffix cases and preservation of an own __proto__ field through persistence. The explicit collision policy and separate TEXT-column follow-up in #1502 clarify the scope. No remaining blocking findings from this review. Approved

PR #1500, opened 2026-09-14
fix(agent-runtime): allow one owner material to be bound to multiple sessions

### LINKED ISSUE #1494: Binding the same owner material to a second session fails with a primary-key violation (material id is globally unique)
### Bug Description

An uploaded owner material can only ever be bound to **one** agent session. Binding the same material to a second session raises a duplicate-key error and the request answers HTTP 500, which makes an owner-level material library unusable for more than one course.

`bindOwnerMaterialsToSession` in `lib/server/agent-runtime/session-materials.ts` decides whether to create a row by looking it up with `(sessionId, id)`:

```ts
if (!(await store.getMaterial(sessionId, id))) {
  await store.createMaterial(sessionId, { id, ... });   // id = owner-side material id
}
```

but `agent_session_materials` has a primary key on `id` alone — verified against a live database with `pg_indexes`, and visible in the pinned DDL in `packages/@openmaic/storage/src/material/pg.ts`:

```sql
CREATE TABLE IF NOT EXISTS agent_session_materials (
  id            TEXT PRIMARY KEY,          -- globally unique
  session_id    TEXT NOT NULL REFERENCES agent_sessions(id) ON DELETE CASCADE,
  ...
```

So the de-duplication key and the actual uniqueness constraint disagree: the code asks "is this material bound to *this* session?" while the schema enforces "this material exists at most once, anywhere".

This blocks any workflow that reuses one upload across sessions — e.g. building a multi-part course series from the same textbook, or re-running a failed session with the same source material — which is the natural way to use an owner-level library.

### Steps to Reproduce

1. Upload one file: `POST /api/materials` → `materialId = mat_X`.
2. Create session A with `materialIds: ["mat_X"]` → 202, binding succeeds.
3. Create session B with `materialIds: ["mat_X"]`.
4. Observe HTTP 500 and:

```
[agent-runtime] request failed under an anonymous owner error:
duplicate key value violates unique constraint "agent_session_materials_pkey"
detail: 'Key (id)=(mat_X) already exists.'
table: 'agent_session_materials'
constraint: 'agent_session_materials_pkey'
```

(Reproduced repeatedly; the sessions were created through the workbench with the same uploaded PDF.)

### Expected Behavior

One of:

- binding is idempotent **per session**, so `mat_X` may be attached to sessions A and B (the row is session-scoped, and every read already goes through `getMaterial(sessionId, id)`), or
- reuse is rejected with a clear 4xx explaining that a material can only be attached once.

A 500 from an unhandled primary-key violation is the one outcome that is definitely not intended.

### Actual Behavior

HTTP 500 with `duplicate key value violates unique constraint "agent_session_materials_pkey"`; the session row is left behind by the soft-delete path.

### Deployment Method

Local / self-hosted `next start` on a standalone build (`output: 'standalone'`), with the embedded PostgreSQL 16.11 managed by the repo.

### OpenMAIC Version

`1.0.2` (commit `98765db`). Also present in `1.0.0` (commit `1e10f60`); no existing issue found.

### Affected Area

Storage / persistence (agent session materials) — plus agent runtime request handling.

### Browser

Chrome (latest) — though the failure is server-side.

### Operating System

macOS 27.0 (arm64), Node v24.18.1.

### Relevant Logs

See the `duplicate key value violates unique constraint` block above.

### Additional Context / Suggested Fix — needs a decision

Two directions with different blast radius, so this probably needs a maintainer's call before a PR:

1. **Make the row session-scoped** — primary key `(session_id, id)`, or keep a surrogate row id with a unique `(session_id, material_id)`. Cleanest for an owner-level library, but requires auditing every `getMaterial(sessionId, id)` caller plus the extraction pipeline, asset principals, and anything else that assumes the material id is globally unique.
2. **Keep it global, make the contract explicit** — generate a fresh session-side id at bind time and record the owner material id in a separate column; return a documented 4xx on reuse. Closer to the current code shape and to the extractor/derived-material model.

Option 1 matches what the calling code already assumes; option 2 is the smaller change.

**Workaround we are using meanwhile:** upload a separate copy of each file per session. The copies are hard-linked on disk (so almost no extra space) but each gets a new `materialId`, which means the per-owner material count limit (`MATERIALS_MAX_COUNT_PER_OWNER`, default 100) has to be raised to cover the duplicates. It works, but it is clearly a workaround rather than a fix.

wyuc:
## Summary

`agent_session_materials.id` is a global primary key, but `bindOwnerMaterialsToSession` de-duplicated on `(sessionId, id)` and then inserted the owner-side material id as the row id. Binding the same owner material to a second session raised a duplicate-key error (23505) that the route surfaced as HTTP 500, so an owner material could only ever be used by one course.

Fixes #1494

## Approach

Row ids stay globally unique: every extraction method (`enqueue` / `claim` / `heartbeat` / `complete` / `settle`) keys on `id` alone, and `derived_from` is a self-referential FK on `id`, so a composite `(session_id, id)` key would have to rework all of that plus a destructive primary-key migration. Instead the binder mints a fresh session row id and stores the shared owner id as metadata:

- New nullable `owner_material_id` column with a partial unique index on `(session_id, owner_material_id)`; added with `ADD COLUMN IF NOT EXISTS` + `CREATE UNIQUE INDEX IF NOT EXISTS`, so existing databases upgrade in place (legacy rows get `NULL`).
- `bindOwnerMaterialsToSession` looks the owner id up for idempotent rebinding; a concurrent rebind is adjudicated by the unique index and the loser adopts the winner's row.
- No FK from `owner_material_id` to the owner library, to keep owner-library deletion decoupled from session rows.
- `@openmaic/storage` 0.29.1 → 0.30.0 (new column, new read-model field, new concrete-store lookup).

Readers audited and unchanged: `getMaterial(sessionId, id)`, `listMaterials`, byte-store key derivation, the agent material tools, and the workbench round-trip (it only ever sends owner ids back).

## Tests

- Backend-neutral contract test: the same owner material bound to two sessions succeeds and both sessions can read it (this replaces the previous assertion that the same id cannot be re-created for another session).
- PGlite in-place upgrade test (drop the column, seed a legacy row, re-run the schema initializer).
- Pinned-schema contract updated for the new column/index.
- `tests/agent-runtime/owner-material-binding.test.ts`: host-level both-sessions test and a real-route 202 regression over PGlite.
- Root `tsc`, package typecheck, eslint, prettier clean; `check-package-version-bumps` passes.


LING-6150:
@wyuc The per-session row ID approach fixes the cross-session primary-key failure, but legacy bindings lose their existing idempotency after upgrade.
Rows created by the previous binder have id = ownerMaterialId and receive owner_material_id = NULL. The new lookup only checks owner_material_id, so rebinding the same upload to the same existing session creates another row and copies the bytes again. I reproduced this using the old binder followed by the new binder: one binding becomes two rows with different IDs.
Please recognize and reuse legacy bindings while preserving their row IDs and extraction state, through a validated fallback or backfill. Add a regression that creates an old-style source binding, rebinds it after upgrade, and asserts that the returned ID and row count remain unchanged. The current migration test checks an old web row’s readability, so it misses this case.
A separate cleanup issue: when concurrent binds race, the loser returns the winner before deleting its own uploaded bytes. I reproduced one database row with two session byte objects. Please clean up the losing object when its key differs from winner.rawAssetId, preserving bytes actually referenced by the winner.

wyuc:
Thanks @LING-6150 — both reproduced and fixed in 6a08c683:

1. **Legacy bindings keep their row.** When the `owner_material_id` lookup misses, the binder now falls back to the session row keyed by the owner upload id and adopts it only when it is unambiguously that binding: `kind = source`, `owner_material_id IS NULL`, matching title, no source URL / derivative / extracted text on the source row (extracted text lives on derived rows, so a finished extraction still qualifies), the deterministic legacy object key, and a byte length that matches the owner record. Adoption stamps `owner_material_id` through a conditional `backfillOwnerMaterialId` that leaves the extraction columns untouched; a lost race re-reads the fast path. Regression: an old-style source row with `extraction_status = done` is rebound after the initializer runs — same id, one row, no byte copy, extraction state intact, `owner_material_id` backfilled — and a bind to a different session still creates a fresh row.
2. **Losing upload is cleaned up.** After adopting the winner, the loser deletes its own object when the key differs from `winner.rawAssetId`, best-effort so a failed cleanup cannot fail a bind that succeeded. Regression: a controllable byte store parks the loser after its upload so the winner commits first; exactly one row and only the winner's object remain.

`@openmaic/storage` goes to 0.30.1 for the new public store method. Storage suite and the host-level binding tests are green; root/package `tsc`, eslint and prettier clean.

LING-6150 (dismissed):
Thanks , I rechecked 6a08c683 and verified that both previous blockers are resolved. Legacy rebinding preserves the original row and extraction state, and concurrent binds clean up the losing object without deleting winner-owned bytes.
The focused host/route and storage/schema tests passed locally. No remaining blocking findings from this review. Approved   @wyuc

PR #1510, opened 2026-09-15
fix(playback): stop superseded scene engines

### LINKED ISSUE #1509: Playback can continue after switching to a non-playable scene
## Bug description

When the classroom switches from a playable scene to a quiz, interactive, or PBL scene with no timeline actions, `PlaybackChromeRoot` clears its engine ref and returns without stopping the previous playback engine. A second race exists when `startLecture()` is still pending: its continuation can start or resume the captured engine after a scene switch has already superseded it.

This can leak stale narration, timers, effects, and progress updates from the previous scene into the new scene.

## Steps to reproduce

1. Open a classroom with a playable slide followed by a quiz/interactive/PBL scene that has no timeline actions.
2. Start playback on the slide.
3. Switch to the non-playable scene while playback is active, or while lecture-session creation is still pending.
4. Observe that the previous engine is not stopped and can continue after the asynchronous session creation resolves.

## Expected behavior

- Scene initialization detaches and stops the previous engine before any asynchronous scene cleanup.
- Callbacks from superseded engines are ignored.
- Pending manual and auto-play continuations do not restart an engine that no longer owns playback.

## Environment

- OpenMAIC `main` at `3edaa4995fd95b4f538abe4ba6421b75a40035b9`
- Reproduced with deterministic component tests under Vitest/jsdom.

Duang777:
## Summary

Stop stale classroom playback engines from surviving a scene switch or being restarted by an asynchronous lecture-session continuation. This prevents narration, timers, effects, and progress callbacks from the previous scene from leaking into the new scene.

This PR is AI-assisted. The final diff was self-reviewed for lifecycle ownership, async races, and sibling call sites before submission.

## Related Issues

Fixes #1509

## Changes

- Detach and stop the current engine before awaiting scene teardown, including when the destination is a non-playable or pending scene.
- Accept progress updates only from the currently installed engine.
- Recheck engine ownership after `startLecture()` in both manual and auto-play paths, and close sessions created for superseded engines.
- Add regression coverage for direct scene switches, delayed manual starts, and delayed auto-play starts.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Run `pnpm exec vitest run tests/components/edit/playback-chrome-root-element-reference-owner.test.ts`.
2. Confirm all 17 tests pass, including the three stale-engine regression cases.
3. Run `pnpm check`, `pnpm lint`, `pnpm exec tsc --noEmit`, and `pnpm check:i18n-keys`.

- Red-green coverage: before the fix, the non-playable switch test observed no `stop()`, and the delayed manual/auto paths resumed the superseded engine.
- After the fix, all 17 focused component tests pass.
- Prettier passes; changed-file ESLint passes with no warnings; TypeScript and i18n checks pass; package version gate passes.
- Sibling sweep covered all three `startLecture()` call sites in `PlaybackChromeRoot`; each now validates engine ownership after awaiting.
- Full `pnpm test` exposed 2 media-orchestrator assertion failures and 4 Quiz runtime timeouts. The exact same six failures reproduce when those files are run against a clean `main` worktree at `3edaa499`, so they are not introduced by this change.
- Browser end-to-end playback was not exercised because it requires a configured classroom/provider environment; the lifecycle races are covered deterministically in jsdom.

### Evidence

- [x] CI checks (`pnpm check`, `pnpm lint`, and `pnpm exec tsc --noEmit`) pass locally
- [x] Focused regression tests pass locally (17/17)
- [ ] Screenshots / recordings attached (not applicable: behavior-only lifecycle fix)

## Checklist

- [x] I have added/updated documentation as needed (no documentation change required)


Duang777:
Implementation note for commit `2605bba7`:

**Root cause**

`PlaybackChromeRoot` checked whether the destination scene had playable actions before stopping the existing engine. A switch to an actionless quiz/interactive/PBL scene therefore nulled the ref and returned while the old engine remained active. In addition, manual and auto-play flows awaited `startLecture()` without confirming that their captured engine still owned playback, allowing an already stopped engine to be restarted after the scene changed.

**Fix**

- Atomically detach the previous engine and call `stop()` at scene initialization entry.
- Reject every progress callback whose engine is no longer installed.
- Revalidate engine identity after asynchronous lecture-session creation in manual and auto-play flows.
- End a session created for a stale engine instead of allowing it to linger.

**Regression coverage**

The added tests deterministically cover:
1. playable -> non-playable scene switch stops the old engine and rejects stale progress;
2. a delayed manual `startLecture()` cannot resume a superseded engine;
3. a delayed auto-play `startLecture()` cannot start a superseded engine.

Focused result: 17/17 passing. Prettier, changed-file ESLint, TypeScript, i18n alignment, and package-version checks pass. The six unrelated failures seen in the full local suite were reproduced unchanged on clean `main` at `3edaa499` (2 media orchestrator assertions and 4 Quiz runtime timeouts). No published package or user-facing text is changed.

wyuc (approved):
Reviewed at `2605bba7`. Approving; the notes below are non-blocking.

**Verified**
- Walked the lifecycle matrix: playable→playable (playing and paused), playable→non-playable (the #1509 case), non-playable→playable, rapid A→B→A with a pending manual start, pending manual and auto `startLecture`, a live discussion in progress, same-scene re-render, StrictMode double effects, user pause/resume, unmount, and imperative `teardown()`. The previous engine is stopped exactly once on every scene initialization, and superseded continuations end their session instead of starting or resuming the old engine.
- Mutation: removing the up-front `previousEngine?.stop()`, restoring the `engineRef.current !== null &&` progress guard, or dropping either the manual or the auto superseded guard each turns a regression test red.
- `tests/playback`, `tests/lib/playback`, `tests/classroom`, `tests/components/edit` → 208 passed; `tests/chat`, `tests/lib/chat`, `tests/edit` → 685 passed. prettier / eslint / `tsc --noEmit` clean.

**Non-blocking notes**
1. **(P2, latent)** Unmount cleanup and `teardown()` call `engineRef.current?.stop()` but leave `engineRef.current` pointing at the stopped engine. A `startLecture` still pending at that moment then passes the `engineRef.current !== engine` guard and calls `continuePlayback()` / `start()` on it (reproduced with a deferred `startLecture` mock). It's unreachable today because `startLecture` never awaits, so the continuation runs within a microtask. It becomes live as soon as session creation gains real async I/O. Suggested fix: set `engineRef.current = null` right after both `stop()` calls, and add a "pending start + teardown/unmount" case.
2. **(P3)** Because `startLecture` reuses an active session for the same scene, a superseded continuation in an A→B→A sequence could `endSession` the session the new A engine has already adopted. Also unreachable while `startLecture` is synchronous; worth re-checking against `lectureSessionIdRef` once it goes async.
3. **(P3)** Only `onProgress` gained the identity guard. The other engine callbacks rely on `stop()` invalidating the playback generation. No leak observed, just noting the guard coverage is partial.

PR #1518, opened 2026-09-15
fix(pptx): preserve tab columns, text insets, and arrow rendering

xuyuanwei678:
## Summary
Fix PPTX import/rendering errors that shifted vertical labels, displayed Wingdings arrow bullets as a victory-hand emoji, and pushed tab-separated answer choices beyond the slide edge.

## Related Issues
User-reported reproduction in a 29-slide programming lesson deck (slides 21–22); no linked issue.

## Changes
- Protect ProseMirror caret separator images from host block-image CSS. Inline formulas at paragraph ends no longer introduce an extra line on entering editing; browser measurements show identical positions/heights for all six paragraphs and identical positions for all five formulas on the interest-theory slide.
- Preserve imported inline KaTeX formulas as atomic editor nodes, reconstructing trusted markup from LaTeX instead of flattening MathML, annotations and rendered glyphs into prose. Keep LaTeX in plain-text clipboard output. Verified all five formulas on slide 4 of the 28-slide interest-theory deck before/during editing and after typing. This preserves formulas while editing surrounding prose; it does not add an inline formula source editor.
- Classify only actual OOXML line presets as connectors. Preserve flowChartConnector circles and flowChartOffpageConnector area shapes, including gradient fills, through serialization and canvas conversion; retain lineInv inverse geometry. Verified both circles on slide 15 of the original deck.
- Add a narrowly scoped, empirical 2D preview for the verified clear/chilly/front-camera/relaxedInset combination on slide 12, restoring the pale gray WPS appearance. This is an editable color approximation, not a 3D lighting implementation; perspective, tilted/extruded shapes and other configurations retain their existing fills.
- Keep font-family outside font-size in editable text to prevent ~1px Latin baseline shifts. Treat fixed-width inline text boxes as editable nodes so mixed fonts do not duplicate their widths; preserve empty boxes through serialization.
- Preserve custom paragraph tab stops and inherited default tab spacing using fixed-width inline spans, keeping all answer choices within their intended columns.
- Measure text with browser font metrics and the emitted font styles; retain native tab handling for paragraphs containing OMML formulas and use approximate metrics for server-side imports.
- Preserve table paragraph spacing when entering cell editing, avoiding row-height redistribution. Browser verification showed zero height delta for all five rows of the demo table.
- Resolve containing paragraphs for list commands invoked inside tab columns, then restore the original caret/selection after the edit. Covers creating, removing, switching lists and cross-paragraph selections.
- Represent imported tab columns as editable inline nodes, preserving equal-width boundaries through round trips, partial formatting, and text insertion/deletion.
- Measure the final CSS font size, including reduced superscript/subscript sizes.
- Include bullet-slot width in the tab cursor and keep column text left-aligned inside centered/right-aligned paragraphs.
- Allow server-estimated columns to grow to their intrinsic text width to prevent overlap; preserve this minimum width and alignment through editor round trips. Server-side stop selection remains approximate.
- Bump importer to 0.2.2, renderer to 0.1.8, and editor to 0.0.6.
- Avoid adding renderer padding when imported text already includes PowerPoint text insets.
- Correct the Wingdings 0xD8 / F0D8 arrow mapping and calculate down-arrow head length from the shape's shorter dimension.
- Add regression coverage for tab columns, style overrides, formulas, text insets, and down-arrow geometry.

## Type of Change

### Steps to reproduce / test
1. Import the reported lesson PPTX into the local slide-renderer demo.
2. Inspect slide 21: priority-label alignment, down-arrow proportions, and arrow bullets.
3. Inspect slide 22: confirm D.20/3 and D.6 are fully visible in the option rows.
4. Run `pnpm --filter @openmaic/importer --filter @openmaic/renderer test`.

- Slide 10: default hyperlink underline is explicit in imported HTML; browser checks confirm it remains visible before, during and after editing. Explicit no-underline and strike-through combinations are covered.
- Slide 8: ordinary year text with a supplemental Wingdings font stays readable; default tab stops become editable columns. Browser checks confirmed 2022年 is present and A/B/C/D option x/y coordinates are identical before and during editing.
- Slides 6 and 7: recover saved font leading for simple one/two-paragraph auto-fit text overlays. Browser screenshots checked; entering edit mode preserves the slide 7 int paragraph top and height exactly. Added guards for mixed sizes, extra paragraph spacing, wrapped/oversized frames and longer bodies.
- Browser visual checks of slides 21 and 22 with the supplied deck.
- Importer: 157 tests passed across 24 files.
- Renderer: 56 tests passed across 15 files.
- Editor: 526 tests passed across 52 files.
- Importer build and local package sync succeeded during implementation; `git diff --check` passed.
- Repository formatting and TypeScript checks passed. ESLint completed with 0 errors and 18 existing warnings outside this diff. Package-version and internal-dependency checks passed.
- Independent follow-up verification: equal-width columns remain separate after editing; superscript AAAA measures about 46.24px and selects the 60px stop.
- Importer/editor builds passed. Independent Chrome reproduction checks confirmed bullet column x=100px, centered/right-aligned start-to-start spacing=100px, and no server-fallback text overlap. Latest review fixes additionally passed focused ESLint and package builds; CI for the new commit is pending.
- SmartArt operator colors remain a known WPS compatibility difference: the original file's styles and cached drawing specify white, and LibreOffice displays white, while WPS displays black. This PR does not change color inheritance. The local black-text compatibility deck and ignored demo files are not included.

- Chrome checks: slide 10 hyperlink baselines match before/during editing across six host fonts; slide 7 identifiers and slide 8 option coordinates remain stable. Caret boundary coordinates, relative font sizes, mixed-font box widths and empty-box round trips verified.

### Evidence
```text
@openmaic/importer: Test Files 24 passed; Tests 157 passed
@openmaic/renderer: Test Files 15 passed; Tests 56 passed
@openmaic/editor: Test Files 52 passed; Tests 526 passed
```
- [x] Manually tested locally

## Checklist


wyuc:
Reviewed `32a2c68c0e` (independent review, plus my own checks on the two findings below). No P0/P1: no data loss, no injection surface, and no backward-compatibility break. Two things look worth fixing before this lands, both small.

**Checked and fine**
- Tests pass as you report: importer 164, renderer 56, editor 573. prettier / eslint (0 errors) / typechecks / package-version and internal-dependency gates all clean; `--frozen-lockfile` install reproduces.
- Mutation-tested four core changes (the `0xd8` mapping, the server-side `min-width:max-content` fallback, the text-inset branch, and inline-container mark normalization) — each one turns its test red, so the new coverage is real rather than vacuous.
- **Schema backward compatibility holds.** Documents persist as HTML (no `Node.fromJSON` anywhere), so dropping the `inlineBlock` mark cannot break archived slides: legacy `display:inline-block;width:…` spans re-parse as `inline_text_box`, empty boxes as `inline_spacer`, and `data-inline-text-box` is purely additive. The legacy round-trip tests cover this.
- Server-estimated vs browser-measured tab columns produce structurally identical documents (same nodes, order and attributes; only the width numbers differ plus `min-width:max-content`), and `parseDOM` only reads `width`, so a server-imported deck cannot parse into a different structure in the browser.

**1. `clear` removes the inline container's own font context**
`commandExecutor.ts:34-48`. The loop deliberately excludes `fontname`/`fontsize` from `materializeInlineMark`, matching the comment ("Container font family/size are the local typography context…"), but the following `tr.removeMark($from.pos, $to.pos)` passes no mark type, so it strips every mark in range. When the range covers the container's start position — which `autoSelectAll` guarantees for the empty-caret toolbar path — the container's own `fontsize`/`fontname` go with it.

Repro: `<p><span style="font-size:2em"><span style="display:inline-block;width:10em">AB</span></span></p>`, caret inside, run `clear` → the container's marks go from `[["fontsize",{"fontsize":"2em"}]]` to `[]`, and it serializes back as `<span data-inline-text-box="true" style="display: inline-block; width: 10em;">AB</span>` with `font-size: 2em` gone. Since the width is in `em`, the box then resolves against a different font size and imported column widths shift. A narrow selection inside the container keeps the context, so the behavior differs between "select a character and clear" and "clear with an empty caret".

Suggested fix: remove `fontname`/`fontsize` only from leaf text nodes (or snapshot the container's two marks and re-add them after the untyped removal), and add a regression case for the empty-selection path — `inlineContainerCommands.test.ts:165` only covers a narrow selection inside the container.

**2. `mapRunSymbols` can now drop a real symbol glyph**
`textSerializer.ts:137-149`. On `main`, any run whose `a:sym` is a symbol font maps every character; here mapping is limited to U+F000–U+F0FF, or to runs whose `a:latin` is also a symbol font. Wingdings 3 arrows sit in the ASCII range (`{` = 0x7B = `▲`), so a run with `<a:latin typeface="Arial"/><a:sym typeface="Wingdings 3"/>` and text `{` now renders a literal `{`, where `main` rendered `▲`.

I agree with narrowing the old behavior — mapping everything is what turned `：（2022年）` into bullets, since unmapped characters fall back to `•`. But the third condition could simply be "the symbol table has an entry for this code point": map when the character is private-use, or the run is a legacy symbol run, or `table[code]` exists. I checked the tables for this: they hold 43/19/8/7 entries and contain no letters, digits or ASCII punctuation, so `：（2022年）` still passes untouched while `{` maps to `▲` again.

**Non-blocking notes**
- `BaseTextElement.tsx:20-22` decides "this content already has PowerPoint insets" from a regex on the first `<div>`'s first `style` attribute. Hand-authored or generated content starting with `<div style="padding: 20px">` loses the default 10px padding, and an import whose padding is written as `padding-block` (or not on the first element) is missed in the other direction. More robust: have the importer mark its wrapper explicitly (e.g. `data-text-inset="true"`) and let the renderer key off that instead of guessing.
- Server-side estimation plus `min-width:max-content` prevents overlap, but the tab cursor still advances on the estimated width, so later columns can drift cumulatively in the browser. Your comment already flags the trade-off; worth stating in the docs that server-imported tab positions are approximate until re-measured in a browser.
- `inline_math.toDOM` re-runs `katex.render` on every serialization, and `RendererTextEditor` serializes on every transaction, so formula-heavy decks pay that repeatedly. Also `attrs: { latex: {} }` has no default, so a node created without `latex` would throw in `toDOM`/`leafText`; a `''` default plus an empty check would make that safe.

Your browser observations against the real deck are the part I can't reproduce, so I've treated them as unverified rather than confirmed — the code-level findings above stand on their own.

xuyuanwei678:
Follow-up on finding 2: we constructed a seven-case PPTX and the user compared the original file in WPS with the importer. The screenshots distinguish two issues:

- **A: `a:latin=Arial`, `a:sym=Wingdings 3`, text U+007B (`{`)** renders as a literal `{` in both WPS and the importer. This sample does not support restoring the old conversion for ordinary ASCII merely because `a:sym` is present.
- **C: explicit Wingdings 3 + U+007B, and D: Arial + Wingdings 3 symbol font + U+F07B** do differ: WPS shows a filled triangle with its right angle at the upper right, whereas our mapping produces `▲`. This is a glyph-mapping error shared by the legacy-byte and private-use paths, separate from the conversion eligibility issue in A.
- The ordinary `n` and date controls remain ordinary text in both screenshots. Adding a “table contains this code point” condition would risk changing those ordinary letters: the existing Wingdings table includes `0x6e` (`n`), and Wingdings 3 includes ASCII punctuation such as `0x7b` (`{`).

I will correct the confirmed `0x7b` mapping to `◥` (U+25E5, BLACK UPPER RIGHT TRIANGLE), add regression coverage for both symbol paths and the ordinary-text controls, and retain the current conversion eligibility rules. No bundled font is needed for this change. This conclusion is limited to the supplied WPS comparison; it is not a claim that all Wingdings mappings have been validated.

Finding 1 (clear formatting removing the container font context) has already been fixed in 5e4909a9, with caret/full/cross-container selection and undo regression coverage.


### Screenshot evidence (before the mapping fix)

The following screenshots were supplied by the user. The importer screenshot predates `0f9a2685`; it is not an after-fix screenshot. That commit corrects the C/D mapping to `◥`; the conversion conditions are unchanged.

**Importer before fix: C/D show ▲**

Importer before fix: C/D show ▲

**WPS reference: C/D show upper-right right-angle triangles (highlighted)**

WPS reference: C/D show upper-right right-angle triangles (highlighted)

xuyuanwei678:
Finding 1 is fixed in **5e4909a9** (`fix(editor): preserve container font context when clearing formatting`).

The clear-format command now preserves the inline container's own `fontname` and `fontsize`, which provide the typography context for relative (`em`) widths. Font overrides are removed only from selected inline leaves; other formatting is still cleared. This covers the empty-caret toolbar path, where `autoSelectAll` expands the selection to the full document.

For font overrides, the implementation replaces only the selected leaf content instead of applying `removeMark` over that range. The latter also visits enclosing inline nodes and can restore container font marks onto descendants during undo.

Regression coverage includes:
- Empty caret / automatic select-all.
- Explicit full selection.
- Cross-container selection, including nested relative font contexts.
- Preserved container fonts and `10em` / `3em` widths, cleared text overrides, unchanged text, and exact document restoration on undo.

Validation at this commit: **576 editor tests passed across 55 files**; editor build, targeted ESLint, and `git diff --check` passed.

xuyuanwei678:
### Follow-up: explicit PPTX text-inset marker

Implemented in **845644b3** for the non-blocking text-inset detection comment:

- The importer emits `data-pptx-text-insets="true"` on the outer `div` when it writes the OOXML text-body padding. This includes explicit zero insets, so the renderer does not add its default padding back.
- The renderer recognizes the explicit marker on the outer wrapper first. For unmarked legacy content, it retains the existing outer-div `padding` detection to preserve saved slides' layout.
- The editor's text-container schema parses and serializes the marker, preserving it through text edits, save, and reopening. Canvas conversion also retains it.

Compatibility trade-off: the legacy fallback intentionally remains. Unmarked hand-authored HTML with outer-div padding can still match that fallback; this change establishes an explicit contract for new imports without reinterpreting old documents. It does not migrate all existing content or claim to eliminate the legacy ambiguity.

Validation:
- Importer: 170 tests passed in the full run, followed by a passing 4-test viewport suite including the additional marker-preservation case.
- Renderer: 60 tests passed, including explicit markers, zero padding, legacy content, and ignoring nested markers.
- Editor: 579 tests passed, including marker/padding preservation across editing and reparsing.
- All three package builds, targeted ESLint, Prettier, and `git diff --check` passed. Local vendor assets have been synchronized.

**Status: committed and pushed to this PR in 845644b3.** Local validation is listed above; CI results for this commit have not been awaited.

The separate server-side Tab limitation documentation and formula serialization performance/invalid-node handling follow-ups are not included in this commit and remain pending.

xuyuanwei678:
The remaining non-blocking follow-ups have been addressed in **566f6cca** (`fix(editor): guard formula serialization and document layout limits`).

### Server-side Tab layout

The importer README now documents that server-side text widths are estimates. `min-width:max-content` lets columns expand to reduce overlap, but subsequent stops still use estimated widths, so actual browser fonts can cause cumulative drift. Existing imported content is not automatically remeasured in the browser. Reimporting the original PPTX in the browser after target fonts are loaded is the current option; automatic remeasurement remains future work. No Tab layout algorithm was changed in this commit.

### Formula serialization safety

- Added an empty default for `latex` and normalized missing/non-string source to an empty formula atom.
- Preserve empty formula markers through serialization and reopening.
- Catch unexpected KaTeX render failures and display escaped source text while retaining the formula atom/source, so surrounding text can still be saved.
- Keep `trust: false` and the existing `throwOnError: false` behavior for invalid LaTeX syntax.
- Added regression coverage for missing/null/numeric/object/empty sources and render failure, including safe escaping and reopening.

### Performance measurement and cache decision

Measured full HTML serialization of one paragraph with 5/20/50/100 distinct synthetic formulas in local headless Chrome, with five warm-ups and 30 measured iterations per size:

| Formulas | Median | P95 |
| --- | --- | --- |
| 5 | 0.4ms | 0.5ms |
| 20 | 1.3ms | 1.6ms |
| 50 | 3.1ms | 3.6ms |
| 100 | 5.9ms | 6.7ms |

These include KaTeX DOM generation and HTML serialization, not paint/layout or a full editor transaction. They are local synthetic measurements, not a cross-device guarantee. The editor already schedules persistence only after document changes, with a 300ms trailing debounce; selection-only transactions do not trigger that persistence path. Based on these measurements, no cache is introduced. The README records the measurement boundaries and recommends profiling representative complex documents/slower devices before adding caching.

Validation: **585 editor tests passed across 55 files**; editor build, targeted ESLint, Prettier and `git diff --check` passed. Commit pushed; CI completion has not been awaited.

xuyuanwei678:
## Full code review: font-size undo regression (P2)

The full review covered all 52 changed files. All 816 existing tests passed, but an additional reproduction confirmed one regression in the editor.

For an imported inline container with `font-size: 2em` and `width: 10em`, changing the font size and undoing does not restore the original document:

- With a partial text selection, undo adds the inherited `2em` mark to the selected text while retaining it on the container, doubling the selected text's effective size.
- With a selection spanning the whole container, the container's font-size context is moved to its text. Its `em`-based width can consequently shrink by half.

The earlier clear-formatting fix does not cover the font-size application path, which still uses ProseMirror's range-based `addMark` on marked inline containers.

I will preserve the container's typography context, apply font changes to the selected inline leaves, and add regression coverage for partial/full selections and undo/redo. No other confirmed new findings were identified in this review.

### Fix pushed in 09fb9567

Font-size and font-family commands now replace only the selected inline leaf content with updated marks. They retain the surrounding container's typography context and avoid range-based `addMark`, whose inverse could restore inherited marks onto descendants.

Regression coverage includes both `inline_text_box` and `pptx_tab_column`, font size and font family, partial selections, select-all, and the existing caret-to-select-all behavior. The tests assert unchanged container attributes/marks and exact document equality after undo and redo.

Validation:
- The new regression tests failed before the fix and pass afterward.
- Editor suite: **587 tests passed across 55 files**.
- Editor build, targeted ESLint, and `git diff --check` passed.

Update: the fix is now committed and pushed to this PR in 09fb9567. Fresh full package validation passed: importer 171, renderer 60, editor 587 tests; editor build and targeted lint also passed.

wyuc:
Thanks — the WPS comparison settles finding 2, and you're right. I retract my suggestion: I claimed the symbol tables contain no letters or ASCII punctuation, which is wrong (Wingdings maps `n`, `u`, `v` among others, and Wingdings 3 maps `{|}~uvwx`), so a "table has an entry" condition would have converted ordinary text. I confirmed that by mutating it in: both of your ordinary-ASCII control tests go red. Keeping the current eligibility rules is correct.

Verified all four follow-ups at `566f6cca`, each by reverting the source change while keeping your new tests:
- `5e4909a9` clear formatting keeps container font context — the 3 new cases fail against the previous implementation.
- `0f9a2685` Wingdings 3 `0x7b → ◥` — both new triangle cases fail with the old glyph.
- `845644b3` explicit `data-pptx-text-insets` marker — the importer, renderer and editor-schema tests each fail without their part of the change; the legacy padding fallback keeps saved slides unchanged.
- `566f6cca` guarded formula serialization — all 6 new cases fail against the previous `inlineMath.ts`; the render-failure fallback uses `textContent`, so it stays escaped.

Suites at this head: importer 171, renderer 60, editor 585.

wyuc (dismissed):
Approving at `566f6cca`.

Both review items are closed — clear formatting preserves container font context (caret, select-all, cross-container, exact undo), and the Wingdings 3 `0x7b` glyph is corrected with the conversion eligibility rules kept (see my comment above, including my retraction on finding 2). The non-blocking notes were addressed too: an explicit text-inset marker with a legacy fallback, a guarded formula serialization path with a `''` default, and documentation of server-side tab-width estimates and formula serialization cost. Each new test was confirmed to fail without its source change. Importer 171, renderer 60, editor 585 tests pass.

wyuc:
@xuyuanwei678 This is marked ready again, but the branch is still at `76d2159153`. Your review comment says the font-size / font-family undo fix is implemented locally but not yet committed or pushed, so it isn't in this PR — merging now would ship that regression (undo doubling the selected text's size, or moving the container's font-size context so its `em` width halves).

Could you push the fix here? I'll verify it the same way as the earlier follow-ups (new tests failing against the previous implementation, full importer/renderer/editor suites) and merge once it's green. If you'd rather land it as a separate follow-up PR, just say so and link it, and I'll take this one as is.

xuyuanwei678:
The font-size / font-family undo fix is now pushed to this PR in commit 09fb9567 (on top of 76d2159153). It is no longer local-only.

Font commands now update selected inline leaves while retaining the container's typography context. Regression coverage checks partial selections, select-all, and caret-to-select-all for both inline text boxes and PPTX tab columns, with exact document restoration on undo and redo.

The new regression tests failed against the previous implementation and pass with the fix. Fresh validation after syncing the current PR branch:
- Importer: 171 tests passed.
- Renderer: 60 tests passed.
- Editor: 587 tests passed.
- Editor build, targeted ESLint, and diff whitespace checks passed.

CI will run against the new commit; I have not waited for its result.

wyuc (approved):
Re-approving at `09fb95676b`.

Verified the font-size / font-family undo fix: font commands now replace only the selected inline leaves and leave the container's typography context untouched, the same approach as the clear-formatting fix. Editor 587, importer 171 and renderer 60 tests pass; with the previous `commandExecutor.ts` restored, both new `inline_text_box` / `pptx_tab_column` undo/redo cases fail, so the regression is guarded. CI is green. Thanks for catching this one in your own review.

PR #1521, opened 2026-09-15
fix(audio): resolve CDN-backed narration consistently

### LINKED ISSUE #1515: [Bug]: CDN-backed narration is silently omitted or mis-timed when local bytes are empty
### Bug Description

Narration records that have an empty local Blob, a valid CDN ossKey, and no stored duration are handled inconsistently across playback and export paths.

The shared audio resolver checks the asset pool and the local compatibility Blob, but it does not fall back to the compatibility row ossKey. This produces three related outcomes from the same valid remote-backed record:

- Classroom ZIP export silently omits the MP3 and writes missing: true.
- Playback reports that no audio started and falls back to the text-duration timer.
- Video export considers the asset present because ossKey exists, estimates its duration from text, and then downloads the real MP3 during collection. The resulting timeline can be shorter than the actual narration and cut it off.

This is a compatibility-path bug. Modern server-backed asset-pool records normally resolve through the pool, but AudioFileRecord still supports ossKey-only rows and the video collector explicitly supports that state.

### Steps to Reproduce

1. Use an AudioFileRecord with:
   - a zero-byte Blob
   - a valid ossKey pointing to an MP3
   - duration unset
2. Reference that record from a speech action by audioId.
3. Ensure the asset-pool lookup misses.
4. Export a classroom .maic.zip.
5. Compile and collect a Hyperframes video export for the same classroom.
6. Start playback from a cold browser state.

A real-world reproduction contained 65 referenced narration clips. The browser compatibility table contained 63 ossKey-only rows, all with zero-byte Blobs and no duration. The exported .maic.zip contained only manifest.json and marked all 65 audio entries missing. The Hyperframes artifact later downloaded 63 MP3 files, but all narration timings were estimated.

### Expected Behavior

- All consumers resolve narration bytes through one consistent chain: asset pool, usable local Blob, then compatibility ossKey.
- Classroom ZIP export includes every fetchable narration MP3.
- Video compilation probes the fetched bytes and uses the real duration before building the timeline.
- Playback starts the same narration from a cold or warm state.
- Truly unresolved referenced audio is surfaced as an export warning instead of a silent success.

### Actual Behavior

- resolveAudioBlob returns null for an ossKey-only compatibility row.
- collectAudioFiles skips the entry, and the classroom manifest records missing: true.
- AudioPlayer.play returns false, which starts the estimated reading timer.
- The video compiler emits estimated-duration even though collection later fetches the same MP3 through ossKey.

### Deployment Method

### OpenMAIC Version

0b2e5028112c8d55af3c7c6437708a37e4f174f8

### Affected Area

Import / export

### Browser

Chromium via Playwright 1.58.2

### Operating System

macOS

### Relevant Logs

```shell
Audio references: 65
Compatibility audio rows: 63
Rows with non-empty local Blob: 0
Rows with ossKey: 63
Rows with stored duration: 0
Asset-pool entries: 0

Exported ZIP entries: 1
Missing audio manifest entries: 65
Present audio manifest entries: 0
```

### Screenshots or Recordings

Not applicable. The exported archive and manifest provide the deterministic failure signal.

### Additional Context

- Related to #1007: global asset-pool migration and compatibility resolution.
- Related to #854: Hyperframes video export.
- Existing tests cover an ossKey-only video collection row when duration is already present, but do not cover the failing combination of empty Blob, ossKey present, and duration missing.
- The downstream live deployment has an additional preload-cache implementation. Its cold-versus-warm playback race is intentionally excluded from this upstream report.

cosarah:
## Summary

Fix compatibility audio rows that retain a valid CDN `ossKey` after their local Blob has been evicted. Playback, classroom ZIP export, and video export now resolve the same bytes, and video timelines probe the fetched audio metadata before compilation.

This PR is AI-assisted and remains in Draft after two independent subagent review passes.

## Related Issues

Closes #1515

## Changes

- Resolve narration in asset-pool → non-empty local Blob → compatibility `ossKey` order.
- Reject failed and zero-byte CDN responses without turning them into silent audio.
- Bound classroom ZIP and video-export audio resolution concurrency through the shared helper.
- Probe fetched audio metadata before compiling video narration timing.
- Keep timeline planning and byte collection consistent when remote resolution fails, preventing estimated timing from being paired with a later-fetched clip.
- Warn when a classroom ZIP still contains unresolved referenced audio, while avoiding false warnings only when legacy URLs rescue every owner of an audio ID.
- Add regression coverage for playback, classroom ZIP contents and missing reporting, timeline duration, precedence, failed remote responses, and mixed audio owners.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Seed a classroom with speech records containing an empty local Blob, a valid CDN `ossKey`, and no stored duration.
2. Start playback from a cold Chromium context and verify each clip reaches its natural end.
3. Export a Hyperframes ZIP, render it through the local render service, and inspect both the manifest and MP4 with `ffprobe`.

- Focused regression suite after review fixes: 31 tests passed.
- Full Vitest suite with one worker: 8,151 passed, 39 skipped.
- TypeScript and Prettier checks passed.
- ESLint completed with 0 errors and 18 pre-existing warnings.
- Both spec and standards subagent reviews completed with no remaining findings.
- Local browser E2E: two real CDN-backed clips played to their full 10.464 s and 9.888 s durations on first playback.
- Exported timeline used measured durations (20.352 s total) and scheduled the second narration after the first ended.
- Local render service produced a 20.375 s H.264/AAC MP4 with equal audio and video stream durations.

### Evidence

- [x] Manually tested locally
- [ ] Screenshots / recordings attached (not applicable; no UI change)

## Checklist

- [ ] I have added/updated documentation as needed (not applicable)


wyuc (approved):
Reviewed at `5a748641`. Approving; the notes below are non-blocking and fine as a follow-up.

**Verified**
- Playback, classroom ZIP export, and video export all resolve through `resolveAudioBlob` (pool → non-empty local Blob → compatibility `ossKey`). Walked every branch: pool hit, local bytes, empty local + ossKey 200 / zero bytes / 4xx-5xx / network error, missing record. A failed remote resolution stays out of `audioById`, so the video path never pairs an estimated duration with a later-fetched clip.
- Tests: `tests/audio`, `tests/export`, `tests/video-export`, `tests/media` → 1472 passed; the 6 changed test files → 63 passed. prettier / eslint / `tsc --noEmit` clean.
- Mutation: removing the zero-byte rejection, swapping resolution order, restoring the old `audioById` retention, and dropping the `fullyRescuedAudioIds` guard each turn a test red.
- Outbound surface: `resolveAudioBlob` reads IndexedDB, so it only runs client-side, and no `app/api` route imports it. Cross-origin `ossKey` fetches go through `/api/proxy-media` (per-hop SSRF validation, pinned dispatcher, manual redirects). `ossKey` is only written from server upload results; classroom import never sets it.
- No semantic conflict with #1523: retained refused-narration bytes live in the local `audioFiles` row with no `ossKey`, so they resolve at the local layer before the new network branch.

**Non-blocking notes**
1. (P2) `lib/media/resolve-audio-bytes.ts` accepts any `200` with a non-empty body. A CDN/gateway error page (`text/html`, `application/json`) comes back as a Blob, gets archived as `.mp3` in the ZIP, and is no longer counted as missing. Given this PR's aim of not treating failed responses as narration, consider rejecting clearly non-audio types (allow `audio/*` and `application/octet-stream`). `resolve-stored-bytes.ts` has the same gap.
2. (P3) The legacy `audioUrl` branch in `lib/video-export-app/timeline-deps.ts` (~L278-292) still keeps a zero-byte blob in `audioById`, unlike the other two paths. It surfaces as `skipped-media` rather than silently, but the zero-byte semantics differ.
3. (P3) Nothing pins the concurrency limits passed at the new call sites (`6` / `4`); the helper is tested on its own, and abort doesn't propagate (each unit is individually time-bounded).
4. (P3) `collectLegacyAudioForExport` re-derives owner counts that `enumerateAssetManifest` already provides via `referenceCounts`. It's correct today, but could drift when the DSL adds an audio slot.
5. (P3) The test "keeps failed shared resolution missing and prevents a later CDN retry" asserts that `result.missing` does *not* contain the audio. The user-visible signal is the `skipped-media` diagnostic, so renaming the test and asserting that diagnostic would make the intent clearer.

PR #1523, opened 2026-09-15
refactor(media): one client-side pool commit primitive; keep refused narration instead of re-billing it

### LINKED ISSUE #1467: Unify the client-side pool commit sequence (media pass, narration adoption, TTS) and give TTS refusals retained-bytes semantics
Follow-up from the review of #1392 (raised by @cosarah).

**Problem.** The "store bytes in the pool → clear the store-full marker → write the reference back → mirror locally" sequence now exists in three places with three refusal semantics:

- the media pass (`lib/media/media-orchestrator.ts`): a refusal for room keeps the synthesized bytes locally so a later Retry re-uploads them without paying the provider again;
- narration adoption (`lib/audio/adopt-cached-narration.ts`): a refusal skips the clip and it is attempted again on a later load;
- fresh TTS synthesis (`lib/hooks/use-scene-generator.ts`): a refusal for room discards the freshly synthesized, already-billed audio before the local cache write, so every retry re-calls the TTS provider.

The third one is the concrete defect: it is exactly the double billing the media pass's retained-bytes contract exists to prevent. Fixing it in place would add a fourth variant.

**Proposal.** Extract one client-side pool commit primitive (`commitToPool({ stageId, bytes, mimeType, slot, ... })`) that owns the sequence and the refusal semantics (retain bytes locally under the placeholder/derived key on a room refusal; clear the marker on success at the seam; write back through `mutateDocument`; mirror locally), and make the three callers use it. The TTS path then gets retained-bytes semantics for free.

**Scope note.** This is a refactor across three modules and belongs in its own PR so #1392 stays reviewable.

wyuc:
Closes #1467.

## What

One client-side primitive, `commitToPool` (`lib/media/commit-to-pool.ts`), now owns the sequence the three pool writers used to spell out separately: store the bytes through `putAsset`'s stage seam → write the allocated id back through the existing `mutateDocument` funnel → mirror locally. It returns `stored | refused-retained | failed` instead of throwing, classifies "the store had no room" in exactly one place (`ASSET_QUOTA_EXCEEDED`), and on a refusal retains the bytes under the caller's slot key so a later Retry re-uploads without paying the provider again. It never reads or writes the store-full marker; that stays inside `putAsset`'s seam (cleared on success) and with the passes that spend model money.

- **Media pass** and **narration adoption**: routed through the primitive with observable behaviour unchanged — no existing test was modified. The video poster goes through it too, so no raw `putAsset` is left on a commit path.
- **Fresh TTS** (the defect the issue names): a room refusal used to discard the already-billed clip, so every retry re-called the provider. It now keeps the clip under its derived key and stamps the action, and the next load's narration adoption uploads it with one pool write and zero provider calls — the same contract a refused image already had.
- Round-2 hardening from review: a sibling narration line's failure no longer rolls back a retained refusal; a retain that cannot write locally demotes the outcome to `failed` (no stamp names bytes that are not there); the mirror step is best-effort in one place.

Browser-only mode is untouched: the primitive short-circuits when the pool is not server-backed.

## Verification

- `tsc --noEmit`, `pnpm run check`, `pnpm run lint` (0 errors), `pnpm build` clean.
- vitest over `tests/media tests/audio tests/hooks tests/classroom tests/persistence`: 1160 passed / 15 skipped (baseline 1142 / 15): +18 tests, all new — 11 primitive contract tests and 7 end-to-end narration tests (refusal → retained bytes → next-load adoption with zero provider calls; sibling failure keeps the retained clip; browser-only unchanged). Each new guard was mutation-checked: reverting its hunk fails exactly that test.
- Independently reviewed with probes on the real code paths (abort interleavings, marker semantics, refusal classification vs. what `putAsset` can throw); no P1/P2.


cosarah (approved):
**Approved.** The refactor is behavior-preserving for the media pass and adoption, fixes #1467 for fresh TTS, and the claimed invariants hold against the code. Verified beyond the PR's own checks:

- `putAsset` now has exactly one caller (`commitToPool`), which makes the `mediaFailureCode` narrowing to two named error classes sound — a quota refusal can only reach the task catch wrapped in `MediaStorageRefusalError`, and no API route emits `ASSET_QUOTA_EXCEEDED` as a generation error code.
- The `assetId === requestId` sentinel is collision-free: pool ids are `ast_`-prefixed, request ids are `tts_…` / `tts_request_s…`.
- The retain sink's row is exactly what adoption reads back: same derived key, `rowBelongsToAction` checks satisfied, and `DERIVED_KEY_ACTION_ID` matches both derived-key shapes, so the ActionsBar single-line regen path is covered by the same contract.
- Rollback boundaries are correct: `rollBackFreshNarration` reclaims only minted allocations; the two epoch-change cleanups run only on scenes that are discarded entirely.
- Reproduced: `tsc --noEmit` clean, eslint/prettier clean on the changed files, vitest 1160 passed / 15 skipped — the +18 is exactly the 11 + 7 new tests, no existing test modified.

Two non-blocking nits, fine for a follow-up:

1. **Orphaned comment** — the "Why a fresh clip never replaces the bytes behind an id it is superseding" block (`lib/hooks/use-scene-generator.ts:557`) was `allocatePooledAudio`'s doc. With that function gone it now floats between two functions while explaining `generateAndStoreTTS`'s handling of `existingAudioId` ~60 lines above. Worth relocating into that branch or trimming.
2. **New cache-residue class** — if a refused clip is later re-synthesized directly (not via adoption) and the pool accepts, the retained row under the derived key stays in Dexie as unreferenced residue once the stamp moves to the `ast_` id. Harmless (nothing names it), just a new residue type this change introduces.

PR #1524, opened 2026-09-15
feat(agent-runtime): workbench generate_image / generate_video write through the asset pool (#1007 part 6)

### LINKED ISSUE #1522: [Task] Workbench generate_image / generate_video write through the asset pool (#1007 part 6)
## Summary

Make the workbench tools `generate_image` and `generate_video` store their bytes in the asset pool and write the allocated asset id into the document, exactly as the classic generation chain has done since #1392. Today they are the only production writers that bypass the pool: they write files under `CLASSROOMS_DIR/<stageId>/media` and put `/api/classroom-media/...` URLs into `PPTImageElement.src` / `PPTVideoElement.src|poster`.

This is part 6 of the #1007 delivery plan and it reverses part of #1242 deliberately, for reasons that no longer hold.

## Why now

#1242 took the pool off the workbench tools because the pool of that time was unsafe on a production path: a reference could be written before its bytes were durable, and nothing reclaimed what a failed generation left behind. Both reasons are gone:

- #1392 established the write discipline: bytes first, then the allocated id, and the document reference is the only "done" signal.
- #1472 / #1473 gave the server the entry lifecycle: a document → asset reference table maintained inside every document write, pending → committed allocations with a TTL, a collector that reclaims expired and unreferenced entries, quota over live entries only, and release on course deletion.

With those in place the local-disk path is the outlier. It has no registry entry, no reference row, no quota, no reclamation, and no owner check on `/api/classroom-media` (only path validation and the optional deployment `ACCESS_CODE`). Every reason to keep it is gone, and one product reason to remove it has arrived: a shared material library across the classic and workbench flows needs one stable identity per asset, and only pool ids give that.

## Facts the design rests on

- The tools run server-side inside the agent runner, which is gated on `OPENMAIC_AGENT_RUNTIME_ENABLED` and a non-empty `DATABASE_URL` (`lib/config/feature-flags.ts`). Whenever they execute, the Pg persistence provider and its `PgAssetStore` exist in-process. No file-only fallback is needed.
- The runner writes the document through the owner-bound Pg document store (`patch_stage` → `putSceneBringingCurrent` → `putScene`; the detached video job → `putScene`). Since #1473 those writes maintain `document_asset_refs` unconditionally, and `enumerateAssetManifest` already enumerates `PPTImageElement.src`, `PPTVideoElement.src|mediaRef|poster` and background images. An `ast_` id written there creates a reference row and commits the pending allocation; a `/api/classroom-media` URL creates nothing.
- The workbench pane reads the document through the client document store, so a functional workbench deployment is one whose client is configured server-backed (`NEXT_PUBLIC_PERSISTENCE=1`); that is the same configuration under which `ast_` ids already render via `HttpAssetStore` in classic mode. No new read path.
- Export resolves pool-first in the browser; the render service consumes a prepared ZIP and never resolves ids itself.

## Design

1. One server-side commit helper, `storeGeneratedAsset({ stageId, bytes, mimeType, kind, poster? })`, wrapping `provider.assetStore.put(principal, blob, meta)` with the shared principal the persistence route uses today. It returns the allocated id (and the poster's), or a typed refusal when the store is full. It is the server-side twin of the classic chain's client-side commit (#1467 is unifying that side).
2. `generate_image`: persist through the helper; the element's `src` becomes the `ast_` id; the tool result carries the id where it carried the URL. The `persistGeneratedImage` injection seam stays; its default changes.
3. `generate_video`: same at job completion for the video bytes and, when present, the poster; the completion patch then names both ids in one `putScene`, which is the write that commits them. The `pending-media` registry and the SSE `media_ready` frame are unchanged — the client fold still only updates the in-memory media store.
4. Refusal semantics: a store-full refusal fails the tool call with an explicit, model-visible error and writes nothing — no local-disk fallback, because a fallback would recreate the two-model situation this removes. Retained-bytes-for-retry, which the classic chain has, is a follow-up for the tools; images are small and video generation is already an async job whose retry policy lives in the runner.
5. Legacy documents holding `/api/classroom-media/...` URLs keep rendering: concrete URLs bypass the pool on the read side, and the route stays for them. No converter, consistent with #1242's decision to stop converting.
6. Nothing changes for browser-only deployments: the tools cannot run there.

## Out of scope

- Principal / ownership (#1413, #1129): assets stay under the shared principal as everywhere else. The reference table now records which document holds which asset, which is the data an ownership model needs; the model itself is the next part.
- Retained bytes on refusal for the tools (see 4).
- Removing `/api/classroom-media` or `persistClassroomMediaBytes` (legacy documents still need them).
- The classic chain and #1467.

## Verification plan

Unit tests for both tools with an injected asset store (success writes the id, refusal surfaces the error and writes nothing); a PostgreSQL contract test that a simulated completion followed by the runner's `putScene` leaves a `document_asset_refs` row and a committed entry; the existing workbench fold and media-lifecycle tests unchanged; `pnpm build`.

wyuc:
Closes #1522. Part 6 of the #1007 delivery plan.

## What

The workbench tools were the last production writers that bypassed the asset pool: they wrote files under `CLASSROOMS_DIR/<stageId>/media` and put `/api/classroom-media/...` URLs into the document, with no registry entry, no reference row, no quota and no reclamation. They now follow the discipline the classic chain has had since #1392 and the server lifecycle from #1472/#1473:

- **One server-side helper**, `storeGeneratedAsset` (`lib/server/store-generated-asset.ts`): `PgAssetStore.put` under the shared principal, returning the allocated id or a typed store-full refusal (the same class the persistence route maps to 507). `stageId` travels in the entry's `meta`; nothing server-side stored it before, for either writer.
- **`generate_image`**: the element's `src` is the `ast_` id; the tool result carries the id where it carried the URL. A refusal fails the call with a model-readable error and writes nothing.
- **`generate_video`**: at job completion the bytes (and the poster, when present) are stored and the completion patch names both ids in the one `putScene` it already did — the write that creates the `document_asset_refs` rows and commits the pending allocations. A poster refusal costs only the poster. `pending-media` and the `media_ready` frame are unchanged.
- **Completion patch invariant** (`patchStageVideoPlaceholder`): every slot holding the job's placeholder takes the allocated id; once `src` holds it, `mediaRef` holds the same id or nothing (a stale generated ref or an imported URL there would shadow the new bytes in `sourceRef`); a user's concrete `src` is never touched and still wins. Covered by a 100-element table (src × mediaRef × poster) run through the real resolver chain.
- **Immediate render preserved**: `useResolvedMediaRef` leases a pool id held on the media task while the document ref is not yet leasable, so the pane renders on arrival as before; `resolveMediaRef` no longer presents a pool id as a URL.
- `generate_scene.media` now admits an allocated id (it refused everything that was not a URL, which would have broken the documented `generate_image → generate_scene` flow); `data:` and placeholders are still refused.
- Legacy documents holding `/api/classroom-media/...` URLs keep rendering (concrete URLs bypass the pool on read); the route stays; no converter.

Browser-only deployments are unaffected: the tools cannot run without `DATABASE_URL`.

## Verification

- New PostgreSQL 16 contract suite `tests/persistence/workbench-tool-asset-references.pg.test.ts` (wired into `storage-pg-contract.yml`): the entry is pending after the store, a `document_asset_refs` row exists and `committed_at` is set after the runner's `putScene`; video and poster committed by the real completion patch's single write; the legacy URL creates neither row nor commit. (That a regeneration retires the previous id's reference row is covered by the pre-existing `document-asset-references.pg.test.ts` from #1473, not by this suite.)
- Unit: both tools with an injected store (id write-back; refusal writes nothing and surfaces `ASSET_QUOTA_EXCEEDED`; poster-only refusal), the 100-element completion table, the immediate-render lease, `generate_scene` admission.
- Full vitest 8250 passed / 43 skipped (baseline 8236 / 43 before this branch); four target dirs 2827 passed; `tsc`, prettier, lint (0 errors) and `pnpm build` clean.
- Four rounds of independent adversarial review with probes on real PostgreSQL and the real resolver chain (pending → committed → reclaimed on expiry; quota refusal; out-of-order job completion; reload without an in-memory task; importer-shaped `mediaRef`).


cosarah (approved):
Approve — clean, well-verified change.

Reviewed in depth: the completion-patch invariant checks out against the real resolver chain (`sourceRef = concreteSrc ?? mediaRef ?? src`, so a stale `mediaRef` would shadow the new id — rule 2 is right), the `refused-retained` demotion semantics are sound, the TTS retained-refusal rollback prevents double billing, and legacy `/api/classroom-media` docs still render. Verified locally: target suites 1749 passed, tsc / eslint / prettier clean.

Three minor points — none blocking. Please respond and merge yourself:

1. PR description vs tests: the description credits the new PG suite with "a regeneration retires the previous id's reference row", but that case lives in the pre-existing `document-asset-references.pg.test.ts` (#1473), not `workbench-tool-asset-references.pg.test.ts`. Coverage exists; just a wording fix.

2. Worth confirming intent: rule 2 also retires a *concrete URL* in `mediaRef` when `src` still held this job's placeholder. The importer round-trips concrete URLs into `mediaRef` (`use-import-classroom.ts`), so an imported video in that mixed state is superseded by the generated one. The `P×MU → N¹` table entry covers it and the comment justifies it ("the element was bound to THIS job on purpose") — just want the author's explicit confirmation that discarding the imported pick there is intended.

3. Drive-by, pre-existing and untouched by this PR: `getPendingMediaTask` / `listPendingMediaTasks` in `pending-media.ts` have no production importers left (tests only). Fine to ignore here.

wyuc:
Thanks for the thorough read. On the three points:

1. Right — that case lives in `document-asset-references.pg.test.ts` from #1473, not in the new suite. Description corrected.

2. Intended, and confirmed. The element was bound to this job on purpose: the placeholder in `src` is the agent's instruction that this element's video is the one being generated, so when the bytes arrive they must be what the element resolves to. An imported concrete URL still sitting in `mediaRef` at that moment would otherwise shadow the result silently — the failure mode the earlier rounds found is "stored, referenced, charged, invisible", and a value-kind exemption is exactly what let it back in once. If a user wants the imported video instead, `src` is the slot that wins and is never touched (rule 3). The `P × MU → N` row is deliberate, and the comment says so for the next reader.

3. Agreed it is pre-existing and out of scope here; noting it so it does not get lost — `getPendingMediaTask` / `listPendingMediaTasks` have only test importers now and can go in a cleanup pass.

Bringing the branch up to date with `main` and merging once CI is green.

PR #1525, opened 2026-09-15
feat(token-plan): add TokenDance one-key preset for every modality

wyuc:
## Summary

Adds [TokenDance](https://tokendance.space) as a Token Plan preset, so a single key can be applied to every modality (LLM, image, video, TTS, web search) from **Settings → Token Plan**, and as a built-in OpenAI-compatible LLM provider configurable through `TOKENDANCE_*` env vars.

TokenDance is a model gateway. Chat and image generation are OpenAI-compatible at `/gateway/v1`, and the same key authenticates vendor-protocol routes on the same host that keep each vendor's wire format. The preset therefore reuses existing adapters with the route prefix as the base URL, following the MiniMax and Volcengine Ark presets:

| Modality | Adapter | Base URL | Models |
|---|---|---|---|
| LLM | built-in `tokendance` (OpenAI-compatible) | `/gateway/v1` | DeepSeek V4.1 Flash, DeepSeek V4 Pro, GLM-5.3, Kimi K3, Qwen3.8 Max, Seed 2.1 Pro, MiniMax M3 |
| Image | `seedream` | `/gateway/ark/v3` | `seedream-5.0-lite`, `seedream-5.0-pro` |
| Video | `minimax-video` | `/gateway/minimax` | `minimax-h3`, `minimax-h3-max` |
| TTS | `minimax-tts` | `/gateway/minimax` | `minimax-speech-2.8-turbo` (default), `minimax-speech-2.8-hd` |
| Web search | `bocha` | `/gateway/bocha` | — |

## Adapter changes (vendor-neutral)

- **Seedream:** `resolveArkRoot` now uses a base URL verbatim when it already ends in a version segment (e.g. `https://gateway.example/ark/v3`), in addition to the existing `/api/...` rule. The default bare host still gets `/api/v3` appended.
- **MiniMax video:** H3-family models (`minimax-h3*`) only accept the v2 task API. They now submit to `POST /v2/video_generation` with a content array and poll `GET /v2/query/video_generation/{id}`, which returns the video URL inline. Other models keep the v1 flow unchanged. For H3 models, the connectivity check probes auth on the v2 query route instead of submitting a (billable) generation.

- **Web-search client base URL allowlist:** `resolveSafeClientWebSearchBaseUrl` also accepts the exact base URL a built-in token plan preset writes for that provider. Without this, applying any plan whose web-search route is not an official vendor host would store a URL the route then rejects with 400. The list is derived from `TOKEN_PLAN_PRESETS`, so no vendor URL is hardcoded in the neutral config. Any other client-supplied URL, including a different path on the same host, is still rejected. A test pins this for every preset.

No provider-neutral layer gains vendor knowledge. The only new vendor token in `lib/server/provider-config.ts` is the env-prefix map entry, which is recorded in the neutrality guard's existing per-file debt table (`['tokendance', 2]`).

## Other changes

- `BuiltInProviderId`, `LLM_ENV_MAP` (`TOKENDANCE_*`), `.env.example`, `public/logos/tokendance.svg`, and a provider name in all 12 locales.
- The TokenDance models are added to `MODELS_WITHOUT_CONFIGURABLE_THINKING`. The gateway serves many vendors' models, and per-model thinking parameter pass-through is not verified yet, so no thinking toggle is exposed for them.
- README / README-zh: a TokenDance one-key quick example next to the other provider examples, **TokenDance** in the supported-provider list, and the Gemini-specific "recommended model" note replaced with a provider-agnostic recommendation to enable every modality.

## Testing

- `pnpm check`, `pnpm lint` on changed files, `npx tsc --noEmit`, `pnpm check:i18n-keys`: pass
- `pnpm test`: 720 files / 8267 tests passed (39 skipped)
- New tests:
  - `apply-token-plan`: the TokenDance preset fills every modality with the expected provider, base URL and models.
  - `video-polled-adapters`: the H3 v2 submit/poll contract, plus a terminal failure.
  - `auth-probe-adapters`: Seedream behind a versioned gateway route, and the H3 auth-only probe.
  - `provider-config`: `TOKENDANCE_*` env mapping.
- Route shapes were checked against the gateway's public model catalogue (`/gateway/v1/models`, which lists each model's supported protocols) and with unauthenticated requests. Every configured route answers 401 (route exists, key required), and the v1 MiniMax video route answers 404, which is why H3 needs the v2 contract. No generation was run with a real key, so an end-to-end check with a TokenDance key is still welcome.

https://claude.ai/code/session_01Qrrq9CPwb718mpouz8Y2KL


wyuc:
@sophietao20-star could you run an end-to-end pass on this with a real TokenDance key? Unit tests and CI are green, but no generation has been run against the live gateway yet.

**Path 1: one-click Token Plan (no `.env.local`)**
1. Settings → Token Plan → TokenDance, paste the key, apply.
2. Confirm all five modalities show as applied (LLM, image, video, TTS, web search).
3. In the TTS and web search settings, select MiniMax TTS and Bocha. Applying a plan does not switch the active provider for these two; that gap already exists for every preset.
4. Generate a classroom with web search on, and check:
   - slides are generated with a TokenDance model (e.g. `deepseek-v4.1-flash`)
   - web search returns results (this path returned 400 before the second commit)
   - illustrations are generated (Seedream 5.0 Lite)
   - narration plays (`minimax-speech-2.8-turbo`)
5. Generate a video with `minimax-h3` at 16:9 and 9:16, and confirm the clip plays at the requested orientation.
6. Use "Test connection" on the image and video providers; it should succeed with a valid key and fail with an invalid one.
7. Remove the plan and confirm the TokenDance provider is reset.

**Path 2: server env**
Use the env block from the README quick example (`TOKENDANCE_*`, `IMAGE_SEEDREAM_*`, `VIDEO_MINIMAX_*`, `TTS_MINIMAX_*`, `BOCHA_*`) with `DEFAULT_MODEL=tokendance:deepseek-v4.1-flash`, restart, and repeat step 4 without entering a key in the UI.

Please note anything that fails, with the provider and the error text. Thanks!

sophietao20-star (changes requested):
## Request changes: TokenDance authentication failures block E2E validation
### Test setup
Tested locally at 858a64d507, following the requested Token Plan setup with a real TD API key. The account has sufficient balance, and no server-side provider configuration overrides the client settings.
### Findings
Manual testing found authentication failures across all five capabilities: text, image, video, TTS, and web search. Errors report unauthorized / API key does not exist; the image connection test explicitly returns HTTP 401.
Course generation also fails with LLM returned empty response, although its underlying error has not yet been confirmed.

Image connection test — HTTP 401

Image

TTS — authentication rejected

Image

Web search — authentication failure blocks course generation

Image

Course generation — generic empty-response error

Image

Code review confirms that the preset configures TD gateway endpoints and the adapters send Bearer credentials. A shared authentication issue is the leading hypothesis, but whether it lies in credential propagation or gateway validation remains unconfirmed.
Please resolve authentication with a valid TD key across all five capabilities. Also preserve meaningful upstream errors in the outline-generation flow rather than potentially reducing them to “empty response.”
E2E acceptance remains blocked. We can rerun the requested validation once authentication is working.

wyuc:
Thanks for the detailed pass. I dug into the errors and I think the blocker is the key itself rather than the wiring, but I need one check from you to be sure.

**What the gateway's two messages mean**

The gateway distinguishes them, and I reproduced both against the live endpoints with no valid key:

- No `Authorization` header at all → `{"code":"unauthorized","message":"缺少 API 密钥"}` ("missing API key")
- A syntactically fine but unknown key → `{"code":"unauthorized","message":"API 密钥不存在"}` ("API key does not exist"), HTTP 401

Your three screenshots (Seedream connection test, MiniMax TTS, Bocha) all show the **second** one. That means the key was transmitted on each of those three code paths and the gateway refused the value, so credential propagation is working and the rejection is happening at the gateway.

I also ruled out attribution as a cause: per the [app attribution docs](https://tokendance.space/docs/app-attribution.md), `X-App-URL` is optional and only used to attribute a call to an app. An invalid key returns the identical error with and without the header, so its absence cannot cause a 401.

**Could you run this with the same key?**

```bash
curl -i https://tokendance.space/gateway/v1/chat/completions \
  -H "Authorization: Bearer YOUR_KEY" -H "content-type: application/json" \
  -d '{"model":"deepseek-v4.1-flash","messages":[{"role":"user","content":"hi"}]}'
```

- If this also returns 401 `API 密钥不存在`, the key is not valid at the gateway and nothing in this PR can fix it. Worth confirming on the [keys page](https://tokendance.space/keys) that the key is enabled, not expired, not deleted, and belongs to the account with balance — and that it was created there rather than copied from somewhere else.
- If this **succeeds** while the app still fails, that is a real bug in this PR and I will fix it. Please say so and I will dig in.

One thing worth ruling out first: the provider panels store the API key exactly as typed, without trimming, so a trailing space or newline picked up when copying would produce exactly this error. Re-pasting carefully (or via the Token Plan page, which trims) is a quick check.

**On the "empty response" error**

Agreed that a real upstream error should never be flattened, so I checked this path specifically. The outline route already surfaces thrown errors verbatim, and `LLM returned empty response` is used only when the stream ends with no text *and* no error. A 401 is not that case: the gateway returns a 401 JSON body for streaming and non-streaming requests alike, plus a `TokenDance-Recovery-Action: reauthorize_api_key` header, which the AI SDK raises and the route reports as-is.

Also, the third and fourth images in your review are the same file (identical checksum), so the course-generation screenshot did not come through and I could not see that error. If it reproduces once auth works, please re-share it (the server log line starting `Outlines attempt ... diagnostics:` records the response length and a preview) and I will make sure the real cause is surfaced.

I have merged the latest `main` into the branch. Happy to rerun the full validation once the key question is settled.

sophietao20-star:
Thanks for the clarification, and apologies for the confusion—the previous key was indeed invalid. I’ll rerun the end-to-end tests with a valid key.

sophietao20-star:
## E2E validation: Passed

Both requested paths passed on `0507aacf5f` with real provider calls. The previous authentication failures and LLM empty-response error did not reproduce with the new key.

### Path 1: Token Plan UI — Passed

- All five capabilities configured successfully.
- Course generation, Bocha search, Seedream images, and MiniMax narration worked.
- H3 landscape and portrait videos generated and played correctly.
- Image/video connection tests accepted the valid key and rejected an invalid key.
- Disabling the plan cleared the TD key and reset the provider configuration.

Course with generated image and narration

H3 landscape playback

H3 portrait playback

### Path 2: Server environment configuration — Passed

Using the README configuration and a fresh browser context with no UI key, course generation, search, images, and narration playback all worked. A supplemental search confirmed ten returned results.

Course generated using server-side configuration

**No blocking issues found in this validation.**

sophietao20-star (approved):
Both requested E2E paths passed with the new key. No blocking issues found in this validation. Results and screenshots: https://github.com/THU-MAIC/OpenMAIC/pull/1525#issuecomment-5692490179

PR #1531, opened 2026-09-16
fix(editor): preserve concurrent scene updates

### LINKED ISSUE #1529: [Bug]: Slide edit session can overwrite concurrent agent updates
## Bug Description

An open slide edit session keeps its own `history.present` snapshot. If an agent updates the same scene in the canonical stage store, the next toolbar operation, renderer snapshot commit, undo, or redo still uses the stale session snapshot and writes it back. This silently removes the agent update.

## Steps to Reproduce

1. Open a slide in Pro edit mode so `useSlideEditSession` is seeded.
2. Update the same scene through the agent/stage-store path, for example by adding an element.
3. Without leaving edit mode, perform a toolbar edit or commit a canvas drag.
4. Inspect the scene content in `useStageStore`.

A deterministic unit reproduction seeds the session, replaces the mocked canonical scene with a version containing an `agent-text` element, then applies a local geometry update.

## Expected Behavior

The local operation is rebased onto the latest canonical scene content. Concurrent agent changes remain present, and undo cannot restore a snapshot older than the agent update.

## Actual Behavior

The session applies the operation to stale `history.present` and writes the resulting whole snapshot through, removing the concurrent agent change. Undo/redo can do the same with older snapshots.

## Deployment Method

Local development (pnpm dev)

## OpenMAIC Version

`b5605ef9` (current `main`)

## Affected Area

Editor / canvas

## Browser

Chrome

## Operating System

macOS

## Relevant Logs

No runtime error is emitted; this is a silent state overwrite.

## Additional Context

The stale read is in `components/edit/surfaces/slide/slide-edit-session.ts`. The renderer compatibility path also needs to derive the local delta from the prior snapshot and replay it onto the latest canonical content.

Duang777:
## Summary

Prevent open slide and quiz edit sessions from overwriting newer same-scene updates in the canonical stage store.

Both sessions now adopt externally updated content as a fresh history baseline before each mutation. Slide renderer snapshots replay their local delta onto the latest canonical slide. Quiz controls pass mutations into the session so they execute against the latest canonical quiz. Undo and redo history created before an external update is invalidated.

AI-assisted: implementation and regression tests were prepared with AI assistance, then reviewed and verified by the author.

## Related Issues

Fixes #1529

## Changes

- refresh slide and quiz edit history from the canonical scene before mutations
- rebase slide renderer snapshot deltas onto concurrent canonical updates
- apply quiz mutations to the latest canonical content
- prevent undo and redo from restoring snapshots older than an external update
- add deterministic regressions for slide toolbar edits, renderer commits, quiz text edits, quiz reorders, and undo

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Seed a slide or quiz edit session from the canonical scene.
2. Replace the canonical scene with a concurrent Agent version containing additional content.
3. Apply a local edit, undo, or redo.
4. Verify the Agent content remains and stale history is not replayed.

- new Quiz concurrency tests fail 3/3 before the fix and pass 3/3 after it
- Quiz surface suites pass 39/39
- all `tests/edit` suites pass 374/374
- `pnpm check`, `pnpm lint` (0 errors), and `pnpm exec tsc --noEmit`
- `git diff --check`

### Evidence

- [x] Manually tested locally
- [ ] Screenshots / recordings attached (not applicable to state-only behavior)

## Checklist

- [x] I have added/updated documentation as needed (not applicable; behavior is covered by regression tests)


Duang777:
Implementation and verification details for commit `02f4b8a6`:

- `freshHistory()` compares the open session with the current canonical slide before every mutation. A concurrent external update becomes a new baseline with empty `past`/`future`, preventing stale undo or redo entries from overwriting it.
- Toolbar and transaction operations execute against that refreshed baseline.
- The legacy renderer still emits a complete snapshot. When the canonical scene changed concurrently, `rebaseSlideEditSnapshot()` derives only the local delta from `previous -> next` and applies those operations to the latest canonical slide instead of replacing the whole document.
- Existing tolerant legacy reducers keep late edits to externally removed elements as no-ops.

Regression evidence:

- Red: the three new concurrency cases failed against the original implementation because `agent-text` disappeared or undo restored the older snapshot.
- Green: focused session/bridge suites pass 28/28.
- Expanded editor regression passes 371/371.
- Prettier, ESLint (0 errors), TypeScript, i18n key alignment, Node engine, package version, and internal dependency checks pass.

Self-review covered canonical/store consistency, stale scene guards, undo/redo invalidation, renderer snapshot replay, and missing-element behavior. No P0-P2 issue was found in this diff.

One related risk remains intentionally outside this focused PR: the quiz edit session has a similar long-lived seed model and should be assessed separately rather than mixed into this slide fix.

Duang777:
Rebased onto current `main` at `2cbd011c` (new head `76976ba5`). The upstream TokenDance preset change had no overlapping files, so the rebase completed without manual hunk resolution. Revalidated: focused editor tests 28/28, TypeScript, Prettier, ESLint, and `git diff --check` all pass. CI has restarted for the rebased commit.

Duang777:
Added Quiz parity in `ef1e587d` after the audit follow-up confirmed the same stale-session overwrite path remained outside the original slide-only fix.

The Quiz session now refreshes from canonical content before every mutation and applies a mutation function to that current version. Concurrent Agent-added questions and unrelated field changes survive local text edits and stale reorders. External updates also invalidate stale undo/redo history.

Regression evidence:
- Red: 3/3 new concurrency cases failed before the fix; Agent-added questions were removed by text edits, reorder, and undo.
- Green: Quiz session tests 9/9, Quiz surface suites 39/39, and all editor tests 374/374.
- TypeScript and Prettier pass. ESLint reports 0 errors; its 18 warnings are pre-existing outside this diff.
- `git diff --check` passes.

This closes the Quiz sibling risk called out in the earlier PR comment.

LING-6150 (changes requested):
Thanks , I reproduced the original stale-session overwrite, but found two remaining blockers on ef1e587d:
1. Slide snapshots that already include the Agent update throw during rebase. The canvas reads canonical content, so after an Agent adds agent-text, a subsequent drag can submit a snapshot already containing it. Diffing that snapshot against the stale session history produces another element.add, then replaying it onto canonical content throws element.add: id "agent-text" already exists. Please account for snapshots that already include external changes and add a regression alongside the existing stale-snapshot case.
2. Quiz option callbacks can target the wrong option. The form still renders history.present, but mutations now run against refreshed canonical content. If the Agent reorders [Apple, Banana] to [Banana, Apple], editing the still-visible Apple input at index 0 changes Banana instead. I reproduced [Apple edited, Apple]. Correct-answer toggles and other index-based operations share this mismatch. Please synchronize the rendered baseline and guard stale positional callbacks so they cannot modify a different option.
The existing 374 editor tests pass, but both additional regressions fail. Holding approval until these paths are corrected.

Duang777:
Addressed both review blockers in `ce9a8ba1`.

- Slide snapshot rebase now treats an `element.add` whose ID already exists in the latest canonical slide as an external change that has already landed. It skips that duplicate add, preserves the canonical element and its newer properties, and still replays independent local renderer edits.
- Quiz rendering now switches to externally updated canonical content immediately and adopts it as the new session baseline. Option callbacks carry the rendered option reference and become no-ops if an external reorder means the recorded index now points to a different option. This guard covers label edits, correct-answer toggles, deletion, and reordering.

Regression evidence:

- Red: the duplicate Slide add originally threw `element.add: id "agent-text" already exists`; the strengthened case also caught stale renderer properties overwriting the canonical element. Five Quiz cases failed before the fix: stale rendering plus the four index-based operations.
- Green: `pnpm exec vitest run tests/edit` passes 51 files and 380 tests.
- `pnpm exec tsc --noEmit`, `pnpm check`, and `git diff --check` pass.
- `pnpm lint` passes with 0 errors; its 18 warnings are pre-existing outside this diff.
- The branch contains current `origin/main` at `2cbd011c`.

AI-assisted disclosure: this follow-up was implemented and tested with AI assistance. I reviewed the diff and test results before pushing.

@LING-6150, please re-review this revision.

LING-6150:
@Duang777 Thanks for the update. On ce9a8ba1, the Quiz regressions pass and the duplicate-add exception is resolved. However, the Slide fix now silently drops a valid local edit.
Reproduction:
1. Open a slide edit session.
2. An Agent adds agent-text at top: 111, and the canvas displays it.
3. Drag that newly added element to top: 222.
4. Commit the renderer snapshot.
The saved position remains 111. I independently reproduced this at the edit-session boundary: the same case passes on base 2cbd011c and fails on the current head.
rebaseSlideEditSnapshot() still derives the delta against stale session history, so the edited element appears as element.add. The new existing-ID check skips that operation entirely, including the user’s geometry change.
Please account for the baseline the renderer actually edited, so already-observed Agent changes can be distinguished from subsequent local edits. Skipping the entire duplicate element loses user changes; replacing it wholesale could overwrite newer canonical properties.
Please add a regression covering a local drag of the Agent-added element itself, while retaining coverage for stale renderer snapshots, preservation of newer canonical properties, and the duplicate-add case.
Holding approval until this remaining Slide case is fixed.

Duang777:
Addressed the remaining Slide blocker in `d0815d03` and synced the branch with current `main` at `f6e350a1` (head `37d56dfd`).

The controlled `SceneProvider` update callback now carries the exact snapshot exposed by that render. `useSlideCanvasController` applies the renderer mutation to that captured baseline, then `commitContent` derives `renderer baseline -> renderer result` and replays only that delta onto the latest canonical slide. This distinguishes an Agent-added element already observed by the renderer from a duplicate local add, while preserving canonical properties changed after that render.

Regression coverage now verifies:
- Agent adds `agent-text` at `top: 111`, the renderer observes it, and a local drag commits `top: 222`.
- A delayed callback from a stale renderer keeps its original render baseline.
- Newer canonical properties survive replay of an independent local renderer edit.
- Stale snapshots and duplicate-add paths remain non-throwing and preserve canonical content.

Red/green evidence: the new Agent-add-then-drag regression failed before the fix with expected `222`, received `111`; it passes after the baseline propagation change. Post-merge verification passes 53 files / 384 tests, including the renderer differential suite. `tsc --noEmit`, changed-file ESLint, Prettier, and `git diff --check` also pass.

AI-assisted disclosure: this follow-up and its tests were implemented with AI assistance. I reviewed the diff and verification results before pushing.

@LING-6150, please re-review this revision.

Duang777:
`main` advanced again while the previous CI run was finishing, so I synced `f1b34e3a` as required (new head `8fbbfe53`). The upstream audio/workflow changes do not overlap this PR.

Post-merge local verification passes 54 files / 389 tests (the 384 editor/renderer regressions plus 5 new upstream audio tests), `tsc --noEmit`, and `git diff --check`. The preceding head also completed all five GitHub checks successfully; CI has restarted for this final sync.

AI-assisted disclosure: this sync and verification were completed with AI assistance and reviewed before pushing.

LING-6150 (approved):
Thanks , I re-reviewed the current head 8fbbfe53. The remaining Slide blocker is resolved: renderer callbacks now carry the exact render baseline, so an Agent-added element can be locally dragged without losing that edit, while newer independent canonical properties are preserved.
I also verified the stale-callback coverage and confirmed that the Quiz positional guards remain intact. I reran the four directly affected regression suites locally: 4 files, 33 tests passed. All current CI checks are green.
main has advanced only through unrelated audio changes, and this PR remains mergeable. No further blocking findings from me. Approving. cc @wyuc

wyuc:
Thanks @Duang777 for the careful work here, and @LING-6150 for two thorough review rounds — the stale-session overwrite is real and the regression tests pin it down precisely.

That said, I'm going to close this rather than merge it. The current editor behaviour is intentionally last-write-wins: it keeps edit sessions simple and predictable until we build a proper collaborative-editing mechanism (an operation log or CRDT with real merge semantics). This PR effectively adds a partial rebase layer — slide delta replay, positional guards for quiz options, and history invalidation — which is a lot of concurrency logic to maintain and would mostly be replaced once that mechanism lands. It also trades one simple behaviour for a few subtler ones (undo history disappearing after an external update, occasionally ignored quiz clicks), and so far we haven't seen this hit in real usage.

If this does come up for users before the collaborative model exists, we'd prefer a much smaller detect-and-refresh step: when a session is about to commit and the canonical scene has changed since the session was seeded, don't merge — reload the editor from the latest content and tell the user their last action wasn't applied. That keeps last-write-wins semantics explicit and is easy to replace later. Happy to review something along those lines, and #1529 can stay open to track it.

PR #1532, opened 2026-09-16
fix(playback): queue widget messages until iframe ready

### LINKED ISSUE #1530: [Bug]: First widget action can be lost before iframe load
## Bug Description

Interactive playback can dispatch its first widget action before the pooled iframe is registered or before the new iframe document has installed its `message` listener. The host currently resolves the sender lazily, but an absent callback returns `null` and a registered-but-not-loaded iframe receives a fire-and-forget `postMessage`; both paths silently lose the action.

## Steps to Reproduce

1. Open an interactive scene whose first timeline item is a widget action such as `widget_setState`.
2. Start playback while the pooled iframe is mounting or reload its `srcDoc`.
3. Delay the iframe load event.
4. Observe the first action is dispatched before the child document is ready.

A deterministic store reproduction requests a scene sender before iframe registration, sends `SET_WIDGET_STATE`, then registers the callback. No later delivery occurs.

## Expected Behavior

Widget messages sent before registration/load are queued per scene and flushed in order after the current iframe document fires `load`. A document reload should return the scene to the not-ready state.

## Actual Behavior

Messages are dropped when no callback exists, or posted too early to a document that has not installed its listener. There is no ready state or pending queue.

## Deployment Method

Local development (pnpm dev)

## OpenMAIC Version

`b5605ef9` (current `main`)

## Affected Area

Playback / presentation

## Browser

Chrome

## Operating System

macOS

## Relevant Logs

No runtime error is emitted; the widget action silently has no effect.

## Additional Context

The race spans `components/edit/PlaybackChromeRoot.tsx`, `lib/store/widget-iframe.ts`, and `components/scene-renderers/InteractiveIframeHost.tsx`. This is distinct from #871, which addressed generated widgets missing the listener contract.

Duang777:
## Summary

Prevent the first interactive widget action from being lost while its pooled iframe is registering or loading.

Widget messages are now queued per scene until the current iframe document fires `load`, then flushed in FIFO order. Replacing the iframe document resets readiness, and unregistration removes stale queued messages.

AI-assisted: implementation and regression tests were prepared with AI assistance, then reviewed and verified by the author.

## Related Issues

Fixes #1530

## Changes

- add per-scene iframe readiness and pending message queues
- return a queueing sender even before the iframe callback is registered
- reset readiness when iframe content changes and flush on `load`
- clear callback, readiness, and queued messages when a scene iframe is removed
- add store contract tests and a host-level load-event regression test

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Resolve a widget sender before its iframe callback is registered.
2. Dispatch one or more widget actions while the iframe is not ready.
3. Register the iframe and dispatch its `load` event.
4. Verify queued actions are delivered once, in order, and later actions send immediately.

- widget store, iframe host, widget action, and playback regression suites (8 files, 65/65 tests)
- `pnpm check`, `pnpm lint` (0 errors), and `pnpm exec tsc --noEmit`
- i18n key, Node engine, package version, and internal dependency range checks
- registration-before-load, FIFO delivery, reload reset, and unregistration cleanup
- focused diff self-review found no P0-P2 issues

### Evidence

- [x] Manually tested locally

## Checklist

- [x] I have added/updated documentation as needed (not applicable; behavior is covered by regression tests)


Duang777:
Implementation and verification details for commit `96b9eea3`:

- `getSendMessage(sceneId)` now returns a stable scene-bound sender before callback registration, so early widget actions enter a per-scene queue instead of disappearing.
- `registerIframe()` marks each newly registered iframe document as not ready.
- `InteractiveIframeHost` registers the sender in a layout effect and marks the document ready from the iframe `load` event.
- `markIframeReady()` removes and flushes the pending batch in FIFO order; subsequent messages send immediately.
- Unregistering a scene clears its callback, readiness flag, and pending queue so stale messages cannot leak into a later document.

Regression evidence:

- Red: all four store contracts and the host load-gating case failed against the original implementation.
- Green: the expanded widget/iframe/playback regression set passes 65/65 across 8 files.
- Prettier, ESLint (0 errors), TypeScript, i18n key alignment, Node engine, package version, and internal dependency checks pass.

Self-review covered registration-before-send, send-before-registration, FIFO order, document reload, unmount cleanup, active-scene fallback, and existing widget action call sites. No P0-P2 issue was found in this diff.

Duang777:
Rebased onto current `main` at `2cbd011c` (new head `1fe1660b`). The upstream TokenDance preset change had no overlapping files, so the rebase completed without manual hunk resolution. Revalidated: focused iframe/widget tests 16/16, TypeScript, Prettier, ESLint, and `git diff --check` all pass. CI has restarted for the rebased commit.

LING-6150 (changes requested):
Thanks , I rechecked the first-action race and found one remaining blocking case in development StrictMode, which is directly relevant to the issue’s reported pnpm dev environment.
PooledIframe unregisters during its layout-effect cleanup, and registerIframe(sceneId, null) deletes the pending queue. During StrictMode’s setup → cleanup → setup replay, this drops widget actions that were queued before the iframe’s first registration.
I reproduced this with a host-level React DOM/jsdom test by intercepting the iframe’s postMessage: the ordinary-mount control delivers the prequeued SET_WIDGET_STATE, while the StrictMode variant loses it. This does not demonstrate a production-build failure, but it means the reported development path is still not fixed.
Could you preserve valid pending actions across same-document StrictMode replay, while still preventing messages from a genuinely removed or replaced iframe lifecycle from leaking into the next one? Please add a StrictMode host regression test covering this case.

Duang777:
Addressed the StrictMode blocker in `7f40a74b`.

`registerIframe()` now returns a cleanup lease. Cleanup waits until the current microtask and clears state only if the same callback still owns the scene registration. A StrictMode setup, cleanup, setup probe therefore keeps the pre-registration queue when the same iframe document immediately reacquires the registration.

Each registration also carries a stable document token derived from `srcDoc` or `src`. A different token clears messages queued for the replaced document, while a matching token preserves valid pending actions for the same document.

Regression evidence:

- Red: the host-level React DOM/jsdom StrictMode case lost the prequeued `SET_WIDGET_STATE` before this change.
- Green: the widget store, iframe host, and widget action suites pass 3 files and 19 tests.
- `pnpm exec tsc --noEmit`, `pnpm check`, and `git diff --check` pass.
- `pnpm lint` passes with 0 errors; its 18 warnings are pre-existing outside this diff.
- The branch contains current `origin/main` at `2cbd011c`.

AI-assisted disclosure: this follow-up was implemented and tested with AI assistance. I reviewed the diff and test results before pushing.

@LING-6150, please re-review this revision.

LING-6150 (approved):
Thanks for the follow-up. Re-reviewed 7f40a74b: the StrictMode queue-loss case is fixed, and the armed-iframe unmount regression now leaves no pending messages after cleanup. The related regression suites pass.
I also verified real classroom playback locally using an existing interactive course: the widget’s object density updates automatically in sequence, from 600 → 1000 → 1500 kg/m³.
The attached GIF shows the classroom replay at 2× speed, including the dialogue and playback controls.

Image

No remaining blockers found in the reviewed scope， approving from my side. cc  @wyuc

Duang777:
Synced the approved branch with current `main` at `f6e350a1` (head `d1c7f5e8`). The upstream storage change has no overlapping files; no iframe behavior changed during the merge.

Post-merge verification passes the same 19 focused regressions: 6 widget iframe store tests, 9 iframe host/StrictMode tests, and 4 widget action tests. `tsc --noEmit`, changed-file ESLint, Prettier, and `git diff --check` also pass.

AI-assisted disclosure: the merge verification and PR update were completed with AI assistance. I reviewed the branch state and test results before pushing.

Duang777:
`main` advanced again while the previous CI run was finishing, so I synced `f1b34e3a` as required (new head `daebdb62`). The upstream audio/workflow changes do not overlap this PR.

Post-merge local verification passes 24 focused tests (the 19 iframe/widget regressions plus 5 new upstream audio tests), `tsc --noEmit`, and `git diff --check`. The preceding head also completed all five GitHub checks successfully; CI has restarted for this final sync.

AI-assisted disclosure: this sync and verification were completed with AI assistance and reviewed before pushing.

wyuc (approved):
Approving. Thanks @Duang777 for the fix and the StrictMode follow-up, and @LING-6150 for the review and the real-classroom verification.

Checked on my side: every caller of `getSendMessage` already treats the sender as optional, so returning the queueing sender instead of `null` doesn't bypass any fallback; messages are only addressed to the mounted scene, so pending queues don't accumulate for scenes that never load; commits after the approval at `7f40a74b` are merges from `main` only.

PR #1553, opened 2026-09-16
fix(web-search): stop leaking Brave's HTML challenge page into errors

### LINKED ISSUE #1552: [Bug]: Brave web search forwards the raw HTML challenge page as the error message
### Summary

Both Brave web-search error paths in `lib/web-search/brave.ts` forward the upstream response body verbatim:

```ts
if (!res.ok) {
  const errorText = await res.text().catch(() => '');
  throw new Error(`Brave Search error (${res.status}): ${errorText || res.statusText}`);
}
```

When Brave throttles or blocks a request it does not answer with a short JSON error — it answers with a full HTML challenge/error page. That page becomes the `Error.message`, propagates up through `searchWithBrave`, and ends up in logs and in the classroom generation UI as a wall of markup. A rate-limit, which is transient and actionable, is also indistinguishable from a real outage.

This is the keyless scrape path most users hit, since Brave is `requiresApiKey: false` (see #641).

### Steps to Reproduce

1. Configure Brave as the web search provider without an API key (keyless scrape mode).
2. Issue enough searches to trip Brave's public-endpoint rate limit, or mock `proxyFetch` to resolve with an HTML body and status 429.
3. Observe the thrown error.

### Current Behaviour

`Error.message` contains the entire HTML document, e.g. `Brave Search error (429): <!doctype html><html>…`. The message is unbounded in length and carries no usable signal.

### Expected Behaviour

- 429 produces a dedicated, actionable message that names the throttling and the available remedies.
- Non-429 failures fall back to the HTTP status text rather than echoing a markup body.
- Short plain-text upstream details are still preserved, since those are genuinely useful.

### Related

- #643 — same class of defect for TTS providers, where a 429 was collapsed into a generic 500.
- #641, #687 — prior fixes to the same keyless Brave path.

### Environment

- Affected file: `lib/web-search/brave.ts` (both `searchWithBraveApi` and `searchWithBraveScrape`)
- Reproduced against `main` @ 2cbd011c

I have a fix ready and will open a PR against this issue.

HuntercodeT:
## Summary

Both Brave web-search error paths forwarded the upstream response body verbatim into `Error.message`. Brave answers a throttled or blocked request with a full HTML challenge page, so that page ended up in logs and in the classroom generation UI, and a transient 429 was indistinguishable from a real outage.

Both paths now go through a shared `formatBraveError` helper, mirroring the existing `formatBochaError` / `formatDoubaoError` pattern in this directory.

## Related Issues

Closes #1552

## Changes

- Add `formatBraveError` in `lib/web-search/brave.ts`, used by both `searchWithBraveApi` and `searchWithBraveScrape`.
- Give HTTP 429 a dedicated message naming the throttling and the three remedies: retry later, switch provider, or configure a Brave API key.
- Drop error bodies that are markup or longer than 200 characters, falling back to the HTTP status text.
- Preserve short plain-text upstream details, which remain genuinely useful (for example an invalid subscription token on the API path).
- Add three regression tests covering the 429 path, an HTML body on a non-429 failure, and a short plain-text API error.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

## Verification

**Goal verified.** `npx vitest run tests/web-search/brave.test.ts` passes, 6 tests including the 3 new ones. Each new test asserts the specific property that was broken: the 429 message contains `429` and `rate-limited` and no `<`, the 503 message is exactly the status text, and the 401 message keeps the upstream detail.

**Regression checked.** The only behaviour change is the text of thrown errors on non-OK responses. Success paths, `parseBraveSearchHtml`, and provider selection are untouched. The existing tests in this file still pass unchanged.

**CI checks run locally.**

| Check | Result |
| --- | --- |
| `pnpm format` | clean |
| `pnpm lint --fix` | no errors on the changed files |
| `npx tsc --noEmit` | 33 errors before and after this change, identical output, all pre-existing on `main` and in unrelated modules |

No user-facing UI strings were added, so no i18n changes are needed. These are thrown `Error` messages in the server-side search layer, consistent with every other provider in `lib/web-search/`.

## Notes

`pnpm lint --fix` also wanted to strip two `eslint-disable-next-line no-console` directives in `packages/@openmaic/renderer/src/snapshot/index.ts` and leave trailing whitespace behind. That is unrelated to this fix, so I reverted it to keep this PR focused. It may be worth a separate look.


HuntercodeT:
Follow-up commit `ce54981b`: deciding whether an error body was worth surfacing still meant reading the whole challenge page into memory first. `readBraveErrorDetail` now checks `content-type` up front and cancels the stream when it is markup, so the large body is never read at all. The string-level markup guard in `formatBraveError` stays as a backstop for responses that lie about their content type.

Added a seventh test that asserts the stream's `cancel()` actually fires. Full suite still green, `tsc --noEmit` output unchanged at the same 33 pre-existing errors.

LING-6150:
@HuntercodeT The raw-body propagation is reproducible with mocked responses, and the sanitization addresses that mechanism. Before proceeding, could you provide a sanitized capture from an actual failing Brave request, including the request mode (keyless or API-key), HTTP status, Content-Type, and a short response-body excerpt?
Please also clarify which UI flow displays the error. The classroom-generation path catches search failures, logs them, and continues without search context, so the claim that the HTML reaches the classroom generation UI needs a concrete reproduction or trace. A screenshot would help if the error is visible there.
Separately, the API-key path already sends a subscription token, so its 429 message should not suggest configuring a Brave API key. Please distinguish the guidance for the two modes and add coverage for an API-key 429.

wyuc:
Thanks @HuntercodeT, and @LING-6150 for the careful pass. Maintainer view on what's needed to land this:

- **A live capture is optional.** Not forwarding an HTML challenge page into `Error.message` and logs is worth doing on its own, and the mocked responses reproduce the mechanism well. Skipping the body read via `content-type` is a nice touch.
- **Required: mode-specific 429 guidance.** The API-key path already sends a subscription token, so its 429 message shouldn't suggest configuring a Brave API key — for that mode, point to the plan's rate limit / retrying later / switching provider instead. Please add a test for an API-key 429 alongside the keyless one.
- **Please soften the description's UI claim.** Classroom generation catches search failures and continues without search context, so describe the impact as logs and error messages rather than the generation UI, unless you have a trace showing otherwise.

With those two changes and green CI, this is good to merge.

wyuc:
@HuntercodeT, gentle ping on this one. Are you still planning to update it along the lines of my comment above? The main thing needed is just not forwarding Brave's HTML challenge/error page into `Error.message` and logs (a bounded, text-only summary with the status is plenty); a live capture is optional.

If we don't hear back in about two weeks, we'll finish it ourselves on top of your branch, keeping you credited as co-author. Thanks!

HuntercodeT:
Thanks @wyuc and @LING-6150, updated in `d2036758`:

- **Mode-specific 429 guidance.** Keyless (scrape) mode still suggests configuring a Brave Search API key. API-key mode now points at the plan's rate limit (retry later, switch provider, or raise the plan limit) and no longer suggests configuring a key. Added an API-key 429 test next to the keyless one.
- **Description softened.** The impact is now described as thrown error messages and server logs rather than the generation UI; the doc comment on `formatBraveError` says the same.

The HTML challenge page is still never forwarded: the body is cancelled when `content-type` is HTML, and anything that is markup or over 200 chars falls back to the status text. `brave.test.ts` passes 8/8; prettier, eslint and `tsc --noEmit` are clean.

LING-6150 (approved):
Thanks for the update!    @wyuc, I’ve checked d2036758, and the updates look consistent with your feedback:
- API-key 429 guidance now points to the plan’s rate limit, with a dedicated regression test.
- The description now limits the impact to error messages and server logs, removing the unverified generation-UI claim.
I independently reran the Brave tests: 8/8 passed, and all six CI checks are green. No remaining blockers from my review , approving.

wyuc (approved):
Re-reviewed d2036758: both round-1 items are fixed (API-key 429 no longer suggests configuring a key; API-key 429 test added), and the challenge-page body cannot reach thrown errors or logs on any branch. Auth failures (401/403) are still reported as-is, not as rate limits. Thanks @HuntercodeT!

Non-blocking follow-ups (P3), happy to take them in a later PR:
- The string-level `isMarkup` backstop in `formatBraveError` has no test: every current HTML test sets `content-type: text/html`, so it short-circuits before reaching it. A `text/plain` + `<!doctype html>` case would pin it.
- `isMarkup` only catches bodies that start with a tag; a short non-HTML-typed body like `Access denied <br> ref 42` is still echoed (capped at 200 chars). Matching `<[a-z!/]` anywhere would close it.
- `/\bhtml\b/` doesn't match `application/xhtml+xml`, so such bodies are read fully before being discarded (no leak, just an extra read).

PR #1571, opened 2026-09-17
fix(importer): include all color attributes in style cache key (#702)

### LINKED ISSUE #702: bug(importer): color cache key collides for scrgbClr/hslClr/sysClr → wrong fill colors on import

Cham1229:
## Problem

In `@openmaic/importer`, `StyleResolver.buildColorCacheKey` builds the color cache key from each node's `localName` + `@val` only. However, `resolveColorUncached` reads several OOXML color types from other attributes:

| color type | identity attributes read | in previous cache key? |
|---|---|---|
| `scrgbClr` | `@r` `@g` `@b` | ❌ |
| `hslClr` | `@hue` `@sat` `@lum` | ❌ |
| `sysClr` | `@lastClr` | ❌ |

Because `@val` is empty for `scrgbClr` and `hslClr`, distinct colors (e.g. red `scrgbClr` vs blue `scrgbClr`) produced identical cache keys (`solidFill||scrgbClr:`), causing subsequent colors to reuse the first resolved color.

## Fix

Key each color node by the full set of value-bearing attributes the resolver actually reads — `val, lastClr, r, g, b, hue, sat, lum` — applied to the top-level node and child/grandchild modifier nodes.

## Verification
- Added regression unit tests in `packages/@openmaic/importer/test/styleResolver.colorCache.test.ts` verifying that distinct `scrgbClr`, `hslClr`, and `sysClr` produce distinct cache keys and colors.
- All 18 test files (110 tests) in `@openmaic/importer` pass cleanly.
- `prettier --check` and `eslint` pass with 0 errors.

Closes #702


LING-6150 (commented):
The cache collision is still reproducible on current main, and the added color attributes address the underlying issue. The four regression tests fail on main and pass at 2e2dd366; the importer suite also passes all 110 tests at that head.
The branch currently conflicts with main in packages/@openmaic/importer/package.json: this PR changes 0.2.1 to 0.2.2, while main is already at 0.2.3. Please update the branch against current main and resolve the version conflict without reverting the newer version.
Before approval, we should also verify the visible result with a PPTX containing distinct scrgbClr, hslClr, and sysClr fills. A small constructed fixture is sufficient; please include the file and before/after import screenshots. Once the conflict is resolved, the updated diff can be re-reviewed.

Cham1229:
Thanks for the thorough review @LING-6150!
I have addressed all the feedback and verified both automated tests and visual results:
1. **Rebase & Version Bump**: Rebased cleanly onto the latest `main` (`a8726ec4`) and bumped `@openmaic/importer` to `0.2.4` in accordance with monorepo release rules. `scripts/check-package-version-bumps.mjs` passes.
2. **Fixture PPTX & Regressions**: Committed `packages/@openmaic/importer/test/fixtures/color-cache-collision.pptx` with 6 shapes across `scrgbClr`, `hslClr`, and `sysClr`. Added unit and fixture import tests in `test/styleResolver.colorCache.test.ts` asserting all 6 shapes resolve to distinct colors.
3. **Bare Node Handling**: In addition to extending `COLOR_KEY_ATTRS` with all color attributes, `resolveColorUncached` in `StyleResolver.ts` now supports bare `scrgbClr` and `hslClr` nodes directly (with strict TypeScript typings).
4. **Visual Verification**:
   - **Before (base `main`)**: Second shape of each color family collided in the cache with the first (Shape 2 and Shape 4 reused Red `#ff0000`; Shape 6 reused Orange `#FF5500`).
   - **After (this PR)**: All 6 shapes preserve their distinct intended colors (`#ff0000` / `#0000ff`, `#ff0000` / `#0000ff`, `#FF5500` / `#00AAFF`).
   
color-cache-collision-comparison

LING-6150:
@Cham1229 Thanks, the substantive review feedback is addressed. I confirmed that the committed fixture is exercised through the full PPTX import path and covers distinct scrgbClr, hslClr, and sysClr fills, including the bare-node cases. The implementation and regression coverage now match the reported failure.
main advanced again after your rebase: it is now at d51f4b8, and @openmaic/importer is already 0.2.5. This branch still conflicts because it sets the package version to 0.2.4.
Please rebase once more onto the latest main and bump the importer to 0.2.6. The source change itself merges cleanly; after that update I’ll do the final head verification.

wyuc (approved):
Approving — thanks @Cham1229, and @LING-6150 for the review. Checked the final head `0e74509`: importer is 0.2.6 against main's 0.2.5, it merges cleanly, and the only changes since LING's last pass are the rebase and version bump. The cache key now covers every attribute `resolveColorUncached` reads, and the fixture exercises distinct `scrgbClr` / `hslClr` / `sysClr` fills through the full import path.

PR #1578, opened 2026-09-18
fix(api): validate pdfContent input

### LINKED ISSUE #1568: generate-flow.md documents pdfContent without a shape; passing a string returns 500 instead of 400
## Summary

`skills/openmaic/references/generate-flow.md` lists `pdfContent` as an optional field but never documents its shape. The API requires an object — `{ text: string; images: string[] }` (`lib/server/classroom-generation.ts`) — so passing a string, which is the natural reading of the docs, fails with an opaque 500 instead of a validation error.

## Reproduction

```
POST /api/generate-classroom
{ "requirement": "…", "pdfContent": "<plain text of the source document>" }
```

## Observed

```
HTTP 500
{"success":false,"errorCode":"INTERNAL_ERROR",
 "error":"Failed to create classroom generation job",
 "details":"Cannot read properties of undefined (reading 'length')"}
```

The same request with `"pdfContent": {"text": "<plain text>", "images": []}` returns `202 queued` immediately.

## Root cause

`buildInputSummary()` in `lib/server/classroom-job-store.ts`:

```ts
pdfTextLength: input.pdfContent?.text.length || 0,
pdfImageCount: input.pdfContent?.images.length || 0,
```

`GenerateClassroomInput.pdfContent` is typed `{ text: string; images: string[] }`, so a string value makes `input.pdfContent.text` `undefined` and the property access throws inside job creation — after the request has already been accepted as valid.

## Expected

1. `skills/openmaic/references/generate-flow.md` should document the accepted shape (what `text` is, and what belongs in `images`), since the skill is the discoverable contract for agents calling the endpoint.
2. `app/api/generate-classroom/route.ts` should validate `pdfContent` and answer `400 INVALID_REQUEST` with an actionable message rather than surfacing an internal `TypeError` as a 500.

## Side note

`pdfContent.images` appears to be counted for the job summary (`pdfImageCount`) but is not consumed by the generation pipeline — the pipeline only reads `pdfContent?.text` (`lib/server/classroom-generation.ts`). If images are intentionally unused, documenting that (and dropping or ignoring the key explicitly) would avoid the impression that image-bearing input is supported.

Qnh233:
## Summary

Validate `pdfContent` at the `POST /api/generate-classroom` boundary so malformed payloads return a 400 instead of reaching job creation and surfacing as a 500. Document the accepted shape and add focused route-level regression coverage.

AI-assisted: yes. The implementation and final diff were reviewed locally before submission.

## Related Issues

Fixes #1568

## Changes

- validate `pdfContent` as an object with `text: string` and `images: string[]`
- return `INVALID_REQUEST` / HTTP 400 for malformed shapes
- document the accepted `pdfContent` contract in `generate-flow.md`
- add route-level regression tests for malformed and valid payloads

## Type of Change

- [x] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Send `POST /api/generate-classroom` with `pdfContent: "plain text"`.
2. Verify the route returns HTTP 400 / `INVALID_REQUEST` after this change.
3. Run the focused API tests and required project checks.

- `pnpm exec vitest run tests/api/generate-classroom.test.ts tests/api/verify-pdf-provider.test.ts` — 14/14 passed
- `pnpm format` completed; relevant files pass Prettier
- `pnpm lint --fix` completed with 0 errors (16 pre-existing warnings remain)
- rebuilt stale local `@openmaic/dsl` output, then `npx tsc --noEmit` passed
- self-reviewed the final diff and corrected the documentation wording to match the validator behavior
- confirmed unrelated install-induced `package.json` changes are not included in this commit

### Evidence

- [x] Manually tested locally

## Checklist


wyuc (approved):
Approving. Validating `pdfContent` at the route boundary turns malformed payloads into a clear `400 INVALID_REQUEST` instead of a 500 from job creation, and the skill docs now state the shape with an example. Verified locally: the route suite passes, and restoring the previous route makes all eight invalid-shape cases fail.

One optional follow-up: `pdfContent: null` (and `""`) used to be silently dropped and now returns 400. Clients that send `null` to mean "no PDF" would start failing, so treating `null` like an absent field might be friendlier. Not blocking. Thanks!

PR #1583, opened 2026-09-18
fix(generation): reject invalid interactive scripts

### LINKED ISSUE #1582: [Bug]: Generated interactive HTML with invalid inline JS is persisted as an unresponsive widget
## Problem or Motivation

Generated interactive HTML can render normally even when one of its classic inline `<script>` blocks contains a JavaScript syntax error. The DOM/CSS remains visible, so buttons and controls look usable, but their event-binding script never executes and the widget appears unresponsive.

Today `generateWidgetContent()` accepts that HTML as valid content as long as the model response contains an extractable HTML document.

### Deterministic reproduction

Given an interactive model response containing:

```html
<button id="go">Go</button>
<script>
document.getElementById("go").addEventListener("click", () => {
  const broken = ;
});
</script>
```

`generateSceneContent()` currently returns non-null content and reports no `invalid-model-output` failure, even though the browser cannot parse the script.

## Proposed Solution

Before persisting generated interactive HTML, compile-check classic inline JavaScript syntax without executing it.

- Reject syntax-invalid classic inline scripts through the existing `invalid-model-output` path.
- Do not execute generated code during validation.
- Skip data scripts such as `application/json` widget config.
- Skip external scripts.
- Skip `type="module"` scripts because module grammar differs from classic Function-body grammar and existing 3D templates use modules.
- Preserve valid interactive HTML behavior.

## Acceptance Criteria

- [ ] An interactive HTML fixture with a visible button and invalid classic inline JS is rejected as `invalid-model-output`.
- [ ] Valid classic inline scripts remain accepted.
- [ ] Widget-config JSON is not interpreted as JavaScript.
- [ ] External scripts and module scripts are not falsely rejected by the classic-script validator.
- [ ] Generation package tests and type checks pass.

## Area

Interactive simulations / classroom generation

Qnh233:
## Summary

Prevent generated interactive widgets from being persisted when their classic inline JavaScript cannot parse. This targets a failure mode where the DOM/CSS renders normally, controls look clickable, but event-binding code never executes and the widget appears unresponsive.

AI-assisted: yes. The implementation, tests, and final diff were reviewed locally before submission.

## Related Issues

Fixes #1582

## Changes

- add a deterministic syntax validator for classic inline `<script>` bodies without executing generated code
- reject syntax-invalid interactive HTML through the existing `invalid-model-output` recovery path
- skip widget-config/data scripts, external scripts, and `type="module"` scripts to avoid applying classic-script grammar where it does not belong
- add validator unit coverage and an end-to-end scene-content regression
- bump `@openmaic/generation` from 0.3.8 to 0.3.9

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Generate an interactive HTML document with a visible button and a classic inline script containing a syntax error.
2. Before this change, `generateSceneContent()` returns non-null content and reports no failure.
3. With this change, the same output is rejected with `invalid-model-output` before persistence.

- deterministic pre-fix reproduction: invalid inline JS was accepted (`returned: true`, `failures: []`)
- deterministic post-fix reproduction: the same HTML is rejected (`returned: false`, `invalid-model-output`)
- `pnpm --filter @openmaic/generation exec vitest run --exclude test/assets.test.ts` — 145/145 passed locally
- `pnpm --filter @openmaic/generation typecheck` — passed
- `npx tsc --noEmit` — passed
- `pnpm format` — passed
- `pnpm lint --fix` — 0 errors; 16 pre-existing warnings remain
- `pnpm check:package-versions upstream/main` — passed (`generation: 0.3.8 -> 0.3.9`)
- relevant-file Prettier check and `git show --check` — passed
- the one excluded generation test is the existing Windows-only path-separator assertion in `test/assets.test.ts` (`\` vs `/`), unrelated to this change

### Evidence

- [x] Manually tested locally
- [ ] Screenshots / recordings attached (not applicable; no UI change)

## Checklist

- [x] I have added/updated tests as needed


Qnh233:
Closing this in favor of a broader runtime-interaction investigation.

A real generated interactive artifact reproduces the user-facing symptom while containing syntactically valid JavaScript: the page renders, but interaction enters generated state/data logic that throws at runtime, leaving controls effectively unresponsive. That means syntax-only validation would address only one subset of the failure class.

I am withdrawing this PR so the next fix can be grounded in the runtime-error / behavioral verification path rather than adding a narrow validator first. No maintainer action needed on this PR.

PR #1588, opened 2026-09-18
fix(importer): bound pptx zip inflation by default [AI-assisted]

AbelWangYaBo:
## Summary

`ZipParseLimits` has been implemented since this parser was written, but no caller ever passes it — both import paths call `parseZip(buffer)` with no argument — so every bound was inert and a `.pptx` could inflate without limit. This activates the bounds that already exist, and adds one that scales with the input instead of with the deck.

## What changed

- **Defaults, applied per field.** A caller that overrides one bound no longer silently drops the others, and `Number.POSITIVE_INFINITY` switches a single bound off.
- **New `maxCompressionRatio`, default 200:1.** The absolute caps have to be generous to fit real decks (a large deck with embedded video legitimately holds hundreds of megabytes), so they cannot be tight on their own. The ratio does not depend on deck size, and 200:1 is already the number `render-service` applies to the archives it unpacks (`render-service/src/config.ts`).
- Exports `DEFAULT_ZIP_PARSE_LIMITS` and `ZipParseLimits` so callers can extend the defaults rather than restate them.
- Adds `test/zipParseLimits.test.ts` — the first tests for this file. The first case is the regression itself: an archive containing a >200:1 entry is rejected by a `parseZip(buffer)` call with **no** limits argument, which is how every caller calls it.
- Version 0.2.4 → 0.3.0, because `^0.2.4` admits `0.2.5` and this changes default behaviour for existing callers.

## The numbers are a judgement call

I sized them as a backstop for archives that are not decks, not as a constraint on decks: 10k entries, 1 GiB per entry, 2 GiB total, 1 GiB of media. If you would rather these live at the call sites, the agent-runtime path already caps the compressed file at 8 MB (`MAX_IMPORT_BYTES` in `lib/server/agent-runtime/import-pptx.ts`) and could take much tighter bounds; the browser import path has no compressed-size cap at all today, and that is what I sized for. `maxCompressionRatio` is the one I would most expect you to want to tune, and I am happy to change any of them.

## Known residual

These are the sizes the archive declares about itself, and JSZip inflates an entry before this module can measure it. An archive that declares small sizes and inflates past them is still rejected — JSZip's own `data_length` check fires — but only after that one allocation. Bounding that before the fact needs a byte-budgeted inflate inside the read path, which is a larger change than activating the bounds. #1361 made the same point about the skill ZIP guard; I have not tried to fix that class here.

## Verification

- `packages/@openmaic/importer`: 34 test files / 255 tests pass, including the 10 added here.
- The repo's own real deck (`test/fixtures/rendering-regression-demo.pptx`) parses unchanged under the defaults. Measured on it: **21 entries, 53,889 bytes uncompressed, worst entry 9.1:1** — against defaults of 10,000 / 2 GiB / 200:1. That measurement is now a test, so it fails if the defaults are ever tightened past a deck that exists.
- Package build passes, and all 8 workspace packages build.
- `npx tsc --noEmit` — 0 errors.
- `prettier` and `eslint` clean on the changed files.
- `node scripts/check-package-version-bumps.mjs e14c3a1` passes (the CI gate).
- **Not done:** I did not run the app end-to-end with a real `.pptx`. The importer suite is the regression check I ran; say the word if you would prefer this in Draft until someone exercises a deck through the UI.

AI-assisted, and self-reviewed with an AI review pass before opening, as `CONTRIBUTING.md` asks.

Generated with AI assistance


AbelWangYaBo:
Self-review pass, as `CONTRIBUTING.md` asks of AI-assisted PRs. I ran an adversarial review over the diff (correctness, regression risk, and whether the new tests can actually fail), and it found three things — all fixed in f02305b rather than left for you to catch:

**1. A supplied `NaN` disabled a bound silently.** `Number(process.env.MAX_ENTRIES)` on an unset variable is `NaN`, `NaN ?? default` keeps it, and `count > NaN` is false — so the check switched off without saying anything. Supplied values are now validated. Integer bounds accept `Infinity` as well, which the doc comment already promised but the validation refused.

**2. The media branch only re-measured entries that declared no size.** An entry whose declared size the archive understated was never compared against what actually arrived; it is now measured for every media entry. That is after the allocation, necessarily.

**3. The comment on the ratio check claimed entries with undeclared sizes are "caught after their read". They are not.** For the lying-central-directory case, the only thing that stops the entry is JSZip's own `uncompressed data size mismatch`, thrown after inflating. The review demonstrated it: an archive declaring 1,000 bytes for an entry whose deflate stream expands to 64 MiB passes every bound in this PR and costs the full allocation before anything objects. The comment and the note on `DEFAULT_ZIP_PARSE_LIMITS` now say that plainly instead, including that the allocation itself needs a byte-budgeted inflate to bound.

The review also asked whether the defaults can be tightened where the input is already bounded — they can. The agent-runtime path caps the *compressed* file at 8 MB (`MAX_IMPORT_BYTES`), so 200:1 still lets a single call retain roughly 1.6 GB, and since neither `parse()` nor `importPptx()` accepts `limits`, a server caller cannot tighten it without bypassing the high-level API. I left that out here to keep the diff reviewable: happy to add the pass-through and a tighter server-side set as a follow-up, or to move the numbers to the call sites entirely if you would rather own them.

Headroom for the numbers as they stand, on the repo's own deck: 21 entries, 53,889 bytes uncompressed, worst entry 9.1:1, against defaults of 10,000 / 2 GiB / 200:1. That measurement is now a test, so it fails if the defaults are ever tightened past a deck that exists.

One expectation-setting note: `patches` pinned to `^0.2.x` will not pick this up — 0.3.0 is outside that caret range. That is deliberate for a default-behaviour change, but it does mean the fix needs a deliberate upgrade.

Generated with AI assistance

wyuc (changes requested):
Thanks @AbelWangYaBo — activating the existing bounds is the right fix, and the per-field default merge, `Infinity` opt-out, NaN rejection and frozen `DEFAULT_ZIP_PARSE_LIMITS` all behave as documented. For an honest archive the bounds trip before any inflation (a 1 GiB all-zero entry is rejected in ~2 ms with no RSS growth). One blocking issue, though.

### Blocking: the 200:1 ratio rejects legitimate decks

`maxCompressionRatio` is checked per entry for every entry, media and XML alike. Already-compressed media (PNG/JPEG/video) sit near 1:1, so the entries that actually reach high ratios are uncompressed bitmaps, PCM audio and repetitive XML — which are ordinary deck content. I built otherwise-valid `.pptx` files and ran them through a default `parseZip(buffer)`:

| Entry | Ratio | Default `parseZip` |
|---|---|---|
| 4000×3000 solid-colour 24-bit BMP | 1027:1 | rejected |
| 1920×1080 white 24-bit BMP | 1020:1 | rejected |
| 1920×1080 white BMP with a grid and text (a typical diagram/screenshot) | 297:1 | rejected |
| 30 s 44.1 kHz stereo silent PCM WAV | 1016:1 | rejected |
| slide XML with 20,000 empty `<a:p>` | 308:1 | rejected |
| 800×600 gradient BMP | 14:1 | accepted |
| six real-world decks (highest entry) | ≤ 19:1 | accepted |

DEFLATE tops out around 1032:1, so any fixed threshold below that can reject a legitimate entry. Because `parse()` and `importPptx()` both call `parseZip(buffer)` with no way to pass limits, affected users can't work around it, and the browser import path reports it as "invalid PPTX file".

Suggested fix: skip the ratio check for `ppt/media/` and `ppt/embeddings/` and rely on the absolute caps there (`maxEntryUncompressedBytes` / `maxMediaBytes` already bound honest archives before inflation). Keep the ratio for XML/text parts — real decks peak around 19:1, so 200 leaves plenty of headroom (500 would also be defensible). A test with a solid-colour BMP and a silent WAV would pin it. On the render-service precedent: that 200:1 applies to our own exported HTML/JS/CSS bundles, whose assets are mostly already compressed, so it doesn't transfer to `.pptx` media.

### Please adjust the description

The bounds only see *declared* sizes. An archive whose central directory under-declares an entry (declared 1000 B, actual 1 GiB) passes every check and is fully inflated before JSZip throws `uncompressed data size mismatch` — peak RSS ~1.19 GB in my run, and with `maxConcurrency` 8 several such entries can inflate in parallel. You already list this as a known residual; please keep the description from implying zip bombs are fully covered.

### Non-blocking

- `maxEntries` isn't validated as an integer (`1.5` or a negative value is accepted); passing `{ integer: true }` as for `maxConcurrency` would match.
- It would help users if limit failures surfaced a distinct message (file too large / unusual compression) instead of the generic invalid-file error.
- The description's test counts are slightly off (the new file has 13 cases; the importer suite is 34 files / 258 tests, all green).

Checked: no conflict with `main`, version 0.2.5 → 0.3.0 passes the version guard, `tsc` and prettier are clean, and reverting the default merge turns 3 of the new tests red.

AbelWangYaBo:
Addressed in cd12e14. Your blocking point was right and my reasoning behind the 200:1 was wrong — I took it from `render-service/src/config.ts` without checking that it applies to their own exported HTML/JS/CSS bundles, where the assets are mostly already compressed. It does not transfer to `.pptx` media, and your measurements make that unambiguous.

**The fix is the one you suggested:** the ratio now applies to text parts only (`ppt/media/` and `ppt/embeddings/` exempt), with binary parts left to `maxEntryUncompressedBytes` and `maxMediaBytes`, which for an honest archive are checked before anything is inflated. The interface doc and the note on `DEFAULT_ZIP_PARSE_LIMITS` both say so, and the old "sized like render-service" sentence is gone.

The exemption is pinned from both sides, so neither direction can regress unnoticed:

- a solid-colour 1920x1080 24-bit BMP, 30 seconds of silent stereo PCM, and an embedded object all parse under the defaults;
- the same low-entropy payload as a text part is still refused;
- a slide part holding 20,000 empty `<a:p/>` — your ~300:1 case — is still refused, since that is the shape the text-part check exists for.

**Also from your review:**

- `maxEntries` now takes `{ integer: true }`, so `1.5` and negatives are refused like `maxConcurrency`; there is a test for it.
- Description rewritten. It no longer implies coverage: it now states plainly that an under-declared entry defeats every bound here and is stopped by JSZip only after inflating, and it carries your ~1.19 GB figure and the point that `maxConcurrency` 8 lets several such entries inflate at once. Correction noted on the counts too — 19 cases in the new file, 35 files / 271 tests in the suite.

**On the distinguishable error message** — I'd rather not fold it in here. Making limit failures surface as something other than a generic invalid-file error means a distinguishable error class plus changes at the call sites, and the browser path loads a prebuilt bundle that has to be re-synced, which is a different kind of change from this one. Happy to do it as a small follow-up PR, or to add it here if you'd prefer to review it in one go — your call.

Re-verified on ed18042: 271 tests pass, `tsc --noEmit` clean, prettier/eslint clean, version gate passes. The PR is mergeable against `main` again after a rebase.

A note on that rebase, since it is why the commit hash moved twice: the branch had to be rebased onto an upstream commit *before* `6914e03` to push it at all — that commit changes a workflow file, and the push token here has no `workflow` scope, so any base at or after it is rejected. The resulting PR diff against `main` is the four files below.

wyuc (approved):
Thanks @AbelWangYaBo — the blocker is resolved, and the rest of the feedback is addressed. Verified independently at cd12e14:

- **Legitimate decks parse under the defaults**: a solid-colour 1920x1080 BMP (~1000:1) and 30 s of silent PCM (~1016:1) under `ppt/media/`, and a constant 1 MiB object under `ppt/embeddings/`, all parse through `parseZip` and the public entry points.
- **The exemption can't be used to smuggle a text part past the bounds**: I tried 19 crafted entry paths (case variants, `./`, leading `/`, backslashes, `..` segments, percent-encoding, doubled slashes). A slide/XML part never gains the exemption. Paths that normalize into `ppt/media/` are consumed as binary media, never as slides, and remain bounded by the absolute caps before inflation. The high-ratio slide XML is still refused.
- **The absolute caps still apply on exempt paths**, before inflation: an over-declared entry and cumulative media over `maxMediaBytes` are both rejected.
- Removing the media exemption, or exempting every path, turns the new tests red; the full importer suite, lint, typecheck and the version guard (`0.2.6 → 0.3.0`) pass.

One small, non-blocking note: the `maxEntries` integer check isn't pinned by a test. Dropping `{ integer: true }` leaves the suite green, because a fractional value is also rejected earlier by the pre-scan. A test that exercises `resolveLimit` directly with `maxEntries: 1.5` would pin it.

Approving. I'll update the branch and merge once CI is green.

PR #1590, opened 2026-09-18
refactor(upload): derive the workbench material MIME policy from the shared format registry

### LINKED ISSUE #1589: Unify the duplicated MIME tables between the classic upload path and the workbench material policy
### Problem or Motivation

#1498 fixed the Kylin generic-Office-MIME upload failure (#1497) on **both** upload paths, but it did so by adding a parallel MIME policy in `lib/workbench/material-upload-policy.ts` next to the format registry in `lib/document/mime.ts`. The review of #1498 flagged the duplication as a non-blocking follow-up:

> the workbench now carries its own extension→MIME table next to the document-path one; worth unifying later so they cannot drift.

The drift has already happened in both directions:

- The workbench accepts `.csv` (`text/csv`) and `.webm` (`video/webm`), but neither format is registered in `DOCUMENT_FORMATS` — the "single source of truth".
- The workbench's alias map lacks `image/jpg`, `text/x-markdown` and `audio/x-wav`, which the document registry already knows. A browser reporting `image/jpg` for a JPEG (or `text/x-markdown` for Markdown) is rejected with 415 by `POST /api/materials` even though the file is supported.
- The generic-MIME fallback set (octet-stream + zip family + `vnd.ms-office`) and the WPS Office alias trio are duplicated verbatim in both files.

### Proposed Solution

Share the **mechanism**, keep the **policy** difference:

- `lib/document/mime.ts` stays the single source of truth. Register `csv` and `webm` in `DOCUMENT_FORMATS`, add `audio/x-m4a` to m4a's `aliasMimes`, and use `M.webm` in `LOCAL_FFMPEG_MEDIA_MIMES`.
- `lib/workbench/material-upload-policy.ts` becomes fully derived:
  - `resolveWorkbenchMaterialMime` delegates to `normalizeDocumentMimeType` (same algorithm: empty/generic → extension fallback, alias → canonical, otherwise verbatim for the whitelist to reject).
  - `WORKBENCH_MATERIAL_MIME_TYPES` / `EXTENSIONS` / `ACCEPT` derive from a list of registry format ids (+ `audio/webm`, which has no distinct extension) via the registry helpers.
  - `MEDIA_MIME_TYPES` (the route's media/document byte-limit split) derives from the registry media ids.
  - All existing export names are kept, so `app/api/materials/route.ts`, `lib/workbench/session-store.ts` and `components/workbench/compose-extras.tsx` need no changes.
- The whitelists intentionally stay separate: the classic path is provider-scoped (depends on the selected extractors), the workbench path is a fixed, extractor-independent list. That is product policy, not duplication.

Adding a format anywhere then becomes a one-line registry change that both paths pick up.

### Acceptance Criteria

- [ ] `DOCUMENT_FORMATS` registers `csv` and `webm`; `audio/x-m4a` is a curated alias.
- [ ] `material-upload-policy.ts` contains no hand-maintained extension→MIME table, alias map, or generic-MIME set — everything derives from `@/lib/document/mime`.
- [ ] All existing behavior is preserved: generic-MIME extension fallback, WPS alias mapping, `x-unknown` + `lesson.pdf` still 415s, 415-before-400 error precedence.
- [ ] Workbench now also accepts `image/jpg`, `text/x-markdown` and stores `audio/x-wav` canonically as `audio/wav` (alias normalization previously missing on this path).
- [ ] Classic-mode whitelists (`SUPPORTED_COURSE_MATERIAL_MIME_TYPES`, provider accept strings) are unchanged — csv/webm are registered but accepted by no document provider.
- [ ] A drift-guard test pins the workbench tables to the registry.
- [ ] `pnpm test`, `tsc --noEmit`, eslint and prettier pass.

### Alternatives Considered

- Moving the registry into a new `lib/upload/mime.ts` shared module — rejected for now: `lib/document/mime.ts` already declares itself the single source of truth and both paths can import it; a rename is churn without behavioral benefit.
- Deriving the workbench whitelist from the provider capability matrix — rejected: the workbench accepts formats no document extractor handles (csv) and its accepted set is intentionally extractor-independent.

### Area

Import / export

cosarah:
## Summary

The review of #1498 flagged a non-blocking follow-up: the workbench now carries its own extension→MIME table next to the document-path one, and they can drift. They already had — in both directions. This PR makes `lib/document/mime.ts` the single source of truth for both upload paths. Closes #1589.

## What changed

**Share the mechanism, keep the policy.** The classic path's whitelist is provider-scoped (depends on the selected extractors); the workbench's is a fixed, extractor-independent list. That difference is product policy and stays. Everything mechanical — extension→MIME table, alias normalization, generic-MIME fallback set — now lives once in the format registry:

- **`lib/document/mime.ts`** — registers `csv` and `webm` (accepted by no document provider, so classic-mode whitelists and accept strings are unchanged) and adds the `audio/x-m4a` alias to m4a. `LOCAL_FFMPEG_MEDIA_MIMES` now uses `M.webm`.
- **`lib/workbench/material-upload-policy.ts`** — rewritten as pure derivation: `resolveWorkbenchMaterialMime` delegates to `normalizeDocumentMimeType` (same algorithm: empty/generic → extension fallback, alias → canonical, otherwise verbatim for the whitelist to reject), and the whitelist / extensions / accept string / media split derive from registry format ids (+ `audio/webm`, which has no distinct extension). All export names are kept, so `app/api/materials/route.ts`, `lib/workbench/session-store.ts`, and `components/workbench/compose-extras.tsx` are untouched. The hand-rolled `MIME_ALIASES`, `MIME_BY_EXTENSION`, `GENERIC_MATERIAL_MIME_TYPES`, `normalizeWorkbenchMaterialMime`, and `MEDIA_MIME_ALIASES` are gone.

## Behavior

Everything #1498 pinned is preserved: generic-MIME extension fallback, WPS alias mapping, `x-unknown` + `lesson.pdf` still 415s, 415-before-400 precedence, and the media/document byte-limit split (`audio/x-wav` normalizes to `audio/wav` before the set check, so the limit outcome is unchanged).

Latent gaps close as a side effect — the workbench gate now accepts the registry's curated aliases its own map didn't know, real browser-reported MIMEs that previously 415'd on `POST /api/materials`: `image/jpg`, `text/x-markdown`, `audio/mp3` (→ `audio/mpeg`), and `audio/x-wav` (stored canonically as `audio/wav`).

## Verification

- **Differential vs main**: 2,419 (mime × filename) pairs compared old-vs-new at route level (`{stored mime, accepted, media-limit}`) — zero accepted→rejected transitions; divergences were exactly the alias closures above plus charset-param inputs, which are unreachable (the route strips params before resolving; `File.type` never carries them). `WORKBENCH_MATERIAL_EXTENSIONS` identical in content and order; `WORKBENCH_MATERIAL_ACCEPT` same string set; classic-path constants (`SUPPORTED_COURSE_MATERIAL_MIME_TYPES`, `COURSE_MATERIAL_ACCEPT`, provider capability sets) byte-identical to main. `mime.ts` is a pure leaf module — no server-only imports reach the client bundle via the new import.
- **End-to-end with real files** (live `next dev` + postgres, real OOXML/PDF/JPEG/WAV/WebM/OLE bytes): 20/20 cases passed — all 15 workbench cases incl. the #1497 `vnd.ms-office` scenario, the WPS aliases, the new alias accepts, and every 415 boundary; all 5 classic cases (mime gate verified via distinguishable post-gate provider errors; mineru/mineru-cloud legacy split intact). Not covered: real MinerU content parsing and real browser picker dialogs (accept-string wiring verified at code level).
- Targeted suites 64/64; `tsc --noEmit`, eslint, prettier clean.
- Full suite: 8359 passed. Two full-suite runs each dropped a *different* handful of 5s-timeout tests in unrelated areas (chat rename UI, chat-storage, scene retry); all of them pass in isolation in this worktree and on unmodified main — machine-load flakiness, not this change.


wyuc (approved):
Approving — thanks @cosarah. Verified with a two-tree differential (merge-base vs head) over ~2.3k MIME × filename cases: no previously accepted upload is now rejected, no active-content type (html/svg/js/xml/css) is accepted, the media/document size-limit classification is unchanged, and the classic path's whitelists, accept strings and extension lists are byte-identical across all 32 provider combinations. The newly accepted inputs are the intended alias fixes (`image/jpg`, `text/x-markdown`, `audio/mp3`, `audio/x-wav` params). Mutating the alias branch or leaking `csv` into a classic whitelist turns the new tests red.

Non-blocking notes:
- Classic `.webm` with an empty/generic MIME now reaches the media path via extension fallback (previously a 400); consistent with explicit `video/webm`, maybe worth a line in the description.
- `MEDIA_MIME_TYPES` no longer lists `audio/x-wav` — it now assumes input has gone through `resolveWorkbenchMaterialMime`; a doc comment would help future callers.
- The accept string lists only some of the aliases the gate now normalizes (harmless, since extensions cover the picker).

PR #1593, opened 2026-09-19
feat: send x-opencode-session for the deepseek provider when configured

nqthiep:
## Summary

Allow the DeepSeek provider to work through the [OpenCode Go](https://opencode.ai/docs/go/) gateway (`https://opencode.ai/zen/go`), which requires callers to send a stable session id in `x-opencode-session` — without it the gateway answers `400`. The header is opt-in via `OPENCODE_GO_SESSION`, so every other endpoint behaves exactly as before.

## Problem

Pointing `DEEPSEEK_BASE_URL` at an OpenAI-compatible gateway that mandates `x-opencode-session` fails every request:

```json
{"error":{"message":"Request is missing x-opencode-session and cannot be routed efficiently. Please see https://opencode.ai/docs/go/#where-can-i-use-it"}}
```

OpenMAIC's DeepSeek provider (`type: 'openai'`) had no way to attach custom headers — `createOpenAI` was built with only `apiKey`/`baseURL`/`name`.

## Solution

Pass the header through `createOpenAI`'s `headers` option, gated on both the provider and the env var:

```ts
...(config.providerId === 'deepseek' && process.env.OPENCODE_GO_SESSION?.trim()
  ? { headers: { 'x-opencode-session': process.env.OPENCODE_GO_SESSION.trim() } }
  : {}),
```

- Scoped to the `deepseek` provider id: other providers (OpenAI native, Azure, Kimi, …) never see the header.
- Opt-in: unset/blank `OPENCODE_GO_SESSION` leaves the client byte-identical.
- The session value itself stays deployment configuration (`.env.local`, e.g. a stable UUID) — the app only forwards it.

Verified live against the gateway: `POST /chat/completions` with the header returns `200` and the expected model; without it, `400`.

## Changes

| File | Change |
| --- | --- |
| `lib/ai/providers.ts` | Conditional `headers` on the OpenAI client for `deepseek` (+6 lines) |
| `tests/ai/openai-sdk-integration.test.ts` | New `OpenCode Go session header` suite: header sent when configured, absent without the opt-in |

## Testing

- New tests: 2/2 pass; mutation-checked (removing the headers block fails exactly the positive test).
- Existing `openai-sdk-integration.test.ts`: 7/7 pass.
- `tsc --noEmit` clean, prettier clean.

## Notes

- No new dependencies, no schema/DB changes.
- Go usage limits still apply per its own billing (session header only affects routing/caching, not quota).

## Checklist

- [x] No new dependencies
- [x] Tests added for the new behavior, mutation-checked
- [x] Type check + formatting clean
- [x] Verified live against the real gateway (200 with header, 400 without)


nqthiep:
Hi there — a tiny follow-up PR from the same self-hosting journey, kept deliberately small so it's easy to review. 🙂

Just to share the backstory: while verifying the reasoning round-trip fixes from #1486 end-to-end. This PR adds exactly the opt-in plumbing needed for that, nothing more: six lines in `providers.ts`, two recorded-header tests, zero behavior change for anyone who doesn't set the env var.

I know everyone's review bandwidth has been generous across the recent rounds (genuinely appreciated — each round made the code meaningfully better), so I tried to keep this one boring on purpose: small diff, tests included, verified live against the real gateway both with and without the header.

Happy to adjust anything — naming, scope, or test placement — whatever makes it easiest to land. Thanks so much for your time and patience! 🙏

nqthiep:
Closing in favor of folding this into #1486 (same effort, one review thread) — the commit now lives on that branch. Sorry for the extra noise, and thanks!

wyuc (changes requested):
Thanks @nqthiep, the opt-in design is right: nothing changes unless `OPENCODE_GO_SESSION` is set, and the on/off tests pass. One thing to fix before merge:

**The header is gated on provider only, not on the endpoint.** `x-opencode-session` is added to every `deepseek` provider request whenever `OPENCODE_GO_SESSION` is set, whatever the base URL is. A request can carry its own `baseUrl` for the deepseek provider, so the server-configured session id would also be sent to non-OpenCode endpoints, including client-controlled ones. Please gate on the resolved base URL too: only attach the header when its host is the OpenCode Go gateway (`opencode.ai`, path under `/zen/go`). Add a test that a deepseek request to another base URL does **not** carry the header.

Nits:
- Document `OPENCODE_GO_SESSION` in `.env.example`.
- The comment says "per conversation", but it's a single deployment-level value, so reword it.
- Add the header to the credential-stripping list used on cross-origin redirects, if the fetch wrapper has one.
- Add a test that other providers are unaffected.

Issue #1526, opened 2026-09-15
[Bug]: local Docker build hangs indefinitely at @openmaic/importer postinstall and causes OrbStack out-of-memory


AI421:
### Environment

- Device: Mac mini M4
- OS: macOS (please add exact version)
- OrbStack: 2.2.3
- Docker CLI: 29.4.0
- OpenMAIC: current main branch (please add commit hash)
- Deployment: Docker Compose
- Host port: `33000:3000`

### Description

A local Docker build of OpenMAIC hangs indefinitely during the dependency installation stage:

```text
[deps 5/5] RUN pnpm install --frozen-lockfile
```

The build consistently stops at the `@openmaic/importer` postinstall script:

```text
@openmaic/importer@0.2.1 build
node -e "require('fs').rmSync('dist',{recursive:true,force:true})" &&
rollup -c &&
tsc --emitDeclarationOnly

src/index.ts → dist/index.umd.js...
```

The step remained at this stage for more than 8 hours (`29067.9s`) without completing.

### Reproduction

```bash
git clone https://github.com/THU-MAIC/OpenMAIC.git
cd OpenMAIC
cp .env.example .env.local
docker compose up -d --build
```

The compose file mapped the application to host port `33000`:

```yaml
ports:
  - "33000:3000"
```

### Observed behavior

After the build stalled, Docker/OrbStack became unresponsive:

```text
failed to receive status: rpc error: code = Unavailable
desc = error reading from server: EOF
```

OrbStack logs showed repeated out-of-memory errors:

```text
global_oom
Out of memory
restarting container container=docker
```

Existing Docker containers temporarily disappeared from `docker ps -a` while OrbStack was stuck restarting. After a full macOS restart, the existing containers became visible again.

### Expected behavior

The Docker build should either:

1. Complete successfully within a reasonable time; or
2. Fail with a clear error instead of consuming the Docker VM memory indefinitely.

### Additional information

The Mac mini M4 is capable of running the existing Docker workload. The failure appears to be related to the local Docker build process, particularly the `@openmaic/importer` postinstall build, rather than GPU support or Apple Silicon compatibility.

Please let me know if additional build logs, system information, or a minimal reproduction are needed.

### Steps to Reproduce

1. Use a Mac mini M4 with OrbStack 2.2.3 and Docker CLI 29.4.0.
2. Clone the OpenMAIC repository:

   ```bash
   git clone https://github.com/THU-MAIC/OpenMAIC.git
[...]

Issue #1506, opened 2026-09-15
[Bug]: local build docker image failed


f39516046:
### Bug Description

execute build cmd as  below:

docker build   --build-arg ALPINE_MIRROR=mirrors.tuna.tsinghua.edu.cn   --build-arg NPM_REGISTRY=https://registry.npmmirror.com   -t openmaic:local . 

can not build successfully. 

build error message:

200.5  EPERM  EPERM: operation not permitted, write
200.5 
200.5 
200.5 
------
Dockerfile:38
--------------------
  37 |     
  38 | >>> RUN --mount=type=cache,id=pnpm-store,target=/root/.local/share/pnpm/store \
  39 | >>>     npm_registry="$NPM_REGISTRY"; \
  40 | >>>     while [ "${npm_registry%/}" != "$npm_registry" ]; do \
  41 | >>>       npm_registry="${npm_registry%/}"; \
  42 | >>>     done; \
  43 | >>>     if [ -n "$npm_registry" ]; then \
  44 | >>>       pnpm config set registry "$npm_registry"; \
  45 | >>>     fi && \
  46 | >>>     pnpm install --frozen-lockfile
  47 |     
--------------------
ERROR: failed to solve: process "/bin/sh -c npm_registry=\"$NPM_REGISTRY\";     while [ \"${npm_registry%/}\" != \"$npm_registry\" ]; do       npm_registry=\"${npm_registry%/}\";     done;     if [ -n \"$npm_registry\" ]; then       pnpm config set registry \"$npm_registry\";     fi &&     pnpm install --frozen-lockfile" did not complete successfully: exit code: 255

### Steps to Reproduce

1. git clone 
2. docker build   --build-arg ALPINE_MIRROR=mirrors.tuna.tsinghua.edu.cn   --build-arg NPM_REGISTRY=https://registry.npmmirror.com   -t openmaic:local . 
3. failed

### Expected Behavior

build successfully

### Actual Behavior

200.5  EPERM  EPERM: operation not permitted, write
200.5 
200.5 
200.5 
------
Dockerfile:38
--------------------
  37 |     
  38 | >>> RUN --mount=type=cache,id=pnpm-store,target=/root/.local/share/pnpm/store \
  39 | >>>     npm_registry="$NPM_REGISTRY"; \
  40 | >>>     while [ "${npm_registry%/}" != "$npm_registry" ]; do \
  41 | >>>       npm_registry="${npm_registry%/}"; \
[...]

PR #1598, opened 2026-09-19
fix(docker): build workspace packages in the builder stage, not during install

wyuc:
## Summary

Moves the workspace package builds out of `pnpm install` in the main Docker image, so the `deps` stage only resolves and links dependencies and the builds run as an explicit step in the `builder` stage. Adds a CI job that builds the main image, which CI did not do before.

Context: #1526 reports the build stalling at the importer's rollup during `pnpm install` in the `deps` stage (with the Docker VM running out of memory), and #1506 reports an `EPERM` at the end of the same step. The root `postinstall` builds nine workspace packages inside that install; `@openmaic/importer`'s rollup + terser pass (UMD/CJS/ES with pdfjs inlined) is by far the heaviest part.

## Changes

- **`Dockerfile` / deps:** `pnpm install --frozen-lockfile --ignore-scripts`. pnpm 10 already skips dependency build scripts, so this only skips the root `postinstall`; `sharp` and `@napi-rs/canvas` still load from their prebuilt binaries.
- **`Dockerfile` / builder:** `RUN NODE_OPTIONS=--max-old-space-size=1024 pnpm run build:packages` runs the same chain explicitly. `public/vendor` is produced here now (by `sync-maic-importer.mjs` at the end of the chain), so `COPY --from=deps /app/public/vendor` is removed. Note that `--max-old-space-size` bounds V8's old space, not the total memory of the step.
- **`package.json`:** the chain moves to a new `build:packages` script (byte-identical to the old `postinstall`), and `postinstall` delegates to it — local `pnpm install` behaviour is unchanged.
- **CI:** new `docker-main-image` job builds the `deps` + `builder` stages (buildx, GHA cache). Because the builder copies the whole context, it runs for every change except docs-only ones (`docs/**`, `*.md`). Not a required check.

## Measurements (same constraints for base and head)

Each cell is a fresh `node:22-alpine` container (aarch64, cgroup v2, 2 CPUs) with the deps-stage apk packages, a hard memory limit and **swap disabled** (`--memory == --memory-swap`), and a pre-warmed pnpm store so only install/build work is compared. *MaxRSS* is GNU time's largest single process; *oom / oom_kill* are read from the container's `memory.events`.

| Step | Limit | Wall | MaxRSS | oom / oom_kill | Exit |
|---|---|---|---|---|---|
| base: `pnpm install --frozen-lockfile` (includes the package builds) | 2500m | 0:58 | 0.98 GiB | 0 / 0 | 0 |
| | 1536m | 1:02 | 0.79 GiB | 0 / 0 | 0 |
| | 1024m | 0:33 | — | 0 / 0 | **134** — V8 `Reached heap limit` in the importer's `rollup -c` (controlled, not an OOM kill) |
| head: deps `pnpm install --frozen-lockfile --ignore-scripts` | 2500m / 1536m / 1024m | ~0:15 | 0.31 / 0.28 / 0.26 GiB | 0 / 0 | 0 |
| head: builder `build:packages` (`--max-old-space-size=1024`) | 2500m | 0:43 | 0.94 GiB | 0 / 0 | 0 |
| | 1536m | 0:43 | 0.92 GiB | 0 / 0 | 0 |
| | 1024m | 0:44 | 0.89 GiB | 0 / 0 | 0 |

What this shows: the `deps` stage no longer carries the build (its peak drops ~4x and it becomes a cacheable dependency-only layer), and the workspace package build completes under a 1 GiB container limit where the current layout fails (the subsequent `pnpm build` was not measured under that limit). At 1024m the difference comes largely from the explicit heap limit — Node otherwise derives a smaller default heap from the cgroup limit. No cgroup OOM kill occurred in any cell, so these measurements do **not** reproduce the VM-level hang / disruption described in #1526; I'm linking it as related rather than claiming it fixed.

The `EPERM: operation not permitted, write` from #1506 did not reproduce here with either the default registry or `registry.npmmirror.com` (cold and warm stores). Its cause is still open; it occurred at the end of the same install step, which no longer runs the builds.

## Verification

- CI `Main image (deps + builder)` builds both stages successfully.
- Local `pnpm install --frozen-lockfile` on this branch still runs `postinstall → build:packages`, builds all packages and syncs `public/vendor/maic-importer`.
- `actionlint` and prettier clean on the changed files; the docs-only filter was checked against docs-only, mixed, source-only, Dockerfile and empty change sets.

Refs #1526. Refs #1506.

https://claude.ai/code/session_01DYifP8wM4XJQ3Hc2qsF6zf


LING-6150 (commented):
Thanks for separating the workspace builds from dependency installation and adding Docker build coverage. The build sequence looks consistent with the original postinstall, and the CI log confirms that the deps and builder stages completed successfully.
One point I’d like to clarify before approval: was the post-change ~953 MiB measurement taken under the same memory limit as the failing baseline? Could you share a base/head comparison under the same constrained conditions, including the commands, architecture, memory/swap limits, peak memory, and exit outcome? Cgroup OOM counters would also help distinguish a controlled failure from an OOM kill.
It doesn’t need to build successfully at 1 GiB—a prompt, clear failure would also meet #1526’s stated expectation. What we still need to establish is whether the change addresses the reported hang and VM disruption. Since --max-old-space-size=1024 limits V8 old space rather than total build memory, it would be helpful to either support the VM-OOM claim with that comparison or narrow the wording to the improvement demonstrated.
Two smaller, non-blocking suggestions:
- The CI path filter currently misses builder inputs such as next.config.ts and application source files. Could we include those, or clarify that the check covers a narrower set of changes?
- Since #1506 reports Ubuntu, I’d suggest leaving the filesystem cause open rather than attributing it to a macOS VM. Keeping Refs #1506 makes sense.
The overall direction looks worthwhile; the constrained-memory comparison would help close the remaining verification gap.

wyuc:
Thanks @LING-6150 — all three points were right, and the earlier numbers weren't comparable. I re-measured base and head under identical constraints and updated the description.

**Same-constraint comparison** (fresh `node:22-alpine` container per cell, aarch64, cgroup v2, 2 CPUs, hard memory limit with swap disabled, pre-warmed pnpm store; OOM counters from `memory.events`):

- **base** `pnpm install --frozen-lockfile` (builds inside install): succeeds at 2500m / 1536m (MaxRSS 0.98 / 0.79 GiB); at **1024m it fails with exit 134, V8 `Reached heap limit` in the importer's `rollup -c`** — a controlled heap failure, `oom_kill=0`.
- **head** deps `--ignore-scripts`: succeeds at 2500m / 1536m / 1024m in ~15 s, MaxRSS 0.26–0.31 GiB.
- **head** builder `build:packages` with `--max-old-space-size=1024`: succeeds at all three limits in ~43 s, MaxRSS 0.89–0.94 GiB.
- No cell recorded `oom` or `oom_kill` > 0.

So you're right on the claim: these runs don't reproduce the VM-level hang/disruption from #1526, and at 1024m the difference is largely the explicit heap limit (Node otherwise sizes a smaller default heap from the cgroup limit) rather than a cap on total memory. I've narrowed the description to what's demonstrated — the deps stage no longer carries the build (~4x lower peak, cacheable dependency-only layer) and the full build completes at a 1 GiB limit where the current layout fails — and switched #1526 to `Refs`.

**CI filter:** agreed — the builder copies the whole context, so the job now runs for every change except docs-only ones (`d585da3`).

**#1506:** removed the macOS-VM attribution; the EPERM cause is left open and it stays `Refs #1506`.

Could you take another look when you have a moment?

LING-6150 (approved):
Thanks for the follow-up and the comparable measurements. The revised scope is much clearer, and switching #1526 to Refs appropriately leaves the original VM-level issue open. I also checked the updated CI filter against seven change scenarios; the previously missed source and configuration changes now trigger the build.
I have no remaining code blockers. Just two small wording points to align with the updated evidence:
- The Dockerfile comment still suggests that the heap setting prevents a VM-level OOM. Could you update it to reflect the distinction already explained in the PR description?
- The measurements demonstrate that the workspace package build completes at 1 GiB. Unless the subsequent pnpm build was also tested under that limit, “workspace package build” would be more precise than “full build.”
Neither requires an implementation change. I have clicked approve first, you can modify the pr body before merge.

PR #1592, opened 2026-09-18
docs: align security and behaviour claims with shipped code (#1587)

Frank-zhu0404:
## Summary

Aligns docs and the persistence auth module comment with shipped behaviour for ACCESS_CODE (fail-open when unset), document reads (capability-by-id via anonymous owner cookie, not `x-learner-key`), and the README action/skill counts (21 / 24).

AI-assisted (Grokbuild + Codex review). Self-reviewed against `middleware.ts`, persistence routes, `Action` union, and `skills/agent-runtime/*/SKILL.md`.

## Related Issues

Closes #1587

## Changes

- `SECURITY.md` / `README.md` / `README-zh.md` / `.env.example`: document that unset `ACCESS_CODE` is fail-open; correct persistence token / cookie / capability-by-id wording; update action count to 21 and skills to 24.
- `lib/persistence/server-auth.ts`: comment-only — same facts as the README warning (no runtime logic change).

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Read #1587 items 1–3 against `main`.
2. Diff this branch: confirm fail-open wording, persistence warning, and counts 21 / 24.
3. Grep docs for stale marketing claims (`28+`, `20 built-in`, `Twenty built-in`, “choose an x-learner-key” as document R/W).

### What you personally verified

- `middleware.ts`: unset `ACCESS_CODE` → `NextResponse.next()` (fail-open).
- Persistence: docs/assets skip authenticator; owner from anonymous cookie; reads capability-by-id.
- Counts: `Action` union has 21 members; `skills/agent-runtime` has 24 `SKILL.md` directories.
- Diff is docs + one comment block only — no behaviour change.
- Did **not** run full `pnpm` CI (docs/comments only); will rely on GitHub CI after opening draft.

### Evidence

- [ ] CI passes (`pnpm check && pnpm lint && npx tsc --noEmit`)
- [x] Manually tested locally (grep / source citation checks above)
- [ ] Screenshots / recordings attached (if UI changes)

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [x] I have added/updated documentation as needed
- [x] My changes do not introduce new warnings

PR #1599, opened 2026-09-19
fix(server): warn when access-code protection is disabled

### LINKED ISSUE #1594: Warn loudly at startup when ACCESS_CODE is unset on a network-exposed deployment
## Problem

When `ACCESS_CODE` is unset, `middleware.ts` lets every request through (fail-open), so the entire API surface is reachable without a credential. That is the intended local-first default, but nothing at runtime tells an operator who has exposed such a deployment to a network that it is unprotected. GHSA-9m7h previously (incorrectly) stated the middleware fails closed, so some self-hosters may assume they are safe. See #1587.

## Proposal

Keep fail-open as the default, but make an unprotected, network-exposed deployment hard to miss:

1. At server startup, if `ACCESS_CODE` is unset **and** the server is listening on a non-loopback address (e.g. `HOSTNAME=0.0.0.0`, or running in a container), log a prominent, one-time warning explaining that all API routes are unauthenticated and how to set `ACCESS_CODE`.
2. Consider surfacing the same state in `GET /api/health` (e.g. `accessCodeConfigured: false`) so deployment checks and monitoring can catch it.
3. Optionally, an opt-in `REQUIRE_ACCESS_CODE=true` that refuses to start when `ACCESS_CODE` is unset, for operators who want fail-closed.

`lib/server/access-code-warning.ts` already warns once when the code is too short; the startup warning could live alongside it.

## Out of scope

Changing the default to fail-closed (it would break the zero-config local experience).

Hosuke:
## Summary

An instance started without `ACCESS_CODE` currently gives operators no startup indication that the access-code gate is disabled. Log one actionable warning from Node instrumentation and expose `accessCodeConfigured` in `/api/health` for deployment probes. Access behavior and health availability are unchanged.

AI-assisted with Codex; independently reviewed by a second Codex agent.

## Related Issues

Fixes #1594

## Changes

- Add a missing-code startup warning with its own once guard, independent of the existing short-code warning.
- Report only a boolean in health, using the same runtime truthiness as middleware (including empty and whitespace values).
- Document the startup warning and health field in `.env.example`.

The warning is deliberately conservative: Next's instrumentation hook does not expose the actual listener address. `next dev` and `next start` default to wildcard binding and ignore `HOSTNAME` (standalone uses it); the private display origin can also map wildcard binding to localhost. A loopback exemption based on these values would miss unprotected deployments. Local-only instances therefore also receive the one-time notice, asking operators to configure the code **before exposing the server to a network**.

## Type of Change

- [x] Documentation update

### Steps to reproduce / test

1. Start with `ACCESS_CODE` unset or empty. The startup log explains the disabled access-code gate and how to configure it; repeated health requests do not repeat the warning.
2. Fetch `/api/health` without credentials: it remains HTTP 200 and includes `accessCodeConfigured: false`.
3. Restart with a configured code. The missing-code warning disappears, health reports `true`, and a protected API request without credentials still receives HTTP 401.

- Five focused suites passed: 51 tests covering the new startup and health behavior, existing access-token/verify handling, and instrumentation's asset-collector integration.
- A separate reviewer reran the 15 new tests and checked middleware parity, Edge exclusion, build-phase handling, and independence of the warning guards; no actionable findings.
- `pnpm check`, `pnpm lint`, `pnpm exec tsc --noEmit`, and `pnpm check:i18n-keys` passed. ESLint reports 18 existing warnings in unchanged files and zero errors.
- Live Next.js smoke on loopback, without a database or provider keys: unset/empty code produces one warning across repeated health requests and reports `false`; restarting with a placeholder code produces no missing-code warning, reports `true`, and rejects an unauthenticated protected API with HTTP 401. Both temporary servers were stopped.
- Full root suite: `pnpm test --maxWorkers=2` passed **740 files / 8,405 tests**, with 12 files / 98 tests skipped by their existing conditions. The initial default-worker run overlapped static checks and hit timing limits in unrelated suites; it was stopped, then the entire suite was rerun with two workers after static checks completed.
- Node engine, internal dependency range, package-version, and diff-whitespace checks passed.

### Evidence

```text
Unset:      GET /api/health -> 200, accessCodeConfigured: false; warning count = 1
Configured: GET /api/health -> 200, accessCodeConfigured: true; warning count = 0
Configured: GET /api/foo    -> 401 (no access cookie)
Test Files  740 passed | 12 skipped (752)
Tests       8405 passed | 98 skipped (8503)
```

- [x] Local checks pass (`pnpm check && pnpm lint && pnpm exec tsc --noEmit`)
- [x] Manually tested locally

## Checklist

- [x] My changes do not introduce new warnings other than the intentional diagnostic described above


LING-6150:
Thanks for the focused fix. I reviewed head `f9acd24f` against #1594 and did not find any blocking issues.

The change preserves the existing access behavior: an unset or empty `ACCESS_CODE` still allows zero-config use, `/api/health` remains accessible without credentials, and protected API requests still require authentication when a code is configured. The new health field matches middleware's runtime check, and neither the warning nor the response exposes the code.

I independently ran seven relevant test files / 71 tests covering startup warnings, health, middleware, access-code verification, token behavior, and instrumentation integration; all passed. I did not independently rerun the full build or live Next.js startup smoke.

One implementation choice worth noting for the final review: the warning also appears on loopback-only deployments. The PR explains why the startup hook cannot reliably determine the actual listener address, and the wording advises configuring the code before network exposure rather than claiming the instance is already exposed. This seems like a reasonable conservative diagnostic while preserving the local-first default.

cc @wyuc to check the scope.

wyuc (dismissed):
Approving — thanks @Hosuke, this covers both parts of #1594. Warning unconditionally rather than trying to infer the bind address is the right call: `next dev` / `next start` ignore `HOSTNAME` and bind a wildcard, so treating a missing value as loopback would be the dangerous direction. The warning never includes the code, and `accessCodeConfigured` gives deployment checks something to assert on.

Two non-blocking thoughts:
- Local zero-config users now see this on every start; suppressing it under `next dev` would keep the signal for production builds without the noise.
- `/api/health` is unauthenticated, so `accessCodeConfigured: false` tells an anonymous caller the deployment is ungated. Minor, since such a deployment is discoverable anyway, but worth being deliberate about.

wyuc:
Merged `main` into your branch to resolve a conflict in `.env.example`: #1592 landed a fail-open note on the same lines, so both notes are now kept (fail-open sentence first, then the runtime-warning lines). No other changes.

wyuc (approved):
Re-approving after the conflict resolution (merge of main; no content change).

PR #1600, opened 2026-09-19
fix(storage): stop chat autosave retries after runtime 401

Hosuke:
## Summary

When a server-backed runtime returns HTTP 401, chat autosave currently keeps retrying with backoff. Pause chat persistence for the current stage visit after that refusal, while retaining unsaved chat state and continuing document saves.

## Related Issues

Related to #1413 — implements the independent client retry slice (item 2). This PR deliberately leaves the issue open: deriving the learner principal on the server (item 1) remains a separate follow-up.

## Changes

- Preserve typed runtime 401 failures through nested chat error causes and distinguish blocked chat changes from transient failures.
- Stop incremental, full-save, and departing-stage chat retries after authentication refusal. Keep chat dirt and the last durable snapshot in memory; reset the pause on a fresh stage load.
- Keep document edits on their normal save cadence, including concurrent flushes. Ignore a late refusal from an earlier visit to the same stage.
- Add scheduler, storage-boundary, and HTTP adapter integration regressions. No authentication policy or server routes change.

## Type of Change

### Steps to reproduce / test

1. Open a classroom with server-backed persistence whose runtime endpoint returns 401.
2. Trigger chat autosave and continue changing chat state. The first refused save reports that autosave is paused; subsequent timers and saves do not keep contacting the runtime.
3. Edit a scene: the document still saves. After authentication is repaired, freshly load the stage and verify chat persistence works again.

### What was verified locally

- 49 focused tests passed across five suites, covering direct/wrapped 401s, transient 503 retries, full/incremental/departing saves, concurrent flushes, retained chat dirt/baseline, and same-stage reloads.
- The integration regression runs the real stage scheduler, stage storage, chat sync, and `HttpRuntimeStore`. A fetch-boundary 401 produces one request across 130 seconds and additional chat changes; a later document edit commits. After a fresh visit, chat writes through an HTTP adapter to `BrowserRuntimeStore` and reads back successfully.
- Full root suite: `pnpm test --maxWorkers=2` — **740 test files passed, 12 skipped; 8,416 tests passed, 98 skipped** (270.54 seconds).
- Browser smoke in Chrome against local `next dev`, with persistence HTTP responses intercepted: opened the classroom, switched scenes, entered a chat draft, and renamed a scene in Pro mode after the 401 pause. The scene update reached `PUT /api/persistence/documents/.../scenes/scene-0`. Runtime request count stayed at 4 over the following 65 seconds (including initial-load requests), with one autosave-pause warning. Screenshots were inspected locally. No live provider or production-auth backend was used.
- Full `pnpm check`, `pnpm lint`, `pnpm exec tsc --noEmit`, and `pnpm check:i18n-keys` passed. Lint reports the existing 18 warnings; none are added by this change.

The pause preserves unsaved chats in the current in-memory visit; it does not make them durable across navigation or reload. This PR does not resolve production learner authentication or add an automatic credential refresh mechanism.

### Evidence

- [x] Local format, lint, TypeScript, and i18n checks pass
- [x] Browser regression smoke completed
- [x] Full unit suite passed

## Checklist

- [x] I have added regression coverage
- [x] My changes do not introduce new lint warnings

AI-assisted with Codex. A separate Codex reviewer found a concurrent-flush timer race; it was fixed, covered by a regression test, and re-reviewed with no outstanding findings.


wyuc:
Thanks @Hosuke — the diagnosis is right and the tests are thorough, but I'm going to close this rather than merge it, for scope reasons.

**The 401 this works around is a symptom of #1413.** In a server-backed deployment the runtime refuses the client because there is no server-derived learner principal — the dev token is all the client has. Once that is fixed (item 1 there), the refusal disappears and a client-side pause state machine has nothing left to protect against. I'd rather spend the effort on that than carry this permanently.

**The cost here is the state, not the idea.** A blocked-save latch, per-visit reset, late-refusal filtering and a blocked-changes concept threaded through the stage store and the storage boundary is a lot of failure-path machinery, and it is the kind that keeps producing subtle bugs: an independent review of this branch already found one — a stale 401 that lands *after* a later successful save still latches, parking a perfectly healthy session for the rest of the visit, because the latch is only cleared on stage reload.

**It also trades a loud failure for a quiet one.** Today the retries are wasteful but visible in the network tab. After this change chat silently stops being saved, with only a server-side log line; recovery within the same visit no longer happens (nothing retries until reload), and a reload is exactly what discards the unsaved chat. Meanwhile document saves continue normally, so the user watches their course changes persist and reasonably assumes the conversation did too.

If the retry storm needs relief before #1413 lands, I'd happily take a much smaller change: stop retrying on a 401 and surface it to the user (a toast or banner saying chat isn't being saved and how to restore it), without the latch plumbing. And for what it's worth, #1599 was exactly the right size for its problem — thanks for that one.

PR #1601, opened 2026-09-19
fix(persistence): correlate 5xx responses with request ids

### LINKED ISSUE #1574: Persistence route answers 5xx (503 on PUT/PATCH/DELETE) with no server-side trace: log them with a request id
**What happens**

On a deployment behind a reverse proxy, the browser saw `PUT /api/persistence/documents/:id/stage`, `PATCH /api/persistence/runtime/sessions/:id/status` and `DELETE /api/persistence/runtime/sessions/:id` answer **503** while `GET`/`POST` on the same route answered 200/201. The chat autosave then looped (new `:generation:N` session per attempt, see #1558).

**Why it could not be diagnosed**

`app/api/persistence/[...path]/route.ts` has exactly one 503 (`PERSISTENCE_DEV_TOKEN_MISSING`) and it does not depend on the method; the storage package never answers 503; and the route logs nothing for a 5xx. From the server side there was no way to tell whether those 503s left the app at all or were produced in front of it — the only evidence was the browser's network tab, with no id to correlate.

**Ask**

Log every 5xx leaving the persistence route with method, path, status, error code and a request id, and echo the id (`x-request-id`) on the response, so a failing save can be matched between browser and server log. A 5xx with no matching server line then provably did not come from the app.

Ai-Eastern:
## Summary

Add correlation IDs to embedded persistence responses and log every returned 5xx with the HTTP method, request path, status, error code, and request ID.

This PR is AI-assisted. I reviewed the complete diff and verified the behavior locally before submission.

## Related Issues

Fixes #1574

## Changes

- Reuse a valid incoming `x-request-id` or generate a UUID, then return it on every persistence response.
- Emit one structured persistence error log for each 5xx response without consuming the client response body.
- Add route-level coverage for 503 correlation logging and invalid request-ID replacement on a non-5xx response.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Run `pnpm vitest run tests/persistence/route.test.ts`.
2. Send a persistence request that returns 5xx and inspect the server log plus the response `x-request-id` header.
3. Send a non-5xx request with an invalid `x-request-id` and confirm it is replaced without an error log.

- Confirmed the new tests failed before the route change because responses had no `x-request-id`.
- Confirmed all 42 persistence route tests pass after the change.
- Confirmed valid incoming IDs are echoed, invalid IDs are replaced with UUIDs, and 4xx responses do not produce 5xx logs.
- Reviewed the diff for response-body consumption, route coverage, and unrelated file changes.

### Evidence

- [ ] Manually tested locally

Local checks:

- `pnpm check` — passed
- `pnpm lint` — 0 errors (existing warnings remain in unchanged files)
- `pnpm vitest run tests/persistence/route.test.ts` — 42 passed

No live server or UI was started; the route handlers were verified directly through route-level tests. Screenshots are not applicable.

## Checklist


wyuc (approved):
Approving — thanks @Ai-Eastern, this is exactly what #1574 asked for. Validating the inbound `x-request-id` against a bounded charset (and falling back to a UUID) keeps log injection out, only 5xx is logged with method/path/error-code/id, and reading the code via `response.clone().json()` is confined to the error path. Both behaviours are covered by tests.

Non-blocking: the echoed id is client-supplied when it passes the pattern, so without a trusted proxy in front two clients could pick the same value. Fine for correlation, worth keeping in mind if it ever feeds something stricter.

PR #1602, opened 2026-09-19
fix(chat): brake autosave after a failed session save

### LINKED ISSUE #1558: Chat autosave re-opens a new runtime session generation on every tick after a failed status/delete (67 generations, ~700 requests in 6 min)
With server-backed persistence, when the runtime store answers a session status `PATCH` or a session `DELETE` with a server error, the chat sync gives up on the current generation and the **next autosave tick opens the same chat as a fresh `…:generation:N` session**. `syncOne` bounds one save (8 attempts × 8 plan steps, with backoff), but every autosave starts a new save, so the schedule restarts from zero.

Measured on a v1.0.3 deployment (6-scene course open for 6 minutes): `session-…:generation:1 … :generation:67` for one scene, ~700 requests to `/api/persistence/runtime/…`, a growing table of empty `runtime_sessions` rows, and a browser tab close to freezing. The store's error body is never surfaced (the client wrapper swallows it), so the operator cannot even see what the server answered.

Proposal (PR follows): a per-stage brake above the sync — after a failed save, skip further saves for that stage until a growing cooldown elapses (5 s → 5 min cap), clear on success, and log the store's status/code/message once per trip.

dyk1454683243-sudo:
## Summary

After a failed chat-session save (for example the store answering `PATCH …/status` or `DELETE` with a server error), every autosave tick re-ran the whole sync and opened a new `…:generation:N` runtime session — 67 generations and ~700 requests in 6 minutes on one course.

This adds a small per-stage brake above `saveChatSessions`: skip while cooling down, clear on success, grow 5 s → 15 s → 45 s → 2 min → 5 min cap, and log the store's HTTP status/code/message once per trip.

[#1559](https://github.com/THU-MAIC/OpenMAIC/pull/1559) sketched this on an older `main` and was closed unmerged. This is a fresh attempt on current `main`. Out of scope: the underlying 503 — this only stops the runaway retry storm.

## Related Issues

**Fixes #1558**

## Changes

- `lib/utils/chat-save-brake.ts`: `chatSaveBrake.allows` / `recordFailure` / `recordSuccess` (`resetAll` for tests).
- `lib/utils/stage-storage.ts`: in `saveStageChats`, skip while braked, release on success, trip on failure. Lock-unavailable errors keep their existing rethrow path and do not trip the brake.
- `tests/utils/chat-save-brake.test.ts`: allow / fail / success / 5-minute cap.
- `tests/document-store/stage-storage-incremental.test.ts`: subsequent autosave skips `saveChatSessions` while cooling down; lock-unavailable rethrow is unchanged.

## Type of Change

- [ ] Breaking change (fix or feature that would change existing functionality)
- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Server-backed persistence on; make `PATCH /api/persistence/runtime/sessions/<id>/status` fail (for example return 503 from a proxy) and keep a course open with a chat.
2. Before: a new `…:generation:N` session every few seconds. After: at most one new attempt per cooldown step; console shows `[ChatSaveBrake] … (HTTP 503 · <code> · <message>); holding autosave for N s`.
3. `vitest run tests/utils/chat-save-brake.test.ts tests/document-store/stage-storage-incremental.test.ts`.

- Re-read current `main`: `saveStageChats` still wraps `saveChatSessions`; a failed save still returns `false` and the next tick would start a fresh generation.
- Local `vitest run` on those two files — 15 passed (allow / fail / growing cooldown / 5-minute cap / success clear / one log line per trip / skip-while-braked / lock-unavailable rethrow unchanged).
- GitHub CI on this PR: Lint/Typecheck/Unit Tests, E2E, Storage Contract, Render Service, and Issue Triage all passed.
- Did not reproduce the live 503 storm in a browser in this environment.

### Evidence

- [x] Unit tests pass (`vitest run` on the files above)
- [ ] Manually tested locally

## Checklist


LING-6150 (changes requested):
Thanks for working on the repeated-save problem. Reducing retries during a storage outage is worthwhile, and the existing 15 focused tests pass. However, placing the brake in the shared saveStageChats path introduces a regression in the departing-course save flow.
Please preserve the existing departing-course retry.
When switching courses, setStage captures the departing course’s pending changes, attempts to persist them, and retries once after 100 ms. If both attempts fail, it logs and drops that snapshot.
With this change, the first failure activates a cooldown of at least 5 seconds. The retry 100 ms later is therefore skipped without calling saveChatSessions, even if storage has recovered. The caller then drops the snapshot.
I checked this at the storage boundary with a transient first-call failure followed by a successful storage response:
- On the base implementation, the second call reaches storage and succeeds.
- At this head, storage is called only once; the second attempt is blocked by the cooldown.
Navigation was already best-effort, but this removes an existing recovery opportunity and can discard chat changes that the previous implementation would have saved. Please scope the brake so it preserves that lifecycle contract, and add a regression test covering a transient failure during the departing-course save and its retry.
Please also clarify the interaction with the existing autosave backoff.
Current base already has consecutiveFlushFailures and an exponential retry delay capped at 30 seconds in lib/store/stage.ts. Could you explain which current path still resets or bypasses that backoff, and how the new per-stage cooldown addresses it? A focused reproduction against current base would help establish the need and intended scope of the additional retry policy.
The direction is useful, but the departing-save regression needs to be resolved before approval.

dyk1454683243-sudo:

PR #1604, opened 2026-09-19
fix(ai): accept Gemini's OpenAI-compatible tool-call stream

### LINKED ISSUE #1556: Gemini OpenAI-compatible endpoint: every tool call fails with 'tool_calls[0].index: expected number, received undefined' (thought_signature not round-tripped)
Setting `OPENAI_BASE_URL` to Google's OpenAI-compatible endpoint (`https://generativelanguage.googleapis.com/v1beta/openai`) and routing the workbench agent driver to `openai:gemini-3.8-flash` (`api: openai-completions`) works for plain chat but **every tool call fails**:

```
Type validation failed … choices[0].delta.tool_calls[0].index: expected number, received undefined
```

Measured chunk from Google (the first tool call of a course-building session, `read …/stage-design/SKILL.md`):
```json
{"choices":[{"delta":{"role":"assistant","tool_calls":[{"extra_content":{"google":{"thought_signature":"…"}},"function":{"arguments":"{\"offset\":1,\"limit\":100,\"path\":\"/app/skills/agent-runtime/stage-design/SKILL.md\"}","name":"read"},"id":"call_297140","type":"function"}]},"index":0}],"model":"gemini-3.8-flash","object":"chat.completion.chunk"}
```

Two gaps versus the OpenAI stream contract: no `tool_calls[].index` (required by `@ai-sdk/openai`'s chunk schema), and `extra_content.google.thought_signature`, which Gemini 3 expects back on the assistant tool call in the next request. The existing compatibility fetch in `lib/ai/providers.ts` is the natural seam. PR follows.

dyk1454683243-sudo:
## Summary

Gemini's OpenAI-compatible endpoint (`OPENAI_BASE_URL=https://generativelanguage.googleapis.com/v1beta/openai`) streams tool-call deltas without `index` (required by `@ai-sdk/openai`'s chunk schema) and with `extra_content.google.thought_signature`, which Gemini 3 expects back on the next request. Plain chat works, but every tool call fails with `Type validation failed … tool_calls[0].index`. This adds a small host-gated shim on the existing `compatFetch` seam.

## Related Issues

**Fixes #1556**

#1557 sketched the same approach (`lib/ai/google-openai-compat.ts` + `compatFetch` wiring + focused vitest) but closed unmerged. This re-implements on current `main` rather than force-pushing that branch.

## Changes

- `lib/ai/google-openai-compat.ts`: `normalizeGoogleToolCallStream` (SSE rewrite: stable `index` per tool-call id, remember thought signature) and `restoreGoogleThoughtSignatures` (put the signature back on assistant `tool_calls` in the next request body); `isGoogleOpenAICompatUrl` guard.
- `lib/ai/providers.ts`: wired into the existing compatibility fetch (same seam as `wrapResponseWithReasoning`), only when `isGoogleOpenAICompatUrl(effectiveBaseUrl)`.
- `tests/ai/google-openai-compat.test.ts`: measured chunk normalisation, SSE keep-alives/`[DONE]` untouched, signature round-trip, URL guard.
- `tests/ai/openai-provider.test.ts`: `compatFetch` wiring — Google host gets the shim; a non-Google custom base URL is unchanged.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. `OPENAI_BASE_URL=https://generativelanguage.googleapis.com/v1beta/openai`, Gemini API key, `OPENAI_MODELS=gemini-3.8-flash`, `MODEL_ROUTES` with `"maic-agent-driver":{"model":"openai:gemini-3.8-flash","api":"openai-completions"}`.
2. In `/workspace` ask for a course. Before: generation fails at the first tool call with `tool_calls[0].index: expected number, received undefined`. After: the tool call validates and the next request carries `extra_content.google.thought_signature`.
3. `vitest run tests/ai` → 150 passed.

- Unit-level with the measured chunk from #1556 (missing `index`, `thought_signature` present).
- SSE keep-alive / `[DONE]` lines are left alone.
- Signature round-trip on the subsequent assistant `tool_calls`.
- URL guard: only `generativelanguage.googleapis.com`; lookalike hosts and `api.openai.com` are unchanged.
- `compatFetch` wiring on Google vs a non-Google custom base URL.
- `vitest run tests/ai` → **150 passed** (17 files).
- `eslint` on the four touched files: clean.

Not verified live against Gemini (no API key in this environment).

### Evidence

- [x] Focused vitest: `tests/ai/google-openai-compat.test.ts` + `tests/ai/openai-provider.test.ts` + `tests/ai/reasoning-sse.test.ts` → 76 passed
- [x] Full `vitest run tests/ai` → 150 passed
- [ ] Manually tested locally

## Checklist


LING-6150 (changes requested):
Thanks for addressing this compatibility issue. The missing `tool_calls[].index` problem is real, and the response normalization allows the reported tool-call shape to pass through the actual OpenAI SDK.

At reviewed head `64a8b751`, however, thought-signature restoration depends entirely on process-local state. This leaves a gap in the existing agent recovery path. There is also a typecheck failure introduced by the new test.

**1. Preserve thought signatures beyond the lifetime of the process cache**

`lib/ai/google-openai-compat.ts` stores signatures in a module-level `Map`, keyed by tool-call ID, with a limit of 1,000 entries. `restoreGoogleThoughtSignatures()` can only restore a signature while it remains in that map.

The review exercised the actual installed AI SDK rather than mocking `createOpenAI`:

1. Feed a Gemini-shaped SSE tool call containing a thought signature and no index through `getModel`.
2. Confirm that the SDK produces a tool call successfully.
3. Send the tool result in a subsequent request while the cache is populated: the signature is restored.
4. Serialize the conversation, clear the process-local signature cache, and replay the same conversation: the outgoing assistant tool call no longer contains `extra_content.google.thought_signature`.

The SDK tool-call output in this test contained `type`, `toolCallId`, `toolName`, and `input`, but no signature-bearing `providerMetadata`. The signature therefore does not travel with the conversation data through this path.

This matters because the agent runtime already restores durable history and resumes execution after interruption. A worker restart, execution in another process, or eviction from the shared cache can remove the only copy of protocol state required for the next tool-result request.

Google's documentation requires the applicable thought signatures to be returned for function calls within the current user turn:
https://ai.google.dev/gemini-api/docs/generate-content/thought-signatures#openai

The missing field on the outgoing replay request was directly reproduced. A live Gemini request was not made to independently reproduce the resulting upstream rejection.

Could you please preserve the signature with the corresponding tool-call/conversation data and restore it when that data is serialized into the next provider request? There is already a provider-metadata path in `lib/agent/runtime/provider-metadata.ts` and `lib/agent/runtime/stream-fn.ts` worth checking for this purpose.

That is an implementation suggestion, not a requirement to introduce a broader abstraction. The required behavior is that a restored conversation can produce the correct request without relying on the original process cache. Increasing the map size would not address restart or cross-process recovery.

**2. Add coverage across the SDK and durable replay boundary**

The current tests demonstrate normalization and a same-process signature round trip. Please extend the coverage to exercise:

- The measured missing-index SSE fixture through the actual SDK.
- The resulting tool call through the runtime's message conversion and serialization path.
- Restoration of that conversation with no previously populated signature cache.
- A subsequent request containing the original signature on the correct assistant tool call.
- Multiple tool calls/steps retaining their respective signatures.
- Non-Google OpenAI-compatible requests remaining unchanged.

The key regression assertion is that the next request remains correct after the conversation has been serialized and restored in a fresh execution context.

The existing 17 AI test files / 150 tests passed during review. The additional actual-SDK check passed for the immediate same-process round trip but failed for cold replay.

**3. Fix the new test's typecheck failure**

The CI failure at `tests/ai/openai-provider.test.ts:832` is caused by the newly added mock being inferred as a zero-argument function, while the test later reads its second call argument:

    fetchMock.mock.calls.at(-1)?.[1]

This produces TS2352 and TS2493. Please give the mock an appropriate fetch argument signature so the request body can be inspected with correct types.

CI log:
https://github.com/THU-MAIC/OpenMAIC/actions/runs/35460813457/job/105944262292

The index normalization is a useful repair and should be retained. Before approval, please make signature preservation work through the existing session lifecycle, add the cold-replay regression coverage, and restore passing typecheck.

dyk1454683243-sudo:

PR #1603, opened 2026-09-19
fix(audio): derive ASR and browser-TTS language defaults from the UI locale

dyk1454683243-sudo:
## Summary

Speech defaults were hardcoded to Chinese/English regardless of the interface language:

- `asrLanguage` defaulted to the literal `'zh-CN'`, and the browser-native recogniser / the settings test button fell back to `'zh-CN'` when the value was empty — a Turkish (or German, …) UI had its microphone transcribed as Chinese until the user found the language menu.
- With no usable TTS voice configured, `PlaybackEngine` set `utterance.lang` via `detectSpeechLang`, whose non-CJK fallback was always `'en-US'`, so a lesson in another language was read by an English voice.

Both now follow the UI locale. Behaviour for the default `zh-CN` locale is unchanged (`defaultASRLanguage('browser-native')` still resolves to `zh-CN`).

This adapts the unmerged sketch from #1549 onto current `main` (TTS language detection now lives in `detectSpeechLang`, including the Vietnamese path).

## Related Issues

**Fixes #1548**.

## Changes

- `lib/audio/asr-language.ts`: `defaultASRLanguage(providerId, locale = defaultLocale)` picks the entry of the provider's `supportedLanguages` matching the locale (full tag → bare language → prefix → `'auto'` → first), so `browser-native` gets `zh-CN`/`tr-TR` and Whisper-style lists get `zh`/`tr`; custom providers use `CUSTOM_ASR_DEFAULT_LANGUAGES`.
- `lib/store/settings.ts` (store default), `lib/hooks/use-audio-recorder.ts`, `lib/hooks/use-browser-asr.ts`, `components/settings/asr-settings.tsx`: the four literals become one call each. `getValidASRLanguage` (#1082) is untouched.
- `lib/playback/engine.ts` + `lib/audio/browser-tts-preview.ts`: non-CJK fallback `'en-US'` → `i18n.language || defaultLocale` (CJK / Vietnamese detection unchanged).
- `tests/audio/asr-language-default.test.ts` and a locale-fallback case in `tests/audio/detect-speech-lang.test.ts`.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] CI/CD or build changes

### Steps to reproduce / test

[...]

PR #1605, opened 2026-09-19
fix(i18n): share cookie/Accept-Language locale between server and client

### LINKED ISSUE #1445: [Bug]: Hydration mismatch on every page load — defaultLocale is hard-coded to zh-CN
### Bug Description

`lib/i18n/types.ts` exports `defaultLocale = 'zh-CN'`:

```ts
export const defaultLocale: Locale = 'zh-CN';
```

The server renders with that locale, but the client initialises i18n from the detected browser language. When the browser is not Chinese, server and client produce different text for the same node and React reports a hydration mismatch on every load.

### Steps to Reproduce

1. Set the browser's preferred language to English (or any non-Chinese locale).
2. Run a production build with `pnpm build && pnpm start`.
3. Open the classroom page.

### Expected Behavior

The server and client agree on the initial locale, so no hydration mismatch occurs.

### Actual Behavior

```
Error: Text content does not match server-rendered HTML.
Warning: An error occurred during hydration. The server HTML was replaced with client content.
  <span>嗨，同学</span>
  <span>Hi, classmate</span>
```

The Chinese string is server-rendered; the client immediately swaps it for English and discards the server HTML for that subtree.

### Deployment Method

### OpenMAIC Version

1.0.1 (`64621adf`)

### Affected Area

Deployment / infrastructure

### Browser

Chrome on Windows 11

### Operating System

Windows 11 (app served from WSL2)

### Relevant Logs

```shell
Error: Hydration failed because the server rendered HTML didn't match the client.
  at span
  at Greeting
```

### Additional Context

Reading the locale from a cookie or `Accept-Language` on the server, or deferring locale-dependent text until after mount, removes the mismatch. Changing `defaultLocale` to `'en-US'` also resolves it for English browsers, but that only moves the same mismatch onto Chinese browsers — the underlying defect is that the server cannot know the client's locale.

### LINKED ISSUE #1536: `<html lang>` is hardcoded to "en" regardless of defaultLocale
`app/layout.tsx` renders `<html lang="en" suppressHydrationWarning>` as a constant, while the interface language defaults to `defaultLocale` (`zh-CN` here, other values in forks). Screen readers announce the page as English and browsers offer to translate a Chinese/Turkish UI "from English".

Expected: the attribute follows `defaultLocale` (language subtag), the same source i18next uses.

I have a one-line fix ready and will open a PR referencing this issue.

dyk1454683243-sudo:
## Summary

The server always rendered with hard-coded `defaultLocale` (`zh-CN`) while the client `I18nProvider` re-detected from `localStorage` / `navigator.language` after (or during) hydration. A non-Chinese browser then hydrated `嗨，同学` as `Hi, classmate` / `Hi, Learner` and React discarded the server HTML for those nodes on every load (#1445). `app/layout.tsx` also hard-coded `<html lang="en">` (#1536).

This follows @wyuc's direction on #1445: one shared initial-locale decision for server and client.

1. Server reads the **locale cookie**, then `Accept-Language`, then `defaultLocale`.
2. Client `I18nProvider` starts from that **server-chosen** value (no first-paint re-detect).
3. The language switcher writes the same cookie (and still updates `localStorage`).
4. `<html lang>` uses the language subtag of that locale.

`defaultLocale` stays `zh-CN` — this does not "fix" English browsers by flipping the default. #1537 (closed, unmerged) only derived `lang` from `defaultLocale` and is an incomplete predecessor.

The root layout loads the locale JSON through a **server-safe** helper (`lib/i18n/load-resource.ts`) and does **not** import `lib/i18n/config` (that module calls `initReactI18next` / `createContext` and cannot run in the RSC graph).

## Related Issues

Fixes #1445
Fixes #1536

## Changes

- `lib/i18n/resolve-locale.ts`: shared `matchLocale` / `resolveRequestLocale` (cookie > Accept-Language > `defaultLocale`; unsupported tags map via prefix) plus `<html lang>` helper and cookie serializer.
- `lib/i18n/request-locale.ts`: server helper over `cookies()` / `headers()`.
- `lib/i18n/load-resource.ts`: locale JSON + workbench merge, importable from Server Components.
- `app/layout.tsx`: async root layout; sets `<html lang>` and passes the resolved locale and its resource bundle into `I18nProvider`.
- `lib/hooks/use-i18n.tsx`: initialise from the server locale; persist explicit `setLocale` to cookie + `localStorage`.
- `tests/i18n/resolve-locale.test.ts`, `request-locale.test.ts`, `i18n-provider-initial-locale.test.ts`: focused vitest coverage.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Prefer English (or any non-Chinese locale) in the browser.
2. `pnpm build && pnpm start` with no `locale` cookie.
3. Open the classroom / home page.
4. Switch language in the picker, reload, confirm the server render and `<html lang>` follow the cookie.

- Focused vitest: cookie > Accept-Language > default; prefix mapping; `getRequestLocale`; I18nProvider first-paint English vs Chinese.
- `pnpm exec eslint` on the touched files and `npx tsc --noEmit` — clean.
- Production `pnpm build` (exit 0) then `pnpm start`.
- `curl` against production:
  - `Accept-Language: en-US` → `<html lang="en">` + `Hi, Learner` (no `嗨，同学`)
  - `Cookie: locale=zh-CN` + English Accept-Language → `<html lang="zh">` + `嗨，同学`
  - `Accept-Language: ja` → `<html lang="ja">` + Japanese copy
- Playwright Chromium (`locale=en-US`) against that production server: no hydration / mismatch console errors on first paint; language switcher writes `locale=zh-CN`; reload keeps `<html lang="zh">` and `嗨，同学`.

## Evidence

- [x] CI checks run locally (`vitest` for the new suites, `eslint` on touched files, `npx tsc --noEmit`)
- [x] Manually tested locally (production start + Playwright)

## Checklist


LING-6150 (changes requested):
Thanks for consolidating the initial locale decision. Resolving cookie -> Accept-Language -> default on the server, passing that locale and its resources to the client, and aligning `<html lang>` is the right direction for #1445 and #1536.

At reviewed head `01eb394b`, there is a reproducible regression when switching to a language whose resources have not yet been loaded.

**1. Keep rendered translations reactive to asynchronous language loading**

In `lib/hooks/use-i18n.tsx`, `setLocale()` calls `applyInitialLocale(newLocale)` and immediately updates React state. `applyInitialLocale()` starts `i18n.changeLanguage()` without waiting for it to complete.

At the same time, the provider no longer uses `useTranslation()` or another subscription that would update React when the language change completes.

For a previously unloaded locale, this produces the following sequence:

1. The context locale changes immediately.
2. `t()` renders before the requested resource bundle is available, so it uses the Chinese fallback.
3. The requested resource bundle finishes loading.
4. The provider does not trigger another React update, so the displayed text can remain in the fallback language.

Because this provider supplies translations to shared UI consumers, the regression is not specific to the greeting used in the reproduction. Consumers can remain on fallback text after the requested resources load, until another render occurs.

This was reproduced in Chromium using the actual provider, React DOM, i18next, and the locale resource loader.

The reproduction used a fresh browser page with English and the Chinese fallback loaded, while the Japanese resource bundle was initially absent. The Japanese resource request was delayed by 300 ms to exercise the asynchronous path. After switching to Japanese, the check waited until i18next confirmed that the Japanese bundle was available, without triggering another interaction.

At that point:

    context locale: ja-JP
    document.documentElement.lang: ja
    rendered greeting: 嗨，Reviewer
    direct i18n translation: こんにちは、Reviewerさん

Switching back to English and then to Japanese displayed the correct Japanese text, because the resource was now cached.

Replacing only the provider with the base implementation in the same browser fixture correctly updated the text after the Japanese resources loaded. This identifies a regression in the new provider behavior rather than a missing translation or a resource-loading failure.

Could you please preserve reactivity when language loading completes, or coordinate loading and state updates so that the selected locale and rendered translations become consistent? Either approach is fine; the required behavior is that the first switch completes correctly without another click or an unrelated re-render.

If the repair awaits loading before committing state, please also ensure that a slower earlier switch cannot overwrite a later user selection.

**2. Add a cold-language-switch regression test**

The existing 96 i18n tests passed during review, but they do not cover this interaction. Testing only English and Chinese can miss it because Chinese is the fallback and may already be loaded.

Please add browser coverage that:

- Starts with one locale and its resources available.
- Switches to a different, previously unloaded locale, such as Japanese.
- Delays that resource load enough to exercise the asynchronous path.
- Verifies that the visible translation updates after loading finishes, without switching away and back or triggering an unrelated render.
- Verifies that the selected locale, visible text, `<html lang>`, and persisted cookie agree after the switch completes.

Please also retain verification of the original fixes: the first server/client render should use the same locale, the cookie should take precedence over Accept-Language, and a reload should preserve the selected language.

The regression above was reproduced in an isolated real-browser component fixture. A complete Next production build and full-app hydration flow were not independently rerun during this review.

**3. Coordinate final validation with #1603**

#1603 should consume the runtime UI locale established by this PR for its ASR and browser-TTS defaults. Keeping the changes in separate PRs is fine, but they should be validated together once the cold-switch regression is fixed.

Please check the combined behavior for a fresh non-Chinese UI and the existing audio fallback paths after language switching. Those fallbacks should consume the resolved UI locale, while an explicitly selected ASR language or browser voice should remain respected.

This does not require adding a new audio preference mode or migrating existing saved ASR preferences. There are separate runtime-locale wiring gaps in #1603; each PR should address its own findings before the combined verification.

The server-owned initial locale approach is worth keeping. Once cold switching updates the UI reliably, the original first-paint/cookie behavior remains covered, and the integration with #1603 is verified, this will be ready for another review.

dyk1454683243-sudo:

southerlyswell:
thankk you for looking into it. Nice to know that any issues I pick up are evaluated :)

PR #1606, opened 2026-09-19
fix(generation): inject KaTeX auto-render when the page ships core

### LINKED ISSUE #1538: Interactive scenes: KaTeX not auto-rendered when the page ships its own KaTeX; first paint raw for init()-filled labels
Two related rendering failures on generated interactive scenes, both observed on real lessons:

1. **Math stays raw when the page already ships KaTeX.** `postProcessInteractiveHtml` (`packages/@openmaic/generation/src/interactive-post-processor.ts`) skips the whole injection when `html.toLowerCase().includes('katex')`. When the model's HTML loads `katex.min.js`/`katex.min.css` itself (common), `auto-render` and the `MutationObserver` are never injected, so `\(2^{-2}\)` and `$…$` are printed literally.

2. **First paint is raw for labels filled by the page's own `init()`.** The injected script renders once on `DOMContentLoaded` and installs the observer afterwards; labels the page fills in the same tick (or right after) land before the observer exists. The math only renders after the first user interaction triggers a mutation (e.g. moving a slider).

Proposed fix (PR follows): guard on `auto-render`/`renderMathInElement` instead of `katex` and inject only the contrib script + observer when the core is already present; after installing the observer, run one more render pass after a short settle and on `window` `load`.

dyk1454683243-sudo:
## Summary

Interactive scenes left math raw (`\\(...\\)` / `$...$`) in two measured cases:

1. When the generated page already loads KaTeX core (`katex.min.js` / `katex.min.css`), `postProcessInteractiveHtml` skipped the whole injection because the guard was `html.includes('katex')`. auto-render and the MutationObserver never arrived.
2. Labels filled by the page's own `init()` on the same tick as `DOMContentLoaded` landed before the observer existed, so the first paint was raw until a later user interaction mutated the DOM.

This is a fresh PR on current `main`. Prior #1539 sketched the same approach and was closed unmerged — that fork is not reopened.

## Related Issues

**Fixes #1538**

## Changes

- Guard on `auto-render` / `renderMathInElement` instead of the bare word `katex`.
- If only KaTeX core is present, inject the missing contrib script + observer before `</body>` (auto-render needs the `katex` global). Do not duplicate core CSS/JS.
- After the observer is installed: `setTimeout(safeRender, 300)` and one more pass on `window` `load`.
- Tests in `packages/@openmaic/generation/test/support-primitives.test.ts`: core-present → contrib injected once; auto-render / `renderMathInElement` present → left alone; post-observer passes sit after the observer.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Generate an interactive scene whose HTML includes `<script src="…/katex.min.js">` and uses `\\(2^{-2}\\)` in labels filled by `init()`.
2. Before: labels stay raw on first load and only render after a control mutation. After: auto-render + observer are injected; math renders on first paint.
3. `pnpm --filter @openmaic/generation test` → 26 files / 150 tests.

- Re-read current `main`: the skip guard is still `html.includes('katex')`; no open PR claimed #1538 (prior #1539 closed unmerged).
- `pnpm --filter @openmaic/generation test` → **150 passed** (26 files), including: core-present → contrib once / core not duplicated / body placement / idempotent; auto-render present → unchanged; `renderMathInElement` present → unchanged; `setTimeout(safeRender, 300)` and `window` `load` after `observer.observe`.
- Did not run a generated interactive scene in a browser in this environment.

### Evidence

- [x] Package tests pass (`pnpm --filter @openmaic/generation test`: 150 passed)
- [ ] Manually tested locally

## Checklist


LING-6150 (changes requested):
Thanks for addressing the missing auto-render injection. I verified that the change fixes a page with synchronously loaded KaTeX core and no auto-render. The generation package’s 150 tests also pass.
Browser checks with KaTeX 0.16.9 exposed two remaining cases that need to be handled before approval:
1. A deferred core script can execute after the injected auto-render script.
For a page containing <script defer src="…/katex.min.js"></script>, placing a normal auto-render script before </body> does not ensure that core has executed. Auto-render initializes first and captures an undefined katex reference. Rendering subsequently fails with:
Cannot read properties of undefined (reading 'ParseError')
The initial render exception also prevents the observer and additional render callbacks from being installed. Please ensure that auto-render initializes after core is available, and add an execution-based regression test for a deferred core script.
2. A stylesheet alone is still treated as an available core library.
includeCore: !lower.includes('katex') returns false for a page containing only katex.min.css. The output therefore receives auto-render but no core JavaScript, and rendering fails. Please distinguish the stylesheet from the core script and cover this case as well.
The package version also needs updating. The current CI failure is the package-version gate: generation’s publishable inputs changed, but its version remains 0.3.10. Please include the required version bump.
One additional clarification: I could not reproduce the proposed init()/observer timing gap with DOMContentLoaded callbacks registered either before or after the injected callback; both cases rendered on base. Could you provide the specific failing HTML fixture for that symptom? The current tests establish that the 300 ms/load callbacks are present, but do not demonstrate that they repair the reported first-render failure.
Please retain the working synchronous-core case while adding browser coverage for these scenarios. That would give us stronger evidence for closing #1538.

dyk1454683243-sudo:

PR #1607, opened 2026-09-19
docs(docker): describe what the builder heap limit does and does not bound

wyuc:
Follow-up to #1598 addressing @LING-6150's wording note: the Dockerfile comment implied the builder's `--max-old-space-size` prevents a VM-level OOM. It only makes the V8 heap limit explicit (instead of derived from the container limit); the measurements in #1598 showed the workspace package build completing under a 1 GiB container limit, not a guarantee against host/VM OOM. Comment-only change.

https://claude.ai/code/session_01DYifP8wM4XJQ3Hc2qsF6zf


LING-6150 (approved):
Thanks for following up. The Dockerfile comment now clearly distinguishes the V8 heap limit from total memory usage and scopes the measured result to the workspace package build. I also confirmed that #1598’s description has been updated accordingly.
Verified that all non-comment Dockerfile instructions are unchanged. This addresses the remaining wording points from my review. Looks good to me.

Issue #847, opened 2026-07-04
[Bug]: Scene content generation fails with GLM(official endpoint) on long(5min+) non-streaming requests


jiangzizi:
### Bug Description

When generating a course with GLM via the OpenAI-compatible BigModel endpoint, `/api/generate/scene-content` can fail with a 500 error during long interactive/code-widget scene generation.

The failure is reproducible when the scene content step takes around 5 minutes before returning a complete non-streaming response. The provider connection is closed with `AI_APICallError: Cannot connect to API: Headers Timeout Error`, and the route returns 500.

<img width="1499" height="732" alt="Image" src="https://github.com/user-attachments/assets/37e52bfa-4a21-4fad-ac2f-6248fa6aa3ce" />

### Steps to Reproduce

1. Configure GLM as the LLM provider using the OpenAI-compatible BigModel endpoint:
   - `GLM_BASE_URL=https://open.bigmodel.cn/api/paas/v4`
   - `DEFAULT_MODEL=glm:glm-5.1` or select `glm:glm-5.1` in the UI
   - Open up thinking mode.
2. Start the app locally with `pnpm dev`.
3. Open the generation preview flow.
4. Enter a prompt that creates a course with an interactive/code-widget scene, for example:
   `我要学习 transformer 架构`
5. Continue generation until the scene-content step reaches a long generation scene such as:
   `整体架构总览`
6. Observe the server logs and the failed scene-content request.

### Expected Behavior

The scene-content generation should complete successfully, including long scenes.

If the model takes a long time to generate, the server should keep the provider connection alive or use streaming internally so the upstream gateway does not close the request before the complete response is returned.

### Actual Behavior

Connection broke after 5 min with the following log
```
[2026-07-04T06:22:28.295Z] [INFO] [Scene Content API] Generating content: "整体架构总览" (slide) [model=glm:glm-5.1]
[2026-07-04T06:27:29.305Z] [ERROR] [Scene Content API] Scene content generation failed [scene="整体架构总览", model=glm:glm-5.1]: AI_APICallError: Cannot connect to API: Headers Timeout Error
```

### Deployment Method

Local development (npm run dev / pnpm dev / yarn dev)

### Browser

Chrome149

### Operating System

MacOS 15.1

### Relevant Logs / Screenshots

```shell
POST /api/generate/scene-actions 200 in 17.6s (compile: 3ms, proxy.ts: 3ms, render: 17.6s)
[...]

PR #1614, opened 2026-09-20
feat: configurable model escalation scheduler

### LINKED ISSUE #1616: [Feature]: configurable model escalation scheduler
## Summary
Optional per-stage model escalation for course generation: when a stage's base model fails with a retryable error (timeout / empty output / network), the engine re-runs the failed scene once with an explicitly selected upgrade model, resolved outside `MODEL_ROUTES`, and records every decision to a durable ledger shown in a new Settings section ("Model Scheduling"). Disabled by default — no config file means unchanged behavior.

## Motivation
Long-running generation stalls on a single bad call (see #847 for the same 5-min timeout failure class this absorbs as a fallback). Operators previously had to edit `MODEL_ROUTES` and manually re-trigger pages.

## Proposal
- Config `data/model-schedule.json` (hot-reload, 250 ms per-call cache, no restart)
- Per-stage policy: `{ base, escalateTo, trigger: onTimeout | onRetryableError, max }`
- Daily escalation budget cap; append-only ledger `data/schedule-events.jsonl`
- Safety boundary: content-safety style rejections never escalate

## Related
- Design discussion: #1612
- Background: #847 / #848 / #1403
- Implementation PR: #1614

## Acceptance criteria
- [ ] Missing config ⇒ original pipeline unchanged
- [ ] Policy-configured stage escalates once on retryable failure and logs the decision
- [ ] Settings panel can edit config, load a 3-tier preset, and view the recent ledger

### Proposed Solution

Auto-fallback when a stage's base model fails with a retryable error:
- Per-stage escalation policy in `data/model-schedule.json` (hot-reload, 250 ms per-call cache, no restart): `{ base, escalateTo, trigger: onTimeout | onRetryableError, max }`
- The failed scene re-runs once with the upgrade model, resolved explicitly outside `MODEL_ROUTES`
- Daily escalation budget cap; append-only ledger `data/schedule-events.jsonl`
- Safety boundary: content-safety style rejections never escalate
- New Settings section "Model Scheduling": JSON editor, 3-tier preset template, recent ledger viewer

Implementation already exists and is verified locally (lint 0 errors, 12-locale i18n aligned, production build passes, engine assert script 8/8). PR: #1614.

### Acceptance Criteria

- [ ] A course generation stalls when a stage's base model times out — verify the scene completes after a single automatic escalation to the configured upgrade model
- [ ] Settings > Model Scheduling shows the config editor, loads the 3-tier preset, and displays the ledger entry for the escalation above
- [ ] Without `data/model-schedule.json`, generation behaves exactly as before (no escalation, no ledger)
- [ ] Renaming/removing the config file disables escalation without a server restart

### Alternatives Considered

- Manual workaround: edit `MODEL_ROUTES` and re-trigger the page when a stage stalls (current behavior)
- Raise HTTP client timeouts (undici headers/body timeout, cf. #1403) — complementary root-cause layer, not a substitute for a model-level fallback

### Area

Classroom generation

### Additional Context

- Design discussion: #1612
- Same 5-minute timeout failure class this absorbs: #847 (open, status:ready) / #848 (stale draft) / #1403
- Implementation PR: #1614 — includes screenshots of the settings panel and ledger

mediasambb:
What

Small, operator-configured fallback-retry for generation calls (per @wyuc's review).

When a call fails with a retryable error — timeout (408), quota (429), capacity
(503), server 5xx, network, or an empty output — the shared LLM layer retries
**once** on a fallback model. Content-safety rejections and other 4xx errors
never fall back; a dead/mis-keyed primary still surfaces its real error.

## Config (server-side env only — no UI, no API, no ledger)

- Per-stage: `MODEL_ROUTES='{"scene-content":{"model":"openai:gpt-5.4","fallback":"qwen:deepseek-v4-pro"}}'`
- Global: `MODEL_FALLBACK='qwen:deepseek-v4-pro'`
- Unset = existing behaviour unchanged.

One mechanism, one place: wired into the shared `callLLM` layer (scene content,
scene actions, agent profiles, quiz grading, web search, classroom generation,
PBL) plus the outlines route. `verify-model` opts out. Aligns with #1034 so
quota and timeout share one fallback. 13 new tests; one log line per fired
fallback.

Closes #1616


wyuc:
Thanks @mediasambb for the thorough write-up, the screenshots and the soak results — the problem is real. One bad call (timeout, empty output) stalling a whole course is the same failure class as #847, and today recovering from it means editing routes and re-triggering pages by hand.

That said, we'd like to take a much simpler route than this PR, so I don't think we should review this implementation as-is. A few reasons:

**It's more machinery than the problem needs.** The fix we're after is essentially *"if the call fails with a retryable error, retry it once on a different model."* Per-stage policy files, hot reload with a cache, a daily escalation budget, an append-only ledger and a settings editor with presets are all additional surface to maintain for what should be a small, predictable fallback.

**It changes server-wide behaviour from the browser.** Which models the server uses is operator configuration (env / `MODEL_ROUTES`), while the settings page holds per-user preferences. The new `/api/model-schedule` GET/PUT/DELETE writes a server-global file with no authentication, so anyone who can reach the site could change which (and how expensive) models the server calls. Writing into `data/` also doesn't fit read-only containers or multi-instance deployments.

**It should be one mechanism, in one place.** #1034 is asking for the same thing with a different trigger (a model out of quota). We'd rather have a single fallback in the shared model-call layer that covers both, for every generation step, instead of logic inside the scene-content route only.

What we'd welcome instead is a small PR along these lines:

- an operator-configured fallback model (e.g. an optional fallback per route, or one global fallback), set on the server like the existing routes;
- on a retryable failure (timeout, empty output, network, quota), retry **once** on the fallback model; content-safety rejections never fall back;
- a log line when it happens; no new UI, no config API, no ledger file;
- unset means behaviour is unchanged.

If you're up for it, it'd be great to align with #1034 on this so the two land as one change — happy to review that. I'll leave this PR open for now in case you'd like to rework it here, but feel free to close it in favour of a fresh, smaller one.

mediasambb:
Thanks @wyuc for the detailed steer — agreed on all three points.
We'll rework this PR into the small fallback-retry you described:
an optional per-route fallback in MODEL_ROUTES plus a global MODEL_FALLBACK, both set server-side via env (no new API, no UI, no ledger);
one retry on the fallback model when the call fails with a retryable error (timeout, empty output, network, quota/429, capacity/503); content-safety rejections never fall back;
one log line when it happens; unset means behaviour is unchanged.
One mechanism, one place: it lands in the shared callLLM layer (covers scene content, scene actions, agent profiles, quiz grading, web search) plus the outlines route via its existing attempt loop, so every generation step is covered. We'll also align with #1034 so the quota and timeout triggers share the same fallback mechanism.
Slimming the branch down now — will update the PR with the reduced diff and refreshed CI shortly.
感谢 @wyuc 的详细指导，以上三点很细致地指出不同的修复路径，个人表示十分赞同。我会把这个 PR 重构成您描述的小型兜底重试：
1) 在 MODEL_ROUTES 里加可选的每路由 fallback + 全局 MODEL_FALLBACK，均通过服务端环境变量配置（无新 API、无 UI、无账本）；
2) 遇到可重试错误（超时/空输出/网络/429 配额/503 容量）时在 fallback 模型上重试一次；内容安全类拒绝永不兜底；
3) 发生时打一条日志；未配置则行为不变。机制放在共享 callLLM 层（覆盖场景内容/场景动作/智能体画像/测验评分/网页搜索）+ outlines 路由的现有重试循环，覆盖所有生成步骤；并与 #1034 对齐共享同一兜底机制。正在精简分支，稍后更新 PR 与 CI。

wyuc (changes requested):
Thanks @mediasambb, this is much closer to what we had in mind: server-side config only, one retry, content-safety never falls back, and nothing changes when unset. The per-stage override makes sense too, since a vision-dependent stage like outlines can't fall back to a text-only global model.

Two things to tighten before we merge, plus one small one:

1. **Put the fallback logic in one place.** `callLLM` has one implementation, and `scene-outlines-stream/route.ts` has a second hand-rolled one inside its retry loop (`maybeFallback` + resetting `attempt = 0` + the extra `!(fellBack && attempt > 1)` loop guard). That loop control is easy to break the next time someone touches outline retries. Please extract one shared "retry once on the fallback model" helper that both the non-streaming path and the outlines stream use, so the retryable-error decision, the fallback resolution and the log line live in a single spot.

2. **Only fall back on genuinely empty output, not on any validation failure.** In `callLLM` Phase 1, any `validate` failure sets `triggerFallback`. That includes caller-supplied validators (format or JSON checks), not just empty output, which quietly turns this into "retry on another model when the output quality is off" and spends the fallback model's quota on it. Please restrict the validation-triggered path to empty or whitespace-only text, which matches the PR description and the outlines route, and add a test that a non-empty result failing a custom validator does **not** fall back.

3. **Nit:** please retitle the PR (e.g. `feat(llm): retry once on a fallback model for retryable failures`). "Escalation scheduler" no longer describes it.

Once these are in, we'll do a final pass and merge. Since this covers the quota case too, we'll close #1034 in favour of this PR.

mediasambb:
Got the suggestions, thanks for the guidance. I'll get started making the adjustments and optimizations now 😊





------------------ Original ------------------
From: wyuc ***@***.***&gt;
Date: Wed,Sep 23,2026 11:52 AM
To: THU-MAIC/OpenMAIC ***@***.***&gt;
Cc: 小魔 ***@***.***&gt;, Mention ***@***.***&gt;
Subject: Re: [THU-MAIC/OpenMAIC] feat: configurable model escalation scheduler (PR #1614)





 
@wyuc requested changes on this pull request.
 
Thanks @mediasambb, this is much closer to what we had in mind: server-side config only, one retry, content-safety never falls back, and nothing changes when unset. The per-stage override makes sense too, since a vision-dependent stage like outlines can't fall back to a text-only global model.
 
Two things to tighten before we merge, plus one small one:
  
 
Put the fallback logic in one place. callLLM has one implementation, and scene-outlines-stream/route.ts has a second hand-rolled one inside its retry loop (maybeFallback + resetting attempt = 0 + the extra !(fellBack &amp;&amp; attempt &gt; 1) loop guard). That loop control is easy to break the next time someone touches outline retries. Please extract one shared "retry once on the fallback model" helper that both the non-streaming path and the outlines stream use, so the retryable-error decision, the fallback resolution and the log line live in a single spot.
 
 
 
Only fall back on genuinely empty output, not on any validation failure. In callLLM Phase 1, any validate failure sets triggerFallback. That includes caller-supplied validators (format or JSON checks), not just empty output, which quietly turns this into "retry on another model when the output quality is off" and spends the fallback model's quota on it. Please restrict the validation-triggered path to empty or whitespace-only text, which matches the PR description and the outlines route, and add a test that a non-empty result failing a custom validator does not fall back.
 
 
 
Nit: please retitle the PR (e.g. feat(llm): retry once on a fallback model for retryable failures). "Escalation scheduler" no longer describes it.
 
  
Once these are in, we'll do a final pass and merge. Since this covers the quota case too, we'll close #1034 in favour of this PR.
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today!

mediasambb:
Hi @wyuc — all three points addressed:

Shared fallback helper: the retryable-error decision (shouldFallbackFor), fallback resolution and the log line (logFallbackFired) now live in lib/server/llm-fallback.ts, used by both the callLLM non-streaming path and the outlines stream; the hand-rolled maybeFallback loop-guard logic is gone.
Only genuinely empty/whitespace-only output triggers the fallback on the validation path — a non-empty result failing a custom validator does not fall back. Added tests for both (50 tests total, incl. the non-empty-validator case).
PR retitled to feat(llm): retry once on a fallback model for retryable failures. Note: CI did not auto-trigger on the latest push (fork workflow approval). Could you approve/trigger it when you get a chance? Thanks!
image

wyuc (changes requested):
Thanks @mediasambb, both round-2 points are addressed: only empty or whitespace output triggers the validation-path fallback, with a test that non-empty validator failures don't fall back, and the decision, resolution and log line now live in `lib/server/llm-fallback.ts`. 401 and content-safety errors still surface unchanged, and the fallback model comes only from server config. The retitle helps too. The remaining hand-rolled loop control in the outlines route is fine to keep.

One blocker, since it means the feature rarely fires in practice:

**P1: `isRetryableLlmError` doesn't recognize the errors the AI SDK actually throws.** `generateText` / `streamText` use `maxRetries = 2` by default. After the SDK's own retries, a 429/503/5xx surfaces as an `AI_RetryError`: no `statusCode`, the real `APICallError` is in `.lastError` / `.errors[]`, and the message reads `Failed after 3 attempts. Last error: ...`. None of the current checks match that, so for every call site on default retries (agent-profiles, quiz-grade, web-search, classroom-generation, PBL planner/simulator, the Claude web-search adapter, ai-sdk-adapter) quota and capacity errors **never** fall back. We reproduced this with the real `ai` package: 429/503/500 → `AI_RetryError` → `isRetryableLlmError === false`. Network failures don't match either. On call sites with `maxRetries: 0`, the SDK throws `APICallError` with no `statusCode` and a message like `Cannot connect to API: connect ECONNRESET`, which only passes when the text happens to contain "TIMEOUT".

Suggested fix: unwrap `RetryError` (`lastError`, then `errors`) and walk `cause`; prefer the SDK's own `APICallError.isRetryable` flag, falling back to status codes (408/409/429/≥500); treat `Cannot connect to API` network errors as retryable. Please rebuild the tests from real `APICallError` and `RetryError` instances (from `@ai-sdk/provider` / `ai`) instead of plain `Error`s, so this can't regress unnoticed. Also sort out the `instanceof TypeError` catch-all, which can treat programming errors as retryable.

Please also fix (P2):
- **Outlines route, `maybeFallback`:** `resolveFallbackModel` isn't wrapped the way `resolveFallbackModelSafe` is in `callLLM`. If the fallback provider is misconfigured, the client ends up seeing `API key required for provider: …` instead of the real "empty response" error. Return `false` on resolution failure and keep `lastError`.
- **Empty-output fallback in `callLLM` is unreachable in production:** no production caller passes `retryOptions`, so `validate` is `undefined` and an empty result is returned as success. Either detect empty output by default when a fallback is configured, or state in the description and `.env.example` that only the outlines route falls back on empty output.

**Also needed: a merge conflict with `main`** in `tests/server/model-routes.test.ts` (from the settings refactor in #1644). Both sides only add cases, so keep both. CI can't run until it's resolved.

Nits: `fallback` is now parsed inside user routes from `x-model-routes`. It's inert because the fallback is read only from operator `MODEL_ROUTES`, but please either exclude it from user routes or add a test pinning "fallback is server-only". A short `.env.example` note that per-stage fallback only applies to stages in `LLM_STAGES` would also help.

mediasambb:
Thanks @wyuc — All round-2 items are now pushed (head dd4684e):

P1 — isRetryableLlmError now classifies real AI SDK errors. unwrapRetryChain unwraps AI_RetryError (preferring lastError, falling back to the last entry of errors) and walks the cause chain with cycle protection. For APICallError we trust the isRetryable flag first, then fall back to status codes 408/409/429/≥500; transport-level failures ("Cannot connect to API", ECONN*, UND_ERR, socket hang up, timeout) are treated as retryable. The blanket instanceof TypeError catch-all is removed, so programming errors like TypeError('x is not a function') no longer trigger a fallback. Tests now construct real APICallError/RetryError instances imported from ai instead of hand-rolled plain Errors.

P2 — outlines route no longer leaks resolution errors. maybeFallback wraps resolveFallbackModel in try/catch and returns false on failure, so the real error (e.g. a provider auth problem) is preserved for the client instead of being masked by an "API key required for provider" fallback error. .env.example also documents that on the shared callLLM layer the fallback fires on retryable errors only, and the empty-output fallback applies to the outlines stream route.

Conflict — model-routes files merged with #1644. lib/server/model-routes.ts and tests/server/model-routes.test.ts now carry both your user-stage additions (parseUserStageRoutes/getUserStageRoute, the three user-route test cases) and the per-stage fallback parsing (our three fallback cases plus a server-only pin), so the conflict is resolved.

Nits — addressed. parseUserStageRoutes explicitly deletes fallback from user-supplied routes (operator-only knob), pinned by a test that feeds fallback through the x-model-routes header and asserts it is dropped; .env.example notes that per-stage fallback only applies to stages listed in LLM_STAGES.

Local verification: tsc --noEmit clean; vitest on the four touched suites → 55/55 passed; prettier + eslint clean on all touched files.

Could you approve the workflow run / trigger CI for the new head when you get a chance? Fork pushes don't start checks until a maintainer approves them, and we'd like to confirm everything is green before your final pass.

PR #1617, opened 2026-09-20
feat: add ask-user step before non-Pro generation

### LINKED ISSUE #1505: [Feature]: Model-initiated clarification when uncertain (ask-user) (non-pro mode)
### Problem or Motivation

OpenMAIC currently has no model-initiated clarification mechanism. When the underlying LLM is uncertain about key parameters—such as audience level, duration, teaching style, language mix, prerequisites, or the scope of an uploaded PDF—it tends to guess and continue generating instead of pausing to ask the user.

### Proposed Solution

Add a lightweight ask-user / clarification capability that the model can call when confidence is low. Concretely, the model would be able to emit a structured action such as an `ask_user` payload containing one or more questions, optional choices, and optional free-text input. When that action is emitted, generation or orchestration should pause, the UI should surface the questions clearly, and the user’s answers should be injected back into context before the pipeline resumes.

### Acceptance Criteria

- [ ] During outline generation, the model can emit a structured `ask_user` action when key parameters are ambiguous
- [ ] Generation or orchestration pauses until the user answers or explicitly skips
- [ ] The UI presents the questions (options plus optional free text) in a clear panel
- [ ] User answers are injected into subsequent generation context and affect the resulting outline or scenes
- [ ] When the requirement is already clear, the model does not over-ask and can proceed without clarification
- [ ] Behavior is documented (when the model should ask versus assume)

### Alternatives Considered

_No response_

### Area

Classroom generation

### Additional Context

For reference, OpenCode implements a similar mid-task `question` tool: the model can pause execution, present structured questions with options (and optional free-text), wait for the user, then continue with the answers in context. Relevant source and docs:

- Repo: https://github.com/anomalyco/opencode
- Tool implementation: https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/tool/question.ts
- Tool prompt / usage guidance: https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/tool/question.txt
- Question service: https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/question/index.ts
- Docs (Tools → question): https://github.com/anomalyco/opencode/blob/dev/packages/web/src/content/docs/tools.mdx

composable-tu:
## Summary

Pre-outline clarification for non-Pro generation, closing #1505. When the request is ambiguous, the model asks a few structured questions (options plus optional free text) before drafting the outline. The pipeline pauses for answers — or an explicit skip — and the answers go into outline generation as authoritative requirements. Clear requests skip this entirely, and outline generation itself never asks.

## Related Issues

Closes #1505

## Changes

The package owns the whole clarification contract (prompt, validation, QA assembly, with unparseable output treated as "no clarification needed"). A new POST /api/generate/clarify route runs the preflight and fails open the same way. In /generation-preview clarification is a regular pipeline step between web-search and outline, rendered inside the same card: timeline motif, question list, footer matching the outline editor (Back to requirements, Skip, Confirm and generate outline).

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Generate with something vague like "teach photosynthesis" and you'll land on the confirming step;
2. pick an option, submit, and the outline request carries the Q/A through. 

Skip goes straight to outlines with nothing injected.

截屏2026-09-20 13 50 06
截屏2026-09-20 13 51 02

### Evidence

- [x] Manually tested locally

## Checklist

- [ ] My code follows the project's coding style
- [ ] I have performed a self-review of my code


composable-tu:
the PR might not fully meet your expectations; if that is the case, please let me know. thank you very much.

wyuc:
Thanks @composable-tu — the failure you describe is real (asking for a 50-slide deck about a document that was never attached, and the model guessing anyway), and the implementation is careful: fail-open on unparseable output, clear requests skipped, outline generation left unchanged. I'm still going to close this, because it doesn't fit where the two modes are meant to sit.

**Classic (non-Pro) generation is deliberately one input, one course.** Its whole value is that you type a sentence and get a course back. A clarification step puts a gate in front of exactly that promise, and because the trigger is the model's judgement, users can't predict when they'll be interrupted — which makes the lightweight path feel unreliable rather than quick.

**Pro mode already owns this.** Pro is a conversation: the agent can and does ask before it builds. Adding an ask-user step to the classic pipeline means maintaining a second, weaker version of a capability that exists natively one mode over. When a classic request really is too vague, the honest answers are "generate anyway and let the user regenerate" (today's behaviour, and cheap) or "this belongs in Pro".

So we'd rather keep the boundary sharp: **clarification lives in Pro; classic stays a single shot.** I'm closing #1505 as well on the same reasoning.

Sorry to close a PR this complete — it's well put together, and the reasoning is about product shape rather than the code. If you'd like to build on Pro's clarification path instead, that's a direction we'd happily look at.

composable-tu:
I understand that the non-Pro version of OpenMAIC is guided by a “single input, single output” philosophy, and I respect that design choice.

At the same time, I wonder whether we might prioritize the final quality of the courseware a little more flexibly. Would it be possible to offer the “ask-user” clarification step as an optional advanced setting—disabled by default—under the non-Pro dialog? That way, users who prefer the current one-shot experience keep it unchanged, while those who want higher-quality results on ambiguous prompts still have a path to improve them.

I raise this only because I care about the non-Pro experience remaining competitive and useful for the majority of users. Happy to discuss further if this direction doesn’t feel right.

composable-tu:
I believe the product philosophy should ultimately serve the quality of the courseware.

PR #1618, opened 2026-09-20
fix(generation): stabilize model picker teardown

### LINKED ISSUE #1447: [Bug]: removeChild NotFoundError when picking a model from the toolbar model picker
### Bug Description

The generation toolbar's model picker is a `Popover` (`modal` defaults to `true`) whose content contains its own popups. Because the popover holds a modal focus trap, opening a nested popup races that trap's teardown; React then tries to remove a DOM node that is already gone.

### Steps to Reproduce

1. Open a classroom.
2. In the generation toolbar, select the model picker (the button showing the current provider and model, around `components/generation/generation-toolbar.tsx:854`).
3. Choose a different provider, then choose a model from the list.

### Expected Behavior

The model is selected, the popover closes, and the toolbar shows the new model.

### Actual Behavior

The selection intermittently throws and the UI can be left in a broken state:

```
NotFoundError: Failed to execute 'removeChild' on 'Node':
  The node to be removed is not a child of this node.
```

### Deployment Method

### OpenMAIC Version

1.0.1 (`64621adf`)

### Affected Area

Classroom generation

### Browser

Chrome on Windows 11

### Operating System

Windows 11 (app served from WSL2)

### Additional Context

Versions: Next.js 16.2.11, React 19.2.3, `@radix-ui/react-popover@1.1.15`, `@radix-ui/react-select@2.2.6`.

Setting `modal={false}` on this `Popover` resolves it locally, which is consistent with the focus-trap race described above. It may be worth checking the other Popovers in the same file, which use the same default.

LeoParkerOu:
## Summary

Make the generation toolbar model picker non-modal and close it before applying a model change, avoiding the focus-trap teardown race with nested controls.

## Related Issues

Fixes #1447

## Changes

- Configure the toolbar model-picker `Popover` with `modal={false}`.
- Close the model picker before updating the selected provider and model.
- Add browser regression coverage for click, Enter, and Space selection.
- Cover nested thinking controls, outside-click dismissal, Escape dismissal, and trigger focus restoration.
- Repeat the affected interactions to exercise the intermittent teardown path.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Open the generation toolbar model picker.
2. Select models using click, Enter, and Space, including after switching providers and opening the nested thinking selector.
3. Verify the picker closes after selection, outside-click dismissal works, Escape restores focus to the trigger, and no page error is raised.

- Prettier, ESLint, and `tsc --noEmit` pass for the changed code.
- The dedicated Playwright suite passes: 5 tests, including ten-cycle stress coverage of the affected interactions.
- Click, keyboard selection, nested popup use, outside-click dismissal, and Escape focus restoration were tested locally.
- The original intermittent Windows 11/WSL2 exception was not reproduced locally; the regression suite exercises the reported focus and teardown paths.

### Evidence

- [x] Manually tested locally

## Checklist


wyuc (approved):
Approving — thanks @LeoParkerOu, and for picking this up so quickly. Two-line production change with the right mechanism: no modal focus trap to tear down, and the picker closes before the provider/model update so the nested control isn't unmounting mid-selection. The e2e coverage is what makes `modal={false}` safe here — outside-click dismissal, Escape, and focus restoration to the trigger are exactly the behaviours that change when the trap goes away, and repeating the interaction exercises the intermittent path.

Issue #1468, opened 2026-09-11
Extract one per-course session hook shared by the classroom page and the workspace pane


wyuc:
Follow-up from the review of #1392 (raised by @cosarah).

**Problem.** #1392 adds per-course session bookkeeping to both classroom surfaces — `app/classroom/[id]/page.tsx` and `components/classroom/ClassroomSurface.tsx` — as parallel copies: the ownership sidecar fetch and `noteStageGenerationOwnership`, teardown on course change (aborting passes, clearing narration allocations, forgetting parked media allocations), and the resume gate. On the page the gate is re-composed inline as `mayStartOwnerGeneration(isServerBackedMediaPersistence(), ownership)` rather than read through `useMayGenerateForStage`.

Two ways this bites later: if the gate gains another input, the inline copy silently diverges and the visitor-must-never-spend invariant can reopen on one surface only; and the next per-course registry added to one file and forgotten in the other leaks the departing course's state.

**Proposal.** Extract one hook owning the per-course session (the pattern `useNarrationAdoption` already established): sidecar fetch + ownership recording + re-ask after transient failure, teardown of every per-course registry, and the resume gate read via `useMayGenerateForStage`. Both surfaces call it; the page's inline gate goes away.

**Scope note.** Pure refactor of two surfaces; deliberately kept out of #1392.

**Related.** The per-element document reload inside `mutateDocument` (one full-document GET per committed element, raised in the same review) is a separate trade-off — the reload under the per-stage lock is what makes each element's write-back independent, so a pass that dies half-way keeps what it committed. A "reuse the loaded document within a pass while keeping per-element commits" variant is worth exploring alongside this, since both touch the same per-course session code.

LeoParkerOu:
I’d like to work on this issue. I’ll first audit the duplicated per-course session logic in the classroom page and workspace pane, then extract a shared hook while preserving the existing ownership gating, retry behavior, cancellation, and per-course cleanup semantics. I’ll add focused regression coverage for both surfaces before preparing a PR.

PR #1619, opened 2026-09-20
refactor(classroom): extract per-course session lifecycle

LeoParkerOu:
## Summary

Extract the per-course classroom session lifecycle into a dedicated hook while preserving the behavior of the shared classroom surface used by both the standalone page and workspace pane.

Since #1485 consolidated both entry points onto `ClassroomSurface`, this applies the requested lifecycle extraction to the current shared-surface architecture rather than recreating separate callers.

## Related Issues

Related to #1468

## Changes

- Add `useClassroomSession` to own the ownership sidecar refresh and generation permission gate.
- Move per-course media, narration, whiteboard, and canvas cleanup into the session hook.
- Stop active generation when the course session is replaced or unmounted.
- Keep page-specific viewer ownership handling while preserving workspace-pane behavior.
- Update focused regression coverage to verify that both entry points use the shared session-owning surface.

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Open a classroom through the standalone classroom route and through the workspace pane.
2. Switch or leave a course while generation or course-scoped state is active.
3. Verify ownership gating, transient sidecar retry, generation cancellation, and per-course cleanup remain unchanged.

- Prettier, ESLint, and `tsc --noEmit` pass for the changed code.
- All classroom tests pass locally: 10 test files, 163 tests.
- The browser classroom load and scene-switching flow passes locally.
- The standalone route and workspace pane both mount the same `ClassroomSurface`, which owns the extracted session hook.
- The separate `mutateDocument` reload trade-off described in #1468 remains outside this refactor.

### Evidence

- [x] Manually tested locally

## Checklist


wyuc (approved):
Approving — thanks @LeoParkerOu. Checked equivalence path by path rather than just reading the diff: first entry, course switch, unmount, standalone page vs workspace pane, non-owner, and both media-persistence modes all behave as before, and every teardown item in the old code has a counterpart in the hook.

Notes, none blocking:

- **Four teardown items have no test coverage** (`revokeObjectUrls`, clearing media tasks, whiteboard history, canvas reset) — deleting all four leaves `tests/classroom` fully green. That gap predates this PR, and you actually *added* assertions for narration allocations and `stopGeneration`, so coverage is net up. A real render harness would close it properly; `linkedom` + `react-dom/client` is already available in the repo and works for this.
- **The session effect's deps drop `variant`.** Today `variant` is a literal so there's no difference, but the old code would re-run teardown and ownership reset if it ever became dynamic. Adding it to the deps would keep that explicit.
- **Two static assertions were loosened** in `owner-generation-gate.test.ts` (the browser-only slice now runs to EOF, and the `loadStart > definition` ordering assertion is gone). They still pin the target strings, just with a weaker boundary.
- The resume effect and `generationStartedRef` reset stay in `ClassroomSurface`, so the lifecycle isn't fully single-owner yet. Fine as an incremental step — I'm leaving #1468 open to track the remainder.

PR #1621, opened 2026-09-20
refactor(export): extract shared speech-text walk into lib/export/narration.ts

### LINKED ISSUE #1142: collectSceneScripts re-derives the speech-text walk used across the export family
## Problem
lib/export/use-export-script.ts:108-114 re-implements the speech-text walk (iterate actions, keep type === 'speech', collect text, join) that already exists as buildSpeakerNotes in lib/export/use-export-pptx.ts:361 and in use-download-subtitles.ts. Three copies of the same domain traversal can drift on what counts as narration.

## Suggested fix
Extract a shared narration-walk helper (e.g. lib/export/narration.ts) consumed by use-export-script, buildSpeakerNotes, and use-download-subtitles. If the existing helpers genuinely cannot share, leave a cross-reference comment at each site naming the canonical walk.

## Evidence
lib/export/use-export-script.ts:113-114 -- if (action.type === 'speech') { parts.push(action.text); } -- same speech-filter walk as buildSpeakerNotes at use-export-pptx.ts:361.

## Metadata
Severity: P2 | Confidence: 75 | Reviewer(s): maintainability | Source: branch feat/download-narration-script @ 77a16c8 (code review of the issue-413 narration script PR). Note: deferred from the PR as a larger cross-file refactor.

zdnemz:
Closes #1142

## What
Adds `lib/export/narration.ts` exporting a single `collectSpeechText(scene, options)` walk. `collectSceneScripts` (script .md/.docx export) and `buildSpeakerNotes` (PPTX speaker notes) now both delegate to it, so the two cannot drift on what counts as narration.

## Why
`use-export-script.ts` re-implemented the speech-text traversal that already existed in `use-export-pptx.ts`, and the two had already drifted — see #1142.

## Correction to the issue's premise
The issue names `use-download-subtitles.ts` as a third walk site. It isn't: it calls `compileSubtitles` from `./build-export-zip`, which has no speech walk either. There were only ever **two** genuine narration walks in `lib/export` (the two above). The `type === 'speech'` checks in `classroom-zip-utils.ts` are manifest transformation, not narration collection, and were left alone.

## The drift is bigger than the issue describes
The issue describes the difference as a trim *filter*. There are actually **two** independent knobs, and the second one changes output even for well-formed scenes:

| | whitespace-only speech | per-part trimming |
|---|---|---|
| `collectSceneScripts` (before) | dropped | **`text.trim()` pushed** |
| `buildSpeakerNotes` (before) | kept | raw text pushed |

The script exporter didn't just filter whitespace-only speech — it pushed the **trimmed** text. So for `text: '  Hello  '` the script export emitted `Hello` while the PPTX notes emitted `  Hello  `. Collapsing this to a single `keepWhitespaceOnly` flag (as the issue and the original plan suggested) would have silently changed the script exporter's output and broken the existing assertion in `tests/export/narration-script-export.test.ts:92` (`'omits whitespace-only speech text and trims kept text'`).

Both knobs are therefore preserved explicitly as `keepWhitespaceOnly` and `trim`, defaulting to the PPTX behaviour:

```ts
collectSpeechText(scene, { keepWhitespaceOnly: false, trim: true }) // script export
collectSpeechText(scene)                                           // pptx notes
```

`collectSceneScripts` keeps its name, signature, `SceneScript` shape and its trim semantics; `buildSpeakerNotes` keeps its name and output. No callers change.

## Behavior change
None for output. Two deliberate robustness changes:
- `collectSpeechText` tolerates a `speech` action with a missing `text` field (the old script walk did `action.text.trim()` and **threw** `TypeError: Cannot read properties of undefined` — verified against `main`). The helper now reads `action.text ?? ''`.
- The parameter is structural/readonly, so `readonly Scene[]` is accepted.

## Verification
- [x] `npx prettier --check` clean, `npx eslint --fix` clean (no changes), `npx tsc --noEmit` clean (needs `NODE_OPTIONS=--max-old-space-size=8192` — default heap OOMs on this repo)
- [x] `npx vitest run tests/export` — 16 files, 230 tests passed (was 15/217; +13 in the new `tests/export/narration.test.ts`)
- [x] New unit tests cover: no actions, undefined actions, null/undefined scene, mixed action types, order, both `keepWhitespaceOnly` values (whitespace-only and empty string), both `trim` values, per-part independent trimming, internal newlines preserved, single-newline join
- [x] **Equivalence harness** — ran the two pre-refactor implementations (copied verbatim from `main`) against the new helper over 11 action-shape cases × both call-site option sets: 21/22 byte-identical. The single non-match is the `missing text field` case where the *old* code crashes and the new one returns a string — the intended robustness fix above, not a regression.
- [ ] Manual: export .md script → narration in order, no empty paragraphs (pending — headless environment)
- [ ] Manual: export .pptx → speaker notes match pre-refactor output (pending, same reason)

## Notes
- Refactor only; no Dockerfile/compose/package changes. Filed by maintainer @SomSamantray with maintainability reviewer metadata.
- AI-assisted PR. Self-reviewed before requesting human review.


LING-6150:
Thanks for the patch and for preserving the existing whitespace behavior of both exporters. This remains primarily a refactor, and our CONTRIBUTING guidelines do not accept refactor-only PRs unless explicitly requested by a maintainer. I couldn’t find such a request in #1142’s discussion.
I’m going to close this under the current contribution policy. cc @wyuc for visibility, please let me know if you’d like to make an exception and continue the review.

wyuc (approved):
Thanks @zdnemz, a careful refactor, and good catch that the two walks differed on two independent knobs rather than one. Verified independently:

- Running the pre-refactor `collectSceneScripts` and `buildSpeakerNotes` (copied from `main`) against the new call sites over 20 action shapes plus a multi-scene case: PPTX notes are byte-identical in every case, and the script export is byte-identical except the missing-`text` cases, where the old walk threw a `TypeError` and the new one returns the remaining parts (or `''`). That's the intended robustness change.
- Dropping `trim` from the script call site, flipping the `keepWhitespaceOnly` default, or changing the join separator each turns tests red; `tests/export`, lint and typecheck pass.

Approving. Small, non-blocking notes:

1. `lib/export/use-export-pptx.ts:364-374`: the old JSDoc ("…and action labels") is still stacked above the new one, so it's orphaned, and the "action labels" claim was never true. It's worth deleting.
2. `lib/video-export/passes/timeline.ts:183-215` has a third, genuinely different speech-text policy (drops whitespace-only but keeps untrimmed text). Leaving it out of scope is right, but it means #1142's single source of truth isn't complete across the whole export family yet; a follow-up could route it through the helper with its own options.
3. Wording only: a missing `text` yields the remaining joined parts when others exist, not always `''`.

I'll update the branch and merge once CI is green.

PR #1624, opened 2026-09-21
fix(classroom): centralize generation resume gate

### LINKED ISSUE #1468: Extract one per-course session hook shared by the classroom page and the workspace pane

LeoParkerOu:
## Summary

Complete the remaining per-course lifecycle ownership work for #1468 by moving classroom generation resume gating into the shared session hook.

## Related Issues

Closes #1468

Follow-up to #1619

## Changes

- Move the owner-only resume gate and one-shot generation latch into `useClassroomSession`.
- Keep the standalone classroom page and workspace classroom pane on the same load, ownership, cleanup, and resume lifecycle.
- Preserve server-job ownership, pending-outline recovery, completed-scene media recovery, and the existing retry affordance.
- Add focused coverage proving the shared hook owns the gate and the classroom surface does not duplicate it.

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactoring
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Mount the standalone classroom page or workspace classroom pane.
2. Load a course with pending generation work.
3. Verify ownership is resolved before resume starts and generation starts at most once.
4. Verify server-job courses remain browser-inert and completed scenes only resume missing media.

- Classroom gate and policy tests passed: 37 tests across 2 files.
- Prettier and ESLint passed for all changed files.
- `git diff --check` passed.
- Full classroom-suite execution was limited locally because workspace packages were not built in this checkout.

## Checklist


wyuc (changes requested):
Thanks @LeoParkerOu — this is the right shape for closing #1468, and most of it checks out independently: the moved resume body is line-for-line identical to the old inline block, the owner gate lives in one place for both the page and the pane and still fails closed (a refused gate neither resumes nor latches, and the effect re-runs when the ownership answer lands), and the full `tests/classroom` suite (163 tests), lint, typecheck and `next build` all pass.

Two things before merge:

### 1. The one-shot latch now fires even when there was nothing to resume

Before, `generationStartedRef` was set only inside the branches that found work: server-job, pending outlines, and the media-only resume. The "nothing to do" case (no pending outlines and no outlines, or no stage) left it unlatched. Because the old effect effectively re-ran on every render (`generateRemaining` changes identity each render, and the surface subscribes to the whole stage store), a pass that found nothing could be followed by one that did.

Now `use-classroom-session.ts` sets the latch whenever the gate passes and then calls `resumeGeneration()`, whether or not it started anything. A pass over an empty or outdated store therefore suppresses every later resume for that course.

Concrete case: switching courses in a mounted surface (neither call site keys `<ClassroomSurface>` by course). On the transition render the reset effect hasn't run yet, `loading`/`error` and the store still describe the outgoing course, and `resumeGeneration()` runs against it. If the outgoing course has no outlines (e.g. one loaded through the server fallback, which sets stage and scenes but not outlines), the no-op latches, and the incoming course's pending work is never auto-resumed until a reload. In server-backed mode an A→B→A switch with a stale `owner` answer can do the same.

It's a missed start rather than a double spend, but it's a regression from the current behaviour. Suggested fix: have `resumeGeneration` return whether it actually started or resumed work (or extract a small decision function), and latch only when it did. That preserves the old "latch only when work was found" rule.

### 2. The tests don't observe the semantics this PR moves

The new assertions in `tests/classroom/owner-generation-gate.test.ts` only check that certain substrings exist in the hook and not in the surface. As a result, each of these mutations leaves all 163 classroom tests green:

- deleting the latch assignment in the hook;
- deleting the latch reset on course change;
- removing `mayGenerate` from the gate input (only `tsc` catches this one).

Please add behavioural coverage, either through a small render harness for `useClassroomSession` or a decision-level unit test, that pins: resume runs only when `mayGenerate` is true; it starts at most once per course; the latch resets on course change; and (after fix 1) a no-op pass doesn't block a later resume. The previous wiring test's dependency-array assertion was also dropped; it's worth keeping an equivalent check.

Happy to re-review once these are in.

PR #1626, opened 2026-09-21
fix(export): include active line geometry in bounds

### LINKED ISSUE #674: getElementRange clips line elements with curve/elbow control points

YizukiAme:
## Summary

Curved and elbow lines can have selection/export bounds smaller than their visible path because `getElementRange` only considers the endpoint maxima. Include endpoint minima and the active path's control hull, while keeping double-elbow routing unchanged and excluding its unused control coordinate.

## Related Issues

Fixes #674. Builds on the approach and review in the closed, unmerged #675, adapted to the current package layout.

## Changes

- Share the renderer's corrected geometry with the app and React editor; preserve exact Bezier visual bounds for alignment guides.
- Translate PPTX endpoints/control points relative to the bounds minimum so the export fits its shape box without moving the line on the slide.
- Add geometry and real PPTX-output regressions; bump renderer/editor patch versions.

## Type of Change

### Steps to reproduce / test

1. Use a line with `start=[0,0]`, `end=[40,0]`, `curve=[20,60]`: the previous range has zero height; the corrected control hull includes the curve.
2. Check horizontal/vertical double elbows, including `end=[100,20]`, `broken2=[50,500]`: the unused `500` must not inflate the bounds or change the route.
3. Export lines with negative control points and nonzero endpoint offsets; unzip the PPTX and check shape origin, path dimensions, and local coordinates.

- Regression tests fail on the original implementation and pass with the fix.
- After rebasing onto current `main`: root suite **8,522 passed, 43 skipped**.
- Renderer suite: 96 passed. Editor suite: 592 passed on rerun. Its first run reported two asynchronous `window is not defined` errors from `EditableSlideCanvas.operate.test.tsx` teardown; rerunning without source changes passed.
- Production Next.js build, renderer/editor builds, i18n, formatting, ESLint, TypeScript, and package-version checks passed. ESLint reports 18 warnings in unchanged files and zero errors.
- Five fixtures rendered through the actual `BaseLineElement` in Chromium; browser-computed path bounds fit the corrected geometry.
- Four real PPTX exports checked, including a cubic whose shape origin becomes `(90,70)` from `(100,100)`, with all path/control points translated to preserve their absolute positions.
- Native PowerPoint rendering was not exercised. Bezier control hulls remain intentionally conservative; the separate Node-only editor command-alignment helper is unchanged.

### Evidence

For a horizontal double elbow with `start=[0,0]`, `end=[100,20]`, `broken2=[-50,500]`, `left=top=100`:

```text
SVG path: M0,0 L-50,0 L-50,20 100,20
Bounds:   x=[50,200], y=[100,120]
PPTX:     origin=(476250,952500), extent=(1428750,190500) EMU
```

- [x] Local formatting, lint, and type checks passed
- [x] Local Chromium geometry and PPTX-output verification completed

## Checklist

- [x] Self-review and independent AI code review completed; no actionable findings remain
- [x] Documentation evaluated; no configuration or public API additions

AI-assisted implementation and review.


wyuc (approved):
Thanks @YizukiAme — this is a clean fix for #674, verified independently:

- The new bounds contain the whole visible path for straight, quadratic, cubic, elbow and double-elbow lines, including negative control points; exact Bezier extrema are still used for alignment guides.
- Double-elbow routing is unchanged and its unused control coordinate is excluded (pushing both coordinates turns 6 tests red).
- The real PPTX-output regression is solid: it unzips the slide XML and pins `a:off`, `a:ext`, the path size and every `a:pt` for four cases, including a minimum smaller than `left`/`top`, so lines keep their absolute position. Dropping the endpoint minima, the control hull or the offset translation each turns tests red.
- Version bumps pass the package check; lint, typecheck and `next build` pass.

Approving. One follow-up I'd like, as a separate PR: two older copies of the endpoint-only range still exist and now disagree with the shared geometry —

1. `packages/@openmaic/editor/src/core/index.ts` (the helper behind `element.align`, which you noted in the description): for a bowed curve, align-bottom can leave the visible path ~30px past the viewport edge.
2. `components/slide-renderer/Editor/Canvas/hooks/useDragElement.ts`: drag snapping and multi-drag union still use endpoint-only boxes.

Pointing both at the renderer's `getElementRange` would make the "one implementation" complete. Minor, optional: the PPTX test could assert `off + pt == original absolute point` directly to make the intent explicit.

PR #1632, opened 2026-09-21
feat(pi): expose interactive static instructions through read_scene

LING-6150:
## Summary

Students can ask about explicit interactive-courseware rules without selecting a component. For example, the word-sorting outline and runtime state omit the full wrong-placement rule: **−5 points and the word returns**. This PR supplies that source-authored text through the existing `read_scene → call_agent` path. Follows #1508.

## Changes

- Reuses bounded, non-executing HTML extraction; labels static instructions and source defaults separately from current-state evidence.
- Retains direct body text and separates table cells, options, and buttons for whole-page reading. Ordinary inline words/units and component-reference behavior are preserved.
- Keeps the 24k scene-evidence body budget. Oversized static text is omitted as a whole with an explicit unavailable note. Outline/scene context survives only if the fallback—including that note—fits; otherwise the read fails without partial evidence.
- No new tool or state protocol; slide, quiz, and PBL boundaries remain unchanged.

## Validation

- **Current `952a7188`: 6 related files, 131/131 offline tests passed**, including extraction, budget boundaries, component compatibility, and route-to-Teacher delegation. TypeScript, ESLint, Prettier, and diff checks passed.
- **Earlier `e78e6a6c`: 22 Pi files, 327/327 tests passed.** The full suite and real-model experiments were not rerun after the focused fixes.
- Historical MaaS evidence: four one-shot QAs, a two-turn buoyancy feedback loop, and one successful P10 action, subject to the provenance and timing limits below.

<details>
<summary>Experiment provenance</summary>

The real-model runs used `df4194f7` plus the patch later committed byte-identically as `df16d7e3..e78e6a6c`; relevant evidence/call-chain files were unchanged between those bases. These runs precede the fixes in `952a7188`.

QA cases 1/4 replayed historical browser observations with refreshed timestamps; cases 2/3 omitted state. All four independently followed `read_scene → call_agent → cue_user`. Requests, SSE answers, and tool traces were captured; full Teacher evidence was reconstructed offline, not captured from provider requests.

The final P10 run returned the original response immediately and observed a clone. The earlier buffered experiment cannot establish production SSE latency.

</details>

## Cross-review — Codex + Claude

Independent reviews of `e78e6a6c` found **no blocking source issues**. Claude identified non-blocking budget degradation and text-joining issues; Codex independently reproduced them against the source/base and implemented the scoped fixes in `952a7188`. Both reviews called for clearer QA provenance, now included above.

The fixes received targeted Codex verification (131 tests), **not a second full cross-review or a Claude re-review**. Dynamic freshness and take-once scene delegation remain existing boundaries.

## Known limitations

- **Transient timing:** P10 reported 11.4 seconds at send time, but only 3.8 remained when actionable guidance was observed in the UI. The action succeeded; “about 11 seconds / definitely enough time” was still unsupported delivery-time wording. This PR does not provide delivery-time resampling or guarantees.
- **CSS layout:** extraction does not infer spacing between visually separated ordinary spans; the buoyancy ticks can still read as `60010001400`.

## Demo

OpenMAIC-buoyancy-teaching-loop-4x

Two turns: Teacher guides the student from 1000 to 1400 kg/m³; after the operation, a fresh sample supports feedback of 4.116 N versus 2.94 N. This demonstrates combined teaching capability; the word-sorting rules establish the separate need for static-source reading.


wyuc (changes requested):
Thanks @LING-6150 — the feature is well scoped and most of it holds up under independent review:

- **Non-executing and bounded**: extraction goes through `linkedom/worker` + parse5 with no script execution, no `eval` and no network (a `fetch` spy and a `window` canary were never touched). `<script>`, `<style>`, `<iframe>`, `<noscript>`, `<template>`, etc. are dropped in both passes; hidden text is kept and explicitly labelled as authored. The limits (4M source, 10k nodes, 20k attributes, depth 256, 512k retained) hold under adversarial near-limit pages in ≤1.6 s, with no hang.
- **Budget and priority**: the 24k scene-evidence bound holds, and outline/state are kept ahead of static text when it doesn't fit (the last commit). Removing the fallback branch, inverting the priority, or leaking `<script>` text each turn tests red.
- Component-reference behaviour and the #1508 observation output are unchanged; the pi/interactive/observation suites (361), lint and typecheck pass.

Two things before merge, both small:

### 1. A parser error can escape the "source could not be safely read" fallback

parse5 throws `RangeError: Invalid code point` on two consecutive lone low surrogates (the minimal input is `<body><p>\udc00\udc00</p></body>`). `extractInteractiveStaticSourceText` (`element-reference.ts`, around L422 / L1068) lets it propagate, and the `instanceof ElementReferenceValidationError` guard in `read-scene.ts:57-63` doesn't catch it. So `execute` rejects, `onEvidence` never fires, and the model gets a raw parser message instead of the designed note. Model-generated HTML can plausibly contain broken surrogates; in a fuzz run, 578 of 600 random malformed inputs hit this. Wrapping any parse failure as `ElementReferenceValidationError` inside the extractor (or broadening that catch) fixes it; a regression test with the surrogate pair would pin it.

### 2. Static text can impersonate the evidence and state labels

The extracted text is inserted into the prompt raw, unfenced and unescaped (`read-scene.ts:66-72`), unlike the page-reported state JSON, which escapes `<` (`interactive-state-evidence.ts:240`). Because the extractor decodes entities, a page whose visible text is entity-encoded can produce literal `</page_reported_state>`, `PAGE-REPORTED STATE`, `Outline description: …` or `Static-source boundary: …` lines inside the same Teacher prompt, i.e. authored courseware can try to fake current state or a different outline. The courseware is model-generated, so this is untrusted input. Please fence the static block with a per-call nonce delimiter and escape (or neutralize) the delimiters and label keywords inside it, consistent with how the state evidence is handled; a test with entity-encoded label text would pin it.

### Non-blocking follow-ups

- Truncation is all-or-nothing: any total over 24k drops 100% of the static text, even with several KB free (a 5k rule plus 20k of base yields no rule). Also, for a static-bearing scene, the fallback note and the longer boundary add ~158 chars, so a base within ~158 chars of the cap now returns `too_large` where it read fine before. Trimming static text to the remaining budget would fix both.
- Extraction is repeated on every `read_scene` call (~0.85–1.3 s near the retained limit); the Director can call it several times per turn, so a per-scene memo would help.
- The static block ships regardless of `isCoursewareReferenceEnabled()`, but its wording points the model to "separately labeled page-reported state evidence", which doesn't exist when the flag is off. Consider softening that sentence when the flag is disabled.

LING-6150:
@wyuc Thanks for the detailed review. Addressed in 2a8b7557:

- Both HTML parsing passes now normalize ordinary parser exceptions into `ElementReferenceValidationError`, preserving the existing unavailable-source fallback, base evidence, and Teacher delegation.
- Static text is now JSON-encoded, with `<` and evidence-label keywords escaped, inside a fresh per-call UUID delimiter. Tests cover entity-decoded labels, copied delimiters, round-trip preservation, Teacher input, and the post-serialization budget.
- The state-boundary wording now makes current facts conditional on supporting state evidence, without implying that such evidence is always present.

Validation: 24 related test files / 361 tests passed, along with TypeScript, changed-file ESLint, Prettier, and `git diff --check`. All seven new regression tests failed before the fixes and passed afterward. No real-model calls were made; this validates evidence framing, not absolute resistance to prompt injection.

The two non-blocking optimizations remain deferred: partial static-text truncation and repeated-read caching. Oversized static text still degrades as a whole, and the availability note still counts toward the 24k budget, including for near-limit base evidence.

wyuc (changes requested):
Re-reviewed 2a8b7557. Both round-1 blockers are fixed. Parser exceptions (including lone surrogates) are normalized on both passes, and the per-call nonce fence plus escaping holds up against lookalike and closing-tag injection. Thanks!

One small blocker left, inside this PR's own goal of accurate evidence framing:

- **P2, `lib/chat/pi/tools/read-scene.ts` ~113-128.** When the static block exceeds `MAX_SCENE_EVIDENCE_CHARS` and is dropped, the fallback still appends the `featureBoundary` chosen while `staticSourceEvidence` was truthy: *"only outline/scene metadata and the separately labeled static-source result above are available"*. That sits right under the *"unavailable because the static text exceeds the scene evidence budget"* note, so the model gets told both things. Pick the boundary after the budget decision (use the "no static text available" wording, or a dedicated "withheld by budget" wording), and add a test for the over-budget case that asserts the boundary line.

Optional P3: the steps after parsing (component/step extraction) could share the same exception normalization, so a future parser change can't reopen the escape.

wyuc (approved):
Re-checked 28701d4b. The over-budget fallback now always ends with the "no Interactive source static text is available" boundary, and the new test asserts the stale "static-source result above" line is gone. The post-parse DOM/text extraction failures also stay on the `ElementReferenceValidationError` path, with a test that the raw error never reaches the evidence text. That covers the P3 too. Thanks @LING-6150!

LING-6150:
@wyuc Thanks for the follow-up review! Fixed in [28701d4b](https://github.com/THU-MAIC/OpenMAIC/commit/28701d4b425751725181a2d652af644485acb910).
- P2: The over-budget fallback now uses the “no static text available” boundary. Tests verify both the returned evidence and the Teacher prompt, while preserving the existing 24k budget checks.
- Optional P3: Extended the extractor’s existing exception normalization to include DOM cleanup and text extraction. Fault-injection tests verify that failures preserve base evidence without leaking raw errors.
Related tests pass 363/363, along with typecheck, lint, and formatting checks. Independent Claude and Codex reviews of this increment found no actionable issues. The PR description is updated.

PR #1633, opened 2026-09-21
fix(audio): prime discussion media element on start gesture

### LINKED ISSUE #1611: [Bug]: Classroom discussion narration goes silent from the second line on mobile (per-line new Audio element refused by autoplay policy)

Frank-zhu0404:
> 🤖 **AI-assisted PR** — written with an AI coding agent (Grok Build) and reviewed by another Grok Build session (`VERDICT=PASS`) before opening. I understand the change and stand behind it.

## Summary

Follow-up to #1610: the shared discussion media element was still first touched **after** the TTS round-trip, so a strict per-element autoplay policy could refuse even the first `play()`. This primes that element with a muted silent clip **synchronously inside the user gesture** that starts discussion (Join / send / interrupt / resume), then pauses and releases the unlock source so later lines can load on the same unlocked element.

## Related Issues

Closes #1611

Related to #1610, #1474, #1477

## Changes

- `lib/audio/discussion-audio.ts`: add `primeDiscussionAudioElement()` (muted silent WAV `play()` + pause + `releaseDiscussionAudioLine`; no-op when a real line is already loaded)
- `lib/hooks/use-discussion-tts.ts`: expose `prime()`
- `components/edit/PlaybackChromeRoot.tsx`: call `discussionTTS.prime()` on Join, chat send/interrupt, topic resume, and discussion resume — before any `await`
- Tests: policy stub requires in-gesture prime; unit coverage for idle unlock + skip-when-loaded; chrome-root call-order checks

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Unit: `pnpm exec vitest run tests/hooks/use-discussion-tts.test.ts tests/audio/discussion-audio-prime.test.ts tests/components/edit/playback-chrome-root-element-reference-owner.test.ts` → 46 passed
2. Local preflight: `pnpm format`, `pnpm lint --fix` (0 errors), `pnpm exec tsc --noEmit`
3. Manual (device): start a classroom discussion on mobile Safari / Chrome; confirm line 2+ remain audible after Join

- Policy tests refuse the first line when `prime()` never ran in the gesture window, and pass when it did; a no-op `prime` turns the success case red
- Priming leaves the element idle (`src` cleared, muted restored, paused)
- Existing single-element reuse from #1610 preserved
- **Not verified on a physical iOS/Android device in this pass** (same honest caveat as the original report) — unit + policy stubs + review PASS are the gate used here

### Evidence

- [x] CI checks run locally (`pnpm format` / `pnpm lint --fix` / `npx tsc --noEmit`)
- [x] Targeted vitest: 46 passed
- [ ] Screenshots / recordings attached (if UI changes) — N/A (audio unlock path; no UI chrome change)

## Checklist


LING-6150 (commented):
@wyuc, this implements the gesture-time priming follow-up you mentioned in #1611. The focused tests pass, but there is still no real-device confirmation that discussion audio fails after #1610 or that this patch resolves it.
My inclination is to hold approval pending that verification. Would you prefer to wait for mobile evidence, or consider this as a defensive change in its own right?

wyuc (changes requested):
Thanks @Frank-zhu0404 — priming the shared element inside the start gesture is the right idea, and the core wiring checks out: `discussionTTS.prime()` runs synchronously as the first statement of the start handlers (before any `await`), playback reuses the same module-scoped element, the priming clip is muted, and a rejected `play()` is caught. Moving the prime after an `await`, or creating a new element for playback, turns the new tests red.

Two things before merge, plus device verification, since that's why #1611 was reopened after #1610:

### 1. The unlock sequence may cancel itself on iOS

`lib/audio/discussion-audio.ts:56-67` starts `play()` and then synchronously `pause()`s, removes `src` and calls `load()`. On iOS Safari, `load()` resets the element to `HAVE_NOTHING`, and an immediate pause can abort the pending play promise, so the element may never be recorded as user-activated. The unlock source is also a zero-sample WAV (`dataSize = 0`, `:27-28`), which may be treated as undecodable. The possible result: an iOS user taps Join, nothing is heard (as intended), and the first real line still rejects with `NotAllowedError`. Suggested hardening: use a short silent WAV with actual samples, and let the unlock `play()` resolve before pausing (or assign the first real clip over the silent one instead of `load()`-resetting the element).

### 2. A voice-first interaction primes outside the gesture

`components/roundtable/index.tsx:392` forwards `onMessageSend` from `onTranscription`, which `lib/hooks/use-audio-recorder.ts:73-84` fires after `await fetch('/api/transcription')`, so `prime()` runs without user activation. If an iOS user's first interaction is a spoken question, the first agent answer is silent. Priming in the click that starts recording (`handleToggleVoice` / `startRecording`, around `:427-440`) as well would cover it.

### Device verification

jsdom can't exercise autoplay policy, so before we merge (and before #1611 is closed), please verify on real devices and attach a short screen recording or notes for each:

- **iOS Safari**, **an iOS in-app WebView** (e.g. WeChat) and **Android Chrome**;
- starting by button and starting by voice;
- first round, second round, and after backgrounding the page and returning.

### Non-blocking

- Dropping the `play()` rejection handler (`discussion-audio.ts:60`) fails no test, because the rejecting fake is a `vi.fn(async …)` whose rejection vitest doesn't surface. A non-mock thenable that records `catch` would pin it.
- `if (audio.src) return` (`:50`) can spend a gesture without unlocking when a previously refused line left `src` set; tracking an "unlocked" flag in the module would be more robust.
- There's no re-prime after `visibilitychange`, which ties into the backgrounding case above.

PR #1634, opened 2026-09-21
fix(generation): reject unusable interactive scripts and surface runtime errors

### LINKED ISSUE #1622: [Bug]: Generated interactive scenes can render but remain unusable after script/runtime failures
## Summary

Generated interactive scenes can render correctly while being unusable because the generated JavaScript fails either:

1. **before execution** (syntax / parse-time failure), or
2. **during interaction** (runtime exception in generated state/data logic).

The current iframe runtime already captures runtime errors, but a broken interactive scene can still reach the user and remain effectively dead.

This is broader than the syntax-only case previously tracked in #1582 / #1583.

## Real reproductions

I reproduced two independent real generated artifacts in a sandboxed browser path.

### Case A: Flink realtime simulator — parse-time failure

The page visually renders a complete Flink simulation UI with sliders, Watermark / Checkpoint indicators, and a **Start Simulation** button.

The generated main script contains:

```js
state counts = new Array(10).fill(0);
```

Browser result:

```text
SyntaxError: Unexpected identifier 'counts'
```

Because the entire classic script fails to parse, `handleMainButton()` is never defined. Clicking the visible button then produces:

```text
ReferenceError: handleMainButton is not defined
```

User-facing symptom: the page looks complete, but clicking Start Simulation does nothing.

### Case B: data-pipeline game — runtime initialization failure

A separate generated game has syntactically valid JavaScript and renders normally.

After clicking **Start Challenge**, generated data references `_REST API_`, while the component registry contains `REST API`. The lookup returns `undefined`, then generated code reads `comp.name` and throws:

```text
TypeError: Cannot read properties of undefined (reading 'name')
```

The interaction starts, partially initializes, then aborts.

## Why this matters

These two artifacts have the same user-facing failure mode but different causes:

```text
render succeeds
  -> control is visible
  -> generated JS is unusable
  -> interaction appears dead / partially dead
```

A syntax-only fix addresses Case A but not Case B.

## Current behavior

- OpenMAIC already captures iframe runtime errors into the scene runtime error store.
- Current main does not appear to consume those errors for recovery.
- #1532 already fixed the separate iframe-ready message queue race, so this issue is not intended to duplicate that work.
- #1053 / #1055 cover deterministic replay / runtime visibility contracts and are related, but this report is specifically about preventing or recovering generated interactive scenes that are already known to be broken.

## Proposed reliability boundary

I locally validated a two-layer approach:

### Layer 1 — generation-time classic inline JS syntax preflight

Before persisting an interactive widget:

- compile-check classic inline `<script>` bodies without executing them;
- skip data scripts such as `application/json`;
- skip external `src` scripts;
- skip `type="module"` scripts;
- route failures through the existing `invalid-model-output` retry path.

The real Flink artifact is rejected before persistence with:

```text
script #9: Unexpected identifier 'counts'
```

### Layer 2 — runtime-error-informed targeted repair

For syntax-valid pages that fail during interaction:

- reuse the existing iframe runtime-error capture;
- provide the current HTML plus concrete fatal runtime errors as repair context;
- regenerate only the interactive HTML;
- keep other scene data unchanged;
- run the repaired output through the same syntax preflight.

With the full runtime-broken game artifact, a local targeted repair produced a page that:
- passed syntax preflight;
- successfully started after clicking;
- rendered all expected component cards;
- emitted no page/runtime errors.

I intentionally kept the repair trigger user-initiated during local validation so normal playback does not automatically spend model quota.

## Validation performed locally

Relevant regression coverage passed:

- iframe / picker / keep-alive / sandbox / repair-path tests: **42/42**
- generation tests excluding the existing Windows-only assets path assertion: **149/149**
- selected PBL / quiz / vocational routing regressions: **39/39**
- generation package build: pass
- generation package typecheck: pass
- root `tsc --noEmit`: pass
- focused ESLint: pass
- `git diff --check`: pass

Syntax preflight cost on the two real artifacts was sub-millisecond (~0.07–0.20 ms per validation).

## Architecture question

The generation-time syntax preflight is a narrow bug-fix boundary.

For runtime recovery, I would like maintainer guidance on the preferred integration point:

- a lightweight runtime failure affordance near the interactive renderer, or
- the existing MAIC Editor / agent editing path (e.g. the current interactive HTML editing plumbing).

I do not want to hard-code the UX/ownership before alignment, especially with #646 evolving the editor/agent architecture.

## Related

- #1582 — previous syntax-only issue, closed after discovering the broader runtime failure class
- #1583 — previous syntax-only PR, withdrawn for the same reason
- #1532 — merged iframe-ready message queue fix; separate failure mode
- #1053 — deterministic interactive replay / contract inspector
- #1055 — runtime visibility contract consumption
- #646 — MAIC Editor / agent architecture

## Expected behavior

A generated interactive scene that is known to be invalid should not be persisted as an apparently functional widget.

If a syntactically valid scene fails at runtime, the captured error should be surfaced into an explicit recovery path rather than leaving a visually complete but unusable interaction.

Frank-zhu0404:
## Summary

Reject classic inline interactive scripts that fail to parse at generation time (so Flink-like `state counts = ...` is not persisted as a dead widget), and surface iframe-captured runtime errors on the active interactive scene instead of leaving a silent, unusable control.

No heuristic JS rewriting, no generation-pipeline smoke runner, and no Fix-with-agent button (workbench `read_stage` is server-side and cannot see the client error store; no existing draft-composer API).

AI-assisted: yes (Grokbuild grok-4.7 xhigh fix + review).

## Related Issues

Closes #1622

## Changes

- Add `findInteractiveScriptSyntaxFailure` in `@openmaic/generation` — compile-check classic inline `<script>` bodies via `Function(source)` (parse only); skip JSON/data, external `src`, and `type="module"` / importmap
- Wire the check into `generateWidgetContent` so failures become `invalid-model-output` before persist / post-process
- Show an i18n runtime-error banner on the active interactive iframe when `useSceneRuntimeErrors` has entries; dismiss hides without clearing the store; `srcDoc` change still `clearScene`
- Bump `@openmaic/generation` 0.3.11 → 0.3.12
- Tests for Flink-like rejection, skip cases, scene-content failure path, and banner/dismiss/srcDoc behavior

## Type of Change

- [ ] Documentation update
- [ ] CI/CD or build changes

### Steps to reproduce / test

1. Generation: produce interactive HTML whose classic inline script contains `state counts = new Array(10).fill(0);` — content generation should report `invalid-model-output` and not persist the widget.
2. Runtime: load a scene whose iframe posts a `runtime-error` — the active interactive chrome should show “This interactive failed to run” with the error text; dismiss hides; changing content clears.
3. Confirm JSON config scripts, external `src`, and `type="module"` are not false-rejected by the preflight.

- `pnpm --filter @openmaic/generation test` → 154/154
- Focused vitest: interactive-runtime-error-banner + scene-runtime-errors → 8/8
- Package typecheck/build and root `tsc --noEmit` (in fix session)
- i18n key parity across locales
- Grokbuild review VERDICT=PASS (0 blocking; 2 non-blocking suggestions on regex attribute edge cases and oldest-vs-newest banner text)
- Live classroom / Playwright e2e not run here (Chrome DevTools / file-watch limits on the box)

### Evidence

- [x] CI passes (`pnpm check && pnpm lint && npx tsc --noEmit`) — local format/lint/tsc in fix session; remote CI will confirm on the PR
- [x] Manually tested locally — unit/package tests above
- [ ] Screenshots / recordings attached (if UI changes) — banner covered by unit tests; no visual screenshot attached

## Checklist


LING-6150 (changes requested):
Thanks ,  the generation-time validation and runtime-error visibility are useful improvements. I found a blocking false rejection in the new script validator.
The regex does not respect quoted attribute values or HTML comments. For example:
<script data-note="a > b">window.widgetRan = true;</script>

<script>window.widgetRan = true;</script>
Both contain valid executable scripts, but the new validator rejects them. I independently reran the generateSceneContent() comparison: base returns content for both; this head returns null with invalid-model-output.
Please extract executable script bodies according to HTML parsing rules and add regressions for these cases. Also, new Function() accepts function-body syntax such as top-level return;, which is invalid in a classic script; please align the syntax check with the intended script grammar.
Separately, please clarify the Closes #1622 wording: this PR adds syntax validation and runtime-error visibility, while the recovery portion remains unimplemented. That can be tracked separately without expanding this patch.

